Issue #45, opened 2023-04-20
Research/Fix ACID failure in storage for writes


hexfusion:
As per #44 we have a possible ACID failure, this issue tracks its resolution.

rkuris:
Fixed with new design

PR #57, opened 2023-04-25
Attempt at fixing #45

rkuris:
Looks like the file being checked isn't the same as the file being created so the test failed. This file does actually exist, but for some reason the test was still just using /tmp for storing the database. Updated the test and let's see if this passes now.


rkuris:
This PR is quite out of date, and the actual problem here is that we aren't waiting for the disk_requester to finish it's work before checking the new state, which makes the test flaky. I'm going to fix this differently.

PR #227, opened 2023-08-28
1/3 Implement improved API implementation for database

rkuris:
One step along the way, broken up for easier review. Changes include:
 - Implements api::DbView for DbRev
 - The API currently returns references; this is challenging with the way shale works, so we're (temporarily) going to return owned values from some APIs


exdx (approved):

firewood/src/db.rs
```diff
@@ -272,6 +273,43 @@ pub struct DbRev<S> {
...
+            .map_err(|e| api::Error::IO(std::io::Error::new(ErrorKind::Other, e)))
+    }
+
+    async fn val<K: api::KeyType>(&self, key: K) -> Result<Option<Vec<u8>>, api::Error> {
+        let obj_ref = self.merkle.get(key, self.header.kv_root);
+        match obj_ref {
+            Err(e) => Err(api::Error::IO(std::io::Error::new(ErrorKind::Other, e))),
+            Ok(obj) => match obj {
+                None => Ok(None),
+                Some(inner) => Ok(Some(inner.as_ref().to_owned())),
```
xinifinity, line 291: nit: Naming is hard.., we discussed to no longer use `inner` as it didn't bring any clarification?

firewood/src/db.rs
```diff
@@ -272,6 +273,43 @@ pub struct DbRev<S> {
...
+            },
+        }
+    }
+
+    async fn single_key_proof<K: api::KeyType, N: AsRef<[u8]> + Send>(
+        &self,
+        _key: K,
+    ) -> Result<Option<Proof<N>>, api::Error> {
+        todo!()
+    }
```
xinifinity, line 301: If these are `todo` maybe we can do it in a follow up with real implementations?

firewood/src/db.rs
richardpringle: I think we need to stop using the `inner` pattern. Naming something `inner` when we've simply added a wrapper type is fine IMO

firewood/src/db.rs
```diff
@@ -272,6 +273,43 @@ pub struct DbRev<S> {
...
+#[async_trait]
+impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
+    async fn root_hash(&self) -> Result<api::HashKey, api::Error> {
+        self.merkle
+            .root_hash(self.header.kv_root)
+            .map(|h| *h)
+            .map_err(|e| api::Error::IO(std::io::Error::new(ErrorKind::Other, e)))
+    }
+
+    async fn val<K: api::KeyType>(&self, key: K) -> Result<Option<Vec<u8>>, api::Error> {
```
richardpringle, line 285: It's infuriating that you *had* to change this.

richardpringle (approved):

firewood/src/db.rs
rkuris: Yeah. This is one of those sweep tasks we can assign as a good first issue. Anyone want to create one?

firewood/src/db.rs
rkuris: That's why this is 1/3 :)

firewood/src/db.rs
richardpringle: I don't think it's trivial to unravel all the ` SomeType { inner: SomeInnerType }` especially with all the misused trait-bounds

PR #287, opened 2023-09-27
Refactor `buffer.rs` for better readability and maintainability

gubatron:
- Introduced `create_state_cache` utility function in tests module to reduce repetition.
- Refactored test functions to use the new utility function.
- ~Adjusted indentations for async operations and error handling. (rustfmt)~

### DIFF
--- a/firewood/src/storage/buffer.rs
+++ b/firewood/src/storage/buffer.rs
@@ -657,6 +657,7 @@ mod tests {
             disk_requester,
         )
         .unwrap()
+        .into()
     }
 
     #[test]
@@ -828,7 +829,7 @@ mod tests {
         disk_requester.init_wal("wal", &root_db_path);
 
         // create a new state cache which tracks on disk state.
-        let state_cache = CachedSpace::new(
+        let state_cache: Arc<CachedSpace> = CachedSpace::new(
             &StoreConfig::builder()
                 .ncached_pages(1)
                 .ncached_files(1)
@@ -838,7 +839,8 @@ mod tests {
                 .build(),
             disk_requester.clone(),
         )
-        .unwrap();
+        .unwrap()
+        .into();
 
         // add an in memory cached space. this will allow us to write to the
         // disk buffer then later persist the change to disk.
--- a/firewood/src/storage/buffer.rs
+++ b/firewood/src/storage/buffer.rs
@@ -646,20 +646,17 @@ mod tests {
         state_path: PathBuf,
         disk_requester: DiskBufferRequester,
     ) -> Arc<CachedSpace> {
-        // create a new state cache which tracks on disk state.
-        Arc::new(
-            CachedSpace::new(
-                &StoreConfig::builder()
-                    .ncached_pages(1)
-                    .ncached_files(1)
-                    .space_id(STATE_SPACE)
-                    .file_nbit(1)
-                    .rootdir(state_path)
-                    .build(),
-                disk_requester,
-            )
-            .unwrap(),
+        CachedSpace::new(
+            &StoreConfig::builder()
+                .ncached_pages(1)
+                .ncached_files(1)
+                .space_id(STATE_SPACE)
+                .file_nbit(1)
+                .rootdir(state_path)
+                .build(),
+            disk_requester,
         )
+        .unwrap()
     }
 
     #[test]
@@ -757,6 +754,7 @@ mod tests {
         // create a new wal directory on top of root_db_fd
         disk_requester.init_wal("wal", &root_db_path);
 
+        // create a new state cache which tracks on disk state.
         let state_cache = create_state_cache(state_path, disk_requester.clone());
 
         // add an in memory cached space. this will allow us to write to the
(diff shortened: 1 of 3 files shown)


firewood/src/storage/buffer.rs
```diff
@@ -642,6 +642,24 @@ mod tests {
             .join("firewood")
     }
 
+    fn create_state_cache(state_path: PathBuf, disk_requester: DiskBufferRequester) -> Arc<CachedSpace> {
```
rkuris, line 645: Every single call to `CachedSpace::new` is wrapped by `Arc::new`, so please make `CachedSpace::new` return an `Arc`. For these one-offs that do not use the configuration, perhaps `CachedSpace::new_small()` which only takes the arguments listed here makes sense.

rkuris (changes requested):

firewood/src/storage/buffer.rs
```diff
@@ -484,8 +484,8 @@ async fn process(
                 wal_cfg.max_revisions,
                 final_path.as_path(),
             )
-            .await
-            .unwrap_or_else(panic_on_intialization_failure_with(&rootpath, &waldir));
+                .await
```
rkuris, line 487: All the code in this repo is formatted using cargo fmt on the latest stable. We have some checks to ensure that it stays that way.

Not sure what you're doing to get this alternate indentation; perhaps not the latest version of rustfmt?

gubatron: Thanks, it seems my intellij was using its own internal formatter and not `rustfmt` as I thought

I've also updated my rustfmt to the latest.

```
$ rustfmt --version
rustfmt 1.6.0-stable (d5c2e9c3 2023-09-13)
```

Updating pull request with all the changes requested in a few mins...

firewood/src/storage/buffer.rs
gubatron: Dear Ron, if I understand correctly, this will make the scope of this PR a bit bigger, _love it_.

I've checked every call to `CachedSpace::new` and you're right, it's always wrapped by `Arc::new`.

>For these one-offs that do not use the configuration
I couldn't find any instance of `CachedSpace::new` that didn't receive a `cfg: &StoreConfig` parameter, perhaps I don't understand what you mean on that one.

For now I'll go ahead and try to make `CachedSpace::new` return an `Arc` wrapped result on a separate commit that will touch the following files:
 - `firewood/src/db.rs`
 - `firewood/src/storage/buffer.rs`

gubatron: [done deal](https://github.com/ava-labs/firewood/pull/287/commits/c97d979a1774a9c98bcb1b3cc6630676e390b267) with the first request Ron.

firewood/src/storage/mod.rs
```diff
@@ -723,10 +723,10 @@ impl CachedSpace {
     pub fn new(
         cfg: &StoreConfig,
         disk_requester: DiskBufferRequester,
-    ) -> Result<Self, StoreError<std::io::Error>> {
+    ) -> Result<Arc<Self>, StoreError<std::io::Error>> {
```
richardpringle, line 726: @rkuris, while not every library does this, I think it's a best practice to follow the standard-library convention of `new` returning `Self` and `try_new` returning `Result<Self, _>`. 

[`Arc<T>` implements `From<T>`](https://doc.rust-lang.org/stable/std/sync/struct.Arc.html#impl-From%3CT%3E-for-Arc%3CT%3E), so if you want to shortcut for `Arc::new`, you can simply call `.into()` on a `CachedSpace` at the call-site. 

@gubatron, could you please change the return type of this function back to what it was? And, if you feel so inclined, feel free to change the method name to `try_new`. 

Thanks again for the contributions!

richardpringle (changes requested):

firewood/src/storage/buffer.rs
```diff
@@ -642,6 +642,23 @@ mod tests {
             .join("firewood")
     }
 
+    fn create_state_cache(
```
rkuris, line 645: nit: rename: `new_cached_space_for_test` would be better.

Nit's do not require fixing, so I'm approving at this point.

rkuris (approved):

firewood/src/storage/buffer.rs
rkuris: @richardpringle 's comment about .into() is much better!

firewood/src/storage/mod.rs
gubatron: I agree, it was cool but kind of weird.
Love the more "Rustacean" approach of implementing it with the `From<T>` trait.

gubatron: Rust allows us to do this in several ways, I wonder which one you like better:

**Annotating the receiving end and using into**
```rust
let state_cache: Arc<CachedSpace> = CachedSpace::new(
            &StoreConfig::builder()
                .ncached_pages(1)
                .ncached_files(1)
                .space_id(STATE_SPACE)
                .file_nbit(1)
                .rootdir(state_path)
                .build(),
            disk_requester.clone(),
        )
        .unwrap()
        .into();
```

**not annotating the declared variable and using  `.into() as Arc<CachedSpace>`**
```rust
let state_cache = CachedSpace::new(
            &StoreConfig::builder()
                .ncached_pages(1)
                .ncached_files(1)
                .space_id(STATE_SPACE)
                .file_nbit(1)
                .rootdir(state_path)
                .build(),
            disk_requester.clone(),
        )
        .unwrap()
        .into() as Arc<CachedSpace>;
```

**or Using `Arc::from`**
```rust
        let state_cache = Arc::from(
            CachedSpace::new(
                &StoreConfig::builder()
                    .ncached_pages(1)
                    .ncached_files(1)
                    .space_id(STATE_SPACE)
                    .file_nbit(1)
                    .rootdir(state_path)
                    .build(),
                disk_requester.clone(),
            )
            .unwrap(),
        );
```

gubatron: I guess it's not a one fit all solution, using a combination of them

richardpringle (approved):

richardpringle (approved):

richardpringle:
Thanks again @gubatron!

gubatron:
Thank you, y'all seem like a really cool team, my pleasure and honor to contribute a small grain of sand.

rkuris:
> Thank you, y'all seem like a really cool team, my pleasure and honor to contribute a small grain of sand.

Plenty more sand needs sweeping, and you gotta start somewhere! Thanks again for doing this.

PR #289, opened 2023-09-28
Remove 'static lifetime on key-type trait

richardpringle:
You should be able to use a borrowed key for lookups

### DIFF
--- a/firewood/src/v2/api.rs
+++ b/firewood/src/v2/api.rs
@@ -8,9 +8,9 @@ use async_trait::async_trait;
 /// A `KeyType` is something that can be xcast to a u8 reference,
 /// and can be sent and shared across threads. References with
 /// lifetimes are not allowed (hence 'static)
-pub trait KeyType: AsRef<[u8]> + Send + Sync + Debug + 'static {}
+pub trait KeyType: AsRef<[u8]> + Send + Sync + Debug {}
 
-impl<T> KeyType for T where T: AsRef<[u8]> + Send + Sync + Debug + 'static {}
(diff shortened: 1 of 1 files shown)


firewood/src/v2/api.rs
```diff
@@ -8,9 +8,9 @@ use async_trait::async_trait;
 /// A `KeyType` is something that can be xcast to a u8 reference,
 /// and can be sent and shared across threads. References with
```
richardpringle, line 9: Why do keys need to be shared across threads, futures?

firewood/src/v2/api.rs
```diff
@@ -8,9 +8,9 @@ use async_trait::async_trait;
...
 /// lifetimes are not allowed (hence 'static)
-pub trait KeyType: AsRef<[u8]> + Send + Sync + Debug + 'static {}
+pub trait KeyType: AsRef<[u8]> + Send + Sync + Debug {}
```
rkuris, line 11: This is going to break the ownership of keys in a number of places. Maybe instead rename this to OwnedKey and add:
```
pub trait Key: AsRef<[u8]> + Send + Sync + Debug {}
```

Sprinkle `Key` in where it's clear we should be borrowing.

rkuris (changes requested):

firewood/src/v2/api.rs
richardpringle: It's not breaking anything. `KeyType + 'static` is a subset of `KeyType` and should be added where necessary, not the other way around.

rkuris: OK that makes sense. Do we need to add any `'static`s anywhere?

rkuris (approved):

PR #291, opened 2023-09-28
Untangling lifetimes and borrows

richardpringle:
- Don't cast lifetimes
- Stop passing in a part of self as a separate arg
- Untangling lifetimes and borrows

It's a miracle that this compiled...

### DIFF
--- a/shale/src/compact.rs
+++ b/shale/src/compact.rs
@@ -1,6 +1,8 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
+use crate::ObjCache;
+
 use super::disk_address::DiskAddress;
 use super::{CachedStore, Obj, ObjRef, ShaleError, ShaleStore, Storable, StoredView};
 use std::fmt::Debug;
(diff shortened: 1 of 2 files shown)


shale/src/compact.rs
```diff
@@ -654,14 +655,16 @@ impl<T: Storable + Send + Sync + Debug + 'static, M: CachedStore + Send + Sync>
...
 
-        Ok(inner.obj_cache.put(inner.get_data_ref(ptr, payload_size)?))
+        Ok(ObjRef::new(Some(obj), cache))
     }
 
     fn flush_dirty(&self) -> Option<()> {
         let mut inner = self.inner.write().unwrap();
         inner.header.flush_dirty();
-        inner.obj_cache.flush_dirty()
+        self.obj_cache.flush_dirty()
```
rkuris, line 667: The guard now lives too long. Should be:
```
self.inner.write().unwrap().header.flush_dirty()
self.obj_cache.flush_dirty()
```

shale/src/lib.rs
```diff
@@ -190,17 +189,12 @@ impl<T: ?Sized + Send + Sync> Deref for Obj<T> {
 /// User handle that offers read & write access to the stored [ShaleStore] item.
 pub struct ObjRef<'a, T: Send + Sync> {
     inner: Option<Obj<T>>,
-    cache: ObjCache<T>,
-    _life: PhantomData<&'a mut ()>,
```
rkuris, line 194: 🏆

rkuris (approved):
Please fix the guard issue before landing

xinifinity (approved):

shale/src/compact.rs
richardpringle, line 667:
```suggestion
        // hold the write lock to ensure that both cache and header are flushed in-sync
        self.obj_cache.flush_dirty()
```

richardpringle: @rkuris, I added a comment here to keep these actions in-sync. Let me know if you still think I should change this

PR #292, opened 2023-09-28
2/3 Implement v2::api for the real DB

rkuris:
Changes:
 - Implement api::Db for Db 😄 
 - Change proposals Batch and BatchOp to use the new v2 ones
 - Rename propose() and commit() in Proposal to propose_sync and commit_sync

Still remaining (part 3)
 - Implement commit
 - Implement range proofs

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -389,20 +389,25 @@ impl api::Db for Db {
 
     type Proposal = Proposal;
 
-    async fn revision(&self, _hash: HashKey) -> Result<Arc<Self::Historical>, api::Error> {
-        todo!()
+    async fn revision(&self, root_hash: HashKey) -> Result<Arc<Self::Historical>, api::Error> {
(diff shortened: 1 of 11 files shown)


firewood/src/db.rs
```diff
@@ -380,6 +383,34 @@ impl Drop for DbInner {
...
 
+#[async_trait]
+impl api::Db for Db {
+    type Historical = DbRev<SharedStore>;
+
+    type Proposal = Proposal;
+
+    async fn revision(&self, root_hash: HashKey) -> Result<Arc<Self::Historical>, api::Error> {
+        if let Some(rev) = self.get_revision(&TrieHash(root_hash)) {
+            Ok(Arc::new(rev.rev))
```
richardpringle, line 394: I think this is correct for this PR, but it seems like we've got some more refactoring to do at some point. All these nested types drive me bananas.

richardpringle (approved):

PR #295, opened 2023-09-28
chore: use `decode` in single key proof verification

xinifinity:
One of the 'pluggable node encoding/decoding' series.

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -310,6 +310,10 @@ impl ExtNode {
         self.1
     }
 
+    pub fn chd_encoded(&self) -> &Option<Vec<u8>> {
+        &self.2
+    }
+
     pub fn chd_mut(&mut self) -> &mut DiskAddress {
         &mut self.1
     }
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -310,8 +310,8 @@ impl ExtNode {
         self.1
     }
 
-    pub fn chd_encoded(&self) -> Option<&Vec<u8>> {
-        self.2.as_ref()
+    pub fn chd_encoded(&self) -> Option<&[u8]> {
+        self.2.as_deref()
     }
 
     pub fn chd_mut(&mut self) -> &mut DiskAddress {
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -310,8 +310,8 @@ impl ExtNode {
         self.1
     }
 
-    pub fn chd_encoded(&self) -> &Option<Vec<u8>> {
-        &self.2
+    pub fn chd_encoded(&self) -> Option<&Vec<u8>> {
+        self.2.as_ref()
     }
 
     pub fn chd_mut(&mut self) -> &mut DiskAddress {
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -148,23 +148,10 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
         mut key_nibbles: NibblesIterator<'a, 0>,
         encoded_node: &[u8],
     ) -> Result<(Option<SubProof>, NibblesIterator<'a, 0>), ProofError> {
-        let items: Vec<Encoded<Vec<u8>>> = bincode::DefaultOptions::new()
-            .deserialize(encoded_node)
-            .map_err(ProofError::DecodeError)?;
-
-        match items.len() {
-            EXT_NODE_SIZE => {
-                let mut items = items.into_iter();
-                let decoded_key: Vec<u8> = items.next().unwrap().decode()?;
-
-                let decoded_key_nibbles = Nibbles::<0>::new(&decoded_key);
-
-                let (cur_key_path, term) =
-                    PartialPath::from_nibbles(decoded_key_nibbles.into_iter());
-                let cur_key = cur_key_path.into_inner();
-
-                let data: Vec<u8> = items.next().unwrap().decode()?;
-
+        let node = NodeType::decode(encoded_node)?;
+        match node {
+            NodeType::Leaf(n) => {
+                let cur_key = &n.path().0;
                 // Check if the key of current node match with the given key
                 // and consume the current-key portion of the nibbles-iterator
                 let does_not_match = key_nibbles.size_hint().0 < cur_key.len()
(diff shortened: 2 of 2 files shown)


firewood/src/proof.rs
```diff
@@ -148,23 +148,30 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
     ) -> Result<(Option<SubProof>, NibblesIterator<'a, 0>), ProofError> {
-        let items: Vec<Encoded<Vec<u8>>> = bincode::DefaultOptions::new()
-            .deserialize(encoded_node)
-            .map_err(ProofError::DecodeError)?;
-
-        match items.len() {
-            EXT_NODE_SIZE => {
-                let mut items = items.into_iter();
-                let decoded_key: Vec<u8> = items.next().unwrap().decode()?;
+        let node = NodeType::decode(encoded_node)?;
```
richardpringle, line 151: Remove the nesting! `locate_subproof` is only called in one place. Change the signature as follows:
```rust
    fn locate_subproof<'a>(
        &self,
        mut key_nibbles: NibblesIterator<'a, 0>,
        node: NodeType,
    ) -> Result<(Option<SubProof>, NibblesIterator<'a, 0>), ProofError> {
```

richardpringle (changes requested):

firewood/src/merkle/node.rs
```diff
@@ -310,6 +310,10 @@ impl ExtNode {
         self.1
     }
 
+    pub fn chd_encoded(&self) -> &Option<Vec<u8>> {
+        &self.2
```
richardpringle, line 314: You should never return references to `Option`. Instead, return `Option<&[u8]>` 
```suggestion
    pub fn chd_encoded(&self) -> Option<&[u8]> {
        self.2.as_ref()
```

richardpringle (changes requested):

firewood/src/merkle/node.rs
```diff
@@ -310,6 +310,10 @@ impl ExtNode {
         self.1
     }
 
+    pub fn chd_encoded(&self) -> Option<&Vec<u8>> {
+        self.2.as_ref()
```
richardpringle, line 314: When it comes to eking out every last drop of performance, it's better to return a pointer with an address and a length rather than return a pointer to another pointer. Just like you should never have a reference to an `Option`, you should probably also never have an _immutable_ reference to a `Vec`. 
```suggestion
    pub fn chd_encoded(&self) -> Option<&[u8]> {
        self.2.as_deref()
```

firewood/src/proof.rs
```diff
@@ -173,34 +179,24 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+                let data = n.chd_encode()[index]
+                    .as_ref()
+                    .ok_or(ProofError::InvalidData)?;
+                self.generate_subproof(data.to_vec())
                     .map(|subproof| (Some(subproof), key_nibbles))
             }
-
-            size => Err(ProofError::DecodeError(Box::new(
-                bincode::ErrorKind::Custom(format!("invalid size: {size}")),
-            ))),
```
richardpringle, line 203: 🥇

firewood/src/proof.rs
```diff
@@ -173,34 +179,24 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                 let index = key_nibbles.next().unwrap() as usize;
-
                 // consume items returning the item at index
-                let data: Vec<u8> = items.into_iter().nth(index).unwrap().decode()?;
-
-                self.generate_subproof(data)
+                let data = n.chd_encode()[index]
+                    .as_ref()
+                    .ok_or(ProofError::InvalidData)?;
+                self.generate_subproof(data.to_vec())
```
richardpringle, line 197: nit:
```suggestion
                    .ok_or(ProofError::InvalidData)?
                    .to_vec();
                self.generate_subproof(data)
```

firewood/src/proof.rs
```diff
@@ -173,34 +179,24 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                 Ok((sub_proof.into(), key_nibbles))
             }
-
-            BRANCH_NODE_SIZE if key_nibbles.size_hint().0 == 0 => Err(ProofError::NoSuchNode),
-
-            BRANCH_NODE_SIZE => {
+            NodeType::Branch(n) => {
+                if key_nibbles.size_hint().0 == 0 {
+                    return Err(ProofError::NoSuchNode);
+                }
```
richardpringle, line 191: Nit:
Every time you have to add more indentation to your code, you're adding complexity for a reader/review. I think it's much better to keep your methods/functions as flat as possible.
```suggestion

            NodeType::Branch(_) if key_nibbles.size_hint().0 == 0  => return Err(ProofError::NoSuchNode),
            
            NodeType::Branch(n) => {
```

firewood/src/proof.rs
```diff
@@ -173,34 +179,24 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
 
-                let sub_proof = if term {
-                    SubProof {
-                        encoded: data,
-                        hash: None,
-                    }
-                } else {
-                    self.generate_subproof(data)?
-                };
+                let sub_proof = self.generate_subproof(data.to_vec())?;
```
richardpringle, line 184: Nit: left the same comment below. You might as well call all the methods you need in the same place
```rust
let data = n.chd_encoded().ok_or(ProofError::InvalidData)?.to_vec();
let sub_proof = self.generate_subproof(data)?;
```

firewood/src/proof.rs
```diff
@@ -146,25 +146,31 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
 
-                let decoded_key_nibbles = Nibbles::<0>::new(&decoded_key);
+                if does_not_match {
+                    return Ok((None, Nibbles::<0>::new(&[]).into_iter()));
+                }
 
-                let (cur_key_path, term) =
-                    PartialPath::from_nibbles(decoded_key_nibbles.into_iter());
-                let cur_key = cur_key_path.into_inner();
+                let data = n.data();
```
richardpringle, line 163: Nit:
```suggestion
                let encoded = n.data().to_vec();
```

Then below becomes:
```rust
                let sub_proof = SubProof {
                    encoded,
                    hash: None,
                };
```

richardpringle (changes requested):
Only one non-nit

firewood/src/merkle/node.rs
richardpringle: This one isn't a nit

rkuris: nit: using unnamed tuples here really hurts from a readability perspective. We only write code once, but read it many times. Please consider naming the elements of `ExtNode` for another PR, probably something like 0->`partial_path`, 1->`disk_address`, and 2->`child_encoded`. Then, all the callers to `ExtNode::path`, `chd` and `chd_encoded` can be unrolled inline.

rkuris (approved):

firewood/src/merkle/node.rs
xinifinity: > This one isn't a nit

sorry I thought I pushed the changes, but it didn't go through... It is now. Thanks!

firewood/src/merkle/node.rs
```diff
@@ -310,6 +310,10 @@ impl ExtNode {
         self.1
     }
 
+    pub fn chd_encoded(&self) -> Option<&[u8]> {
+        self.2.as_deref()
+    }
```
richardpringle, line 315: 🚀

firewood/src/proof.rs
```diff
@@ -174,33 +160,41 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+                if does_not_match {
+                    return Ok((None, Nibbles::<0>::new(&[]).into_iter()));
+                }
+                let data = n.chd_encoded().ok_or(ProofError::InvalidData)?.to_vec();
+                let sub_proof = self.generate_subproof(data)?;
 
-            BRANCH_NODE_SIZE => {
+                Ok((sub_proof.into(), key_nibbles))
+            }
+            NodeType::Branch(_) if key_nibbles.size_hint().0 == 0 => Err(ProofError::NoSuchNode),
```
richardpringle, line 187: 😍

richardpringle (approved):

PR #296, opened 2023-09-28
Move methods to functions that don't reference self

richardpringle:
- Make generate_proof a function instead of a method
- Make locate_subproof a function instead of a method
- Make SubProof private

### DIFF
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -128,7 +128,7 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
                 .get(&cur_hash)
                 .ok_or(ProofError::ProofNodeMissing)?;
             let node = NodeType::decode(cur_proof.as_ref())?;
-            let (sub_proof, traversed_nibbles) = self.locate_subproof(key_nibbles, node)?;
+            let (sub_proof, traversed_nibbles) = locate_subproof(key_nibbles, node)?;
             key_nibbles = traversed_nibbles;
 
             cur_hash = match sub_proof {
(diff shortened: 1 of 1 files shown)


firewood/src/proof.rs
richardpringle, line 1: Can make this stuff pub again if I need to in a future PR

richardpringle:
Merge #295 first, it'll be easier for me to rebase this one

firewood/src/proof.rs
```diff
@@ -102,11 +102,95 @@ const BRANCH_NODE_SIZE: usize = 17;
 /// to a single proof step. If reaches an end step during proof verification,
 /// the hash value will be none, and the encoded value will be the value of the
 /// node.
-pub struct SubProof {
+struct SubProof {
     encoded: Vec<u8>,
     hash: Option<[u8; 32]>,
 }
 
+fn generate_subproof(data: Vec<u8>) -> Result<SubProof, ProofError> {
```
rkuris, line 110: nit: `SubProof::try_new(data)`

firewood/src/proof.rs
```diff
@@ -102,11 +102,95 @@ const BRANCH_NODE_SIZE: usize = 17;
...
+    match data.len() {
```
rkuris, line 111: nit: `let sub_hash = match data.len() {` then `Ok(SubProof..., Some(sub_hash)` at the end

firewood/src/proof.rs
```diff
@@ -102,11 +102,95 @@ const BRANCH_NODE_SIZE: usize = 17;
...
+    }
+}
+
+fn locate_subproof<'a>(
+    mut key_nibbles: NibblesIterator<'a, 0>,
+    encoded_node: &[u8],
+) -> Result<(Option<SubProof>, NibblesIterator<'a, 0>), ProofError> {
+    let items: Vec<Encoded<Vec<u8>>> = bincode::DefaultOptions::new()
+        .deserialize(encoded_node)
+        .map_err(ProofError::DecodeError)?;
```
rkuris, line 142: This loses the reason why the deserialize fails, although I'm not sure that's easy to fix.

rkuris (approved):

firewood/src/proof.rs
richardpringle: Can do that in a follow-up. All I'm doing in this PR is moving the code that already exists since it didn't make sense the way it was laid out before.

firewood/src/proof.rs
richardpringle: Agree, but same response as above. I think it's easier to see a commit in the history that just moves a function without changing any of the code.

firewood/src/proof.rs
richardpringle: Did you look at the deleted code? I haven't changed anything. Also, I don't think we lose any information; `ProofError::DecodeError` contains the `bincode::Error`.

firewood/src/proof.rs
rkuris: Sure. I usually do this as a separate commit -- move in the first commit, then rename and refactor next. If I see the function move then I can study the differences separately if I so desire.

Otherwise I think these smaller nit comments get lost (at least for me). They are all, however, nits so nothing stops merging here.

PR #301, opened 2023-09-30
Renames for readability

rkuris:
is_encoded_big renamed to is_encoded_bigger_than_hash_len

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -133,7 +133,7 @@ impl BranchNode {
                 Some(c) => {
                     let mut c_ref = store.get_item(*c).unwrap();
 
-                    if c_ref.is_encoded_big::<S>(store) {
+                    if c_ref.is_encoded_bigger_than_hash_len::<S>(store) {
                         list[i] = Encoded::Data(
                             bincode::DefaultOptions::new()
                                 .serialize(&&(*c_ref.get_root_hash::<S>(store))[..])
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/node.rs
```diff
@@ -133,7 +133,7 @@ impl BranchNode {
                 Some(c) => {
                     let mut c_ref = store.get_item(*c).unwrap();
 
-                    if c_ref.is_encoded_big::<S>(store) {
+                    if c_ref.is_encoded_bigger_than_hash_len::<S>(store) {
```
richardpringle, line 136: This is kind of a nit? But then this rename is also kind of a nit... so I think a nit on a nit cancels out, making this a real change suggestion.
```suggestion
                    if c_ref.is_encoded_longer_than_hash::<S>(store) {
```

We are talking about length, so we should use the terms "longer" and "shorter" for comparison. And if we use the term "longer," `len` is already implied for `hash`, so we can drop it from the name.

richardpringle (changes requested):

JacobEverly (approved):

firewood/src/merkle/node.rs
```diff
@@ -133,7 +133,7 @@ impl BranchNode {
                 Some(c) => {
                     let mut c_ref = store.get_item(*c).unwrap();
 
-                    if c_ref.is_encoded_big::<S>(store) {
+                    if c_ref.is_encoded_longer_than_hash_len::<S>(store) {
```
richardpringle, line 136: You don't need the `len` on the end because `longer` implies length...

richardpringle (approved):

firewood/src/merkle/node.rs
rkuris: True, but "Longer than hash" is less readable IMO.

PR #302, opened 2023-10-02
Remove line about Docker in README

richardpringle:
While there may be some corner cases that still don't work, the
examples, benchmarks, and tests all run in Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,113 +40,123 @@ versions with no additional cost at all.
 
 ## Terminology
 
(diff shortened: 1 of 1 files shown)


rkuris (approved):
nit: reformat in separate commit so I can easily find the actual change here.

richardpringle:
>nit: reformat in separate commit so I can easily find the actual change here.

Sorry, didn't even mean to push formatting. I forgot my auto-formatter was on

PR #303, opened 2023-10-02
chore: use `decode` in range proof verification

xinifinity:
With this, now `proof` no longer deserialize the node directly, and all encoding/decoding logic is contained inside `node.rs`.

### DIFF
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -448,11 +448,11 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
         end_node: bool,
     ) -> Result<(DiskAddress, Option<SubProof>, usize), ProofError> {
         let node = NodeType::decode(buf)?;
-        let addr = merkle
-            .new_node(Node::new(node.clone()))
-            .map(|node| node.as_ptr())
+        let new_node = merkle
+            .new_node(Node::new(node))
             .map_err(ProofError::InvalidNode)?;
-        match node {
+        let addr = new_node.as_ptr();
+        match new_node.inner() {
             NodeType::Leaf(n) => {
                 let cur_key = &n.path().0;
                 // Check if the key of current node match with the given key.
@@ -460,9 +460,8 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
                     return Ok((addr, None, 0));
                 }
 
-                let encoded = n.data().to_vec();
                 let subproof = Some(SubProof {
-                    encoded,
+                    encoded: n.data().to_vec(),
                     hash: None,
                 });
                 Ok((addr, subproof, cur_key.len()))
@@ -486,10 +485,10 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
             NodeType::Branch(n) => {
                 // Check if the subproof with the given key exist.
                 let index = key[0] as usize;
-                let Some(data) = n.chd_encode()[index].clone() else {
+                let Some(data) = &n.chd_encode()[index] else {
                     return Ok((addr, None, 1));
                 };
-                let subproof = generate_subproof(data)?;
+                let subproof = generate_subproof(data.to_vec())?;
                 Ok((addr, Some(subproof), 1))
             }
         }
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -4,7 +4,6 @@
 use std::cmp::Ordering;
 use std::ops::Deref;
 
-use bincode::Options;
 use nix::errno::Errno;
 use sha3::Digest;
 use shale::disk_address::DiskAddress;
@@ -12,15 +11,11 @@ use shale::ShaleError;
 use shale::ShaleStore;
 use thiserror::Error;
 
-use crate::merkle::Encoded;
 use crate::nibbles::Nibbles;
 use crate::nibbles::NibblesIterator;
 use crate::{
     db::DbError,
-    merkle::{
-        to_nibble_array, BranchNode, ExtNode, LeafNode, Merkle, MerkleError, Node, NodeType,
-        PartialPath, NBRANCH,
-    },
+    merkle::{to_nibble_array, Merkle, MerkleError, Node, NodeType},
     merkle_util::{new_merkle, DataStoreError, MerkleSetup},
     v2::api::Proof,
 };
@@ -94,9 +89,6 @@ impl From<DbError> for ProofError {
         }
     }
 }
-
-const EXT_NODE_SIZE: usize = 2;
-const BRANCH_NODE_SIZE: usize = 17;
 
 /// SubProof contains the encoded value and the hash value of a node that maps
 /// to a single proof step. If reaches an end step during proof verification,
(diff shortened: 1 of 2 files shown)


firewood/src/proof.rs
```diff
@@ -4,23 +4,18 @@
 use std::cmp::Ordering;
 use std::ops::Deref;
 
-use bincode::Options;
```
rkuris, line 7: 🚀

firewood/src/proof.rs
```diff
@@ -455,93 +447,51 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                    .into_iter()
-                    .flat_map(to_nibble_array)
-                    .collect();
-
-                let (cur_key_path, term) = PartialPath::decode(&cur_key_path);
-                let cur_key = cur_key_path.into_inner();
-
-                let data: Vec<u8> = items.next().unwrap().decode()?;
-
+        let node = NodeType::decode(buf)?;
```
rkuris, line 450: nit: naming: `decode` -> `new` or, even beter, `impl From<&[u8]> for `NodeType`. Would also be okay with `deserialize`.

firewood/src/proof.rs
```diff
@@ -455,93 +447,51 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+        let addr = merkle
+            .new_node(Node::new(node.clone()))
+            .map(|node| node.as_ptr())
+            .map_err(ProofError::InvalidNode)?;
+        match node {
```
rkuris, line 455: Avoid the clone here:
```suggestion
        let new_node = merkle
            .new_node(Node::new(node))
            .map_err(ProofError::InvalidNode)?;
        let addr = new_node.as_ptr();
        match new_node.inner() {
```

firewood/src/proof.rs
```diff
@@ -455,93 +447,51 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                };
-
-                let cur_key_len = cur_key.len();
-
-                let ext_ptr = get_ext_ptr(merkle, term, Data(data), CurKey(cur_key))?;
-
-                Ok((ext_ptr, subproof, cur_key_len))
+                let encoded = n.data().to_vec();
+                let subproof = Some(SubProof {
+                    encoded,
```
rkuris, line 465: nit: would have just done `encoded: n.data().to_vec()` rather than having a local

firewood/src/proof.rs
```diff
@@ -455,93 +447,51 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+            NodeType::Branch(_) if key.is_empty() => Err(ProofError::NoSuchNode),
+            NodeType::Branch(n) => {
                 // Check if the subproof with the given key exist.
                 let index = key[0] as usize;
-
-                let Some(data) = chd_encoded[index].clone() else {
-                    let branch_ptr = build_branch_ptr(merkle, value, chd_encoded)?;
-
-                    return Ok((branch_ptr, None, 1));
+                let Some(data) = n.chd_encode()[index].clone() else {
```
rkuris, line 489: Seems like we can avoid this clone too

rkuris (approved):
Approved, but consider removing the two new clones.

firewood/src/proof.rs
xinifinity: Probably will do in a follow up.

richardpringle: I think the `decode` static method is aptly named. `new` usually does not contain much logic and is generally just used to set fields. `TryFrom` might make more sense, but even native types like `u64` have static methods like [`from_ne_bytes`](https://doc.rust-lang.org/stable/std/primitive.u64.html#method.from_ne_bytes) instead of implementing the `TryFrom` trait. 

Usually `From` and `Into` and their `Try*` equivalents are more like type-casts. A `Vec<u8>` may contain the data to create a `NodeType`, but they are two very different types. 

Checkout [all the `TryFrom` implementors](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html#implementors) to see the pattern.

firewood/src/proof.rs
```diff
@@ -455,93 +447,50 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                let cur_key_len = cur_key.len();
-
-                let ext_ptr = get_ext_ptr(merkle, term, Data(data), CurKey(cur_key))?;
-
-                Ok((ext_ptr, subproof, cur_key_len))
+                let subproof = Some(SubProof {
+                    encoded: n.data().to_vec(),
+                    hash: None,
+                });
+                Ok((addr, subproof, cur_key.len()))
```
richardpringle, line 467: nit:
```suggestion
                let subproof = SubProof {
                    encoded: n.data().to_vec(),
                    hash: None,
                };

                Ok((addr, subproof.into(), cur_key.len()))
```

firewood/src/proof.rs
```diff
@@ -455,93 +447,50 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                    chd_encoded[i] = Some(data).filter(|data| !data.is_empty());
-                }
-
-                // If the node is the last one to be decoded, then no subproof to be extracted.
-                if end_node {
-                    let branch_ptr = build_branch_ptr(merkle, value, chd_encoded)?;
-
-                    return Ok((branch_ptr, None, 1));
+                // Check if the key of current node match with the given key.
+                if key.len() < cur_key.len() || &key[..cur_key.len()] != cur_key {
```
richardpringle, line 473: Nit, repeated logic should live in a function. If the function is well-named, we wouldn't need a code comment. 

```rust
pub trait ContainsOtherExt {
    fn contains_other(&self, other: &Self) -> bool;
}

impl<T> ContainsOtherExt for &[T] {
    fn contains_other(&self, other: &Self) -> bool {
        self.len() >= other.len() && &self[..other.len()] == other
    }
}
```
Then the code reads like English
```rust
if !key.contains_other(cur_key) {
// ...
```

firewood/src/proof.rs
```diff
@@ -455,93 +447,50 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+                    return Ok((addr, None, 0));
                 }
 
-                if key.is_empty() {
-                    return Err(ProofError::NoSuchNode);
-                }
+                let encoded = n.chd_encoded().ok_or(ProofError::InvalidData)?.to_vec();
```
richardpringle, line 477: This allocation is unfortunate... is it possible to have a `new_node.into_inner()` such that you can remove fields from the parent?

xinifinity: Will try to address it in a follow up.

firewood/src/proof.rs
rkuris: That's because there isn't a type specific to endianness to convert from, and `TryFrom` doesn't take any arguments -- there has to be a concrete type, so that's appropriate. However, here there aren't any other arguments.

firewood/src/proof.rs
xinifinity: Tried it out, but it seems cannot avoid the `clone()`, as it gives `error cannot move out of a shared reference`.

PR #304, opened 2023-10-03
Switch to unbounded channel

rkuris:
This reduces the number of blocking calls for a spike, plus I can't imagine why anyone would want to tune this.


xinifinity (approved):

richardpringle:
>plus I can't imagine why anyone would want to tune this

Unbounded channels should only be used when necessary (here). No back pressure is not a great pattern in-general and can lead to some issues. 

I also remember something about bounded channels being more performant because of their implementation, but my memory sometimes lies to me.

PR #305, opened 2023-10-04
chore: naming the elements of `ExtNode`

xinifinity:
Follow up of [discussion](https://github.com/ava-labs/firewood/pull/295#discussion_r1341615543)

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -243,15 +243,19 @@ impl LeafNode {
 }
 
 #[derive(PartialEq, Eq, Clone)]
-pub struct ExtNode(
-    pub(super) PartialPath,
-    pub(super) DiskAddress,
-    pub(super) Option<Vec<u8>>,
-);
+pub struct ExtNode {
+    path: PartialPath,
+    chd: DiskAddress,
+    chd_encoded: Option<Vec<u8>>,
+}
 
 impl Debug for ExtNode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        write!(f, "[Extension {:?} {:?} {:?}]", self.0, self.1, self.2)
+        write!(
+            f,
+            "[Extension {:?} {:?} {:?}]",
+            self.path, self.chd, self.chd_encoded
+        )
     }
 }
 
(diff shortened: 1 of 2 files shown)


firewood/src/merkle/node.rs
```diff
@@ -243,15 +243,19 @@ impl LeafNode {
...
-pub struct ExtNode(
-    pub(super) PartialPath,
-    pub(super) DiskAddress,
-    pub(super) Option<Vec<u8>>,
-);
+pub struct ExtNode {
+    path: PartialPath,
+    chd: DiskAddress,
+    chd_encoded: Option<Vec<u8>>,
+}
```
richardpringle, line 250: 🥇

firewood/src/merkle/node.rs
```diff
@@ -243,15 +243,19 @@ impl LeafNode {
...
 
 impl Debug for ExtNode {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        write!(f, "[Extension {:?} {:?} {:?}]", self.0, self.1, self.2)
+        write!(
+            f,
+            "[Extension {:?} {:?} {:?}]",
+            self.path, self.chd, self.chd_encoded
```
richardpringle, line 257: More of an FYI than comment or change-request:
To prevent forgetting to update implementations like these, it's best to destructure the fields such that you get a compiler error if you add a field. 
```suggestion
        let Self { path, chd, chd_encoded } = self;

        write!(
            f,
            "[Extension {path:?} {chd:?} {chd_encoded:?}]",
```

firewood/src/merkle/node.rs
```diff
@@ -243,15 +243,19 @@ impl LeafNode {
...
 #[derive(PartialEq, Eq, Clone)]
-pub struct ExtNode(
-    pub(super) PartialPath,
-    pub(super) DiskAddress,
-    pub(super) Option<Vec<u8>>,
-);
+pub struct ExtNode {
+    path: PartialPath,
+    chd: DiskAddress,
+    chd_encoded: Option<Vec<u8>>,
```
richardpringle, line 249: nit:
```suggestion
    child: DiskAddress,
    child_encoded: Option<Vec<u8>>,
```

firewood/src/merkle/node.rs
```diff
@@ -299,27 +303,35 @@ impl ExtNode {
     }
 
     pub fn new(path: Vec<u8>, chd: DiskAddress, chd_encoded: Option<Vec<u8>>) -> Self {
-        ExtNode(PartialPath(path), chd, chd_encoded)
+        ExtNode {
+            path: PartialPath(path),
```
richardpringle, line 307: Just a question, why not force the caller to pass in a partial-path so that it's more explicit?

firewood/src/merkle/node.rs
richardpringle: After reading through all of the code here, I definitely think it's an improvement, but we really shouldn't be using a getter pattern with what is effectively an internal type like this. I don't think it provides any benefit. I would make all of these fields `pub(super)` and the struct itself `pub(crate)` assuming that the node-structure never actually leaks outside of the database.

richardpringle (approved):
I think this is a slight improvement, but I think it also brought light to some other less than ideal patterns

PR #307, opened 2023-10-05
Rely on mutable iterators instead of offsets

richardpringle:
Small quality of life improvement

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -7,6 +7,7 @@ use shale::{disk_address::DiskAddress, ObjRef, ShaleError, ShaleStore};
 use std::{
     collections::HashMap,
     io::Write,
+    iter::once,
     sync::{atomic::Ordering, OnceLock},
 };
 use thiserror::Error;
@@ -53,7 +54,7 @@ pub struct Merkle<S> {
     store: Box<S>,
 }
 
-impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
+impl<S: ShaleStore<Node>> Merkle<S> {
     pub fn get_node(&self, ptr: DiskAddress) -> Result<ObjRef<Node>, MerkleError> {
         self.store.get_item(ptr).map_err(Into::into)
     }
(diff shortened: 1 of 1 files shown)


firewood/src/merkle.rs
```diff
@@ -371,13 +367,23 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+
         // walk down the merkle tree starting from next_node, currently the root
-        for (key_nib_offset, key_nib) in key_nibbles.into_iter().enumerate() {
+        loop {
+            let Some(key_nib) = key_nibbles.next() else {
+                break;
+            };
+
             // special handling for extension nodes
             if nskip > 0 {
```
rkuris, line 382: nit: this could be optimized with `iter::skip(nskip)`, maybe even tossing `nskip` and just doing it when it's set

firewood/src/merkle.rs
```diff
@@ -408,18 +412,17 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
                 NodeType::Leaf(n) => {
                     // we collided with another key; make a copy
                     // of the stored key to pass into split
                     let n_path = n.0.to_vec();
                     let n_value = Some(n.1.clone());
-                    let rem_path = key_nibbles
-                        .into_iter()
-                        .skip(key_nib_offset)
-                        .collect::<Vec<_>>();
+                    let rem_path = once(key_nib).chain(key_nibbles).collect::<Vec<_>>();
```
rkuris, line 425: nit: very nice, but took some digesting; maybe rename `key_nib` -> `current_nibble`

rkuris (approved):

firewood/src/merkle.rs
richardpringle: haha I'm in the middle of tossing it for a follow up

firewood/src/merkle.rs
richardpringle: Will also do in a follow-up. I didn't love it either, but I wanted to do away with offsets to make the overall function more digestible.

PR #308, opened 2023-10-05
API implementation part 3

rkuris:
Probably easier to review commit-by-commit

### DIFF
--- a/firewood/Cargo.toml
+++ b/firewood/Cargo.toml
@@ -32,7 +32,7 @@ parking_lot = "0.12.1"
 serde = { version = "1.0", features = ["derive"] }
 sha3 = "0.10.2"
 thiserror = "1.0.38"
-tokio = { version = "1.21.1", features = ["rt", "sync", "macros"] }
+tokio = { version = "1.21.1", features = ["rt", "sync", "macros", "rt-multi-thread"] }
 typed-builder = "0.16.0"
 bincode = "1.3.3"
 
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -37,6 +37,7 @@ use std::{
     sync::Arc,
     thread::JoinHandle,
 };
+use tokio::task::block_in_place;
 
 mod proposal;
 
@@ -390,7 +391,8 @@ impl api::Db for Db {
     type Proposal = Proposal;
 
     async fn revision(&self, root_hash: HashKey) -> Result<Arc<Self::Historical>, api::Error> {
-        if let Some(rev) = self.get_revision(&TrieHash(root_hash)) {
+        let rev = block_in_place(|| self.get_revision(&TrieHash(root_hash)));
+        if let Some(rev) = rev {
             Ok(Arc::new(rev.rev))
         } else {
             Err(api::Error::HashNotFound {
--- a/firewood/src/v2/api.rs
+++ b/firewood/src/v2/api.rs
@@ -176,10 +176,6 @@ pub trait Proposal: DbView {
     type Proposal: DbView + Proposal;
 
     /// Commit this revision
-    ///
-    /// # Return value
-    ///
-    /// * A reference to a new historical view
     async fn commit(self: Arc<Self>) -> Result<(), Error>;
 
     /// Propose a new revision on top of an existing proposal
--- a/firewood/src/v2/api.rs
+++ b/firewood/src/v2/api.rs
@@ -96,7 +96,7 @@ pub struct Proof<N>(pub HashMap<HashKey, N>);
 pub trait Db {
     type Historical: DbView;
 
-    type Proposal: DbView + Proposal<Self::Historical>;
+    type Proposal: DbView + Proposal;
 
     /// Get a reference to a specific view based on a hash
     ///
@@ -172,15 +172,15 @@ pub trait DbView {
 /// [DbView], which means you can fetch values from it or
 /// obtain proofs.
 #[async_trait]
-pub trait Proposal<T: DbView>: DbView {
-    type Proposal: DbView + Proposal<T>;
+pub trait Proposal: DbView {
+    type Proposal: DbView + Proposal;
 
     /// Commit this revision
     ///
     /// # Return value
     ///
     /// * A reference to a new historical view
-    async fn commit(self: Arc<Self>) -> Result<Arc<T>, Error>;
+    async fn commit(self: Arc<Self>) -> Result<(), Error>;
 
     /// Propose a new revision on top of an existing proposal
     ///
--- a/firewood/src/v2/propose.rs
+++ b/firewood/src/v2/propose.rs
@@ -139,7 +139,7 @@ impl<T: api::DbView + Send + Sync> api::DbView for Proposal<T> {
 }
 
 #[async_trait]
-impl<T: api::DbView + Send + Sync> api::Proposal<T> for Proposal<T> {
+impl<T: api::DbView + Send + Sync> api::Proposal for Proposal<T> {
     type Proposal = Proposal<T>;
 
     async fn propose<K: KeyType, V: ValueType>(
@@ -150,12 +150,12 @@ impl<T: api::DbView + Send + Sync> api::Proposal<T> for Proposal<T> {
         Ok(Proposal::new(ProposalBase::Proposal(self), data))
     }
 
-    async fn commit(self: Arc<Self>) -> Result<Arc<T>, api::Error> {
+    async fn commit(self: Arc<Self>) -> Result<(), api::Error> {
         // TODO: commit should modify the db; this will only work for
         // emptydb at the moment
         match &self.base {
             ProposalBase::Proposal(base) => base.clone().commit().await,
-            ProposalBase::View(v) => Ok(v.clone()),
+            ProposalBase::View(_) => Ok(()),
         }
     }
 }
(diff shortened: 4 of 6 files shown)


firewood/Cargo.toml
```diff
@@ -32,7 +32,7 @@ parking_lot = "0.12.1"
 serde = { version = "1.0", features = ["derive"] }
 sha3 = "0.10.2"
 thiserror = "1.0.38"
-tokio = { version = "1.21.1", features = ["rt", "sync", "macros"] }
+tokio = { version = "1.21.1", features = ["rt", "sync", "macros", "rt-multi-thread"] }
```
richardpringle, line 35: 🚀

firewood/src/db.rs
```diff
@@ -390,7 +391,8 @@ impl api::Db for Db {
     type Proposal = Proposal;
 
     async fn revision(&self, root_hash: HashKey) -> Result<Arc<Self::Historical>, api::Error> {
-        if let Some(rev) = self.get_revision(&TrieHash(root_hash)) {
+        let rev = block_in_place(|| self.get_revision(&TrieHash(root_hash)));
```
richardpringle, line 394: 💩

richardpringle: Why `block_in_place` over `spawn_blocking`. The latter is usually used when one is unsure of the outer context in which the blocking call is made. In other words, this is only faster if there aren't a bunch of other tasks running on the same thread. 

My intuition tells me that performance would be much more predictable using `spawn_blocking` but *could be* faster using `block_in_place`. I'm not necessarily opposed to it, but I think it should be justified.

firewood/src/v2/api.rs
```diff
@@ -96,7 +96,7 @@ pub struct Proof<N>(pub HashMap<HashKey, N>);
 pub trait Db {
     type Historical: DbView;
 
-    type Proposal: DbView + Proposal<Self::Historical>;
+    type Proposal: DbView + Proposal;
```
richardpringle, line 99: 🚀

firewood/src/v2/api.rs
```diff
@@ -172,15 +172,15 @@ pub trait DbView {
...
+pub trait Proposal: DbView {
+    type Proposal: DbView + Proposal;
 
     /// Commit this revision
     ///
     /// # Return value
     ///
     /// * A reference to a new historical view
-    async fn commit(self: Arc<Self>) -> Result<Arc<T>, Error>;
+    async fn commit(self: Arc<Self>) -> Result<(), Error>;
```
richardpringle, line 183: Need to update the doc-comments

firewood/src/v2/propose.rs
```diff
@@ -150,12 +150,12 @@ impl<T: api::DbView + Send + Sync> api::Proposal<T> for Proposal<T> {
         Ok(Proposal::new(ProposalBase::Proposal(self), data))
     }
 
-    async fn commit(self: Arc<Self>) -> Result<Arc<T>, api::Error> {
+    async fn commit(self: Arc<Self>) -> Result<(), api::Error> {
```
richardpringle, line 153: Wasn't this supposed to return the latest view upon success? Now we have to go back to the database and get the latest view?

firewood/tests/v2api.rs
richardpringle, line 1: I think you should just rename this file to API. `v2` is just a placeholder for the code, right? We definitely aren't going to write any v1 tests at this point.

firewood/tests/v2api.rs
```diff
@@ -0,0 +1,69 @@
...
+                .file_nbit(15)
+                .block_nbit(8)
+                .max_revisions(10)
+                .build(),
+        )
+        .truncate(true)
+        .build();
+    let db = Arc::new(testdb(cfg).await?);
+    let empty_hash = db.root_hash().await?;
+    assert_ne!(empty_hash, [0; 32]);
```
richardpringle, line 28: Do we not have a unit test somewhere for this?

firewood/tests/v2api.rs
```diff
@@ -0,0 +1,69 @@
...
+    // ensure the latest hash is different
+    let latest = db.root_hash().await?;
+    assert_ne!(empty_hash, latest);
+
+    // fetch the view of the latest
+    let view = db.revision(latest).await.unwrap();
+
+    // check that the key/value is there
+    let value = view.val(b"smoke").await.unwrap();
+    assert_eq!(value, Some(b"test".to_vec()));
```
richardpringle, line 47: Should only define the key and value once then use that definition everywhere. 

```suggestion
    let key = b"smoke";
    let value = b"test";

    // insert a single key/value
    let batch_put = BatchOp::Put {
        key,
        value,
    };
    let proposal: Arc<firewood::db::Proposal> = db.propose(vec![batch_put]).await?.into();
    proposal.commit().await?;

    // ensure the latest hash is different
    let latest = db.root_hash().await?;
    assert_ne!(empty_hash, latest);

    // fetch the view of the latest
    let view = db.revision(latest).await.unwrap();

    // check that the key/value is there
    assert_eq!(view.val(key).await.unwrap().unwrap(), value);
```
you can compare `Vec<T>` with `&[T]`.

firewood/tests/v2api.rs
```diff
@@ -0,0 +1,69 @@
+use std::{error::Error, path::PathBuf, sync::Arc};
+
+use firewood::{
+    db::{BatchOp, Db as PersistedDb, DbConfig, DbError, WalConfig},
+    v2::api::{Db, DbView, Proposal},
+};
+
+#[tokio::test(flavor = "multi_thread")]
```
richardpringle, line 8: The test would be simpler if you didn't use the test macro and simply
```rust
fn smoke() -> Result<(), Box<dyn Error>> {
    let cfg = // ...
    let tmpdbpath = tmp_dir().join("testdb");
    let db = PersistedDb::new(tmpdbpath, &cfg).unwrap();
    
    tokio::block_on(async move {
        // then put your test logic here. 
    });
}
```

Then `spawn_blocking` wouldn't be necessary

firewood/src/db.rs
```diff
@@ -390,7 +391,8 @@ impl api::Db for Db {
...
+        if let Some(rev) = rev {
```
richardpringle, line 395:
```suggestion
        block_in_place(|| {
             self
                  .get_revision(&TrieHash(root_hash)))
                  .map(Arc::new)
                  .ok_or(api::Error::HashNotFound { provided: root_hash }
        });
```
That's my best guess at formatting

firewood/src/db.rs
richardpringle: Please respond before next round of review

firewood/src/db.rs
richardpringle: nit ^

firewood/src/v2/api.rs
richardpringle: Please make this change before next review

firewood/src/v2/propose.rs
richardpringle: simply a question, should not block merge

firewood/tests/v2api.rs
richardpringle: Please do this... or `v2_api.rs` or `api_v2.rs`... but I think it should be removed.

firewood/tests/v2api.rs
richardpringle: Not important, just a suggestion

firewood/tests/v2api.rs
richardpringle: I think this belongs elsewhere, but shouldn't hold up PR

firewood/tests/v2api.rs
richardpringle: Please do this before merge

richardpringle (changes requested):

firewood/src/db.rs
rkuris: The trouble here is that I can't move `self` into the closure, and `block_in_place` can borrow self since the thread is guaranteed to be done with it before returning.

I think if this is slower or faster is in the eye of the beholder (or of the benchmark tool). Since we don't know, and there are lifetime issues, this seems like a decent choice.

firewood/src/v2/propose.rs
rkuris: The issue here is that T is specified by the caller. This really needs T specified by the implementation. It's difficult to implement it to return the specific type asked for by the caller, so that's why I got rid of T completely.

This can be done, but I don't think it's critical so I'm just removing the generic here as a start (you'd have to do that anyway).

firewood/tests/v2api.rs
rkuris: Can do this as a followup, especially once v2 is the new standard.

firewood/tests/v2api.rs
rkuris: This is an integration/smoke test for v2::api::root_hash, which isn't really tested elsewhere right now.

richardpringle: It should be! We.can remove this once it is, though.

firewood/tests/v2api.rs
```diff
@@ -0,0 +1,70 @@
...
+                .build(),
+        )
+        .truncate(true)
+        .build();
+    let db = Arc::new(testdb(cfg).await?);
+    let empty_hash = db.root_hash().await?;
+    assert_ne!(empty_hash, [0; 32]);
+
+    // insert a single key/value
+    let (key, value) = (b"smoke", b"test");
```
richardpringle, line 34: Nice

richardpringle (approved):

PR #311, opened 2023-10-06
Cleanup merkle-insert code

richardpringle:
- Methods should actually use self
- Cleanup merkle insert

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -40,7 +40,7 @@ macro_rules! write_node {
     ($self: expr, $r: expr, $modify: expr, $parents: expr, $deleted: expr) => {
         if let Err(_) = $r.write($modify) {
             let ptr = $self.new_node($r.clone())?.as_ptr();
-            $self.set_parent(ptr, $parents);
+            set_parent(ptr, $parents);
             $deleted.push($r.as_ptr());
             true
         } else {
@@ -154,20 +154,6 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
             self.dump_(root, w)?;
         };
         Ok(())
-    }
-
-    fn set_parent(&self, new_chd: DiskAddress, parents: &mut [(ObjRef<'_, Node>, u8)]) {
-        let (p_ref, idx) = parents.last_mut().unwrap();
-        p_ref
-            .write(|p| {
-                match &mut p.inner {
-                    NodeType::Branch(pp) => pp.chd[*idx as usize] = Some(new_chd),
-                    NodeType::Extension(pp) => *pp.chd_mut() = new_chd,
-                    _ => unreachable!(),
-                }
-                p.rehash();
-            })
-            .unwrap();
     }
 
     #[allow(clippy::too_many_arguments)]
@@ -345,7 +331,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         // observation:
         // - leaf/extension node can only be the child of a branch node
         // - branch node can only be the child of a branch/extension node
-        self.set_parent(new_chd, parents);
+        set_parent(new_chd, parents);
         Ok(None)
     }
 
@@ -527,7 +513,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                         chd_encoded: Default::default(),
                     })))?
                     .as_ptr();
-                self.set_parent(branch, &mut parents);
+                set_parent(branch, &mut parents);
             }
         }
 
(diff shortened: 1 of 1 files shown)


firewood/src/merkle.rs
```diff
@@ -156,20 +158,6 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
-                match &mut p.inner {
-                    NodeType::Branch(pp) => pp.chd[*idx as usize] = Some(new_chd),
-                    NodeType::Extension(pp) => *pp.chd_mut() = new_chd,
-                    _ => unreachable!(),
-                }
-                p.rehash();
-            })
-            .unwrap();
-    }
-
```
richardpringle, line 172: Moved

firewood/src/merkle.rs
```diff
@@ -355,58 +343,73 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         val: Vec<u8>,
         root: DiskAddress,
     ) -> Result<(), MerkleError> {
+        let (parents, deleted) = self.insert_and_return_updates(key, val, root)?;
```
richardpringle, line 346: Had to split this out because of `&self` vs `&mut self` borrowing issues. The alternative was putting the majority of the code in curly braces... I think it's better this way, but open to alternatives.

firewood/src/merkle.rs
```diff
@@ -1182,6 +1186,20 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+        .write(|p| {
+            match &mut p.inner {
+                NodeType::Branch(pp) => pp.chd[*idx as usize] = Some(new_chd),
+                NodeType::Extension(pp) => *pp.chd_mut() = new_chd,
+                _ => unreachable!(),
+            }
+            p.rehash();
+        })
+        .unwrap();
+}
```
richardpringle, line 1201: No changes from https://github.com/ava-labs/firewood/pull/311/files#diff-6a6e300f3d9050248dbafa719531f14eabecc1bfac2169bb2527c81ee5474ecaL159 other than removing `self` reference that was unused.

xinifinity (approved):

firewood/src/merkle.rs
```diff
@@ -34,13 +34,15 @@ pub enum MerkleError {
     ParentLeafBranch,
     #[error("removing internal node references failed")]
     UnsetInternal,
+    #[error("error updating nodes with: {0}")]
```
rkuris, line 37: nit: remove "with"; adds no value and might make the sentence grammatically awkward!

firewood/src/merkle.rs
```diff
@@ -1182,6 +1186,20 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
     }
 }
 
+fn set_parent(new_chd: DiskAddress, parents: &mut [(ObjRef<'_, Node>, u8)]) {
```
rkuris, line 1189: This seems so strange. We pass in a mutable array of parents, then immediately only get the last parent and modify it? Also idx (the second u8) is never mutated. Maybe:
`fn set_parent(new_chd: DiskAddress, mut parent: ObjRef<'_, Node>, parent_node_idx: u8)`?

Followup?

firewood/src/merkle.rs
```diff
@@ -355,58 +343,73 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         val: Vec<u8>,
         root: DiskAddress,
     ) -> Result<(), MerkleError> {
+        let (parents, deleted) = self.insert_and_return_updates(key, val, root)?;
+
+        for mut r in parents {
```
rkuris, line 348: 🏆 
I was thinking: `parents.iter().for_each(|mut &r| r.write(|u| u.rehash())?);` but yours is more readable.

rkuris (approved):

firewood/src/merkle.rs
richardpringle: Yeah, I looked at changing this, but honestly (for now), I think it's best if it stays the way that it is. `set_parent` is called inside that really awful `write_node` macro which is why I didn't change it. 

Step 1: get rid of the macro altogether (I'm not sure that it provides a ton of value)
Step 2: fix `set_parent`. 

Just my opinion... But yeah, I completely agree this is super weird, but to change this now would have been too big a change IMO

firewood/src/merkle.rs
richardpringle: You could do it with `iter_mut` and `try_for_each`, but that's a bigger change than what I wanted to do. You can see the original code on the left somewhere below:
```rust
        for (mut r, _) in parents.into_iter().rev() {
            r.write(|u| u.rehash()).unwrap();
        }
```

The reversed iterator is now returned from `insert_and_return_updates` and I drop the indexes as they aren't used here.

PR #314, opened 2023-10-07
Fix up example, use new async interface

rkuris:
We need more examples; this is a good start.

The old example wasn't super useful, it was probably a better integration test.


firewood/examples/rev.rs
xinifinity, line 1: instead of completely removing it, maybe good to move to integration tests?

rkuris: It would need complete rewriting I think. The interface it's using is the old interface and what it actually does is peek inside, not using the external interface, so it doesn't even make a good integration test.

xinifinity: Hmm, which part of logic (old APIs) you think it is peek inside?

rkuris:
All of them. We should be deprecating the old ones and only exposing the
new ones. This will get more important as we start pushing async down the
stack.

firewood/examples/rev.rs
rkuris: I think we want to deprecate the old APIs, so our examples and even our integration tests should use the new ones.

xinifinity: yeah, I agree, but the point here is to not forget about adding equivalent integration test for V2, maybe create an issue to track it if we just delete in the PR?

rkuris: Created https://github.com/ava-labs/firewood/issues/318 for tracking

firewood/examples/insert.rs
```diff
@@ -0,0 +1,74 @@
...
+
+fn string_to_range(input: &str) -> Result<RangeInclusive<usize>, Box<dyn Error + Sync + Send>> {
+    //<usize as std::str::FromStr>::Err> {
+    let parts: Vec<&str> = input.split('-').collect();
+    match parts.len() {
+        1 => Ok(input.parse()?..=input.parse()?),
+        2 => Ok(parts[0].parse()?..=parts[1].parse()?),
+        _ => Err("Too many dashes in input string".into()),
+    }
+}
```
richardpringle, line 33: It's amazing how little code you need to be able to run this from the command line with all these options.

.github/workflows/ci.yaml
```diff
@@ -145,8 +145,8 @@ jobs:
         # benchmarks were not being done in --release mode, we can enable this again later
         # - name: Run benchmark example
         #   run: RUST_BACKTRACE=1 cargo run --example benchmark -- --nbatch 100 --batch-size 1000
-      - name: Run rev example
-        run: RUST_BACKTRACE=1 cargo run --example rev
+      - name: Run insert example
+        run: RUST_BACKTRACE=1 cargo run --example insert
```
richardpringle, line 149: release?

firewood/examples/insert.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let db = tokio::task::spawn_blocking(move || {
+        Db::new("rev_db", &cfg).expect("db initiation should succeed")
+    })
+    .await
+    .unwrap();
+
+    let keys = args.batch_keys;
+    let start = Instant::now();
+    for _ in 0..args.inserts {
```
richardpringle, line 50: nit
```suggestion

    for _ in 0..args.inserts {
```

firewood/examples/insert.rs
```diff
@@ -0,0 +1,74 @@
...
+                )
+            })
+            .map(|(key, value)| BatchOp::Put { key, value })
+            .collect();
+        let proposal: Arc<firewood::db::Proposal> = db.propose(batch).await.unwrap().into();
+        proposal.commit().await?;
+    }
+    let duration = start.elapsed();
+    println!("Inserted {keys} in {duration:?}");
+    Ok(())
```
richardpringle, line 73: nit:
```suggestion
    }
    
    let duration = start.elapsed();
    println!("Inserted {keys} in {duration:?}");
    
    Ok(())
```

firewood/examples/insert.rs
```diff
@@ -0,0 +1,74 @@
...
+        Db::new("rev_db", &cfg).expect("db initiation should succeed")
+    })
+    .await
+    .unwrap();
+
+    let keys = args.batch_keys;
+    let start = Instant::now();
+    for _ in 0..args.inserts {
+        let keylen = rand::thread_rng().gen_range(args.keylen.clone());
+        let datalen = rand::thread_rng().gen_range(args.datalen.clone());
```
richardpringle, line 52: I have no idea how quick these actually are, but if you're communicating between threads, it's not trivial, right? You should precompute all the lengths and iterate over the pairs.

```rust
let key_and_data_lengths = (0..args.inserts)
    .map(|_| (
        rand::thread_rng().gen_range(args.keylen.clone()),
        rand::thread_rng().gen_range(args.datalen.clone()),
    ))
    .collect::<Vec<(usize, size)>>();
    
let start = Instant::now();
    
for (keylen, datalen) in key_and_data_lengths {
  // ...
```

firewood/examples/insert.rs
```diff
@@ -0,0 +1,74 @@
...
+            .map(|_| {
+                (
+                    rand::thread_rng()
+                        .sample_iter(&Alphanumeric)
+                        .take(keylen as usize)
+                        .collect::<Vec<u8>>(),
+                    rand::thread_rng()
+                        .sample_iter(&Alphanumeric)
+                        .take(datalen as usize)
+                        .collect::<Vec<u8>>(),
```
richardpringle, line 63: Wait, you're also measuring both of these allocations. You should pre-allocate the key and value vectors and only measure transforming them into a batch.

richardpringle (changes requested):
You're measuring a couple things you don't need to be measuring! 

Even if they don't have a significant effect on the results, it should be clear that you don't want to measure the creation of keys and values.

firewood/examples/insert.rs
richardpringle: actual change request

firewood/examples/insert.rs
richardpringle: actual change request

richardpringle:
> You're measuring a couple things you don't need to be measuring!
> 
> Even if they don't have a significant effect on the results, it should be clear that you don't want to measure the creation of keys and values.

@rkuris, I marked the two comments that are actual change requests, everything else is a nit or just a comment

firewood/examples/insert.rs
rkuris: The goal here wasn't an actual benchmark. If I wanted that, I'd have used the benchmarking stuff that's built in that supports these kinds of assumptions. That's why this is in the examples directory.

Your point that it probably shouldn't count the time for the generation of the keys is valid, though. I'm more tempted to just removing the timings and call it a true example though. Thoughts?

firewood/examples/insert.rs
rkuris: Same comment as above.

firewood/examples/insert.rs
richardpringle: Maybe just add a comment about the time? 

Can be done in a follow up

richardpringle (approved):

PR #316, opened 2023-10-10
chore: proof cleanup

xinifinity:
follow up of https://github.com/ava-labs/firewood/pull/303

### DIFF
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -454,23 +454,21 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
         let addr = new_node.as_ptr();
         match new_node.inner() {
             NodeType::Leaf(n) => {
-                let cur_key = &n.path().0;
-                // Check if the key of current node match with the given key.
-                if key.len() < cur_key.len() || &key[..cur_key.len()] != cur_key {
+                let cur_key: &[u8] = &n.path().0;
+                if !key.contains_other(&cur_key) {
                     return Ok((addr, None, 0));
                 }
 
-                let subproof = Some(SubProof {
+                let subproof = SubProof {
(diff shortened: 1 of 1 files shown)


firewood/src/proof.rs
```diff
@@ -454,23 +454,21 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
         let addr = new_node.as_ptr();
         match new_node.inner() {
             NodeType::Leaf(n) => {
-                let cur_key = &n.path().0;
-                // Check if the key of current node match with the given key.
-                if key.len() < cur_key.len() || &key[..cur_key.len()] != cur_key {
+                let cur_key: &[u8] = &n.path().0;
+                if !key.contains_other(&cur_key) {
```
rkuris, line 458: This passes `&&[u8]` to `contains_other` which means we have to double-dereference it in the function.
```suggestion
                let cur_key = n.path().0.as_ref();
                if !key.contains_other(cur_key) {
```
Then later do this:
```
   fn contains_other(&self, other: Self) -> bool {
       self.len() >= other.len() && &self[..other.len()] == other
```
and change the trait

firewood/src/proof.rs
```diff
@@ -454,23 +454,21 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                     encoded: n.data().to_vec(),
                     hash: None,
-                });
-                Ok((addr, subproof, cur_key.len()))
+                };
+                Ok((addr, subproof.into(), cur_key.len()))
             }
             NodeType::Extension(n) => {
-                let cur_key = &n.path.0;
+                let cur_key: &[u8] = &n.path.0;
```
rkuris, line 469: same as above

rkuris (approved):
Approved with changes

PR #317, opened 2023-10-11
Pluggable encoding for nodes

### LINKED ISSUE #293: Abstract the encoding of a node away from the storage of the node
We need a way to change the node encoding easily, particularly if this is a bottleneck.

xinifinity:
Closes #293.

With this one can use any serializer/deserializer for node encoding/decoding and adding a new encoding format by implement `serde` for `EncodedNode`.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -8,7 +8,7 @@ pub use crate::{
 };
 use crate::{
     file,
-    merkle::{Merkle, MerkleError, Node, Proof, ProofError, TrieHash, TRIE_HASH_LEN},
+    merkle::{Bincode, Merkle, MerkleError, Node, Proof, ProofError, TrieHash, TRIE_HASH_LEN},
     storage::{
         buffer::{DiskBuffer, DiskBufferRequester},
         CachedSpace, MemStoreR, SpaceWrite, StoreConfig, StoreDelta, StoreRevMut, StoreRevShared,
@@ -275,7 +275,7 @@ impl<T: MemStoreR + 'static> Universe<Arc<T>> {
 #[derive(Debug)]
 pub struct DbRev<S> {
     header: shale::Obj<DbHeader>,
-    merkle: Merkle<S>,
+    merkle: Merkle<S, Bincode>,
 }
 
 #[async_trait]
@@ -321,7 +321,7 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
     pub fn stream<K: KeyType>(
         &self,
         start_key: Option<K>,
-    ) -> Result<merkle::MerkleKeyValueStream<'_, S>, api::Error> {
+    ) -> Result<merkle::MerkleKeyValueStream<'_, S, Bincode>, api::Error> {
         self.merkle
             .get_iter(start_key, self.header.kv_root)
             .map_err(|e| api::Error::InternalError(Box::new(e)))
@@ -333,7 +333,7 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
         Some(())
     }
 
-    fn borrow_split(&mut self) -> (&mut shale::Obj<DbHeader>, &mut Merkle<S>) {
+    fn borrow_split(&mut self) -> (&mut shale::Obj<DbHeader>, &mut Merkle<S, Bincode>) {
         (&mut self.header, &mut self.merkle)
     }
 
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -6,8 +6,8 @@ use crate::v2::api;
 use futures::{Stream, StreamExt, TryStreamExt};
 use sha3::Digest;
 use std::{
-    cmp::Ordering, collections::HashMap, future::ready, io::Write, iter::once, sync::OnceLock,
-    task::Poll,
+    cmp::Ordering, collections::HashMap, future::ready, io::Write, iter::once, marker::PhantomData,
+    sync::OnceLock, task::Poll,
 };
 use thiserror::Error;
 
@@ -15,7 +15,10 @@ mod node;
 pub mod proof;
 mod trie_hash;
 
-pub use node::{BranchNode, Data, ExtNode, LeafNode, Node, NodeType, PartialPath};
+pub use node::{
+    BinarySerde, Bincode, BranchNode, Data, EncodedNode, EncodedNodeType, ExtNode, LeafNode, Node,
+    NodeType, PartialPath,
+};
 pub use proof::{Proof, ProofError};
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
@@ -39,6 +42,8 @@ pub enum MerkleError {
     UnsetInternal,
     #[error("error updating nodes: {0}")]
     WriteError(#[from] ObjWriteError),
+    #[error("merkle serde error: {0}")]
+    BinarySerdeError(String),
 }
 
 macro_rules! write_node {
@@ -55,11 +60,12 @@ macro_rules! write_node {
 }
 
 #[derive(Debug)]
-pub struct Merkle<S> {
+pub struct Merkle<S, T> {
     store: Box<S>,
+    phantom: PhantomData<T>,
 }
 
-impl<S: ShaleStore<Node>> Merkle<S> {
+impl<S: ShaleStore<Node>, T> Merkle<S, T> {
     pub fn get_node(&self, ptr: DiskAddress) -> Result<ObjRef, MerkleError> {
         self.store.get_item(ptr).map_err(Into::into)
     }
@@ -73,11 +79,72 @@ impl<S: ShaleStore<Node>> Merkle<S> {
     }
 }
 
-impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
+impl<'de, S, T> Merkle<S, T>
+where
+    S: ShaleStore<Node> + Send + Sync,
+    T: BinarySerde,
+    EncodedNode<T>: serde::Serialize + serde::Deserialize<'de>,
+{
     pub fn new(store: Box<S>) -> Self {
-        Self { store }
+        Self {
+            store,
+            phantom: PhantomData,
+        }
     }
 
+    // TODO: use `encode` / `decode` instead of `node.encode` / `node.decode` after extention node removal.
+    #[allow(dead_code)]
+    fn encode(&self, node: &ObjRef) -> Result<Vec<u8>, MerkleError> {
+        let encoded = match node.inner() {
+            NodeType::Leaf(n) => EncodedNode::new(EncodedNodeType::Leaf(n.clone())),
+            NodeType::Branch(n) => {
+                // pair up DiskAddresses with encoded children and pick the right one
+                let encoded_children = n.chd().iter().zip(n.children_encoded.iter());
+                let children = encoded_children
+                    .map(|(child_addr, encoded_child)| {
+                        child_addr
+                            // if there's a child disk address here, get the encoded bytes
+                            .map(|addr| self.get_node(addr).and_then(|node| self.encode(&node)))
+                            // or look for the pre-fetched bytes
+                            .or_else(|| encoded_child.as_ref().map(|child| Ok(child.to_vec())))
+                            .transpose()
+                    })
+                    .collect::<Result<Vec<Option<Vec<u8>>>, MerkleError>>()?
+                    .try_into()
+                    .expect("MAX_CHILDREN will always be yielded");
+
+                EncodedNode::new(EncodedNodeType::Branch {
+                    children,
+                    value: n.value.clone(),
+                })
+            }
+
+            NodeType::Extension(_) => todo!(),
+        };
+
+        Bincode::serialize(&encoded).map_err(|e| MerkleError::BinarySerdeError(e.to_string()))
+    }
+
+    #[allow(dead_code)]
+    fn decode(&self, buf: &'de [u8]) -> Result<NodeType, MerkleError> {
+        let encoded: EncodedNode<T> =
+            T::deserialize(buf).map_err(|e| MerkleError::BinarySerdeError(e.to_string()))?;
+
+        match encoded.node {
+            EncodedNodeType::Leaf(leaf) => Ok(NodeType::Leaf(leaf)),
+            EncodedNodeType::Branch { children, value } => {
+                let path = Vec::new().into();
+                let value = value.map(|v| v.0);
+                Ok(NodeType::Branch(
+                    BranchNode::new(path, [None; BranchNode::MAX_CHILDREN], value, *children)
+                        .into(),
+                ))
+            }
+        }
+    }
+}
+
+impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
     pub fn init_root(&self) -> Result<DiskAddress, MerkleError> {
         self.store
             .put_item(
@@ -1072,7 +1139,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         &mut self,
         key: K,
         root: DiskAddress,
-    ) -> Result<Option<RefMut<S>>, MerkleError> {
+    ) -> Result<Option<RefMut<S, T>>, MerkleError> {
         if root.is_null() {
             return Ok(None);
         }
@@ -1151,7 +1218,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         &self,
         key: Option<K>,
         root: DiskAddress,
-    ) -> Result<MerkleKeyValueStream<'_, S>, MerkleError> {
+    ) -> Result<MerkleKeyValueStream<'_, S, T>, MerkleError> {
         Ok(MerkleKeyValueStream {
             key_state: IteratorState::new(key),
             merkle_root: root,
@@ -1269,13 +1336,15 @@ impl<'a> Default for IteratorState<'a> {
 /// This iterator is not fused. If you read past the None value, you start
 /// over at the beginning. If you need a fused iterator, consider using
 /// std::iter::fuse
-pub struct MerkleKeyValueStream<'a, S> {
+pub struct MerkleKeyValueStream<'a, S, T> {
     key_state: IteratorState<'a>,
     merkle_root: DiskAddress,
-    merkle: &'a Merkle<S>,
+    merkle: &'a Merkle<S, T>,
 }
 
-impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyValueStream<'a, S> {
+impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync, T> Stream
+    for MerkleKeyValueStream<'a, S, T>
+{
     type Item = Result<(Vec<u8>, Vec<u8>), api::Error>;
 
     fn poll_next(
@@ -1531,10 +1600,10 @@ fn set_parent(new_chd: DiskAddress, parents: &mut [(ObjRef, u8)]) {
 
 pub struct Ref<'a>(ObjRef<'a>);
 
-pub struct RefMut<'a, S> {
+pub struct RefMut<'a, S, T> {
     ptr: DiskAddress,
     parents: ParentAddresses,
-    merkle: &'a mut Merkle<S>,
+    merkle: &'a mut Merkle<S, T>,
 }
 
 impl<'a> std::ops::Deref for Ref<'a> {
@@ -1548,8 +1617,8 @@ impl<'a> std::ops::Deref for Ref<'a> {
     }
 }
 
-impl<'a, S: ShaleStore<Node> + Send + Sync> RefMut<'a, S> {
-    fn new(ptr: DiskAddress, parents: ParentAddresses, merkle: &'a mut Merkle<S>) -> Self {
+impl<'a, S: ShaleStore<Node> + Send + Sync, T> RefMut<'a, S, T> {
+    fn new(ptr: DiskAddress, parents: ParentAddresses, merkle: &'a mut Merkle<S, T>) -> Self {
         Self {
             ptr,
             parents,
@@ -1621,7 +1690,7 @@ mod tests {
         assert_eq!(n, nibbles);
     }
 
-    fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>> {
+    fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>, Bincode> {
         const RESERVED: usize = 0x1000;
 
         let mut dm = shale::cached::DynamicMem::new(0x10000, 0);
@@ -1669,9 +1738,11 @@ mod tests {
         })
     }
 
+    #[test_case(leaf(Vec::new(), Vec::new()) ; "empty leaf encoding")]
     #[test_case(leaf(vec![1, 2, 3], vec![4, 5]) ; "leaf encoding")]
-    #[test_case(branch(b"value".to_vec(), vec![1, 2, 3].into()) ; "branch with value")]
-    #[test_case(branch(b"value".to_vec(), None); "branch without value")]
+    #[test_case(branch(b"value".to_vec(), vec![1, 2, 3].into()) ; "branch with chd")]
+    #[test_case(branch(b"value".to_vec(), None); "branch without chd")]
+    #[test_case(branch(Vec::new(), None); "branch without value and chd")]
     #[test_case(extension(vec![1, 2, 3], DiskAddress::null(), vec![4, 5].into()) ; "extension without child address")]
     fn encode_(node: Node) {
         let merkle = create_test_merkle();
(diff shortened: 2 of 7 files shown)


rkuris (commented):
More thorough review shortly, but we should use the type system better here -- lets make a trait for each of the node types with methods 'serialize' and 'deserialize' on them.

firewood/src/merkle/node.rs
```diff
@@ -824,3 +670,182 @@ impl Storable for Node {
...
+}
+
+pub struct BinCodec;
+
+impl NodeCodec for BinCodec {
+    fn encode<S: ShaleStore<Node>>(node: &NodeType, store: &S) -> Vec<u8> {
+        match node {
+            NodeType::Leaf(n) => BinCodec::encode_leaf(n),
+            NodeType::Extension(n) => BinCodec::encode_ext(n, store),
+            NodeType::Branch(n) => BinCodec::encode_branch(n, store),
```
richardpringle, line 687: 🤔 

Something is off here, but it's not immediately obvious what it is. 

Using a ZST (Zero-Sized Type) to specify the encoding type is the right way to do things as you've done... But I think the way that you actually want to do this is more like the following:
```rust
struct Merke {
    // ...
    format: PhantomData<EncodingFormat>,
}

// shortcut so we don't have to write this out every time
type Store = ShaleStore<Node>;

// could come up with a better name here
pub trait Encode<EncodingFormat> {
    type Error;
    
    fn encode<S: Store>(self, store: S) -> Result<Vec<u8>, Self::Error>;
    
    fn decode(bytes: &[u8]) -> Result<Self, Self::Error>;
}
```
Then implement `Encode` for `NodeType`, `ExtNode`, `LeafNode`, and `BranchNode`, where `NodeType` just delegates to the inner implementation. 

I'm not sure if you need to bind `EncodingFormat` with a marker-trait. If the compiler complains, you could call the marker-trait `EncodingFormat` and change the generic argument to `E: EncodingFormat`, but *I don't think* it'll complain. 
    
However, I could be wrong about where to place the outermost `EncodingFormat` generic, but reviewing the code in this PR seems to point me toward the tree itself. 

 I think you will have to further parameterize `get_encoded` as well, but I could be wrong. When you are done with this PR, there should be no way to serialize with one `Codec` and deserialize with another.

richardpringle: Further thoughts (that I'm not completely sure of):
There may be a better way to do all of this with something like the following:
```rust
pub trait BinaryCodec {
    // depending on the Serializer and Deserializer implementations, these will likely always be the same, but should not be required to be the same
    type SerializeError;
    type DeserializeError = Self::SerializeError;

    fn serialize<T: Serialize>(&self, t: &T) -> Result<T, Self::SerializeError>;
    
    fn deserialize<'de, T: Deserialize<'de>>(&self, bytes: &[u8]) -> Result<T, Self::DeserializeError>;
}

#[derive(Default)]
struct Bincode(bincode::DefaultOptions); 

impl BinaryCodec for Bincode {
    type SerializeError: bincode::Error;

    fn serialize<T: Serialize>(&self, t: &T) -> Result<T, Self::SerializeError> {
        self.0.clone().serialize(t)
    }
    
    fn deserialize<'de, T: Deserialize<'de>>(&self, bytes: &[u8]) -> Result<T, Self::DeserializeError> {
        self.0.clone().deserialize(bytes)
    }
}
```
Or something along those lines. Then, our node types can simply implement `Serialize` and `Deserialize` and the `Merkle` type would be generic over the `BinaryCodec`. It's likely that we would want to attach a `&'static Bincode` to the database itself to avoid copying the config on every new tree (view).

richardpringle (changes requested):

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+                            let chd_encoded = self.get_encoded(&node)?;
+                            children[pos] = Some(chd_encoded);
+                        }
+                        None => {
+                            if let Some(v) = &n.children_encoded[pos] {
+                                children[pos] = Some(v.to_vec());
+                            }
+                        }
+                    }
+                }
```
richardpringle, line 118: In general, I actually find it kind of difficult to read all the branching here. This one isn't that bad, but you have several contexts to keep in your head at any one time. You're on the `Branch` branch of the match statement, then you're looping over children. Now I check to see if the child is avalailable and if not, go look for the pre-encoded child.

What if you used the functional approach instead?

```rust
                // let's look at pairs of children
                let encoded_children = n.chd().iter().zip(n.children_encoded.iter());
                let encoded_children = encoded_children.map(|(child_addr, encoded_child)| {
                    child_addr
                         // if there's a child here, get the encoded bytes
                        .map(|addr| self.get_node(addr).and_then(|node| self.get_encoded(&node)))
                        // or look for the pre-fetched bytes
                        .or_else(|| encoded_child.as_ref().map(|child| Ok(child.to_vec())))
                        .transpose()
                });
                
               // now I don't care about how I got them, but I know I have all my encoded children above

                // now create the empty array
                let mut children: Box<[Option<Vec<u8>>; MAX_CHILDREN]> = Default::default();

                // and just set each child
                for (child, encoded_child) in children.iter_mut().zip(encoded_children) {
                    *child = encoded_child?;
                }
```

With the above code, you deal with any possible branching early on, then just move on. It's just as efficient since iterators are lazy. We don't have to worry about indexing into any array, `zip` takes care of lining everything up for us. 

I know that `map`, `and_then`, `or_else`, and `transpose` aren't the easiest things to understand if you aren't used to these combinators, but it's nice to construct the iterator then think about next steps instead of doing multiple things at once.

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+            }
+
+            NodeType::Extension(_) => todo!(),
+        };
+        Ok(self.serde.serialize(&encoded).unwrap())
+    }
+
+    #[allow(dead_code)]
+    fn decode(&self, buf: &'de [u8]) -> Result<NodeType, MerkleError> {
+        let encoded: EncodedNode<T> = self.serde.deserialize(buf).unwrap();
```
richardpringle, line 135: Need to solve this unwrap. This code is absolutely fallible. I recommend (even when prototyping) to handle errors. The part that you can defer is the error struct itself. [`Infallible`](https://doc.rust-lang.org/stable/std/convert/enum.Infallible.html) is really useful for this. But here, you could just add a `MerkleError` variant.

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+        match encoded.node {
```
richardpringle, line 136: nit
```suggestion
        let encoded: EncodedNode<T> = self.serde.deserialize(buf).unwrap();

        match encoded.node {
```

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+            EncodedNodeType::Leaf(leaf) => Ok(NodeType::Leaf(LeafNode::new(leaf.path, leaf.data))),
```
richardpringle, line 137:
```suggestion
            EncodedNodeType::Leaf(leaf) => Ok(NodeType::Leaf(leaf)),
```

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+            EncodedNodeType::Branch { children, value } => {
+                let path = Vec::new().into();
+                let value = value.map(|v| v.0);
```
richardpringle, line 140: 🤔, you shouldn't need to do this, the branch is just going to wrap the `Vec` in `Data` again.

firewood/src/merkle.rs
```diff
@@ -1661,7 +1734,7 @@ mod tests {
         assert_eq!(n, nibbles);
     }
 
-    fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>> {
+    fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>, bincode::DefaultOptions> {
```
richardpringle, line 1737: A type alias would probably be nice here.

firewood/src/merkle.rs
```diff
@@ -1904,7 +1997,8 @@ mod tests {
     async fn single_value_range_proof() {
         const RANDOM_KEY: u8 = 42;
 
-        let mut merkle: Merkle<CompactSpace<Node, DynamicMem>> = create_test_merkle();
+        let mut merkle: Merkle<CompactSpace<Node, DynamicMem>, bincode::DefaultOptions> =
+            create_test_merkle();
```
richardpringle, line 2001: I don't think any of these need explicit types, do they?

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+impl Serialize for EncodedNode<bincode::DefaultOptions> {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        use serde::ser::Error;
+        let other = bincode::DefaultOptions::new();
+
+        match &self.node {
+            EncodedNodeType::Leaf(n) => {
+                let s = [
+                    Encoded::Raw(from_nibbles(&n.path.encode(true)).collect()),
+                    Encoded::Raw(n.data.to_vec()),
```
richardpringle, line 744: Did you try getting rid of `Encoded`?

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+}
+
+impl<'de> Deserialize<'de> for EncodedNode<bincode::DefaultOptions> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error;
+
+        let other = bincode::DefaultOptions::new();
```
richardpringle, line 791: Same comment about putting this into the scope where it's needed.

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+        let items: Vec<Encoded<Vec<u8>>> = Deserialize::deserialize(deserializer)?;
+        let len = items.len();
+        match len {
+            LEAF_NODE_SIZE => {
+                let mut items = items.into_iter();
+                let path = match items.next().unwrap() {
```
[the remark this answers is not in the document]
richardpringle: Actually, better to use `expect` because the comment is built-in.

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+                    Encoded::Raw(path) => path,
+                    Encoded::Data(path) => Err(D::Error::custom(format!(
+                        "incorrect encoded type for leaf node path {:?}",
+                        path
+                    )))?,
+                };
```
richardpringle, line 803: Or better yet
```suggestion
                let Some(Encoded::Raw(path)) = match items.next() else {
                    return Err(D::Error::custom("incorrect encoded type for leaf node path".into())),
                };
```

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+                Ok(Self {
+                    node,
+                    phantom: PhantomData,
+                })
+            }
+            BRANCH_NODE_SIZE => {
+                let mut items = items;
+
+                // we've already validated the size, that's why we can safely unwrap
+                let data = match items.pop().unwrap() {
```
richardpringle, line 826: I think I may have written code like this before, but it would be better to just use an iterator and get the children, then the data.

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+
+                // we popped the last element, so their should only be NBRANCH items left
+                for (i, chd) in items.into_iter().enumerate() {
+                    let chd = match chd {
+                        Encoded::Raw(chd) => chd,
+                        Encoded::Data(chd) => other
+                            .deserialize(chd.as_ref())
+                            .map_err(|e| D::Error::custom(format!("bincode error: {e}")))?,
+                    };
+                    children[i] = Some(chd).filter(|chd: &Vec<u8>| !chd.is_empty());
```
richardpringle, line 848:
```suggestion
                    children[i] = Some(chd).filter(|chd| !chd.is_empty());
```

firewood/src/proof.rs
```diff
@@ -451,7 +451,7 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
     ///                to be present.
     fn decode_node<S: ShaleStore<Node> + Send + Sync>(
         &self,
-        merkle: &Merkle<S>,
+        merkle: &Merkle<S, bincode::DefaultOptions>,
```
richardpringle, line 454: You're going to need to make these parameters generic as well.

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+                    phantom: PhantomData,
+                })
+            }
+            size => Err(D::Error::custom(format!("invalid size: {size}"))),
+        }
+    }
+}
+pub trait BinarySerde {
+    type SerializeError: Debug;
+    type DeserializeError: Debug;
```
richardpringle, line 865:
```suggestion
    type SerializeError: serde::ser::Error;
    type DeserializeError: serde::de::Error;
```

firewood/src/merkle.rs
```diff
@@ -37,6 +41,8 @@ pub enum MerkleError {
     UnsetInternal,
     #[error("error updating nodes: {0}")]
     WriteError(#[from] ObjWriteError),
+    #[error("merkle bincode error: {0:?}")]
+    Bincode(#[from] Box<bincode::ErrorKind>),
```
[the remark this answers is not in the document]
richardpringle: Could be a `Box<dyn serde::ser::Error>` for serialization and a separate variant for deserialization, `Box<dyn serde::de::Error>`

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+                Ok(Self {
+                    node,
+                    phantom: PhantomData,
+                })
+            }
+            size => Err(D::Error::custom(format!("invalid size: {size}"))),
+        }
+    }
+}
+pub trait BinarySerde {
```
richardpringle, line 863: nit: I think this trait should live in `merkle.rs`

firewood/src/merkle/node.rs
```diff
@@ -717,6 +718,176 @@ impl Storable for Node {
...
+    type SerializeError: Debug;
+    type DeserializeError: Debug;
+
+    fn serialize<T: Serialize>(&self, t: &T) -> Result<Vec<u8>, Self::SerializeError>;
+
+    fn deserialize<'de, T: Deserialize<'de>>(
+        &self,
+        bytes: &'de [u8],
+    ) -> Result<T, Self::DeserializeError>;
+}
```
richardpringle, line 873: I'm thinking it would be best if there were only one concrete way to create the serializer.
```suggestion
pub trait BinarySerde {
    type SerializeError: serde::ser::Error;
    type DeserializeError: serde::de::Error;
    
    fn new() -> Self;
    
   fn serialize<T: Serialize>(t: &T) -> Result<Vec<u8>, Self::SerializeError> {
       Self::new().serialize_impl(t)
   }

    fn deserialize<'de, T: Deserialize<'de>>(bytes: &'de [u8],) -> Result<T, Self::DeserializeError> {
        Self::new().deserialize_impl(bytes)
    }

    fn serialize_impl<T: Serialize>(t: &T) -> Result<Vec<u8>, Self::SerializeError>;

    fn deserialize_impl<'de, T: Deserialize<'de>>(
        &self,
        bytes: &'de [u8],
    ) -> Result<T, Self::DeserializeError>;
}
```

I'm not sure that this is *the best* solution, but you wan't to make sure there's only one way to construct the serializer.

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
     }
 }
 
-impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
-    pub fn new(store: Box<S>) -> Self {
-        Self { store }
+impl<'de, S: ShaleStore<Node> + Send + Sync, T: BinarySerde> Merkle<S, T>
+where
+    EncodedNode<T>: serde::Deserialize<'de>,
```
richardpringle, line 88: Please put all trait bounds in the `where` clause.

richardpringle (changes requested):
You'll have to pick through and figure out which comments are important and which aren't. I tried to mark the nits with "nit:".

xinifinity (commented):
Address most of them, working on the rest.

firewood/src/merkle.rs
```diff
@@ -71,11 +83,70 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+                            if let Some(v) = &n.children_encoded[pos] {
+                                children[pos] = Some(v.to_vec());
+                            }
+                        }
+                    }
+                }
+                EncodedNode {
+                    node: EncodedNodeType::Branch {
+                        children,
+                        value: n.value.clone(),
```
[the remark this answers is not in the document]
xinifinity: Same comment as above.

xinifinity (commented):
Sorry, had to do a force push as rebasing is a nightmare...

firewood/src/merkle/node.rs
```diff
@@ -459,6 +463,203 @@ impl Storable for Node {
     }
 }
 
+pub struct EncodedNode<T: ?Sized> {
```
rkuris, line 466: Q: What cases are we using that aren't `Sized`?

firewood/src/merkle/node/branch.rs
```diff
@@ -19,6 +19,7 @@ pub type DataLen = u32;
 pub type EncodedChildLen = u8;
 
 const MAX_CHILDREN: usize = 16;
+pub const SIZE: usize = MAX_CHILDREN + 1;
```
rkuris, line 22: nit: naming: at least CHILDREN_SIZE, but that's not quite right either.

firewood/src/merkle/node.rs
xinifinity: Good catch, I think this bound is not necessary.

firewood/src/db.rs
```diff
@@ -322,7 +322,7 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
     pub fn stream<K: KeyType>(
         &self,
         start_key: Option<K>,
-    ) -> Result<merkle::MerkleKeyValueStream<'_, S>, api::Error> {
+    ) -> Result<merkle::MerkleKeyValueStream<'_, S, bincode::DefaultOptions>, api::Error> {
```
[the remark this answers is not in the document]
xinifinity: Yeah, I plan to make it generic in a follow up when we actually has more than one implementation of `BinarySerde`.

firewood/src/merkle.rs
```diff
@@ -73,11 +79,76 @@ impl<S: ShaleStore<Node>> Merkle<S> {
...
+                        .transpose()
+                });
+
+                // now create the empty array, and set each child
+                let mut children: Box<[Option<Vec<u8>>; BranchNode::MAX_CHILDREN]> =
+                    Default::default();
+
+                for (child, encoded_child) in children.iter_mut().zip(encoded_children) {
+                    *child = encoded_child?;
+                }
```
rkuris, line 118:
```suggestion
                // pair up DiskAddresses with encoded children and pick the right one
                let children = n
                    .chd()
                    .iter()
                    .zip(n.children_encoded.iter())
                    // matched pairs of (DiskAddress, EncodedChild) here
                    .map(|(child_addr, encoded_child)| {
                        child_addr
                            // if there's a child disk address here, get the encoded bytes
                            .map(|addr| self.get_node(addr).and_then(|node| self.encode(&node)))
                            // or look for the pre-fetched bytes
                            .or_else(|| encoded_child.as_ref().map(|child| Ok(child.to_vec())))
                            .transpose()
                    })
                    .collect::<Result<Vec<Option<Vec<u8>>>, MerkleError>>()?
                    .try_into()
                    .expect("MAX_CHILDREN will always be yielded");
```

firewood/src/merkle.rs
```diff
@@ -1726,6 +1801,21 @@ mod tests {
         assert_eq!(encoded, new_node_encoded);
     }
 
+    #[test_case(leaf(Vec::new(), Vec::new()) ; "empty leaf encoding")]
+    #[test_case(leaf(vec![1, 2, 3], vec![4, 5]) ; "leaf encoding")]
+    #[test_case(branch(b"value".to_vec(), vec![1, 2, 3].into()) ; "branch with chd")]
+    #[test_case(branch(b"value".to_vec(), None); "branch without chd")]
+    #[test_case(branch(Vec::new(), None); "branch without value and chd")]
+    fn node_encode_decode_(node: Node) {
```
rkuris, line 1809: q: what does the trailing _ mean here?

firewood/src/merkle/node.rs
```diff
@@ -459,6 +463,203 @@ impl Storable for Node {
...
+                for (i, c) in children.iter().enumerate() {
+                    if let Some(c) = c {
+                        if c.len() >= TRIE_HASH_LEN {
+                            let serialized_hash =
+                                Bincode::serialize(&Keccak256::digest(c).to_vec())
+                                    .map_err(|e| S::Error::custom(format!("bincode error: {e}")))?;
+                            list[i] = Encoded::Data(serialized_hash);
+                        } else {
+                            list[i] = Encoded::Raw(c.to_vec());
+                        }
```
rkuris, line 516: You just need as `as_ref`:
```suggestion
                for (i, c) in children
                    .iter()
                    .enumerate()
                    .filter_map(|(i, c)| c.as_ref().map(|c| (i, c)))
                {
                    if c.len() >= TRIE_HASH_LEN {
                        let serialized_hash = Bincode::serialize(&Keccak256::digest(c).to_vec())
                            .map_err(|e| S::Error::custom(format!("bincode error: {e}")))?;
                        list[i] = Encoded::Data(serialized_hash);
                    } else {
                        list[i] = Encoded::Raw(c.to_vec());
```

firewood/src/merkle/node.rs
```diff
@@ -459,6 +463,203 @@ impl Storable for Node {
...
+                        "incorrect encoded type for leaf node path",
+                    ));
+                };
+                let Some(Encoded::Raw(data)) = items.next() else {
+                    return Err(D::Error::custom(
+                        "incorrect encoded type for leaf node data",
+                    ));
+                };
+                let path_nibbles = Nibbles::<0>::new(&path);
+                let path = PartialPath::from_nibbles(path_nibbles.into_iter()).0;
```
rkuris, line 558: nit: I would have combined these:
```suggestion
                let path = PartialPath::from_nibbles(Nibbles::<0>::new(&path).into_iter()).0;
```

firewood/src/merkle/node.rs
```diff
@@ -459,6 +463,203 @@ impl Storable for Node {
...
+                    node,
+                    phantom: PhantomData,
+                })
+            }
+            BranchNode::MSIZE => {
+                let mut children: [Option<Vec<u8>>; BranchNode::MAX_CHILDREN] = Default::default();
+                let mut value: Option<Data> = Default::default();
+                let len = items.len();
+
+                for (i, chd) in items.into_iter().enumerate() {
```
rkuris, line 573: nit: there are a few ways of handling the last item on  vector differently besides enumerating:

 - `let last = pop(items);` (although this requires a mutable items)
 - create an iterator from `items.chunks(MAX_CHILDREN)`, then the first item are the regular children and the last item is in a separate slice. The slices can form new iterators.

This is only a nit but it avoids the `mut children` and `mut value` as you can do map/collect for them.

firewood/src/merkle/node.rs
rkuris: we can and should move this later

firewood/tests/merkle.rs
```diff
@@ -16,7 +16,7 @@ fn merkle_build_test<K: AsRef<[u8]> + std::cmp::Ord + Clone, V: AsRef<[u8]> + Cl
     items: Vec<(K, V)>,
     meta_size: u64,
     compact_size: u64,
-) -> Result<MerkleSetup<Store>, DataStoreError> {
+) -> Result<MerkleSetup<Store, Bincode>, DataStoreError> {
```
rkuris, line 19: nit: I don't love the name here as it implies it has to be the bincode implementation, but I'm okay with this for now.

rkuris (approved):
Everything are nits, although I strongly recommend the big change in merkle.rs (collect the whole thing into the array) which IMO improves readability.

firewood/src/merkle.rs
xinifinity: Was following the test name of the other suite. Maybe @richardpringle can jump in here? :)

PR #319, opened 2023-10-13
Use bincode for serialization

rkuris:
Performance improvement of about 10%

    Performance for walfile serialization/deserialization went from taking
    14.5% of the entire time for inserting 150k rows, now reduced to about 4%

### DIFF
--- a/firewood/src/storage/mod.rs
+++ b/firewood/src/storage/mod.rs
@@ -6,6 +6,7 @@ use self::buffer::DiskBufferRequester;
 use crate::file::File;
 use nix::fcntl::{flock, FlockArg};
 use parking_lot::RwLock;
+use serde::{Deserialize, Serialize};
 use shale::{CachedStore, CachedView, SendSyncDerefMut, SpaceId};
 use std::{
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


rkuris:
Old and new flamegraphs show a dramatic improvement

flamegraph
flamegraph-old

xinifinity (approved):

firewood/src/storage/mod.rs
```diff
@@ -78,78 +79,19 @@ impl Ash {
...
-            })
-            .flat_map(|(space_id_bytes, undo_len, ash_bytes)| {
-                space_id_bytes
-                    .into_iter()
-                    .chain(undo_len.to_le_bytes())
-                    .chain(ash_bytes)
-            });
-
-        len_bytes.into_iter().chain(ash_record_bytes).collect()
+        bincode::serialize(self).unwrap().into()
```
richardpringle, line 87: We should just inline this. Types that implement Serialize shouldn't also have a serialize method.

richardpringle (approved):
Fine for now, but it would be great if you could just use Rust-Analyzer to inline both of these method calls. 

We should also put some effort into eliminating unwraps as it's just more work we are going to have to do later

PR #321, opened 2023-10-16
Add verify option to insert example

rkuris:
This example now allows you to read-verify some of the inserted values. It does make a copy of the inserted values since propose consumes the batch.


richardpringle (commented):
I think these changes make the example harder to read, to be honest.

rkuris:
> I think these changes make the example harder to read, to be honest.

Is this better?

richardpringle (approved):

PR #323, opened 2023-10-17
Make Db::new async, force use of new API

rkuris:
This somewhat-large diff pushes some of the async down the stack a little. It's not complete but it's a good stop point for the next push down.


firewood/src/db.rs
```diff
@@ -411,6 +412,10 @@ impl api::Db for Db {
     ) -> Result<Self::Proposal, api::Error> {
         block_in_place(|| self.new_proposal(batch)).map_err(Into::into)
     }
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
```
richardpringle, line 418: 🤔 I think I would prefer to see this inline even if it's repeated. But I'm definitely curious to see where this is needed.

firewood/src/db.rs
```diff
@@ -436,8 +441,19 @@ pub struct Db {
 impl Db {
     const PARAM_SIZE: u64 = size_of::<DbParams>() as u64;
 
+    #[allow(clippy::new_ret_no_self)]
+    pub async fn new<P: AsRef<Path>>(
+        db_path: P,
+        cfg: &DbConfig,
+    ) -> Result<impl api::Db, api::Error> {
```
richardpringle, line 448: There shouldn't be a forced indirection here. You should be able to use the concrete database. This is a little awkward.

firewood/src/file.rs
richardpringle, line 1: 👍

firewood/src/storage/buffer.rs
richardpringle, line 1: 👍

richardpringle, line 1: 👍

firewood/src/v2/api.rs
```diff
@@ -122,6 +122,8 @@ pub trait Db {
         &self,
         data: Batch<K, V>,
     ) -> Result<Self::Proposal, Error>;
+
+    fn as_any(&self) -> &dyn Any;
```
richardpringle, line 126: What witchcraft are you planning here?

firewood/src/v2/api.rs
```diff
@@ -172,7 +174,7 @@ pub trait DbView {
 /// [DbView], which means you can fetch values from it or
 /// obtain proofs.
 #[async_trait]
-pub trait Proposal: DbView {
+pub trait Proposal: DbView + Send + Sync {
```
richardpringle, line 177: There's nothing inherent about a proposal that should necessitate `Send` and `Sync` trait bounds. Something seems off here. It would be better if we could put extra bounds on the generic types at the location where `Send` and `Sync` are actually necessary. 

Unless `Send` and `Sync` are necessary for the actual trait implementation, but I'm not sure that that is the case, is it?

firewood/tests/db.rs
```diff
@@ -2,47 +2,28 @@
...
     ($e: ident) => {{
         let mut s = Vec::new();
-        $e.kv_root_hash().unwrap();
-        $e.kv_dump(&mut s).unwrap();
+        let boxed_any: Box<&dyn Any> = Box::new(&$e);
+        boxed_any
+            .downcast_ref::<firewood::db::Db>()
+            .unwrap()
+            .kv_dump(&mut s)
+            .unwrap();
```
richardpringle, line 20: This is gross... I see why you "have" to do it, but it doesn't make it any less gross... Something doesn't seem right here. We shouldn't have to hack together tests like this. `kv_dump` is already a hack and now we have a hack on a hack.

firewood/tests/db.rs
```diff
@@ -57,15 +38,23 @@ fn test_basic_metrics() {
...
-    let db = Db::new("test_revisions_db2", &cfg.truncate(true).build()).unwrap();
-    let metrics = db.metrics();
+    let db = firewood::db::Db::new("test_revisions_db2", &cfg.truncate(true).build())
+        .await
+        .unwrap();
+    let metrics = db
+        .as_any()
+        .downcast_ref::<firewood::db::Db>()
+        .unwrap()
+        .metrics();
```
richardpringle, line 48: If you returned the concrete type from `new`, you wouldn't have to do this. I don't see why you would want to erase the `Db` type to begin with. If we want to limit the public API, that's easy enough to do without erasing type information.

firewood/tests/db.rs
```diff
@@ -2,47 +2,28 @@
...
-}
-
-impl<P: AsRef<Path> + ?Sized> Drop for Db<'_, P> {
-    fn drop(&mut self) {
-        // if you're using absolute paths, you have to clean up after yourself
-        if self.1.as_ref().is_relative() {
-            remove_dir_all(self.1).expect("should be able to remove db-directory");
-        }
-    }
-}
```
richardpringle, line 42: Why did you delete this? There are probably better ways to do this for sure, but if we don't have some automated way to clean up after ourselves in tests, it will inevitably get missed. It's easy enough to make the `new` function `async`.

firewood/tests/db.rs
```diff
@@ -57,15 +38,23 @@ fn test_basic_metrics() {
...
+        .downcast_ref::<firewood::db::Db>()
+        .unwrap()
+        .metrics();
     assert_eq!(metrics.kv_get.hit_count.get(), 0);
-    db.kv_get("a").ok();
+    let root = db.root_hash().await.unwrap();
+    let rev = db.revision(root).await.unwrap();
+    rev.val("a").await.ok();
     assert_eq!(metrics.kv_get.hit_count.get(), 1);
 }
```
richardpringle, line 54: You create a directory without cleaning it up if I'm not mistaken

firewood/tests/db.rs
```diff
@@ -100,10 +89,11 @@ fn test_revisions() {
...
     };
     for i in 0..10 {
-        let db =
-            PersistedDb::new("test_revisions_db", &cfg.clone().truncate(true).build()).unwrap();
+        let db = PersistedDb::new("test_revisions_db", &cfg.clone().truncate(true).build())
+            .await
+            .unwrap();
         let mut dumped = VecDeque::new();
-        let mut hashes: VecDeque<TrieHash> = VecDeque::new();
+        let mut hashes: VecDeque<firewood::v2::api::HashKey> = VecDeque::new();
```
richardpringle, line 96: Nit: prefer qualified path use.
```suggestion
        let mut hashes: VecDeque<HashKey> = VecDeque::new();
```

firewood/tests/db.rs
```diff
@@ -117,48 +107,34 @@ fn test_revisions() {
                     };
                     batch.push(write);
                 }
-                let proposal = db.new_proposal(batch).unwrap();
-                proposal.commit_sync().unwrap();
+                let proposal = Arc::new(db.propose(batch).await.unwrap());
+                proposal.commit().await.unwrap();
```
richardpringle, line 111: Not for this PR, but we definitely should not require that a `Proposal` be wrapped in an `Arc` here.

firewood/tests/db.rs
```diff
@@ -117,48 +107,34 @@ fn test_revisions() {
...
-                .for_each(|(b, a)| {
-                    if &a != b {
-                        print!("{a}\n{b}");
-                        panic!("not the same");
-                    }
-                });
+            for (dump, hash) in dumped.iter().zip(hashes.iter().cloned()) {
+                let rev = db.revision(hash).await.unwrap();
+                assert_eq!(kv_dump!(rev), *dump);
+            }
```
richardpringle, line 123: wen async iterators...

firewood/tests/db.rs
```diff
@@ -117,48 +107,34 @@ fn test_revisions() {
...
-        println!("i = {i}");
+        let db = PersistedDb::new("test_revisions_db", &cfg.clone().truncate(false).build())
+            .await
+            .unwrap();
+        for (dump, hash) in dumped.iter().zip(hashes.iter().cloned()) {
+            let rev = db.revision(hash).await.unwrap();
+            assert_eq!(*dump, kv_dump!(rev), "not the same: pass {i}");
+        }
     }
 }
```
richardpringle, line 134: There are folders that are sticking around here that shouldn't be.

firewood/tests/db.rs
```diff
@@ -266,73 +231,50 @@ fn db_proposal() -> Result<(), DbError> {
...
-                key: b"another_k_1",
-                value: "another_v_1".as_bytes().to_vec(),
-            }];
-            let another_proposal = proposal.clone().propose_sync(another_batch)?;
-            let rev = another_proposal.get_revision();
-            assert_val!(rev, "k", "v");
-            assert_val!(rev, "another_k_1", "another_v_1");
-            Ok(())
-        });
-    });
```
richardpringle, line 311: I don't think that you should delete this; it's our only test that assures we can access nodes from multiple threads. While I think it's not really appropriately placed, it should remain as along as it's not properly replaced.

firewood/tests/v2api.rs
```diff
@@ -1,15 +1,15 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-use std::{error::Error, path::PathBuf, sync::Arc};
```
richardpringle, line 4: Why?

richardpringle (changes requested):
I know you've already pushed changes, but I figure I will leave this review here now anyway

firewood/tests/db.rs
rkuris: Let's chat about this. Originally, I wanted `db.propose` to return an `Arc<Proposal>` because you really can't do anything with a Proposal that isn't wrapped in an `Arc`.

rkuris: Documenting offline convo:
- I definitely feel like we should return `Arc<Proposal>` from `propose`
- @richardpringle believes we should leave it since it's easy to change later and we might have a use case for a bare proposal
- I don't see any use case for that, and I'd rather change the API earlier rather than later
- No consensus so it stays the way it is

firewood/tests/db.rs
rkuris: Yes, except it's non-deterministic. For this test to actually catch something, it should be in a loop of multiple threads running simultaneously, not just a single lookup.

firewood/tests/v2api.rs
rkuris: We need the other `v2::api::Error` type and this one conflicts. I'll clean this up some and make it clearer.

firewood/tests/db.rs
rkuris: I moved the databases into the CARGO_TARGET_DIR/tmp directory, which get wiped by `cargo clean` and others. Actually removing the test directories is a separate ticket.

firewood/tests/db.rs
richardpringle: While I will admit that my way of doing this was not the most elegant, I think it's a bit of a regression not to have test files cleaned up. I'm all for deleting this code in favour of something better, but I would argue that that "something better" is not just stuffing everything into a temp-dir to be manually cleaned by the user (or the OS on restart).

firewood/tests/db.rs
```diff
@@ -99,11 +101,18 @@ fn test_revisions() {
             .collect();
         key
     };
+
+    let mut tmpdir: PathBuf = std::env::var_os("CARGO_TARGET_DIRY")
+        .unwrap_or(temp_dir().into())
+        .into();
+    tmpdir.push("/tmp/testdb");
```
richardpringle, line 108: Is it time for `crate::test_utils`? Maybe in a follow up?

firewood/tests/db.rs
```diff
@@ -99,11 +101,18 @@ fn test_revisions() {
             .collect();
         key
     };
+
+    let mut tmpdir: PathBuf = std::env::var_os("CARGO_TARGET_DIRY")
```
richardpringle, line 105: Same type... but are you actually trying to stuff databases in `CARGO_TARGET_DIR` instead of `CARGO_TARGET_TMPDIR`?

firewood/tests/db.rs
```diff
@@ -99,11 +101,18 @@ fn test_revisions() {
...
+        .unwrap_or(temp_dir().into())
+        .into();
+    tmpdir.push("/tmp/testdb");
+
     for i in 0..10 {
-        let db =
-            PersistedDb::new("test_revisions_db", &cfg.clone().truncate(true).build()).unwrap();
+        let db = Db::new(tmpdir.clone(), &cfg.clone().truncate(true).build())
```
richardpringle, line 111:
```suggestion
        let db = Db::new(&tmpdir, &cfg.clone().truncate(true).build())
```

firewood/src/db.rs
```diff
@@ -436,13 +436,23 @@ pub struct Db {
 impl Db {
     const PARAM_SIZE: u64 = size_of::<DbParams>() as u64;
 
-    /// Open a database.
-    pub fn new<P: AsRef<Path>>(db_path: P, cfg: &DbConfig) -> Result<Self, DbError> {
-        // TODO: make sure all fds are released at the end
+    pub async fn new<P: AsRef<Path> + Send + Sync + 'static>(
```
richardpringle, line 439: This is over-constrained
```suggestion
    pub async fn new<P: AsRef<Path>>(
```

firewood/src/db.rs
```diff
@@ -436,13 +436,23 @@ pub struct Db {
...
-            let _ = std::fs::remove_dir_all(db_path.as_ref());
+            let _ = tokio::fs::remove_dir_all(db_path.as_ref()).await;
         }
 
+        block_in_place(|| Db::new_internal(db_path, cfg.clone()))
+            .map_err(|e| api::Error::InternalError(e.into()))
+    }
+
+    /// Open a database.
+    fn new_internal<P: AsRef<Path> + Send + Sync + 'static>(
```
richardpringle, line 452: This is over-constrained
```suggestion
    fn new_internal<P: AsRef<Path>>(
```

richardpringle (changes requested):

firewood/tests/db.rs
richardpringle: It's not deterministic if things are being read in parallel, but the essence of the test still assures that you can at least attempt to do it. And we haven't had this test flake out, so it appears that parallel reads are actually possible. 

The async equivalent is as follows:
```rust
    let t1 = tokio::spawn({
        let proposal = proposal.clone();

        async move {
            let another_batch = vec![BatchOp::Put {
                key: b"another_k_t1",
                value: "another_v_t1".as_bytes().to_vec(),
            }];
            let another_proposal = proposal.clone().propose(another_batch).await.unwrap();
            let rev = another_proposal.get_revision();
            assert_val!(rev, "k", "v");
            assert_val!(rev, "another_k_t1", "another_v_t1");
            // The proposal is invalid and cannot be committed
            assert!(Arc::new(another_proposal).commit().await.is_err());
        }
    });

    let t2 = tokio::spawn({
        let proposal = proposal.clone();

        async move {
            let another_batch = vec![BatchOp::Put {
                key: b"another_k_t2",
                value: "another_v_t2".as_bytes().to_vec(),
            }];
            let another_proposal = proposal.clone().propose(another_batch).await.unwrap();
            let rev = another_proposal.get_revision();
            assert_val!(rev, "k", "v");
            assert_val!(rev, "another_k_t2", "another_v_t2");
        }
    });

    let _ = tokio::join!(t1, t2);
```

firewood/tests/db.rs
richardpringle: I don't think this is working as expected. When I run tests on main, I am left with a clean working directory. When I run tests on this branch, I get artifacts created by the tests.

richardpringle:
```console
➜  firewood git:(rkuris/async-down-stack) ✗ git status 
On branch rkuris/async-down-stack
Your branch is up to date with 'origin/rkuris/async-down-stack'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        firewood/test_db_proof/
        firewood/test_db_proposal/

nothing added to commit but untracked files present (use "git add" to track)
```

rkuris: Yeah I just missed these. Fixed in the next commit.

firewood/src/db.rs
```diff
@@ -755,10 +767,10 @@ impl Db {
     }
 
     /// Create a proposal.
-    pub fn new_proposal<K: KeyType, V: ValueType>(
+    pub(crate) fn new_proposal<K: KeyType, V: ValueType>(
         &self,
         data: Batch<K, V>,
-    ) -> Result<Proposal, DbError> {
+    ) -> Result<proposal::Proposal, DbError> {
```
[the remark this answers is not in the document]
rkuris: Yes, that will be more dead code!!

firewood/src/file.rs
```diff
@@ -74,7 +75,7 @@ impl Deref for File {
     }
 }
 
-pub fn touch_dir(dirname: &str, rootdir: &Path) -> Result<PathBuf, std::io::Error> {
+pub(crate) fn touch_dir(dirname: &str, rootdir: &Path) -> Result<PathBuf, std::io::Error> {
```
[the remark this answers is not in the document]
rkuris: It's coming :)

firewood/tests/db.rs
rkuris: Yes, follow up

firewood/tests/db.rs
rkuris: In a test, but yeah, good habit

firewood/examples/insert.rs
```diff
@@ -107,7 +105,7 @@ fn get_keys_to_verify(batch: &Batch<Vec<u8>, Vec<u8>>, pct: u16) -> HashMap<Vec<
 }
 
 async fn verify_keys(
-    db: &Db,
+    db: &impl firewood::v2::api::Db,
```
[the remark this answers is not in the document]
richardpringle: Not quite.

>Note: For function parameters, generic type parameters and `impl Trait` are not exactly equivalent. With a generic parameter such as `<T: Trait>`, the caller has the option to explicitly specify the generic argument for `T` at the call site using [*GenericArgs*](https://doc.rust-lang.org/beta/reference/paths.html#paths-in-expressions), for example, `foo::<usize>(1)`. If `impl Trait` is the type of any function parameter, then the caller can't ever provide any generic arguments when calling that function. This includes generic arguments for the return type or any const generics.
>
>Therefore, changing the function signature from either one to the other can constitute a breaking change for the callers of a function.


https://doc.rust-lang.org/beta/reference/types/impl-trait.html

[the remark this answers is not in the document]
richardpringle: But again, no need to change now

firewood/src/storage/buffer.rs
richardpringle, line 1: >I don't think these changes should be committed, there's no benefit to them. We should only make tests async that actually need to be async. We can do that on an as-needed basis.

I really don't think this code should be merged, to be honest. There's no reason for it. You can just stash these changes locally and re-apply them when you need them.

firewood/src/v2/api.rs
richardpringle:
```console
➜  firewood git:(main) gh pr checkout 323
Switched to branch 'rkuris/async-down-stack'
Your branch is up to date with 'origin/rkuris/async-down-stack'.
Already up to date.
➜  firewood git:(rkuris/async-down-stack) sd '(DbView) \+ Send \+ Sync' '$1' firewood/src/v2/api.rs
➜  firewood git:(rkuris/async-down-stack) ✗ git --no-pager diff
diff --git a/firewood/src/v2/api.rs b/firewood/src/v2/api.rs
index 0ddacbe..dbc9ff4 100644
--- a/firewood/src/v2/api.rs
+++ b/firewood/src/v2/api.rs
@@ -172,7 +172,7 @@ pub trait DbView {
 /// [DbView], which means you can fetch values from it or
 /// obtain proofs.
 #[async_trait]
-pub trait Proposal: DbView + Send + Sync {
+pub trait Proposal: DbView {
     type Proposal: DbView + Proposal;

     /// Commit this revision
➜  firewood git:(rkuris/async-down-stack) ✗ cargo check -p firewood
    Checking firewood v0.0.4 (/Users/richardpringle/code/firewood/firewood)
    Finished dev [unoptimized + debuginfo] target(s) in 0.40s
➜  firewood git:(rkuris/async-down-stack) ✗
```

FYI
This might not be the case for you, but I never remember how to use `sed`, I much prefer [`sd`](https://github.com/chmln/sd).

`ripgrep` is also much better than `grep` 😉

firewood/tests/db.rs
```diff
@@ -266,73 +266,74 @@ fn db_proposal() -> Result<(), DbError> {
...
-            assert_val!(rev, "another_k_1", "another_v_1");
-            Ok(())
-        });
+            assert_val!(rev, "another_k_t2", "another_v_t2");
+            assert!(Arc::new(another_proposal).commit().await.is_err());
+        }
     });
+    let (first, second) = tokio::join!(t1, t2);
+    first.unwrap();
+    second.unwrap();
```
richardpringle, line 314: 👍

richardpringle (changes requested):
I'm going to approve this, but I really don't think the changes in `firewood/src/storage/buffer.rs` should be checked in. Tests should only be made `async` on an as-needed basis.

richardpringle:
> I'm going to approve this, but I really don't think the changes in `firewood/src/storage/buffer.rs` should be checked in. Tests should only be made `async` on an as-needed basis.

I swear I hit the approve button... weird. But `Send + Sync` needs to be removed, so this is fine like this. Would love to discuss the `buffer.rs` changes, but they shouldn't block the PR

firewood/src/storage/buffer.rs
rkuris: The problem is that a future diff will make these functions wait for I/O and will need to support async. I'm just getting an early start.

firewood/src/v2/api.rs
rkuris: Without this, you get this when you wrap the Proposal in an Arc:
```
usage of an `Arc` that is not `Send` or `Sync`
the trait `Send` is not implemented for `<impl Db as Db>::Proposal`
the trait `Sync` is not implemented for `<impl Db as Db>::Proposal`
required for `Arc<<impl Db as Db>::Proposal>` to implement `Send` and `Sync`
consider using an `Rc` instead or wrapping the inner type with a `Mutex`
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
`#[warn(clippy::arc_with_non_send_sync)]` on by default
```

richardpringle (approved):

PR #328, opened 2023-10-23
Insert test

richardpringle:
- Use leaf and branch constructors for Node
- Clean up merkle tests
- Add insert tests

I got a little carried away cleaning things up, but now it's a lot easier to add new tests

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -79,11 +79,11 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
     pub fn init_root(&self) -> Result<DiskAddress, MerkleError> {
         self.store
             .put_item(
-                Node::new(NodeType::Branch(BranchNode {
+                Node::branch(BranchNode {
                     children: [None; NBRANCH],
                     value: None,
                     children_encoded: Default::default(),
-                })),
+                }),
                 Node::max_branch_node_size(),
             )
             .map_err(MerkleError::Shale)
@@ -219,7 +219,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
             let new_branch_address = self.put_node(new_branch)?.as_ptr();
 
             if idx > 0 {
-                self.put_node(Node::new(NodeType::Extension(ExtNode {
+                self.put_node(Node::from(NodeType::Extension(ExtNode {
                     path: PartialPath(matching_path[..idx].to_vec()),
                     child: new_branch_address,
                     child_encoded: None,
@@ -347,7 +347,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                 .as_ptr();
 
             if !prefix.is_empty() {
-                self.put_node(Node::new(NodeType::Extension(ExtNode {
+                self.put_node(Node::from(NodeType::Extension(ExtNode {
                     path: PartialPath(prefix.to_vec()),
                     child: branch_address,
                     child_encoded: None,
@@ -427,10 +427,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                         // insert the leaf to the empty slot
                         // create a new leaf
                         let leaf_ptr = self
-                            .put_node(Node::new(NodeType::Leaf(LeafNode(
-                                PartialPath(key_nibbles.collect()),
-                                Data(val),
-                            ))))?
+                            .put_node(Node::leaf(PartialPath(key_nibbles.collect()), Data(val)))?
                             .as_ptr();
                         // set the current child to point to this leaf
                         node.write(|u| {
@@ -565,11 +562,11 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                 chd[idx as usize] = Some(c_ptr);
 
                 let branch = self
-                    .put_node(Node::new(NodeType::Branch(BranchNode {
+                    .put_node(Node::branch(BranchNode {
                         children: chd,
                         value: Some(Data(val)),
                         children_encoded: Default::default(),
-                    })))?
+                    }))?
                     .as_ptr();
 
                 set_parent(branch, &mut parents);
@@ -609,10 +606,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                     // from: [p: Branch] -> [b (v)]x -> [Leaf]x
                     // to: [p: Branch] -> [Leaf (v)]
                     let leaf = self
-                        .put_node(Node::new(NodeType::Leaf(LeafNode(
-                            PartialPath(Vec::new()),
-                            val,
-                        ))))?
+                        .put_node(Node::leaf(PartialPath(Vec::new()), val))?
                         .as_ptr();
                     p_ref
                         .write(|p| {
@@ -625,10 +619,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                     // from: P -> [p: Ext]x -> [b (v)]x -> [leaf]x
                     // to: P -> [Leaf (v)]
                     let leaf = self
-                        .put_node(Node::new(NodeType::Leaf(LeafNode(
-                            PartialPath(n.path.clone().into_inner()),
-                            val,
-                        ))))?
+                        .put_node(Node::leaf(PartialPath(n.path.clone().into_inner()), val))?
                         .as_ptr();
                     deleted.push(p_ptr);
                     set_parent(leaf, parents);
@@ -649,7 +640,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                             //                           \____[Leaf]x
                             // to: [p: Branch] -> [Ext] -> [Branch]
                             let ext = self
-                                .put_node(Node::new(NodeType::Extension(ExtNode {
+                                .put_node(Node::from(NodeType::Extension(ExtNode {
                                     path: PartialPath(vec![idx]),
                                     child: c_ptr,
                                     child_encoded: None,
@@ -778,7 +769,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                                     // from: [Branch] -> [Branch]x -> [Branch]
                                     // to: [Branch] -> [Ext] -> [Branch]
                                     n.children[b_idx as usize] = Some(
-                                        self.put_node(Node::new(NodeType::Extension(ExtNode {
+                                        self.put_node(Node::from(NodeType::Extension(ExtNode {
                                             path: PartialPath(vec![idx]),
                                             child: c_ptr,
                                             child_encoded: None,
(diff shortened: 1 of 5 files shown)


firewood/src/merkle/node.rs
```diff
@@ -842,3 +844,78 @@ impl Storable for Node {
         }
     }
 }
+
+#[cfg(test)]
+pub(super) mod tests {
```
richardpringle, line 849: There are other ways to do this, but this was the fastest. A lot of our unit tests need cleanup anyway, so I hope this is fine for now.

firewood/src/merkle.rs
```diff
@@ -1465,58 +1341,81 @@ mod test {
...
+        let node_ref = merkle.put_node(node).unwrap();
+        let encoded = node_ref.get_encoded(merkle.store.as_ref());
+        let new_node = Node::from(NodeType::decode(encoded).unwrap());
+        let new_node_encoded = new_node.get_encoded(merkle.store.as_ref());
+
+        assert_eq!(encoded, new_node_encoded);
+    }
+
+    #[test]
+    fn insert_and_retrieve() {
```
richardpringle, line 1379: new test

firewood/src/merkle.rs
```diff
@@ -1465,58 +1341,81 @@ mod test {
...
+
+        merkle.insert(key, val.to_vec(), root).unwrap();
+
+        let fetched_val = merkle.get(key, root).unwrap();
+
+        assert_eq!(fetched_val.as_deref(), val.as_slice().into());
+    }
+
+    #[test]
+    fn insert_and_retrieve_multiple() {
```
richardpringle, line 1394: I guess I should explain why this is necessary, what do you think?

firewood/src/merkle/node.rs
```diff
@@ -421,6 +421,20 @@ impl NodeType {
...
+            root_hash: OnceLock::new(),
+            is_encoded_longer_than_hash_len: OnceLock::new(),
+            encoded: OnceLock::new(),
+            inner,
+            lazy_dirty: AtomicBool::new(false),
+        };
+        s.rehash();
+        s
+    }
+}
```
richardpringle, line 436: Constructors `leaf` and `branch` should be used instead of `Node::new` (deleted below), but when decoding an encoded node, we're left with the arbitrary `NodeType` in some places. This is an easy way to convert between types. 

In general, there's a really bad pattern used here that doesn't work that well with Rust. Whatever needs to do the locking should wrap the types they need locked instead of having wrapper types with locks. So eventually, `NodeType` will be completely refactored into `Node` and only the `Cache` should manage locks. But that's all for another day.

firewood/src/merkle/node.rs
```diff
@@ -842,3 +844,78 @@ impl Storable for Node {
...
+        node.dehydrate(&mut bytes).unwrap();
+
+        let mut mem = PlainMem::new(node.dehydrated_len(), 0x00);
+        mem.write(0, &bytes);
+
+        let hydrated_node = Node::hydrate(0, &mem).unwrap();
+
+        assert_eq!(node, hydrated_node);
+    }
+}
```
richardpringle, line 921: These tests were taken from `merkle.rs`. I got rid of a lot of repeated code using `test_case`, but I did not change any of the logic. If you (dear reviewer) could please verify that this is the case, I would be most appreciative.

firewood/src/merkle/partial_path.rs
```diff
@@ -71,3 +71,21 @@ impl PartialPath {
...
+    #[test_case(&[0x0, 0x1, 0x2], false)]
+    #[test_case(&[0x1, 0x2], true)]
+    #[test_case(&[0x1], true)]
+    fn test_encoding(steps: &[u8], term: bool) {
+        let path = PartialPath(steps.to_vec()).encode(term);
+        let (decoded, decoded_term) = PartialPath::decode(&path);
+        assert_eq!(&decoded.0, &steps);
+        assert_eq!(decoded_term, term);
+    }
+}
```
richardpringle, line 91: This test was moved from `merkle.rs` as well.

firewood/src/merkle/trie_hash.rs
```diff
@@ -4,18 +4,15 @@
...
 };
 
 pub const TRIE_HASH_LEN: usize = 32;
+const U64_TRIE_HASH_LEN: u64 = TRIE_HASH_LEN as u64;
 
 #[derive(PartialEq, Eq, Clone)]
 pub struct TrieHash(pub [u8; TRIE_HASH_LEN]);
 
-impl TrieHash {
-    const MSIZE: u64 = 32;
```
richardpringle, line 16: Wasn't pub anyway... previous code was fragile and `MSIZE` and `TRIE_HASH_LEN` could have diverged.

firewood/src/merkle/trie_hash.rs
```diff
@@ -26,22 +23,21 @@ impl std::ops::Deref for TrieHash {
...
 
     fn dehydrated_len(&self) -> u64 {
-        Self::MSIZE
+        U64_TRIE_HASH_LEN
     }
 
-    fn dehydrate(&self, to: &mut [u8]) -> Result<(), ShaleError> {
-        Cursor::new(to).write_all(&self.0).map_err(ShaleError::Io)
+    fn dehydrate(&self, mut to: &mut [u8]) -> Result<(), ShaleError> {
+        to.write_all(&self.0).map_err(ShaleError::Io)
```
richardpringle, line 40: `&mut [u8]` already implements `std::io::Write`. No point in using a `Cursor` if we just throw it away afterwards.

firewood/src/merkle/trie_hash.rs
```diff
@@ -50,3 +46,18 @@ impl Debug for TrieHash {
...
+    #[test]
+    fn test_dehydrate() {
+        let zero_hash = TrieHash([0u8; TRIE_HASH_LEN]);
+
+        let mut to = [1u8; TRIE_HASH_LEN];
+        zero_hash.dehydrate(&mut to).unwrap();
+
+        assert_eq!(&to, &zero_hash.0);
+    }
+}
```
richardpringle, line 63: This test existed in `merkle.rs` before

firewood/src/merkle.rs
```diff
@@ -1308,136 +1299,21 @@ pub fn from_nibbles(nibbles: &[u8]) -> impl Iterator<Item = u8> + '_ {
...
+    fn to_nibbles(bytes: Vec<u8>, nibbles: Vec<u8>) {
         let n: Vec<_> = bytes.into_iter().flat_map(to_nibble_array).collect();
         assert_eq!(n, nibbles);
     }
 
-    const ZERO_HASH: TrieHash = TrieHash([0u8; TRIE_HASH_LEN]);
-
-    #[test]
-    fn test_hash_len() {
-        assert_eq!(TRIE_HASH_LEN, ZERO_HASH.dehydrated_len() as usize);
```
richardpringle, line 1330: This test was deleted. It's impossible to create a `TrieHash` who's length is different than `TRIE_HASH_LENGTH` now. 

I don't think tests exist for checking that the actual `dehydrated_len` is the serialized length (assuming that's what this means), but this test didn't test that either.

firewood/src/merkle/trie_hash.rs
rkuris: Seems meaningless.

rkuris (approved):
Almost all nits, otherwise looks good.

firewood/src/merkle.rs
```diff
@@ -219,7 +219,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
             let new_branch_address = self.put_node(new_branch)?.as_ptr();
 
             if idx > 0 {
-                self.put_node(Node::new(NodeType::Extension(ExtNode {
+                self.put_node(Node::from(NodeType::Extension(ExtNode {
```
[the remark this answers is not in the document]
richardpringle: exactly

firewood/src/merkle/node.rs
```diff
@@ -842,3 +844,78 @@ impl Storable for Node {
...
+    use super::*;
+    use shale::cached::PlainMem;
+    use test_case::test_case;
+
+    pub fn leaf(path: Vec<u8>, data: Vec<u8>) -> Node {
+        Node::leaf(PartialPath(path), Data(data))
+    }
+
+    pub fn branch(
+        repeated_disk_adderss: usize,
```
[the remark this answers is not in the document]
richardpringle: No, haven't you ever heard of an adderss? 😅

firewood/src/merkle.rs
```diff
@@ -79,11 +79,11 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
     pub fn init_root(&self) -> Result<DiskAddress, MerkleError> {
         self.store
             .put_item(
-                Node::new(NodeType::Branch(BranchNode {
+                Node::branch(BranchNode {
```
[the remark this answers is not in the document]
richardpringle: Agree with the comment, but the method isn't new, so I'll leave this one for another time (otherwise I'm growing the diff)

firewood/src/merkle.rs
```diff
@@ -1465,58 +1341,81 @@ mod test {
...
-            let new_node = Node::new(NodeType::decode(r).unwrap());
-            let new_encoded = new_node.get_encoded(merkle.store.as_ref());
-            assert_eq!(r, new_encoded);
+    fn branch(value: Vec<u8>, encoded_child: Option<Vec<u8>>) -> Node {
+        let children = Default::default();
+        let value = Some(value).map(Data);
+        let mut children_encoded = <[Option<Vec<u8>>; NBRANCH]>::default();
+
+        if let Some(child) = encoded_child {
+            children_encoded[0] = Some(child);
```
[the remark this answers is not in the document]
richardpringle: I think the former is easier to follow. If this weren't inside test-code, I would agree with getting rid of the `mut`. Since this is used for tests though, I'm not concerned about someone coming in and mutating.

firewood/src/merkle.rs
richardpringle: 🤔, you're absolutely right, however, by the same argument, one could say that many unit tests are actually covered by integration tests and are therefore unnecessary. 

There are three separate outcomes when looking at these tests together:
1. both tests pass (everything works, hurray)
2. the `_multiple` test fails and the other tests pass (something is wrong with multiple insertion specifically, it's probably something to do with splitting nodes)
3. both tests fail (nothing is working)

And with that said, I've realized that this test needs to be altered as it would not cause a split, but I am going to do that in the next PR!

firewood/src/merkle/trie_hash.rs
```diff
@@ -4,18 +4,15 @@
 use shale::{CachedStore, ShaleError, Storable};
 use std::{
     fmt::{self, Debug},
-    io::{Cursor, Write},
+    io::Write,
 };
 
 pub const TRIE_HASH_LEN: usize = 32;
+const U64_TRIE_HASH_LEN: u64 = TRIE_HASH_LEN as u64;
```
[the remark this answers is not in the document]
richardpringle: You're not sure? I hate it haha... but I think it's better than what was there before. This is all due to some shale madness though. When shale finally only deals with `usize`, the compiler will help guide us through where changes need to be made and this code will disappear. At the very least, `const` isn't `pub`, so the bad code is isolated.

firewood/src/merkle/trie_hash.rs
richardpringle: It's definitely not meaningless. It's making sure that `dehydrate` writes everything to its input parameter.

PR #329, opened 2023-10-24
More merkle.rs cleanup

richardpringle:
- Make remove and get use the same logic
- Make get_mut and get share the same logic
- Consolidate node-fetching logic

There's still a little more work to do, but the testing logic isn't trival so I left a TODO

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -32,6 +32,7 @@ use std::{
     io::{Cursor, ErrorKind, Write},
     mem::size_of,
     num::NonZeroUsize,
+    ops::Deref,
     os::fd::{AsFd, BorrowedFd},
     path::Path,
     sync::Arc,
@@ -287,10 +288,7 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
         let obj_ref = self.merkle.get(key, self.header.kv_root);
         match obj_ref {
             Err(e) => Err(api::Error::IO(std::io::Error::new(ErrorKind::Other, e))),
-            Ok(obj) => match obj {
-                None => Ok(None),
-                Some(inner) => Ok(Some(inner.as_ref().to_owned())),
-            },
+            Ok(obj) => Ok(obj.map(|inner| inner.deref().to_owned())),
         }
     }
 
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -869,10 +869,8 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         }
 
         let (found, parents, deleted) = {
-            let mut parents = Vec::new();
-            let node_ref = self.get_node_by_key(self.get_node(root)?, key, |node_ref, nib| {
-                parents.push((node_ref, nib))
-            })?;
+            let (node_ref, mut parents) =
+                self.get_node_and_parents_by_key(self.get_node(root)?, key)?;
 
             let Some(mut node_ref) = node_ref else {
                 return Ok(None);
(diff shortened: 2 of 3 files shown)


firewood/src/db.rs
richardpringle, line 1: Oddest thing... I added a test and the compiler started failing here, specifically with `inner.as_ref`... no idea why adding a test would cause that

firewood/src/merkle.rs
```diff
@@ -1418,4 +1354,57 @@ mod tests {
...
+    }
+
+    #[test]
+    fn remove_many() {
+        let mut merkle = create_test_merkle();
+        let root = merkle.init_root().unwrap();
+
+        // insert values
+        for key_val in u8::MIN..=u8::MAX {
+            let key = vec![key_val];
```
rkuris, line 1387: nit:
```suggestion
            let key = &[key_val];
```
Then change the usages to just `key` instead of `&key`.

shale/src/lib.rs
```diff
@@ -206,11 +206,17 @@ impl<'a, T: Send + Sync> ObjRef<'a, T> {
 
         Ok(())
     }
+
+    pub fn into_ptr(self) -> DiskAddress {
+        let ptr = self.deref().as_ptr();
+        ptr
```
rkuris, line 212: nit; why the temp?
```suggestion
        self.deref().as_ptr()
```

shale/src/lib.rs
```diff
@@ -206,11 +206,17 @@ impl<'a, T: Send + Sync> ObjRef<'a, T> {
...
+    }
 }
 
 impl<'a, T: Send + Sync> Deref for ObjRef<'a, T> {
     type Target = Obj<T>;
     fn deref(&self) -> &Obj<T> {
+        // TODO: Something is seriously wrong here but I'm not quite sure about the best approach for the fix
```
rkuris, line 219: Not following. What's wrong? The unwrap()? Or the idea that there should be a deref on this at all?

rkuris (changes requested):

shale/src/lib.rs
richardpringle: `unwrap` inside `deref` is horrible. The assumption is that something that implements `Deref` can simply be used as the `Deref` target. In other words, it's a wrapper type to  **strictly** adds functionality. In this case, `ObjRef` the "added" functionality is only valid if the type is initialized. In other words, it's "sometimes" added functionality. 

Anyway, I added to `into_ptr` method so that I don't have to manually drop the reference.

firewood/src/merkle.rs
```diff
@@ -864,83 +864,49 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
         }
 
-        let mut deleted = Vec::new();
-        let mut parents: Vec<(ObjRef<Node>, _)> = Vec::new();
-        let mut u_ref = self.get_node(root)?;
-        let mut nskip = 0;
-        let mut found = None;
+        let (found, parents, deleted) = {
+            let mut parents = Vec::new();
+            let node_ref = self.get_node_by_key(self.get_node(root)?, key, |node_ref, nib| {
```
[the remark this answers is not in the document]
richardpringle: I don't disagree with your reasoning, but I can't see a clean way to implement both methods without duplicating the logic. 

It's literally one line in the middle of loop, but the issue is that the one line is a mutation that depends on external (to the loop) state. 

Unless you're cool with `get_node_by_key_with_callback` as the actual implementation, then I could wrap that in `get_node_by_key` and `get_node_and_parents_by_key`. But in order to follow those methods, you're just going to look at the callback method anyway. It's more cognitive overhead to have three methods compared to one. If these methods were pub, it would be a different story. 

Unless you can think of a way to implement both methods without duplicating code and without using a closure, I think the change should stay.

Happy to discuss further though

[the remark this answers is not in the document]
richardpringle: I don't think generic-bloat is relevant in private methods. There's almost no way the compiler doesn't just inline private methods like this.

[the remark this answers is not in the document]
rkuris: Maybe add a parameter `Option<&mut parents>` and pass `None` when you don't want them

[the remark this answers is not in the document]
richardpringle: addressed:
https://github.com/ava-labs/firewood/pull/329/commits/6952f067e2ae25e0de4a42cf184f92aafff2b84a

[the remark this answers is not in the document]
rkuris: 🚀 so much better!!!

firewood/src/merkle.rs
```diff
@@ -21,6 +21,10 @@ pub use node::{BranchNode, Data, ExtNode, LeafNode, Node, NodeType, NBRANCH};
 pub use partial_path::PartialPath;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
+type ObjRef<'a> = shale::ObjRef<'a, Node>;
+type ParentRefs<'a> = Vec<(ObjRef<'a>, u8)>;
```
rkuris, line 25: Ahhh, this is soooo much nicer to read!

rkuris (approved):

firewood/src/merkle.rs
richardpringle: Clippy made me do it haha

firewood/src/merkle.rs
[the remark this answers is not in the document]
richardpringle: Agreed!

PR #330, opened 2023-10-24
Check all features

richardpringle:
Make sure clippy checks all-features


rkuris (approved):
I'm not sure I love this, since now we could theoretically have a failure that depends on some feature being turned off. Of course, I suppose in that case we'd catch it during our regular development, so I guess this is okay.

PR #335, opened 2023-10-30
Remove MSIZE from DiskAddress

rkuris:
Now actually uses the size of the struture for computing how many bytes to write.


shale/src/disk_address.rs
```diff
@@ -159,13 +160,9 @@ impl std::ops::BitAnd<usize> for DiskAddress {
     }
 }
 
-impl DiskAddress {
-    const MSIZE: u64 = 8;
```
richardpringle, line 163:
```rust
const MSIZE: u64 = size_of::<Self>() as u64;
```
Then, you can leave the rest of the code as it was. 

You still want the three places where this was previously used to be the same value, so it doesn't make sense to call the same function three different times.

richardpringle (changes requested):
I know it would be insane for someone to mess this up, but if you use the same constant, it's a little more clear that the function-return, view-size, and error all contain the exact same thing

richardpringle (approved):

xinifinity (approved):

PR #336, opened 2023-11-01
Remove TypedView trait

rkuris:
This trait was only implemented by StoredView and was always wrapped in a dyn. This removes the vtable.

The other change here requires that items that are Storable are Send, Sync, and Debug, which greatly simplifies the trait bounds.

### DIFF
--- a/shale/src/compact.rs
+++ b/shale/src/compact.rs
@@ -521,9 +521,7 @@ impl<T: Storable, M: CachedStore> CompactSpace<T, M> {
     }
 }
 
-impl<T: Storable + 'static, M: CachedStore + Send + Sync> ShaleStore<T>
-    for CompactSpace<T, M>
-{
+impl<T: Storable + 'static, M: CachedStore + Send + Sync> ShaleStore<T> for CompactSpace<T, M> {
     fn put_item(&self, item: T, extra: u64) -> Result<ObjRef<'_, T>, ShaleError> {
         let size = item.dehydrated_len() + extra;
         let addr = self.inner.write().unwrap().alloc(size)?;
--- a/shale/src/compact.rs
+++ b/shale/src/compact.rs
@@ -227,7 +227,7 @@ impl<M: CachedStore> CompactSpaceInner<M> {
         StoredView::ptr_to_obj(self.meta_space.as_ref(), ptr, CompactDescriptor::MSIZE)
     }
 
-    fn get_data_ref<U: Storable + Debug + Send + Sync + 'static>(
+    fn get_data_ref<U: Storable + 'static>(
         &self,
         ptr: DiskAddress,
         len_limit: u64,
@@ -493,12 +493,12 @@ impl<M: CachedStore> CompactSpaceInner<M> {
 }
 
 #[derive(Debug)]
-pub struct CompactSpace<T: Send + Sync, M> {
+pub struct CompactSpace<T: Storable, M> {
     inner: RwLock<CompactSpaceInner<M>>,
     obj_cache: ObjCache<T>,
 }
 
-impl<T: Storable + Send + Sync, M: CachedStore> CompactSpace<T, M> {
+impl<T: Storable, M: CachedStore> CompactSpace<T, M> {
     pub fn new(
         meta_space: Arc<M>,
         compact_space: Arc<M>,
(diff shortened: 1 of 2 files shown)


shale/src/compact.rs
```diff
@@ -227,7 +227,7 @@ impl<M: CachedStore> CompactSpaceInner<M> {
         StoredView::ptr_to_obj(self.meta_space.as_ref(), ptr, CompactDescriptor::MSIZE)
     }
 
-    fn get_data_ref<U: Storable + Debug + Send + Sync + 'static>(
```
richardpringle, line 230: 🙌

shale/src/lib.rs
```diff
@@ -93,43 +93,18 @@ pub trait CachedStore: Debug + Send + Sync {
...
 
 /// A wrapper of `TypedView` to enable writes. The direct construction (by [Obj::from_typed_view]
 /// or [StoredView::ptr_to_obj]) could be useful for some unsafe access to a low-level item (e.g.
 /// headers/metadata at bootstrap or part of [ShaleStore] implementation) stored at a given [DiskAddress]
 /// . Users of [ShaleStore] implementation, however, will only use [ObjRef] for safeguarded access.
 #[derive(Debug)]
-pub struct Obj<T: ?Sized + Send + Sync> {
-    value: Box<dyn TypedView<T>>,
+pub struct Obj<T: Storable> {
+    value: Box<StoredView<T>>,
```
richardpringle, line 102: I think you can remove the `Box` now. Up to you whether you want to do it in this PR or a follow up.

shale/src/lib.rs
```diff
@@ -93,43 +93,18 @@ pub trait CachedStore: Debug + Send + Sync {
...
-    /// reflected in the underlying [CachedStore]).
-    fn is_mem_mapped(&self) -> bool;
-}
 
 /// A wrapper of `TypedView` to enable writes. The direct construction (by [Obj::from_typed_view]
 /// or [StoredView::ptr_to_obj]) could be useful for some unsafe access to a low-level item (e.g.
 /// headers/metadata at bootstrap or part of [ShaleStore] implementation) stored at a given [DiskAddress]
 /// . Users of [ShaleStore] implementation, however, will only use [ObjRef] for safeguarded access.
 #[derive(Debug)]
-pub struct Obj<T: ?Sized + Send + Sync> {
```
richardpringle, line 127: But why? Why would this ever be bound by `?Sized`? When would you ever be dealing with something going in and out a shale-store that wasn't sized? 😞

shale/src/lib.rs
```diff
@@ -155,7 +130,7 @@ impl<T: ?Sized + Send + Sync> Obj<T> {
     }
 
     #[inline(always)]
-    pub fn from_typed_view(value: Box<dyn TypedView<T>>) -> Self {
+    pub fn from_typed_view(value: Box<StoredView<T>>) -> Self {
```
richardpringle, line 132: Same comment about the `Box`

shale/src/lib.rs
```diff
@@ -252,7 +227,7 @@ pub trait ShaleStore<T: Send + Sync> {
 /// A stored item type that can be decoded from or encoded to on-disk raw bytes. An efficient
 /// implementation could be directly transmuting to/from a POD struct. But sometimes necessary
 /// compression/decompression is needed to reduce disk I/O and facilitate faster in-memory access.
-pub trait Storable {
+pub trait Storable: Debug + Send + Sync {
```
richardpringle, line 229: Why put the bounds here? Is there anything in the trait that actually requires `Debug`, `Send`, or `Sync`? 

```suggestion
pub trait Storable {
```

I checked clippy this time too and everything passed without this.

richardpringle (changes requested):
Just one change-request (the "suggestion")

shale/src/lib.rs
rkuris: Follow up! But yeah it can probably be removed.

shale/src/lib.rs
rkuris: Hence I removed `?Sized`. There is no reason for it, and arguably `T` should be `Sized`.

richardpringle (approved):

PR #338, opened 2023-11-01
Move shale into firewood as a module

rkuris:
This is strictly a move and module import change; no code changes


firewood/tests/merkle.rs
```diff
@@ -5,10 +5,11 @@ use firewood::{
     merkle::Node,
     merkle_util::{new_merkle, DataStoreError, MerkleSetup},
     proof::ProofError,
+    // TODO: we should not be using shale from an integration test
```
richardpringle, line 8: Do you mean, "for an integration test"? I think an integration test *is* the place to use shale, isn't it? It's in the unit tests that we don't want to be using these types (but we're always tied to a `ShaleStore` implementation regardless.

richardpringle (changes requested):
Usually I wouldn't care about this, but the change from `shale::SomeType` to `crate::shale::SomeType` is going to cause necessarily tedious merge conflicts if we touch the same code. Can we try to minimize the diff my importing `crate::shale::{self, ...`, please?

firewood/tests/merkle.rs
rkuris: No. Integration tests test the public interface. We might be able to use it from a functional test though.

richardpringle (approved):
🚀

PR #339, opened 2023-11-01
Enable manual trigger of default-cache build

richardpringle:
Let's see if this works
docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/default-branch-cache.yaml
+++ b/.github/workflows/default-branch-cache.yaml
@@ -2,6 +2,7 @@
 name: default-branch-cache
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main


.github/workflows/default-branch-cache.yaml
```diff
@@ -2,6 +2,7 @@
 name: default-branch-cache
 
 on:
+  workflow_dispatch:
```
rkuris, line 5: Appears incorrect. See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
```suggestion
on workflow_dispatch:
```

nit: please link docs on these kinds of changes in the commit comments somewhere

rkuris (changes requested):

.github/workflows/default-branch-cache.yaml
richardpringle: >please link docs

Noted, updated the description. 

Your suggestion is incorrect though. I edited the file in the browser and there were no complaints

richardpringle: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-activity-types-and-filters-with-multiple-events

rkuris (approved):

PR #340, opened 2023-11-02
Rename hydrate/dehydrate to serialize/deserialize

rkuris:
These names are more rust idiomatic.

Another option for these is serialize_for_storage, which might be helpful if we need serde implementations for these types.


richardpringle (approved):
🚀

PR #341, opened 2023-11-03
DiskAddress is already a more efficient Option<>

rkuris:
This removes Option<DiskAddress> from BranchNodes

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -199,7 +199,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
             let new_node = Node::leaf(PartialPath(new_node_path.to_vec()), Data(val));
             let leaf_address = self.put_node(new_node)?.as_ptr();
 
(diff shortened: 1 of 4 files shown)


richardpringle (changes requested):
Please no... I don't disagree with the change, but please not now lol

xinifinity (approved):
LGTM, although, talked to @richardpringle offline, it seems this conflicts with his major refactor/work on Merkle with https://github.com/ava-labs/firewood/pull/344. Not sure how easy or does it make sense to rebase on top?

richardpringle (changes requested):
Upon further review, I don't think that we should do this. It doesn't make sense to have an uninitialized `DiskAddress`. The only valid place that I see this happening is related to `Extension` nodes and currently, calling `serialize` on a `DiskAddress` that has not been initialized will panic. 

Instead, `DiskAddress` should simply wrap `NonZeroUsize`, and parent types can use `Option<DiskAddress>`.

rkuris:
Lets change DiskAddress to not be an Option.

PR #342, opened 2023-11-03
Streaming iterator spike

rkuris:
This is ready for feedback.

There are still some unfinished parts, all annotated with TODO. Notably:
 - You should be able to start at a key that doesn't exist
 - You should be able to start at the beginning of the trie
 - We can be more efficient on building and maintaining current_key

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1256,9 +1256,7 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
                 };
 
                 let returned_key_value = match last_node.inner() {
-                    NodeType::Branch(branch) => {
-                        (key, branch.value.to_owned().unwrap().to_vec())
-                    }
+                    NodeType::Branch(branch) => (key, branch.value.to_owned().unwrap().to_vec()),
                     NodeType::Leaf(leaf) => (key, leaf.1.to_vec()),
                     NodeType::Extension(_) => todo!(),
                 };
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -6,7 +6,6 @@ use crate::v2::api;
 use crate::{nibbles::Nibbles, v2::api::Proof};
 use futures::Stream;
 use sha3::Digest;
-use std::pin::pin;
 use std::{
     cmp::Ordering,
     collections::HashMap,
@@ -1231,24 +1230,23 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
     type Item = Result<(Vec<u8>, Vec<u8>), api::Error>;
 
     fn poll_next(
-        self: std::pin::Pin<&mut Self>,
+        mut self: std::pin::Pin<&mut Self>,
         _cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        let mut pinned_self = pin!(self);
         // Note that this sets the key_state to StartAtBeginning temporarily
         //  - if you get to the end you get Ok(None) but can fetch again from the start
         //  - if you get an error, you'll get Err(...), but continuing to fetch starts from the top
         // If this isn't what you want, then consider using [std::iter::fuse]
-        let found_key = match std::mem::take(&mut pinned_self.key_state) {
+        let found_key = match std::mem::take(&mut self.key_state) {
             IteratorState::StartAtBeginning => todo!(),
             IteratorState::StartAtKey(key) => {
                 // TODO: support finding the next key after K
-                let root_node = pinned_self
+                let root_node = self
                     .merkle
-                    .get_node(pinned_self.merkle_root)
+                    .get_node(self.merkle_root)
                     .map_err(|e| api::Error::InternalError(e.into()))?;
 
-                let (found_node, parents) = pinned_self
+                let (found_node, parents) = self
                     .merkle
                     .get_node_and_parents_by_key(root_node, &key)
                     .map_err(|e| api::Error::InternalError(e.into()))?;
@@ -1265,7 +1263,7 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
                     NodeType::Extension(_) => todo!(),
                 };
 
-                pinned_self.key_state = IteratorState::Iterating { last_node, parents };
+                self.key_state = IteratorState::Iterating { last_node, parents };
 
                 return Poll::Ready(Some(Ok(returned_key_value)));
             }
@@ -1283,7 +1281,7 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
 
                             parents.push((last_node, child_position as u8)); // remember where we walked down from
 
-                            let current_node = pinned_self
+                            let current_node = self
                                 .merkle
                                 .get_node(child_address)
                                 .map_err(|e| api::Error::InternalError(e.into()))?;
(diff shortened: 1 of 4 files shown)


firewood/src/merkle.rs
```diff
@@ -1175,6 +1180,180 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                    .skip(1)
+                    .map(|parent| parent.1)
+                    .tuples()
+                    .map(|(hi, lo)| (hi << 4) + lo)
+                    .collect::<Vec<u8>>();
+                pinned_self.current_key = current_key.into(); // Some(current_node.inner().as_leaf().unwrap().0.to_vec());
+                pinned_self
+                    .current_key
+                    .as_mut()
+                    .unwrap()
```
richardpringle, line 1332: You have the key above, why not extend it before turning it into an `Option`?

firewood/src/merkle.rs
```diff
@@ -1175,6 +1180,180 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                    .extend(current_node.inner().as_leaf().unwrap().0.to_vec());
```
richardpringle, line 1333: You already have access to the leaf here in this part of the match arm, why not `_leaf` -> `leaf` and use `leaf.0.to_vec()`?

firewood/src/merkle.rs
```diff
@@ -1175,6 +1180,180 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+
+                NodeType::Leaf(leaf) => {
+                    Poll::Ready(Some(Ok((key.as_ref().to_vec(), leaf.1.to_vec()))))
+                }
+                NodeType::Extension(_) => todo!(),
+            };
+        }
+        // The current node might be none if the tree is empty or we happen to be at the end
+        let current_node = match pinned_self.current_node.take() {
+            None => return Poll::Ready(None),
```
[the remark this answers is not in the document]
rkuris: I can't do the empty key test yet because it's not implemented. I need to add code that walks down the leftmost branch every time and returns that key, which doesn't currently exist.

I'm thinking about not using the existing get_node_and_parents_by_key or refactoring it so it supports:
- Option<K> so you can get the lowest leaf node
- returning the next highest key if the key can't be found (maybe as an option? or wrapping in another function to discard the key if it isn't found?)

firewood/src/merkle.rs
```diff
@@ -1175,6 +1180,180 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+impl<'a, K: AsRef<[u8]> + Unpin, S: shale::ShaleStore<node::Node> + Send + Sync> Stream
+    for MerkleKeyValueStream<'a, K, S>
+{
+    type Item = Result<(Vec<u8>, Vec<u8>), api::Error>;
+
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        let mut pinned_self = pin!(self);
```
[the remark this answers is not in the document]
rkuris: I can't add `mut self` because that doesn't match the trait, so this doesn't work.

firewood/src/merkle.rs
```diff
@@ -1175,6 +1180,180 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+        }
+        // The current node might be none if the tree is empty or we happen to be at the end
+        let current_node = match pinned_self.current_node.take() {
+            None => return Poll::Ready(None),
+            Some(node) => node,
+        };
+
+        let next_node = match current_node.inner() {
+            NodeType::Branch(branch) => {
+                // previously rendered the value from a branch node, so walk down to the first child
```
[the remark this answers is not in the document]
rkuris: Yeah this was really confusing. I'm redoing it.

firewood/src/merkle.rs
rkuris: Fixed in a rewrite

firewood/src/merkle.rs
```diff
@@ -1175,6 +1180,180 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                    }
+                }
+                // recompute current_key
+                // TODO: Can we keep current_key updated as we walk the tree instead of building it from the top all the time?
+                let current_key = pinned_self
+                    .parents
+                    .iter()
+                    .skip(1)
+                    .map(|parent| parent.1)
+                    .tuples()
```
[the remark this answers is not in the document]
rkuris: I didn't move this to the nibbles code yet, but did get rid of `tuples`

firewood/src/merkle.rs
```diff
@@ -1388,6 +1569,33 @@ mod tests {
...
+        }
+
+        let mut it = merkle.get_iter(Some([u8::MIN]), root).unwrap();
+        for k in u8::MIN..=u8::MAX {
+            let next = it.next().await.unwrap().unwrap();
+            assert_eq!(next.0, next.1);
+            assert_eq!(next.1, vec![k]);
+        }
+        assert!(it.next().await.is_none())
+    }
```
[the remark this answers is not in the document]
rkuris: This test will happen after `StartAtBeginning` is implemented in the next PR.

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: I was able to get this to work!

firewood/src/merkle.rs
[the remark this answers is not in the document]
richardpringle: That's only one case. There are cases here that can be tested now. The logic that has been added here should be accompanied by tests to test the cases.

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,220 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                                    let addr = children[child_position as usize].unwrap();
+
+                                    // we push (node, u8::MAX) because we will add 1, which will wrap to 0
+                                    let child = self
+                                        .merkle
+                                        .get_node(addr)
+                                        .map(|node| (node, u8::MAX))
+                                        .map_err(|e| api::Error::InternalError(e.into()))?;
+
+                                    // TODO: If the branch has a value, then we shouldn't keep_going
```
xinifinity, line 1338: Hmm, would it be better to fix this issue in this PR instead of leaving a TODO? It seems a straightforward fix.

firewood/src/merkle.rs
```diff
@@ -1270,10 +1487,12 @@ pub fn from_nibbles(nibbles: &[u8]) -> impl Iterator<Item = u8> + '_ {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use futures::StreamExt;
     use node::tests::{extension, leaf};
     use shale::{cached::DynamicMem, compact::CompactSpace, CachedStore};
     use std::sync::Arc;
     use test_case::test_case;
+    //use itertools::Itertools;
```
xinifinity, line 1495: nit: Why not remove this line?

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,220 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
     pub fn flush_dirty(&self) -> Option<()> {
         self.store.flush_dirty()
     }
+
+    pub fn get_iter<K: AsRef<[u8]>>(
+        &self,
+        key: Option<K>,
+        root: DiskAddress,
+    ) -> Result<MerkleKeyValueStream<'_, S>, MerkleError> {
+        // TODO: if DiskAddress::is_null() ...
```
xinifinity, line 1187: Will all this TODOs be tracked in a issue so that we won't forget about it? All of them seem a must do.

xinifinity (commented):
Mostly about the TODOs which some of them would be straightforward to fix in this PR, I'm good with tracking all non fixed TODOs in a issue as well, since they seem essential to have.

rkuris:
Abandoning this in favor of #346, which contains all these commits

firewood/src/merkle.rs
rkuris: This was fixed in #346

firewood/src/merkle.rs
rkuris: Fixed in #346

firewood/src/merkle.rs
rkuris: Fixed in #346

PR #343, opened 2023-11-06
make storable generic

richardpringle:
We were always using `Cursor` before, but we should be able to write to bytes to anything that implements the `std::io::Write` trait. 

- Make `to_dehydrated` generic
- Make `serialize` generic over Write


richardpringle:
Can't seem to get this to work (which is worrying)

PR #344, opened 2023-11-07
Add path to Branch-Nodes (don't use Extension-Nodes)

richardpringle:
- Add path to branch-node
- Split leaf, branch, and extension node files
- NodeType should box the branch-node
- Use leaf-node size
- Use an enum instead of numbers
- Refactor node-serialization
- WIP

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -25,6 +25,8 @@ pub use proof::{Proof, ProofError};
 pub use stream::MerkleKeyValueStream;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
+use self::node::write_branch;
+
 type NodeObjRef<'a> = shale::ObjRef<'a, Node>;
 type ParentRefs<'a> = Vec<(NodeObjRef<'a>, u8)>;
 type ParentAddresses = Vec<(DiskAddress, u8)>;
@@ -276,6 +278,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             .zip(n_path.iter())
             .position(|(a, b)| a != b);
 
+        #[allow(clippy::indexing_slicing)]
         let new_child_address = if let Some(idx) = split_index {
             // paths diverge
             let new_split_node_path = n_path.split_at(idx + 1).1;
@@ -298,10 +301,8 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
             let mut chd = [None; BranchNode::MAX_CHILDREN];
 
-            #[allow(clippy::indexing_slicing)]
             let last_matching_nibble = matching_path[idx];
-            #[allow(clippy::indexing_slicing)]
-            (chd[last_matching_nibble as usize] = Some(leaf_address));
+            chd[last_matching_nibble as usize] = Some(leaf_address);
 
             let address = match &node_to_split.inner {
                 NodeType::Extension(u) if u.path.len() == 0 => {
@@ -311,8 +312,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                 _ => node_to_split_address,
             };
 
-            #[allow(clippy::indexing_slicing)]
-            (chd[n_path[idx] as usize] = Some(address));
+            chd[n_path[idx] as usize] = Some(address);
 
             let new_branch = Node::from_branch(BranchNode {
                 path: PartialPath(matching_path[..idx].to_vec()),
@@ -516,11 +516,12 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
                     let overlap = Overlap::from(&n.path, &key_remainder);
 
+                    #[allow(clippy::indexing_slicing)]
                     match (overlap.unique_a.len(), overlap.unique_b.len()) {
                         // same node, overwrite the data
                         (0, 0) => {
                             node.write(|node| {
-                                *node.inner.data_mut() = Data(val);
+                                node.inner.set_data(Data(val));
                                 node.rehash();
                             })?;
                         }
@@ -638,6 +639,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                 }
 
                 NodeType::Branch(n) if n.path.len() == 0 => {
+                    #[allow(clippy::indexing_slicing)]
                     match n.children[next_nibble as usize] {
                         Some(c) => (node, c),
                         None => {
@@ -649,15 +651,18 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                                     Data(val),
                                 )))?
                                 .as_ptr();
+
                             // set the current child to point to this leaf
-                            #[allow(clippy::unwrap_used)]
-                            node.write(|u| {
-                                #[allow(clippy::indexing_slicing)]
-                                let uu = u.inner.as_branch_mut().unwrap();
-                                uu.children[next_nibble as usize] = Some(leaf_ptr);
-                                u.rehash();
-                            })
-                            .unwrap();
+                            #[allow(clippy::indexing_slicing)]
+                            node.write(|node| {
+                                node.rehash();
+
+                                let f = write_branch(|branch| {
+                                    branch.children[next_nibble as usize] = Some(leaf_ptr);
+                                });
+
+                                f(node)
+                            })?;
 
                             break None;
                         }
@@ -672,11 +677,12 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
                     let overlap = Overlap::from(&n.path, &key_remainder);
 
+                    #[allow(clippy::indexing_slicing)]
                     match (overlap.unique_a.len(), overlap.unique_b.len()) {
                         // same node, overwrite the data
                         (0, 0) => {
                             node.write(|node| {
-                                *node.inner.data_mut() = Data(val);
+                                node.inner.set_data(Data(val));
                                 node.rehash();
                             })?;
 
@@ -705,11 +711,16 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                                     ));
 
                                     let new_leaf = self.put_node(new_leaf)?.as_ptr();
-                                    node.write(|node| {
-                                        let branch = node.inner.as_branch_mut().unwrap();
-                                        branch.children[next_nibble as usize] = Some(new_leaf);
 
+                                    #[allow(clippy::indexing_slicing)]
+                                    node.write(|node| {
                                         node.rehash();
+
+                                        let f = write_branch(|branch| {
+                                            branch.children[next_nibble as usize] = Some(new_leaf);
+                                        });
+
+                                        f(node)
                                     })?;
 
                                     break None;
@@ -1240,29 +1251,20 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             let (node, mut parents) =
                 self.get_node_and_parents_by_key(self.get_node(root)?, key)?;
 
-            let Some(node) = node else {
+            let Some(mut node) = node else {
                 return Ok(None);
             };
 
             let data = match &node.inner {
-                NodeType::Branch(n) => {
-                    let data = n.value.clone();
+                NodeType::Branch(branch) => {
+                    let data = branch.value.clone();
+                    let children = branch.children;
 
                     if data.is_none() {
                         return Ok(None);
                     }
 
-                    // TODO: fix mutability issue here
-                    let mut node = self.get_node(node.as_ptr())?;
-
-                    node.write(|branch| {
-                        branch.inner.as_branch_mut().unwrap().value = None;
-                    })?;
-
-                    let branch = node.inner.as_branch().unwrap();
-
-                    let children: Vec<_> = branch
-                        .children
+                    let children: Vec<_> = children
                         .iter()
                         .enumerate()
                         .filter_map(|(i, child)| child.map(|child| (i, child)))
@@ -1272,6 +1274,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                     if children.len() == 1 && !parents.is_empty() {
                         let branch_path = &branch.path.0;
 
+                        #[allow(clippy::indexing_slicing)]
                         let (child_index, child) = children[0];
                         let mut child = self.get_node(child)?;
 
@@ -1292,7 +1295,15 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
                         deleted.push(node.as_ptr());
                     } else {
-                        node.write(|node| node.rehash())?
+                        node.write(|node| {
+                            node.rehash();
+
+                            let f = write_branch(|branch| {
+                                branch.value = None;
+                            });
+
+                            f(node)
+                        })?
                     }
 
                     data
@@ -1304,13 +1315,17 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                     // TODO: handle unwrap better
                     deleted.push(node.as_ptr());
 
-                    let (mut parent, child_index) = parents.pop().unwrap();
+                    let (mut parent, child_index) = parents.pop().expect("parents is never empty");
 
-                    parent.write(|parent| {
-                        parent.inner.as_branch_mut().unwrap().children[child_index as usize] = None;
-                    })?;
+                    #[allow(clippy::indexing_slicing)]
+                    parent.write(write_branch(|parent| {
+                        parent.children[child_index as usize] = None;
+                    }))?;
 
-                    let branch = parent.inner.as_branch().unwrap();
+                    let branch = parent
+                        .inner
+                        .as_branch()
+                        .expect("parents are always branch nodes");
 
                     let children: Vec<_> = branch
                         .children
@@ -1324,6 +1339,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                         (1, None, true) => {
                             let parent_path = &branch.path.0;
 
+                            #[allow(clippy::indexing_slicing)]
                             let (child_index, child) = children[0];
                             let child = self.get_node(child)?;
 
@@ -1584,6 +1600,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                         });
                     };
 
+                    #[allow(clippy::indexing_slicing)]
                     match n.children[nib as usize] {
                         Some(c) => c,
                         None => return Ok(None),
@@ -1926,6 +1943,7 @@ impl<'a, T: PartialEq> Overlap<'a, T> {
     fn from(a: &'a [T], b: &'a [T]) -> Self {
         let mut split_index = 0;
 
+        #[allow(clippy::indexing_slicing)]
         for i in 0..std::cmp::min(a.len(), b.len()) {
             if a[i] != b[i] {
                 break;
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -512,7 +512,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                         .chain(key_nibbles.clone())
                         .collect::<Vec<_>>();
 
-                    let overlap = Overlap::from(&n.path, &key_remainder);
+                    let overlap = PrefixOverlap::from(&n.path, &key_remainder);
 
                     #[allow(clippy::indexing_slicing)]
                     match (overlap.unique_a.len(), overlap.unique_b.len()) {
@@ -669,7 +669,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                         .chain(key_nibbles.clone())
                         .collect::<Vec<_>>();
 
-                    let overlap = Overlap::from(&n.path, &key_remainder);
+                    let overlap = PrefixOverlap::from(&n.path, &key_remainder);
 
                     #[allow(clippy::indexing_slicing)]
                     match (overlap.unique_a.len(), overlap.unique_b.len()) {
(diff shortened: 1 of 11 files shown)


firewood/src/merkle.rs
```diff
@@ -248,7 +232,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                             node_to_split,
                             |u| {
                                 match &mut u.inner {
-                                    NodeType::Leaf(u) => u.1 = Data(val),
+                                    NodeType::Leaf(u) => u.data = Data(val),
```
rkuris, line 235: nit: renaming `LeafNode(key, data)` to `LeafNode{key, data}` is great, but should be in a seperate diff.

firewood/src/merkle.rs
```diff
@@ -277,7 +261,9 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                                             result = Err(e);
                                         }
                                     }
-                                    NodeType::Branch(_) => unreachable!(),
+                                    NodeType::Branch(u) => {
```
rkuris, line 264: why is this possible now? A comment would be helpful

firewood/src/merkle.rs
```diff
@@ -319,7 +306,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                     }
                     // insert path is greather than the path of the leaf
                     (Ordering::Greater, Some(n_value)) => {
-                        let leaf = Node::leaf(
+                        let leaf = Node::from_leaf(
```
rkuris, line 309: This is from PartialPath and Data, not from a LeafNode, which is what `from_leaf` implies. Maybe `Node::leaf_from_path_and_data`

firewood/src/merkle.rs
```diff
@@ -338,28 +325,17 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
-                self.put_node(Node::from(NodeType::Extension(ExtNode {
-                    path: PartialPath(prefix.to_vec()),
-                    child: branch_address,
-                    child_encoded: None,
-                })))?
-                .as_ptr()
-            } else {
-                branch_address
-            }
+            self.put_node(Node::from_branch(BranchNode {
```
rkuris, line 332: (comment only) this is fine, and quite different semantics from `from_leaf`, which doesn't require a `LeafNode`

firewood/src/merkle.rs
```diff
@@ -425,31 +400,11 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                 // For a Branch node, we look at the child pointer. If it points
                 // to another node, we walk down that. Otherwise, we can store our
                 // value as a leaf and we're done
```
rkuris, line 402: This comment seems related to the deleted code below

firewood/src/merkle.rs
```diff
@@ -465,6 +420,87 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                     break None;
                 }
 
+                NodeType::Branch(n) if n.path.len() == 0 => {
```
rkuris, line 423: Can you move this back? I can't easily tell what changed. Alternatively, do the move in one commit and then make the changes in another.

firewood/src/merkle.rs
```diff
@@ -473,7 +509,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                         .collect::<Vec<_>>();
                     let n_path_len = n_path.len();
 
-                    if let Some(v) = self.split(
+                    if let Some((node, v)) = self.split(
```
rkuris, line 512: nit: rename `v` (preferably in another diff)

rkuris (changes requested):

firewood/src/merkle.rs
```diff
@@ -120,9 +113,9 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         Ok(if let Some(root) = root {
             let mut node = self.get_node(root)?;
             let res = node.get_root_hash::<S>(self.store.as_ref()).clone();
-            if node.lazy_dirty.load(Relaxed) {
+            if node.is_dirty() {
```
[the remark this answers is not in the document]
richardpringle: I explicitly said that this was the case. The issue is that I moved around files to make things easier to reason about, and I did that when I already had a bunch of "branch with path" code in place. 

Again, I will state that the goal here is to finish the implementation, know that it works (priority number one), and then break down the PR into smaller PRs if and only if it makes it easier to review. 

Again, I was the first person to identify that I made the mistake of adding the path without properly spiking on all the places where it was necessary to make the changes. I had LARGE swaths of code that were in a half-working state when I realized that I needed to refactor some of this other code to make it easier to grok. I continued to refactor as I went further down the rabbit hole, with the explicit priority of solving my initial goal. A lot of this refactoring was necessary because of floating constants that only made sense based on some other code in some random place. As I was trying to solve the task at hand, it became unwieldy to traverse up and down the large files. It was also easier to edit multiple files at the same time when they were split up.

firewood/src/merkle.rs
richardpringle: This is a mistake on my part, I should be passing in a `Leaf` there.

firewood/src/merkle.rs
richardpringle: See comment above

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: I know. I'm just marking those areas that can use this refactor, and why it's just a nit.

firewood/src/merkle.rs
richardpringle: Because branch nodes can split now

firewood/src/merkle.rs
```diff
@@ -1675,7 +2237,14 @@ mod tests {
             let key = &[key_val];
             let val = &[key_val];
 
-            let removed_val = merkle.remove(key, root).unwrap();
+            if key_val == 254 {
+                dbg!("hello");
+            }
```
richardpringle, line 2242: Delete me

firewood/src/merkle/node.rs
```diff
@@ -146,11 +153,19 @@ impl NodeType {
...
         }
     }
+
+    pub fn data_mut(&mut self) -> &mut Data {
+        match self {
+            NodeType::Branch(u) => u.value.as_mut().unwrap(),
+            NodeType::Leaf(node) => &mut node.data,
+            NodeType::Extension(_) => unreachable!(),
+        }
+    }
```
richardpringle, line 168: Not sure I used this anywhere, should delete

firewood/src/merkle/node.rs
```diff
@@ -229,7 +244,7 @@ impl Node {
                 is_encoded_longer_than_hash_len: OnceLock::new(),
                 inner: NodeType::Branch(
                     BranchNode {
-                        // path: vec![].into(),
+                        path: vec![].into(),
```
richardpringle, line 247: 🤔  I have to see where this is being called. The branch-node size is dynamic now so this function doesn't really make sense.

firewood/src/merkle/node.rs
```diff
@@ -945,34 +977,26 @@ pub(super) mod tests {
...
+    #[test_case(leaf(vec![0x01, 0x02, 0x03], vec![0x04, 0x05]); "leaf_node")]
     #[test_case(extension(vec![0x01, 0x02, 0x03], DiskAddress::from(0x42), None); "extension with child address")]
     #[test_case(extension(vec![0x01, 0x02, 0x03], DiskAddress::null(), vec![0x01, 0x02, 0x03].into()) ; "extension without child address")]
-    #[test_case(extension_from_hash(None, Some(vec![0x01, 0x02, 0x03]), Some(false), vec![0x01, 0x02, 0x03], DiskAddress::from(0x42), None); "extension with encoded")]
-    #[test_case(extension_from_hash(Some(TrieHash([0x01; 32])), None, Some(true), vec![0x01, 0x02, 0x03], DiskAddress::from(0x42), None); "extension with hash")]
-    #[test_case(branch(0x0a, b"hello world".to_vec().into(), None); "branch with data")]
-    #[test_case(branch(0x0a, None, vec![0x01, 0x02, 0x03].into()); "branch without data")]
-    #[test_case(branch_with_hash(Some(TrieHash([0x01; 32])), None, Some(true), 0x0a, b"hello world".to_vec().into(), None); "branch with hash value")]
-    #[test_case(branch_with_hash(None, Some(vec![0x01, 0x02, 0x03]), Some(false), 0x0a, b"hello world".to_vec().into(), None); "branch with encoded")]
-    fn test_encoding(node: Node) {
```
richardpringle, line 975: @xinifinity, I'm going to have to get you to explain what you're doing with these. I think you unnecessarily combined tests. We should test both with and without hashes if that makes sense. 

In any case, removing your test changes made rebasing easier. I figured I could add them back before this PR is merged.

firewood/src/merkle/node/branch.rs
```diff
@@ -57,16 +58,16 @@ impl Debug for BranchNode {
 
 impl BranchNode {
     pub const MAX_CHILDREN: usize = MAX_CHILDREN;
-    pub const MSIZE: usize = Self::MAX_CHILDREN + 1;
+    pub const MSIZE: usize = Self::MAX_CHILDREN + 2;
```
richardpringle, line 61: What is this `2`... There should at least be a comment.

firewood/src/merkle/node/branch.rs
```diff
@@ -112,6 +113,10 @@ impl BranchNode {
     pub(super) fn decode(buf: &[u8]) -> Result<Self, Error> {
         let mut items: Vec<Encoded<Vec<u8>>> = bincode::DefaultOptions::new().deserialize(buf)?;
 
+        let path = items.pop().unwrap().decode()?;
```
richardpringle, line 116: Need to handle the empty `items` case with an error.

firewood/src/merkle/node/branch.rs
```diff
@@ -140,8 +142,8 @@ impl BranchNode {
     }
 
     pub(super) fn encode<S: ShaleStore<Node>>(&self, store: &S) -> Vec<u8> {
-        // TODO: add path to encoded node
-        let mut list = <[Encoded<Vec<u8>>; Self::MAX_CHILDREN + 1]>::default();
+        // path + children + value
```
richardpringle, line 145: This comment should be re-ordered to match the order in which the values are encoded.

firewood/src/merkle/proof.rs
```diff
@@ -440,6 +450,17 @@ fn locate_subproof(
...
         }
         NodeType::Branch(n) => {
+            let partial_path = &n.path.0;
+
+            let does_not_match = key_nibbles.size_hint().0 < partial_path.len()
+                || !partial_path
+                    .iter()
+                    .all(|val| key_nibbles.next() == Some(*val));
+
+            if dbg!(does_not_match) {
```
richardpringle, line 460: whoops

firewood/src/merkle/proof.rs
```diff
@@ -512,13 +533,30 @@ fn unset_internal<K: AsRef<[u8]>, S: ShaleStore<Node> + Send + Sync, T: BinarySe
...
+    let mut fork_left = Ordering::Equal;
+    let mut fork_right = Ordering::Equal;
 
     let mut index = 0;
     loop {
         match &u_ref.inner() {
+            #[allow(clippy::indexing_slicing)]
             NodeType::Branch(n) => {
+                // If either the key of left proof or right proof doesn't match with
+                // shortnode, stop here and the forkpoint is the shortnode.
```
richardpringle, line 545: fix `shortnode` comment as that's an extension or a leaf

firewood/src/merkle/proof.rs
```diff
@@ -512,13 +533,30 @@ fn unset_internal<K: AsRef<[u8]>, S: ShaleStore<Node> + Send + Sync, T: BinarySe
...
+                let path = &*n.path;
+
+                if path.len() > 0 {
```
richardpringle, line 548: Add a comment saying you can't call chunks on an empty slice.

firewood/src/merkle/proof.rs
```diff
@@ -640,6 +732,9 @@ fn unset_internal<K: AsRef<[u8]>, S: ShaleStore<Node> + Send + Sync, T: BinarySe
                 return Ok(false);
             }
 
+            let node = n.chd();
+            let cur_key = n.path.clone().into_inner();
```
richardpringle, line 736: This looks wrong... I think I just moved this code. I think we can borrow the path here.

firewood/src/merkle/stream.rs
```diff
@@ -97,35 +102,63 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+
+                            node_result
+                                .map(|node| match dbg!(node.inner()) {
+                                    NodeType::Branch(branch) => {
+                                        let mut partial_path_iter = branch.path.iter();
+                                        let next_nibble = nibbles
+                                            .map(|nibble| (Some(nibble), partial_path_iter.next()))
+                                            .find(|(a, b)| a.as_ref() != *b);
+
+                                        match dbg!(next_nibble) {
```
richardpringle, line 126: Gotta scan for all the `dbg!` I've left behind

firewood/src/shale/compact.rs
```diff
@@ -310,6 +313,13 @@ impl<M: CachedStore> CompactSpaceInner<M> {
         let mut offset = addr - hsize;
         let header_payload_size = {
             let header = self.get_header(DiskAddress::from(offset as usize))?;
+
+            if header.is_freed {
+                dbg!(addr);
+                dbg!(hsize);
+                dbg!(offset);
+            }
```
richardpringle, line 321: delete me

firewood/tests/merkle.rs
```diff
@@ -64,12 +66,19 @@ fn test_root_hash_fuzz_insertions() -> Result<(), DataStoreError> {
...
         let mut items = Vec::new();
         for _ in 0..10 {
             let val: Vec<u8> = (0..8).map(|_| rng.borrow_mut().gen()).collect();
             items.push((keygen(), val));
         }
+
+        if i == 6 {
+            eprintln!("{:?}", &items);
+            dbg!("this one breaks");
+        }
```
richardpringle, line 80: delete me

firewood/tests/merkle.rs
```diff
@@ -169,38 +173,59 @@ fn test_root_hash_random_deletions() -> Result<(), DataStoreError> {
...
         items_ordered.shuffle(&mut *rng.borrow_mut());
         let mut merkle = new_merkle(0x100000, 0x100000);
+
         for (k, v) in items.iter() {
+            dbg!(&k, &v);
             merkle.insert(k, v.to_vec())?;
         }
+
+        for (k, v) in items.iter() {
+            dbg!(k);
```
richardpringle, line 196: delete me

firewood/src/merkle/node.rs
richardpringle: I guess the value could have always grown too, so the behaviour is consistent

firewood/src/merkle/node.rs
```diff
@@ -315,6 +330,16 @@ impl Node {
     }
 }
 
+pub(crate) fn write_branch<F: FnOnce(&mut Box<BranchNode>)>(f: F) -> impl FnOnce(&mut Node) {
```
rkuris, line 333: Unless I'm missing something, this function:
 - is named "write_branch" but does not write anything
 - seems to convert a Node into a Box<BranchNode>
 - wraps some arbitrary function with the mapping logic

This signature appears to be needlessly complicated: all callers but one seem to save then invoke the returned closure. The one that doesn't expects a closure, which would be pretty easy to code.

Maybe a method on `Node` that does the heavy lifting would improve readability (inside `impl Node`):

```
    // Assume this node is a branch, and return a mutable boxed BranchNode
    // Caution: panics if you call it with a different kind of node
    pub(crate) fn as_branch_mut(&mut self) -> &mut Box<BranchNode> {
        self.inner_mut().as_branch_mut().expect("must be a branch node")
    }
```
This changes this call sequence, repeated a few times in the code:
```
let f = write_branch(|branch| {
    branch.children[next_nibble as usize] = Some(new_leaf);
});
f(node)
```
Into:
```
node.as_branch_mut().children[next_nibble as usize] = Some(new_leaf);
```

xinifinity (commented):
Do you mind updating the description accordingly with the latest changes? Thanks!

firewood/src/merkle.rs
```diff
@@ -1472,6 +1931,37 @@ pub fn from_nibbles(nibbles: &[u8]) -> impl Iterator<Item = u8> + '_ {
     nibbles.chunks_exact(2).map(|p| (p[0] << 4) | p[1])
 }
 
+#[derive(Debug)]
+struct Overlap<'a, T> {
```
xinifinity, line 1935: I think here deserves a comment to indicate where to check the overlap part. If I understand correctly based on `from`, we are only checking from the start and don't care about overlap if in the middle or end.

xinifinity (commented):
Partial review as a first pass.

firewood/src/merkle.rs
```diff
@@ -1564,14 +2089,14 @@ mod tests {
...
+    #[test_case(Bincode::new(), branch(b"", b"value", None); "branch without chd with Bincode")]
+    #[test_case(Bincode::new(), branch_without_data(b"", None); "branch without value and chd with Bincode")]
     #[test_case(PlainCodec::new(), leaf(Vec::new(), Vec::new()) ; "empty leaf encoding with PlainCodec")]
     #[test_case(PlainCodec::new(), leaf(vec![1, 2, 3], vec![4, 5]) ; "leaf encoding with PlainCodec")]
-    #[test_case(PlainCodec::new(), branch(Some(b"value".to_vec()), vec![1, 2, 3].into()) ; "branch with chd with PlainCodec")]
-    #[test_case(PlainCodec::new(), branch(Some(b"value".to_vec()), Some(Vec::new())); "branch with empty chd with PlainCodec")]
-    #[test_case(PlainCodec::new(), branch(Some(Vec::new()), vec![1, 2, 3].into()); "branch with empty value with PlainCodec")]
+    #[test_case(PlainCodec::new(), branch(b"", b"value", vec![1, 2, 3].into()) ; "branch with chd with PlainCodec")]
+    #[test_case(PlainCodec::new(), branch(b"", b"value", Some(Vec::new())); "branch with empty chd with PlainCodec")]
+    #[test_case(PlainCodec::new(), branch(b"", b"", vec![1, 2, 3].into()); "branch with empty value with PlainCodec")]
```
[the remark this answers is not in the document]
rkuris: Rather than rolling each permutation by hand, the `test_case` crate also has `test_matrix` which can test all the combinations for you:
```
   #[test_matrix(
        [b"", b"key"],
        [b"", b"value"],
        [None, Some(vec![1,2,3])]
    )]
    fn encodes_then_decode_branches(key: &'static [u8], value: &'static [u8], vec: Option<Vec<u8>>) {
        encode(branch(key, value, vec));
    }

    #[test_matrix(
        [vec![], vec![1, 2. 3]],
        [vec![], vec![1, 2, 3]]
    )]
    fn encode_then_decode_leaves(key: Vec<u8>, value: Vec<u8>) {
        encode(leaf(key, value))
    }
```

firewood/src/merkle.rs
```diff
@@ -1564,14 +2075,14 @@ mod tests {
 
     #[test_case(Bincode::new(), leaf(Vec::new(), Vec::new()) ; "empty leaf encoding with Bincode")]
     #[test_case(Bincode::new(), leaf(vec![1, 2, 3], vec![4, 5]) ; "leaf encoding with Bincode")]
-    #[test_case(Bincode::new(), branch(Some(b"value".to_vec()), vec![1, 2, 3].into()) ; "branch with chd with Bincode")]
-    #[test_case(Bincode::new(), branch(Some(b"value".to_vec()), None); "branch without chd with Bincode")]
-    #[test_case(Bincode::new(), branch(None, None); "branch without value and chd with Bincode")]
+    #[test_case(Bincode::new(), branch(b"", b"value", vec![1, 2, 3].into()) ; "branch with chd with Bincode")]
+    #[test_case(Bincode::new(), branch(b"", b"value", None); "branch without chd with Bincode")]
```
rkuris, line 2079: Missing this failing test case:
```
#[test_case(Bincode::new(), branch(b"", b"value", Some(Vec::new())); "branch with empty chd with Bincode")]
```

Strongly recommend using this to catch them all:
```
 #[test_matrix(
        [PlainCodec::new(), Bincode::new()],
        [
            leaf(Vec::new(), Vec::new()),
            leaf(vec![1, 2, 3], vec![4, 5]),
            branch(b"", b"value", None),
            branch(b"", b"value", Some(Vec::new())),
            branch(b"", b"value", vec![1, 2, 3].into()),
        ]
    )]
```

firewood/src/merkle.rs
richardpringle: Good call out. I'll change the name.

firewood/src/merkle.rs
richardpringle: This is failing on `main` and is out of the scope of this PR.
I created a new issue to cover this: https://github.com/ava-labs/firewood/issues/512
As stated in the issue description, it's unclear whether or not it's worth spending time fixing the failing test as an empty vector is never a valid encoded-node.

firewood/src/merkle.rs
```diff
@@ -1528,17 +2018,44 @@ mod tests {
         create_generic_test_merkle::<Bincode>()
     }
 
-    fn branch(value: Option<Vec<u8>>, encoded_child: Option<Vec<u8>>) -> Node {
+    fn branch(path: &[u8], value: &[u8], encoded_child: Option<Vec<u8>>) -> Node {
```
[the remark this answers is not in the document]
richardpringle: I don't think this comment is relevant anymore due to changes that landed in `main`. Please re-open if that's not the case.

firewood/src/merkle.rs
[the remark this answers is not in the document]
richardpringle: See issue #512 for the `test-matrix` comment. 

@xinifinity 
>It doesn't seem we have a test case for branch with non empty path?

Correct, the new encoding/decoding implementation isn't currently used anywhere. Can we push the implementation out after this PR lands? It's impossible that this doesn't get caught because all the proofs will start failing once we switch to the new encoding/decoding. It should also be trivial to implement

firewood/src/merkle.rs
```diff
@@ -1547,11 +2064,19 @@ mod tests {
...
-    #[test_case(branch(None, None); "branch without value and chd")]
+    #[test_case(branch(b"", b"value", vec![1, 2, 3].into()) ; "branch with chd")]
+    #[test_case(branch(b"", b"value", None); "branch without chd")]
+    #[test_case(branch_without_data(b"", None); "branch without value and chd")]
+    #[test_case(branch(b"", b"", None); "branch without path value or children")]
+    #[test_case(branch(b"", b"value", None) ; "branch with value")]
+    #[test_case(branch(&[2], b"", None); "branch with path")]
+    #[test_case(branch(b"", b"", vec![1, 2, 3].into()); "branch with children")]
+    #[test_case(branch(&[2], b"value", None); "branch with path and value")]
+    #[test_case(branch(b"", b"value", vec![1, 2, 3].into()); "branch with value and children")]
```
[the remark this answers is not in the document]
richardpringle: Please use permalinks... the code changed and now I don't know what line you were talking about. 
Maybe [this one](https://github.com/ava-labs/firewood/pull/344/files#diff-6a6e300f3d9050248dbafa719531f14eabecc1bfac2169bb2527c81ee5474ecaR2071)?

Looks the same. This will get fixed with #512

firewood/src/merkle.rs
```diff
@@ -1630,6 +2155,60 @@ mod tests {
...
+
+            let fetched_val = merkle.get(key, root).unwrap();
+
+            assert_eq!(fetched_val.as_deref(), val.as_slice().into());
+        }
+
+        for (key, val) in key_val {
+            let fetched_val = merkle.get(key, root).unwrap();
+
+            assert_eq!(fetched_val.as_deref(), val.as_slice().into());
```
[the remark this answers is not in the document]
richardpringle: We insert a value, then make sure that it's still there. Then we check to see if all the values are there at the end. In an improper implementation, it's possible that inserting new nodes will improperly move nodes (through splitting etc) and values will be inadvertently moved.

xinifinity (approved):

rkuris:
We decided to land this despite completing the review as it is for some reason blocking adding a configuration flag to enable extension nodes

PR #345, opened 2023-11-10
Rkuris/range proofs

rkuris:
The generic N can't be specified by the caller, since the method needs to create owned things of type N, so this generic was removed from the trait.

There is a subtle difference between `Into::into` and `Box::new` in that the former is not Send for some reason.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -305,7 +305,7 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
             .map_err(|e| api::Error::IO(std::io::Error::new(ErrorKind::Other, e)))
     }
 
-    async fn range_proof<K: api::KeyType, V, N>(
+    async fn range_proof<K: api::KeyType, V>(
         &self,
         _first_key: Option<K>,
         _last_key: Option<K>,
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -322,7 +322,7 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
     ) -> Result<merkle::MerkleKeyValueStream<'_, S>, api::Error> {
         self.merkle
             .get_iter(start_key, self.header.kv_root)
-            .map_err(|e| api::Error::InternalError(e.into()))
+            .map_err(|e| api::Error::InternalError(Box::new(e)))
     }
 
     fn flush_dirty(&mut self) -> Option<()> {
@@ -454,7 +454,7 @@ impl Db {
         }
 
         block_in_place(|| Db::new_internal(db_path, cfg.clone()))
-            .map_err(|e| api::Error::InternalError(e.into()))
+            .map_err(|e| api::Error::InternalError(Box::new(e)))
     }
 
     /// Open a database.
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -351,7 +351,7 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
 
         // remove the last key from middle and do a proof on it
         let last_key = match middle.pop() {
-            None => return Err(api::Error::RangeProofError("Range too small".to_owned())),
+            None => return Err(api::Error::RangeTooSmall),
             Some((last_key, _)) => self.single_key_proof(last_key).await,
         };
 
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -29,7 +29,6 @@ use crate::{
 use async_trait::async_trait;
 use bytemuck::{cast_slice, AnyBitPattern};
 
-use futures::{future::ready, StreamExt as _, TryStreamExt as _};
 use metered::metered;
 use parking_lot::{Mutex, RwLock};
 use std::{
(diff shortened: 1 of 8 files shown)


firewood/src/db.rs
```diff
@@ -301,17 +306,76 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+    ) -> Result<Option<api::RangeProof<Vec<u8>, Vec<u8>>>, api::Error> {
+        let mut it = self.stream(first_key.as_ref())?;
+        let first_key = match it.next().await {
+            None => return Ok(None),
+            Some(Err(e)) => return Err(e),
+            Some(Ok((key, _))) => self.single_key_proof(key).await?,
+        };
+        let Some(first_key) = first_key else {
+            return Ok(None);
+        };
```
richardpringle, line 323:
```suggestion
        let Some((key, _)) = it.next().await.transpose()? else {
            return Ok(None);
        };
        
        let Some(first_key) = self.single_key_proof(key).await? else {
            return Ok(None);
        };
```

firewood/src/db.rs
```diff
@@ -301,17 +306,76 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+        let mut found = 0;
+
+        // middle is mutable because the last key will be removed before returning
```
richardpringle, line 326: This comment isn't necessary

firewood/src/db.rs
```diff
@@ -301,17 +306,76 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+                            && kv
+                                .as_ref()
+                                .map(|(k, _)| {
+                                    <Vec<u8> as AsRef<[u8]>>::as_ref(k)
+                                        < last_key.as_ref().unwrap().as_ref()
+                                }) // (b2)
+                                .unwrap_or(false)) // (b3)
+                })
+            })
+            .try_collect::<Vec<(Vec<u8>, Vec<u8>)>>()
```
richardpringle, line 349: I think what you really want is:
```rust
        let mut middle = stream
            .take_while(|kv_result| {
                let Some(last_key) = last_key.as_ref() else {
                    return ready(true);
                };

                let result = kv_result
                    .as_ref()
                    .map(|(key, _)| key.as_slice() < last_key.as_ref())
                    .unwrap_or(true); // always return Err(_)

                ready(result)
            })
            .take(limit)
            .try_collect::<Vec<_>>()
            .await?;
```

firewood/src/db.rs
```diff
@@ -301,17 +306,76 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+            .await?;
+
+        let last_key = match middle.pop() {
+            None => return Err(api::Error::RangeProofError("Range too small".to_owned())),
```
richardpringle, line 353: If there's only ever one message in a `RangeProofError`, we should just have this message as part of the display. 

Alternatively, it's easy to put `&'static str` as the data-type in the Error-variant.

richardpringle (changes requested):

firewood/src/db.rs
rkuris: Very nice! Love it!

firewood/src/db.rs
```diff
@@ -301,17 +306,76 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+        let Some(first_key) = first_key else {
+            return Ok(None);
+        };
+        let mut found = 0;
+
+        // middle is mutable because the last key will be removed before returning
+        let mut middle = it
+            .take_while(|kv| {
+                future::ready({
+                    // we stop if this returns true. We return true if:
```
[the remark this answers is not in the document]
rkuris: This was fixed. This PR isn't really ready for review yet.

firewood/src/db.rs
rkuris: I refactored this. Take should probably be first, since it's way cheaper.

firewood/src/db.rs
```diff
@@ -301,17 +306,76 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+        };
+        let Some(first_key) = first_key else {
+            return Ok(None);
+        };
+        let mut found = 0;
+
+        // middle is mutable because the last key will be removed before returning
+        let mut middle = it
+            .take_while(|kv| {
+                future::ready({
```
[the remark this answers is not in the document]
rkuris: I was able to shorten it quite a bit so I don't know that this is still necessary.

xinifinity (commented):
Sorry, was trying to review, but not sure if the PR has the correct base and which code is committed, do you mind doing another rebase with the current code? Thanks!

rkuris:
> Sorry, was trying to review, but not sure if the PR has the correct base and which code is committed, do you mind doing another rebase with the current code? Thanks!

Done! Was a bit of a challenge to rebase so I just cherry picked and force pushed.

firewood/src/db.rs
```diff
@@ -305,13 +306,64 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+        };
+
+        let Some(first_key) = self.single_key_proof(key).await? else {
+            return Ok(None);
+        };
+
+        // we stop streaming if either we hit the limit or the key returned was larger
+        // than the largest key requested
+        let mut middle = stream
+            .take(limit.unwrap_or(usize::MAX))
```
richardpringle, line 334: This needs to come after the `take_while`

firewood/src/db.rs
```diff
@@ -305,13 +306,64 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
             .map_err(|e| api::Error::IO(std::io::Error::new(ErrorKind::Other, e)))
     }
 
-    async fn range_proof<K: api::KeyType, V, N>(
+    async fn range_proof<K: api::KeyType, V>(
```
richardpringle, line 309: This should really be implemented in `merkle.rs` and tested there as well. You should be able to generate a `RangeProof` on `Merkle` and the `Db::range_proof` should just delegate.

firewood/src/db.rs
richardpringle: See my comment above about `merkle.rs`. If we put this functionality there, it's also easy to test. we should have tests before merging. Not necessarily tests for proving a range-proof, but tests that give us confidence that the filters work properly (first-key, last-key, and limit).

firewood/src/db.rs
richardpringle: Much better

richardpringle (changes requested):
Sorry, it's definitely extra work, but I didn't think about moving this until I thought about tests. And the `take` call is in the wrong place.

firewood/src/db.rs
richardpringle: Ah, you're right, my bad. I was thinking of `filter` (my brain was playing tricks on me).

firewood/src/db.rs
xinifinity: +1

firewood/src/db.rs
```diff
@@ -305,13 +306,64 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+            .try_collect::<Vec<(Vec<u8>, Vec<u8>)>>()
+            .await?;
+
+        // remove the last key from middle and do a proof on it
+        let last_key = match middle.pop() {
+            None => return Err(api::Error::RangeTooSmall),
+            Some((last_key, _)) => self.single_key_proof(last_key).await,
+        };
+
+        let Ok(Some(last_key)) = last_key else {
```
xinifinity, line 358: nit: last_key_proof.

firewood/src/db.rs
```diff
@@ -305,13 +306,64 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
-    async fn range_proof<K: api::KeyType, V, N>(
+    async fn range_proof<K: api::KeyType, V>(
         &self,
-        _first_key: Option<K>,
-        _last_key: Option<K>,
-        _limit: usize,
-    ) -> Result<Option<api::RangeProof<K, V, N>>, api::Error> {
-        todo!()
+        first_key: Option<K>,
+        last_key: Option<K>,
```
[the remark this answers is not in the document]
rkuris: This is a good point. We fetch the first_key even if it's less than the last_key, and we should just return an empty. I created an issue to track this.

rkuris:
The linter error here is fixed in #355

xinifinity (approved):

firewood/src/db.rs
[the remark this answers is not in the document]
richardpringle: I think the current behaviour is fine. We basically don't do any work and leave it to the caller to pass in sane values. That's the nice thing about iterators

richardpringle (approved):

PR #346, opened 2023-11-11
Rkuris/streaming iterator from start

rkuris:
This walks down the leftmost branch every time, recording the path down, then figuring out what the key and value is at the bottom, then resumes.

Also refactors get_key_from_parents, as this is now done 3 times in the code, sometimes when we add a leaf key value, and other times when not.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1256,9 +1256,7 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
                 };
 
                 let returned_key_value = match last_node.inner() {
-                    NodeType::Branch(branch) => {
-                        (key, branch.value.to_owned().unwrap().to_vec())
-                    }
+                    NodeType::Branch(branch) => (key, branch.value.to_owned().unwrap().to_vec()),
                     NodeType::Leaf(leaf) => (key, leaf.1.to_vec()),
                     NodeType::Extension(_) => todo!(),
                 };
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1323,33 +1323,32 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
                 match last_node.inner() {
                     NodeType::Branch(branch) => {
                         // previously rendered the value from a branch node, so walk down to the first available child
-                        if let Some(child_position) =
+                        let Some(child_position) =
                             branch.children.iter().position(|&addr| addr.is_some())
-                        {
-                            let child_address = branch.children[child_position].unwrap();
-
-                            parents.push((last_node, child_position as u8)); // remember where we walked down from
-
-                            let current_node = self
-                                .merkle
-                                .get_node(child_address)
-                                .map_err(|e| api::Error::InternalError(e.into()))?;
-
-                            let found_key = key_from_parents(&parents);
-
-                            self.key_state = IteratorState::Iterating {
-                                // continue iterating from here
-                                last_node: current_node,
-                                parents,
-                            };
-
-                            found_key
-                        } else {
+                        else {
                             // Branch node with no children?
                             return Poll::Ready(Some(Err(api::Error::InternalError(Box::new(
                                 MerkleError::ParentLeafBranch,
                             )))));
-                        }
+                        };
+                        let child_address = branch.children[child_position].unwrap();
+
+                        parents.push((last_node, child_position as u8)); // remember where we walked down from
+
+                        let current_node = self
+                            .merkle
+                            .get_node(child_address)
+                            .map_err(|e| api::Error::InternalError(e.into()))?;
+
+                        let found_key = key_from_parents(&parents);
+
+                        self.key_state = IteratorState::Iterating {
+                            // continue iterating from here
+                            last_node: current_node,
+                            parents,
+                        };
+
+                        found_key
                     }
                     NodeType::Leaf(leaf) => {
                         let mut next = parents.pop();
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1238,7 +1238,55 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
         //  - if you get an error, you'll get Err(...), but continuing to fetch starts from the top
         // If this isn't what you want, then consider using [std::iter::fuse]
         let found_key = match std::mem::take(&mut self.key_state) {
-            IteratorState::StartAtBeginning => todo!(),
+            IteratorState::StartAtBeginning => {
+                let root_node = self
+                    .merkle
+                    .get_node(self.merkle_root)
+                    .map_err(|e| api::Error::InternalError(e.into()))?;
+                let mut last_node = root_node;
+                let mut parents = vec![];
+                loop {
+                    match last_node.inner() {
+                        NodeType::Branch(branch) => {
+                            let leftmost_position =
+                                branch.children.iter().position(|&addr| addr.is_some());
+
+                            let Some(leftmost_position) = leftmost_position else {
+                                // we already exhausted the branch node. This happens with an empty trie
+                                // ... or a corrupt one
+                                return if parents.is_empty() {
+                                    // empty trie
+                                    Poll::Ready(None)
+                                } else {
+                                    // branch with NO children, not at the top
+                                    Poll::Ready(Some(Err(api::Error::InternalError(Box::new(
+                                        MerkleError::ParentLeafBranch,
+                                    )))))
+                                };
+                            };
+
+                            let next = self
+                                .merkle
+                                .get_node(branch.children[leftmost_position].unwrap())
+                                .map_err(|e| api::Error::InternalError(e.into()))?;
+
+                            parents.push((last_node, leftmost_position as u8));
+
+                            last_node = next;
+                        }
+                        NodeType::Leaf(_) => break,
+                        NodeType::Extension(_) => todo!(),
+                    }
+                }
+
+                // last_node should have a leaf; compute the key and value
+                let current_key =
+                    key_from_parents_and_leaf(&parents, last_node.inner().as_leaf().unwrap());
+
+                self.key_state = IteratorState::Iterating { last_node, parents };
+
+                current_key
+            }
             IteratorState::StartAtKey(key) => {
                 // TODO: support finding the next key after K
                 let root_node = self
@@ -1284,10 +1332,7 @@ impl<'a, S: shale::ShaleStore<node::Node> + Send + Sync> Stream for MerkleKeyVal
                                 .get_node(child_address)
                                 .map_err(|e| api::Error::InternalError(e.into()))?;
 
-                            let found_key = parents[1..]
-                                .chunks_exact(2)
-                                .map(|parents| (parents[0].1 << 4) + parents[1].1)
-                                .collect::<Vec<u8>>();
+                            let found_key = key_from_parents(&parents);
 
                             self.key_state = IteratorState::Iterating {
                                 // continue iterating from here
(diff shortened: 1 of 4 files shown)


rkuris:
Depends on streaming iterator spike PR

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                return Poll::Ready(Some(Ok(returned_key_value)));
+            }
+            IteratorState::Iterating {
+                last_node,
+                mut parents,
+            } => {
+                match last_node.inner() {
+                    NodeType::Branch(branch) => {
+                        // previously rendered the value from a branch node, so walk down to the first available child
+                        if let Some(child_position) =
```
richardpringle, line 1323:
```rust
let Some(child_position) = branch.children.iter().position(|&addr| addr.is_some()) else {
    // Branch node with no children?
    return Poll::Ready(Some(Err(api::Error::InternalError(Box::new(
        MerkleError::ParentLeafBranch,
    )))));
};

// then the rest of the logic can follow.

File: firewood/src/merkle.rs [hunk 2]
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                                    let children = parent.inner().as_branch().unwrap().chd();
+
+                                    // we use wrapping_add here because the value might be u8::MAX indicating that
+                                    // we want to go down branch
+                                    let mut child_position = child_position.wrapping_add(1);
+
+                                    if let Some(found_offset) = children[child_position as usize..]
+                                        .iter()
+                                        .position(|&addr| addr.is_some())
+                                    {
```
richardpringle, line 1367:
```rust
let Some(found_offset) = ... else {
```
^^ this syntax would be nicer here. However, if you prefer `if let Some(..) = expression { `, please save `expression` into a variable instead of allowing it to span multiple lines.

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                        loop {
+                            match next {
+                                None => return Poll::Ready(None),
+                                Some((parent, child_position)) => {
+                                    // Assume all parents are branch nodes
+                                    let children = parent.inner().as_branch().unwrap().chd();
+
+                                    // we use wrapping_add here because the value might be u8::MAX indicating that
+                                    // we want to go down branch
+                                    let mut child_position = child_position.wrapping_add(1);
```
richardpringle, line 1362: If you made `child_position` an `Option<u8>`, you wouldn't have to have special values like this. Then instead, you can simply call `child_position.unwrap_or(0)` and the code would be a lot clearer. 

For the initial setup, the only change you would need is:
```rust
let mut next = parents.pop().map(|(parent, index)| (parent, index.into())));
```

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                        }
+                    }
+                    NodeType::Leaf(leaf) => {
+                        let mut next = parents.pop();
+                        loop {
+                            match next {
+                                None => return Poll::Ready(None),
+                                Some((parent, child_position)) => {
+                                    // Assume all parents are branch nodes
+                                    let children = parent.inner().as_branch().unwrap().chd();
```
richardpringle, line 1358: 🤔, not for this PR, but once we remove the extension, the parent array can literally just be an array of `BranchNode` instead of `NodeType`.

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                                    parents.push((parent, child_position));
+
+                                    if stop_descending {
+                                        break;
+                                    }
+                                }
+                            }
+                        }
+                        // recompute current_key
+                        // TODO: Can we keep current_key updated as we walk the tree instead of building it from the top all the time?
```
richardpringle, line 1403: Should be able to, good TODO.

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+// The default state is to start at the beginning
+impl<'a> Default for IteratorState<'a> {
+    fn default() -> Self {
+        Self::StartAtBeginning
+    }
+}
+pub struct MerkleKeyValueStream<'a, S> {
+    key_state: IteratorState<'a>,
+    merkle_root: DiskAddress,
+    merkle: &'a Merkle<S>,
```
richardpringle, line 1226: 🤔 this could be problematic. In the RPC API, expects that we create and cache iterators for RPC clients... If the iterator has an immutable borrow of `Merkle`, then we can't write to the trie while the iterator exists. 

We should leave a note about this in the doc-comments.

firewood/src/merkle.rs
```diff
@@ -1388,6 +1672,39 @@ mod tests {
...
+        let mut merkle = create_test_merkle();
+        let root = merkle.init_root().unwrap();
+
+        // insert all values from u8::MIN to u8::MAX, with the key and value the same
+        for k in u8::MIN..=u8::MAX {
+            merkle.insert([k], vec![k], root).unwrap();
+        }
+
+        for start in [Some([u8::MIN]), None] {
+            let mut it = merkle.get_iter(start, root).unwrap();
```
richardpringle, line 1695: It would be better to have separate unit tests for a variety of start cases. 
 - no starting key
 - a starting key lower than the first key in the trie
 - a starting key at the first key in the trie
 - a starting key greater than the starting key in the trie

richardpringle (changes requested):

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,285 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+    type Item = Result<(Vec<u8>, Vec<u8>), api::Error>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        // Note that this sets the key_state to StartAtBeginning temporarily
+        //  - if you get to the end you get Ok(None) but can fetch again from the start
+        //  - if you get an error, you'll get Err(...), but continuing to fetch starts from the top
+        // If this isn't what you want, then consider using [std::iter::fuse]
```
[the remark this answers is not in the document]
rkuris: Actually it is. Iterators are free to continue after returning None. The nomicon somewhere mentioned that you should not attempt to optimize the case after you return None, and that the behavior is implementation specific. I went looking for the reference but only found circumstantial stuff. I can keep digging if you don't believe me.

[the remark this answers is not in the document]
richardpringle: Discussed offline, this is an easy change to make later if we want to do that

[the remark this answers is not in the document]
richardpringle: Please add these comments to the public type that implements `Stream` then you can resolve this comment

firewood/src/merkle.rs
rkuris: I wasn't 100% sure what you were asking for here.  [This commit](https://github.com/ava-labs/firewood/pull/346/commits/dcfb3386858fba6dbd2a73b0470f76f7dedd21cf) was my interpretation. I'm not sure it's any better.

firewood/src/merkle.rs
rkuris: I coded this and found it far less readable. You end up with stuff like this (from key_from_parents):
`parents[0].1.unwrap() << 4 + parents[1].1.unwrap()`

I think this might be doable if a new type were added which contained an `ObjRef` and a `Option<u8>` with a bunch of helper methods, but that would be a lot more code. So I think this is a trade off between readability and volume of code.

Another possible approach I explored was to pre-add the 1 to the u8 when it's stored, leaving you with values from 0..=16 in the position field, but this unnecesarily complicates computing the current key.

firewood/src/merkle.rs
richardpringle: I didn't say that Case 1 wasn't covered, I said 
>It would be better to have separate unit tests

You can use `test_case` to have separate inputs and thus separate tests instead of looping over a two-element array. 

>I'm not sure why we need case 4.

To make sure that the code behaves as expected, especially in the event of a change. We need to make sure that keys are skipped properly.

richardpringle (changes requested):
There are a couple of stylistic suggestions, I guess those can all be considered nits. 

I do, however, think that if it's possible to write code that doesn't panic, we should absolutely be writing code that cannot panic, even in the event of code changes. It's always better to have the compiler uphold your invariants.

As for the tests, splitting case coverage into separate unit tests also makes refactoring and debugging much easier. Unit tests should cover a single case (within reason) such that it's obvious what failure is occurring. Right now, if we wanted to change the behaviour of the iterator starting over from the beginning, we would have to modify a test that covers multiple other cases. We could introduce a bug into our test coverage of a case that we did not want to change and cause a regression.

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,288 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+                let mut last_node = root_node;
+                let mut parents = vec![];
+                let leaf = loop {
+                    match last_node.inner() {
+                        NodeType::Branch(branch) => {
+                            if let Some((leftmost_position, leftmost_address)) = branch
+                                .children
+                                .iter()
+                                .enumerate()
+                                .find(|&addr| addr.1.is_some())
```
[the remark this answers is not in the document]
rkuris: I don't like this because it will get undone once we change DiskAddress in the node [as in this PR](https://github.com/ava-labs/firewood/pull/341). I want to just remove the unwrap later (or land #341 first).

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,288 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+
+                                    if let Some(found_offset) = found_offset {
+                                        child_position += found_offset as u8;
+                                    } else {
+                                        next = parents.pop();
+                                        continue;
+                                    }
+
+                                    let addr = children[child_position as usize].unwrap();
+
```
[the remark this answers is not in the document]
rkuris: This is another case of removing the Option being easier. Can we just land #341 ?

firewood/src/merkle.rs
```diff
@@ -1175,6 +1178,288 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+            } => {
+                let value = match last_node.inner() {
+                    NodeType::Branch(branch) => branch.value.to_owned().unwrap().to_vec(),
+                    NodeType::Leaf(leaf) => leaf.1.to_vec(),
+                    NodeType::Extension(_) => todo!(),
+                };
+
+                (found_key, value)
+            }
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
rkuris: I wasn't able to get this to work. It's a bit tricky because of what's borrowed and moved. I'll create a task for fixing this in a followup.

firewood/src/merkle.rs
xinifinity: Will the merkle tree stay immutable even after traversed one time? I think it should, as we will use a new iterator for a changed merkle?

richardpringle (approved):
I would strongly have preferred no panicking code because I don't think that #341 should be merged. I will leave a comment in that PR as to why.

rkuris:
> I would strongly have preferred no panicking code because I don't think that #341 should be merged. I will leave a comment in that PR as to why.

Fair point; lets make these changes before merging then.

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: Fixed per discussion later on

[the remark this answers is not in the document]
rkuris: Applied this

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: Changed per the suggestion

richardpringle (approved):
🥇 

Thanks for the patience

PR #357, opened 2023-11-18
README updates

rkuris:
- async is in
 - range proofs almost done
 - Clarified position on io_uring and others
 - change proofs instead of range proofs for changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,18 +117,19 @@ The focus of this milestone will be to support synchronization to other
 instances to replicate the state. A synchronization library should also
 be developed for this milestone.
 
-- [ ] :runner: Add support for Ava Labs generic test tool via grpc client
+- [x] Migrate to a fully async interface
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,18 +117,19 @@ The focus of this milestone will be to support synchronization to other
 instances to replicate the state. A synchronization library should also
 be developed for this milestone.
 
-- [ ] :runner: Add support for Ava Labs generic test tool via grpc client
+- [x] Migrate to a fully async interface
```
richardpringle, line 120: This definitely isn't done. Sure, the interface uses async signatures, but none of the actual execution is async.

richardpringle (changes requested):

README.md
richardpringle: After a discussion offline, I've changed my mind here.

richardpringle (approved):

PR #359, opened 2023-11-20
Add proof verification test

richardpringle:
Add a simple proof verification test

### DIFF
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -92,6 +92,8 @@ impl From<DbError> for ProofError {
 /// to a single proof step. If reaches an end step during proof verification,
 /// the hash value will be none, and the encoded value will be the value of the
 /// node.
+
+#[derive(Debug)]
 struct SubProof {
     encoded: Vec<u8>,
     hash: Option<[u8; 32]>,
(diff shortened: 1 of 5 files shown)


firewood/src/proof.rs
```diff
@@ -113,24 +115,33 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-        }
+        };
+
+        let value = match NodeType::decode(&value_node) {
+            Ok(NodeType::Branch(branch)) => branch.value().as_ref().map(|v| v.to_vec()),
+            Ok(NodeType::Leaf(leaf)) => leaf.data().to_vec().into(),
+            _ => return Ok(value_node.into()),
+        };
+
+        Ok(value)
```
richardpringle, line 144: There are currently tests that only pass by chance of leaf-nodes not containing paths. This assures that those will always pass

firewood/src/proof.rs
```diff
@@ -113,24 +115,33 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                 }) => hash,
                 _ => return Ok(None),
             };
-        }
+        };
+
+        let value = match NodeType::decode(&value_node) {
+            Ok(NodeType::Branch(branch)) => branch.value().as_ref().map(|v| v.to_vec()),
+            Ok(NodeType::Leaf(leaf)) => leaf.data().to_vec().into(),
+            _ => return Ok(value_node.into()),
```
richardpringle, line 141: Weird case with extension nodes here... This will go away when extension nodes go away

firewood/src/proof.rs
rkuris: scary

rkuris (approved):

PR #360, opened 2023-11-20
split node files

richardpringle:
- Add path to branch-node
- Split leaf, branch, and extension node files
- Remove path again


firewood/src/merkle/node.rs
```diff
@@ -3,27 +3,41 @@
...
         OnceLock,
     },
 };
 
+mod branch;
+mod extension;
+mod leaf;
+mod partial_path;
+
+pub use branch::{BranchNode, MAX_CHILDREN, SIZE as BRANCH_NODE_SIZE};
```
rkuris, line 24: pub(crate) or pub(super). We don't want these as part of the public API.

firewood/src/merkle/node.rs
```diff
@@ -374,14 +133,15 @@ impl NodeType {
...
-            EXT_NODE_SIZE => {
+            EXTENSION_NODE_SIZE => {
                 let mut items = items.into_iter();
                 let decoded_key: Vec<u8> = items.next().unwrap().decode()?;
 
                 let decoded_key_nibbles = Nibbles::<0>::new(&decoded_key);
 
                 let (cur_key_path, term) =
-                    PartialPath::from_nibbles(decoded_key_nibbles.into_iter());
+                    dbg!(PartialPath::from_nibbles(decoded_key_nibbles.into_iter()));
```
rkuris, line 143: remove dbg

firewood/src/merkle/node.rs
```diff
@@ -526,39 +286,45 @@ impl Node {
 
 impl Storable for Node {
     fn deserialize<T: CachedStore>(addr: usize, mem: &T) -> Result<Self, ShaleError> {
-        const META_SIZE: usize = 32 + 1 + 1;
+        const META_SIZE: usize = TRIE_HASH_LEN + 1 + 1;
```
rkuris, line 289: +1 for replacing 32 with TRIE_HASH_LEN. Any idea what these other +1s are for? Can we const them?

firewood/src/merkle/node.rs
```diff
@@ -526,39 +286,45 @@ impl Node {
...
+
         let meta_raw =
             mem.get_view(addr, META_SIZE as u64)
                 .ok_or(ShaleError::InvalidCacheView {
                     offset: addr,
                     size: META_SIZE as u64,
                 })?;
-        let attrs = meta_raw.as_deref()[32];
+
+        let attrs = meta_raw.as_deref()[TRIE_HASH_LEN];
```
rkuris, line 298: every use of `meta_raw `calls `as_deref()`. Can we move this to the assignment?

firewood/src/merkle/node.rs
```diff
@@ -727,12 +505,15 @@ impl Storable for Node {
                     .collect();
 
                 let (path, _) = PartialPath::decode(&nibbles);
-                let value = Data(remainder.as_deref()[path_len as usize..].to_vec());
-                Ok(Self::new_from_hash(
+                let data = Data(remainder.as_deref()[path_len as usize..].to_vec());
```
rkuris, line 508: Same thing with `remainder`. We immediately call `as_deref()` on every use.

firewood/src/merkle/node.rs
```diff
@@ -760,13 +542,13 @@ impl Storable for Node {
...
                     1 + 8
-                        + n.path.dehydrated_len()
+                        + n.path.serialized_len()
                         + match n.chd_encoded() {
                             Some(v) => 1 + v.len() as u64,
                             None => 1,
                         }
                 }
-                NodeType::Leaf(n) => 1 + 4 + n.0.dehydrated_len() + n.1.len() as u64,
+                NodeType::Leaf(n) => 1 + 4 + n.path.serialized_len() + n.data.len() as u64,
```
rkuris, line 551: nit: move to new diff

firewood/src/merkle/node.rs
```diff
@@ -792,6 +574,7 @@ impl Storable for Node {
 
         match &self.inner {
             NodeType::Branch(n) => {
+                // TODO: add path
```
rkuris, line 577: Not sure what this means?

rkuris (commented):
This is challenging to review, because I can't tell if changes were made to the data structures without extracting them and doing a different set of diffs.

Would it be possible to split this into 3 parts:
(1) rename leaf.0 and leaf.1 to leaf.key and leaf.data
(2) rename NBRANCH to MAX_CHILDREN
(3) move different node types to different files

firewood/src/merkle/node.rs
richardpringle: https://github.com/ava-labs/firewood/issues/362

It's a bit of a tangled web. I will do this later if it's alright with you. The `node` mod itself isn't public, so I don't think any of these can be accessed outside the crate, but I agree that we should be more strict in case we ever change the visibility of the `node` mod.

firewood/src/merkle/node.rs
richardpringle: I have more PRs coming where this gets cleaned up

firewood/src/merkle/node.rs
richardpringle: All this code is going away in a future PR

richardpringle:
@rkuris 

>Would it be possible to split this into 3 parts

It would be possible, but it's a lot of work for very little value. You have already identified the three major changes in this PR, and you can rest assured that the functionality hasn't changed when the tests pass. You'll notice that I did not change any of the integration test files here. 

You may also notice that there are currently three commits here, one of which will be very easy to remove when continuing to get the roughly 2500 lines of #344 applied. If I didn't have that many more lines to go, I wouldn't push back, but this is the tip of the ice-berg.

rkuris:
> It would be possible, but it's a lot of work for very little value. You have already identified the three major changes in this PR, and you can rest assured that the functionality hasn't changed when the tests pass. You'll notice that I did not change any of the integration test files here.

There was some value here, but to review it I had to split it up locally, identifying a few problems I would have otherwise missed; comments coming. Please consider splitting these up in the future; it literally took just over an hour.

firewood/src/merkle/node/partial_path.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl std::ops::Deref for PartialPath {
+    type Target = [u8];
+    fn deref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+impl From<Vec<u8>> for PartialPath {
```
rkuris, line 32: I don't like this addition. It's now easy for someone to put non-nibbles into a PartialPath. If you really think this is needed as a convenience, it should be TryFrom, and each byte should be checked.

This also ties PartialPath more strongly to a Vec, which is something I'd like to change.

firewood/src/merkle/node/branch.rs
```diff
@@ -0,0 +1,184 @@
...
+        let mut chd_encoded: [Option<Vec<u8>>; MAX_CHILDREN] = Default::default();
+
+        // we popped the last element, so their should only be NBRANCH items left
+        for (i, chd) in items.into_iter().enumerate() {
+            let data = chd.decode()?;
+            chd_encoded[i] = Some(data).filter(|data| !data.is_empty());
+        }
+
+        // TODO: add path
+        let path = Vec::new().into();
```
rkuris, line 124: PartialPath::new()

firewood/src/merkle/node/leaf.rs
```diff
@@ -0,0 +1,75 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use test_case::test_case;
+
+    // these tests will fail if the encoding mechanism changes and should be updated accordingly
+    #[test_case(0b10 << 4, vec![0x12, 0x34], vec![1, 2, 3, 4]; "even length")]
+    // first nibble is part of the prefix
+    #[test_case((0b11 << 4) + 2, vec![0x34], vec![2, 3, 4]; "odd length")]
+    fn encode_regression_test(prefix: u8, path: Vec<u8>, nibbles: Vec<u8>) {
```
rkuris, line 61: nit: at a minimum add a comment in the PR  that this is a new, not moved, test

rkuris: Is this just testing `partial_path`? If you're checking to see if the node is storing it correctly, rather than hardcoding this here (which can change if partial_path changes), perhaps use partial_path and compare that encoding?

firewood/src/merkle/node/leaf.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use std::fmt::{Debug, Error as FmtError, Formatter};
+
+use bincode::Options;
+
+use super::{Data, Encoded};
+use crate::merkle::{from_nibbles, PartialPath};
+
+#[derive(PartialEq, Eq, Clone)]
+pub struct LeafNode {
```
rkuris, line 12: nit: LeafNode shouldn't escape the crate, so the inner ones can just be pub, but this one should be `pub(crate)`.

firewood/src/merkle/node/leaf.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl LeafNode {
+    pub fn new<P: Into<PartialPath>, D: Into<Data>>(path: P, data: D) -> Self {
+        Self {
+            path: path.into(),
+            data: data.into(),
+        }
+    }
+
+    pub fn path(&self) -> &PartialPath {
```
rkuris, line 31: nit: since these are `pub fn` we should reduce the visibility of the members themselves, or get rid of this.

firewood/src/merkle/node/leaf.rs
rkuris, line 61: Is this just testing partial_path? If so, can't we move these tests there? If not, a better choice here would be to call partial_path in this test

firewood/src/merkle/node/partial_path.rs
```diff
@@ -0,0 +1,129 @@
...
+// See the file LICENSE.md for licensing terms.
+
+use super::Flags;
+use crate::nibbles::NibblesIterator;
+use std::{
+    fmt::{self, Debug},
+    iter::once,
+};
+
+// TODO: use smallvec
```
rkuris, line 11: comment only: We might even get by with Box<[u8]> since we almost always know how big it will be when we start. I'm not sure how often these grow.

rkuris (changes requested):

firewood/src/merkle/node/branch.rs
richardpringle: This code is temporary

firewood/src/merkle/node/leaf.rs
richardpringle: agreed, added to https://github.com/ava-labs/firewood/issues/362

firewood/src/merkle/node/leaf.rs
richardpringle: Agreed (both `path` and `data` methods are redundant

https://github.com/ava-labs/firewood/issues/364

firewood/src/merkle/node/leaf.rs
richardpringle: Yes, it looks like this is just testing partial-path

I've created https://github.com/ava-labs/firewood/issues/365 to move the test

firewood/src/merkle/node/partial_path.rs
richardpringle: `Box<[u8]>` would definitely be an improvement, but the goal of `smallvec` or something like it would be quick stack-copies since we actually copy these partial-paths a lot. 

Something like `Bytes` might be more of a benefit, though, where we store the longest keys  on the heap and reference subsets of the keys in the nodes themselves. Again, the goal is a stack-copy when creating a node (branch of leaf) instead of having to allocate copied data.

firewood/src/merkle/node/partial_path.rs
richardpringle: I created [#366](https://github.com/ava-labs/firewood/issues/366) suggesting to add convenience methods for extending and shrinking `PartialPath`. The current problem is that in order to perform any change to the path, you essentially need to copy the `PartialPath`, and then make changes to the internal `Vec` and reset the value. 

Adding `From` was a convenience for getting the rest of the "branch with path" implementation further along. I will improve this once it's complete.

firewood/src/shale/mod.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-use disk_address::DiskAddress;
```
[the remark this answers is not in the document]
richardpringle: 34199e5169205ffd9df8344bc1152b03d827e258

rkuris (approved):
approved with new issues

xinifinity (approved):

PR #361, opened 2023-11-20
Move range-proof-builder

richardpringle:
Having implemented methods in different files is confusing. We should stick to types implemented in different files only


firewood/src/merkle.rs
```diff
@@ -1190,6 +1190,79 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
             merkle: self,
         })
     }
+
+    pub(crate) async fn range_proof<K: api::KeyType + Send + Sync>(
```
rkuris, line 1194: crate->super; this method is only needed by parents of merkle, not anywhere in the crate.

rkuris (dismissed):
Let's lower the visibility back down to super.

I'm not sure I agree with the sentiment here. merkle.rs is huge already and we end up with more merge conflicts when every method is in the same file.

rkuris (approved):

PR #371, opened 2023-11-22
Use bitflags for leaf-node attributes

richardpringle:
Bitflags are a lot easier to read and reason about. I believe that they will effectively compile down to the exact same assembly.

Further, we're using an enum to check the node-type. This makes the check for a valid node-type more explicit. This is the Rust way: parse your data and early return if it's invalid, then match on verified valid types.

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -196,15 +196,23 @@ impl From<NodeType> for Node {
     }
 }
 
-impl Node {
-    const BRANCH_NODE: u8 = 0x0;
-    const EXT_NODE: u8 = 0x1;
-    const LEAF_NODE: u8 = 0x2;
+bitflags! {
+    struct NodeAttributes: u8 {
+        const ROOT_HASH_VALID      = 0b001;
+        const IS_ENCODED_BIG_VALID = 0b010;
+        const LONG                 = 0b100;
+    }
+}
 
-    const ROOT_HASH_VALID_BIT: u8 = 1 << 0;
+impl Node {
+    const BRANCH_NODE: u8 = 0;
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/node.rs
```diff
@@ -196,15 +196,23 @@ impl From<NodeType> for Node {
...
+    const EXT_NODE: u8 = 1;
+    const LEAF_NODE: u8 = 2;
 
-    const ROOT_HASH_VALID_BIT: u8 = 1 << 0;
+    const ROOT_HASH_VALID_BIT: u8 = 0b01;
     // TODO: why are these different?
-    const IS_ENCODED_BIG_VALID: u8 = 1 << 1;
-    const LONG_BIT: u8 = 1 << 2;
+    const IS_ENCODED_BIG_VALID: u8 = 0b10;
+    const LONG_BIT: u8 = 0b100;
```
richardpringle, line 215: The six `const`s above will be removed in a follow up

firewood/src/merkle/node.rs
```diff
@@ -196,15 +196,23 @@ impl From<NodeType> for Node {
...
+    }
+}
+
 impl Node {
-    const BRANCH_NODE: u8 = 0x0;
-    const EXT_NODE: u8 = 0x1;
-    const LEAF_NODE: u8 = 0x2;
+    const BRANCH_NODE: u8 = 0;
+    const LEAF_NODE: u8 = 1;
+    const EXT_NODE: u8 = 2;
```
richardpringle, line 210: Order here changes on purpose. The important part is that it is the same as below

xinifinity (approved):

PR #372, opened 2023-11-22
Use more strict types in node-(de)serialization

richardpringle:
Use more strict types in node (de)serialization for storage. Strict type conversions eliminate some potential bugs in logic at compile-time.

This is the precursor to a more major refactor, moving storage (de)serialization implementations into the individual node-types.

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -10,6 +10,7 @@ use sha3::{Digest, Keccak256};
 use std::{
     fmt::Debug,
     io::{Cursor, Read, Write},
+    mem::size_of,
     sync::{
         atomic::{AtomicBool, Ordering},
         OnceLock,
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/node.rs
```diff
@@ -305,6 +297,17 @@ impl Node {
     }
 }
 
+#[repr(C)]
+struct Meta {
+    root_hash: [u8; TRIE_HASH_LEN],
+    attrs: NodeAttributes,
+    is_encoded_longer_than_hash_len: Option<bool>,
+}
```
richardpringle, line 305: This is 100% stack-based and can actually be (de)serialized in one shot. I didn't do that here though, that can be done later if necessary.

firewood/src/merkle/node.rs
```diff
@@ -368,45 +369,55 @@ impl Storable for Node {
         match meta_raw.as_deref()[TRIE_HASH_LEN + 1].try_into()? {
             NodeTypeId::Branch => {
                 // TODO: add path
+                // TODO: figure out what this size is?
```
richardpringle, line 372: I figure it out later, I promise.

xinifinity (approved):

PR #373, opened 2023-11-22
Delegate branch-storage logic to the branch mod

richardpringle:
There is a code move here and well as some code changes. They are unfortunately in one commit. Breaking it up would mean re-implementing it and possibly introducing a bug. 

I will be doing the same thing with both leaf and extension nodes.

Upon reviewing my own code, I see that this is actually just serialization. There's a `todo!()` that will come later for deserialization.

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -22,7 +22,7 @@ mod extension;
 mod leaf;
 mod partial_path;
 
-pub use branch::{BranchNode, MAX_CHILDREN, SIZE as BRANCH_NODE_SIZE};
+pub use branch::BranchNode;
 pub use extension::ExtNode;
 pub use leaf::{LeafNode, SIZE as LEAF_NODE_SIZE};
 pub use partial_path::PartialPath;
@@ -114,7 +114,7 @@ impl NodeType {
                 }
             }
             // TODO: add path
-            BRANCH_NODE_SIZE => Ok(NodeType::Branch(BranchNode::decode(buf)?.into())),
+            BranchNode::MSIZE => Ok(NodeType::Branch(BranchNode::decode(buf)?.into())),
             size => Err(Box::new(bincode::ErrorKind::Custom(format!(
                 "invalid size: {size}"
             )))),
@@ -216,7 +216,7 @@ impl Node {
                 inner: NodeType::Branch(
                     BranchNode {
                         // path: vec![].into(),
-                        children: [Some(DiskAddress::null()); MAX_CHILDREN],
+                        children: [Some(DiskAddress::null()); BranchNode::MAX_CHILDREN],
                         value: Some(Data(Vec::new())),
                         children_encoded: Default::default(),
                     }
@@ -370,7 +370,7 @@ impl Storable for Node {
             NodeTypeId::Branch => {
                 // TODO: add path
                 // TODO: figure out what this size is?
-                let branch_header_size = MAX_CHILDREN as u64 * 8 + 4;
+                let branch_header_size = BranchNode::MAX_CHILDREN as u64 * 8 + 4;
                 let node_raw = mem.get_view(addr + Meta::SIZE, branch_header_size).ok_or(
                     ShaleError::InvalidCacheView {
                         offset: addr + Meta::SIZE,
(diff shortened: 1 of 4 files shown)


firewood/src/merkle/node.rs
```diff
@@ -408,9 +409,10 @@ impl Storable for Node {
                     ))
                 };
 
-                let mut chd_encoded: [Option<Vec<u8>>; MAX_CHILDREN] = Default::default();
+                let mut chd_encoded: [Option<Vec<u8>>; BranchNode::MAX_CHILDREN] =
+                    Default::default();
 
-                let offset = if raw_len == u32::MAX as u64 {
+                let offset = if raw_len == u32::MAX {
```
rkuris, line 415: 😆

firewood/src/merkle/node/branch.rs
```diff
@@ -21,6 +28,29 @@ pub struct BranchNode {
     pub(crate) children_encoded: [Option<Vec<u8>>; MAX_CHILDREN],
 }
 
+enum BranchDataLength {
+    None,
+    Length(u32),
+}
+
```
rkuris, line 35: This is confusing; why is this not an `Option<u32>`, or even better, `Option<NonZeroU32>`?

If zero is a valid value, consider https://crates.io/crates/nonmax

This way you get niche optimization with None so the size is smaller and serialization is easier.

firewood/src/merkle/node/branch.rs
```diff
@@ -179,3 +212,59 @@ impl BranchNode {
             .unwrap()
     }
 }
+
```
rkuris, line 215: I will have to review this later to verify it didn't change.

firewood/src/shale/disk_address.rs
```diff
@@ -30,6 +30,8 @@ impl DerefMut for DiskAddress {
 }
 
 impl DiskAddress {
+    pub const MSIZE: u64 = size_of::<Self>() as u64;
```
rkuris, line 33: why was the visibility increase needed?

rkuris (commented):
Partial review, will finish on the moved bits later

firewood/src/shale/disk_address.rs
richardpringle: It's use in (de)serialization

firewood/src/merkle/node/branch.rs
richardpringle: Would it be less confusing if it was 
```rust
enum BranchDataLength {
    NoData,
    /// `0` is a valid length, *empty* data and *no* data are treated separately
    Length(u32),
}
```
[`Shepmaster`](https://stackoverflow.com/users/155423/shepmaster) put this pretty nicely [in a comment on SO](https://stackoverflow.com/questions/30414068/can-i-use-the-null-pointer-optimization-for-my-own-non-pointer-types#comments-30414145):
>I'll just have to do more profiling to see if it's truly worth it. Using less bytes seems like an obvious win; using less bytes **and** mandatory math everywhere is less sure-fire.

(emphasis mine)

In any case, I didn't end up using this code. I'll delete it.

firewood/src/merkle/node/branch.rs
richardpringle: You actually do not have to verify that it didn't change. What you have to verify is that we have tests that make sure that things that we serialize can also be deserialized since we have no burden of backward compatibility (yet).

richardpringle: 🤔 , now that I say that... I should have explicit tests for this. This is actually tested in all the `merkle.rs` tests because of [this line](https://github.com/ava-labs/firewood/blob/313cae7852d5ea66af1b43452634a4639bd1dfec/firewood/src/merkle.rs#L1682), but that's pretty fragile.

I created a new issue: https://github.com/ava-labs/firewood/issues/375

firewood/src/merkle/node/branch.rs
richardpringle: deleted https://github.com/ava-labs/firewood/pull/373/commits/313cae7852d5ea66af1b43452634a4639bd1dfec

firewood/src/shale/disk_address.rs
rkuris: `pub(crate)` please

rkuris (approved):
approved, almost all nits

xinifinity (approved):

richardpringle:
@rkuris, I see your nits and raise you one
https://github.com/ava-labs/firewood/issues/394

I think storage could be done in a much more idiomatic way.

PR #378, opened 2023-11-23
Move proof to be a sub-module of merkle

richardpringle:
Proofs only make sense in the context of a merkle-tree, therefore they should be a submodule of the merkle module.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -8,14 +8,13 @@ pub use crate::{
 };
 use crate::{
     file,
-    merkle::{Merkle, MerkleError, Node, TrieHash, TRIE_HASH_LEN},
-    proof::ProofError,
+    merkle::{Merkle, MerkleError, Node, Proof, ProofError, TrieHash, TRIE_HASH_LEN},
     storage::{
         buffer::{DiskBuffer, DiskBufferRequester},
         CachedSpace, MemStoreR, SpaceWrite, StoreConfig, StoreDelta, StoreRevMut, StoreRevShared,
         ZeroStore, PAGE_SIZE_NBIT,
     },
-    v2::api::{self, HashKey, KeyType, Proof, ValueType},
+    v2::api::{self, HashKey, KeyType, ValueType},
 };
 use crate::{
     merkle,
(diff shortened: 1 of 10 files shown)


rkuris (changes requested):
The API returns proofs, so we want api::Proof to be visible to callers. Otherwise, the whole merkle module has to be pub.

richardpringle:
> The API returns proofs, so we want api::Proof to be visible to callers. Otherwise, the whole merkle module has to be pub.

I thought the `api` module was just a mechanism for serving two apis at the same time. We shouldn't have a separate module for the public API, there's a reason that no one does this. 

>Otherwise, the whole merkle module has to be pub.

This specifically is incorrect. The whole purpose of `pub use merkle::Proof` (called in the `db.rs`) solves this problem. The `Proof` type and the `impl Proof` should live in the same place.

xinifinity (approved):

rkuris:
> > The API returns proofs, so we want api::Proof to be visible to callers. Otherwise, the whole merkle module has to be pub.
> 
> I thought the `api` module was just a mechanism for serving two apis at the same time. We shouldn't have a separate module for the public API, there's a reason that no one does this.
> 
> > Otherwise, the whole merkle module has to be pub.
> 
> This specifically is incorrect. The whole purpose of `pub use merkle::Proof` (called in the `db.rs`) solves this problem. The `Proof` type and the `impl Proof` should live in the same place.

This only works if `merkle::Proof` is also pub, which is what I want to avoid. This does not compile:

```
pub use inner::Thing;

mod inner {
    struct Thing {}
}
```

richardpringle:
> > > The API returns proofs, so we want api::Proof to be visible to callers. Otherwise, the whole merkle module has to be pub.
> > 
> > 
> > I thought the `api` module was just a mechanism for serving two apis at the same time. We shouldn't have a separate module for the public API, there's a reason that no one does this.
> > > Otherwise, the whole merkle module has to be pub.
> > 
> > 
> > This specifically is incorrect. The whole purpose of `pub use merkle::Proof` (called in the `db.rs`) solves this problem. The `Proof` type and the `impl Proof` should live in the same place.
> 
> This only works if `merkle::Proof` is also pub, which is what I want to avoid. This does not compile:
> 
> ```
> pub use inner::Thing;
> 
> mod inner {
>     struct Thing {}
> }
> ```

but this does
https://github.com/ava-labs/firewood/pull/378/files#diff-648f516579918c7c3dddf3fd5eec6807c064b1ae8f8411d2c029bd804f7ef871R4

rkuris (approved):

PR #379, opened 2023-11-27
Add shared-path-on-insert test

richardpringle:
This is an important test case when it comes to making sure that path behaviour remains the same when moving paths to branches.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1957,4 +1957,48 @@ mod tests {
 
         assert_eq!(verified, Some(value2.to_vec()));
     }
+
+    // this was a specific failing case
+    #[test]
+    fn shared_path_on_insert() {
+        type Bytes = &'static [u8];
+        let pairs: Vec<(Bytes, Bytes)> = vec![
+            (
+                &[1, 1, 46, 82, 67, 218],
+                &[23, 252, 128, 144, 235, 202, 124, 243],
+            ),
+            (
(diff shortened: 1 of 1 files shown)


firewood/src/merkle.rs
```diff
@@ -1957,4 +1957,36 @@ mod tests {
...
+            ),
+        ];
+
+        let mut merkle = create_test_merkle();
+        let root = merkle.init_root().unwrap();
+
+        for (key, val) in pairs {
+            let val = val.to_vec();
+            merkle.insert(key, val, root).unwrap()
+        }
```
rkuris, line 1990: What are we actually testing here? That the insert succeeds? Shouldn't we walk the merkle afterwards and make sure it was built correctly?

rkuris (changes requested):

firewood/src/merkle.rs
richardpringle: Good catch, I must have deleted the check in the process of pulling this out of its original branch.

xinifinity: Did you push the check? Sorry I don't see it yet?

richardpringle: I pushed it to the wrong place... whoops

richardpringle: pushed

xinifinity (approved):
LL

rkuris (approved):

PR #380, opened 2023-11-27
Centralize trie-traversal for proof generation

richardpringle:
Trie traversal has historically been re-implmented in every place that requires such logic. This makes it very difficult to implement any changes in the trie structure itself. Traversal has been partially centralized prior to this PR, but this _should_ finish the job.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -970,7 +970,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         node_ref: ObjRef<'a>,
         key: K,
     ) -> Result<Option<ObjRef<'a>>, MerkleError> {
-        self.get_node_by_key_with_callback(node_ref, key, |_, _| {})
+        self.get_node_by_key_with_callbacks(node_ref, key, |_, _| {}, |_, _| {})
     }
 
     fn get_node_and_parents_by_key<'a, K: AsRef<[u8]>>(
@@ -979,9 +979,14 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         key: K,
     ) -> Result<(Option<ObjRef<'a>>, ParentRefs<'a>), MerkleError> {
         let mut parents = Vec::new();
-        let node_ref = self.get_node_by_key_with_callback(node_ref, key, |node_ref, nib| {
-            parents.push((node_ref, nib));
-        })?;
+        let node_ref = self.get_node_by_key_with_callbacks(
+            node_ref,
+            key,
+            |_, _| {},
+            |node_ref, nib| {
+                parents.push((node_ref, nib));
+            },
+        )?;
 
         Ok((node_ref, parents))
     }
@@ -992,18 +997,24 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
         key: K,
     ) -> Result<(Option<ObjRef<'a>>, ParentAddresses), MerkleError> {
         let mut parents = Vec::new();
-        let node_ref = self.get_node_by_key_with_callback(node_ref, key, |node_ref, nib| {
-            parents.push((node_ref.into_ptr(), nib));
-        })?;
+        let node_ref = self.get_node_by_key_with_callbacks(
+            node_ref,
+            key,
+            |_, _| {},
+            |node_ref, nib| {
+                parents.push((node_ref.into_ptr(), nib));
+            },
+        )?;
 
         Ok((node_ref, parents))
     }
 
-    fn get_node_by_key_with_callback<'a, K: AsRef<[u8]>>(
+    fn get_node_by_key_with_callbacks<'a, K: AsRef<[u8]>>(
         &'a self,
         mut node_ref: ObjRef<'a>,
         key: K,
-        mut loop_callback: impl FnMut(ObjRef<'a>, u8),
+        mut start_loop_callback: impl FnMut(DiskAddress, u8),
+        mut end_loop_callback: impl FnMut(ObjRef<'a>, u8),
     ) -> Result<Option<ObjRef<'a>>, MerkleError> {
         let mut key_nibbles = Nibbles::<1>::new(key.as_ref()).into_iter();
 
@@ -1011,6 +1022,8 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
             let Some(nib) = key_nibbles.next() else {
                 break;
             };
+
+            start_loop_callback(node_ref.as_ptr(), nib);
 
             let next_ptr = match &node_ref.inner {
                 NodeType::Branch(n) => match n.children[nib as usize] {
@@ -1045,7 +1058,7 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
                 }
             };
 
-            loop_callback(node_ref, nib);
+            end_loop_callback(node_ref, nib);
 
             node_ref = self.get_node(next_ptr)?;
         }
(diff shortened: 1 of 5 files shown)


firewood/src/merkle.rs
```diff
@@ -1090,76 +1103,36 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
-        // Skip the sentinel root
-        let root = self
-            .get_node(root)?
-            .inner
-            .as_branch()
-            .ok_or(MerkleError::NotBranchNode)?
-            .children[0];
-        let mut u_ref = match root {
-            Some(root) => self.get_node(root)?,
-            None => return Ok(Proof(proofs)),
```
rkuris, line 1109: this needs a test case. It's not clear if it's handled correctly with the new code. The case here is that there is a sentinel node but no first child.

firewood/src/merkle.rs
```diff
@@ -1989,4 +1962,216 @@ mod tests {
             merkle.insert(key, val, root).unwrap()
         }
     }
+
+    #[test]
+    fn overwrite_leaf() {
```
rkuris, line 1967: It's unclear what this test has to do with the above changes.

firewood/src/merkle.rs
```diff
@@ -1989,4 +1962,216 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            merkle.get(&key_2, root).unwrap().as_deref(),
+            Some(val_2.as_slice())
+        );
+    }
+
+    #[test]
+    fn single_key_proof_with_one_node() {
```
rkuris, line 2141: How is this different from `test_one_element_proof` in the tests directory?

firewood/src/merkle.rs
```diff
@@ -1989,4 +1962,216 @@ mod tests {
...
+
+        let proof = merkle.prove(key, root).unwrap();
+
+        let verified = proof.verify(key, root_hash.0).unwrap();
+
+        assert_eq!(verified, Some(value.to_vec()));
+    }
+
+    #[test]
+    fn two_key_proof_without_shared_path() {
```
rkuris, line 2158: Does this cover something not covered in `test_proof`?

firewood/src/merkle.rs
```diff
@@ -992,18 +997,24 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
         Ok((node_ref, parents))
     }
 
-    fn get_node_by_key_with_callback<'a, K: AsRef<[u8]>>(
+    fn get_node_by_key_with_callbacks<'a, K: AsRef<[u8]>>(
         &'a self,
         mut node_ref: ObjRef<'a>,
         key: K,
-        mut loop_callback: impl FnMut(ObjRef<'a>, u8),
+        mut start_loop_callback: impl FnMut(DiskAddress, u8),
```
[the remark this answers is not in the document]
richardpringle: I could pass an `&ObjRef`, but I can't pass the owned type until after I'm done using it. So I could call the `as_ptr` in the callback itself if you prefer, but I can't make both callbacks take an owned type.

firewood/src/merkle.rs
richardpringle: It's not. Would you like me to delete the other test?

firewood/src/merkle.rs
richardpringle: I don't know and it's impossible to tell what the structure of the tree is in that test without debugging. There is no guarantee that there are nodes that don't share a path.

firewood/src/merkle.rs
richardpringle: It's an extra test case that wasn't previously covered.

firewood/src/merkle/proof.rs
```diff
@@ -118,33 +118,29 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+                }) if key_nibbles.is_empty() => return Ok(Some(encoded)),
                 // The trie doesn't contain the key.
                 Some(SubProof {
                     hash: Some(hash), ..
                 }) => hash,
                 _ => return Ok(None),
             };
-        };
-
-        let value = match NodeType::decode(&value_node) {
```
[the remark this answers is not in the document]
xinifinity: hmm, I see. I think Subproof should only contains un-decoded value, so not sure if change it to `enum` is a good idea.

firewood/src/merkle.rs
rkuris: Sure. Those tests belong here anyway.

firewood/src/merkle.rs
rkuris: It would be easy to add a check for that in the other test and fold this one into that. It does generate enough data that it seems extremely likely there will be non-shared path proofs.

I like the idea of testing more than just this one case.

firewood/src/merkle.rs
```diff
@@ -992,18 +997,24 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
+        mut end_loop_callback: impl FnMut(ObjRef<'a>, u8),
```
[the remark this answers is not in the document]
richardpringle: https://github.com/ava-labs/firewood/issues/399

firewood/src/merkle.rs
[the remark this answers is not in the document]
richardpringle: #399 solves this too

firewood/src/merkle.rs
richardpringle: https://github.com/ava-labs/firewood/pull/380/files#diff-6a6e300f3d9050248dbafa719531f14eabecc1bfac2169bb2527c81ee5474ecaR1841

firewood/src/merkle.rs
```diff
@@ -1090,76 +1103,36 @@ impl<S: ShaleStore<Node> + Send + Sync> Merkle<S> {
...
-                        break;
-                    }
-                    if !remaining_path.take(n_path.len()).eq(n_path.iter().cloned()) {
-                        // contents aren't the same
-                        break;
+        let node = self.get_node_by_key_with_callbacks(
+            root_node,
+            key,
+            {
+                let (mut is_root, nodes) = (true, &mut nodes);
```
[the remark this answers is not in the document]
richardpringle: Without this (I removed it), we're potentially unnecessarily allocating an extra value... which made me think that we actually know the max capacity of the `Vec` ahead of time. But all this is moot with #399

firewood/src/merkle.rs
richardpringle: But that goes against proper unit testing principles. Unit tests should just test one thing. I'm not dogmatic about this; I've written tests that do more, but I had an absolute nightmare trying to figure out why `test_proof` was failing before. With this case, it would be ultra clear why it's failing; there's no extra complexity. 

We also had another test that was only passing because the leaf nodes that were created didn't have partial paths BY CHANCE. We should know the exact state of the thing we are testing. We should *also* fuzz test (using proper fuzzer, not just random values), but fuzzing and unit testing should be completely separate IMO. They serve a different purpose.

firewood/src/merkle/proof.rs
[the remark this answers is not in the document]
richardpringle: I just did this now. The issue with the code before is that the states weren't well represented. We never used `encoded` if hash was `Some`. 
cdf026aab295322ac3a2abe17c778ca3a00e3270

firewood/src/merkle/node/branch.rs
```diff
@@ -164,6 +164,11 @@ impl BranchNode {
                         list[i] = Encoded::Raw(child_encoded.to_vec());
                     }
                 }
+
+                // I'm not sure that this case makes sense. Why would there be an encoded child if there isn't a
```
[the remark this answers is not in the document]
richardpringle: 5a9e0d9d7810749fbc9984ead072905a9606126f

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: Yeah lets defer to #399

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: Go #399 🥂

firewood/src/merkle.rs
rkuris: I think that's a test for `range_proof` that does the same thing? I don't believe it gets to `proof` since it can't find the first key. Is that incorrect?

firewood/src/merkle.rs
rkuris: Fair point. We should rename that test fuzz or something.

rkuris: I do wonder if there are other cases besides this one we aren't covering though, but I can't offhand think of any.

firewood/src/merkle.rs
richardpringle: bad link... edited

rkuris (approved):

PR #381, opened 2023-11-27
Growth ring file handling improvements

rkuris:
Unwrap removals and better type safety during file traversal

### DIFF
--- a/growth-ring/src/wal.rs
+++ b/growth-ring/src/wal.rs
@@ -9,19 +9,21 @@ use futures::{
     Future,
 };
 
-use std::cell::{RefCell, UnsafeCell};
 use std::convert::{TryFrom, TryInto};
 use std::mem::MaybeUninit;
 use std::num::NonZeroUsize;
 use std::pin::Pin;
+use std::{
+    cell::{RefCell, UnsafeCell},
+    ffi::OsStr,
+    path::{Path, PathBuf},
+};
 use std::{
     collections::{hash_map, BinaryHeap, HashMap, VecDeque},
     marker::PhantomData,
 };
 
 pub use crate::walerror::WalError;
-
-const FILENAME_FMT: &str = r"[0-9a-f]+\.log";
 
 enum WalRingType {
     #[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


growth-ring/src/wal.rs
```diff
@@ -59,8 +61,23 @@ type WalFileId = u64;
 pub type WalBytes = Box<[u8]>;
 pub type WalPos = u64;
 
-fn get_fid(fname: &str) -> WalFileId {
-    scan_fmt!(fname, "{x}.log", [hex WalFileId]).unwrap()
+// convert XXXXXX.log into number from the XXXXXX (in hex)
+fn get_fid(fname: &Path) -> Result<WalFileId, WalError> {
+    let wal_err: WalError = WalError::Other("not a log file".to_string());
```
richardpringle, line 66: In theory, this is an allocation with every call to this function. I think it would be better to repeat the error creation.

growth-ring/src/wal.rs
```diff
@@ -59,8 +61,23 @@ type WalFileId = u64;
...
+
+    if fname.extension() != Some(OsStr::new("log")) {
+        return Err(wal_err);
+    }
+
+    u64::from_str_radix(
+        fname
+            .file_stem()
+            .unwrap_or(OsStr::new(""))
```
richardpringle, line 75: Can you use `unwrap_or_default` instead?

growth-ring/src/wal.rs
```diff
@@ -59,8 +61,23 @@ type WalFileId = u64;
...
+            .to_str()
+            .unwrap_or(""),
```
richardpringle, line 77: Same here

growth-ring/src/wal.rs
```diff
@@ -1301,3 +1314,25 @@ impl WalLoader {
...
+mod test {
+    use super::*;
+    use test_case::test_case;
+
+    #[test_case("foo", Err("not a log file"); "no log extension")]
+    #[test_case("foo.log", Err("not a log file"); "invalid digit found in string")]
+    #[test_case("0000001.log", Ok(1); "happy path")]
+    #[test_case("1.log", Ok(1); "no leading zeroes")]
+
+    fn test_get_fid(input: &str, expected: Result<u64, &str>) {
```
richardpringle, line 1328: Nit: prefixing test functions with `test_` is a stutter, since they are already in the `tests` module.

richardpringle (approved):
All comments are nits

PR #390, opened 2023-11-28
Merkle test move only, no code changes

rkuris:
This moves the merkle integration tests into unit tests, to allow removing the merkle code from the external interface.


richardpringle (changes requested):
Can we hold off on merging these in here for now? I would really rather see these tests cleaned up before they are moved. The style of testing in this file is very different than everything that's currently in `src/merkle.rs`.

rkuris:
Move then clean. All the merkle tests could use some attention and this
gets them all in one place.

rkuris:
Richard has 10 open PRs and doesn't want this merged due to possible conflicts.

PR #392, opened 2023-11-29
Small cleanups

richardpringle:
Prefer using fully qualified paths


rkuris (changes requested):
This makes HashKey not visible unless the caller uses merkle::HashKey. The merkle directory won't be public.

PR #397, opened 2023-11-30
growth-ring: Add remaining lint checks

rkuris:
Removed a couple of index dereferences and unwraps.

Unfortunately, there's one spot where it isn't easy to add the lint check, so there is a block that disables it. It's still a lot better having it in most of the project and not in this block. (search for TODO: not allowed for the spot).

Rewriting the loop to be more idiomatic is probably the right move here, but that's outside the scope of this fix.

### DIFF
--- a/growth-ring/Cargo.toml
+++ b/growth-ring/Cargo.toml
@@ -35,4 +35,6 @@ crate-type = ["dylib", "rlib", "staticlib"]
 
 [lints.clippy]
 unwrap_used = "warn"
+indexing_slicing = "warn"
+explicit_deref_methods = "warn"
 missing_const_for_fn = "warn"
--- a/growth-ring/src/wal.rs
+++ b/growth-ring/src/wal.rs
@@ -289,13 +289,17 @@ impl<F: WalFile + 'static, S: WalStore<F>> WalFilePool<F, S> {
         })
     }
 
-    #[allow(clippy::unwrap_used)]
     async fn read_header(&self) -> Result<Header, WalError> {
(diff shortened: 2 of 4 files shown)


growth-ring/Cargo.toml
```diff
@@ -35,4 +35,6 @@ crate-type = ["dylib", "rlib", "staticlib"]
 
 [lints.clippy]
 unwrap_used = "warn"
+indexing_slicing = "warn"
```
xinifinity, line 38: Maybe do not enable warning on `indexing_slicing` until we have fix most of them, as there are a bunch of places we need to explicitly disable them.

xinifinity: Sounds good, not feeling strong about this.

xinifinity (approved):

richardpringle (approved):
Index slicing is a little less trivial than disallowing `unwrap`. 

There are times when you call a method to get a slice of a specific length and as long as your index is smaller than said length, you're fine indexing into the slice. It's these cases where the compiler wouldn't even do a bounds check. 

Do we want to keep this type of behaviour and add the lint-ignore attribute?

It's possible that this never comes up. I can already think of places I do this but could do it without indexing. But in the case where it does, it would suck to make a function return a `Result` if it didn't need to.

I'm approving the PR because I agree with the change in general, but I would like to have a strategy in case the aforementioned scenario does come up.

rkuris:
> Index slicing is a little less trivial than disallowing `unwrap`.
Please have a look at the other 2 proposed lint changes going forward that were approved in https://github.com/ava-labs/firewood/pull/356 to ensure we're all in agreement on direction before I continue with these changes.

PR #401, opened 2023-12-01
Rkuris/deny new unsafe growth ring

rkuris:
Final change for growth ring to avoid any new unsafes without a lint override. There are still a few unsafes that need fixing in another PR.

Please review #400 first!


growth-ring/Cargo.toml
```diff
@@ -33,6 +33,9 @@ name = "growthring"
 path = "src/lib.rs"
 crate-type = ["dylib", "rlib", "staticlib"]
 
+[lints.rust]
+unsafe_code = "deny"
```
richardpringle, line 37: 🚀

PR #407, opened 2023-12-04
Rename rpc -> grpc-testtool

rkuris:


xinifinity (commented):
As discussed, we may use the same crate not only for testing but for VM integration over grpc as well?

rkuris:
> As discussed, we may use the same crate not only for testing but for VM integration over grpc as well?

We can always rename this again if that happens.

richardpringle (approved):
I'm approving this... but I would prefer `grpc-test-tool`

PR #414, opened 2023-12-05
Add additional proto files to testtool

rkuris:
There is a process server that we need to implement as well as the two other RPCs.

This just builds the proto specs. A bit tricky since it includes prometheus metrics.


rkuris:
We are switching to a String for the metrics

PR #423, opened 2023-12-07
Orchestrator interface changes

rkuris:
Changed to use --options so we don't have to parse all database specific options (we might add more to merkledb). This uses a json blob for options. We can parse additional ones as they become supported from the orchestrator.

Also added span-level tracing for debugging in a second commit. This required sprinkling `#[derive(Debug)]` on the components.

### DIFF
--- a/grpc-testtool/src/bin/process-server.rs
+++ b/grpc-testtool/src/bin/process-server.rs
@@ -84,8 +84,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
         .filter(None, args.log_level)
         .init();
 
+    tracing_subscriber::fmt::init();
+
     // log to the file and to stderr
-    eprintln!("Database-Server listening on: {}", args.grpc_port);
     info!("Starting up: Listening on {}", args.grpc_port);
 
     let svc = Arc::new(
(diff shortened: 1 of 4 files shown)


grpc-testtool/src/bin/process-server.rs
```diff
@@ -91,19 +81,22 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
         .format_target(true)
-        .target(Target::Pipe(Box::new(logfile)))
         .filter(None, args.log_level)
         .init();
 
-    log_panics::Config::new()
-        .backtrace_mode(log_panics::BacktraceMode::Unresolved)
-        .install_panic_hook();
-
     // log to the file and to stderr
```
xinifinity, line 89: nit: we no longer log to file

grpc-testtool/src/bin/process-server.rs
```diff
@@ -3,82 +3,72 @@
...
     let args = Opts::parse();
 
-    match std::fs::create_dir_all(&args.log_dir) {
-        Ok(()) => {}
-        Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {}
-        Err(e) => panic!("Unable to create directory\n{}", e),
-    }
-
-    // set up the log file
-    let logfile = std::fs::OpenOptions::new()
```
xinifinity, line 76: Why do we remove the part to log to file?

xinifinity (approved):

grpc-testtool/src/bin/process-server.rs
rkuris: Because the orchestrator now captures stdout and stderr

Issue #399, opened 2023-11-30
Look into the best way to run a callback on the parent node when traversing the trie


richardpringle:
from #380 

Even better than that comment, I can't imagine the case where we are using both callbacks, so this is probably an enum:
```rust
enum Callback {
    None,
    Infix(FnMut(ObjRef<'a>, u8),
    PostFix(FnMut(ObjRef<'a>, u8)
};
fn get_node_by_key_with_callback<'a, K: AsRef<[u8]>>(
        &'a self,
        mut node_ref: ObjRef<'a>,
        key: K,
        callback: Callback,
```

_Originally posted by @rkuris in https://github.com/ava-labs/firewood/pull/380#discussion_r1409949526_

The above isn't actually possible since we can't take ownership of the node when traversing the trie. After looking into this a little further, it seems like the only difference between the two cases is that the traversed-node always goes through the callback when called at the start, and won't get passed to the callback when early returning. 

If we renamed this method to `traverse`, and return `bool` that's true when all nibbles matched on nodes and false otherwise, we could get all the behaviour that we want at the call-site. That might require popping off the back of a `Vec` of traversed nodes, but that logic will be hidden by wrapper functions.

richardpringle:
On second thought, it would be way better to just implement an iterator (stream?)
```rust
// need two lifetimes otherwise the key has to live as long as the trie itself
fn get_keyed_iter<'a, 'b, K: AsRef<[u8]> + 'b>(&'a self, root: DiskAddress, key: K) -> KeyedIter<'a, 'b> { .. }
```
Then getting all the parents and the final node would look like this:
```rust
let root_node = self.get_node(root_addr)?;

let (parents, found_node) = merkle
    .get_keyed_iter(root_addr, &key)
    // `current` would be Option<ObjRef<Node>> 
    // depending on whether or not there is a node at `nib`
    .fold((Vec::new(), Some(root_node)), |(mut parents, parent), (current, nib)| {
        if let Some(node) = parent {
            parents.push((node, nib))
        }
    
        (parents, current)
    });
```
again, I want to emphasize that this will be surrounded by a wrapper function. 

The above has the same behaviour as pushing all the parents and returning `Option<NodeRef>` for the found-node. 

For proofs, we skip the root node anyway

```rust
let nodes = merkle.get_keyed_iter(root_addr, &key).filter_map(|(node, _)| node).collect();
```
We could actually go further with the latter case such that and create the hash-map directly.

```rust
let proofs = merkle
[...]

PR #424, opened 2023-12-07
Make `MerkleKeyValueStream` handle branches with values

richardpringle:
This was not the easiest thing to implement. Once I got the tests passing, I went over it a few times. 

It actually looks like we should move the starting logic into the `get_iter` method, but I will leave that for another day. I honestly tried to make this a much smaller change, but there was no way to get it to work without re-writing everything. 

There are two commits here. The first one is the move (feel free to verify that all the tests pass with the pure move). The second is the addition of tests and changes in behaviour.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1400,7 +1400,6 @@ pub fn from_nibbles(nibbles: &[u8]) -> impl Iterator<Item = u8> + '_ {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::StreamExt;
     use node::tests::{extension, leaf};
     use shale::{cached::DynamicMem, compact::CompactSpace, CachedStore};
     use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


firewood/src/merkle.rs
```diff
@@ -1690,7 +1412,7 @@ mod tests {
         assert_eq!(n, nibbles);
     }
 
-    fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>, Bincode> {
+    pub(super) fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>, Bincode> {
```
richardpringle, line 1415: This isn't ideal but it was the quickest way to re-use this code.

firewood/src/merkle/stream.rs
```diff
@@ -0,0 +1,505 @@
...
+                        .collect::<Result<Vec<_>, _>>()
+                        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+
+                    // means the keys definitely match
+                    if let Some(found_node) = found_node {
+                        let (value, visit_last) = match found_node.inner() {
+                            NodeType::Extension(_) => unreachable!(),
+
+                            NodeType::Branch(branch) => {
+                                let value = branch.value.as_ref().map(|v| v.to_vec());
```
richardpringle, line 121: 🤔, this is a little awkward. `get_node_by_key_with_callbacks` will only return nodes with data. 

This will go away once #399 is implemented.

firewood/src/merkle/stream.rs
```diff
@@ -0,0 +1,505 @@
...
+                            parents,
+                        };
+
+                        return self.poll_next(_cx);
+                    }
+
+                    let key_from_parents = key_from_nibble_iter(nibble_iter_from_parents(&parents));
+                    let mut visit_last = key == &key_from_parents;
+
+                    if (key.as_slice()) < (key_from_parents.as_slice()) {
```
richardpringle, line 150: Clippy should have removed these brackets for me. Leaving a comment here for myself

firewood/src/merkle/stream.rs
```diff
@@ -0,0 +1,528 @@
...
+pub(super) enum IteratorState<'a> {
+    /// Start iterating at the beginning of the trie,
+    /// returning the lowest key/value pair first
+    StartAtBeginning,
+    /// Start iterating at the specified key
+    StartAtKey(Vec<u8>),
+    /// Continue iterating after the given `next_node` and parents
+    Iterating {
+        visit_last: bool,
+        next_result: Option<(Vec<u8>, Vec<u8>)>,
```
rkuris, line 21: The state should not include the next key, only location information as to where we are.

As we discussed in-person, we should attempt to fix the existing stream iterator unless there's some reason why we can't.

Next steps are to add a unit test and debug the existing stream iterator code.

rkuris (changes requested):

PR #425, opened 2023-12-08
Add Merkledb encoding schema

### LINKED ISSUE #294: Improve compatibility with MerkleDB, especially in the area of hash calculations
Now we have the pluggable encoding, we can implement the MerkleDB encoding schema.

xinifinity:
With pluggable encoding, implement the MerkleDB matching node encoding schema.

Closes #294.

### DIFF
--- a/firewood/benches/hashops.rs
+++ b/firewood/benches/hashops.rs
@@ -6,7 +6,7 @@
 use criterion::{criterion_group, criterion_main, profiler::Profiler, BatchSize, Criterion};
 use firewood::{
     db::{BatchOp, DbConfig},
-    merkle::{Merkle, MerkleWithEncoder, TrieHash, TRIE_HASH_LEN},
+    merkle::{MerkleWithEncoder, TrieHash, TRIE_HASH_LEN},
     shale::{
         cached::PlainMem,
         compact::{CompactHeader, CompactSpace},
@@ -104,7 +104,7 @@ fn bench_merkle<const N: usize>(criterion: &mut Criterion) {
                     )
                     .unwrap();
 
-                    let merkle: MerkleWithEncoder = Merkle::new(Box::new(store));
+                    let merkle = MerkleWithEncoder::new(Box::new(store));
                     let root = merkle.init_root().unwrap();
 
                     let keys: Vec<Vec<u8>> = repeat_with(|| {
--- a/firewood/benches/hashops.rs
+++ b/firewood/benches/hashops.rs
@@ -6,7 +6,7 @@
 use criterion::{criterion_group, criterion_main, profiler::Profiler, BatchSize, Criterion};
 use firewood::{
     db::{BatchOp, DbConfig},
-    merkle::{Merkle, TrieHash, TRIE_HASH_LEN},
+    merkle::{Bincode, Merkle, Node, TrieHash, TRIE_HASH_LEN},
     shale::{
         cached::PlainMem,
         compact::{CompactHeader, CompactSpace},
@@ -104,7 +104,8 @@ fn bench_merkle<const N: usize>(criterion: &mut Criterion) {
                     )
                     .unwrap();
 
-                    let merkle = Merkle::new(Box::new(store));
+                    let merkle: Merkle<CompactSpace<Node, PlainMem>, Bincode> =
+                        Merkle::new(Box::new(store));
                     let root = merkle.init_root().unwrap();
 
                     let keys: Vec<Vec<u8>> = repeat_with(|| {
--- a/firewood/benches/hashops.rs
+++ b/firewood/benches/hashops.rs
@@ -6,7 +6,7 @@
 use criterion::{criterion_group, criterion_main, profiler::Profiler, BatchSize, Criterion};
 use firewood::{
     db::{BatchOp, DbConfig},
-    merkle::{Bincode, Merkle, Node, TrieHash, TRIE_HASH_LEN},
+    merkle::{Merkle, MerkleWithEncoder, TrieHash, TRIE_HASH_LEN},
     shale::{
         cached::PlainMem,
         compact::{CompactHeader, CompactSpace},
@@ -104,8 +104,7 @@ fn bench_merkle<const N: usize>(criterion: &mut Criterion) {
                     )
                     .unwrap();
 
-                    let merkle: Merkle<CompactSpace<Node, PlainMem>, Bincode> =
-                        Merkle::new(Box::new(store));
+                    let merkle: MerkleWithEncoder = Merkle::new(Box::new(store));
                     let root = merkle.init_root().unwrap();
 
                     let keys: Vec<Vec<u8>> = repeat_with(|| {
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1501,17 +1501,15 @@ mod tests {
     #[test_case(leaf(Vec::new(), Vec::new()) ; "empty leaf encoding")]
     #[test_case(leaf(vec![1, 2, 3], vec![4, 5]) ; "leaf encoding")]
     #[test_case(branch(b"value".to_vec(), vec![1, 2, 3].into()) ; "branch with chd")]
-    #[test_case(branch(b"value".to_vec(), None); "branch without chd")]
-    #[test_case(branch(Vec::new(), None); "branch without value and chd")]
+    #[test_case(branch(Vec::new(), vec![1, 2, 3].into()); "branch without value")]
     fn node_encode_decode_plain_(node: Node) {
         let merkle = create_test_merkle::<PlainCodec>();
         let node_ref = merkle.put_node(node.clone()).unwrap();
         let encoded = merkle.encode(&node_ref).unwrap();
-        println!("encoded: {:?}", encoded);
-        // let new_node = Node::from(merkle.decode(encoded.as_ref()).unwrap());
-        // let new_node_hash = new_node.get_root_hash(merkle.store.as_ref());
-        // let expected_hash = node.get_root_hash(merkle.store.as_ref());
-        // assert_eq!(new_node_hash, expected_hash);
+        let new_node = Node::from(merkle.decode(encoded.as_ref()).unwrap());
+        let new_node_hash = new_node.get_root_hash(merkle.store.as_ref());
+        let expected_hash = node.get_root_hash(merkle.store.as_ref());
+        assert_eq!(new_node_hash, expected_hash);
     }
 
     #[test]
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1451,10 +1451,7 @@ mod tests {
         Merkle::new(store)
     }
 
-    pub(super) fn create_test_merkle<'de>() -> Merkle<CompactSpace<Node, DynamicMem>, Bincode>
-    where
-        EncodedNode<Bincode>: serde::Serialize + serde::Deserialize<'de>,
-    {
+    pub(super) fn create_test_merkle() -> Merkle<CompactSpace<Node, DynamicMem>, Bincode> {
         create_generic_test_merkle::<Bincode>()
     }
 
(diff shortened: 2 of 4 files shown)


richardpringle:
@xinifinity, could you please add a proper PR description?

firewood/src/merkle.rs
```diff
@@ -1761,7 +1767,21 @@ mod tests {
...
     #[test_case(branch(Vec::new(), None); "branch without value and chd")]
     fn node_encode_decode_(node: Node) {
-        let merkle = create_test_merkle();
+        let merkle = create_test_merkle::<Bincode>();
+        let node_ref = merkle.put_node(node.clone()).unwrap();
+        let encoded = merkle.encode(&node_ref).unwrap();
+        let new_node = Node::from(merkle.decode(encoded.as_ref()).unwrap());
+        let new_node_hash = new_node.get_root_hash(merkle.store.as_ref());
+        let expected_hash = node.get_root_hash(merkle.store.as_ref());
+        assert_eq!(new_node_hash, expected_hash);
```
richardpringle, line 1776: This is really hard to read in a big block like this. Can we add _some_ spacing? Doesn't have to be as much as I would add, any is better than none.

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,93 @@ pub enum EncodedNodeType {
     },
 }
 
+// TODO: probably can merge with `EncodedNodeType`.
```
richardpringle, line 498: 👍

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,93 @@ pub enum EncodedNodeType {
...
+
+// Note that the serializer passed in should always be the same type as T in EncodedNode<T>.
+impl Serialize for EncodedNode<PlainCodec> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let n = match &self.node {
+            EncodedNodeType::Leaf(n) => {
+                let data = Some(n.data.to_vec()).filter(|d| !d.is_empty());
```
richardpringle, line 514: You should be able to store empty data

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,93 @@ pub enum EncodedNodeType {
...
+                let chd: Vec<(u64, Vec<u8>)> = Default::default();
+                let path = from_nibbles(&n.path.encode(true)).collect();
```
richardpringle, line 516: 🤔, we don't need the prefix to match with `MerkleDB`, right?

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,93 @@ pub enum EncodedNodeType {
...
+                    data,
+                    path: Vec::new(),
+                }
+            }
+        };
+
+        let mut s = serializer.serialize_struct("", 3)?;
+        s.serialize_field("", &n.chd)?;
+        s.serialize_field("", &n.data)?;
+        s.serialize_field("", &n.path)?;
```
richardpringle, line 547: Can't you just serialize a tuple here?

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,93 @@ pub enum EncodedNodeType {
...
+                    .enumerate()
+                    .filter_map(|(i, c)| c.as_ref().map(|c| (i, c)))
+                {
+                    if c.len() >= TRIE_HASH_LEN {
+                        let hash = Keccak256::digest(c).to_vec();
+                        chd.push((i as u64, hash));
+                    } else {
+                        chd.push((i as u64, c.to_vec()));
+                    }
+                }
```
richardpringle, line 533: Can you just do this with an iterator, calling `collect` at the end to turn it into a `Vec`?

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,93 @@ pub enum EncodedNodeType {
...
+
+                let data = value.clone().map(|v| v.0);
```
richardpringle, line 535: You should just me cloning the value. I think you want to call `.as_ref().map(|v| v.to_vec())`

richardpringle (changes requested):

firewood/src/merkle.rs
```diff
@@ -1723,7 +1729,7 @@ mod tests {
 
     fn branch(value: Vec<u8>, encoded_child: Option<Vec<u8>>) -> Node {
         let children = Default::default();
-        let value = Some(Data(value));
+        let value = Some(Data(value)).filter(|data| !data.is_empty());
```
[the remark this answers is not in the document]
xinifinity: This is because the test has branch node with [empty value](https://github.com/ava-labs/firewood/pull/425/files#diff-6a6e300f3d9050248dbafa719531f14eabecc1bfac2169bb2527c81ee5474ecaL1745). Want to cover it correctly.

firewood/src/merkle/node.rs
xinifinity: Yeah I guess empty value is different from zero value in this case. Tried to match with MerkleDB but I think your suggestion is actually right with a second thought.

firewood/src/merkle/node.rs
xinifinity: I think we need, but we can revisit this after we remove the extension node with correct path stored.

firewood/src/merkle/node.rs
xinifinity: Initially did as a tuple, but I found it is easier to do as a struct without implementation the visitor and directly use `derive`.

xinifinity: Plus we probably can just use one struct `EncodedBranchNode` if we only have branch node in the end.

richardpringle (changes requested):
Two (non-nit) change requests and a couple of questions too

firewood/src/merkle.rs
```diff
@@ -1760,13 +1775,31 @@ mod tests {
...
+        let merkle = create_test_merkle();
+        let node_ref = merkle.put_node(node.clone()).unwrap();
         let expected_hash = node.get_root_hash(merkle.store.as_ref());
+
+        let encoded = merkle.encode(&node_ref).unwrap();
+        let new_node = Node::from(merkle.decode(encoded.as_ref()).unwrap());
+        let new_node_hash = new_node.get_root_hash(merkle.store.as_ref());
+        
         assert_eq!(new_node_hash, expected_hash);
     }
```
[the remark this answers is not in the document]
xinifinity: Thought about the same thing, was thinking need to make the test case generic to be able to do so, but I guess an option is also fine.

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,92 @@ pub enum EncodedNodeType {
...
+                    path: Vec::new(),
+                }
+            }
+        };
+
+        let mut s = serializer.serialize_struct("", 3)?;
+        s.serialize_field("", &n.chd)?;
+        s.serialize_field("", &n.data)?;
+        s.serialize_field("", &n.path)?;
+        s.end()
```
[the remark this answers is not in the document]
xinifinity: I thought I have to use `serialize_struct` in this case LOL

firewood/src/merkle/node.rs
```diff
@@ -666,6 +751,37 @@ impl BinarySerde for Bincode {
...
+    type DeserializeError = Self::SerializeError;
+
+    fn new() -> Self {
+        Self(bincode::DefaultOptions::new())
+    }
+
+    fn serialize_impl<T: Serialize>(&self, t: &T) -> Result<Vec<u8>, Self::SerializeError> {
+        // Serializes the object directly into a Writer without include the length.
+        let mut writer = Vec::new();
+        self.0.serialize_into(&mut writer, t)?;
```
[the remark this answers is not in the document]
xinifinity: Simply just to match with MerkleDB encoding, which doesn't include the length.

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,92 @@ pub enum EncodedNodeType {
...
+
+            let path = PartialPath::from_nibbles(Nibbles::<0>::new(&node.path).into_iter()).0;
+            let node = EncodedNodeType::Leaf(LeafNode { path, data });
+            Ok(Self::new(node))
+        } else {
+            let mut children: [Option<Vec<u8>>; BranchNode::MAX_CHILDREN] = Default::default();
+            let value = node.data.map(Data).filter(|data| !data.is_empty());
+
+            for (i, chd) in node.chd {
+                children[i as usize] = Some(chd).filter(|chd| !chd.is_empty());
```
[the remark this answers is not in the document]
xinifinity: Good point, I should not filter this.

firewood/src/merkle.rs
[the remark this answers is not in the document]
xinifinity: Hmm, tried to do it but encountered difficulties, it seems this requires giving different types for `merkle`  based on condition. Do you have some suggestions?

firewood/src/merkle/node.rs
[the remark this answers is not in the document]
richardpringle: 🤔, might want to still use the length serializer to figure out capacity. 

Can you create an issue for this, "Pre-determine Vec capacity for Serialization". 

We should benchmark both cases. Don't need to do it now though.

firewood/benches/hashops.rs
```diff
@@ -104,7 +104,7 @@ fn bench_merkle<const N: usize>(criterion: &mut Criterion) {
                     )
                     .unwrap();
 
-                    let merkle = Merkle::new(Box::new(store));
+                    let merkle = MerkleWithEncoder::new(Box::new(store));
```
richardpringle, line 107: 🙌

firewood/src/merkle.rs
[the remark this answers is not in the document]
richardpringle: I think I see the issue here. I think you should delete the `Node::encode` method since we now have `Merle::Encode` (same goes for decode). 

Also, I don't think that checking the root hash is necessary. If `node == new_node`, then `node.get_root_hash(store) == new_node.get_root_hash(store)` when using the same `store`. You've already verified that the inputs to `get_root_hash` are the same, and I think it's overkill to test for determinism. Even if we do want to test for determinism, that should be done in a separate test.

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,92 @@ pub enum EncodedNodeType {
...
+                    .enumerate()
+                    .filter_map(|(i, c)| {
+                        c.as_ref().map(|c| {
+                            if c.len() >= TRIE_HASH_LEN {
+                                (i as u64, Keccak256::digest(c).to_vec())
+                            } else {
+                                (i as u64, c.to_vec())
+                            }
+                        })
+                    })
```
richardpringle, line 531:
```suggestion
                    .filter_map(|(i, c)| c.as_ref().map(|c| (i as u64, c)))
                    .map(|(i, c)| if c.len() >= TRIE_HASH_LEN {
                        (i, Keccak256::digest(c).to_vec())
                    } else {
                        (i, c.to_vec())
                    })
```

firewood/src/merkle/node.rs
```diff
@@ -490,6 +495,92 @@ pub enum EncodedNodeType {
...
+            } else {
+                Data(Vec::new())
+            };
+
+            let path = PartialPath::from_nibbles(Nibbles::<0>::new(&node.path).into_iter()).0;
+            let node = EncodedNodeType::Leaf(LeafNode { path, data });
+            Ok(Self::new(node))
+        } else {
+            let mut children: [Option<Vec<u8>>; BranchNode::MAX_CHILDREN] = Default::default();
+            let value = node.data.map(Data).filter(|data| !data.is_empty());
```
richardpringle, line 569: I think I asked this somewhere else, but how do we different empty data and no data?

richardpringle (changes requested):

firewood/src/merkle/node.rs
xinifinity: Good question, this is following existing encoding style as https://github.com/ava-labs/firewood/blob/main/firewood/src/merkle/node.rs#L590 and https://github.com/ava-labs/firewood/blob/main/firewood/src/merkle/node/branch.rs#L119. But probably I should not do that for MerkleDB encoding schema. I will remove it.

firewood/src/merkle.rs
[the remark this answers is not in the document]
xinifinity: :+1, however I cannot remove `Node::encode` as the new encode/decode method doesn't cover extension node. https://github.com/ava-labs/firewood/blob/main/firewood/src/merkle.rs#L98

firewood/src/merkle/node.rs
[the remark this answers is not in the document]
xinifinity: https://github.com/ava-labs/firewood/issues/431

firewood/src/merkle.rs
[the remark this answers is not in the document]
richardpringle: Can we create an issue for this, please?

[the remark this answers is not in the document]
xinifinity: https://github.com/ava-labs/firewood/issues/432

firewood/src/merkle/node.rs
xinifinity: Created an issue for fixing existing Bincode encoding schema https://github.com/ava-labs/firewood/issues/433

richardpringle (approved):

PR #430, opened 2023-12-11
Handle Extension-nodes in Iterator

richardpringle:
- Move MerkleStream to submodule
- Make MerkleStream handle branches with values
- Stop prefetching next node in iterator

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -318,21 +318,12 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
 }
 
 impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
-    pub fn stream<K: KeyType>(
-        &self,
-    ) -> Result<merkle::MerkleKeyValueStream<'_, S, Bincode>, api::Error> {
-        self.merkle
-            .iter(self.header.kv_root)
-            .map_err(|e| api::Error::InternalError(Box::new(e)))
+    pub fn stream(&self) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
+        self.merkle.iter(self.header.kv_root)
     }
 
-    pub fn stream_from<K: KeyType>(
-        &self,
-        start_key: K,
-    ) -> Result<merkle::MerkleKeyValueStream<'_, S, Bincode>, api::Error> {
-        self.merkle
-            .iter_from(start_key, self.header.kv_root)
-            .map_err(|e| api::Error::InternalError(Box::new(e)))
+    pub fn stream_from(&self, start_key: Vec<u8>) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
+        self.merkle.iter_from(self.header.kv_root, start_key)
     }
 
     fn flush_dirty(&mut self) -> Option<()> {
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -320,10 +320,18 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
 impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
     pub fn stream<K: KeyType>(
         &self,
-        start_key: Option<K>,
     ) -> Result<merkle::MerkleKeyValueStream<'_, S, Bincode>, api::Error> {
         self.merkle
-            .get_iter(start_key, self.header.kv_root)
+            .iter(self.header.kv_root)
+            .map_err(|e| api::Error::InternalError(Box::new(e)))
+    }
+
+    pub fn stream_from<K: KeyType>(
+        &self,
+        start_key: K,
+    ) -> Result<merkle::MerkleKeyValueStream<'_, S, Bincode>, api::Error> {
+        self.merkle
+            .iter_from(start_key, self.header.kv_root)
             .map_err(|e| api::Error::InternalError(Box::new(e)))
     }
 
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -322,7 +322,10 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
         self.merkle.iter(self.header.kv_root)
     }
 
-    pub fn stream_from(&self, start_key: Vec<u8>) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
+    pub fn stream_from(
+        &self,
+        start_key: Box<[u8]>,
+    ) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
         self.merkle.iter_from(self.header.kv_root, start_key)
     }
 
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1221,13 +1221,24 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
-    pub(crate) fn get_iter<K: AsRef<[u8]>>(
+    pub(crate) fn iter(
         &self,
-        key: Option<K>,
         root: DiskAddress,
     ) -> Result<MerkleKeyValueStream<'_, S, T>, MerkleError> {
         Ok(MerkleKeyValueStream {
-            key_state: IteratorState::new(key),
+            key_state: IteratorState::new(),
+            merkle_root: root,
+            merkle: self,
+        })
+    }
+
+    pub(crate) fn iter_from<K: AsRef<[u8]>>(
+        &self,
+        key: K,
+        root: DiskAddress,
+    ) -> Result<MerkleKeyValueStream<'_, S, T>, MerkleError> {
+        Ok(MerkleKeyValueStream {
+            key_state: IteratorState::with_key(key),
             merkle_root: root,
             merkle: self,
         })
@@ -1245,9 +1256,11 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             return Ok(None);
         }
 
-        let mut stream = self
-            .get_iter(first_key, root)
-            .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+        let mut stream = match first_key {
+            Some(key) => self.iter_from(key, root),
+            None => self.iter(root),
+        }
+        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
 
         // fetch the first key from the stream
         let first_result = stream.next().await;
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -24,7 +24,6 @@ pub use node::{
     NodeType, PartialPath,
 };
 pub use proof::{Proof, ProofError};
-use stream::IteratorState;
 pub use stream::MerkleKeyValueStream;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
@@ -1221,27 +1220,16 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
-    pub(crate) fn iter(
-        &self,
-        root: DiskAddress,
-    ) -> Result<MerkleKeyValueStream<'_, S, T>, MerkleError> {
-        Ok(MerkleKeyValueStream {
-            key_state: IteratorState::new(),
-            merkle_root: root,
-            merkle: self,
-        })
+    pub(crate) fn iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
+        MerkleKeyValueStream::new(self, root)
     }
 
-    pub(crate) fn iter_from<K: AsRef<[u8]>>(
+    pub(crate) fn iter_from(
         &self,
-        key: K,
         root: DiskAddress,
-    ) -> Result<MerkleKeyValueStream<'_, S, T>, MerkleError> {
-        Ok(MerkleKeyValueStream {
-            key_state: IteratorState::with_key(key),
-            merkle_root: root,
-            merkle: self,
-        })
+        key: Vec<u8>,
+    ) -> MerkleKeyValueStream<'_, S, T> {
+        MerkleKeyValueStream::from_key(self, root, key)
     }
 
     pub(super) async fn range_proof<K: api::KeyType + Send + Sync>(
@@ -1252,15 +1240,14 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         limit: Option<usize>,
     ) -> Result<Option<api::RangeProof<Vec<u8>, Vec<u8>>>, api::Error> {
         // limit of 0 is always an empty RangeProof
-        if let Some(0) = limit {
+        if limit == Some(0) {
             return Ok(None);
         }
 
         let mut stream = match first_key {
-            Some(key) => self.iter_from(key, root),
+            Some(key) => self.iter_from(root, key.as_ref().to_vec()),
             None => self.iter(root),
-        }
-        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+        };
 
         // fetch the first key from the stream
         let first_result = stream.next().await;
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1227,7 +1227,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
     pub(crate) fn iter_from(
         &self,
         root: DiskAddress,
-        key: Vec<u8>,
+        key: Box<[u8]>,
     ) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::from_key(self, root, key)
     }
@@ -1245,7 +1245,8 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         }
 
         let mut stream = match first_key {
-            Some(key) => self.iter_from(root, key.as_ref().to_vec()),
+            // TODO: fix the call-site to force the caller to do the allocation
+            Some(key) => self.iter_from(root, key.as_ref().to_vec().into_boxed_slice()),
             None => self.iter(root),
         };
 
@@ -1264,7 +1265,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             .map_err(|e| api::Error::InternalError(Box::new(e)))?;
         let limit = limit.map(|old_limit| old_limit - 1);
 
-        let mut middle = vec![(first_key, first_data)];
+        let mut middle = vec![(first_key.into_vec(), first_data)];
 
         // we stop streaming if either we hit the limit or the key returned was larger
         // than the largest key requested
@@ -1283,8 +1284,9 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
                     };
 
                     // keep going if the key returned is less than the last key requested
-                    ready(kv.0.as_slice() <= last_key.as_ref())
+                    ready(&*kv.0 <= last_key.as_ref())
                 })
+                .map(|kv_result| kv_result.map(|(k, v)| (k.into_vec(), v)))
                 .try_collect::<Vec<(Vec<u8>, Vec<u8>)>>()
                 .await?,
         );
--- a/firewood/src/merkle/stream.rs
+++ b/firewood/src/merkle/stream.rs
@@ -359,7 +359,7 @@ mod helper_types {
 
     /// Enums enable stack-based dynamic-dispatch as opposed to heap-based `Box<dyn Trait>`.
     /// This helps us with match arms that return different types that implement the same trait.
-    /// It's possible that https://github.com/rust-lang/rust/issues/63065 will make this unnecessary.
+    /// It's possible that [rust-lang/rust#63065](https://github.com/rust-lang/rust/issues/63065) will make this unnecessary.
     ///
     /// And this can be replaced by the `either` crate from crates.io if we ever need more functionality.
     pub(super) enum Either<T, U> {
--- a/firewood/src/merkle/stream.rs
+++ b/firewood/src/merkle/stream.rs
@@ -16,11 +16,7 @@ pub(super) enum IteratorState<'a> {
     /// Start iterating at the specified key
     StartAtKey(Vec<u8>),
     /// Continue iterating after the given `next_node` and parents
-    Iterating {
-        visit_last: bool,
-        next_result: Option<(Vec<u8>, Vec<u8>)>,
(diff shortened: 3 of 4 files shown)


firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,296 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+        .skip(1) // always skip the sentinal node
+        .flat_map(|(parent, child_nibble)| match parent.inner() {
+            NodeType::Branch(_) => vec![*child_nibble],
+            NodeType::Extension(extension) => extension.path.to_vec(),
+            NodeType::Leaf(leaf) => leaf.path.to_vec(),
+        })
+}
+
+fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Vec<u8> {
+    let mut data = Vec::with_capacity(nibbles.size_hint().0);
```
richardpringle, line 335: Not going to change this right away, but noticed reviewing my own code

```suggestion
    let mut data = Vec::with_capacity(nibbles.size_hint().0 / 2);
```

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,293 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+
+        *node = ParentNode::Visited(next_parent);
+        *pos = next_pos;
+    }
+
+    Ok(true.into())
+}
+
+/// create an iterator over the key-nibbles from all parents _excluding_ the sentinal node.
+fn nibble_iter_from_parents<'a>(parents: &'a [(ObjRef, u8)]) -> impl Iterator<Item = u8> + 'a {
```
rkuris, line 320: I think this does a lot of allocations, since you are creating a vec for each component as you traverse. I think it's possible to return a `impl Iterator<Item = &'a u8>` with something like this (importing the `either` crate here, but there are other ways):

```
fn nibble_iter_from_parents<'a>(parents: &'a [(ObjRef, u8)]) -> impl Iterator<Item = &'a u8> {
    parents
        .iter()
        .skip(1) // always skip the sentinal node
        .flat_map(|(parent, child_nibble)| match parent.inner() {
            NodeType::Branch(_) => either::Either::Left(std::iter::once(child_nibble)),
            NodeType::Extension(extension) => either::Either::Right(extension.path.iter()),
            NodeType::Leaf(leaf) => either::Either::Right(leaf.path.iter()),
        })
}
```

There is then some work to always use references in the callers, of course, but that actually seemed a bit easier as you're using `copied()` to match this in.

rkuris (changes requested):
Partial review

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,293 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                if let NodeType::Branch(branch) = root.inner() {
+                    if let Some(value) = branch.value.as_ref() {
+                        let key = vec![];
+                        let value = value.to_vec();
+
+                        self.key_state = IteratorState::Iterating {
+                            parents: vec![(root, 0)],
+                        };
+
+                        return Poll::Ready(Some(Ok((key, value))));
```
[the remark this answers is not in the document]
richardpringle: Yeah, I think this code is wrong. This is supposed to be the case where's there's data in the root, but "root" is not actually the root-node, it's the sentinel-node. 

I'll add the test for the empty key and delete this code (it's dead code).

firewood/src/merkle/stream.rs
richardpringle: `either` is a good idea, will try it out.

rkuris: Why did you decide not to go with an iterator over `&'a u8` as suggested above?

rkuris: > Why did you decide not to go with an iterator over `&'a u8` as suggested above?

Answering my own question: it's actually faster since `size_of::<u8>` < `size_of::<&u8>`

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+impl<T> From<T> for MustUse<T> {
+    fn from(t: T) -> Self {
+        Self(t)
+    }
+}
+
+impl<T: Not> Not for MustUse<T> {
+    type Output = T::Output;
+
+    fn not(self) -> Self::Output {
```
rkuris, line 273: Q: Is calling `not` sufficient in your use case for `must_use`, or does this need annotating as well?

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
-    while let (Some(hi), Some(lo)) = (iter.next(), iter.next()) {
+        .skip(1) // always skip the sentinal node
+        .flat_map(|(parent, child_nibble)| match parent.inner() {
+            NodeType::Branch(_) => Either::Left(std::iter::once(*child_nibble)),
+            NodeType::Extension(extension) => Either::Right(extension.path.iter().copied()),
+            NodeType::Leaf(leaf) => Either::Right(leaf.path.iter().copied()),
+        })
+}
+
+enum Either<T, U> {
```
rkuris, line 319: I have a similar set of questions to the previous set around `MustUse`; I'll avoid repeating them here.

I read that `Either` was removed from `std` during rust development because it was thought that the overly generic enums of `Left` and `Right` produced less readable code. For those users that still wanted it, they moved it over to the `either` crate.

Q1: Why did you choose to reimplement `Either` here rather than use the `either` crate?

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            Self::Left(left) => left.next(),
+            Self::Right(right) => right.next(),
+        }
+    }
+}
+
+fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Vec<u8> {
```
rkuris, line 339: Q: why is this here and not in the implementation of `Nibbles`, perhaps as `Nibbles::into_vec(&self)`?

rkuris (commented):
Still need to review the algorithm more, but I added a few questions in the meantime.

firewood/src/merkle/stream.rs
richardpringle: Yeah, I think I generally prefer to copy early if I'm going to need to copy (not clone), even if it was a `u64`

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
-        .collect::<Vec<u8>>()
+fn get_children_iter(branch: &BranchNode) -> impl Iterator<Item = (DiskAddress, u8)> {
+    branch
+        .children
+        .into_iter()
+        .enumerate()
+        .filter_map(|(pos, child_addr)| child_addr.map(|child_addr| (child_addr, pos as u8)))
+}
+
+#[must_use]
```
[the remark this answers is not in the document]
richardpringle: A1: yes
A2: I can change it to a `bool` if you prefer, it definitely does not need to be generic
A3: I put zero thought into its location. Happy to move it.

To be honest, I'm not a huge fan of this code, but it works, and it prevents someone from ignoring the result and forgetting to early-return. I have some ideas on how I could improve the ergonomics here, but great is the enemy of good enough, so I've been holding off.

firewood/src/merkle/stream.rs
richardpringle: Yep, otherwise clippy would fail because of `-Dwarnings` in CI. I actually wasn't sure about this either, so I made sure I got the warning when I didn't "use" the `MustUse`

firewood/src/merkle/stream.rs
richardpringle: The `either` crate comes with a bunch of baggage that we don't need. I would argue that if we find more use cases for it other than this one place, then we should pull it in, but I don't think we need to pull in a whole crate just to get a little stack-based dynamic-dispatch that can be implemented in fewer than 20-lines of code.

firewood/src/merkle/stream.rs
richardpringle: I took the `key_from_parents_and_leaf` function and split it up such that I could generically pass in an iterator (sometimes with a chained component, sometimes without). 

Do you know if there's anywhere else we need this functionality?

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                Poll::Ready(next_result)
             }
+
             IteratorState::StartAtKey(key) => {
-                // TODO: support finding the next key after K
-                let root_node = self
-                    .merkle
-                    .get_node(self.merkle_root)
+                if key.is_empty() {
+                    self.key_state = IteratorState::StartAtBeginning;
```
rkuris, line 77: I'm not sure this is strictly necessary, but since there is code here, can you add a test case for it? I don't see one where we start an iterator with an empty key.

firewood/src/merkle/stream.rs
```diff
@@ -1,25 +1,22 @@
...
     StartAtBeginning,
     /// Start iterating at the specified key
     StartAtKey(Vec<u8>),
-    /// Continue iterating after the given last_node and parents
-    Iterating {
-        last_node: ObjRef<'a>,
-        parents: Vec<(ObjRef<'a>, u8)>,
-    },
+    /// Continue iterating after the given `next_node` and parents
+    Iterating { parents: Vec<(ObjRef<'a>, u8)> },
```
rkuris, line 19: I can't add a comment to line 30 below, which is the implementation of Default for IteratorState. I couldn't find a use case where you're using this in the new code, so I believe the code that implements `Default` can be removed.

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                let (found_node, mut parents) = merkle
                     .get_node_and_parents_by_key(root_node, &key)
                     .map_err(|e| api::Error::InternalError(Box::new(e)))?;
 
-                let Some(last_node) = found_node else {
-                    return Poll::Ready(None);
-                };
+                if let Some(found_node) = found_node {
+                    let value = match found_node.inner() {
+                        NodeType::Branch(branch) => branch.value.as_ref(),
```
rkuris, line 91: Can we add a test case that has a value in a branch to cover this?

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                        NodeType::Leaf(leaf) => Some(&leaf.data),
+                        NodeType::Extension(_) => None,
```
rkuris, line 93: Q: Is it possible that `found_node` is actually a `NodeType::Extension`? If so, under what conditions? Can we add a test case for that?

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+    loop {
+        match node.inner() {
+            New(NodeType::Leaf(leaf)) => {
+                let value = leaf.data.to_vec();
+                return Ok(Some((node.into_node(), value)));
+            }
+
+            Visited(NodeType::Leaf(_)) | Visited(NodeType::Extension(_)) => {
+                let Some((next_parent, next_pos)) = visited_parents.pop() else {
+                    return Ok(None);
```
rkuris, line 198: I wasn't able to follow the logic here. Can you add some comments as to why we're returning `Ok(None)`?

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
         .iter()
-        .map(|parent| parent.1)
-        .chain(leaf.path.to_vec());
-    let mut data = Vec::with_capacity(iter.size_hint().0);
-    while let (Some(hi), Some(lo)) = (iter.next(), iter.next()) {
+        .skip(1) // always skip the sentinal node
+        .flat_map(|(parent, child_nibble)| match parent.inner() {
+            NodeType::Branch(_) => Either::Left(std::iter::once(*child_nibble)),
+            NodeType::Extension(extension) => Either::Right(extension.path.iter().copied()),
+            NodeType::Leaf(leaf) => Either::Right(leaf.path.iter().copied()),
```
rkuris, line 315: Q: How do `parents` contain a `Leaf`? Isn't this case `unreachable!()`?

rkuris (changes requested):
Round 2 of questions

firewood/src/merkle/stream.rs
rkuris: Fair. What are your thoughts about making it easy to remove later by perhaps sticking it in it's own module (can even be in the same source file)?

firewood/src/merkle/stream.rs
rkuris: I think there are other callers that need to convert nibbles back into bytes, but I'd have to look. I'm worried about testing the case where there are an odd number of nibbles and someone calls this method, and I think those tests don't belong here either.

firewood/src/merkle/stream.rs
```diff
@@ -54,236 +48,301 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
         mut self: std::pin::Pin<&mut Self>,
         _cx: &mut std::task::Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        // Note that this sets the key_state to StartAtBeginning temporarily
-        let found_key = match std::mem::take(&mut self.key_state) {
+        let MerkleKeyValueStream {
```
rkuris, line 51: nit: why not `let Self {` ...?

firewood/src/merkle/stream.rs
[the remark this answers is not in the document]
rkuris: Bool would be better. This could use some doc comments as well, particularly around why you're doing this and why it's important in this particular case. Rust doesn't have a good way to force a must use of the inner part of the result.

firewood/src/merkle/stream.rs
richardpringle: If you want me to do that, I can, but I really don't see the benefit. All you have to do is delete the type.

firewood/src/merkle/stream.rs
richardpringle: I'm not sure how you would test that without creating an invalid tree. 

The only other conversion *from* nibbles is for `PartialPath`, that I'm aware of.

firewood/src/merkle/stream.rs
rkuris: Your choice. For each comment, here are your options:

- You disagree with the review comment, in which case you should state why and resolve the comment.
- You agree with the review comment but it's not stopping the commit because you'll fix it later. If you don't have a follow-on commit pending, you'll create an issue.
- You agree with the review comment and will fix it before landing.
- There's a comment that refers to style or preference or a suggestion that doesn't matter to the reviewer very much (a "nit"). You can choose to ignore these, or treat them like one of the other 3. I will sometimes ignore them if they are marked nit.

If you don't see the benefit, then you've met the first bullet and you should resolve this comment.

firewood/src/merkle/stream.rs
richardpringle: I see what you're saying about testing. I can add some tests depending on the outcome of the offline discussion about https://github.com/ava-labs/firewood/pull/435

richardpringle: Resolving this as the discussion is now moved to #435

firewood/src/merkle/stream.rs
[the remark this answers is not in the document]
richardpringle: fixed

firewood/src/merkle/stream.rs
richardpringle: Great suggestion, thanks again

firewood/src/merkle/stream.rs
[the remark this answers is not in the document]
richardpringle: I moved both `Either` and `MustUse` to a sub-mod within the same file called `helper_types`. I left `MustUse` as generic because the `#[must_use]` attribute itself is generic. When I read
```rust
#[must_use]
struct MustUse(bool) 
```
I ask myself what the purpose of the `bool` is. Whereas with the generic type, I know that I have some wrapper that I must do *something* with. It's also obvious that it only serves *one* purpose since `T` could be anything including `()`.

firewood/src/merkle/stream.rs
richardpringle: I moved the `Either` type to a module

firewood/src/merkle/stream.rs
richardpringle: deleted

firewood/src/merkle/stream.rs
richardpringle: removed

firewood/src/merkle/stream.rs
richardpringle: done!

firewood/src/merkle/stream.rs
richardpringle: The simple answer is no, `found_node` can't be an extension. But it is problematic when the key lands us on an extension and I have added a test case for it.

firewood/src/merkle/stream.rs
richardpringle: `Fused` is intended (it was easier to write that way now that we aren't calling `std::mem::take` on `IteratorState`. It would be possible to have a cycling stream where we would just wrap the stream with an additional type that resets the state once the end is reached. If we find such a use case, it will be trivial to write. 

As for this particular line you originally commented on, it is interesting. If we assume there is always going to be a sentinel-node, we'll never hit this case. However, if we ignore that assumption, this function doesn't care about the existence of a sentinel node or not. If there are no other nodes to traverse, simply return. It's the more general solution and works in both cases.

richardpringle: Oh, and I added the `FusedStream` implementation as suggested.

firewood/src/merkle/stream.rs
richardpringle: Bad naming... as discussed in the walkthrough, `parents` as been renamed to `visited_node_path`.

firewood/src/merkle/stream.rs
```diff
@@ -1,25 +1,22 @@
...
-    /// Continue iterating after the given last_node and parents
-    Iterating {
-        last_node: ObjRef<'a>,
-        parents: Vec<(ObjRef<'a>, u8)>,
-    },
+    /// Continue iterating after the given `next_node` and parents
+    Iterating { parents: Vec<(ObjRef<'a>, u8)> },
 }
 impl IteratorState<'_> {
     pub(super) fn new<K: AsRef<[u8]>>(starting: Option<K>) -> Self {
```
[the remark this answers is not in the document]
richardpringle: done!

firewood/src/db.rs
```diff
@@ -318,13 +318,15 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
...
+    pub fn stream_from(
         &self,
-        start_key: Option<K>,
-    ) -> Result<merkle::MerkleKeyValueStream<'_, S, Bincode>, api::Error> {
-        self.merkle
-            .get_iter(start_key, self.header.kv_root)
-            .map_err(|e| api::Error::InternalError(Box::new(e)))
+        start_key: Box<[u8]>,
+    ) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
+        self.merkle.iter_from(self.header.kv_root, start_key)
```
richardpringle, line 329: This is a significant improvement to the API. It was such a pain specifying the type on the `None` when it could be any number of things. It's too bad there's no `default` generic, like with structs and enums.

firewood/src/merkle.rs
```diff
@@ -1264,7 +1265,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             .map_err(|e| api::Error::InternalError(Box::new(e)))?;
         let limit = limit.map(|old_limit| old_limit - 1);
 
-        let mut middle = vec![(first_key, first_data)];
+        let mut middle = vec![(first_key.into_vec(), first_data)];
```
richardpringle, line 1268: I wasn't sure about this, but we can change it later.  `into_vec` does not allocate.

firewood/src/merkle/stream.rs
```diff
@@ -394,22 +559,216 @@ mod tests {
...
+
+            assert_eq!(&*next.0, &*next.1);
+            assert_eq!(&*next.0, key);
+        }
+
+        check_stream_is_done(stream).await;
+    }
+
+    #[tokio::test]
+    async fn start_at_key_greater_than_all_others() {
```
xinifinity, line 735: thanks a lot for the thorough test lists! On top of that may be `start_at_key_smaller_than_all_others` ? I guess this may be already covered, but just to make it explicitly covered.

firewood/src/merkle/stream.rs
```diff
@@ -1,290 +1,398 @@
...
-                    return Poll::Ready(None);
-                };
+                Poll::Ready(next)
+            }
+        }
+    }
+}
+
+enum NodeRef<'a> {
+    New(ObjRef<'a>),
```
xinifinity, line 163: ❤️

firewood/src/merkle/stream.rs
```diff
@@ -1,290 +1,398 @@
...
+
+        visited_path.push((next_node, 0));
+
+        (key, value)
+    });
+
+    Ok(next)
+}
+
+fn find_next_node_with_data<'a, S: ShaleStore<Node>, T>(
```
xinifinity, line 210: nit: as the core function, I feel it deserves some high level comments describe how things work.

xinifinity (approved):
My review was mostly focus on the algorithm and the test cases coverage and all comments are nits that can be ignored or addressed in a follow up.

rkuris (approved):
Removing "requested changes" status. Hao has completed the review here.

PR #435, opened 2023-12-13
Add nibbles_into_bytes for an arbitrary iterator

rkuris:
This method is useful for converting nibbles back into their corresponding bytes.

Added some simple tests, both in the success case and the failure case.

### DIFF
--- a/firewood/src/nibbles.rs
+++ b/firewood/src/nibbles.rs
@@ -140,23 +140,8 @@ impl<'a, const LEADING_ZEROES: usize> DoubleEndedIterator for NibblesIterator<'a
     }
 }
 
-pub(crate) trait IntoBytes: Iterator<Item = u8> {
-    fn nibbles_into_bytes(&mut self) -> Vec<u8> {
-        let mut data = Vec::with_capacity(self.size_hint().0 / 2);
-
-        while let (Some(hi), Some(lo)) = (self.next(), self.next()) {
-            data.push((hi << 4) + lo);
-        }
-
-        data
-    }
-}
-impl<T: Iterator<Item = u8>> IntoBytes for T {}
-
 #[cfg(test)]
 mod test {
-    use crate::nibbles::IntoBytes;
-
     use super::Nibbles;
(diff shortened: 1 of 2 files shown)


rkuris:
This PR came to be from the review of #430

After applying this diff, the code in that PR `key_from_nibbles_iter` moves here, so the one caller can just do:
```rust
let key = nibble_iter_from_parents(parents).chain(node_path_iter).nibbles_into_bytes();
```

firewood/src/nibbles.rs
```diff
@@ -140,8 +140,23 @@ impl<'a, const LEADING_ZEROES: usize> DoubleEndedIterator for NibblesIterator<'a
...
+        let mut data = Vec::with_capacity(self.size_hint().0 / 2);
+
+        while let (Some(hi), Some(lo)) = (self.next(), self.next()) {
+            data.push((hi << 4) + lo);
+        }
+
+        data
+    }
+}
+impl<T: Iterator<Item = u8>> IntoBytes for T {}
```
richardpringle, line 154: I think this would be great if we had `Item = Nibble` where `#[repr(transparent)] struct Nibble(u8)` (where you can only create the nibble if it's within the correct bounds).

Instead, I think something like this would make the call more explicit

```rust
/// CAUTION: only use with nibble iterators  
pub(crate) fn nibbles_into_bytes<Iter: Iterator<Item = u8>>(iter: Iter) -> Vec<u8> {
    let mut data = Vec::with_capacity(self.size_hint().0 / 2);

    while let (Some(hi), Some(lo)) = (self.next(), self.next()) {
        data.push((hi << 4) + lo);
    }

    data
}
```

richardpringle (changes requested):
Could we identify where this would be used outside of `stream.rs` before we merge? If the use case only exists in one module, I think the abstraction should live there until we find more use cases.

firewood/src/nibbles.rs
rkuris: Created #440 for this larger change

xinifinity (approved):

richardpringle (approved):

Issue #433, opened 2023-12-12
Differentiate between empty data and no data for Bincode encode schema


xinifinity:
As mentioned [here](https://github.com/ava-labs/firewood/pull/425#discussion_r1424432630) fix https://github.com/ava-labs/firewood/blob/main/firewood/src/merkle/node.rs#L590 and https://github.com/ava-labs/firewood/blob/main/firewood/src/merkle/node/branch.rs#L119.

rkuris:
No longer using bincode with new design

PR #436, opened 2023-12-13
Make encode test generic to cover all schemas

xinifinity:


firewood/src/merkle.rs
```diff
@@ -1491,28 +1491,22 @@ mod tests {
...
-    }
-
-    #[test_case(leaf(Vec::new(), Vec::new()) ; "empty leaf encoding")]
-    #[test_case(leaf(vec![1, 2, 3], vec![4, 5]) ; "leaf encoding")]
-    #[test_case(branch(Some(b"value".to_vec()), vec![1, 2, 3].into()) ; "branch with chd")]
-    #[test_case(branch(Some(b"value".to_vec()), Some(Vec::new())); "branch with empty chd")]
-    #[test_case(branch(Some(Vec::new()), vec![1, 2, 3].into()); "branch with empty value")]
-    fn node_encode_decode_plain(node: Node) {
-        let merkle = create_generic_test_merkle::<PlainCodec>();
+    #[test_case(Bincode::new(), leaf(Vec::new(), Vec::new()) ; "empty leaf encoding with Bincode")]
```
rkuris, line 1494: nit: this might work better with the [test-matrix](https://docs.rs/test-case/latest/test_case/#test-matrix) option, which seems to have a way of creating a cartesian product of possible tests.

I didn't test that it would work in this use case, but it might simplify this code and ensure that if we add new ones we don't miss both test cases.

firewood/src/merkle/node.rs
```diff
@@ -722,6 +722,7 @@ pub trait BinarySerde {
     ) -> Result<T, Self::DeserializeError>;
 }
 
+#[derive(Default)]
```
rkuris, line 725: nit: I'm not sure why this was added here.

Q: Will all new codecs require `Default`? If so, we might want to be more explicit elsewhere.

rkuris (approved):

firewood/src/merkle.rs
xinifinity: yeah, thought about that, although unless https://github.com/ava-labs/firewood/issues/433 is addressed, some test cases cannot be merged.

firewood/src/merkle/node.rs
xinifinity: no, added now only for test cases usage.

firewood/src/merkle.rs
richardpringle: woooow... `test-matrix` looks really nice and should definitely be used where we can.

Issue #426, opened 2023-12-08
Document how to setup and run docker on the mac for development


rkuris:
Ideally we use the same setup. Currently I'm using a separate VM for this.

richardpringle:
TL;DR (putting this here for now, will make the proper docs though)

- Setup a dev-environment using your repo-directory as the project directory [^1].
- Configure Docker-Desktop to use `VirtioFS`
- Move your `CARGO_TARGET_DIR` to somewhere that **is not** source-mapped onto the host. 

Then do all the normal Linux stuff. Download `rustup` to get cargo, etc. Set up your build-essentials. These will not be things that I will add to the doc, instead, whoever is setting their stuff up for the first time should add as they see fit. I don't remember all the different things I had to do, but they are no different than setting up any Linux environment. 

If you are seeing build failures, there's a concurrency issue with building onto mapped volumes. You need the target-dir to only live in the container. On your host machine, you don't need to change the behaviour. For step by step debugging, you'll need to use a remote-vs-code instance that you can start from docker-desktope. Otherwise, I do all my development from the host machine, and keep a terminal open for running the entire test suite in the container. Most of the `Merkle` tests, etc. can run fine in the mac environment. 

[^1]: https://docs.docker.com/desktop/dev-environments/

richardpringle:
![image](https://github.com/ava-labs/firewood/assets/3286504/83d6b66d-19e3-4b59-bc73-f67cf68d7329)

richardpringle:
![image](https://github.com/ava-labs/firewood/assets/3286504/2f573c42-10df-4af3-b253-2522820b3ce6)

richardpringle:
![image](https://github.com/ava-labs/firewood/assets/3286504/88c981cb-42b9-4b99-acec-fbca31cca652)

PR #437, opened 2023-12-13
Add docker instructions

richardpringle:
This is pretty basic. I did this a while ago so I have no idea what steps I'm missing. 

For anyone that follows these instructions, I would encourage you to PR my branch with necessary changes before we merge the PR. 

[RENDERED](https://github.com/ava-labs/firewood/blob/docker-docs/README.docker.md)

### DIFF
--- a/README.docker.md
+++ b/README.docker.md
@@ -1,0 +1,82 @@
+# Docker on Mac Compatibility
+
+Note:
+Docker compatiblitiy is a work in progress. Please PR any changes here if you find a better way to do this.
+
+## Steps
+
+### Step 1
+
+Install docker-desktop ([guide](https://docs.docker.com/desktop/install/mac-install/))
+
+### Step 2
+
+Setup a dev-environment ([guide](https://docs.docker.com/desktop/dev-environments/set-up/#set-up-a-dev-environment))
+
+Here, you want to specifically pick a local-directory (the repo's directory)
+
+![image](https://github.com/ava-labs/firewood/assets/3286504/83d6b66d-19e3-4b59-bc73-f67cf68d7329)
+
+This is best because you can still do all your `git` stuff from the host.
+
+### Step 3
+
+You will need the `Dev Containers` VSCODE extension, authored by Microsoft for this next step.
+
+Open your dev-environment with VSCODE. Until you do this, the volume might not be properly mounted. If you (dear reader) know of a better way to do this, please open a PR. VSCODE is very useful for its step-by-step debugger, but other than that, you can run whatever IDE you would like in the host environment and just open a shell in the container to run the tests.
+
+![image](https://github.com/ava-labs/firewood/assets/3286504/88c981cb-42b9-4b99-acec-fbca31cca652)
+
+### Step 4
+
+Open a terminal in vscode OR exec into the container directly as follows
+
+```sh
+# you don't need to do this if you open the terminal from vscode
+# the container name here is "firewood-app-1", you should be able to see this in docker-desktop
+docker exec -it --privileged -u root firewood-app-1 zsh
+```
+
+Once you're in the terminal you'll want to install the Rust toolset. You can find instructions [here](https://rustup.rs/)
+
+**!!! IMPORTANT !!!**
+
+Make sure you read the output of any commands that you run. `rustup` will likely ask you to `source` a file to add some tools to your `PATH`.
+
+You'll also need to install all the regular linux dependencies (if there is anything from this list that's missing, please add to this README)
+
+```sh
+apt update
+apt install vim
+apt install build-essential
+apt install protobuf-compiler
+```
+
+### Step 5
+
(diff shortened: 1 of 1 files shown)


richardpringle:
closes https://github.com/ava-labs/firewood/issues/426

rkuris (dismissed):
These instructions are missing a lot of steps. Some known issues:

 - Rust is not installed in the VM. This should get installed by docker or provide instructions on how to do it
 - Plugins for vscode are missing

I added some changes to the instructions already, but this needs more TLC

rkuris:
@richardpringle what are your thoughts about making this a draft? I don't think it's ready for review from others at this point; it's still some effort to get it into the correct state.

rkuris:
Another thought here: since docker is probably a temporary thing, how do you feel about a `README.docker.md` that contains these instructions, which will avoid rendering all these images and instructions from the main page.

Another choice might be to move these to `README.build.md` and include the other build instructions.

richardpringle:
> @richardpringle what are your thoughts about making this a draft? I don't think it's ready for review from others at this point; it's still some effort to get it into the correct state.

Why isn't it ready for review? Could you post the errors you run into? Sure the container might be missing rustup, but you have to know how to get that no matter what environment you're working on and there are plenty of well written tutorials out there for how to do that. 

If you could post any issues that you're running into, I might be able to solve them quickly.

richardpringle:
> Another thought here: since docker is probably a temporary thing, how do you feel about a `README.docker.md` that contains these instructions, which will avoid rendering all these images and instructions from the main page.
> 
> Another choice might be to move these to `README.build.md` and include the other build instructions.

Not opposed

README.md
```diff
@@ -123,7 +123,7 @@ be developed for this milestone.
       and Firewood.
 - [ ] :runner: Support replicating the full state with corresponding range proofs that
       verify the correctness of the data.
-- [ ] Pluggable IO subsystem (tokio\_uring, monoio, etc)
+- [ ] Pluggable IO subsystem (tokio_uring, monoio, etc)
```
rkuris, line 126: Unfortunately, an underscore is a special character in markdown, so not having the escape means the markdown renderer has to look for another underscore later in the file. Different implementations may behave differently as to how far forward it will search for the closing underscore.

As of right now, I don't think there is another underscore anywhere else in the file.

Why was this backslash removed?

README.docker.md
```diff
@@ -0,0 +1,82 @@
...
+apt install vim
+apt install build-essential
+apt install protobuf-compiler
+```
+
+### Step 5
+
+**!!! IMPORTANT !!!**
+
+You need to create a separate `CARGO_TARGET_DIR` that isn't volume mounted onto the host. `VirtioFS` has some concurrency issues when dealing with sequential writes and reads to a volume that is mounted to the host. You can put a directory here for example: `/root/target`.
```
xinifinity, line 59: nit: `VirtioFS` (the default filesystem being used)

README.docker.md
```diff
@@ -0,0 +1,82 @@
...
+
+There are a couple of places where this can be setup. If you're a `zsh` user, you should add `export CARGO_TARGET_DIR=/root/target` to either `/root/.zshrc` or `/root/.bashrc`.
+After adding the line, don't forget to `source` the file to make sure your current session is updated.
+
+### Step 6
+
+Navigate to `/com.docker.devenvironments.code ` and run `cargo test`. If it worked, you are most of the way there! If it did not work, there are a couple of common issues. If the code will not compile, it's possible that your target directory isn't set up properly. Check inside `/root/target` to see if there are any build artifacts. If not, you might need to call `source ~/.zshrc` again (sub in whatever your preferred shell is).
+
+Now for vscode, you need to configure your `rust-analyzer` in the "remote-environment" (the Docker container). There are a couple of places to do this. First, you want to open `/root/.vscode-server/Machine/settings.json` and make sure that you have the following entry:
```
xinifinity, line 70: It seems in my case I don't need to manually configure `rust-analyzer` and I cannot find`/root/.vscode-server/Machine/settings.json`.

xinifinity (approved):
Thanks a lot for the thorough instruction! Somehow `rust-analyzer` is automatically setup in my environment so didn't verify that part.

README.md
xinifinity: +1

richardpringle: Whoops, that was my formatter. I think an underscore surrounded by letters cannot be used for italics which is why my formatter does that. 

I didn't mean to add this change though, I'll remove it.

README.docker.md
richardpringle: Sorry @xinifinity, I merged this without addressing your comment here. Should I add a comment saying that you should create the file if it doesn't already exist? It's possible that it's already grabbing the value from your `.bashrc` file, so maybe that's why it's not needed. @dboehm-avalabs and I did not test the two configurations in isolation.

Issue #448, opened 2023-12-20
`Obj.flush_dirty` should not be  called in `drop` if not dirty.


xinifinity:
This will save additionally operation to set the `dirty` field and extra call stacks.

PR #450, opened 2023-12-20
Only call `Obj.flush_dirty` in `drop` if it is actually dirty

xinifinity:
Closes https://github.com/ava-labs/firewood/issues/448.

After the change, according to the flamegraph of `insert` example using [sync API](https://github.com/ava-labs/firewood/commit/bfbd148dbefe49d232c55963a4afb91d89a8e868), ~3% time reduced in `drop_in_place` call and ~2% in `flush_dirty`.

Before 
insert_sync_flamegraph_before

After
insert_sync_flamegraph_after

The flamegraph is generated using `cargo flamegraph --example insert-sync
` the default is to use `release` profile

### DIFF
--- a/firewood/src/shale/mod.rs
+++ b/firewood/src/shale/mod.rs
@@ -149,7 +149,9 @@ impl<T: Storable> Obj<T> {
 
 impl<T: Storable> Drop for Obj<T> {
     fn drop(&mut self) {
-        self.flush_dirty()
+        if self.dirty.is_some() {
+            self.flush_dirty()
+        }
     }
 }
 
--- a/firewood/src/shale/mod.rs
+++ b/firewood/src/shale/mod.rs
@@ -134,15 +134,13 @@ impl<T: Storable> Obj<T> {
     }
 
     pub fn flush_dirty(&mut self) {
-        if !self.value.is_mem_mapped() {
-            if let Some(new_value_len) = self.dirty.take() {
-                let mut new_value = vec![0; new_value_len as usize];
-                // TODO: log error
-                self.value.write_mem_image(&mut new_value).unwrap();
-                let offset = self.value.get_offset();
-                let bx: &mut dyn CachedStore = self.value.get_mut_mem_store();
-                bx.write(offset, &new_value);
-            }
+        if let Some(new_value_len) = self.dirty.take() {
+            let mut new_value = vec![0; new_value_len as usize];
+            // TODO: log error
+            self.value.write_mem_image(&mut new_value).unwrap();
+            let offset = self.value.get_offset();
+            let bx: &mut dyn CachedStore = self.value.get_mut_mem_store();
+            bx.write(offset, &new_value);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


richardpringle:
🤔, that's interesting that you got a speed up because the `flush_dirty` method already checks if there's a value there. Can you get the same speedup if you swap the `if` statements inside the `flush_dirty` method?

firewood/src/shale/mod.rs
```diff
@@ -149,7 +149,9 @@ impl<T: Storable> Obj<T> {
 
 impl<T: Storable> Drop for Obj<T> {
     fn drop(&mut self) {
-        self.flush_dirty()
+        if self.dirty.is_some() {
+            self.flush_dirty()
+        }
```
richardpringle, line 154: Upon further review, it appears that `is_mem_mapped()` will always return false (part of the `flush_dirty` method above). That method should be removed entirely. It's a trait method (part of `Storable`), which means there's an indirection every time it's called and it's possible that the optimizer has a hard time figuring out that there isn't a single non-default implementation of the method (which just defaults to `false`).

Further, I wouldn't expect `self.dirty.take()` to take longer if `self.dirty == None`, so you shouldn't need that extra check. Did you see the 3% drop in time in `release` mode? 

I suspect that part of the slowdown here is also due to the double indirection in `StoredView::mem` (the property). Its type is `Box<dyn SendSyncDerefMut<Target = dyn CachedStore>>`. We have to go to two different vtables just to call the `CachedStore::write` trait-method. 

I wonder how difficult it would be to remove that `dyn SendSyncDerefMut` though? It's definitely a type of lifetime hack.

Anyway, enough thinking out loud. The one concrete thing that should be changed and measured is the removal of `is_mem_mapped` and see if you get a decrease in time spend. 

Please post the result in the conversation section along with the compiler flags used so that we can keep ourselves honest about compiling in release-mode.

xinifinity: yeah, I'm using release mode as that's the default for cargo `flamegraph`.

Remove is_mem_mapped() is not showing perf decrease, so added it as well. But I do think `self.dirty.take()` has a cost here even for `self.dirty == None`, probably because this is a very critical path even a tiny diff is shown.

firewood/src/shale/mod.rs
```diff
@@ -134,22 +134,22 @@ impl<T: Storable> Obj<T> {
...
-        if !self.value.is_mem_mapped() {
-            if let Some(new_value_len) = self.dirty.take() {
-                let mut new_value = vec![0; new_value_len as usize];
-                // TODO: log error
-                self.value.write_mem_image(&mut new_value).unwrap();
-                let offset = self.value.get_offset();
-                let bx: &mut dyn CachedStore = self.value.get_mut_mem_store();
-                bx.write(offset, &new_value);
-            }
+        if let Some(new_value_len) = self.dirty.take() {
```
richardpringle, line 137: This might seem like a nit, but I actually don't think it is... your change is an optimization for calling `flush_dirty` and should therefore exist in the `flush_dirty` code instead.

```suggestion
        // faster than calling `self.dirty.take()` on a `None`
        if self.dirty.is_none() {
            return;
        }

        if let Some(new_value_len) = self.dirty.take() {
```

richardpringle (changes requested):
Small suggestion, but it means that we get the same speed up everywhere `flush_dirty` is called, not just on drop.

richardpringle (approved):

PR #454, opened 2023-12-21
Delay block in place

rkuris:
We can delay calling `block_in_place` until we actually block. This results in a small performance gain, probably more noticeable when we start doing reads in parallel.

Before:
```
Your branch is up to date with 'origin/main'.
❯ cargo run --quiet --release --example insert -- -i 1000 -s 1
Generated and inserted 1000 batches of size 1 in 5.022415705s
❯ cargo run --quiet --release --example insert -- -i 1000 -s 1
Generated and inserted 1000 batches of size 1 in 5.410185903s
❯ cargo run --quiet --release --example insert -- -i 1000 -s 1
Generated and inserted 1000 batches of size 1 in 5.059761131s
```

After
```
❯ cargo run --quiet --release --example insert -- -i 1000 -s 1
Generated and inserted 1000 batches of size 1 in 4.911608567s
❯ cargo run --quiet --release --example insert -- -i 1000 -s 1
Generated and inserted 1000 batches of size 1 in 4.906468629s
❯ cargo run --quiet --release --example insert -- -i 1000 -s 1
Generated and inserted 1000 batches of size 1 in 4.918876902s
```

### DIFF
--- a/firewood/src/storage/buffer.rs
+++ b/firewood/src/storage/buffer.rs
@@ -623,7 +623,7 @@ impl DiskBufferRequester {
             .send(BufferCmd::CollectAsh(nrecords, resp_tx))
             .map_err(StoreError::Send)
             .ok();
-        resp_rx.blocking_recv().map_err(StoreError::Receive)
+        block_in_place(|| resp_rx.blocking_recv().map_err(StoreError::Receive))
     }
 
     /// Register a cached space to the buffer.
--- a/firewood/src/storage/buffer.rs
+++ b/firewood/src/storage/buffer.rs
@@ -20,6 +20,7 @@ use growthring::{
     walerror::WalError,
     WalFileImpl, WalStoreImpl,
 };
+use tokio::task::block_in_place;
 use tokio::{
     sync::{
         mpsc,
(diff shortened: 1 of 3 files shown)


xinifinity (commented):
Seems there are tests failing?

firewood/src/storage/buffer.rs
```diff
@@ -569,7 +570,7 @@ impl DiskBufferRequester {
             .send(BufferCmd::GetPage((space_id, page_id), resp_tx))
             .map_err(StoreError::Send)
             .ok();
-        resp_rx.blocking_recv().unwrap()
+        block_in_place(move || resp_rx.blocking_recv().unwrap())
```
richardpringle, line 573: Just guessing that you don't need this here because you're not in an async function

xinifinity: +1 will it work if not in an async call?

richardpringle (approved):

firewood/src/storage/buffer.rs
rkuris: Without this, all the tests fail. This code does nothing if you're not in an async context. If you are, then it moves you to a blockable thread to prevent interference with the async runtime.

rkuris:
> Seems there are tests failing?
Should be good now. I found another instance where we making a blocking call in an async context.

xinifinity (approved):

Issue #449, opened 2023-12-20
`Proposal.commit` should consume the committed DbRev directly as `DbRevInner.base_revision`


xinifinity:
During `Proposal.commit` we should directly move the underlying DbRev to be committed to `DbRevInner.base_revision`, to
1. avoid allocation/deallocation of `Merkle`,
2. it may also save some root hash calculation. 

This requires to map a `StoreRevMut` to `StoreRevShared`.

xinifinity:
Current progress captured in [hao/perf ](https://github.com/ava-labs/firewood/tree/hao/perf)branch. The summary is 

1.  The clean approach is to combine `StoreRevMut` and `StoreRevShared` into one struct, example from @rkuris 
```
#[derive(Debug)]
struct Revision<S> {
    base_space: RwLock<Arc<dyn MemStoreR>>,
    delta: StoreDelta,
    state: S,
}

#[derive(Debug)]
struct RevisionStateMutable {
    prev_deltas: Arc<RwLock<StoreRevMutDelta>>,
}

impl From<Revision<RevisionStateMutable>> for Revision<RevisionStateImmutable> {
    fn from(value: Revision<RevisionStateMutable>) -> Self {
        Self {
            base_space: value.base_space,
            delta: value.delta,
            state: RevisionStateImmutable{},
        }
    }
}

#[derive(Debug)]
struct RevisionStateImmutable {}
```
However, for a quick check we are doing hacks to covert StoreRevMut` to `StoreRevShared`.

2.  Another problem is `CompactSpaceInner` wraps the underlying space with `Arc` so we cannot simply consume the underlying space. 

3. The other option is to build a batch directly on DB instead of using Proposal in the case no batch is built on other uncommitted one, if we cannot fix `Proposal` in short term.

xinifinity:
Actually the `Arc` wrap is redundant in `CompactSpaceInner`. Removed it now can see at least ~40% time reduced.

```
cargo run --release --quiet --example insert -- -i 1000
Generated and inserted 1000 batches of size 1 in 3.48867496s
```

PR #456, opened 2023-12-22
`Proposal.commit` consume the underlying `DbRev` directly to the base committed revision

xinifinity:
Close https://github.com/ava-labs/firewood/issues/449

With this change we can reduce around ~(40% - 50%) of time using the `insert` example as the benchmark.

```
~/src/firewood$ cargo run --release --quiet --example insert -- -i 1000
Generated and inserted 1000 batches of size 1 in 3.106775653s
~/src/firewood$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
~/src/firewood$ cargo run --release --quiet --example insert -- -i 1000
Generated and inserted 1000 batches of size 1 in 6.001036863s
```

However there are serval improvements need to happen after this PR,

1.  It takes the short route to convert a `StoreRevMut` to `StoreRevShared`, a more idiomatic way to combine these two into one struct should happen.
2. It disabled the recursive commit feature of proposal as we now consume the proposal when we commit.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -58,6 +58,7 @@ const MAGIC_STR: &[u8; 16] = b"firewood v0.1\0\0\0";
 
 pub type MutStore = CompactSpace<Node, StoreRevMut>;
 pub type SharedStore = CompactSpace<Node, StoreRevShared>;
+// pub type SharedStore = CompactSpace<Node, StoreRev>;
 
 #[derive(Debug)]
 #[non_exhaustive]
@@ -829,7 +830,7 @@ impl Db {
             m: Arc::clone(&self.inner),
             r: Arc::clone(&self.revisions),
             cfg: self.cfg.clone(),
-            rev: Box::new(rev),
+            rev,
             store,
             committed: Arc::new(Mutex::new(false)),
             parent,
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -139,11 +139,11 @@ impl SubUniverse<StoreRevShared> {
     }
 }
 
-impl SubUniverse<Arc<StoreRevMut>> {
-    fn new_from_other(&self) -> SubUniverse<Arc<StoreRevMut>> {
+impl SubUniverse<StoreRevMut> {
+    fn new_from_other(&self) -> SubUniverse<StoreRevMut> {
         SubUniverse {
-            meta: Arc::new(StoreRevMut::new_from_other(self.meta.as_ref())),
-            payload: Arc::new(StoreRevMut::new_from_other(self.payload.as_ref())),
+            meta: StoreRevMut::new_from_other(&self.meta),
+            payload: StoreRevMut::new_from_other(&self.payload),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


firewood/src/db.rs
```diff
@@ -381,6 +381,15 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
...
 }
 
+impl DbRev<MutStore> {
+    pub fn into_shared(self) -> DbRev<SharedStore> {
+        DbRev {
+            header: self.header,
+            merkle: self.merkle.into(),
+        }
+    }
+}
```
rkuris, line 391: If you just implement From, you get into() for free. This change plus changing proposal.rs to call into() instead of into_shared() just works.
```suggestion
impl From<DbRev<MutStore>> for DbRev<SharedStore> {
    fn from(value: DbRev<MutStore>) -> Self {
        DbRev {
            header: value.header,
            merkle: value.merkle.into(),
        }
    }
}
```

firewood/src/db.rs
```diff
@@ -462,7 +471,7 @@ impl Db {
     }
 
     /// Open a database.
-    fn new_internal<P: AsRef<Path>>(db_path: P, cfg: DbConfig) -> Result<Self, DbError> {
+    pub fn new_internal<P: AsRef<Path>>(db_path: P, cfg: DbConfig) -> Result<Self, DbError> {
```
rkuris, line 474: nit: Let's not make this available outside the crate. We have a visibility problem already.
```suggestion
    pub(crate) fn new_internal<P: AsRef<Path>>(db_path: P, cfg: DbConfig) -> Result<Self, DbError> {
```

firewood/src/merkle.rs
```diff
@@ -71,6 +72,16 @@ pub struct Merkle<S, T> {
     phantom: PhantomData<T>,
 }
 
+impl<T> Merkle<MutStore, T> {
```
rkuris, line 75: Same pattern here. If you `impl<T> From<Merkle<MutStore, T>> for Merkle<SharedStore, T>` you get into (and from, if you happen to need it) for free.

firewood/src/shale/compact.rs
```diff
@@ -216,24 +218,36 @@ impl Storable for CompactSpaceHeader {
...
-    meta_space: Arc<M>,
-    compact_space: Arc<M>,
+    meta_space: M,
+    compact_space: M,
     header: CompactSpaceHeaderSliced,
     alloc_max_walk: u64,
     regn_nbit: u64,
 }
 
+impl CompactSpaceInner<StoreRevMut> {
```
rkuris, line 228: Same here

firewood/src/shale/compact.rs
```diff
@@ -522,14 +536,24 @@ impl<T: Storable, M: CachedStore> CompactSpace<T, M> {
     }
 }
 
+impl CompactSpace<Node, StoreRevMut> {
```
rkuris, line 539: and here

firewood/src/storage/mod.rs
```diff
@@ -401,6 +401,22 @@ impl StoreRevMut {
         }
     }
 
+    pub fn into_shared(self) -> StoreRevShared {
```
rkuris, line 404: Probably here too

rkuris (approved):
Lots of followup comments. The conversions can be more idiomatic.

xinifinity:
> Lots of followup comments. The conversions can be more idiomatic.

Sounds good, will address in a follow up. Thanks!

PR #460, opened 2024-01-02
Use nibbles iterator in proof code

richardpringle:
While this isn't strictly necessary, it makes dealing with branchs with paths much simpler. I also believe that the code for handling leaves may have previously been wrong. It checked that leaf's partial path was contained by the key, but I think it was possible for the key to contain the partial-path without completely matching. I could be wrong, but using the nibbles iterator is much simpler IMO.

### DIFF
--- a/firewood/src/merkle/proof.rs
+++ b/firewood/src/merkle/proof.rs
@@ -466,8 +466,25 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
     ///
     /// # Arguments
     ///
+    /// * `merkle` - The merkle tree.
+    ///
+    /// * `key_nibbles` - The nibbles iterator of the key.
+    ///
+    /// * `buf` - The encoded node.
+    ///
     /// * `end_node` - A boolean indicates whether this is the end node to decode, thus no `key`
     ///                to be present.
+    ///
+    ///  # Returns `Result<(addr, subproof, nibbles), ProofError>`
+    ///
+    /// * `addr` - The new address of the node after it's decoded and inserted into the cache.
+    ///
+    /// * `subproof` - An `Option<SubProof>`, `None` if the sub-proof does not exist or the node's partial-path
+    /// does not match the given key
+    ///
+    /// * `nibbles` - The remaining nibbles iterator of the key or an empty iterator if the node's partial-path
+    /// does not match the given key
+    ///
     fn decode_node<'a, S: ShaleStore<Node> + Send + Sync, T: BinarySerde>(
         &self,
         merkle: &Merkle<S, T>,
--- a/firewood/src/merkle/proof.rs
+++ b/firewood/src/merkle/proof.rs
@@ -492,7 +492,11 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
                 Ok((addr, subproof.into(), key_nibbles))
             }
             NodeType::Extension(n) => {
-                let paths_match = n.path.iter().copied().eq(&mut key_nibbles);
+                let paths_match = n
+                    .path
+                    .iter()
+                    .copied()
+                    .all(|nibble| Some(nibble) == key_nibbles.next());
 
                 if !paths_match {
                     return Ok((addr, None, Nibbles::new(&[]).into_iter().peekable()));
--- a/firewood/src/merkle/proof.rs
+++ b/firewood/src/merkle/proof.rs
@@ -3,8 +3,6 @@
 
 use std::cmp::Ordering;
 use std::collections::HashMap;
-use std::iter::Peekable;
-use std::ops::Deref;
 
 use crate::shale::{disk_address::DiskAddress, ShaleError, ShaleStore};
 use crate::v2::api::HashKey;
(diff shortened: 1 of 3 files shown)


firewood/src/merkle/proof.rs
```diff
@@ -331,15 +337,15 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-            if u_ref.inner().as_branch().is_some() {
-                #[allow(clippy::indexing_slicing)]
-                (branch_index = chunks[key_index] as usize);
+            #[allow(clippy::unwrap_used)]
+            if node_ref.inner().as_branch().is_some() {
+                child_index = key_nibbles.next().unwrap() as usize;
             }
 
-            key_index += size;
+            // key_index += size;
```
richardpringle, line 348: whoops, forgot to delete this

richardpringle:
🤔  all these tests pass locally...

rkuris:
These fail on my machine as well. with RUST_BACKTRACE=1 I get:
```

---- test_all_elements_proof stdout ----
thread 'test_all_elements_proof' panicked at /home/rk/open-source/firewood/firewood/src/shale/compact.rs:630:30:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:127:5
   3: core::option::Option<T>::unwrap
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/option.rs:935:21
   4: <firewood::shale::compact::CompactSpace<T,M> as firewood::shale::ShaleStore<T>>::get_item
             at ./src/shale/compact.rs:630:24
   5: firewood::merkle::Merkle<S,T>::get_node
             at ./src/merkle.rs:87:9
   6: firewood::merkle::proof::unset_node_ref
             at ./src/merkle/proof.rs:917:21
   7: firewood::merkle::proof::unset_node_ref
             at ./src/merkle/proof.rs:957:13
   8: firewood::merkle::proof::unset_node_ref
             at ./src/merkle/proof.rs:951:13
   9: firewood::merkle::proof::unset_node_ref
             at ./src/merkle/proof.rs:951:13
  10: firewood::merkle::proof::unset_internal
             at ./src/merkle/proof.rs:748:13
  11: firewood::merkle::proof::Proof<N>::verify_range_proof
             at ./src/merkle/proof.rs:237:13
  12: firewood::merkle_util::MerkleSetup<S,T>::verify_range_proof
             at ./src/merkle_util.rs:115:9
  13: merkle::test_all_elements_proof
             at ./tests/merkle.rs:689:5
  14: merkle::test_all_elements_proof::{{closure}}
             at ./tests/merkle.rs:659:33
  15: core::ops::function::FnOnce::call_once
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/ops/function.rs:250:5
  16: core::ops::function::FnOnce::call_once
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

richardpringle:
> 🤔 all these tests pass locally...

I lied...

firewood/src/merkle/proof.rs
```diff
@@ -473,41 +480,52 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                 // Check if the subproof with the given key exist.
-                #[allow(clippy::indexing_slicing)]
-                let index = key[0] as usize;
-                #[allow(clippy::indexing_slicing)]
-                let Some(data) = &n.chd_encode()[index] else {
-                    return Ok((addr, None, 1));
-                };
-                let subproof = generate_subproof(data.to_vec())?;
-                Ok((addr, Some(subproof), 1))
+                let index = *key_nibbles.peek().unwrap() as usize;
```
rkuris, line 519: New code should not call `unwrap()`

Choices:
 - Don't use the `if` restriction on the switch, and do that test here (preferred). You can easily bail if the `peek` returns `None`.
 - Since there's no way this unwrap could ever fail, then use `expect("explanation")` to explain why this unwrap means "logically, this can't happen so this is a bug in our code".
 
 I don't llike the second choice but it's better than adding another unwrap.

firewood/src/merkle/proof.rs
```diff
@@ -473,41 +480,52 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                if !key.contains_other(cur_key) {
-                    return Ok((addr, None, 0));
+                let paths_match = n
+                    .path
+                    .iter()
+                    .copied()
+                    .all(|nibble| Some(nibble) == key_nibbles.next());
+
+                if !paths_match {
+                    return Ok((addr, None, Nibbles::new(&[]).into_iter().peekable()));
```
rkuris, line 501: question (please answer by adding a comment if possible)

Why are we returning an empty nibbles iterator instead of None or something like that?

firewood/src/merkle/proof.rs
```diff
@@ -352,31 +353,31 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
 
-                            chd_ptr = self.decode_node(merkle, cur_key, proof.as_ref(), true)?.0;
+                            chd_ptr = self
+                                .decode_node(merkle, key_nibbles, proof.as_ref(), true)?
+                                .0;
 
                             // Link the child to the parent based on the node type.
-                            match &u_ref.inner() {
+                            #[allow(clippy::indexing_slicing)]
+                            match &node_ref.inner() {
```
rkuris, line 371: q: what does the TODO comment mean on the next line? Isn't it obsolete at this point?

firewood/src/merkle/proof.rs
```diff
@@ -352,31 +353,31 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
                 }
                 Some(p) => {
                     // Return when reaching the end of the key.
-                    if key_index == chunks.len() {
-                        #[allow(clippy::indexing_slicing)]
-                        (cur_key = &chunks[key_index..]);
+                    if key_nibbles.size_hint().0 == 0 {
```
rkuris, line 356: Nit: the other case is much easier to handle earlier (find the comment "The trie doesn't contain the key") so it should be moved here, reducing the indendation and complexity of this block.

firewood/src/merkle/proof.rs
```diff
@@ -330,16 +333,14 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-            u_ref = merkle.get_node(chd_ptr)?;
+            node_ref = merkle.get_node(chd_ptr)?;
 
             // If the new parent is a branch node, record the index to correctly link the next child to it.
-            if u_ref.inner().as_branch().is_some() {
-                #[allow(clippy::indexing_slicing)]
-                (branch_index = chunks[key_index] as usize);
+            #[allow(clippy::unwrap_used)]
+            if node_ref.inner().as_branch().is_some() {
+                child_index = key_nibbles.next().unwrap() as usize;
```
rkuris, line 341: New code should not use `unwrap()`.

It's not clear why this unwrap is safe. Why are we sure that there is another nibble at this point?

xinifinity:
> It checked that leaf's partial path was contained by the key, but I think it was possible for the key to contain the partial-path without completely matching.

Are you saying the previous logic may not check if it reach the end of the key if it is a leaf?

firewood/src/merkle/proof.rs
```diff
@@ -458,13 +457,21 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
     ///
     /// * `end_node` - A boolean indicates whether this is the end node to decode, thus no `key`
     ///                to be present.
-    fn decode_node<S: ShaleStore<Node> + Send + Sync, T: BinarySerde>(
+    // TODO: return the child-index
```
xinifinity, line 460: What is the TODO for?

firewood/src/merkle/proof.rs
```diff
@@ -473,41 +480,52 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
 
         match new_node.inner() {
             NodeType::Leaf(n) => {
-                let cur_key = n.path().0.as_ref();
+                let paths_match = n.path.iter().copied().eq(&mut key_nibbles);
 
-                if !key.contains_other(cur_key) {
```
xinifinity, line 478: It looks like in the leaf node case, the old code only check if it contains the leaf's partial path, which probably is what we desired? Because to handle the non-existing proof case.

rkuris (dismissed):
Please answer the listed questions. This will require another pass for approval.

firewood/src/merkle/proof.rs
richardpringle: It's to handle paths in branches. They aren't handled in this PR.

firewood/src/merkle/proof.rs
richardpringle: Not necessary anymore, will delete

firewood/src/merkle/proof.rs
richardpringle: I can't make it do the same thing as extension nodes. I actually think this code was incorrect as `key` could be `[1, 2, 3]` and `cur_key` could be `[2, 3]`. This case should early return but it wouldn't have with the old code, right?

firewood/src/merkle/proof.rs
richardpringle: I'll add doc-comments to the function. 

It's better to handle iteration generically. An empty iterator means there are no more nibbles to iterate over and we should handle the result accordingly.

firewood/src/merkle/proof.rs
richardpringle: Yeah, this needs to be fixed. Working on it

firewood/src/merkle/proof.rs
richardpringle: Doc comments added. I'm going to resolve this comment as you can open new comments on the new doc-comments directly

firewood/src/merkle/proof.rs
richardpringle: fixed

firewood/src/merkle/proof.rs
richardpringle: I think doing this made the code a little harder to review FYI, but the resultant code is much nicer... I probably should have done this in a follow up

firewood/src/merkle/proof.rs
richardpringle: fixed

firewood/src/merkle/proof.rs
rkuris: Not sure I understand this comment. `contains_other` is a horribly-named function that just checks that it starts with the same thing. These tests show that `[1,2,3]` and `[2,3]` do not return `true`.

```
    // false tests
    assert!(![1,2,3].as_slice().contains_other(&[2,3]));
    assert!(![2,3].as_slice().contains_other(&[1,2,3]));
    
    // true tests
    assert!([1,2,3].as_slice().contains_other(&[1,2]));
```

rkuris:
There is now a dead public method named `contains_other` that should be removed from `proof.rs` (at the end).

firewood/src/merkle/proof.rs
richardpringle: Ah, "contains_other" implies a subset and was misleading, you are correct. In any case, the new code matches the old behaviour now

firewood/src/merkle/proof.rs
```diff
@@ -350,103 +345,118 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+
+                    // Decode the last subproof to get the value.
+                    let node_ref = if let SubProof::Hash(p_hash) = p {
+                        let proof = proofs_map
+                            .get(&p_hash)
+                            .ok_or(ProofError::ProofNodeMissing)?;
+
+                        // If the new parent is a branch node, record the index to correctly link the next child to it.
+                        if node_ref.inner().as_branch().is_some() {
+                            let Some(index) = key_nibbles.next().map(|i| i as usize) else {
```
rkuris, line 383: It seems like if we get here, we know that `key_nibbles.size_hint().0 == 0`, otherwise we would have matched the other match statement. I think this means the `next` method is guaranteed to return `None`. If that's not true, can you explain why?

richardpringle:
> There is now a dead public method named `contains_other` that should be removed from `proof.rs` (at the end).

removed

firewood/src/merkle/proof.rs
```diff
@@ -258,261 +257,146 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                            Ok(None)
-                        } else {
-                            Err(ProofError::NodeNotInTrie)
-                        };
-                    };
-
-                    cur_hash = hash;
-                    #[allow(clippy::indexing_slicing)]
-                    (cur_key = &chunks[key_index..]);
+                    n.chd_encoded().map(Some).ok_or(ProofError::InvalidData)?
```
rkuris, line 347: This line needs a comment. I'm not sure what's happening here.

firewood/src/merkle_util.rs
```diff
@@ -64,7 +64,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T: BinarySerde> MerkleSetup<S, T> {
             .map_err(|_err| DataStoreError::GetError)
     }
 
-    pub const fn get_root(&self) -> DiskAddress {
+    pub const fn get_sentinel(&self) -> DiskAddress {
```
rkuris, line 67: question: Why is this pub?

Why not return the actual node here? I think all callers will always want to fetch that node right away anyway.

As always, feel free to defer that to another PR. I realize you only changed the name.

If you keep this name, I'd prefer `get_sentinel_address` or similar.

firewood/src/merkle/proof.rs
```diff
@@ -563,26 +447,25 @@ fn locate_subproof(
...
-                .ok_or(ProofError::InvalidData)?
-                .to_vec();
+                .ok_or(ProofError::InvalidData)?;
             generate_subproof(data).map(|subproof| (Some(subproof), key_nibbles))
         }
     }
 }
 
-fn generate_subproof(encoded: Vec<u8>) -> Result<SubProof, ProofError> {
+fn generate_subproof_hash(encoded: &[u8]) -> Result<[u8; 32], ProofError> {
```
rkuris, line 456: nit: Proof code uses the type TrieHash for this. Can we reuse it?

rkuris (commented):
Partial review, needs another pass

firewood/src/merkle/proof.rs
```diff
@@ -258,261 +257,146 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-        let mut branch_index = 0;
-
-        loop {
-            let cur_proof = proofs_map
-                .get(&cur_hash)
-                .ok_or(ProofError::ProofNodeMissing)?;
-            // TODO(Hao): (Optimization) If a node is already decode we don't need to decode again.
-            let (mut chd_ptr, sub_proof, size) =
-                self.decode_node(merkle, cur_key, cur_proof.as_ref(), false)?;
+        let mut child_index = 0;
```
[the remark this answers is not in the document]
richardpringle: 0 is used for the sentinel node because the root-node is always at the 0th child of the sentinel; then, this is only set and used for branch nodes after that. 

The reason we have a variable here is that we have to use the nibble from the iterator twice, once to get the encoded-child and then again set child-address on the parent. 

Where does `(u64, Vec)` come from here and what would it be used for? Is that the encoded-node? And did you mean `u8`? I'm a little confused with what you're suggesting, but I agree that `child_index` is a little awkward here.

firewood/src/merkle/proof.rs
```diff
@@ -258,261 +257,146 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                         // insert the leaf to the empty slot
                         #[allow(clippy::unwrap_used)]
-                        u_ref
-                            .write(|u| {
-                                let uu = u.inner_mut().as_branch_mut().unwrap();
+                        parent_node_ref
+                            .write(|node| {
                                 #[allow(clippy::indexing_slicing)]
-                                (uu.chd_mut()[branch_index] = Some(chd_ptr));
+                                let node = node.inner_mut().as_branch_mut().unwrap();
```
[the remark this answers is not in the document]
richardpringle: I can use `expect`. This pattern is used everywhere because it's the only way to write to the `inner` value of the node. We pattern match on `node.inner` but have to call `write` on `node`.

firewood/src/merkle/proof.rs
richardpringle: Will do this in a follow up

xinifinity (approved):
Thanks a lot for making the changes. I think it improves the readability a lot!

xinifinity (approved):

firewood/src/merkle/proof.rs
```diff
@@ -258,261 +257,146 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+                        })
+                        .unwrap();
+
+                    child_node
                 }
 
+                NodeType::Extension(n) => merkle.get_node(n.chd())?,
+
                 // We should not hit a leaf node as a parent.
                 _ => return Err(ProofError::InvalidNode(MerkleError::ParentLeafBranch)),
```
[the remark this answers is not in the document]
richardpringle: We should not panic there. I think it would make sense to return a "database is corrupt" error here, but if we panic, that doesn't give the database user the opportunity to gracefully handle anything else they have in flight. It should be the user's decision whether or not to panic.

firewood/src/merkle/proof.rs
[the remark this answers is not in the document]
rkuris: Sorry, I wasn't super clear. I'm suggesting that you preserve the iterator instead of the offset if you're on a branch node and are iterating that branch node. Then, to get the next child, all you have to do is advance that iterator, possibly using `as_ref()` to avoid terminating the iterator or `clone()` if you want to just look at it and not advance it.

This would let you do some additional simplifications like if the value is in the branch node, just initialize but don't search the iterator. It would also prevent you from going backwards as was shown in the recent StreamIterator bug.

firewood/src/merkle/proof.rs
```diff
@@ -56,6 +56,8 @@ pub enum ProofError {
     Shale(ShaleError),
     #[error("invalid root hash")]
     InvalidRootHash,
+    #[error("failed writing the node")]
+    WriteError,
```
rkuris, line 60:
```suggestion
WriteError(#[from] ObjWriteError)
```
I think that right now `ObjWriteError` doesn't actually save the original error, but it should, and when it does, it would be nice to have it propagated here. This also avoids the `map_err` with a discard and lets you use ?

This also fixes the case that someone changes `write` to return something other than `ObjWriteError` and doesn't check all the callers for their handling.

firewood/src/merkle/proof.rs
```diff
@@ -258,261 +259,150 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+    merkle: &'a Merkle<S, T>,
+    proofs_map: &HashMap<[u8; 32], N>,
+    child_hash: &[u8; 32],
+) -> Result<ObjRef<'a>, ProofError> {
+    let child_proof = proofs_map
+        .get(child_hash)
+        .ok_or(ProofError::ProofNodeMissing)?;
+    let child_node = NodeType::decode(child_proof.as_ref())?;
+    merkle
+        .put_node(Node::from(child_node))
```
rkuris, line 402: This returns `MerkleError` and `ProofError` seems to already have a derived `into()` method from this type, so I think you can just use ? here. If that's wrong, just ignore this comment.

rkuris (approved):
Minor comments remain. Would be good to improve the error propagation though.

PR #471, opened 2024-01-05
Layer the object cache to avoid recomputing serialized values

rkuris:


rkuris:
Going in a different direction here

PR #474, opened 2024-01-05
Persist `is_encoded_longer_than_hash_len` property of a Node

xinifinity:
Closes https://github.com/ava-labs/firewood/issues/472

This PR ensures we store `is_encoded_longer_than_hash_len` property of a Node and reload when reading a node from disk, so that the persisted root hash can be used. It also ensures root hash calculation is done before we flushing the writes in a proposal so it can be persisted.

With this change, the performance of small key insertion with large number of batches dramatically improved. For example with 1000 batches of size 10, dropped from 27s to 1s.  Before,
```
$ cargo run --release --quiet --example insert -- -b 10 -n 1000 -s 0
Generated and inserted 1000 batches of size 10 in 27.246639082s
```

After,
```
$ cargo run --release --quiet --example insert -- -b 10 -n 1000 -s 0
Generated and inserted 1000 batches of size 10 in 1.521389874s
```

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -157,6 +157,7 @@ impl NodeType {
 pub struct Node {
     pub(super) root_hash: OnceLock<TrieHash>,
     encoded: OnceLock<Vec<u8>>,
+    is_encoded_longer_than_hash_len: OnceLock<bool>,
     // lazy_dirty is an atomicbool, but only writers ever set it
     // Therefore, we can always use Relaxed ordering. It's atomic
     // just to ensure Sync + Send.
@@ -173,6 +174,7 @@ impl PartialEq for Node {
         let Node {
             root_hash,
             encoded,
+            is_encoded_longer_than_hash_len: _,
             lazy_dirty: _,
             inner,
         } = self;
@@ -187,6 +189,7 @@ impl Clone for Node {
     fn clone(&self) -> Self {
         Self {
             root_hash: self.root_hash.clone(),
+            is_encoded_longer_than_hash_len: self.is_encoded_longer_than_hash_len.clone(),
             encoded: self.encoded.clone(),
             lazy_dirty: AtomicBool::new(self.is_dirty()),
             inner: self.inner.clone(),
@@ -199,6 +202,7 @@ impl From<NodeType> for Node {
         let mut s = Self {
             root_hash: OnceLock::new(),
             encoded: OnceLock::new(),
+            is_encoded_longer_than_hash_len: OnceLock::new(),
             inner,
             lazy_dirty: AtomicBool::new(false),
         };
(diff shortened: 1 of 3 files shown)


firewood/src/merkle/node.rs
```diff
@@ -157,6 +157,7 @@ impl NodeType {
 pub struct Node {
     pub(super) root_hash: OnceLock<TrieHash>,
     encoded: OnceLock<Vec<u8>>,
+    is_encoded_longer_than_hash_len: OnceLock<bool>,
```
rkuris, line 160: We took this out because the `OnceLock` was really slow. If we need to add it back in, then we should create a structure that has both `Vec<u8>` and `bool` in it, and wrap that in the `OnceLock`. There is no point having two of them.

rkuris (changes requested):

firewood/src/merkle/node.rs
```diff
@@ -298,7 +313,6 @@ impl Node {
 struct Meta {
     root_hash: [u8; TRIE_HASH_LEN],
     attrs: NodeAttributes,
-    is_encoded_longer_than_hash_len: Option<bool>,
```
richardpringle, line 301: I see the issue here. This is supposed to be the `NodeTypeId`, can we add that here.

firewood/src/merkle/node.rs
```diff
@@ -360,24 +374,46 @@ impl Storable for Node {
...
+            };
+
+        let meta_raw = mem
+            .get_view(offset, 1_u64)
+            .ok_or(ShaleError::InvalidCacheView {
+                offset,
+                size: 1_u64,
+            })?;
+
+        offset += 1;
```
richardpringle, line 393: If you add the type-id to `Meta`, you don't need this

firewood/src/merkle/node.rs
```diff
@@ -360,24 +374,46 @@ impl Storable for Node {
...
+
         #[allow(clippy::indexing_slicing)]
-        match meta_raw.as_deref()[TRIE_HASH_LEN + 1].try_into()? {
+        match meta_raw.as_deref()[0].try_into()? {
```
richardpringle, line 396: If you add the `NodeTypeId` to `Meta`, you can leave this code as is. The idea here was actually to create the `Meta` struct first then use its properties... not sure why I didn't do that. Regardless, dealing with all the meta-data should be done in one atomic step. There should only be one read from `mem`, then `mem` should be passed to the node deserialization.

richardpringle (dismissed):

rkuris (approved):
Approved since we'll address my other comment at a later date

firewood/src/merkle/node.rs
rkuris: We will do this in a followup, if we find it matters.

PR #475, opened 2024-01-05
Make ObjRef Debug

rkuris:
Really helps with debugging

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -177,7 +177,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             .map(|node| node.as_ptr())
     }
 
-    pub fn get_store(&self) -> &dyn ShaleStore<Node> {
+    pub fn get_store(&self) -> &S {
         self.store.as_ref()
     }
 
--- a/firewood/src/shale/mod.rs
+++ b/firewood/src/shale/mod.rs
@@ -165,12 +165,13 @@ impl<T: Storable> Deref for Obj<T> {
 }
 
 /// User handle that offers read & write access to the stored [ShaleStore] item.
-pub struct ObjRef<'a, T: Storable> {
+#[derive(Debug)]
+pub struct ObjRef<'a, T: Storable + Debug> {
     inner: Option<Obj<T>>,
     cache: &'a ObjCache<T>,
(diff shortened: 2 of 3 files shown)


firewood/src/merkle.rs
```diff
@@ -177,7 +177,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             .map(|node| node.as_ptr())
     }
 
-    pub fn get_store(&self) -> &dyn ShaleStore<Node> {
+    pub fn get_store(&self) -> &S {
```
richardpringle, line 180: 🥇

firewood/src/shale/mod.rs
```diff
@@ -165,12 +165,13 @@ impl<T: Storable> Deref for Obj<T> {
 }
 
 /// User handle that offers read & write access to the stored [ShaleStore] item.
-pub struct ObjRef<'a, T: Storable> {
+#[derive(Debug)]
+pub struct ObjRef<'a, T: Storable + Debug> {
```
richardpringle, line 169: It would be much better if we could avoid further constraining types themselves (as opposed to implementations). 

If you need to debug the `Drop` implementation, I think it makes sense to just add this code in temporarily while you're debugging.

It's a lot more difficult to remove constraints on types once you rely on them in parent-types.

```suggestion
pub struct ObjRef<'a, T: Storable> {
```

firewood/src/shale/mod.rs
```diff
@@ -200,7 +201,7 @@ impl<'a, T: Storable> Deref for ObjRef<'a, T> {
     }
 }
 
-impl<'a, T: Storable> Drop for ObjRef<'a, T> {
+impl<'a, T: Storable + Debug> Drop for ObjRef<'a, T> {
```
richardpringle, line 204:
```suggestion
impl<'a, T: Storable> Drop for ObjRef<'a, T> {
```

richardpringle (commented):
Leaving this as a "comment" review because it is a strongly suggested change, not an absolutely necessary change.

firewood/src/shale/mod.rs
xinifinity: +1, makes sense

xinifinity (approved):

PR #476, opened 2024-01-05
add support for start key argument in fwdctl dump command

danlaine:
Adds a new optional `START_KEY` argument. The `START_KEY` is not expected to be hex. The string is turned into bytes with `.as_bytes()`. Tested manually and it works.

### DIFF
--- a/fwdctl/src/dump.rs
+++ b/fwdctl/src/dump.rs
@@ -21,6 +21,16 @@ pub struct Options {
         help = "Name of the database"
     )]
     pub db: String,
+
+    /// The key to start dumping from (if no key is provided, start from the beginning).
+    /// Defaults to None.
+    #[arg(
+        required = false,
+        value_name = "START_KEY",
+        default_value_t = String::from(""),
+        help = "Start dumping from this key (inclusive)."
+    )]
+    pub start_key: String,
 }
 
 pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
@@ -32,7 +42,12 @@ pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
     let db = Db::new(opts.db.clone(), &cfg.build()).await?;
     let latest_hash = db.root_hash().await?;
     let latest_rev = db.revision(latest_hash).await?;
-    let mut stream = latest_rev.stream();
+    let start_key_bytes = opts.start_key.as_bytes();
+    let start_key = match start_key_bytes.is_empty() {
+        true => Box::new([]),
+        false => start_key_bytes.to_vec().into_boxed_slice(),
+    };
+    let mut stream = latest_rev.stream_from(start_key);
     loop {
         match stream.next().await {
             None => break,
--- a/fwdctl/src/dump.rs
+++ b/fwdctl/src/dump.rs
@@ -9,7 +9,6 @@ use firewood::{
 use futures_util::StreamExt;
 use log;
 use std::borrow::Cow;
-use std::num::ParseIntError;
 
 #[derive(Debug, Args)]
 pub struct Options {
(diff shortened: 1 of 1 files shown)


fwdctl/src/dump.rs
```diff
@@ -21,6 +21,16 @@ pub struct Options {
...
+
+    /// The key to start dumping from (if no key is provided, start from the beginning).
+    /// Defaults to None.
+    #[arg(
+        required = false,
+        value_name = "START_KEY",
+        default_value_t = String::from(""),
+        help = "Start dumping from this key (inclusive)."
+    )]
+    pub start_key: String,
```
rkuris, line 33: Use `Option<Box<[u8]>>`. We should describe the input format more specifically -- this should be a set of hex bytes which we should parse and produce a binary key.

This is pretty hard to get the parsing right, but it's possible to set the type here and implement a custom parser for the start_key, something like this:

```
fn key_parser(s: &str) -> Result<Box<[u8]>, ParseIntError> {
    let mut result = Vec::with_capacity(s.len() / 2);
    let mut iter = s.chars().peekable();
    while iter.peek().is_some() {
        let chunk = iter.by_ref().take(2).collect::<String>();
        result.push(u8::from_str_radix(&chunk, 16)?);
    }
    Ok(result.into_boxed_slice())
}
```

Then, you just set the attribute `value_parser = key_parser` in the `arg` section above and it works.

This code was only tested against some happy paths, and I see two small problems with it, but it's probably fine for this. The problems are:
 - If the string is an odd length, it "works" kinda; we should enforce that each slice is 2 characters
 - The error sucks because it doesn't tell you what part of the string was wrong, just that there was something wrong

It's also not super performant but I think that doesn't matter for this case.

fwdctl/src/dump.rs
```diff
@@ -32,7 +42,12 @@ pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
     let db = Db::new(opts.db.clone(), &cfg.build()).await?;
     let latest_hash = db.root_hash().await?;
     let latest_rev = db.revision(latest_hash).await?;
-    let mut stream = latest_rev.stream();
+    let start_key_bytes = opts.start_key.as_bytes();
+    let start_key = match start_key_bytes.is_empty() {
```
rkuris, line 46: All this work will just happen in the args, you'll just have to do:
`let start_key = opts.start_key.clone().unwrap_or(Box::new([]));`

Or something like that. You could use the other iterator by mapping instead of starting with an empty slice, with a little more work:
```
    let iter = opts
        .start_key
        .as_ref()
        .map(|key| latest_rev.stream_from(key.clone()))
        .unwrap_or(latest_rev.stream());
```

fwdctl/src/dump.rs
danlaine: I think it makes more sense to have the start key string be converted into bytes the same way that we do in `fwdctl insert` so I did that. If you feel strongly we could do hex

fwdctl/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 
 [dependencies]
 firewood = { version = "0.0.4", path = "../firewood" }
-clap = { version = "4.4.18", features = ["cargo", "derive"] }
```
rkuris, line 8: Version updates should happen in a different PR.

rkuris (changes requested):
Code changes are fine, but lets not mix the upgrade of clap in with code changes.

fwdctl/Cargo.toml
danlaine: moved to https://github.com/ava-labs/firewood/pull/521

rkuris (approved):

richardpringle (approved):

richardpringle (approved):

PR #478, opened 2024-01-08
simplify `start_at_key_greater_than_all_others`

danlaine:
Rewrote this test for simplicity. After the rewrite, we simply insert `0x00` and then assert that if we start iterating from `0x01`, we should iterate zero times. Right now, this test fails because the stream implementation is incorrect.


richardpringle (changes requested):
I know this is draft, but you shouldn't delete this test, you should just add your test.

danlaine:
resolved by https://github.com/ava-labs/firewood/pull/498

PR #482, opened 2024-01-09
Only use an iterator instead of an index

richardpringle:
Up for discussion whether or not we want to merge this. It creates dead-code (the `else` case). While the code is dead in its current state, the advantage of this is that if someone changes the surrounding code, we hit the `else` case instead of using the wrong index.

### DIFF
--- a/firewood/src/merkle/proof.rs
+++ b/firewood/src/merkle/proof.rs
@@ -273,38 +273,43 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
             .get_node(sentinel)
             .map_err(|_| ProofError::NoSuchNode)?;
 
-        let mut key_nibbles = Nibbles::<0>::new(key.as_ref()).into_iter();
+        let mut key_nibbles = Nibbles::<1>::new(key.as_ref()).into_iter().peekable();
 
         let mut child_hash = root_hash;
         let proofs_map = &self.0;
-        let mut child_index = 0;
 
         let sub_proof = loop {
             // Link the child to the parent based on the node type.
             // if a child is already linked, use it instead
             let child_node = match &parent_node_ref.inner() {
                 #[allow(clippy::indexing_slicing)]
-                NodeType::Branch(n) => match n.chd()[child_index] {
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/proof.rs
```diff
@@ -273,38 +273,43 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
-                                    .inner_mut()
-                                    .as_branch_mut()
-                                    .expect("parent_node_ref is a branch");
-                                node.chd_mut()[child_index] = Some(child_node.as_ptr());
-                            })
-                            .map_err(|_| ProofError::WriteError)?;
-
-                        child_node
+                NodeType::Branch(n) => {
+                    let Some(child_index) = key_nibbles.next().map(usize::from) else {
```
rkuris, line 287: +1 for moving this case to the top, reducing the overall complexity

firewood/src/merkle/proof.rs
```diff
@@ -273,38 +273,43 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
+                            let child_node = decode_subproof(merkle, proofs_map, &child_hash)?;
+
+                            // insert the leaf to the empty slot
+                            parent_node_ref
+                                .write(|node| {
+                                    #[allow(clippy::indexing_slicing)]
+                                    let node = node
+                                        .inner_mut()
+                                        .as_branch_mut()
+                                        .expect("parent_node_ref is a branch");
```
rkuris, line 304: Is this backwards?

I think that if we expect parent_node_ref only ever contains branches, the code should self-document this with:
```suggestion
                                        .expect("parent_node_ref must only contains branches but it did not");
```

Otherwise, I think the error message is perhaps more misleading, because if you were to get this message,  it will just say "parent_node_ref is a branch" when it actually is not.

rkuris (approved):
Approved with strong suggestion for an error message change

firewood/src/merkle/proof.rs
richardpringle: It's the parent-node that must be a branch. I fixed the message.

Issue #481, opened 2024-01-09
Follow up nits for #460


richardpringle:
- [x] https://github.com/ava-labs/firewood/pull/460#discussion_r1445256795
        https://github.com/ava-labs/firewood/pull/483
  
  - [x] https://github.com/ava-labs/firewood/pull/460#discussion_r1445266411
      https://github.com/ava-labs/firewood/pull/482
      
  - [x] https://github.com/ava-labs/firewood/pull/460#discussion_r1445357153
      https://github.com/ava-labs/firewood/pull/484

  plus error handling
   - [x] https://github.com/ava-labs/firewood/pull/460#discussion_r1446214503
   - [x] https://github.com/ava-labs/firewood/pull/460#discussion_r1446241638
         https://github.com/ava-labs/firewood/pull/485
         
Can close once PRs are merged

PR #484, opened 2024-01-09
Use `HashKey` alias instead of plain array

richardpringle:
Another nit from #481


rkuris (approved):
Approved but fix the unused TRIE_HASH_LEN warning

danlaine (approved):

PR #487, opened 2024-01-09
Persist the encoded value of a Node if its length is less than the hash

xinifinity:
Close https://github.com/ava-labs/firewood/issues/473

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -261,6 +261,7 @@ impl Node {
 
     pub(super) fn rehash(&mut self) {
         self.encoded = OnceLock::new();
+        self.is_encoded_longer_than_hash_len = OnceLock::new();
         self.root_hash = OnceLock::new();
     }
 
@@ -282,6 +283,7 @@ impl Node {
 
     pub(super) fn new_from_hash(
         root_hash: Option<TrieHash>,
+        encoded: Option<Vec<u8>>,
         is_encoded_longer_than_hash_len: Option<bool>,
         inner: NodeType,
     ) -> Self {
@@ -290,7 +292,10 @@ impl Node {
                 Some(h) => OnceLock::from(h),
                 None => OnceLock::new(),
             },
-            encoded: OnceLock::new(),
+            encoded: match encoded {
+                Some(e) => OnceLock::from(e),
+                None => OnceLock::new(),
+            },
             is_encoded_longer_than_hash_len: match is_encoded_longer_than_hash_len {
                 Some(v) => OnceLock::from(v),
                 None => OnceLock::new(),
@@ -309,10 +314,14 @@ impl Node {
     }
 }
 
+const ENCODED_MAX_LEN: usize = TRIE_HASH_LEN;
+
 #[repr(C)]
 struct Meta {
     root_hash: [u8; TRIE_HASH_LEN],
     attrs: NodeAttributes,
+    encoded_len: [u8; size_of::<u64>()],
+    encoded: [u8; ENCODED_MAX_LEN],
     type_id: NodeTypeId,
 }
 
@@ -375,8 +384,27 @@ impl Storable for Node {
             None
         };
 
+        let mut start_index = TRIE_HASH_LEN + 1;
+        let end_index = start_index + size_of::<u64>();
         #[allow(clippy::indexing_slicing)]
-        let type_id: NodeTypeId = meta_raw.as_deref()[TRIE_HASH_LEN + 1].try_into()?;
+        let encoded_len = u64::from_le_bytes(
+            meta_raw.as_deref()[start_index..end_index]
+                .try_into()
+                .expect("invalid slice"),
+        );
+
+        start_index = end_index;
+        let mut encoded: Option<Vec<u8>> = None;
+        if encoded_len > 0 {
+            #[allow(clippy::indexing_slicing)]
+            let value: Vec<u8> =
+                meta_raw.as_deref()[start_index..start_index + encoded_len as usize].into();
+            encoded = Some(value);
+        }
+
+        start_index += ENCODED_MAX_LEN;
+        #[allow(clippy::indexing_slicing)]
+        let type_id: NodeTypeId = meta_raw.as_deref()[start_index].try_into()?;
 
         let is_encoded_longer_than_hash_len =
             if attrs.contains(NodeAttributes::IS_ENCODED_BIG_VALID) {
@@ -393,6 +421,7 @@ impl Storable for Node {
 
                 Ok(Self::new_from_hash(
                     root_hash,
+                    encoded,
                     is_encoded_longer_than_hash_len,
                     inner,
                 ))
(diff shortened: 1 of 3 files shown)


firewood/src/merkle/node.rs
```diff
@@ -375,8 +384,27 @@ impl Storable for Node {
...
         };
 
+        let mut start_index = TRIE_HASH_LEN + 1;
+        let end_index = start_index + size_of::<u64>();
+        #[allow(clippy::indexing_slicing)]
+        let encoded_len = u64::from_le_bytes(
+            meta_raw.as_deref()[start_index..end_index]
+                .try_into()
+                .expect("invalid slice"),
+        );
```
rkuris, line 394: Avoid indexing_slicing for new code. There are three ways around this that I can think of, possibly more:
```rust
        let encoded_len = u64::from_le_bytes(
            meta_raw.as_deref().get(start_index..end_index)
                .expect("reason why meta_raw must be at least end_length bytes goes here")
                .try_into()
                .expect("invalid slice"),
        );
```

Another fix would be to use an iterator, with a minus of a memory allocation here (you'd need to delete end_index as well):
```rust
        let encoded_len = u64::from_le_bytes(
            meta_raw
                .as_deref()
                .into_iter()
                .skip(start_index)
                .take(size_of::<u64>())
                .collect::<Vec<_>>()
                .try_into()
                .expect("why we know start_index and end_index are valid offsets goes here"),
        );
```
The best fix for this is to use the bytemuck crate. We're converting a `&[u8]` to a `u64` which is what bytemuck does efficiently and safely:

```rust
        let encoded_len = u64::from_le_bytes(
            *bytemuck::from_bytes::<[u8; size_of::<u64>()]>(meta_raw
                .as_deref()
                .get(start_index..end_index)
                .expect("why we know start_index and end_index are valid offsets goes here"),
                )
        );
```

rkuris (approved):
I am approving this only because the existing code has so many indexing_slicing cases and you're just following the pattern, but please consider refactoring this to avoid all the possible panics.

firewood/src/merkle/node.rs
richardpringle: >The best fix for this is to use the bytemuck crate.

I think this is true, but I think you want to use the [`try_cast`](https://docs.rs/bytemuck/latest/bytemuck/fn.try_cast.html) function with `Meta` and go straight from `meta_raw` to a `Meta` struct. You need to use the macros to implement the proper traits for the `Meta` struct. This was always my intention, but I just didn't get there.

firewood/src/merkle/node.rs
```diff
@@ -309,10 +314,14 @@ impl Node {
     }
 }
 
+const ENCODED_MAX_LEN: usize = TRIE_HASH_LEN;
+
 #[repr(C)]
 struct Meta {
     root_hash: [u8; TRIE_HASH_LEN],
     attrs: NodeAttributes,
+    encoded_len: [u8; size_of::<u64>()],
```
richardpringle, line 323: If you look at the individual node-types, whenever I'm encoding a length, I use a type alias so that if you make a change to the size of the thing that encodes the length, you'll get those updates everywhere. 
```rust
type EncodedLen = u64;
```

Then this line becomes
```rust
    encoded_len: EncodedLen,
```

firewood/src/merkle/node.rs
```diff
@@ -375,8 +384,27 @@ impl Storable for Node {
             None
         };
 
+        let mut start_index = TRIE_HASH_LEN + 1;
```
richardpringle, line 387: We should be using a cursor here instead (that's if you don't want to use `bytemuck`)

firewood/src/merkle/node.rs
```diff
@@ -375,8 +384,27 @@ impl Storable for Node {
...
+        let end_index = start_index + size_of::<u64>();
+        #[allow(clippy::indexing_slicing)]
+        let encoded_len = u64::from_le_bytes(
```
richardpringle, line 390: These are the dangerous parts. If you change the type of the length, you have to make sure to update everywhere. Please use the aforementioned alias.

firewood/src/merkle/node.rs
```diff
@@ -375,8 +384,27 @@ impl Storable for Node {
...
+            meta_raw.as_deref()[start_index..end_index]
+                .try_into()
+                .expect("invalid slice"),
```
richardpringle, line 393: A `Cursor` here with `read_exaxt` would make it so you wouldn't need to index. You would just create a buffer of the proper size and read into that. A good example can be seen in `branch.rs`.

firewood/src/merkle/node.rs
```diff
@@ -440,10 +471,14 @@ impl Storable for Node {
...
         };
 
+        let mut encoded_len: u64 = 0;
+        let mut encoded = None;
         if let Some(b) = self.encoded.get() {
             attrs.insert(if b.len() > TRIE_HASH_LEN {
                 NodeAttributes::LONG
             } else {
+                encoded_len = b.len() as u64;
+                encoded = Some(b);
```
richardpringle, line 481: This code is now really hard to read because too many things are happening inside the same branching logic. Try to flatten things out. 

```rust
let (encoded_len, encoded) = if let Some(b) = self.encoded.get() {
    let (encoded_len, endoded, attribute) = if b.len() > TRIE_HASH_LEN {
        (0, None, NodeAttributes::LONG)
    } else {
        (b.len() as u64, Some(b), NodeAttributes::IS_ENCODED_BIG_VALID)
    };
    
    attrs.insert(attribute);
    
    (encoded_len, encoded)
} else if Some(b) = self.is_encoded_longer_than_hash_len.get() {
    attrs.insert(if *b {
        NodeAttributes::LONG
    } else {
        NodeAttributes::IS_ENCODED_BIG_VALID
    });
    
    (0, None)
} else {
    (0, None)
};
```

There are other ways to write this that might be better, but the point is that you should be setting a value while you're inserting another value. It might be clear right now, but it will be more difficult to change.

firewood/src/merkle/node.rs
```diff
@@ -457,6 +492,15 @@ impl Storable for Node {
         #[allow(clippy::unwrap_used)]
         cur.write_all(&[attrs.bits()]).unwrap();
 
+        cur.write_all(&encoded_len.to_le_bytes())?;
+        if let Some(encoded) = encoded {
+            cur.write_all(encoded)?;
+            let remaining_len = ENCODED_MAX_LEN - encoded_len as usize;
+            cur.write_all(&vec![0; remaining_len])?;
```
richardpringle, line 499: If we have the encoded-length, why are we making this a constant size?

xinifinity: Because the limitation of shale, this size has to be known ahead.

firewood/src/merkle/node.rs
richardpringle: @xinifinity ^^ I edited what I said

firewood/src/merkle/node.rs
richardpringle: Oh, otherwise you have to write a new node when encoding?

PR #488, opened 2024-01-10
Clippy Warning fix with latest rust upgrade

xinifinity:
After upgrade rust version, the following shown as clippy warning.


rkuris (approved):
nit: paste the warning in the description. I didn't see these warnings with my version of rust.

PR #489, opened 2024-01-10
Logging in firewood, initial version

rkuris:
Needed logging for easy reporting of frequency of serialization and deserialization, so added some logging framework, behind a feature flag.

Also a tiny change to remove mutability of offset when deserializing, which arguably could have been done in a different commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,12 @@ There are several examples, in the examples directory, that simulate real world
 use-cases. Try running them via the command-line, via `cargo run --release
 --example simple`.
 
+## Logging
+
+If you want logging, enable the `logging` feature flag, and then set RUST\_LOG accordingly.
+See the documentation for [env\_logger](https://docs.rs/env_logger/latest/env_logger/) for specifics.
+We currently have very few logging statements, but this is useful for print-style debugging.
+
 ## Release
 
 See the [release documentation](./RELEASE.md) for detailed information on how to release Firewood.
--- a/firewood/src/logger.rs
+++ b/firewood/src/logger.rs
@@ -6,7 +6,7 @@
 // static shortcut
 
 #[cfg(feature = "logger")]
-pub use log::{trace, debug, info, warn, error};
+pub use log::{debug, error, info, trace, warn};
 
 #[cfg(not(feature = "logger"))]
 pub use noop_logger::{debug, error, info, trace, warn};
--- a/firewood/src/logger.rs
+++ b/firewood/src/logger.rs
@@ -1,0 +1,87 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+#[cfg(feature = "logger")]
+#[macro_export]
+macro_rules! trace {
+    // trace!(target: "my_target", key1 = 42, key2 = true; "a {} event", "log")
+    // trace!(target: "my_target", "a {} event", "log")
+    (target: $target:expr, $($arg:tt)+) => (log::log!(target: $target, log::Level::Trace, $($arg)+));
+
+    // trace!("a {} event", "log")
+    ($($arg:tt)+) => (log::log!(log::Level::Trace, $($arg)+))
+}
+#[cfg(not(feature = "logger"))]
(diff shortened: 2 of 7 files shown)


firewood/src/merkle/node.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
 use bincode::{Error, Options};
 use bitflags::bitflags;
 use enum_as_inner::EnumAsInner;
+use log::debug;
```
richardpringle, line 11: You should pull this out into an internal module where you do the conditional compilation. There, you can export your own debug macro that will just be a `() => {}` (not a no-op... but a no-code? haha) if the flag isn't set. 

That way you can get rid of the feature flag and won't forget about it. You can do the same thing with the other log-levels as well.

firewood/src/merkle/node.rs
```diff
@@ -427,6 +431,7 @@ impl Storable for Node {
     }
 
     fn serialize(&self, to: &mut [u8]) -> Result<(), ShaleError> {
+        debug!("[{self:p}] Serializing node");
```
richardpringle, line 434: Forgot a feature flag here. This shouldn't happen if you follow my comment from above

README.md
```diff
@@ -143,6 +143,12 @@ There are several examples, in the examples directory, that simulate real world
 use-cases. Try running them via the command-line, via `cargo run --release
 --example simple`.
 
+## Logging
+
+If you want logging, enable the `logging` feature flag, and then set RUST\_LOG accordingly.
+See the documentation for [env\_logger](https://docs.rs/env_logger/latest/env_logger/) for specifics.
```
richardpringle, line 149: You actually don't need the `\` before the `_` when the `_` is surrounded by letters.

richardpringle (commented):
So... I'm leaving this as a comment for now, but I strongly suggest these changes as they're a pretty significant QoL improvement every time you want to add a log.

firewood/src/shale/compact.rs
```diff
@@ -561,6 +564,8 @@ impl<T: Storable, M: CachedStore> CompactSpace<T, M> {
             }),
             obj_cache,
         };
+        #[cfg(logger)]
+        debug!("[{cs:p} New cache");
```
xinifinity, line 568: nit: "New compact space" instead?

firewood/src/shale/compact.rs
```diff
@@ -584,6 +589,9 @@ impl<T: Storable + Debug + 'static + PartialEq, M: CachedStore + Send + Sync> Sh
         #[allow(clippy::unwrap_used)]
         let addr = self.inner.write().unwrap().alloc(size)?;
 
+        #[cfg(feature = "logger")]
+        debug!("{self:p} New item at {addr} size {size}");
```
xinifinity, line 593: Nit: "Put item" instead of "New item".

README.md
rkuris: True, but it doesn't hurt anything either.

firewood/src/merkle/node.rs
rkuris: Fixed in latest

firewood/src/logger.rs
richardpringle, line 1: This might be a little more maintainable:
```rust

#[cfg(feature = "logger")]
pub use log::{debug, error, info, trace, warn};

#[cfg(not(feature = "logger"))]
#[macro_use]
mod noop_logger {
    #[macro_export]
    macro_rules! info {
        ($(target: $a:expr,)? $b:tt) => {};
    }
    #[macro_export]
    macro_rules! trace {
        ($(target: $a:expr,)? $b:tt) => {};
    }
    #[macro_export]
    macro_rules! debug {
        ($(target: $a:expr,)? $b:tt) => {};
    }
    #[macro_export]
    macro_rules! warn {
        ($(target: $a:expr,)? $b:tt) => {};
    }
    #[macro_export]
    macro_rules! error {
        ($(target: $a:expr,)? $b:tt) => {};
    }
}
```
The imports need to change slightly if you do this

richardpringle (approved):

PR #494, opened 2024-01-12
Use fast crc32 crate instead in growth-ring

xinifinity:
Closes https://github.com/ava-labs/firewood/issues/495.

While examine the flamegraph for inserts of large batch size I found the more number of batches, the more time it spent on crc32 checksum calculation in wal. e.g. for 5000 batches of insert of size 10 it took ~20% time. This PR updates to use a faster crc32 crate.

Before the change,
$ cargo run --release --quiet --example insert -- -n 5000 -b 10 -s 0
Generated and inserted 5000 batches of size 10 in 19.322914592s

flamegrah,
flamegraph_5000_old

After,
cargo run --release --quiet --example insert -- -n 5000 -b 10 -s 0
Generated and inserted 5000 batches of size 10 in 14.522793177s
flamegraph_5000

### DIFF
--- a/growth-ring/src/wal.rs
+++ b/growth-ring/src/wal.rs
@@ -3,6 +3,7 @@
 
 use async_trait::async_trait;
 use bytemuck::{cast_slice, Pod, Zeroable};
+use crc32fast::Hasher;
 use futures::{
     future::{self, FutureExt, TryFutureExt},
     stream::StreamExt,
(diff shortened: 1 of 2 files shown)


richardpringle (approved):
🚀

growth-ring/src/wal.rs
```diff
@@ -918,7 +927,11 @@ impl WalLoader {
     }
 
     fn verify_checksum_(data: &[u8], checksum: u32, p: &RecoverPolicy) -> Result<bool, WalError> {
-        if checksum == CRC32.checksum(data) {
+        let mut hasher = Hasher::new();
```
rkuris, line 930: nit: rather than do this 3 times, an inline function would perform the same and be less code.

rkuris (approved):

PR #499, opened 2024-01-14
Refactor stream implementation

danlaine:
This PR refactors the stream implementation. The iterator maintains a stack where each element represents an ongoing iteration of a branch node. In each iteration, the element at the top of the stack tells us which key-value pair to traverse next.

The special casing in the "setup" code that deals with branch/extension nodes which are the last element of `path_to_key` is kind of annoying and not ideal, but we can probably refactor this further if/when we implement a function that returns an iterator over all nodes on the path to a given key.

### DIFF
--- a/firewood/src/merkle/stream.rs
+++ b/firewood/src/merkle/stream.rs
@@ -11,6 +11,7 @@ use crate::{
 use futures::{stream::FusedStream, Stream};
 use std::{collections::VecDeque, task::Poll};
 type Key = Box<[u8]>;
+use crate::v2::api::Error::ChildNotFound;
 type Value = Vec<u8>;
 
 struct BranchIterator {
@@ -103,22 +104,6 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
 
                 let mut branch_iter_stack: Vec<BranchIterator> = vec![];
 
-                // if key.len() == 0 {
-                //     let root_branch = root_node.inner().as_branch().unwrap();
-                //     let child: DiskAddress = root_branch.children[0].unwrap();
-
-                //     let children_iter = std::iter::once((child, 0));
-
-                //     branch_iter_stack.push(BranchIterator {
-                //         key_nibbles: vec![],
-                //         children_iter: Box::new(children_iter),
-                //     });
-
-                //     self.key_state = IteratorState::Iterating { branch_iter_stack };
-
-                //     return self.poll_next(_cx);
-                // }
-
                 // (disk address, index) for each node we visit along the path to the node
                 // with [key], where [index] is the next nibble in the key.
                 let mut path_to_key = vec![];
@@ -166,7 +151,15 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
                                 // should be the child at [pos] or the child at [pos + 1].
 
                                 // Get the child at [pos], if any.
-                                let child = branch.children[pos as usize]
+                                let Some(child) = branch.children.get(pos as usize) else {
+                                    // This should never happen -- [pos] should never be OOB.
+                                    return Poll::Ready(Some(Err(api::Error::InternalError(
+                                        Box::new(ChildNotFound),
+                                    ))));
+                                };
+
+                                // Convert child from address --> node.
+                                let child = child
                                     .map(|child_addr| merkle.get_node(child_addr))
                                     .transpose()
                                     .map_err(|e| api::Error::InternalError(Box::new(e)))?;
--- a/firewood/src/merkle/stream.rs
+++ b/firewood/src/merkle/stream.rs
@@ -130,50 +130,51 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
                 while let Some((node, pos)) = path_to_key.pop_front() {
                     match node.inner() {
                         NodeType::Leaf(_) => (),
-                        NodeType::Branch(branch) => {
-                            // Figure out whether we want to start iterating over this
-                            // node's children at [pos] or [pos + 1].
-                            if !path_to_key.is_empty() {
-                                // The next element in [path_to_key] will handle the child at [pos]
-                                // so we can start iterating at [pos + 1].
-                                let children_iter = get_children_iter(branch)
-                                    .filter(move |(_, child_pos)| child_pos > &pos);
+                        NodeType::Branch(branch) if path_to_key.is_empty() => {
+                            // This is the last node in the path to [key].
+                            // Figure out whether to start iterating over this node's
+                            // children at [pos] or [pos + 1].
 
-                                branch_iter_stack.push(BranchIterator {
-                                    key_nibbles: matched_key_nibbles.clone(),
-                                    children_iter: Box::new(children_iter),
-                                });
+                            // Get the child at [pos], if any.
+                            let Some(child) = branch.children.get(pos as usize) else {
+                                // This should never happen -- [pos] should never be OOB.
+                                return Poll::Ready(Some(Err(api::Error::InternalError(
+                                    Box::new(api::Error::ChildNotFound),
+                                ))));
+                            };
+
+                            let child = child
+                                .map(|child_addr| merkle.get_node(child_addr))
+                                .transpose()
+                                .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+
+                            let comparer = if child.is_none() {
+                                // The child doesn't exist; we don't need to iterate over [pos].
+                                |a: &u8, b: &u8| a > b
                             } else {
-                                // Get the child at [pos], if any.
-                                let Some(child) = branch.children.get(pos as usize) else {
-                                    // This should never happen -- [pos] should never be OOB.
-                                    return Poll::Ready(Some(Err(api::Error::InternalError(
-                                        Box::new(api::Error::ChildNotFound),
-                                    ))));
-                                };
+                                // The child does exist; the first key to iterate over must be at [pos].
+                                |a: &u8, b: &u8| a >= b
+                            };
 
-                                let child = child
-                                    .map(|child_addr| merkle.get_node(child_addr))
-                                    .transpose()
-                                    .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+                            let children_iter = get_children_iter(branch)
+                                .filter(move |(_, child_pos)| comparer(child_pos, &pos));
 
-                                let comparer = if child.is_none() {
-                                    // The child doesn't exist; we don't need to iterate over [pos].
-                                    |a: &u8, b: &u8| a > b
-                                } else {
-                                    // The child does exist; the first key to iterate over must be at [pos].
-                                    |a: &u8, b: &u8| a >= b
-                                };
+                            branch_iter_stack.push(BranchIterator {
+                                key_nibbles: matched_key_nibbles.clone(),
+                                children_iter: Box::new(children_iter),
+                            });
+                            matched_key_nibbles.push(pos);
+                        }
+                        NodeType::Branch(branch) => {
+                            // The next element in [path_to_key] will handle the child at [pos]
+                            // so we can start iterating at [pos + 1].
+                            let children_iter = get_children_iter(branch)
+                                .filter(move |(_, child_pos)| child_pos > &pos);
 
-                                let children_iter = get_children_iter(branch)
-                                    .filter(move |(_, child_pos)| comparer(child_pos, &pos));
-
-                                branch_iter_stack.push(BranchIterator {
-                                    key_nibbles: matched_key_nibbles.clone(),
-                                    children_iter: Box::new(children_iter),
-                                });
-                            }
-
+                            branch_iter_stack.push(BranchIterator {
+                                key_nibbles: matched_key_nibbles.clone(),
+                                children_iter: Box::new(children_iter),
+                            });
                             matched_key_nibbles.push(pos);
                         }
                         NodeType::Extension(extension) => {
@@ -353,55 +354,6 @@ fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Key {
 
     data.into_boxed_slice()
 }
-
-// This code should be either used or removed.
-// TODO how can we use Either instead of `Box<dyn Trait>`?
-// mod helper_types {
-//     use std::ops::Not;
-
-//
-//     /// Enums enable stack-based dynamic-dispatch as opposed to heap-based `Box<dyn Trait>`.
-//     /// This helps us with match arms that return different types that implement the same trait.
-//     /// It's possible that [rust-lang/rust#63065](https://github.com/rust-lang/rust/issues/63065) will make this unnecessary.
-//     ///
-//     /// And this can be replaced by the `either` crate from crates.io if we ever need more functionality.
-//     pub(super) enum Either<T, U> {
-//         Left(T),
-//         Right(U),
-//     }
-
-//     impl<T, U> Iterator for Either<T, U>
-//     where
-//         T: Iterator,
-//         U: Iterator<Item = T::Item>,
-//     {
-//         type Item = T::Item;
-
-//         fn next(&mut self) -> Option<Self::Item> {
-//             match self {
-//                 Self::Left(left) => left.next(),
-//                 Self::Right(right) => right.next(),
-//             }
-//         }
-//     }
-
-//     #[must_use]
-//     pub(super) struct MustUse<T>(T);
-
-//     impl<T> From<T> for MustUse<T> {
-//         fn from(t: T) -> Self {
-//             Self(t)
-//         }
-//     }
-
-//     impl<T: Not> Not for MustUse<T> {
-//         type Output = T::Output;
-
-//         fn not(self) -> Self::Output {
-//             self.0.not()
-//         }
-//     }
-// }
 
 // CAUTION: only use with nibble iterators
 trait IntoBytes: Iterator<Item = u8> {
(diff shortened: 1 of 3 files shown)


firewood/src/merkle/stream.rs
```diff
@@ -88,202 +104,180 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                            branch_iter_stack.push(BranchIterator {
+                                key_nibbles: child_key_nibbles.clone(), // TODO reduce¸cloning
+                                children_iter: Box::new(children_iter),
+                            });
+
+                            // If there's a value, return it.
+                            if let Some(value) = branch.value.as_ref() {
+                                let value = value.to_vec();
+                                return Poll::Ready(Some(Ok((
+                                    key_from_nibble_iter(child_key_nibbles.into_iter().skip(1)), // skip the sentinel node leading 0
```
richardpringle, line 246: Can we just construct `child_key_nibbles` without the leading 0 in the first place?

firewood/src/merkle/stream.rs
```diff
@@ -470,6 +420,40 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn no_start_key() {
+        let mut merkle = create_test_merkle();
+        let root = merkle.init_root().unwrap();
+
+        for i in (0..256).rev() {
+            for j in (0..256).rev() {
+                let key = vec![i as u8, j as u8];
```
richardpringle, line 430:
```suggestion
        for i in 0..=u8::MAX {
            for j in 0..=u8::MAX {
                let key = vec![i, j];
```
Why do you care about insertion order?

firewood/src/merkle/stream.rs
```diff
@@ -470,6 +420,40 @@ mod tests {
...
+
+                merkle.insert(key, value, root).unwrap();
+            }
+        }
+
+        let mut stream = merkle.iter(root);
+
+        for i in 0..256 {
+            for j in 0..256 {
+                let expected_key = vec![i as u8, j as u8];
```
richardpringle, line 441:
```suggestion
        for i in 0..=u8::MAX {
            for j in 0..=u8::MAX {
                let expected_key = vec![i, j];
```

richardpringle (commented):
Pretty good! I like the pattern. I'm pretty sure we can use a concrete type for `children_iter` though.

firewood/src/merkle/stream.rs
danlaine: We'd have to special case so that we don't push the 0 index that points from the sentinel node to the root but it's certainly possible

firewood/src/merkle/stream.rs
danlaine: Added a comment -- this is just to ensure that the iterator doesn't somehow return key-value pairs in their insertion order, although I suppose this isn't really necessary.

firewood/src/merkle/stream.rs
```diff
@@ -1,28 +1,44 @@
...
-use std::task::Poll;
-
+use std::{collections::VecDeque, task::Poll};
 type Key = Box<[u8]>;
 type Value = Vec<u8>;
 
-enum IteratorState<'a> {
+struct BranchIterator {
+    // The nibbles of the key at this node.
+    key_nibbles: Vec<u8>,
```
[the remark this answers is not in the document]
danlaine: I tried making this a `Box<[u8]>` but ran into some compiler issues. Maybe you could help me out with that next time we chat?

firewood/src/merkle/stream.rs
```diff
@@ -88,202 +104,180 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                let mut path_to_key = vec![];
+
+                let node_at_key = merkle
+                    .get_node_by_key_with_callbacks(
+                        root_node,
+                        &key,
+                        |node_addr, i| path_to_key.push((node_addr, i)),
+                        |_, _| {},
+                    )
+                    .map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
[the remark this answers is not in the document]
danlaine: We don't declare `node_at_key` anymore so not sure if this is still relevant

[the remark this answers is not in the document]
richardpringle: My comment wouldn't compile anyway haha

firewood/src/merkle/stream.rs
```diff
@@ -88,197 +100,239 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                                    .transpose()
+                                    .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+
+                                let comparer = if child.is_none() {
+                                    // The child doesn't exist; we don't need to iterate over [pos].
+                                    |a: &u8, b: &u8| a > b
+                                } else {
+                                    // The child does exist; the first key to iterate over must be at [pos].
+                                    |a: &u8, b: &u8| a >= b
+                                };
```
richardpringle, line 166:
```suggestion
                                let comparer = if child.is_none() {
                                    <u8 as PartialEq>::gt
                                } else {
                                    <u8 as PartialEq>::ge
                                };
```
This probably compiles to the same thing, but you don't need a closure here; you can use a function pointer that's already defined.

firewood/src/merkle/stream.rs
```diff
@@ -88,197 +100,239 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                            }
+
+                            matched_key_nibbles.push(pos);
+                        }
+                        NodeType::Extension(extension) => {
+                            if !path_to_key.is_empty() {
+                                // Add the extension node's path to the key nibbles.
+                                matched_key_nibbles.extend(extension.path.iter());
+                                continue;
+                            }
```
richardpringle, line 184: Same comment as above. I think it would be easier to read the degenerate case as a separate branch 
```rust
NodeType::Extension(extension) if path_to_key.is_empty() => // ...
NodeType::Extension(extension) => // ...
```

firewood/src/merkle/stream.rs
```diff
@@ -88,197 +100,239 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                                    branch_iter_stack.push(BranchIterator {
+                                        key_nibbles,
+                                        children_iter: Box::new(iter),
+                                    });
+                                    break;
+                                };
+
+                                match next_extension_nibble.cmp(&next_key_nibble) {
+                                    std::cmp::Ordering::Equal => (), // The nibbles match; check the next one.
+                                    std::cmp::Ordering::Less => break, // [extension]'s child is before [key]. Skip it.
```
richardpringle, line 220: nit: put these comments above their respective match-arms.

firewood/src/merkle/stream.rs
```diff
@@ -88,197 +100,239 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
             }
+            IteratorState::Iterating { branch_iter_stack } => {
+                while let Some(mut branch_iter) = branch_iter_stack.pop() {
+                    // [node_addr] is the next node to visit.
+                    // It's the child at index [pos] of [node_iter].
+                    let Some((node_addr, pos)) = branch_iter.children_iter.next() else {
+                        // We visited all this node's descendants.
+                        // Go back to its parent.
+                        continue;
+                    };
```
richardpringle, line 261: 🤔 ... `branch_iter_stack` could actually be a `FlatMap` such that it actually just outputs the children directly. 

I think... in any case, this can be for another PR.

firewood/src/merkle/stream.rs
```diff
@@ -410,6 +423,8 @@ use super::tests::create_test_merkle;
 #[cfg(test)]
 #[allow(clippy::indexing_slicing, clippy::unwrap_used)]
 mod tests {
+    use std::vec;
```
richardpringle, line 426: I don't think you need this.

firewood/src/merkle/stream.rs
```diff
@@ -344,51 +354,54 @@ fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Key {
...
+//     }
+
+//     impl<T: Not> Not for MustUse<T> {
+//         type Output = T::Output;
+
+//         fn not(self) -> Self::Output {
+//             self.0.not()
+//         }
+//     }
+// }
```
richardpringle, line 404: You can delete

firewood/src/merkle/stream.rs
```diff
@@ -88,197 +100,239 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                    )
+                    .map_err(|e| api::Error::InternalError(Box::new(e)))?;
 
-                let found_key = nibble_iter_from_parents(&visited_node_path);
-                let found_key = key_from_nibble_iter(found_key);
+                // Convert (address, index) pairs to (node, index) pairs.
+                let mut path_to_key = path_to_key
+                    .into_iter()
+                    .map(|(node, pos)| merkle.get_node(node).map(|node| (node, pos)))
+                    .collect::<Result<VecDeque<_>, _>>()
```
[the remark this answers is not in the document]
danlaine: I used the first suggestion (except I use a bool for `is_last` instead of checking `i == 0`)

xinifinity:
> This PR refactors the stream implementation. The iterator maintains a stack where each element represents an ongoing iteration of a branch node. In each iteration, the element at the top of the stack tells us which key-value pair to traverse next.
> 
> The special casing in the "setup" code that deals with branch/extension nodes which are the last element of `path_to_key` is kind of annoying and not ideal, but we can probably refactor this further if/when we implement a function that returns an iterator over all nodes on the path to a given key.

Do you mind sharing what is the major motivation for the refactor? Is it for readability or for better performance?

danlaine:
> > This PR refactors the stream implementation. The iterator maintains a stack where each element represents an ongoing iteration of a branch node. In each iteration, the element at the top of the stack tells us which key-value pair to traverse next.
> > The special casing in the "setup" code that deals with branch/extension nodes which are the last element of `path_to_key` is kind of annoying and not ideal, but we can probably refactor this further if/when we implement a function that returns an iterator over all nodes on the path to a given key.
> 
> Do you mind sharing what is the major motivation for the refactor? Is it for readability or for better performance?

Just readability. I haven't tested the performance of this relative to the existing implementation.

richardpringle:
How hard do you think it would be to pop the starting logic into its own method?
```rust
impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'a, S, T> {
    type Item = Result<(Key, Value), api::Error>;

    fn poll_next(
        mut self: std::pin::Pin<&mut Self>,
        _cx: &mut std::task::Context<'_>,
    ) -> Poll<Option<Self::Item>> {
        // destructuring is necessary here because we need mutable access to `key_state`
        // at the same time as immutable access to `merkle`
        let Self {
            key_state,
            merkle_root,
            merkle,
        } = &mut *self;

        match key_state {
            IteratorState::StartAtKey(key) => {
                self.start_at(key)? // might look slightly different than this
                self.poll_next(_cx)
            }
            IteratorState::Iterating { branch_iter_stack } => {
                // ...
            }
        }
    }
}
```
This is definitely something that I should have done in the previous pass.

firewood/src/v2/api.rs
```diff
@@ -82,6 +82,15 @@ pub enum Error {
 
     #[error("Range too small")]
     RangeTooSmall,
+
+    #[error("Child not found")]
+    ChildNotFound,
+
+    #[error("Extension node is at the root; should be branch node")]
```
xinifinity, line 89: This is only referring to sentinel root? as his can happen for the real root, right? The sentinel root we force it to be branch node.

firewood/src/merkle/stream.rs
```diff
@@ -465,6 +444,84 @@ mod tests {
         check_stream_is_done(merkle.iter(root)).await;
     }
 
+    #[tokio::test]
+    async fn no_start_key() {
```
xinifinity, line 448: Is this somewhat overlapped with test at L407? If so, can we incorporate into one?

xinifinity (commented):
This PR will also work with the extension node removal PR, right?

firewood/src/v2/api.rs
danlaine: updated name

firewood/src/merkle/stream.rs
danlaine: Done

danlaine:
> This PR will also work with the extension node removal PR, right?

Not as it is right now. We will need to update the traversal logic in this PR to account for the fact that branch nodes may have partial paths in them.

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+/// Returns the state of an iterator over merkle, which has the given
+/// root_node, after it's been initialized. Iteration starts at the given key.
+fn get_iterator_intial_state<S: ShaleStore<Node> + Send + Sync, T>(
+    merkle: &Merkle<S, T>,
+    root_node: DiskAddress,
+    key: &[u8],
+) -> Result<IteratorState, api::Error> {
+    let root_node = merkle
+        .get_node(root_node)
+        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
rkuris, line 64: The error handling is poor here; if there is an IO error it gets mapped to an internal error. The right fix for this is to write `From` implementations for `MerkleError` and `ShaleError` for `v2::api::Error`.

This should be done in a follow up though.

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+            root_node,
+            key,
+            |node_addr, i| path_to_key.push((node_addr, i)),
+            |_, _| {},
+        )
+        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+
+    // Get each node in [path_to_key]. Mark the last node as being the last
+    // so we can use that information in the while loop below.
+    let path_to_key = path_to_key
```
rkuris, line 85: There is a much easier way than doing all this work up front, see next two comments.

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+        .into_iter()
+        .rev()
+        .enumerate()
+        .rev()
+        .map(|(i, (node, pos))| merkle.get_node(node).map(|node| (i == 0, (node, pos))));
+
+    let mut matched_key_nibbles = vec![];
+
+    for node in path_to_key {
```
rkuris, line 94:
```suggestion
        let mut path_iterator = path_to_key.into_iter().peekable();

    while let Some(node) = path_iterator.next() {
```

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+        let (is_last, (node, pos)) = match node {
+            Ok(result) => result,
+            Err(e) => return Err(api::Error::InternalError(Box::new(e))),
+        };
+
+        match node.inner() {
+            NodeType::Leaf(_) => (),
+            NodeType::Branch(branch) if is_last => {
```
rkuris, line 102:
```suggestion
            NodeType::Branch(branch) if path_iterator.peek().is_none() => {
```

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+                // This is the last node in the path to [key].
+                // Figure out whether to start iterating over this node's
+                // children at [pos] or [pos + 1].
```
rkuris, line 105: 🏆 level commenting

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+
+                // Get the child at [pos], if any.
+                let Some(child) = branch.children.get(pos as usize) else {
+                    // This should never happen -- [pos] should never be OOB.
+                    return Err(api::Error::InternalError(Box::new(
+                        api::Error::ChildNotFound,
+                    )));
+                };
```
rkuris, line 113: This is a valid case for a panic, as the code prevents this from ever happening.

See https://blog.burntsushi.net/unwrap/#so-one-should-never-use-unwrap-or-expect

We have a rule for prohibiting unwrap() but not expect() exactly for this case.
```suggestion
                // Get the child at [pos], if any.
                let child = branch.children.get(pos as usize).expect("pos cannot be OOB");
```

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+
+                let child = child
+                    .map(|child_addr| merkle.get_node(child_addr))
+                    .transpose()
+                    .map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
rkuris, line 118: same comment about errors as above. Should be fixed in a followup

firewood/src/merkle/stream.rs
```diff
@@ -1,47 +1,228 @@
...
+                let comparer = if child.is_none() {
+                    // The child doesn't exist; we don't need to iterate over [pos].
+                    <u8 as PartialOrd>::gt
+                } else {
+                    // The child does exist; the first key to iterate over must be at [pos].
+                    <u8 as PartialOrd>::ge
+                };
+
+                let children_iter = get_children_iter(branch)
+                    .filter(move |(_, child_pos)| comparer(child_pos, &pos));
```
rkuris, line 129: Comments significantly improve readability but in this case I'd just include the logic in the closure, which makes it even more readable:
```suggestion
                  let children_iter = get_children_iter(branch).filter(move |(_, child_pos)| {
                    if child.is_none() {
                        child_pos > &pos
                    } else {
                        child_pos >= &pos
                    }
                });
```

rkuris (changes requested):
Partial review, will complete the remainder from L130 on after standups

danlaine:
Marking as WIP. Going to try to implement this by implementing using a node iterator. See https://github.com/ava-labs/firewood/issues/399.

danlaine:
closing in favor of https://github.com/ava-labs/firewood/pull/517

PR #500, opened 2024-01-18
use bytemuck for metadata

richardpringle:
These code changes are from my suggestions [here](https://github.com/ava-labs/firewood/pull/487#event-11447833272) (on @xinifinity's PR). 

Using a stack-based struct allows for simplified serialization and deserialization and leaves less room for error compared to manually dealing with lengths. 

It would further be beneficial to refactor both `Extension` and `Branch` nodes such that all their metadata is stored in one chunk at the start of the serialized bytes. However, changing either type of branch is a little more involved and beyond the scope of this PR

- Use bytemuck for node-metadata
- Use bytemuck for leaf-node serialization

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -1,13 +1,14 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-use crate::logger::trace;
 use crate::{
+    logger::trace,
     merkle::from_nibbles,
     shale::{disk_address::DiskAddress, CachedStore, ShaleError, ShaleStore, Storable},
 };
 use bincode::{Error, Options};
 use bitflags::bitflags;
+use bytemuck::{CheckedBitPattern, NoUninit, Pod, Zeroable};
 use enum_as_inner::EnumAsInner;
 use serde::{
     de::DeserializeOwned,
@@ -213,6 +214,8 @@ impl From<NodeType> for Node {
 }
 
 bitflags! {
+    #[derive(Debug, Clone, Copy, Pod, Zeroable)]
+    #[repr(transparent)]
     struct NodeAttributes: u8 {
         const ROOT_HASH_VALID      = 0b001;
         const IS_ENCODED_BIG_VALID = 0b010;
@@ -315,14 +318,13 @@ impl Node {
     }
 }
 
-const ENCODED_MAX_LEN: usize = TRIE_HASH_LEN;
-
-#[repr(C)]
+#[derive(Clone, Copy, CheckedBitPattern, NoUninit)]
+#[repr(C, packed)]
 struct Meta {
     root_hash: [u8; TRIE_HASH_LEN],
     attrs: NodeAttributes,
-    encoded_len: [u8; size_of::<u64>()],
-    encoded: [u8; ENCODED_MAX_LEN],
+    encoded_len: u64,
+    encoded: [u8; TRIE_HASH_LEN],
     type_id: NodeTypeId,
 }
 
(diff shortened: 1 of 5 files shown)


firewood/src/merkle/node.rs
```diff
@@ -451,87 +444,83 @@ impl Storable for Node {
...
+        let root_hash = self.root_hash.get().map(|hash| hash.0);
+        let encoded = self.encoded.get();
+        let encoded_len = encoded.map(Vec::len).map(|len| len as u64);
+
+        let mut attrs = match (encoded_len, self.is_encoded_longer_than_hash_len.get()) {
+            // the raw encoded-node is longer than a hash
+            (None, Some(true)) => NodeAttributes::LONG,
+            (Some(len), _) if len > TRIE_HASH_LEN as u64 => NodeAttributes::LONG,
+            //
+            (Some(_), _) | (None, Some(false)) => NodeAttributes::IS_ENCODED_BIG_VALID,
```
richardpringle, line 466: @xinifinity, do you know what this is used for? I know the attribute is used to set `is_encoded_longer_than_hash_len` here, https://github.com/ava-labs/firewood/blob/bdb82d704b8a305023185b42221cc146ed6ebdf9/firewood/src/merkle/node.rs#L412-L419

But it seems like that's only checked in two places:
 - https://github.com/ava-labs/firewood/blob/bdb82d704b8a305023185b42221cc146ed6ebdf9/firewood/src/merkle/node/branch.rs#L153-L170
  - https://github.com/ava-labs/firewood/blob/bdb82d704b8a305023185b42221cc146ed6ebdf9/firewood/src/merkle/node/extension.rs#L55-L68
  
  But it seems we could just use the `Long` attribute in that case instead.

richardpringle: To clarify, it's apparent from this `match` statement that either the `Long` flag is set OR the `IS_ENCODED_BIG_VALID`, but never both. Can we just get rid of it?

rkuris (changes requested):
This code is too difficult to read; I rewrote the handling of `root_hash` as an example.

I started on a rewrite of the handling of `encoded` and `encoded_len`, but I'm not sure if this is worth working on right now. The std::array::from_fn needs work.

firewood/src/merkle/node.rs
```diff
@@ -451,87 +444,83 @@ impl Storable for Node {
...
+                    .map(Vec::as_slice)
+                    .unwrap_or_default()
+                    .iter()
+                    .copied();
+
+                move |_| iter.next().unwrap_or_default()
+            });
+
+            (encoded_len.unwrap_or_default(), encoded)
+        };
```
rkuris, line 490: I scratched my head on this one for a while; I think this is equivalent code but my confidence is not 100%:
```suggestion
        // have: encoded_len: Option<u64> and encoded: Option<&Vec<u8>>
        // If either of these are None, encoded is [0u8; 32]
        // encoded_len just unwrapped
        let (encoded_len, encoded) = match (encoded_len, encoded) {
            (Some(encoded_len), Some(encoded)) => (encoded_len, encoded.as_slice().try_into().unwrap_or([0; TRIE_HASH_LEN])),
            _ => (0, [0u8; TRIE_HASH_LEN]),
        };
```

xinifinity (approved):
Nice improvements!

I thought I pushed the review comment last Thursday, it turned out to be not...

firewood/src/merkle/node.rs
```diff
@@ -451,87 +444,83 @@ impl Storable for Node {
...
-                cur.write_all(&h.0)?;
-                NodeAttributes::ROOT_HASH_VALID
-            }
-            None => {
-                cur.write_all(&[0; 32])?;
-                NodeAttributes::empty()
-            }
+        let mut cursor = Cursor::new(to);
+
+        let root_hash = self.root_hash.get().map(|hash| hash.0);
```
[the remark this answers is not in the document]
xinifinity: I can understand this probably because I wrote the first version. To better the understanding, I do agree to separate the `root_hash` and `encoded` logic respectively and leave block comments may also help. Also, with the simplifying of NodeAttributes it may be easier to read as well.

richardpringle:
Waiting on 505 to land before opening this back up for review

PR #505, opened 2024-01-29
Simplify node-attribute flags for storage

richardpringle:
Based on some of the findings and feedback from #500, I think this simplifies storage

I changed up the way the flags work. We have ENCODED_IS_LONG and ENCODED_LENGTH_IS_KNOWN. These represent the three states of is_len_greater_than_hash_len.

Ignoring any other bits, we have:
```
00 -> None
01 -> Some(false)
10 -> Some(true)
11 -> Some(true)
```
Obviously, there's a redundant case, but we simply ignore the `ENCODED_LENGTH_IS_KNOWN` flag if `ENCODED_IS_LONG` is set.

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -216,7 +216,7 @@ bitflags! {
     struct NodeAttributes: u8 {
         const HAS_ROOT_HASH           = 0b001;
         const ENCODED_LENGTH_IS_KNOWN = 0b010;
-        const ENCODED_IS_LONG         = 0b100;
+        const ENCODED_IS_LONG         = 0b110;
     }
 }
 
@@ -409,13 +409,12 @@ impl Storable for Node {
         #[allow(clippy::indexing_slicing)]
         let type_id: NodeTypeId = meta_raw.as_deref()[start_index].try_into()?;
 
-        let is_encoded_longer_than_hash_len = if attrs.contains(NodeAttributes::ENCODED_IS_LONG) {
-            Some(true)
-        } else {
-            attrs
-                .contains(NodeAttributes::ENCODED_LENGTH_IS_KNOWN)
-                .then_some(false)
-        };
+        let is_encoded_longer_than_hash_len =
+            if attrs.contains(NodeAttributes::ENCODED_LENGTH_IS_KNOWN) {
+                attrs.contains(NodeAttributes::ENCODED_IS_LONG).into()
+            } else {
+                None
+            };
 
         match type_id {
             NodeTypeId::Branch => {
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/node.rs
```diff
@@ -315,7 +315,7 @@ impl Node {
     }
 }
 
-const ENCODED_MAX_LEN: usize = TRIE_HASH_LEN;
+const ENCODED_MAX_LEN: usize = TRIE_HASH_LEN - 1;
```
xinifinity, line 318: Hmm, `is_encoded_longer_than_hash_len` is a bad naming, do you mind changed it to `is_encoded_longer_than_or_equal_to_hash_len` or keep it as it is an update `node::is_encoded_longer_than_hash_len()` to really reflect the naming. Thanks!

firewood/src/merkle/node.rs
```diff
@@ -214,9 +214,9 @@ impl From<NodeType> for Node {
 
 bitflags! {
     struct NodeAttributes: u8 {
-        const ROOT_HASH_VALID      = 0b001;
-        const IS_ENCODED_BIG_VALID = 0b010;
-        const LONG                 = 0b100;
+        const HAS_ROOT_HASH           = 0b001;
+        const ENCODED_LENGTH_IS_KNOWN = 0b010;
+        const ENCODED_IS_LONG         = 0b100;
```
xinifinity, line 219: Why can't we remove `ENCODED_IS_LONG` as if not `ENCODED_LENGTH_IS_KNOWN` then it is longer than hash value.

xinifinity (approved):

firewood/src/merkle/node.rs
richardpringle: `ENCODED_LENGTH_IS_KNOWN` isn't the best name, but right now, there's a case where `Node` has a value for `is_encoded_longer_than_hash` but doesn't have a value for `encoded` or `root_hash`. I'm not sure if there's code that would lead to this being the case, but that context shouldn't be captured in here. So we definitely need three separate states, or `Node` itself should be changed.

firewood/src/merkle/node.rs
richardpringle: I created an issue
https://github.com/ava-labs/firewood/issues/508

firewood/src/merkle/node.rs
xinifinity: That should never happen as `is_encoded_longer_than_hash` is generated based on the comparison between `encoded` and `root_hash`. But I see your point. However, this makes this simplifying not so meaningful. The result code is a bit more confusing..

richardpringle: >The result code is a bit more confusing

Before, `IS_ENCODED_BIG_VALID` is ONLY set when we've *already* calculated the length of `encoded` and know that the length is **smaller** than that of a hash. I don't see how you could possibly make things more confusing than that haha.

Now, we have one flag for whether or not we know the length of the hash and another flag for whether the hash is long. I don't see how that's more confusing. 

However, I think the way that we're both looking at this is wrong. We are coupling node-logic and serialization-logic. If we manually create an invalid node with `is_encoded_length_longer_than_hash`, when the `encoded` value is actually shorter, the node should still serialize and deserialize to the same thing. 

I think the flags should remain as they currently are, but I will change the logic.

xinifinity (approved):

PR #514, opened 2024-02-05
Update tokio

richardpringle:
Looks like dependabot must not be configured properly


richardpringle:
@rkuris 

> Can we figure out why dependabot is not updating it instead? Are there other libraries we should be checking for updates on that it is also missing?

We should check all dependencies, dependabot is not running. 

I can't see anything wrong with the `dependabot.yml` file. You're the only one in this repo with settings perms to be able to check there (at least I don't know who else has perms). 

I'll close this PR, feel free to re-open if you change your mind.

PR #516, opened 2024-02-05
Pin minor versions, not bugfix versions

rkuris:
[Cargo dependency rules](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) say:

> The string "0.1.12" is a version requirement. Although it looks like a specific version of the time crate, it actually specifies a range of versions and allows SemVer compatible updates. An update is allowed if the new version number does not modify the left-most non-zero number in the major, minor, patch grouping. In this case, if we ran cargo update time, cargo should update us to version 0.1.13 if it is the latest 0.1.z release, but would not update us to 0.2.0. If instead we had specified the version string as 1.0, cargo should update to 1.1 if it is the latest 1.y release, but not 2.0. The version 0.0.x is not considered compatible with any other version.

### DIFF
--- a/firewood/Cargo.toml
+++ b/firewood/Cargo.toml
@@ -16,41 +16,43 @@ homepage = "https://avalabs.org"
 readme = "../README.md"
 
 [dependencies]
-aquamarine = "0.5.0"
-async-trait = "0.1.57"
-bytemuck = { version = "1.13.1", features = ["derive"] }
-enum-as-inner = "0.6.0"
+aquamarine = "0.5"
+async-trait = "0.1"
+bytemuck = { version = "1.13", features = ["derive"] }
+enum-as-inner = "0.6"
 growth-ring = { version = "0.0.4", path = "../growth-ring" }
 libaio = {version = "0.0.4", path = "../libaio" }
-futures = "0.3.24"
-hex = "0.4.3"
-lru = "0.12.0"
-metered = "0.9.0"
-nix = {version = "0.27.1", features = ["fs", "uio"]}
-parking_lot = "0.12.1"
+futures = "0.3"
+hex = "0.4"
+lru = "0.12"
+metered = "0.9"
+nix = {version = "0.27", features = ["fs", "uio"]}
+parking_lot = "0.12"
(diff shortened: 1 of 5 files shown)


xinifinity (commented):
Do you have an idea why dependabot didn't catch any of this? Thanks!

rkuris:
> Do you have an idea why dependabot didn't catch any of this? Thanks!

I think so, but I'm not sure. The debug logs from dependabot show this:
```
  updater | 2024/02/06 20:04:18 INFO Checking if tokio  needs updating
  updater | 2024/02/06 20:04:18 INFO Latest version is 1.36.0
  updater | 2024/02/06 20:04:22 INFO Requirements to unlock update_not_possible
  updater | 2024/02/06 20:04:22 INFO Requirements update strategy bump_versions_  if_necessary
```
I think the "update_not_possible" means that it is not currently supported to update multiple packages.

xinifinity:
> > Do you have an idea why dependabot didn't catch any of this? Thanks!
> 
> I think so, but I'm not sure. The debug logs from dependabot show this:
> 
> ```
>   updater | 2024/02/06 20:04:18 INFO Checking if tokio  needs updating
>   updater | 2024/02/06 20:04:18 INFO Latest version is 1.36.0
>   updater | 2024/02/06 20:04:22 INFO Requirements to unlock update_not_possible
>   updater | 2024/02/06 20:04:22 INFO Requirements update strategy bump_versions_  if_necessary
> ```
> 
> I think the "update_not_possible" means that it is not currently supported to update multiple packages.

Got it, I'm fine to update all dependencies once manually to fix this. However can we not removing the patch version? If we keep everything up to date, will dependabot be able to auto generate update PR moving forward?

richardpringle:
I would lock all pre-1.0 dependencies to a specific patch version as they do not need to adhere to any stability guarantees (API or otherwise). 

>4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

[source](https://semver.org/#spec-item-4)

rkuris:
Going in a different direction here

PR #517, opened 2024-02-06
Implement iterator over nodes; refactor key-value iterator

danlaine:
alternate implementation to #499.

This PR implements an iterator over nodes. The iterator can be configured to start at a given key. Uses this node iterator to refactor the key-value iterator implementation.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -25,8 +25,6 @@ pub use proof::{Proof, ProofError};
 pub use stream::MerkleKeyValueStream;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
-use self::stream::MerkleNodeStream;
-
 type NodeObjRef<'a> = shale::ObjRef<'a, Node>;
 type ParentRefs<'a> = Vec<(NodeObjRef<'a>, u8)>;
 type ParentAddresses = Vec<(DiskAddress, u8)>;
@@ -1727,20 +1725,6 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         key: Box<[u8]>,
     ) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::from_key(self, root, key)
-    }
-
-    #[allow(dead_code)]
-    pub(crate) fn node_iter(&self, root: DiskAddress) -> MerkleNodeStream<'_, S, T> {
-        MerkleNodeStream::new(self, root)
-    }
-
-    #[allow(dead_code)]
-    pub(crate) fn node_iter_from(
-        &self,
-        root: DiskAddress,
-        key: Box<[u8]>,
-    ) -> MerkleNodeStream<'_, S, T> {
-        MerkleNodeStream::from_key(self, root, key)
     }
 
     pub(super) async fn range_proof<K: api::KeyType + Send + Sync>(
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -22,7 +22,7 @@ pub use node::{
     NodeType, PartialPath,
 };
 pub use proof::{Proof, ProofError};
-pub use stream::MerkleKeyValueStream2;
+pub use stream::MerkleKeyValueStream;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
 type NodeObjRef<'a> = shale::ObjRef<'a, Node>;
@@ -1715,16 +1715,16 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
-    pub(crate) fn iter(&self, root: DiskAddress) -> MerkleKeyValueStream2<'_, S, T> {
-        MerkleKeyValueStream2::new(self, root)
+    pub(crate) fn iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
+        MerkleKeyValueStream::new(self, root)
     }
 
     pub(crate) fn iter_from(
         &self,
         root: DiskAddress,
         key: Box<[u8]>,
-    ) -> MerkleKeyValueStream2<'_, S, T> {
-        MerkleKeyValueStream2::from_key(self, root, key)
+    ) -> MerkleKeyValueStream<'_, S, T> {
+        MerkleKeyValueStream::from_key(self, root, key)
     }
 
     pub(super) async fn range_proof<K: api::KeyType + Send + Sync>(
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -25,6 +25,8 @@ pub use proof::{Proof, ProofError};
 pub use stream::MerkleKeyValueStream;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
+use self::stream::MerkleNodeStream;
+
 type NodeObjRef<'a> = shale::ObjRef<'a, Node>;
 type ParentRefs<'a> = Vec<(NodeObjRef<'a>, u8)>;
 type ParentAddresses = Vec<(DiskAddress, u8)>;
@@ -1715,16 +1717,30 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
-    pub(crate) fn iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
+    pub(crate) fn key_value_iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::new(self, root)
     }
 
-    pub(crate) fn iter_from(
+    pub(crate) fn key_value_iter_from(
         &self,
         root: DiskAddress,
         key: Box<[u8]>,
     ) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::from_key(self, root, key)
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn node_iter(&self, root: DiskAddress) -> MerkleNodeStream<'_, S, T> {
+        MerkleNodeStream::new(self, root)
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn node_iter_from(
+        &self,
+        root: DiskAddress,
+        key: Box<[u8]>,
+    ) -> MerkleNodeStream<'_, S, T> {
+        MerkleNodeStream::from_key(self, root, key)
     }
 
     pub(super) async fn range_proof<K: api::KeyType + Send + Sync>(
@@ -1750,8 +1766,8 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
         let mut stream = match first_key {
             // TODO: fix the call-site to force the caller to do the allocation
-            Some(key) => self.iter_from(root, key.as_ref().to_vec().into_boxed_slice()),
-            None => self.iter(root),
+            Some(key) => self.key_value_iter_from(root, key.as_ref().to_vec().into_boxed_slice()),
+            None => self.key_value_iter(root),
         };
 
         // fetch the first key from the stream
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1721,7 +1721,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         MerkleKeyValueStream::new(self, root)
     }
 
-    pub(crate) fn key_value_iter_from(
+    pub(crate) fn key_value_iter_from_key(
         &self,
         root: DiskAddress,
         key: Box<[u8]>,
@@ -1766,7 +1766,9 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
         let mut stream = match first_key {
             // TODO: fix the call-site to force the caller to do the allocation
-            Some(key) => self.key_value_iter_from(root, key.as_ref().to_vec().into_boxed_slice()),
+            Some(key) => {
+                self.key_value_iter_from_key(root, key.as_ref().to_vec().into_boxed_slice())
+            }
             None => self.key_value_iter(root),
         };
 
--- a/firewood/src/merkle/stream.rs
+++ b/firewood/src/merkle/stream.rs
@@ -261,7 +261,7 @@ fn get_iterator_intial_state<S: ShaleStore<Node> + Send + Sync, T>(
                     }
                 };
 
-                for next_partial_key_nibble in partial_key.iter().copied() {
+                for next_partial_key_nibble in partial_key.iter() {
                     let Some(next_key_nibble) = unmatched_key_nibbles.next() else {
                         // Ran out of [key] nibbles so [child] is after [key]
                         branch_iter_stack.push(BranchIterator::Unvisited {
@@ -277,7 +277,7 @@ fn get_iterator_intial_state<S: ShaleStore<Node> + Send + Sync, T>(
                         return Ok(MerkleNodeStreamState::Initialized { branch_iter_stack });
                     };
 
-                    if next_partial_key_nibble > next_key_nibble {
+                    if next_partial_key_nibble > &next_key_nibble {
                         // The leaf's key and the key diverged, and the
                         // leaf is greater, so we can stop here.
                         branch_iter_stack.push(BranchIterator::Unvisited {
@@ -290,7 +290,7 @@ fn get_iterator_intial_state<S: ShaleStore<Node> + Send + Sync, T>(
                                 .collect(),
                         });
                         return Ok(MerkleNodeStreamState::Initialized { branch_iter_stack });
-                    } else if next_partial_key_nibble < next_key_nibble {
+                    } else if next_partial_key_nibble < &next_key_nibble {
                         // [child] is before [key]
                         return Ok(MerkleNodeStreamState::Initialized { branch_iter_stack });
                     }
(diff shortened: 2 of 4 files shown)


firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+    // of each loop iteration.
+    let mut node_addr = root_node;
 
-                    self.key_state = IteratorState::Iterating {
-                        check_child_nibble,
-                        visited_node_path,
-                    };
+    // Invariant: [matched_key_nibbles] is the key of `node` at the start
+    // of each loop iteration.
+    let mut matched_key_nibbles = vec![];
```
richardpringle, line 200: Can you use `Vec::with_capacity` here to minimize reallocations?

richardpringle: 🤔... maybe that wouldn't be good

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+) -> Result<MerkleNodeStreamState<'a>, api::Error> {
+    // Invariant: `node`'s key is a prefix of `key`.
+    let mut node = merkle
+        .get_node(root_node)
+        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
 
-                    visited_node_path.push((found_node, 0));
+    // Invariant: This is `node`'s address at the start
+    // of each loop iteration.
+    let mut node_addr = root_node;
```
richardpringle, line 196: Can you use `node.as_ptr()` instead? Or is there a time when `node_addr` does not point to `node`?

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+                    NodeType::Branch(branch) => &branch.path,
+                    NodeType::Leaf(leaf) => &leaf.path,
+                    NodeType::Extension(_) => {
+                        panic!("extension nodes shouldn't exist")
+                    }
+                };
 
-                Poll::Ready(next)
+                let (comparison, new_unmatched_key_nibbles) =
+                    is_prefix(partial_key.iter(), unmatched_key_nibbles);
```
richardpringle, line 267: nit: 
`is_prefix` implies a `bool` as a return type. I say this as someone who frequently gives functions terrible names. What about `compare_as_prefix` or something along those lines?

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+                        // `child` is a prefix of `key`.
+                        matched_key_nibbles.extend(partial_key.iter().copied());
+                        node = child;
+                        node_addr = child_addr;
+                    }
+                    Ordering::Greater => {
+                        // `child` is after `key`.
+                        let child = merkle
+                            .get_node(child_addr)
+                            .map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
richardpringle, line 285: 🤔 , did `child` or `child_addr` change here? Don't you already have the `child`?

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+                        iter_stack.push(IterationNode::Unvisited {
+                            key: matched_key_nibbles
+                                .iter()
+                                .copied()
+                                .chain(partial_key.iter().copied())
+                                .collect(),
+                            node: child,
+                        });
```
richardpringle, line 293: nit: I think it's easier to read single-line property assigns
```suggestion
                        let key = matched_key_nibbles
                            .iter()
                            .copied()
                            .chain(partial_key.iter().copied())
                            .collect();
                        iter_stack.push(IterationNode::Unvisited {
                            key,
                            node: child,
                        });
```

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+                        iter_stack.push(IterationNode::Unvisited {
+                            key: matched_key_nibbles
+                                .iter()
+                                .copied()
+                                .chain(leaf.path.iter().copied())
+                                .collect(),
+                            node,
+                        });
+                    }
+                }
```
richardpringle, line 318: I'm surprised clippy didn't catch this, but it's just an if-statement

```suggestion
                if is_prefix(leaf.path.iter(), unmatched_key_nibbles).0 == Ordering::Greater {
                   let node = merkle
                        .get_node(node_addr)
                        .map_err(|e| api::Error::InternalError(Box::new(e)))?;

                    // The leaf's key > `key`, so we can stop here.
                    let key = matched_key_nibbles
                        .iter()
                        .copied()
                        .chain(leaf.path.iter().copied())
                        .collect();
                    
                    iter_stack.push(IterationNode::Unvisited {
                        key,
                        node,
                    });
                }
```

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
-        }
+            NodeType::Leaf(leaf) => {
+                let (comparison, _) = is_prefix(leaf.path.iter(), unmatched_key_nibbles);
+
+                match comparison {
+                    Ordering::Less | Ordering::Equal => {}
+                    Ordering::Greater => {
+                        let node = merkle
+                            .get_node(node_addr)
+                            .map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
richardpringle, line 307: 🤔, am I missing something or are you redundantly populating `node`?

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
-        visited_node_path: Vec<(NodeObjRef<'a>, u8)>,
+/// Represents an ongoing iteration over a node and its children.
+enum IterationNode<'a> {
+    // This node has not been returned yet.
+    Unvisited {
+        /// The key (as nibbles) of this node.
+        key: Key,
+        node: NodeObjRef<'a>,
+    },
+    // This node has been returned. Track which child to visit next.
```
richardpringle, line 25: I think you can also use triple `/` here too
```suggestion
    /// This node has not been returned yet.
    Unvisited {
        /// The key (as nibbles) of this node.
        key: Key,
        node: NodeObjRef<'a>,
    },
    /// This node has been returned. Track which child to visit next.
```

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+                                    panic!("extension nodes shouldn't exist")
+                                }
+                            };
+
+                            let child_key: Box<[u8]> = key
+                                .iter()
+                                .copied()
+                                .chain(once(pos))
+                                .chain(partial_path)
+                                .collect();
```
richardpringle, line 160: 🥇

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+
+                            // There may be more children of this node to visit.
+                            iter_stack.push(iter_node);
+
+                            let child = merkle
+                                .get_node(child_addr)
+                                .map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
richardpringle, line 167: 🤔 Already have the child again?

richardpringle (changes requested):
So, I really don't want to send you chasing any red herrings, but *I think* you can minimize the usage of `node_addr`/`child_addr` such that you aren't unnecessarily fetching nodes.

firewood/src/merkle/stream.rs
danlaine: removed

firewood/src/merkle/stream.rs
danlaine: renamed `compare_partial_path`

firewood/src/merkle/stream.rs
danlaine: removed

firewood/src/merkle/stream.rs
danlaine: removed

firewood/src/merkle/stream.rs
danlaine: removed `node_addr` -- not needed

firewood/src/merkle/stream.rs
danlaine: 🤷 Honestly you would know better than me lol

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+            merkle,
+        } = &mut *self;
 
-                if !found_next_node {
-                    return Ok(None);
-                }
+        match state {
+            MerkleKeyValueStreamState::Uninitialized(key) => {
+                // TODO how to not clone here?
+                let iter = MerkleNodeStream::from_key(merkle, *merkle_root, key.clone());
```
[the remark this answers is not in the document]
danlaine: I'll defer to you and the rest of the team on how to handle this. For the time being I removed the TODO since I don't think it's a blocker.

firewood/src/merkle/stream.rs
richardpringle: I was thinking that `with_capacity` would be better if you know the max size, but you're cloning the `matched_key_nibbles` as it grows, so you don't want a bunch of oversized vectors that will never grow. That would cause excess memory usage in an environment where memory is prized.

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+                // (i.e. if we should run this loop body again)
+                #[allow(clippy::indexing_slicing)]
+                let child_addr = match branch.children[nib as usize] {
+                    Some(c) => c,
+                    None => {
+                        // There is no child at `nib`.
+                        // We'll visit `node`'s first child at index > `nib`
+                        // first (if it exists).
+                        return Ok(MerkleNodeStreamState::Initialized { iter_stack });
+                    }
```
rkuris, line 244:
```suggestion
                let Some(child_addr) = branch.children[nib as usize] else {
                    // There is no child at `nib`.
                    // We'll visit `node`'s first child at index > `nib`
                    // first (if it exists).
                    return Ok(MerkleNodeStreamState::Initialized { iter_stack });
                };
```

rkuris (changes requested):

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,449 @@ use crate::{
...
+        };
+        // `nib` is the first nibble after [matched_key_nibbles].
+
+        match &node.inner {
+            NodeType::Branch(branch) => {
+                // The next nibble in `key` is `nib`,
+                // so all children of `node` with a position > `nib`
+                // should be visited since they are after `key`.
+                iter_stack.push(IterationNode::Visited {
+                    key: matched_key_nibbles.clone().into_boxed_slice(),
```
richardpringle, line 228: For a follow up PR:
I think it might make sense to do the following:
```suggestion
                    key: matched_key_nibbles.iter().copied().collect(),
```
I think you did this elsewhere in the PR too. The reason that I would suggest to do this instead of calling `clone` for turning a `Vec` into a `Box<[u8]>`, is that there are no guarantees that `Vec::clone` doesn't also create a new `Vec` with the same capacity, then turn it into a `Box<[u8]>`, dropping the extra capacity. 

In practice, that's not how the current version of `Vec::clone` works, but the only thing that's guaranteed is the API between Rust versions. 

I'm sure that you get the exact same output assembly, but I think it's generally better to be explicit about size (by using an iterator).

firewood/src/merkle/stream.rs
```diff
@@ -485,19 +489,179 @@ mod tests {
...
+        //        00 <-- branch with no value
+        //     0/  D|   \F
+        //    00   0D0   F <-- leaf with no partial path
+        //  0/ \F
+        //  1   F
+        //
+        // Note the 0000 branch has no value and the F0F0
+        // The number next to each branch is the position of the child in the branch's children array.
+
+        // Test cases for when the iterator has no start key
```
richardpringle, line 567: I strongly suggest you split these test cases up so that they are run as separate tests instead of checking different cases all in the same test. You can always have a separate function to create the same trie, but having a failing test case prevent a subsequent test does not lead to a great debugging experience.

richardpringle (approved):

firewood/src/merkle.rs
```diff
@@ -1715,18 +1717,32 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
...
-    pub(crate) fn iter_from(
+    pub(crate) fn key_value_iter_from(
         &self,
         root: DiskAddress,
         key: Box<[u8]>,
     ) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::from_key(self, root, key)
     }
 
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
danlaine: I thought it made sense to expose this new functionality even though it's not used right now, but I can remove it since it's not used right now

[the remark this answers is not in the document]
danlaine: Hmm if I remove `#[allow(dead_code)]` then it complains that the code is unused, but if I remove the function, I can't test the node iterator, which I should do because it's used by the key-value iterator

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
     fn with_key(key: Key) -> Self {
-        Self::StartAtKey(key)
+        Self::Uninitialized(key)
     }
 }
 
-/// A MerkleKeyValueStream iterates over keys/values for a merkle trie.
-pub struct MerkleKeyValueStream<'a, S, T> {
-    key_state: IteratorState<'a>,
+/// Iterates over the nodes in `merkle, whose root is `merkle_root,
```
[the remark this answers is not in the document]
danlaine: agreed, kind of redundant. Removed.

firewood/src/merkle/stream.rs
rkuris: Another good choice is: `key: Box::from(matched_key_nibbles.as_slice()),` since [Box::from can take a slice](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from-20)

firewood/src/merkle.rs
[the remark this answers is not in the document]
rkuris: Mark them #[cfg(test)] if they are only used in testing, or perhaps only have them defined in the test module.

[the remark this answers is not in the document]
danlaine: moved to test module and did a bit of renaming

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,471 @@ use crate::{
...
+    state: MerkleNodeStreamState<'a>,
     merkle_root: DiskAddress,
     merkle: &'a Merkle<S, T>,
 }
 
-impl<'a, S: ShaleStore<Node> + Send + Sync, T> FusedStream for MerkleKeyValueStream<'a, S, T> {
+impl<'a, S: ShaleStore<Node> + Send + Sync, T> FusedStream for MerkleNodeStream<'a, S, T> {
     fn is_terminated(&self) -> bool {
-        matches!(&self.key_state, IteratorState::Iterating { visited_node_path, .. } if visited_node_path.is_empty())
+        matches!(&self.state, MerkleNodeStreamState::Initialized { iter_stack } if iter_stack.is_empty())
```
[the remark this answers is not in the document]
danlaine: added

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,432 @@ use crate::{
...
+                        matched_key_nibbles.extend(partial_key.iter().copied());
+                        node = child;
+                    }
+                    Ordering::Greater => {
+                        // `child` is after `key`.
+                        let key = matched_key_nibbles
+                            .iter()
+                            .copied()
+                            .chain(partial_key.iter().copied())
+                            .collect();
```
rkuris, line 262: nit; doubtful it makes any difference but seems easier to digest
```suggestion
                        let key = matched_key_nibbles
                            .iter()
                            .chain(partial_key.iter())
                            .copied()
                            .collect();
```

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,432 @@ use crate::{
...
+                    }
+                }
             }
-        }
+            NodeType::Leaf(leaf) => {
+                if compare_partial_path(leaf.path.iter(), unmatched_key_nibbles).0
+                    == Ordering::Greater
+                {
+                    // `child` is after `key`.
+                    let key = matched_key_nibbles
```
rkuris, line 274: nit: since we're doing exactly the same thing earlier, perhaps a helper function would be useful here

firewood/src/merkle/stream.rs
```diff
@@ -7,399 +7,432 @@ use crate::{
...
-                    .iter()
-                    .copied()
-                    .chain(std::iter::once(*child_nibble)),
-            ),
-            NodeType::Extension(extension) => Either::Right(extension.path.iter().copied()),
-            NodeType::Leaf(leaf) => Either::Right(leaf.path.iter().copied()),
-        })
+/// Returns an iterator that returns (child_addr, pos) for each non-empty child of `branch`,
+/// where `pos` is the position of the child in `branch`'s children array.
+fn get_children_iter(branch: &BranchNode) -> impl Iterator<Item = (DiskAddress, u8)> {
```
rkuris, line 430: nit: this seems a little odd, in that we had to reverse the output of `enumerate()`. Seems like these tuples should be `(u8, DiskAddress)` to match the way `enumerate()` works.

nit: naming could be better. Typically `get` is not used in rust. Some ideas:
- `enumerated_children_iter`
- `as_enumerated_children_iter`

firewood/src/merkle/stream.rs
```diff
@@ -527,14 +706,75 @@ mod tests {
...
+                    (expected_key.into_boxed_slice(), expected_value),
+                    "i: {}, j: {}",
+                    i,
+                    j,
+                );
+            }
+        }
+        check_stream_is_done(stream).await;
+
+        // Test with start key
```
rkuris, line 749: nit: new test?

rkuris (approved):
Approved, all are remaining nits. There is one extra allocation you should probably remove before merging.

richardpringle (approved):

richardpringle (approved):

richardpringle:
🥇

PR #519, opened 2024-02-07
Add an upgrade check to CI

rkuris:
This PR adds a check to validate that nothing can be upgraded. Example output (tested with outdated version of tokio):
```
Run cargo upgrade --locked
    Updating 'https://github.com/rust-lang/crates.io-index' index
    Checking virtual workspace's dependencies
    Checking firewood's dependencies
name  old req compatible latest new req
====  ======= ========== ====== =======
    Checking fwdctl's dependencies
tokio 1.35.0  1.36.0     1.36.0 1.36.0 
    Checking growth-ring's dependencies
    Checking libaio's dependencies
    Checking rpc's dependencies
Error: cannot upgrade due to `--locked`
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,12 +5,13 @@
 
 version: 2
 updates:
-  - package-ecosystem: cargo
+  - package-ecosystem: "cargo"
+    versioning-strategy: "increase"
     directory: "/"
     schedule:
-      interval: daily
+      interval: "daily"
       time: "05:00"
-      timezone: America/Los_Angeles
+      timezone: "America/Los_Angeles"
     open-pull-requests-limit: 10
     allow:
-      - dependency-type: all
+      - dependency-type: "all"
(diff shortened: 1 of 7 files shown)


.github/dependabot.yml
```diff
@@ -5,12 +5,13 @@
 
 version: 2
 updates:
-  - package-ecosystem: cargo
+  - package-ecosystem: "cargo"
+    versioning-strategy: "increase"
```
xinifinity, line 9: 👍

xinifinity (approved):

richardpringle:
This will encourage upgrading dependencies *with* code changes instead of separating dependency updates as separate commits on `main`.

rkuris:
> This will encourage upgrading dependencies _with_ code changes instead of separating dependency updates as separate commits on `main`.

Unfortunately, unless dependabot is more dependable, I think this is the lesser of two evils. Most of the time dependabot will do the right thing. When it doesn't, we'll have to manually update things.

We can avoid simultaneous upgrades by code review. This happens already when we get a new rust release with a new lint, for example.

PR #529, opened 2024-02-15
Add proto file for service exposing views/proposals

danlaine:
This PR adds a `.proto` file which defines a service whose API allows a client to create and manipulate views/proposals on a merkledb


grpc-testtool/proto/merkle/merkle.proto
```diff
@@ -0,0 +1,102 @@
...
+  CommitError err = 1;
+}
+
+message NewViewRequest {
+  bytes root_id = 1;
+}
+
+message NewViewResponse {
+  // If not given, the parent view is the current database revision.
+  optional uint32 view_id = 1;
```
rkuris, line 45: I still want a `view_id` even if it's the current database revision at the time of this call.

This means I can continue to use the view even if I commit more stuff

grpc-testtool/proto/merkle/merkle.proto
```diff
@@ -0,0 +1,102 @@
...
+    rpc ViewGet(ViewGetRequest) returns (ViewGetResponse);
+    rpc ViewNewIteratorWithStartAndPrefix(ViewNewIteratorWithStartAndPrefixRequest) returns (ViewNewIteratorWithStartAndPrefixResponse);
+    rpc ViewRelease(ViewReleaseRequest) returns (google.protobuf.Empty);
+}
+
+message NewProposalRequest {
+  // If not given, the parent view is the current database revision.
+  optional uint32 parent_view_id = 1;
+  repeated PutRequest puts = 2;
+  repeated bytes deletes = 3;
```
rkuris, line 24: Should be a DeleteRequest, also from rpcdb.proto

rkuris (changes requested):

grpc-testtool/proto/merkle/merkle.proto
danlaine: This was a copy-paste error. Now it always returns a `view_id`.

grpc-testtool/proto/merkle/merkle.proto
danlaine: I'd be fine having this just be `repeated bytes` instead of `repeated DeleteRequest` where `DeleteRequest` just has one `bytes` field. If you feel strongly I can add that type though

rkuris: Ah, I see, that's just the key to delete. It's fine.

nit: comment that this is the key value

danlaine:
Converting to a draft since we will probably use the proto definitions in https://github.com/ava-labs/firewood/pull/533 for the first pass

rkuris (approved):

PR #530, opened 2024-02-15
Path iterator

danlaine:
Implements an iterator over the nodes on the path to a given key. Will make subsequent PRs to use this instead of `get_node_by_key_with_callbacks`.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1673,7 +1673,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             return Ok(Proof(proofs));
         }
 
-        let path_iterator = PathIterator::new(key.as_ref(), self, root);
+        let path_iterator = PathIterator::new(key.as_ref(), self, root)?;
 
         let nodes: Vec<DiskAddress> = path_iterator
             .map(|result| match result {
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1676,7 +1676,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             return Ok(Proof(proofs));
         }
 
-        let path_iterator = self.path_iter_by_key(root, key.as_ref())?;
+        let path_iterator = self.path_iter(root, key.as_ref())?;
 
         let nodes: Vec<DiskAddress> = path_iterator
             .map(|result| match result {
@@ -1714,7 +1714,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
-    pub fn path_iter_by_key<'b>(
+    pub fn path_iter<'b>(
         &self,
         root: DiskAddress,
         key: &'b [u8],
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -25,8 +25,6 @@ pub use proof::{Proof, ProofError};
 pub use stream::MerkleKeyValueStream;
 pub use trie_hash::{TrieHash, TRIE_HASH_LEN};
 
-use self::stream::MerkleNodeStream;
-
 type NodeObjRef<'a> = shale::ObjRef<'a, Node>;
 type ParentRefs<'a> = Vec<(NodeObjRef<'a>, u8)>;
 type ParentAddresses = Vec<(DiskAddress, u8)>;
@@ -1727,20 +1725,6 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         key: Box<[u8]>,
     ) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::from_key(self, root, key)
-    }
-
-    #[allow(dead_code)]
-    pub(crate) fn node_iter(&self, root: DiskAddress) -> MerkleNodeStream<'_, S, T> {
-        MerkleNodeStream::new(self, root)
-    }
-
-    #[allow(dead_code)]
-    pub(crate) fn node_iter_from(
-        &self,
-        root: DiskAddress,
-        key: Box<[u8]>,
-    ) -> MerkleNodeStream<'_, S, T> {
-        MerkleNodeStream::from_key(self, root, key)
     }
 
     pub(super) async fn range_proof<K: api::KeyType + Send + Sync>(
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1719,7 +1719,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         root: DiskAddress,
         key: &'b [u8],
     ) -> Result<PathIterator<'_, 'b, S, T>, MerkleError> {
-        PathIterator::new(key, self, root)
+        PathIterator::new(self, root, key)
     }
 
     pub(crate) fn key_value_iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1715,11 +1715,11 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
-    pub(crate) fn iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
+    pub(crate) fn key_value_iter(&self, root: DiskAddress) -> MerkleKeyValueStream<'_, S, T> {
         MerkleKeyValueStream::new(self, root)
     }
 
-    pub(crate) fn iter_from(
+    pub(crate) fn key_value_iter_from_key(
         &self,
         root: DiskAddress,
         key: Box<[u8]>,
@@ -1750,8 +1750,10 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
         let mut stream = match first_key {
             // TODO: fix the call-site to force the caller to do the allocation
-            Some(key) => self.iter_from(root, key.as_ref().to_vec().into_boxed_slice()),
-            None => self.iter(root),
+            Some(key) => {
+                self.key_value_iter_from_key(root, key.as_ref().to_vec().into_boxed_slice())
+            }
+            None => self.key_value_iter(root),
         };
 
         // fetch the first key from the stream
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1675,11 +1675,13 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
         let mut nodes = Vec::new();
 
-        let path_iterator = PathIterator::new(Box::from(key.as_ref()), self, root);
+        let path_iterator = PathIterator::new(key.as_ref(), self, root);
         for result in path_iterator {
             match result {
-                Ok((_, node)) => {
-                    nodes.push(node.as_ptr());
+                Ok((node_key, node)) => {
+                    if node_key.starts_with(key.as_ref()) {
+                        nodes.push(node.as_ptr());
+                    }
                 }
                 Err(e) => return Err(e),
             }
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1675,28 +1675,29 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
 
         let mut nodes = Vec::new();
 
-        let path_iterator = PathIterator::new(key.as_ref(), self, root);
-        for result in path_iterator {
-            match result {
-                Ok((node_key, node)) => {
-                    if node_key.starts_with(key.as_ref()) {
-                        nodes.push(node.as_ptr());
-                    }
-                }
-                Err(e) => return Err(e),
-            }
-        }
-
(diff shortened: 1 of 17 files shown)


firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+/// longest prefix of the key or the node that is where the key would be inserted if
+/// it were in the trie. Note this means the final returned node may not be a prefix
+/// of the key we're seeking.
+pub struct PathIterator<'a, 'b, S, T> {
+    state: PathIteratorState<'b>,
+    merkle: &'a Merkle<S, T>,
+}
+
+impl<'a, 'b, S: ShaleStore<Node> + Send + Sync, T> PathIterator<'a, 'b, S, T> {
+    pub(super) fn new(
```
richardpringle, line 425:
```suggestion
    pub(super) fn try_new(
```

firewood/src/merkle.rs
```diff
@@ -1671,23 +1673,17 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
...
+        let path_iterator = PathIterator::new(key.as_ref(), self, root)?;
 
-        if let Some(node) = node {
-            nodes.push(node.as_ptr());
-        }
+        let nodes: Vec<DiskAddress> = path_iterator
+            .map(|result| match result {
+                Ok((_, node)) => Ok(node.as_ptr()),
+                Err(e) => Err(e),
+            })
```
richardpringle, line 1682:
```suggestion
            .map(|result| result.map(|(_, node)| node.as_ptr()))
```

firewood/src/merkle.rs
```diff
@@ -1671,23 +1673,17 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
...
-
-        let mut nodes = Vec::new();
-
-        let node = self.get_node_by_key_with_callbacks(
-            root_node,
-            key,
-            |node, _| nodes.push(node),
-            |_, _| {},
-        )?;
+        let path_iterator = PathIterator::new(key.as_ref(), self, root)?;
```
richardpringle, line 1676: 🤔 I think the idiomatic thing to do here would be to expose a method called something like `fn node_iter_by_key(&self, root, key: Key)`. 

I think this would constrain the way the iterator is created. I think there was also talk of eventually attaching the `root` to `Merkle` so that you don't have to pass it in every time too.

I'm not sure if this really matters if the iterator is never really called outside of this module, but it is is, I think that would be the more convenient way to construct it.

richardpringle: And huge nit... but typically, I like to order args such that the most frequently changed come last. So, `merkle` -> `root` -> `key`... 

But as I said, huge nit... and I don't even remember to always follow this.

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
     }
 }
 
+enum PathIteratorState<'a> {
+    Iterating {
+        /// The key, as nibbles, of the node at [address], without the
```
richardpringle, line 401: I think you want this
```suggestion
        /// The key, as nibbles, of the node at `address`, without the
```
Or maybe this:
```suggestion
        /// The key, as nibbles, of the node at [Self::Iterating::address], without the
```
or maybe you need to specify the whole name
```suggestion
        /// The key, as nibbles, of the node at [PathIteratorState::Iterating::address], without the
```
I can't remember, you'll have to play with it.

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+        /// node's partial path (if any) at the end.
+        /// Invariant: If this node has a parent, the parent's key is a
+        /// prefix of the key we're seeking.
+        /// Note the node at [address] may not have a key which is a
+        /// prefix of the key we're seeking.
```
richardpringle, line 406: I would use the term "iterating over" instead of "seeking". 

The note is also kind of confusing. I think it would be better to describe the situation where the key is not a prefix of the key we're iterating over. When do we run into this case? I would expect the path-iterator to strictly return nodes that *are* on the path. 

Is it because the address is the address of the next node to check?

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+        address: DiskAddress,
+    },
+    Exhausted,
+}
+
+/// Iterates over all nodes on the path to a given key starting from the root.
+/// If the key isn't in the trie, the final returned node is either the node with the
+/// longest prefix of the key or the node that is where the key would be inserted if
+/// it were in the trie. Note this means the final returned node may not be a prefix
+/// of the key we're seeking.
```
richardpringle, line 418: I think I understand the point above now. If my understanding is correct, the only time the returned key *is not* a prefix is when it's the initial key the iterator was created with.

If that's the case, it would be better to say the key for each node could be one of two things, either the key we're iterating over or a prefix of that key. 

However, I might be missing something here...

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+    merkle: &'a Merkle<S, T>,
+}
+
+impl<'a, 'b, S: ShaleStore<Node> + Send + Sync, T> PathIterator<'a, 'b, S, T> {
+    pub(super) fn new(
+        key: &'b [u8],
+        merkle: &'a Merkle<S, T>,
+        merkle_root: DiskAddress,
+    ) -> Result<Self, MerkleError> {
+        let sentinel_node = merkle.get_node(merkle_root)?;
```
richardpringle, line 430: Instead of passing in the root, you could clean up the API by passing in the sentinel node. Then you can return `Self` instead of a `Result`. 

Doing this causes you to write a little more repeated code, but it's a lot more explicit. I would rather have the extra line that makes it explicit why there's a failure *before* calling the constructor rather than after. This pattern is especially useful in places where you can retry (which I don't think is applicable here). 

This isn't actually a Rust-specific pattern, but I've found Rust code bases (libraries or apps) a lot easier to maintain if you decouple fallible and infallible code as much as possible.

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+        match state {
+            PathIteratorState::Exhausted => None,
+            PathIteratorState::Iterating {
+                matched_key,
+                unmatched_key,
+                address,
+            } => {
+                let node = match merkle.get_node(*address) {
+                    Ok(node) => node,
+                    Err(e) => return Some(Err(e)),
```
richardpringle, line 473: Saves you ONE WHOLE CHARACTER!!! Haha, not sure if you knew you could do this, that's the only reason the suggestion is here, take it or leave it. 
```suggestion
                    e @ Err(_) => return Some(e),
```

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                    .collect();
+
+                let (comparison, unmatched_key) =
+                    compare_partial_path(partial_path.iter(), unmatched_key);
+
+                match comparison {
+                    Ordering::Less | Ordering::Greater => {
+                        self.state = PathIteratorState::Exhausted;
+                        Some(Ok((node_key, node)))
+                    }
```
richardpringle, line 495: 👍

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                                .copied()
+                                .collect();
+                            *address = child;
+
+                            Some(Ok((node_key, node)))
+                        }
+                        NodeType::Leaf(_) => {
+                            self.state = PathIteratorState::Exhausted;
+                            Some(Ok((node_key, node)))
+                        }
```
richardpringle, line 524: Nit, I would put this at the top. I tend to find code easier to read if the degenerate case is written first (when it's possible).

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                            let Some(child) = branch.children[next_unmatched_key_nibble as usize] else {
+                                // There's no child at the index of the next nibble in the key.
+                                // The node we're seeking isn't in the trie.
+                                self.state = PathIteratorState::Exhausted;
+                                return Some(Ok((node_key, node)));
+                            };
+
+                            *matched_key = node_key
+                                .iter()
+                                .chain(once(&next_unmatched_key_nibble))
```
richardpringle, line 514: Another FYI, not actually a suggested change:
```suggestion
                                .chain(Some(&next_unmatched_key_nibble))
```
Checkout the `once` implementation:
https://doc.rust-lang.org/stable/src/core/iter/sources/once.rs.html#55-57

I think I prefer using an `Option` if I don't have to call `into_iter` just so I don't have to import another function, but 🤷

firewood/src/merkle/stream.rs
```diff
@@ -396,6 +396,140 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleKeyValueStream<'
...
+                                .copied()
```
richardpringle, line 515: Another silly suggestion that saves you a single character:
```suggestion
                                .copied()
                                .chain(once(next_unmatched_key_nibble))
```

firewood/src/merkle/stream.rs
```diff
@@ -476,6 +610,121 @@ mod tests {
...
+    #[test_case(&[13]; "prefix of singleton key")]
+    #[test_case(&[13, 37]; "match singleton key")]
+    #[test_case(&[13, 37,1]; "suffix of singleton key")]
+    #[test_case(&[255]; "no key nibbles match singleton key")]
+    #[tokio::test]
+    async fn path_iterate_singleton_merkle(key: &[u8]) {
+        let mut merkle = create_test_merkle();
+        let root = merkle.init_root().unwrap();
+
+        merkle.insert(vec![0x13, 0x37], vec![0x42], root).unwrap();
```
richardpringle, line 633: Great number selection

firewood/src/merkle/stream.rs
```diff
@@ -476,6 +610,121 @@ mod tests {
...
+            key,
+            vec![0x00, 0x00, 0x00, 0x00, 0x00, 0x00].into_boxed_slice()
+        );
+        assert_eq!(
+            node.inner().as_branch().unwrap().value,
+            Some(vec![0x00, 0x00, 0x00].into()),
+        );
+
+        assert!(stream.next().is_none());
+    }
```
richardpringle, line 726: All these tests are 👍 👍

richardpringle (commented):
Haha some of these are nothing comments, just FYIs.

firewood/src/merkle.rs
```diff
@@ -1671,23 +1673,17 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
...
-        }
+        let nodes: Vec<DiskAddress> = path_iterator
+            .map(|result| match result {
+                Ok((_, node)) => Ok(node.as_ptr()),
+                Err(e) => Err(e),
+            })
+            .collect::<Result<Vec<DiskAddress>, MerkleError>>()?;
 
         // Get the hashes of the nodes.
-        for node in nodes.into_iter().skip(1) {
```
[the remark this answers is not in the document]
danlaine: The sentinel node is never returned by this iterator

firewood/src/merkle/stream.rs
danlaine: no longer returns a Result so I'm dismissing this

firewood/src/merkle/stream.rs
danlaine: updated the comment, lmk what you think

firewood/src/merkle.rs
```diff
@@ -1715,6 +1720,14 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
         self.store.flush_dirty()
     }
 
+    pub fn path_iter<'a, 'b>(
```
richardpringle, line 1723: haha much better than my name

firewood/src/merkle.rs
```diff
@@ -1715,6 +1720,14 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
...
+        &'a self,
+        sentinel_node: NodeObjRef<'a>,
```
richardpringle, line 1725: 🚀

firewood/src/merkle/stream.rs
richardpringle: To be honest, I find the `Note` part of the comment a little confusing, but it doesn't matter, I understand the code!

richardpringle (approved):

xinifinity (approved):

PR #533, opened 2024-02-16
Rkuris/implement merkle grpc

rkuris:


grpc-testtool/proto/merkle/merkle.proto
```diff
@@ -0,0 +1,101 @@
...
+  uint32 view_id = 1;
+  bytes key = 2;
+}
+
+message ViewGetResponse {
+  bytes value = 1;
+  GetError err = 2;
+}
+
+message ViewNewIteratorWithStartAndPrefixRequest {
```
danlaine, line 66: Should this and `ViewNewIteratorWithStartAndPrefixResponse` be here? We don't expose a way to use the iterator once you have its `iterator_id`.

rkuris: Discussed offline

PR #534, opened 2024-02-16
Use path iterator in `merkle.prove` and `merkle.get_node_by_key`

danlaine:
Removes 2 instances of usage of `get_node_by_key_with_callbacks`. Will remove other usages in subsequent PRs because they are more involved changes.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1673,7 +1673,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
             return Ok(Proof(proofs));
         }
 
-        let path_iterator = PathIterator::new(key.as_ref(), self, root);
+        let path_iterator = PathIterator::new(key.as_ref(), self, root)?;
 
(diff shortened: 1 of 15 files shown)


firewood/src/merkle.rs
```diff
@@ -1504,7 +1504,21 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
...
+            Some(Err(e)) => Err(e),
+            Some(Ok((node_key, node))) => {
+                let key_nibbles = Nibbles::<0>::new(key).into_iter();
+                if key_nibbles.eq(node_key.iter().copied()) {
+                    Ok(Some(node))
+                } else {
+                    Ok(None)
+                }
+            }
+        }
```
richardpringle, line 1521: FYI
```rust
path_iter.last().transpose().and_then(|(node_key, node)| {
    let key_nibbles = Nibbles::<0>::new(key).into_iter(); 
    key_nibbles.eq(node_key.iter().copied()).then(node)
})

PR #536, opened 2024-02-16
Take out the updates step for now

rkuris:
This is more hassle than it's worth and we need a better long term solution. I wish dependabot worked correctly.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,29 +102,6 @@ jobs:
       - name: Clippy
         run: cargo clippy --tests --examples --benches --all-features -- -D warnings
 
-  upgrades:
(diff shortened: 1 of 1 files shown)


danlaine:
Alternatively we could add `continue-on-error: true` to the `Cargo upgrade` step. I think that will make it so that CI won't be marked as failed even if this step fails.

danlaine (approved):

rkuris:
> Alternatively we could add `continue-on-error: true` to the `Cargo upgrade` step. I think that will make it so that CI won't be marked as failed even if this step fails.

Perhaps, but I think this check should be done somewhere else on a weekly or bi-weekly basis or something, not on each build.

PR #540, opened 2024-02-16
Name the diskbuffer thread

rkuris:
The name is helpful when (1) the stack overflows; (2) debugging; or (3) looking at process status output.

Also, we don't need block_in_place since all we're doing is creating a new thread to run the diskbuffer.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -520,12 +520,14 @@ impl Db {
         let disk_requester = DiskBufferRequester::new(sender);
         let buffer = cfg.buffer.clone();
         #[allow(clippy::unwrap_used)]
-        let disk_thread = block_in_place(|| {
-            Some(std::thread::spawn(move || {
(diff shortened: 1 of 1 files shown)


rkuris:
Output afterwards when running benchmarks for a long time:
```
     Running benches/insert.rs (/home/rk/open-source/firewood/target/release/deps/insert-7f7888487ea0cf5a)
Using process-server ../target/release/process-server
2024-02-17T00:02:05 [INFO] - Starting up: Listening on 5000
Benchmarking insert/1: Collecting 20 samples in estimated 30.060 s (18k iterations)
thread 'DiskBuffer' has overflowed its stack
fatal runtime error: stack overflow
thread 'main' panicked at grpc-testtool/benches/insert.rs:213:26:
batch insert succeeds: Status { code: Unknown, message: "transport error", source: Some(tonic::transport::Error(Transport, hyper::Error(Io, Custom { kind: BrokenPipe, error: "stream closed because of a broken pipe" }))) }
```

PR #544, opened 2024-02-20
Add the iterator to the public async API

rkuris:


firewood/src/db.rs
```diff
@@ -316,6 +319,19 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
             .await
             .map_err(|e| api::Error::InternalError(Box::new(e)))
     }
+
+    #[allow(refining_impl_trait)]
+    async fn iter<K: KeyType, V>(
+        &self,
+        first_key: Option<K>,
```
richardpringle, line 326: Because `KeyType` is generic, it might make this API awkward to call with `None` as you might have to specify what type of `None::<&[u8]>` even though the type has no bearing on the output. 

I would just make two methods `iter()` and `iter_from(key)`. That would also line up better with the pervasive convention of the `iter` signature not taking any arguments.

richardpringle (commented):
I know this is in draft, but came up in my notifications so thought I would leave a comment

firewood/src/db.rs
rkuris: This is a great idea

firewood/src/v2/api.rs
```diff
@@ -170,6 +171,41 @@ pub trait DbView {
...
+        &self,
+        first_key: Option<K>,
+    ) -> Result<impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>, Error>;
+
+    /// Obtain a stream over the keys/values of this view, starting from the beginning
+    fn iter(
+        &self,
+    ) -> impl Future<Output = Result<impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>, Error>>
+           + Send {
+        // impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>  {
```
xinifinity, line 196: nit: remove the commented out code

firewood/src/db/proposal.rs
```diff
@@ -299,4 +301,18 @@ impl api::DbView for Proposal {
     {
         todo!();
     }
+
+    #[allow(refining_impl_trait)]
```
xinifinity, line 305: Why allow this? instead of directly using `Result<MerkleKeyValueStream<CompactSpace<Node, storage::StoreRevMut>, Bincode>, api::Error>` in the trait?

firewood/src/v2/api.rs
```diff
@@ -170,6 +171,41 @@ pub trait DbView {
         last_key: Option<K>,
         limit: Option<usize>,
     ) -> Result<Option<RangeProof<Vec<u8>, Vec<u8>>>, Error>;
+
+    /// Obtain a stream over the keys/values of this view, using an optional starting point
+    ///
+    /// # Arguments
+    ///
+    /// * `first_key` - If None, start at the lowest key
```
xinifinity, line 179: I'm not sure why we need to provide the three APIs all together, why not just 1) one `iter_option` or 2)`iter` and `iter_from`. I would prefer the latter though.

firewood/src/db/proposal.rs
rkuris: Because not all the implementations use that exact signature. They all are iterators that return keys and values, but the three implementations return "refined" implementations:

A `DbRev<S>` returns `Result<MerkleKeyValueStream<S, Bincode>, api::Error>`
A `Proposal` return `Result<MerkleKeyValueStream<CompactSpace<Node, storage::StoreRevMut>, Bincode>, api::Error>`
`EmptyDb` returns `Result<EmptyStreamer, Error>`

If we made the trait specific, all of those implementations would be forced to use those concrete types.

Also, this allows us to keep those inner types a bit more hidden from the user. As a user of the API, I only care that it returns a tuple of keys and values, not that it implements a `KeyValueStream` on a `CompactSpace` for `Node`s. This might be important for us, but it's probably meaningless for an API user.

firewood/src/v2/api.rs
rkuris: Richard previously pointed out that it would be better to do the latter, but rather than have each implementation have two new methods, you can only add one (the iter_option) and both automatically take the default implementation from the trait.

firewood/src/v2/api.rs
```diff
@@ -170,6 +171,41 @@ pub trait DbView {
...
+        self.iter_option(Option::<Box<[u8]>>::None)
+    }
+
+    /// Obtain a stream over the key/values, starting at a specific key
+    fn iter_from<K: KeyType + 'static>(
+        &self,
+        first_key: K,
+    ) -> impl Future<Output = Result<impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>, Error>>
+           + Send {
+        // impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>  {
```
danlaine, line 206:
```suggestion
```

danlaine (commented):
Not sure if this PR is still a WIP or not. We should probably add unit tests for the implementations other than EmptyDb before merging

rkuris:
> Not sure if this PR is still a WIP or not. We should probably add unit tests for the implementations other than EmptyDb before merging

Discussed in standup. This isn't actually adding an iterator, it just validates that we get the keys to the iterator methods correctly, so emptydb is probably sufficient.

danlaine (approved):

xinifinity (approved):

firewood/src/v2/emptydb.rs
```diff
@@ -146,4 +167,22 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn empty_streamer_start_at() -> Result<(), Error> {
+        let emptydb = EmptyDb {};
+        let rev = emptydb.revision(ROOT_HASH).await?;
+        let mut iter = rev.iter_from(b"ignored").await?;
+        assert!(iter.next().await.is_none());
+        Ok(())
+    }
```
richardpringle, line 187: These are both testing a hard-coded value... I'm not sure it's worth having these tests here unless there's actual functionality to test.

firewood/src/v2/api.rs
```diff
@@ -170,6 +171,40 @@ pub trait DbView {
...
+    /// # Arguments
+    ///
+    /// * `first_key` - If None, start at the lowest key
+    ///
+    /// # Note
+    ///
+    /// If you always want to start at the beginning, [DbView::iter] is easier to use
+    /// If you always provide a key, [DbView::iter_from] is easier to use
+    ///
+    async fn iter_option<K: KeyType>(
```
richardpringle, line 186: Why is this async?

firewood/src/v2/api.rs
```diff
@@ -170,6 +171,40 @@ pub trait DbView {
...
+        &self,
+        first_key: Option<K>,
+    ) -> Result<impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>, Error>;
+
+    /// Obtain a stream over the keys/values of this view, starting from the beginning
+    fn iter(
+        &self,
+    ) -> impl Future<Output = Result<impl Stream<Item = Result<(Box<[u8]>, Vec<u8>), Error>>, Error>>
```
richardpringle, line 194: `Stream` should be an associated type, then you wouldn't have to use `#[allow(refining_impl_trait)]`

firewood/src/merkle/stream.rs
```diff
@@ -328,7 +328,7 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> FusedStream for MerkleKeyValueStr
 }
 
 impl<'a, S, T> MerkleKeyValueStream<'a, S, T> {
-    pub(super) fn new(merkle: &'a Merkle<S, T>, merkle_root: DiskAddress) -> Self {
+    pub(crate) fn new(merkle: &'a Merkle<S, T>, merkle_root: DiskAddress) -> Self {
```
richardpringle, line 331: Why did this change? This shouldn't be called outside of `Merkle`

firewood/src/v2/api.rs
rkuris: Good question, it doesn't need to be! Will address in a followup.

firewood/src/merkle/stream.rs
rkuris: Great catch, was doing some debugging when I changed this.

xinifinity (approved):

PR #548, opened 2024-02-21
Add `into_inner` method so we don't have to `clone`

richardpringle:
Add an `into_inner` method to `ObjRef<'_, Node>` and make it so `inner` can only be `None` inside the `shale` module. This allows us to efficiently use the `Node` values without cloning when we are going to drop anwyay. This is also useful for some of the iterator stuff as paths no longer need to be cloned when building keys.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1519,7 +1519,7 @@ impl<S: ShaleStore<Node> + Send + Sync, T> Merkle<S, T> {
     ) -> Result<NodeObjRef<'a>, MerkleError> {
         if let Err(ObjWriteSizeError) = write_result {
             let old_node_address = node.as_ptr();
-            node = self.put_node(node.clone())?;
+            node = self.put_node(node.into_inner())?;
             deleted.push(old_node_address);
 
             set_parent(node.as_ptr(), parents);
@@ -2323,7 +2323,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn single_key_proof_with_one_node() {
         let mut merkle = create_test_merkle();
         let root = merkle.init_root().unwrap();
@@ -2362,7 +2361,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn update_leaf_with_larger_path() -> Result<(), MerkleError> {
         let path = vec![0x00];
         let data = vec![0x00];
@@ -2382,7 +2380,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn update_leaf_with_larger_data() -> Result<(), MerkleError> {
         let path = vec![0x00];
         let data = vec![0x00];
@@ -2402,7 +2399,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn update_branch_with_larger_path() -> Result<(), MerkleError> {
         let path = vec![0x00];
         let data = vec![0x00];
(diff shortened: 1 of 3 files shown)


firewood/src/shale/mod.rs
```diff
@@ -207,16 +233,16 @@ impl<'a, T: Storable + Debug> Deref for ObjRef<'a, T> {
...
-        let mut inner = self.inner.take().unwrap();
-        let ptr = inner.as_ptr();
-        let mut cache = self.cache.lock();
-        match cache.pinned.remove(&ptr) {
-            Some(true) => {
-                inner.dirty = None;
-            }
-            _ => {
-                cache.cached.put(ptr, inner);
+        if let Some(mut inner) = self.inner.take() {
```
richardpringle, line 236: If `inner == None`, There's nothing to do in the cache anyway.

firewood/src/shale/mod.rs
```diff
@@ -171,12 +184,15 @@ impl<T: Storable> Deref for Obj<T> {
...
+    /// WARNING:
+    /// [`Self::inner`] should only be `None` when calling `into_inner` and `drop`.
     inner: Option<Obj<T>>,
     cache: &'a ObjCache<T>,
 }
 
 impl<'a, T: Storable + Debug> ObjRef<'a, T> {
-    const fn new(inner: Option<Obj<T>>, cache: &'a ObjCache<T>) -> Self {
+    const fn new(inner: Obj<T>, cache: &'a ObjCache<T>) -> Self {
+        let inner = Some(inner);
```
richardpringle, line 195: You should never be able to create an uninitialized `ObjRef`... what would it be "referencing" then? Even if we did want this behaviour, there was no way to initialize an uninitialized `ObjRef` which will just crash the program when dropped.

In its previous state, passing in `None` was guaranteed to crash the program.

firewood/src/shale/mod.rs
```diff
@@ -171,12 +184,15 @@ impl<T: Storable> Deref for Obj<T> {
 /// User handle that offers read & write access to the stored [ShaleStore] item.
 #[derive(Debug)]
 pub struct ObjRef<'a, T: Storable> {
+    /// WARNING:
+    /// [`Self::inner`] should only be `None` when calling `into_inner` and `drop`.
```
danlaine, line 188: Is this true? Don't we panic if we call `into_inner` and `inner` is `None`?

richardpringle: 🤔, I think I need to more specific. Should only be set to `None` inside `drop` or _after_ calling `into_inner`. Will update.

firewood/src/merkle.rs
```diff
@@ -2323,7 +2323,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
```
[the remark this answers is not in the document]
richardpringle: Code was approved but there was a suggestion to add tests. I added tests but couldn't get them to pass. The runtime would crash when attempting to drop the parent node (in the case of these tests, the sentinel). I verified that this behaviour "expected" on `main` prior to merging the ignored tests, using the `write_node!` macro (how we were updating nodes before I introduced a regression). 

To be perfectly honest, I don't quite understand why there was a panic while dropping the sentinel. I spent a couple of hours trying to figure it out to no avail. In any case dropping a node with `inner == None` should never be an issue since there's nothing to actually put in the cache. 

Put another way, I don't quite know what the `insert` code does differently that allows it to work on `main` while the tests fail.

rkuris (approved):
This is better than it was before, but please look at the comment I added for another idea.

firewood/src/shale/mod.rs
richardpringle: Updated

danlaine (approved):

PR #550, opened 2024-02-21
Use `Key` type instead of `Box<[u8]>`

danlaine:


fwdctl/src/dump.rs
```diff
@@ -1,15 +1,15 @@
...
-
 use clap::Args;
 use firewood::{
     db::{Db, DbConfig, WalConfig},
+    merkle::Key,
     v2::api::{self, Db as _},
 };
 use futures_util::StreamExt;
 use log;
+use std::borrow::Cow;
```
rkuris, line 12: nit: why was this suddenly needed?

rkuris (approved):

fwdctl/src/dump.rs
danlaine: It used to be imported above but it seems that the format-on-save moved this line for some reason

Issue #432, opened 2023-12-12
Remove `node.encode`


xinifinity:
We should use `Merkle.encode` / `Merkle.decode` instead of `Node.encode` / `Node.decode` after extension node removal.

rkuris:
Resolved with new design

PR #553, opened 2024-02-21
remove dead encode/decode code

danlaine:


xinifinity (commented):
Oops, https://github.com/ava-labs/firewood/issues/432 is actually the opposite :). We should remove the deacode annotation. And  use the encode/decode method from Merkle instead. https://github.com/ava-labs/firewood/pull/553/files#diff-6a6e300f3d9050248dbafa719531f14eabecc1bfac2169bb2527c81ee5474ecaL112

PR #555, opened 2024-02-23
Fix the demo example in `growth-ring`

xinifinity:
WIP not ready for review.

Currently, the demo example in `growth-ring` failed with `there is no reactor running, must be called from the context of a Tokio 1.x runtime` error.


richardpringle:
You just need to create a runtime then call `block_on` from the tokio-runtime. Easier to show you than explain 
https://github.com/ava-labs/firewood/commit/ff329451947ed2f594509a23d85a47cda84948bb

PR #564, opened 2024-02-26
rename MerkleSetup to InMemoryMerkle

danlaine:


firewood/src/db.rs
```diff
@@ -375,7 +375,8 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
         values: Vec<V>,
     ) -> Result<bool, ProofError> {
         let hash: [u8; 32] = *self.kv_root_hash()?;
-        let valid = proof.verify_range_proof(hash, first_key, last_key, keys, values)?;
+        let valid =
+            proof.verify_range_proof::<K, V, Bincode>(hash, first_key, last_key, keys, values)?;
```
richardpringle, line 379:
```suggestion
            proof.verify_range_proof::<_, _, Bincode>(hash, first_key, last_key, keys, values)?;
```
It's a best practice to let the compiler infer as much as possible to minimize the diff in case you decide to change anything.

xinifinity (approved):

firewood/src/merkle_util.rs
```diff
@@ -36,12 +36,53 @@ pub enum DataStoreError {
     ProofEmptyKeyValuesError,
 }
 
-pub struct MerkleSetup<S, T> {
+type InMemoryStore = CompactSpace<Node, DynamicMem>;
```
richardpringle, line 39: 👍

firewood/src/merkle_util.rs
```diff
@@ -36,12 +36,53 @@ pub enum DataStoreError {
...
     root: DiskAddress,
-    merkle: Merkle<S, T>,
+    merkle: Merkle<InMemoryStore, T>,
 }
 
-impl<S: ShaleStore<Node> + Send + Sync, T: BinarySerde> MerkleSetup<S, T> {
+impl<T> InMemoryMerkle<T>
+where
+    T: BinarySerde,
+    EncodedNode<T>: serde::Serialize + serde::de::DeserializeOwned,
```
richardpringle, line 49: I think this type is over-constrained. There shouldn't be any need to use `BinarySerde` or `EncodedNode<T>` methods in the constructor, should there?

firewood/src/merkle/proof.rs
```diff
@@ -146,14 +146,20 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
         last_key: K,
         keys: Vec<K>,
         vals: Vec<V>,
-    ) -> Result<bool, ProofError> {
+    ) -> Result<bool, ProofError>
+    where
+        K: AsRef<[u8]>,
+        V: AsRef<[u8]>,
+        T: BinarySerde,
+        EncodedNode<T>: serde::Serialize + serde::de::DeserializeOwned,
```
richardpringle, line 161: Are there any constraints here that can be removed? Or all they all needed? Am I just missing something with where `T` gets used?

danlaine: I think we need this for our call to `InMemoryMerkle::new` which needs it for the call to `Merkle::new`

firewood/src/merkle_util.rs
richardpringle: Ah you're right... `Merkle::new` is over-constrained... At some point, that should be cleaned up so that you don't need all these extra constraints everywhere, doesn't need to be this PR though

rkuris (approved):
Agree with Richard that we have these overconstrained, but that can be fixed in a followup.

PR #568, opened 2024-02-27
[WIP] remove `Encoded` type

danlaine:
Not ready for review yet

### DIFF
--- a/firewood/src/merkle/node.rs
+++ b/firewood/src/merkle/node.rs
@@ -11,7 +11,6 @@ use bitflags::bitflags;
 use bytemuck::{CheckedBitPattern, NoUninit, Pod, Zeroable};
 use enum_as_inner::EnumAsInner;
 use serde::{
-    de::DeserializeOwned,
     ser::{SerializeSeq, SerializeTuple},
     Deserialize, Serialize,
 };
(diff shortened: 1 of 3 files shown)


firewood/src/merkle/node.rs
```diff
@@ -728,42 +693,26 @@ impl<'de> Deserialize<'de> for EncodedNode<Bincode> {
...
 
             BranchNode::MSIZE => {
-                let path = items
-                    .pop()
-                    .unwrap_or_default()
-                    .deserialize::<Bincode>()
-                    .map_err(D::Error::custom)?;
+                // We checked the length above so unwrap is safe.
+                #[allow(clippy::unwrap_used)]
+                let path = items.pop().unwrap();
```
rkuris, line 698:
```suggestion
                let path = items.pop().expect("length was checked above");
```

rkuris (approved):
This code is still messy but it's soooo much better now!
🔢 🐌 

I would probably have done something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0b4b1c17ff55fddacbf9251ae90da5fe  or maybe just grab the first and second elements some other way.

xinifinity (commented):
I think this basically changed the result encoded value using Bincode encoding format. However, it may be fine because we are not keeping hash value compatibility with any now.

PR #569, opened 2024-02-27
Allow kv iterator to use any NodeSource (merkle, DbRev, Proposal)

rkuris:
Streaming just needs a source of nodes. Currently, the only node source
is a merkle. Future diffs will add a DbRev or a Proposal as a node
source, allowing these to be the backing storage for a
MerkleKeyValueStream.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -283,7 +283,7 @@ pub struct DbRev<S> {
 
 #[async_trait]
 impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
-    type Stream<'a> = MerkleKeyValueStream<'a, S, Bincode> where Self: 'a;
+    type Stream<'a> = MerkleKeyValueStream<'a, Merkle<S, Bincode>> where Self: 'a;
 
     async fn root_hash(&self) -> Result<api::HashKey, api::Error> {
         self.merkle
@@ -336,11 +336,14 @@ impl<S: ShaleStore<Node> + Send + Sync> api::DbView for DbRev<S> {
 }
 
 impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
-    pub fn stream(&self) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
+    pub fn stream(&self) -> merkle::MerkleKeyValueStream<'_, Merkle<S, Bincode>> {
         self.merkle.key_value_iter(self.header.kv_root)
     }
 
-    pub fn stream_from(&self, start_key: Key) -> merkle::MerkleKeyValueStream<'_, S, Bincode> {
+    pub fn stream_from(
+        &self,
+        start_key: Key,
+    ) -> merkle::MerkleKeyValueStream<'_, Merkle<S, Bincode>> {
         self.merkle
             .key_value_iter_from_key(self.header.kv_root, start_key)
     }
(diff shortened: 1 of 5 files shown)


firewood/src/db.rs
```diff
@@ -373,9 +376,9 @@ impl<S: ShaleStore<Node> + Send + Sync> DbRev<S> {
     }
 
     /// Verifies a range proof is valid for a set of keys.
-    pub fn verify_range_proof<N: AsRef<[u8]> + Send, K: AsRef<[u8]>, V: AsRef<[u8]>>(
+    pub fn verify_range_proof<NS: AsRef<[u8]> + Send, K: AsRef<[u8]>, V: AsRef<[u8]>>(
```
danlaine, line 379: Nit: I don't want to get too hung up on names, but why `NS`? Based on this PR, my brain is primed to think NS is short for "node source," which it doesn't appear to be in this case.

firewood/src/merkle/stream.rs
```diff
@@ -178,8 +180,8 @@ impl<'a, S: ShaleStore<Node> + Send + Sync, T> Stream for MerkleNodeStream<'a, S
 
 /// Returns the initial state for an iterator over the given `merkle` with root `root_node`
 /// which starts at `key`.
-fn get_iterator_intial_state<'a, S: ShaleStore<Node> + Send + Sync, T>(
-    merkle: &'a Merkle<S, T>,
+fn get_iterator_intial_state<'a, N: NodeSource>(
```
danlaine, line 183: Much cleaner! 👍

firewood/src/db.rs
rkuris: Good point; I just quickly resolved the name collision and it needs a better letter.

firewood/src/merkle.rs
```diff
@@ -66,6 +66,10 @@ macro_rules! write_node {
     };
 }
 
+pub trait NodeSource {
+    fn get_node(&self, ptr: DiskAddress) -> Result<NodeObjRef, MerkleError>;
+}
```
richardpringle, line 71: This seems a little odd to me (probably because there's only a single implementation right now). What, other than a `Merkle<S, T>` would ever implement this trait? 

Seems to me that a `DbRev` and a `Proposal` would always create a stream by talking to their underlying `Merkle<S, T>`?

I think it's a lot easier to find the right abstraction when you have at least one more implementation.

And I'm not saying this is wrong, I can only go on my intuition without seeing another use.

rkuris: The issue is that iterators need to have a reference back to something that holds nodes, since `get_node` returns a reference to a node with a lifetime of their parent. Iterators keep around either a `DbRev` or a `Proposal`, not a`merkle` per-se. In order for the iterator to live long enough between a RPC call, a reference to something that lives longer than that is needed.

I am open for other ideas on how to fix this, maybe we can talk about this during standup.

richardpringle: Yeah, happy to talk on a call. By no means do I think my comment should hold up the PR, just some food for thought. I don't see why a `Proposal` couldn't return a `MerkleKeyValueStream` that's not generic over a `Merkle<S, T>`, though. Maybe that means that the underlying `DbRev` has to expose that, but the actual `get_node` functionality actually always comes from the `Store`. From what I understand, the store will always be wrapped in some `Merkle<_, T>`, unless there's something else that's generic over the store that I'm just missing (very possible).

rkuris:
Decided against this

rkuris:
Decided against this

PR #572, opened 2024-03-04
Make the merkle tests idempotent

rkuris:
This should help with making the tests repeatable in case of failure.

Renamed functions and parameters to make this more readable as well.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -144,7 +144,7 @@ where
             }
         };
 
-        Bincode::serialize(&encoded).map_err(|e| MerkleError::BinarySerdeError(e.to_string()))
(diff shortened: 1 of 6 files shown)


xinifinity (commented):
I would like to have separate inputs for that if the purpose of this PR is only for easy debugging. e.g. the tests will be run in two modes, fixed or randomized. I agree we should have fuzz tests instead but removing this means less coverage. So we should remove it after the fuzz tests are in place.

rkuris:
> I would like to have separate inputs for that if the purpose of this PR is only for easy debugging. e.g. the tests will be run in two modes, fixed or randomized. I agree we should have fuzz tests instead but removing this means less coverage. So we should remove it after the fuzz tests are in place.

We talked about this in standup. The right answer is to have the seed randomized and report the seed when it fails. This is the best of both worlds.

danlaine (approved):

PR #570, opened 2024-02-29
Proposals are immutable

rkuris:
We shouldn't keep the StoreRevMut around in the Proposal, especially now since you can't iterate on a mutable revision.

PR #574, opened 2024-03-05
Make sure free space has a writeable CachedStore

rkuris:
The database header was backed by a StoreRevShared on new revisions, which meant that writes to the database header were not persisted. This happens because the merkle meta and payloads were backed by a `StoreRevShared`. The write method on `StoreRevShared` was a no-op.

This change prevents this problem from reappearing in several ways:
 - The `CachedStore` trait now has an `is_writeable` method to indicate whether writes should be allowed to this cache. This is only tested when debug mode is active, and ends out throwing the error earlier to make finding errant writes easier. Otherwise, this gets flushed when it goes out of scope, which can happen a lot later.
 - The `write` method of CachedStore now returns a `ShaleError`. These always return `Ok` except for the `StoreRevShared` implementation, which always returns an error.

Once these changes were made, it found the free space management bugs which are fixed in this change.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -406,7 +406,8 @@ impl DbRev<MutStore> {
 }
 
 impl From<DbRev<MutStore>> for DbRev<SharedStore> {
-    fn from(value: DbRev<MutStore>) -> Self {
+    fn from(mut value: DbRev<MutStore>) -> Self {
+        value.flush_dirty();
         DbRev {
             header: value.header,
             merkle: value.merkle.into(),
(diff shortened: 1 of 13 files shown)


firewood/src/db.rs
```diff
@@ -406,7 +406,8 @@ impl DbRev<MutStore> {
 }
 
 impl From<DbRev<MutStore>> for DbRev<SharedStore> {
-    fn from(value: DbRev<MutStore>) -> Self {
+    fn from(mut value: DbRev<MutStore>) -> Self {
+        value.flush_dirty();
```
xinifinity, line 410: why do w need to flush dirty here?

xinifinity (commented):
Which tests are failed because the free space management bugs?

firewood/src/db.rs
rkuris: Without this change, it would be possible to forget to flush the MutStore before consuming it and converting it to a SharedStore. This could result in deferring those writes to the underlying unwriteable shared store.

This only happens once the `DbRev` is finalized and ready for sharing, so at this point we're sure there won't be any more writes to the store. Seems like a great time to flush it.

danlaine (approved):

xinifinity:
> > Which tests are failed because the free space management bugs?
> 
> This is a bit hard to answer the way it's phrased. Let me attempt to be more specific see if that helps.
> 
> No tests were failing before any changes were made (obvioiusly, as the main branch runs and passes all the tests).
> 
> When debugging [my immutable proposal PR](https://github.com/ava-labs/firewood/pull/570) I found that writes weren't actually happening. It looked like they weren't happening because we were writing to a `StoreRevShared` `CachedStore`, so I added some changes to catch writes to the shared store.
> 
> Once I made the first set of changes to `CachedStore` against main (the part that propogates the error from write), many tests failed. Any test that created a new revision failed. They failed when dropping `ObjRef` objects, and so it was difficult to tell what the issue actually was. At that point, I added the `is_writeable` trait method and caught the updates when the call to `write` happened and the object was dirtied. This identified that the header writes weren't actually happening, and made it easy to see that the merkle was backed by a `StoreRevShared` when a new revision was being created, so I fixed that as part of this diff as well.
> 
> Since the header writes weren't happening, it would be possible to write a test that creates a corrupt database on disk. However, we currently have no way of detecting the corruption. We don't have any test that validate that there aren't any unallocated or overlapping regions of the database (a "fsck" filesystem type check tool). This is probably a good idea, but is definitely outside the scope of this diff. I created #577 as a followup.

Thanks a lot for the explanation! Did you get a chance to see how much performance degradation this cause?

xinifinity (approved):

rkuris:
> Thanks a lot for the explanation! Did you get a chance to see how much performance degradation this cause?

Yes. Nothing measurable. I did a code review with Dan this morning and he pointed out that this only happens during database creation, not at each commit.

PR #579, opened 2024-03-06
dont panic on error during disk buffer creation

danlaine:
Adds handling for errors from `DiskBuffer::new` instead of panicking. To do this, we add an enum variant `Aio` on `DbError`  and `AioError` on `ProofError` (followed existing naming conventions)


rkuris (approved):
This is fine but aio is likely to disappear at some point as (1) it's linux only -- no mac (2) it isn't very performant

xinifinity (approved):

PR #586, opened 2024-03-11
remove `ShaleStore` trait

danlaine:
Removes the `ShaleStore` trait and replaces its usage with `CompactSpace`, which was the only implementation of `ShaleStore`.


richardpringle (approved):
LGTM

xinifinity (approved):

PR #588, opened 2024-03-12
rename `PartialPath` to `Path`

danlaine:
This type is just a bunch of nibbles -- it can actually be a full path too. (And we will use it as such in subsequent encoding/decoding changes.)

This PR also renames `from_nibbles` to `nibbles_to_bytes_iter` which I think is more clear.

### DIFF
--- a/firewood/src/merkle/node/path.rs
+++ b/firewood/src/merkle/node/path.rs
@@ -63,12 +63,12 @@ impl Path {
     // TODO: remove all non `Nibbles` usages and delete this function.
     // I also think `Path` could probably borrow instead of own data.
     //
-    /// returns a tuple of the decoded partial path and whether the path is terminal
+    /// Returns the decoded partial path
     pub fn decode(raw: &[u8]) -> Self {
         Self::from_iter(raw.iter().copied())
     }
 
-    /// returns a tuple of the decoded partial path and whether the path is terminal
+    /// Returns the decoded partial path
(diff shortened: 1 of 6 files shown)


firewood/src/merkle/node/path.rs
```diff
@@ -60,14 +61,14 @@ impl PartialPath {
     }
 
     // TODO: remove all non `Nibbles` usages and delete this function.
-    // I also think `PartialPath` could probably borrow instead of own data.
+    // I also think `Path` could probably borrow instead of own data.
     //
-    /// returns a tuple of the decoded partial path and whether the path is terminal
+    /// Returns the decoded partial path
```
xinifinity, line 66: nit: is it still partial path?

firewood/src/merkle/node/path.rs
```diff
@@ -60,14 +61,14 @@ impl PartialPath {
...
     pub fn decode(raw: &[u8]) -> Self {
         Self::from_iter(raw.iter().copied())
     }
 
-    /// returns a tuple of the decoded partial path and whether the path is terminal
+    /// Returns the decoded partial path
```
xinifinity, line 71: Same here.

danlaine: fixed

firewood/src/merkle/node/path.rs
danlaine: fixed

xinifinity (approved):

PR #589, opened 2024-03-13
remove `EncodedNodeType`

danlaine:
Once again removing types I don't think we really need. Removes `EncodedNodeType` and `EncodedBranchNode` (which we kind of abused. It wasn't always a branch node.)

I think this code more cleanly aligns with merkledb now.


xinifinity (approved):

richardpringle:
This is nice

PR #590, opened 2024-03-14
remove `Data` type

danlaine:
Another type that I don't think needs to exist. One less thing for readers of this code to have to remember.


firewood/src/merkle/node/branch.rs
```diff
@@ -61,21 +61,7 @@ impl BranchNode {
     pub const MAX_CHILDREN: usize = MAX_CHILDREN;
     pub const MSIZE: usize = Self::MAX_CHILDREN + 2;
 
-    pub fn new(
```
danlaine, line 64: This function would've been `const` after this change so I just removed it. I think it's a bit silly to have a `new` function that just puts its parameters into fields. Harder to read that way imo.

xinifinity (approved):

PR #592, opened 2024-03-14
proof nits

danlaine:
* Rename `parent_node_ref` to `parent_node`. It's a `NodeObjRef`, and so is `child_node`, but the latter doesn't have `_ref` at the end. I think the suffix is unnecessary.
* Rename `proofs_map` to `proof_nodes_map`. This is more clear because the elements in the map are nodes, not proofs.
* Remove `decode_subproof`. This function was small, only used in one place, and not commented.
* Minor changes to a few comments

### DIFF
--- a/firewood/src/merkle/proof.rs
+++ b/firewood/src/merkle/proof.rs
@@ -308,6 +308,7 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
                         // If the child already resolved, then use the existing node.
                         Some(node) => merkle.get_node(node)?,
                         None => {
+                            // Look up the child's encoded bytes and decode to get the child.
                             let child_node_bytes = proof_nodes_map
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/proof.rs
```diff
@@ -308,10 +308,16 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
                         // If the child already resolved, then use the existing node.
                         Some(node) => merkle.get_node(node)?,
                         None => {
-                            let child_node = decode_subproof(merkle, proofs_map, &child_hash)?;
+                            let child_node_bytes = proof_nodes_map
```
xinifinity, line 311: nit: can you add a comment for the block to indicate this is to decode subproof to get the node?

richardpringle (approved):

PR #597, opened 2024-03-20
remove PayloadSize type

danlaine:
All else being equal, having fewer types is better because it makes the code easier to understand. I don't think we gain anything by having a type `PayloadSize`.


firewood/src/shale/compact.rs
```diff
@@ -73,11 +71,11 @@ impl Storable for CompactHeader {
 
 #[derive(Debug)]
 struct CompactFooter {
-    payload_size: PayLoadSize,
+    payload_size: u64,
 }
 
 impl CompactFooter {
-    const MSIZE: u64 = std::mem::size_of::<PayLoadSize>() as u64;
+    const MSIZE: u64 = 8; // u64.to_le_bytes() returns [u8; 8]
```
xinifinity, line 78: Why not do `std::mem::size_of::<u8>();` to avoid magic number here?

xinifinity: > it's the serialized length of it, which is actually determined by the length of u64.to_le_bytes() which is always 8. But then so is std::mem::size_of::<PayLoadSize>()

I'm a bit confused, if that is the case why `std::mem::size_of::<PayLoadSize>() as u64` would be different and work?

danlaine: I'm not sure I follow. What I'm saying is `8 == std::mem::size_of::<PayLoadSize>() == std::mem::size_of::<u64>() == u64.to_le_bytes().len()`

xinifinity: In that case, I would vote any approaches that can avoid magic numbers. I'm even fine with`u64.to_le_bytes().len()`, as this is serialized length as you suggested.

rkuris: Why not just:
```rust
    const MSIZE: u64 = (u64::BITS / u8::BITS) as u64;
```

You're looking for the number of bytes in a u64, but rust only provides `BITS`, so you can just divide to get it.

rkuris: Actually, on second thought, this is supposed to be the size of the whole `CompactFooter`, not the size of `payload_size`, so `#[repr(transparent)]` the CmopactFooter, then `std::mem::size_of::<CompactFooter>() as u64`, just like we do with all the other MSIZE constants.

rkuris (approved):
Approved after small fixes

firewood/src/shale/compact.rs
danlaine: > Actually, on second thought, this is supposed to be the size of the whole CompactFooter

It's supposed to be the _serialized length_ of the `CompactFooter`, not the _in-memory_ length. But these are the same in this case (`8`).

What does using `#[repr(transparent)]` buy us, exactly? I think in this case we don't actually care whether the layout of `CompactFooter` is the same as the layout of `u64`.

xinifinity (approved):

PR #599, opened 2024-03-20
fix/clarify MSIZE declarations

danlaine:
This PR should make it easier to understand the memory format of shale types.

* Removes magic numbers
* Defines constants for offsets
* Fixes places where we assume that a `usize` is 8 bytes


firewood/src/shale/compact.rs
```diff
@@ -24,7 +24,9 @@ pub struct CompactHeader {
 }
 
 impl CompactHeader {
-    pub const MSIZE: u64 = 17;
+    const IS_FREED_OFFSET: usize = 8;
```
rkuris, line 27: We can do sooo much better than this once offset_of is ready. It's slated for the next version of rust. On nightly, you can do:
```rust
    const IS_FREED_OFFSET: usize = core::mem::offset_of!(CompactHeader, is_freed);
```

I think it would be better to just wait for that to be on stable. It's "any day now".

Also, there is NO reason to have the is_freed bool before DiskAddress. This means unaligned accesses or padding, so swapping these two would be a good idea.

Since we are relying on the internal representation not being touched, we also need `#[repr(C)]` on `CompactHeader`. Otherwise, the compiler can (and in some cases will) reorder things or pack them differently on different architetures. Fortunately we're mostly on ix86 right now or this code may not even work.

firewood/src/shale/compact.rs
```diff
@@ -44,13 +46,19 @@ impl Storable for CompactHeader {
                 size: Self::MSIZE,
             })?;
         #[allow(clippy::indexing_slicing)]
-        let payload_size =
-            u64::from_le_bytes(raw.as_deref()[..8].try_into().expect("invalid slice"));
+        let payload_size = u64::from_le_bytes(
```
rkuris, line 49: We're doing all this work to specify a specific endianness and I think that's not necessary. I think we should be using the native endianness and just write the bytes out. This becomes something like:
```rust
        let payload_size = *bytemuck::cast_slice(&derefed[..Self::IS_FREED_OFFSET]).first().expect("wrong size");
```

You can probably do this with `bytemuck::cast` which would be more readable.

rkuris (approved):
Lots more work to do here, but this code is better than the old code, so I'm approving it. Please review the comments before landing and decide if you want to increase the scope or not.

xinifinity (approved):

PR #602, opened 2024-03-20
align path serialization with merkledb

danlaine:
Changes path serialization to work the same way as merkledb. Namely, write the length of the path _in nibbles_ and then write the path bytes out, where each byte is 2 nibbles (except the last byte, which has 1 nibble, in the event there are an odd number of nibbles.)


firewood/src/merkle/node.rs
```diff
@@ -402,6 +407,99 @@ impl<T> PartialEq for EncodedNode<T> {
...
+            return Err(serde::de::Error::custom(
+                "path length is not a multiple of 4",
+            ));
+        }
+
+        // Add BITS_PER_NIBBLE so that if there are an odd number of nibbles,
+        // we need to read the final byte.
+        // e.g. if `path_bits_len` is 12 we need to read 2 bytes
+        // If we didn't add BITS_PER_NIBBLE `path_bytes_len` would be 1.
+        let path_bytes_len = (path_bits_len + BITS_PER_NIBBLE) / BITS_PER_BYTE;
```
StephenButtolph, line 452: Is this read from an untrusted source? seems like this could overflow.

firewood/src/merkle/node.rs
```diff
@@ -402,6 +407,99 @@ impl<T> PartialEq for EncodedNode<T> {
...
+                .ok_or(serde::de::Error::custom("not enough bytes"))?;
+
+            // Convert the byte to 2 nibbles
+            path.push(byte >> 4);
+            path.push(byte & 0b_0000_1111);
+        }
+
+        if path_bits_len % BITS_PER_BYTE != 0 {
+            // The last byte only contained one nibble.
+            path.pop();
```
StephenButtolph, line 468: Should we be verifying that we are removing `0` here rather than an unexpected value?

danlaine: done

firewood/src/merkle/node.rs
danlaine: Added a max path length check to prevent malicious serialized paths from causing huge mallocs. Not the most elegant fix but we're planning on rewriting the node serialization/deserialization eventually so I'm OK with this as a stopgap.

firewood/src/merkle/node.rs
```diff
@@ -45,6 +46,19 @@ bitflags! {
     }
 }
 
+const NIBBLES_PER_BYTE: usize = 2;
+const BITS_PER_NIBBLE: u64 = 4;
+const BITS_PER_BYTE: u64 = 8;
```
richardpringle, line 51: Just an FYI, you can do math at compile time.
Also, why are the types different here? Might make sense to add a comment.

richardpringle, line 51: Why are the types different here? Might make sense to add a comment.

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
     }
 }
 
+#[derive(Debug)]
+struct PathWithBitsPrefix(Path);
```
richardpringle, line 420: Why not 
```rust
struct PathWithBitsPrefix<T>(T, Path);
```
?
(I just use `T`, because I don't know the actual size, could be a `u64` or a `[u8; N]`)

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
...
+        D: serde::Deserializer<'de>,
+    {
+        // TODO danlaine: This is slightly abusing the serde API.
+        // We don't know the number of bytes in the tuple until we've read the first u64.
+        // We should implement a custom serializer/deserializer for EncodedNode.
+        deserializer.deserialize_tuple(usize::MAX, TupleVisitor)
+    }
+}
+
+struct TupleVisitor;
```
richardpringle, line 434: It's actually common to define the `Visitor` directly in the `deserialize` implementation if it's not meant to be used anywhere else. Also, this name implies that the visitor is more generic than it actually is.

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
...
+
+impl<'de> Visitor<'de> for TupleVisitor {
+    type Value = PathWithBitsPrefix;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+        formatter.write_str("tuple")
```
richardpringle, line 440: Should add a TODO here, because you aren't just expecting a generic `"tuple"`.

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
...
+        // If we didn't add BITS_PER_NIBBLE `path_bytes_len` would be 1.
+        // Note `path_bytes_len` >= 1 because we assert `path_bits_len` > 0.
+        let path_bytes_len = (path_bits_len + BITS_PER_NIBBLE) / BITS_PER_BYTE;
+
+        let mut path = Vec::with_capacity(NIBBLES_PER_BYTE * path_bytes_len as usize);
+
+        for _ in 0..path_bytes_len {
+            let byte = seq
+                .next_element::<u8>()?
+                .ok_or(serde::de::Error::custom("not enough bytes"))?;
```
richardpringle, line 479: I think this error might not be very informative when you actually hit it. I could be wrong, but I'm pretty sure if you're deserializing a node and hit the error, you'll just get `"not enough bytes"`, not `"not enough bytes when attempting to deserialize a path"`.

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
...
+
+            // Convert the byte to 2 nibbles
+            path.push(byte >> 4);
+            path.push(byte & 0b_0000_1111);
```
richardpringle, line 483: I think `0x0f` is sufficient, no?

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
...
+        }
+
+        if path_bits_len % BITS_PER_BYTE != 0 {
+            // The last byte only contained one nibble.
+            let padding_nibble = path.pop().expect("path_bytes_len > 0");
+            if padding_nibble != 0 {
+                return Err(serde::de::Error::custom("padding nibble is not 0"));
+            }
+        }
```
richardpringle, line 492: Why not do this check outside the loop and determine the range based on the check?

firewood/src/merkle/node.rs
```diff
@@ -402,6 +416,113 @@ impl<T> PartialEq for EncodedNode<T> {
...
+}
+
+impl Serialize for PathWithBitsPrefix {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let path_length_bits: u64 = (self.0.len() as u64)
+            .checked_mul(4u64)
+            .expect("too many bits to serialize");
```
richardpringle, line 505: Why not propagate the error?

PR #605, opened 2024-03-21
[WIP] remove merkle.decode; update Proof type to have NodeType inside

danlaine:


firewood/src/merkle/proof.rs
```diff
@@ -274,13 +274,11 @@ impl<N: AsRef<[u8]> + Send> Proof<N> {
...
                             // Look up the child's encoded bytes and decode to get the child.
-                            let child_node_bytes = proof_nodes_map
+                            let child_node = proof_nodes_map
                                 .get(&child_hash)
                                 .ok_or(ProofError::ProofNodeMissing)?;
 
-                            let child_node = merkle.decode(child_node_bytes.as_ref())?;
-
-                            let child_node = merkle.put_node(Node::from(child_node))?;
+                            let child_node = merkle.put_node(Node::from(child_node.to_owned()))?; // todo avoid clone
```
xinifinity, line 281: As we are passing `NodeType` now, and we didn't actually assign the value to `chd_encode()` so this won't work. Particularly [this line](https://github.com/ava-labs/firewood/pull/605/files#diff-c89e2e8208c9c45b3d2ea693acb23b3fb6827335f509032d23ce214ce4763d7cL403). Also we should store the hashed value of the child instead at least for the `proof`.

PR #606, opened 2024-03-22
WIP

richardpringle:


firewood/src/merkle/node.rs
```diff
@@ -417,7 +417,14 @@ impl<T> PartialEq for EncodedNode<T> {
 }
 
 #[derive(Debug)]
-struct PathWithBitsPrefix(Path);
+struct PathWithBitsPrefix(u64, Path);
```
richardpringle, line 420: I actually don't think you need the extra field which makes the diff way smaller

firewood/src/merkle/node.rs
```diff
@@ -500,21 +510,18 @@ impl Serialize for PathWithBitsPrefix {
     where
         S: serde::Serializer,
     {
-        let path_length_bits: u64 = (self.0.len() as u64)
-            .checked_mul(4u64)
-            .expect("too many bits to serialize");
+        let path_length_bits = self.0;
```
richardpringle, line 513: Can set the length here instead of storing it

danlaine (approved):

PR #610, opened 2024-03-25
Rename `Compact` types

danlaine:
* Rename `CompactHeader` to `ChunkHeader`
* Rename `CompactSpace` to `Store`
* Rename `CompactFooter` to `ChunkFooter`
* Rename `CompactDescriptor` to `ChunkDescriptor`
* Rename `CompactSpaceHeader` to `StoreHeader`
* Rename `CompactSpaceInner` to `StoreInner`

I made these changes because:
1. I think "store" is more straightforward than "space" -- a store stores things. Which isn't to say "space" is wrong -- it would the phrase "linear space" makes sense, but I think is more complicated than necessary.
2. I don't think we need to stutter these names. This is the only store implementation. If we want we could alias the imports

Reviewers, please make sure these names are accurate


xinifinity (approved):
Approved, as I cannot think of a better name and it is improvements on existing one :)

PR #621, opened 2024-03-29
Remove Undo/Redo Persistence

xinifinity:
As we identified the syncer layer doesn't require the database (firewood) to replay older revisions from disk if they are not already in memory. This PR removes the logic to no longer persist the redo/undo from previous revisions for replay. Note that currently these redo/undo are not used for crash recovery and only used for replay older revisions.

As persisting redo/undo is one of the major bottleneck for fast writes, with this removal,  it provides at least 1x performance improvements. 

Before,
```
~/src/firewood-cp/firewood$ cargo run --release  --example insert --  -n 10000 -b 10 -s 0
     Running `target/release/examples/insert -n 10000 -b 10 -s 0`
Generated and inserted 10000 batches of size 10 in 48.60570351s
```

After,
```
~/src/firewood-cp/firewood$ cargo run --release  --example insert --  -n 10000 -b 10 -s 0
     Running `target/release/examples/insert -n 10000 -b 10 -s 0`
Generated and inserted 10000 batches of size 10 in 21.398385009s
```

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -10,12 +10,11 @@ use crate::{
     file,
     merkle::{
         Bincode, Key, Merkle, MerkleError, MerkleKeyValueStream, Proof, ProofError, TrieHash,
-        TRIE_HASH_LEN,
     },
     storage::{
         buffer::{DiskBuffer, DiskBufferRequester},
-        CachedStore, MemStoreR, StoreConfig, StoreDelta, StoreRevMut, StoreRevShared, StoreWrite,
-        ZeroStore, PAGE_SIZE_NBIT,
+        CachedStore, MemStoreR, StoreConfig, StoreDelta, StoreRevMut, StoreRevShared,
+        PAGE_SIZE_NBIT,
     },
     v2::api::{self, HashKey, KeyType, ValueType},
 };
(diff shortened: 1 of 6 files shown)


firewood/src/db.rs
```diff
@@ -240,11 +208,11 @@ struct Universe<T> {
 }
 
 impl Universe<StoreRevShared> {
-    fn to_mem_store_r(&self) -> Universe<Arc<impl MemStoreR>> {
-        Universe {
-            merkle: self.merkle.to_mem_store_r(),
-        }
-    }
+    // fn to_mem_store_r(&self) -> Universe<Arc<impl MemStoreR>> {
```
danlaine, line 211: I think this can be deleted

firewood/src/db.rs
```diff
@@ -870,85 +818,27 @@ impl Db {
...
-                    root_hash_store
-                        .get_view(0, TRIE_HASH_LEN as u64)
-                        .expect("get view failed")
-                        .as_deref()
-                })
-                .map(|data| TrieHash(data[..TRIE_HASH_LEN].try_into().unwrap()))
-                .position(|trie_hash| &trie_hash == root_hash));
-        }
-
+        let nback = revisions.root_hashes.iter().position(|r| r == root_hash);
```
danlaine, line 833: Nit: Can this line and the one below be combined by adding `?` to the end of this line?

firewood/tests/db.rs
```diff
@@ -146,16 +145,16 @@ async fn test_revisions() {
...
-        for (dump, hash) in dumped.iter().zip(hashes.iter().cloned()) {
-            let rev = db.revision(hash).await.unwrap();
-            rev.root_hash().await.unwrap();
-            assert_eq!(
-                *dump,
-                block_in_place(|| kv_dump!(rev)),
-                "not the same: pass {i}"
-            );
-        }
+        // let db = db.reopen().await;
```
danlaine, line 148: This can be deleted

firewood/src/storage/buffer.rs
```diff
@@ -790,42 +726,21 @@ mod tests {
...
-        assert!(disk_requester.collect_ash(1).unwrap().is_empty());
-
         // get RO view of the buffer from the beginning.
         let view = mut_store.get_view(0, hash.len() as u64).unwrap();
         assert_eq!(view.as_deref(), hash);
 
         // Commit the change. Take the delta from cached store,
         // then apply changes to the CachedStore.
-        let (redo_delta, wal) = mut_store.delta();
+        let (redo_delta, _) = mut_store.delta();
```
danlaine, line 735: It looks like the second return value of `delta` is only ever used in tests. Should we remove it?

firewood/src/db/proposal.rs
```diff
@@ -225,33 +225,23 @@ impl Proposal {
...
+            },
+            BufferWrite {
+                store_id: store.merkle.meta.id(),
+                delta: merkle_meta_redo,
+            },
+            BufferWrite {
+                store_id: rev_inner.root_hash_staging.id(),
+                delta: root_hash_redo,
+            },
+        ]));
```
danlaine, line 244: I'm not familiar with this part of the code -- what are these lines doing? It looks like they have something to do with "redo"s. Do we still need these if we don't need to construct prior revisions from disk? If so what are the things we're storing here used for?

firewood/src/storage/buffer.rs
```diff
@@ -342,43 +334,31 @@ async fn init_wal(
 
 async fn run_wal_queue(
```
danlaine, line 335: Similar question to above -- what does this do now that we're removing undo/redo persistence?

firewood/src/db/proposal.rs
xinifinity: Here is needed as it is actually persisting the new changes from this revision.

firewood/src/storage/buffer.rs
xinifinity: Same as above, it is actually persisting the dirty pages.

PR #624, opened 2024-04-02
implement node store; add `InMemReadWriteLinearStore` type for testing

danlaine:


firewood/src/storage/linear/mod.rs
```diff
@@ -96,3 +96,99 @@ impl<S: ReadLinearStore> ReadLinearStore for LinearStore<S> {
         todo!()
     }
 }
+
+pub mod tests {
```
richardpringle, line 100: missing `#[cfg(test)]`

firewood/src/storage/linear/mod.rs
```diff
@@ -96,3 +96,99 @@ impl<S: ReadLinearStore> ReadLinearStore for LinearStore<S> {
...
+                self.bytes.resize(offset + object.len(), 0);
+            }
+            self.bytes[offset..offset + object.len()].copy_from_slice(object);
+            Ok(object.len())
+        }
+    }
+
+    impl ReadLinearStore for InMemReadWriteLinearStore {
+        fn stream_from(&self, addr: u64) -> Result<impl Read, std::io::Error> {
+            Ok(&self.bytes[addr as usize..])
```
richardpringle, line 128: This can panic. I know it's for testing, but it should probably be consistent with how the real one works (assuming the real one won't panic)

firewood/src/storage/linear/mod.rs
```diff
@@ -96,3 +96,99 @@ impl<S: ReadLinearStore> ReadLinearStore for LinearStore<S> {
...
+        assert_eq!(read_bytes, vec![12]);
+        assert_eq!(num_read_bytes, 1);
+
+        // Read from past the end of the store
+        let mut reader = store.stream_from(store.size().unwrap()).unwrap();
+        let mut read_bytes = vec![];
+        let num_read_bytes = reader.read_to_end(&mut read_bytes).unwrap();
+        assert_eq!(read_bytes, vec![]);
+        assert_eq!(num_read_bytes, 0);
+    }
```
richardpringle, line 193: These should all be separate tests. If the first test fails, you'll have no idea how many (if any) the other tests fail and will have to fix each test one at a time instead of all the different tests that fail at once.

firewood/src/storage/linear/mod.rs
```diff
@@ -96,3 +96,79 @@ impl<S: ReadLinearStore> ReadLinearStore for LinearStore<S> {
...
+            }
+            self.bytes[offset..offset + object.len()].copy_from_slice(object);
+            Ok(object.len())
+        }
+    }
+
+    impl ReadLinearStore for InMemReadWriteLinearStore {
+        fn stream_from(&self, addr: u64) -> Result<impl Read, std::io::Error> {
+            if addr as usize >= self.bytes.len() {
+                return Err(std::io::Error::new(
```
rkuris, line 132: Reads streaming past the eof should just return an eof, not an error. You can just return an empty cursor.

This happens when reading from a proposal that has extended the file size. It then switches back to the base image to look for bytes past that, which do not exist.

firewood/src/storage/linear/mod.rs
```diff
@@ -96,3 +96,79 @@ impl<S: ReadLinearStore> ReadLinearStore for LinearStore<S> {
...
+
+            Ok(&self.bytes[addr as usize..])
+        }
+
+        fn size(&self) -> Result<u64, std::io::Error> {
+            Ok(self.bytes.len() as u64)
+        }
+    }
+
+    #[test_case(&[],0, &[1, 2, 3],&[1, 2, 3]; "write to empty store")]
```
rkuris, line 146: change the test so you can just bundle all the writes and then do reads (untested, but should work?)
This lets you test the empty case as well.
```rust
#[test_case(&[(b"123", 0), (b"45", 1)], (0, b"145"))] // overwrite some bytes
#[test_case(&[(b"123", 0), (b"456", 1)], (0, b"1456"))] // extend
#[test_case(&[(b"123", 0), (b"456", 3)], (0, b"123456"))] // only extend

// uncovered cases
#[test_case(&[(b"123", 0), (b"456", 4)], (0, b"123\0456"))] // zero fill
#[test_case(&[], (0, []))] // empty case
#[test_case(&[], (42, []))] // read past eof

// more cases here
fn in_mem_write(mods: &[(u64, &[u8])], expected: (u64, &[u8])) -> Result<(), Error> {
    let mut store = InMemReadWriteLinearStore::default();
    for mod in mods {
        store.write(mod.1, mod.0)?;
    }
    let mut actual = Vec::new();
    store.stream_from(expected.0)?;
    let nread = reader.read_to_end(&mut actual)?;
    assert_eq!(actual.len(), nread);
    assert_eq!(actual, expected.1);
}
```

rkuris (changes requested):

firewood/src/storage/linear/mod.rs
danlaine: refactored along these lines

firewood/src/storage/linear/mod.rs
danlaine: done

firewood/src/storage/linear/tests/mod.rs
```diff
@@ -67,3 +68,65 @@ fn reparent() {
...
+}
+
+impl ReadLinearStore for InMemReadWriteLinearStore {
+    fn stream_from(&self, addr: u64) -> Result<Box<dyn Read>, std::io::Error> {
+        let bytes = if addr as usize >= self.bytes.len() {
+            // Out of bounds. Return an empty cursor.
+            self.bytes[0..0].to_owned()
+        } else {
+            self.bytes[addr as usize..].to_owned()
+        };
```
rkuris, line 101: The warnings here are clues you can let the system do the checks for you:
```suggestion
        let bytes = self
            .bytes
            .get(addr as usize..)
            .unwrap_or_default()
            .to_owned();
```

Arguably the compiler would otherwise have to check the bounds twice (this might get optimized away?)

rkuris (approved):
Bit of a challenge to review due to lots of moved code.

One suggestion in there about using get() instead of bounds checking yourself

xinifinity (approved):

danlaine:
superseded by https://github.com/ava-labs/firewood/pull/637

PR #625, opened 2024-04-02
New node store interface

rkuris:
Work in progress for the new node store interface


firewood/src/storage/linear/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn stream_from(&self, addr: u64) -> Result<impl Read, Error>;
+    fn size(&self) -> Result<u64, Error>;
+}
+
+/// Some linear stores support updates
+pub(super) trait WriteLinearStore: Debug {
+    fn write(&mut self, offset: u64, object: &[u8]) -> Result<usize, Error>;
+}
+
+impl<ReadWrite: ReadLinearStore + Debug> WriteLinearStore for LinearStore<ReadWrite> {
```
danlaine, line 84: Should this be `ReadWrite: ReadLinearStore + WriteLinearStore + Debug`?
```suggestion
impl<ReadWrite: ReadLinearStore + Debug> WriteLinearStore for LinearStore<ReadWrite> {
```

rkuris: > Should this be `ReadWrite: ReadLinearStore + WriteLinearStore + Debug`?
Yes. Fixed when these methods were implemented in an upcoming diff.

firewood/src/storage/linear/proposed.rs
```diff
@@ -0,0 +1,526 @@
...
+            offset,
+            data: Box::from(object),
+        };
+
+        // we track areas to delete here, which happens when we combine
+        let mut merged_offsets_to_delete = Vec::new();
+
+        // TODO: combining a change where we constantly append to a file might be too
+        // expensive, so consider skipping the relocations if they don't overlap at all
+        for existing in self.new.range(..=insert.offset).rev() {
```
danlaine, line 125: Does this for loop need to iterate more than once? Don't we only need to consider the first diff that this iterator returns? Is it possible any subsequent diff overlaps with the insert?

firewood/src/storage/linear/proposed.rs
```diff
@@ -0,0 +1,526 @@
...
+                        },
+                    );
+                }
+                insert.offset = *existing.0;
+                merged_offsets_to_delete.push(insert.offset);
+                insert.data = new_data.into_boxed_slice();
+            }
+        }
+        // TODO: search for modifications after this one that can be merged
+        for existing in self.new.range(insert.offset + 1..) {
```
danlaine, line 162: Same question as above -- does this loop need to iterate more than once?

firewood/src/storage/linear/proposed.rs
```diff
@@ -0,0 +1,526 @@
...
+            // existing ----smme---- s=existing.0 e=existing.0+existing.1.len()
+            // adding   --smme------ s=updated_offset e=new_end
+            let new_end = insert.offset + insert.data.len() as u64;
+            if new_end >= *existing.0 {
+                // we have some overlap
+                let mut new_data = vec![];
+                for resulting_bytes_offset in insert.offset.. {
+                    new_data.push(
+                        if resulting_bytes_offset >= insert.offset
```
danlaine, line 171: `resulting_bytes_offset >= insert.offset` is always true

danlaine (commented):
I know this is still a WIP but leaving some feedback from what I've read so far

firewood/src/storage/linear/current.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
rkuris: Actually we need to do all of them. If you change this, you get a build failure.

firewood/src/storage/linear/proposed.rs
rkuris: Yeah it should be impossible with the current implementation. I think this code should just disappear though. I'm not sure it's worth coalescing left.

Consider the bulk insert case -- we keep appending new nodes. I don't want to keep reallocating all the stuff we are about to write repeatedly. It's almost like this makes sense to do sometimes. Figuring out what sometimes is seems challenging.

firewood/src/storage/linear/proposed.rs
rkuris: this one definitely does! Imagine you have two changes:
 `omomo`
and you're writing over both of them:
 `MMMM-`
we want to do this loop twice in that case to consume both original modifications.

firewood/src/storage/linear/proposed.rs
rkuris: good catch (the optimizer should have removed this test anyway, but no point cluttering the code)

firewood/src/storage/linear/proposed.rs
```diff
@@ -0,0 +1,526 @@
...
+
+        // TODO: combining a change where we constantly append to a file might be too
+        // expensive, so consider skipping the relocations if they don't overlap at all
+        for existing in self.new.range(..=insert.offset).rev() {
+            let old_end = *existing.0 + existing.1.len() as u64;
+            // found    smme--------- s=existing.0 e=old_end
+            // original ----smmme---- s=new_start e=new_start+object.len() [0 overlap]
+            // original --smmmme----- with overlap
+            // new      smmmmmmme----
+            if insert.offset <= old_end {
```
danlaine, line 131: Should this be `<` instead of `<=`?

IIUC `old_end` is the first index _after_ the `existing` diff ends. For example, if `existing.0` and `existing.1` are `0` and `&[1,2,3]` respectively, then `old_end` is `3`, and index `3` is after this diff. And IIUC `insert.offset` is the first index affected by the diff. So if `insert.offset == old_end` then these diffs don't overlap.

firewood/src/storage/linear/mod.rs
richardpringle: Does it ever make sense to have a linear store that's `Read` but not `Write`? Can simplify your trait bounds by making `trait WriteLinearStore: ReadLinearStore { ... }`

rkuris: Fixed in newest code

firewood/src/storage/linear/proposed.rs
rkuris: Consensus is that we shouldn't coalesce blocks that do not overlap (i.e., adjacent blocks) on either side.

firewood/src/storage/linear/current.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    pub(crate) fn reparent(self, parent: Arc<LinearStore<P>>) -> Historical<P> {
+        Historical::from_current(self.old, parent)
+    }
+}
+
+impl<P: ReadLinearStore> ReadLinearStore for Current<P> {
+    fn stream_from(
+        &self,
+        _addr: u64,
```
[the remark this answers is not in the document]
rkuris: No. At this level, everything is just bytes. `LinearAddress` can't be 0, but LinearStores do store things at address 0. Think of `LinearAddress` as `NodeOrFreeSpaceHeaderOrFreeSpaceLinearAddress` which can't be 0, but that's too hard to type. And I refuse to use `ObjectLinearAddress` :)

danlaine (approved):

PR #629, opened 2024-04-10
First Implementation of Persisted Store

xinifinity:
A single file backed persisted store using tokio async IO.


firewood/src/storage/persisted/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+            .open(&path)
+            .await?;
+        Ok(Self {
+            path,
+            fd: Mutex::new(file),
+        })
+    }
+
+    // TODO: impl ReadLinearStore when it is async
+    pub async fn stream_from(&self, addr: u64, length: usize) -> Result<Bytes, Error> {
```
xinifinity, line 42: Removed the premature page cache.

rkuris, line 42: The point of `stream_from` is to return a readable stream (something that implements the `Read` trait or perhaps in your case `AsyncRead`). You can't pass in a length, otherwise this is just a read method.

firewood/src/storage/persisted/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        let mut file = self.fd.lock().await;
+        file.seek(SeekFrom::Start(addr)).await?;
+        let mut buffer = vec![0; length];
+        file.read_exact(&mut buffer).await?;
+        Ok(Bytes::from(buffer))
+    }
+
+    pub async fn write_all(&mut self, writes: &[StoreWrite]) -> Result<(), Error> {
```
rkuris, line 50: How does this hook up to the coordinator, and how do you get notified when all these writes complete?

rkuris (changes requested):
How will this work next week when we replace shale? Who calls these methods?

It looks like you're trying to implement an async FileBacked, but this doesn't have any of the things that would make it worth using. What we need is some way to send a bunch of writes somewhere and perform them in parallel.

This is what's in the doc:
```
The coordinator during commit handles the write path asynchronously. It will pass the writes for each proposal to the disk write scheduler and let it send to a separate thread to flush the write by calling PersistedStore.write_all. And for each batch from a proposal, when the write is done an acknowledgement will be sent back to the scheduler, so that the caller can know the writes are complete.
```

Where is the thread that will be used by the coordinator, how do you initialize it, and how do you pass the data that is to be written to it? Also, how do you get notified that it's done?

xinifinity:
> How will this work next week when we replace shale? Who calls these methods?
> 
> It looks like you're trying to implement an async FileBacked, but this doesn't have any of the things that would make it worth using. What we need is some way to send a bunch of writes somewhere and perform them in parallel.
> 
> This is what's in the doc:
> 
> ```
> The coordinator during commit handles the write path asynchronously. It will pass the writes for each proposal to the disk write scheduler and let it send to a separate thread to flush the write by calling PersistedStore.write_all. And for each batch from a proposal, when the write is done an acknowledgement will be sent back to the scheduler, so that the caller can know the writes are complete.
> ```
> 
> Where is the thread that will be used by the coordinator, how do you initialize it, and how do you pass the data that is to be written to it? Also, how do you get notified that it's done?

Working on the disk scheduler PR as a follow up for handing async write in parallel, this is the initial PR to lay out the direct IO handing in PersistedStore.

firewood/src/storage/persisted/mod.rs
xinifinity: This will be handled by the scheduler PR.

firewood/src/storage/persisted/mod.rs
xinifinity: This is intended to replace `FileBacked` in the read path. However as the `LinearStore` signature is sync, will update its signature to match when `LinearStore` switch to async. Which definitely can come later.

rkuris: If that's the case, you'll need to implement the other methods. We don't want a length here. The length is determined by how many bytes they read from the returned trait.

firewood/src/storage/persisted/mod.rs
rkuris: Not sure I'm following. Do you want the scheduler to also be async? For the prototype we agreed to keep everything sync, and this async code would get fired off to do the IO.

As it is today, we have no async context so we can't use this method.

I think the input to this is the same structure used by `new` in `Proposed`. Without that, I think we'd have to make a new structure which would involve some copies, unless you have some other idea.

PR #631, opened 2024-04-11
[WIP] Implement `linear::Historical`

danlaine:


firewood/src/storage/linear/historical.rs
```diff
@@ -11,12 +11,22 @@ use super::{LinearStore, ReadLinearStore};
 /// A [Historical] [LinearStore] supports read operations only
 #[derive(Debug)]
 pub(crate) struct Historical<P: ReadLinearStore> {
-    old: BTreeMap<u64, Box<[u8]>>,
-    parent: Arc<LinearStore<P>>,
+    pub(crate) old: BTreeMap<u64, Box<[u8]>>,
```
rkuris, line 14: This means we bleed the implementation details of Historical outside this file. Why do we need to do that?

firewood/src/storage/linear/proposed.rs
```diff
@@ -32,16 +33,40 @@ impl<P: ReadLinearStore, M> Proposed<P, M> {
     }
 }
 
+#[derive(Debug)]
+struct Layer<'a, P: ReadLinearStore> {
+    parent: Arc<LinearStore<P>>,
+    diffs: &'a BTreeMap<u64, Box<[u8]>>,
```
rkuris, line 39: not sure how this will work with `Proposed`, which needs both the old and the new data. The old sticks around in case it commits, and gets migrated into a `Current`. If we don't keep the old data, we'd have have to recreate it on commit.

PR #633, opened 2024-04-17
Refactor linear storage types

danlaine:
* Removes `current.rs`.
* Removes `LinearStore` type and replaces with `ImmutableLinearStore` and `MutableLinearStore` types.
* Changes `parent` field of `Historical`, `Proposed` to be `Arc<dyn ReadLinearStore>` instead of `LinearStore`.


firewood/Cargo.toml
```diff
@@ -37,6 +37,7 @@ bincode = "1.3.3"
 bitflags = { version = "2.4.2", features = ["bytemuck"] }
 env_logger = { version = "0.11.2", optional = true }
 log = { version = "0.4.20", optional = true }
+tempdir = "0.3.7"
```
rkuris, line 40: move to dev-dependencies as it's only used in tests I think

firewood/src/storage/linear/mod.rs
```diff
@@ -68,47 +70,60 @@ mod filebacked;
...
-
-/// All linearstores support reads
-pub(super) trait ReadLinearStore: Send + Sync + Debug {
-    fn stream_from(&self, addr: u64) -> Result<Box<dyn Read + '_>, Error>;
-    fn size(&self) -> Result<u64, Error>;
+pub(super) enum ImmutableLinearStore {
+    Historical(Historical),
+    Proposed(Arc<Proposed<Immutable>>),
+    FileBacked(FileBacked),
+    Invalid, // TODO do we need this variant?
```
rkuris, line 79: Lets say no

firewood/src/storage/linear/mod.rs
```diff
@@ -68,47 +70,60 @@ mod filebacked;
...
     }
 }
 
-/// Some linear stores support updates
-pub(super) trait WriteLinearStore: Debug {
-    fn write(&mut self, offset: u64, object: &[u8]) -> Result<usize, Error>;
+#[derive(Debug)]
+pub(super) enum MutableLinearStore {
+    Proposed(Proposed<Mutable>),
+    Invalid, // TODO do we need this variant? If not we can remove MutableLinearStore
```
rkuris, line 105: Yet another reason to remove Invalid. Let's make these invalids actually work, even though it is probably a VM error to use them. They can't be committed ever of course, and we will drop references to them once we know they can't be committed

firewood/src/storage/linear/proposed.rs
```diff
@@ -7,39 +7,55 @@ use std::io::{Error, Read};
...
 /// which could be a another [Proposed] or is [FileBacked](super::filebacked::FileBacked)
 /// The M type parameter indicates the mutability of the proposal, either read-write or readonly
+/// TODO update comment
 #[derive(Debug)]
-pub(crate) struct Proposed<P: ReadLinearStore, M> {
-    pub(crate) new: BTreeMap<u64, Box<[u8]>>,
-    old: BTreeMap<u64, Box<[u8]>>,
-    pub(crate) parent: Arc<LinearStore<P>>,
-    phantom: PhantomData<M>,
+pub(crate) struct Proposed<T> {
```
rkuris, line 19: nit: T -> M

I like seeing M for mutability, which is all this is

firewood/src/storage/linear/proposed.rs
```diff
@@ -164,51 +180,67 @@ impl<P: ReadLinearStore> WriteLinearStore for Proposed<P, Mutable> {
...
+        proposed.stream_from(0).unwrap().read_exact(&mut data)?;
+        assert_eq!(Some(data.as_slice()), TEST_DATA.get(..data.len()));
 
         Ok(())
     }
 
     #[test]
     fn smoke_mutate() -> Result<(), std::io::Error> {
-        let mut sut: Proposed<ConstBacked, Mutable> = ConstBacked::new(ConstBacked::DATA).into();
+        let parent = new_temp_filebacked(TEST_DATA);
```
rkuris, line 228: Not sure I love this, as it's now creating files whereas before it didn't need to. We aren't testing filebacked, we're just testing mutate.

rkuris (changes requested):

firewood/src/storage/linear/historical.rs
```diff
@@ -4,28 +4,31 @@
...
 #[derive(Debug)]
-pub(crate) struct Historical<P: ReadLinearStore> {
+pub(crate) struct Historical {
     /// (offset, value) for every area of this LinearStore that is modified in
     /// the revision after this one (i.e. `parent`).
     /// For example, if the first 3 bytes of this revision are [0,1,2] and the
     /// first 3 bytes of the next revision are [4,5,6] then this map would
     /// contain [(0, [0,1,2])].
-    pub(crate) changed_in_parent: BTreeMap<u64, Box<[u8]>>,
+    changed_in_parent: BTreeMap<u64, Box<[u8]>>,
```
richardpringle, line 22: If there's a time to use `bytes::Bytes`, it's right now. It's the most used sub-crate by any "networking" crate (in other words, any crate that's copying around byte slices). It's in the top 50 most downloaded crates of all time for good reason. 

@danlaine, this comment isn't just for you.

firewood/src/storage/linear/filebacked.rs
```diff
@@ -9,39 +9,61 @@
...
+    use tempdir::TempDir;
+
+    pub fn new_temp_filebacked(data: &[u8]) -> FileBacked {
+        let temp_dir = TempDir::new("test_historical_stream_from").unwrap();
+        let db_path = temp_dir.path().join("db");
+        let mut store = FileBacked::new(db_path).unwrap();
+        store.write(0, data).unwrap();
+        store
+    }
+}
```
richardpringle, line 69: I've done this before, and it was definitely something I regretted. `tests` modules should never have anything other than private visibility. Having a function written in the parent module that's only callable from tests is perfectly fine. 
```suggestion
#[cfg(test)]
#[allow(clippy::unwrap_used)]
pub(super) fn new_temp_filebacked(data: &[u8]) -> FileBacked {
    let temp_dir = tempdir::TempDir::new("test_historical_stream_from").unwrap();
    let db_path = temp_dir.path().join("db");
    let mut store = FileBacked::new(db_path).unwrap();
    store.write(0, data).unwrap();
    store
}
```

firewood/src/storage/linear/historical.rs
```diff
@@ -34,22 +37,27 @@ impl<P: ReadLinearStore> Historical<P> {
...
+        addr: u64,
+    ) -> Result<Box<dyn std::io::Read + '_>, std::io::Error> {
+        Ok(Box::new(LayeredReader::new(
+            addr,
+            Layer::new(self.parent.clone(), &self.changed_in_parent),
+        )))
     }
 
-    fn size(&self) -> Result<u64, std::io::Error> {
+    pub(super) const fn size(&self) -> Result<u64, std::io::Error> {
```
richardpringle, line 51: It doesn't really make sense to have this as a const method. When are you ever going to be creating `Historical` in a const context (at compile time)?

Or was this a clippy suggestion? I would actually ignore the rule for this if that's the case.

firewood/src/storage/linear/filebacked.rs
richardpringle: Alternatively, you could put `#[cfg(not(test))]` above `new` and make this function a static method.

firewood/src/storage/linear/mod.rs
```diff
@@ -68,47 +70,60 @@ mod filebacked;
...
 
-    fn size(&self) -> Result<u64, Error> {
-        self.deref().size()
+    pub fn size(&self) -> Result<u64, Error> {
+        match self {
+            ImmutableLinearStore::Historical(historical) => historical.size(),
+            ImmutableLinearStore::Proposed(proposed) => proposed.size(),
+            ImmutableLinearStore::FileBacked(filebacked) => filebacked.size(),
+            ImmutableLinearStore::Invalid => Err(std::io::ErrorKind::InvalidData.into()),
+        }
```
richardpringle, line 98: 🤔, if you're only methods are delegating, why not just use a trait object? Then you don't need the match statement. You can implement `ReadLinearStore` for `Arc<T>` where `T: ReadLinearStore`.

firewood/src/storage/linear/proposed.rs
richardpringle: I would literally call it `Mutability` then. Sometimes it's beneficial to deviate a little from the conventions 😉

firewood/src/storage/linear/proposed.rs
```diff
@@ -7,39 +7,55 @@ use std::io::{Error, Read};
...
+pub(crate) struct Proposed<T> {
+    new: BTreeMap<u64, Box<[u8]>>,
+    pub(super) old: BTreeMap<u64, Box<[u8]>>,
+    pub(super) parent: Arc<ImmutableLinearStore>,
+    state: PhantomData<T>,
 }
 
-impl<P: ReadLinearStore, M> Proposed<P, M> {
-    pub(crate) fn new(parent: Arc<LinearStore<P>>) -> Self {
+impl From<Proposed<Mutable>> for Proposed<Immutable> {
```
richardpringle, line 26: This is a nit, but I think it would be more explicit using a method named something like `freeze`. 

The call site goes from this
```rust
let proposal: Proposal<Immutable> = proposal.into();
// or 
let proposal = Proposal::<Immutable>::from(proposal);
```
to
```rust
let proposal = proposal.freeze();
```
Let Rust-Analyzer give you the type hints from there.

firewood/src/storage/linear/mod.rs
```diff
@@ -68,47 +70,60 @@ mod filebacked;
...
     }
 }
 
-/// Some linear stores support updates
-pub(super) trait WriteLinearStore: Debug {
-    fn write(&mut self, offset: u64, object: &[u8]) -> Result<usize, Error>;
+#[derive(Debug)]
+pub(super) enum MutableLinearStore {
+    Proposed(Proposed<Mutable>),
```
richardpringle, line 104: From what I've seen, you already have mutability built in. Let's say for a second that `Proposal` isn't generic over its mutability: `Proposal` (no `<T>`). 

Your write method requires mutable access, right? I can see here that the `Proposal` is owned and you can call `write` whenever you want. Your `ImmutableLinearStore` has an `Arc<Proposal>`. That's already immutable. You're not gaining anything (at least not that I can see) with the generic. 

Can you point me to the code where this is being used in a way that the type system can't already facilitate?

firewood/src/storage/linear/proposed.rs
```diff
@@ -7,39 +7,55 @@ use std::io::{Error, Read};
...
 #[derive(Debug)]
-pub(crate) struct Proposed<P: ReadLinearStore, M> {
-    pub(crate) new: BTreeMap<u64, Box<[u8]>>,
-    old: BTreeMap<u64, Box<[u8]>>,
-    pub(crate) parent: Arc<LinearStore<P>>,
-    phantom: PhantomData<M>,
+pub(crate) struct Proposed<T> {
+    new: BTreeMap<u64, Box<[u8]>>,
+    pub(super) old: BTreeMap<u64, Box<[u8]>>,
+    pub(super) parent: Arc<ImmutableLinearStore>,
```
richardpringle, line 22: If you put a `Arc<MutableStore>` here, it would still be immutable.

richardpringle (commented):
So... 😅, we've re-implemented two different features built into the language. we're doing dynamic dispatch via enums and implementing mutability via generics. 

Is it really worth it? If so why/how? 

I know I said I didn't mind the enum thing, but I didn't realize that it was only being used as a type wrapper.

firewood/src/storage/linear/proposed.rs
danlaine: Proposed no longer has a generic. It implements `ReadLinearStore` and `WriteLinearStore` now

firewood/src/storage/linear/proposed.rs
danlaine: removed

firewood/src/storage/linear/filebacked.rs
danlaine: removed

firewood/src/storage/linear/historical.rs
danlaine: I'm not opposed but think that should be a separate PR

rkuris: We can create an issue for this and do it separately. The nice bit about `Bytes` is that you get nearly free copies and moves. We need this at commit time, when `old` becomes `was` in the newly created `LinearStore`.

firewood/src/storage/linear/mod.rs
```diff
@@ -71,44 +68,15 @@ mod proposed;
 #[cfg(test)]
 mod tests;
 
-#[derive(Debug)]
-pub(super) struct LinearStore<S: ReadLinearStore> {
-    state: S,
-}
-
 /// All linearstores support reads
-pub(super) trait ReadLinearStore: Send + Sync + Debug {
```
rkuris, line 80: Unfortunately, you should not remove this. If you run against the nightly version of clippy, it will complain that you could have a `ReadLinearStore` that is not `Send` or `Sync`, and since you wrap that in an `Arc` (and we probably will need these to be Send/Sync) we should specifically say that this is required.

firewood/src/storage/linear/mod.rs
```diff
@@ -71,44 +68,15 @@ mod proposed;
...
-impl<S: ReadLinearStore> ReadLinearStore for LinearStore<S> {
-    fn stream_from(&self, addr: u64) -> Result<Box<dyn Read + '_>, Error> {
-        self.state.stream_from(addr)
-    }
-
-    fn size(&self) -> Result<u64, Error> {
-        todo!()
-    }
-}
+pub(super) trait ReadWriteLinearStore: ReadLinearStore + WriteLinearStore + Debug {}
```
rkuris, line 82: As Richard pointed out, you can't have a `WriteLinearStore` that doesn't allow for reads (completely meaningless). So, lets just do:
```rust
pub(super) trait WriteLinearStore: ReadLinearStore {
    fn write(&mut self, offset: u64, object: &[u8]) -> Result<usize, Error>;
}
```
There is no need to repeat `Debug` or `Send` or `Sync` since trait bounds are inherited.

firewood/src/storage/linear/proposed.rs
```diff
@@ -4,42 +4,40 @@
...
-/// The P type parameter indicates the state of the linear store for it's parent,
-/// which could be a another [Proposed] or is [FileBacked](super::filebacked::FileBacked)
-/// The M type parameter indicates the mutability of the proposal, either read-write or readonly
 #[derive(Debug)]
-pub(crate) struct Proposed<P: ReadLinearStore, M> {
-    pub(crate) new: BTreeMap<u64, Box<[u8]>>,
-    old: BTreeMap<u64, Box<[u8]>>,
-    pub(crate) parent: Arc<LinearStore<P>>,
-    phantom: PhantomData<M>,
+pub(crate) struct Proposed {
```
rkuris, line 14: I don't like removing M here. This means all `Proposed` can be mutated by mistake. Lets avoid this problem by having two types:
`ProposedMutable` and `ProposedImmutable` which the compiler can do for you with `Proposed<Mutable>` and `Proposed<Immutable>` and some `PhantomData`. Then, we only implement `WriteLinearStore` on `Proposed<Mutable>`.

firewood/src/storage/linear/proposed.rs
```diff
@@ -53,20 +51,11 @@ impl<P: ReadLinearStore, M: Send + Sync + Debug> ReadLinearStore for Proposed<P,
...
-/// Marker that the Proposal is mutable
-#[derive(Debug)]
-pub(crate) struct Mutable;
-
-/// Marker that the Proposal is immutable
-#[derive(Debug)]
-pub(crate) struct Immutable;
-
-impl<P: ReadLinearStore> WriteLinearStore for Proposed<P, Mutable> {
+impl WriteLinearStore for Proposed {
```
rkuris, line 54: Needs to be bound by mutability. I don't want *all* `Proposed` types to have the `write` method.

rkuris (changes requested):

firewood/src/storage/linear/proposed.rs
danlaine: I removed this based on @richardpringle 's [comment](https://github.com/ava-labs/firewood/pull/633#discussion_r1569619102) which I think makes a reasonable point. If someone has a `&Proposed`, then they can't mutate it, but if they have a `&mut proposed` then they can. If they have a `dyn ReadLinearStore` they can't mutate it, but if they have a `dyn WriteLinearStore` they can. 

What does having these types `Proposed<Mutable>` and `Proposed<Immutable>` buy us that `mut` and the `dyn`s don't?

firewood/src/storage/linear/proposed.rs
danlaine: Same response as in https://github.com/ava-labs/firewood/pull/633#discussion_r1574805130 -- doesn't the use (or lack of use) of `mut` already give us a way of controlling when a `Proposed` is `write`able?

firewood/src/storage/linear/proposed.rs
rkuris: It buys us that you can't call `write` on an immutable proposal, and this is caught at compile time.

firewood/src/storage/linear/proposed.rs
danlaine: https://github.com/ava-labs/firewood/pull/633#discussion_r1575131164

rkuris (approved):
Seems fine. I'm not sure I fully understand why we need `Layer` to be a concrete type. It feels more like a trait to me.

PR #634, opened 2024-04-19
Rkuris/use bytes attempt

rkuris:


firewood/src/storage/linear/historical.rs
```diff
@@ -0,0 +1,80 @@
...
+pub(crate) struct Historical<P: ReadLinearStore> {
+    /// (offset, value) for every area of this LinearStore modified in
+    /// the revision after this one (i.e. `parent`).
+    /// That is, what each area `was` in this revision.
+    /// For example, if the first 3 bytes of this revision are `[0,1,2]` and the
+    /// first 3 bytes of the next revision are `[4,5,6]` then this map contains
+    /// `[(0, [0,1,2])]`.
+    pub(crate) was: BTreeMap<u64, Bytes>,
+    /// The state of the revision after this one.
+    pub(crate) parent: Arc<LinearStore<P>>,
```
richardpringle, line 24: I think you need a trait object here, don't you?

firewood/src/storage/node.rs
```diff
@@ -0,0 +1,230 @@
...
+
+impl<T: WriteLinearStore + ReadLinearStore> NodeStore<T> {
+    /// Allocate space for a [Node] in the [LinearStore]
+    fn create(&mut self, node: &Node) -> Result<LinearAddress, Error> {
+        let serialized =
+            bincode::serialize(node).map_err(|e| Error::new(ErrorKind::InvalidData, e))?;
+        // TODO: search for a free space block we can reuse
+        let addr = self.page_store.size()?;
+        self.page_store
+            .write(addr, Bytes::copy_from_slice(&serialized))?;
```
richardpringle, line 85: Is there no way to create bytes without copying here?

firewood/src/storage/node.rs
```diff
@@ -0,0 +1,230 @@
...
+            bincode::serialize(node).map_err(|e| Error::new(ErrorKind::InvalidData, e))?;
+        if serialized.len() != size {
+            // this node is a different size, so allocate a new node
+            // TODO: we could do better if the new node is smaller
+            let new_address = self.create(node)?;
+            self.delete(addr)?;
+            Err(UpdateError::NodeMoved(new_address))
+        } else {
+            self.page_store
+                .write(addr.into(), Bytes::copy_from_slice(&serialized))?;
```
richardpringle, line 106: Same here?

firewood/src/storage/node.rs
```diff
@@ -0,0 +1,230 @@
...
+    const fn new(reader: T) -> Self {
+        Self {
+            reader,
+            bytes_read: 0,
+        }
+    }
+}
+
+impl<T: Read> Read for ReaderWrapperWithSize<T> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
```
richardpringle, line 182: Why is buf mutable on a read? Is that just the read API?

richardpringle (commented):
Just a quick skim. I left a couple of questions. I didn't checkout the code or look at some of the type signatures from the crate functions used so some of the answers might be obvious. 

Happy to see the use of bytes::Bytes though

rkuris:
Not doing it this way for now

PR #636, opened 2024-04-24
Add the write scheduler for parallel batch writes to disk

xinifinity:
This adds a DiskScheduler and DiskRequester. The DiskRequester send the batch writes to the separate write thread which should spawned by the Coordinator (caller) and be processed by the DiskScheduler. Along each batch writes a notify is attached so that once the writes are complete the caller will be notified.

### DIFF
--- a/firewood/src/storage/persisted/scheduler.rs
+++ b/firewood/src/storage/persisted/scheduler.rs
@@ -1,0 +1,138 @@
+use super::{PersistedStore, StoreWrite};
+use std::sync::Arc;
+use thiserror::Error;
+use tokio::sync::{mpsc, Notify};
+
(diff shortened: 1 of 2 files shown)


firewood/src/storage/persisted/scheduler.rs
```diff
@@ -0,0 +1,138 @@
...
+
+                        // Ack the writes are complete.
+                        notifier.notify_one();
+                    }
+                }
+            })
+            .await;
+
+        // when finished process all requests, wait for any pending-futures to complete
+        local_pool.await;
```
rkuris, line 72: How do we get here? `run_until` will block until the future completes, which is in a `loop` with no `break` statement.

PR #647, opened 2024-05-15
Enabling changes for `merkle.remove`

danlaine:
Broken out of #644 to make reviewing easier.

* Changes the value type of `modified` to `Option`, where `None` indicates the node was deleted.
* Allows setting the root of a node store to `None`.

### DIFF
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -198,7 +198,7 @@ impl<T: WriteLinearStore> HashedNodeStore<T> {
         Ok(())
     }
 
-    pub fn set_root(&mut self, root_addr: LinearAddress) -> Result<(), Error> {
+    pub fn set_root(&mut self, root_addr: Option<LinearAddress>) -> Result<(), Error> {
         self.nodestore.set_root(root_addr)
     }
 }
@@ -265,7 +265,7 @@ mod test {
             value: Box::new(*b"abc"),
         });
         let addr = hns.create_node(node).unwrap();
-        hns.set_root(addr).unwrap();
+        hns.set_root(Some(addr)).unwrap();
 
         let frozen = hns.freeze().unwrap();
         assert_ne!(frozen.root_hash().unwrap(), Default::default());
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -4,6 +4,7 @@
 use sha3::{Digest, Keccak256};
 use std::collections::HashMap;
 use std::io::Error;
+use std::io::ErrorKind::NotFound;
 use std::iter::once;
 use std::sync::{Arc, OnceLock};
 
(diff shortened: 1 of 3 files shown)


firewood/src/hashednode.rs
```diff
@@ -50,15 +53,24 @@ impl<T: ReadLinearStore> From<NodeStore<T>> for HashedNodeStore<T> {
 impl<T: ReadLinearStore> HashedNodeStore<T> {
     pub fn read_node(&self, addr: LinearAddress) -> Result<Arc<Node>, Error> {
         if let Some(modified_node) = self.modified.get(&addr) {
-            Ok(modified_node.0.clone())
+            if let Some((modified_node, _)) = modified_node {
+                Ok(modified_node.clone())
+            } else {
+                Err(Error::new(std::io::ErrorKind::Other, "Node not found"))
```
rkuris, line 59: Although I think this code shouldn't be needed at all (see earlier comment) at a minimum this should be the same as when a node is read that has been freed, see `read_node` in nodestore which does this:
```
Err(Error::new(
                ErrorKind::InvalidData,
                "Attempted to read a freed area",
            )),
```

firewood/src/hashednode.rs
```diff
@@ -22,7 +23,9 @@ use storage::{ReadLinearStore, WriteLinearStore};
 #[derive(Debug)]
 pub struct HashedNodeStore<T: ReadLinearStore> {
     nodestore: NodeStore<T>,
-    modified: HashMap<LinearAddress, (Arc<Node>, u8)>,
+    // Linear Address -> (Node, Node Size Index)
+    // None means the node was deleted.
+    modified: HashMap<LinearAddress, Option<(Arc<Node>, u8)>>,
```
rkuris, line 28: I think this is not needed. Why can't you just remove it from modified when it is deleted?

Nobody should ever reference it once it has been deleted in any case.

rkuris (changes requested):

firewood/src/hashednode.rs
danlaine: What does the `delete_node` implementation look like without this? 

> Why can't you just remove it from modified when it is deleted?
What if the node being deleted wasn't modified in this `HashedNodeStore`? Then removing it from `modified` would be a no-op.

If the node _is_ in `modified`, and we delete it, what happens then? When do we tell the underlying node store to delete the node at the given address?

rkuris: > If the node _is_ in `modified`, and we delete it, what happens then? When do we tell the underlying node store to delete the node at the given address?

Yes. We should just delete it from the linear store.

danlaine: >  If you don't do this, then the space for it can't be reused by another node in the same batch, for example.

Yeah that's definitely a bug in this PR. Going to close this. I removed these changes from #644

PR #660, opened 2024-05-23
Hash refactor

danlaine:
* Move hashing logic out of the `impl HashedNodeStore` because we don't actually use any of the `HashedNodeStore`'s fields.
* Add type `HashPreimage`. This generalizes the existing logic, which only knows how to hash a `&Node`. I think this will be useful in implementing proofs.
* Renames `hasher` in instances where it's actually not necessarily a hasher (it could be a vec we're writing the node preimage into)

### DIFF
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -356,8 +356,8 @@ impl<'a, K: Iterator<Item = u8> + Clone, V: AsRef<[u8]>> HashPreimage<'a, K, V> 
 }
 
 /// Returns the SHA-256 hash of the `preimage`.
-pub(crate) fn _hash<'a, K: Iterator<Item = u8> + Clone, V: AsRef<[u8]>>(
-    preimage: HashPreimage<'a, K, V>,
+pub(crate) fn _hash<K: Iterator<Item = u8> + Clone, V: AsRef<[u8]>>(
+    preimage: HashPreimage<'_, K, V>,
 ) -> TrieHash {
     let mut hasher: Sha256 = Sha256::new();
     preimage.write(&mut hasher);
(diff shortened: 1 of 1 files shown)


firewood/src/hashednode.rs
```diff
@@ -312,104 +311,140 @@ impl HasUpdate for Vec<u8> {
...
     }
 }
 
-fn add_value_to_hasher<H: HasUpdate, T: AsRef<[u8]>>(hasher: &mut H, value: Option<T>) {
-    let Some(value) = value else {
-        let value_exists: u8 = 0;
-        hasher.update([value_exists]);
-        return;
+/// Returns the SHA-256 hash of the `preimage`.
+pub(crate) fn _hash<K: Iterator<Item = u8> + Clone, V: AsRef<[u8]>>(
```
rkuris, line 359: Why is this `_hash` instead of `hash`?

firewood/src/hashednode.rs
```diff
@@ -312,104 +311,140 @@ impl HasUpdate for Vec<u8> {
...
+pub fn hash_node(node: &Node, path_prefix: &Path) -> TrieHash {
+    let mut hasher: Sha256 = Sha256::new();
+    write_hash_preimage(node, path_prefix, &mut hasher);
+    hasher.finalize().into()
+}
+
+/// Returns the serialized representation of `node` used as the pre-image
+/// when hashing the node. The node is at the given `path_prefix`.
+pub fn hash_preimage(node: &Node, path_prefix: &Path) -> Box<[u8]> {
+    let mut buf = vec![];
```
rkuris, line 415: nit: Maybe `vec::with_capacity(node.value.len())` or something to avoid reallocating the buffer every time.

rkuris (approved):

firewood/src/hashednode.rs
danlaine: Changed to pre-allocate capacity

firewood/src/hashednode.rs
danlaine: It was unused. Removed.

PR #662, opened 2024-05-24
Implement proof generation and verification

danlaine:
This PR implements a first pass on proof generation and verification. It:
* Removes dead code
* Adds Proof, ProofNode types
* Changes the `ValueDigest` type to be generic
* Removes the `Preimage` type. I couldn't figure out how to implement a blanket implementation for `to_hash` and `write` with the additional generics `V` and `H`.
* Implements `Merkle.prove` and `Proof.verify`


firewood/src/db.rs
```diff
@@ -79,7 +77,7 @@ impl<T: ReadLinearStore> api::DbView for HistoricalRev<T> {
     async fn single_key_proof<K: api::KeyType>(
         &self,
         _key: K,
-    ) -> Result<Option<Proof<Vec<u8>>>, api::Error> {
+    ) -> Result<Option<Proof>, api::Error> {
```
rkuris, line 80: Removing this generic means each ProofNode can be of a different type, which isn't what we want I think. All ProofNodes within a proof should either be owned or borrowed (currently all of them are borrowed).

I'd rather see the generic here and have ProofNode use the type from it's parent.

firewood/src/hashednode.rs
```diff
@@ -310,43 +309,47 @@ impl HasUpdate for Vec<u8> {
...
+    Value(T),
+    /// TODO this variant will be used when we deserialize a proof node
+    /// from a remote Firewood instance. The serialized proof node they
+    /// send us may the hash of the value, not the value itself.
+    _Hash(T),
 }
 
-trait Hashable {
+pub(crate) trait Hashable {
+    type T: AsRef<[u8]>;
```
rkuris, line 323: nit: associated types usually have better names, suggest ValueDigestType or something like that

firewood/src/hashednode.rs
```diff
@@ -310,43 +309,47 @@ impl HasUpdate for Vec<u8> {
...
-    fn write(self, buf: &mut impl HasUpdate);
+    fn write(&self, buf: &mut impl HasUpdate);
 }
 
 // Implement Preimage for all types that implement Hashable
-impl<T> Preimage for T
+impl<T: Hashable> Preimage for T
 where
-    T: Hashable,
+    T::T: AsRef<[u8]>,
```
rkuris, line 344: Either use a `where` clause or bounds, not both

`impl<T: Hashable + AsRef<[u8]>>`

or
```
impl<T> Preimage for T
where
    T: Hashable,
    T: AsRef<[u8]>,
```

firewood/src/merkle.rs
```diff
@@ -23,8 +23,8 @@ pub type Value = Vec<u8>;
 
 #[derive(Debug, Error)]
 pub enum MerkleError {
-    #[error("uninitialized")]
-    Uninitialized,
+    #[error("can't generate proof for empty node")]
```
rkuris, line 26: 🏆

firewood/src/merkle.rs
```diff
@@ -796,10 +807,33 @@ impl<'a, T: PartialEq> PrefixOverlap<'a, T> {
...
+    use std::time::{SystemTime, UNIX_EPOCH};
+
     use super::*;
+    use rand::{rngs::StdRng, Rng, SeedableRng};
     use storage::MemStore;
     use test_case::test_case;
 
+    // Returns n random key-value pairs.
+    fn generate_random_kvs(seed: u64, n: usize) -> Vec<(Vec<u8>, Vec<u8>)> {
+        println!("Used seed: {}", seed);
```
rkuris, line 819: nit: `println!` -> `eprintln!` if you need this

firewood/src/merkle.rs
```diff
@@ -1048,15 +1082,56 @@ mod tests {
...
+        let proof = merkle.prove(b"any-key");
+        assert!(matches!(proof.unwrap_err(), MerkleError::Empty));
+    }
+
+    #[test]
+    fn single_key_proof() {
+        let mut merkle = create_in_memory_merkle();
 
-    //         let proof = merkle.prove(b"any-key", root_addr).unwrap();
+        let seed = SystemTime::now()
```
rkuris, line 1096: Definitely print this seed! If the test fails, the person debugging the test can set the seed manually. Even better, something like this:
```
         let seed = std::env::var("FIREWOOD_TEST_SEED")
             .ok()
             .map_or_else(
                 || None,
                 |s| Some(str::parse(&s).expect("couldn't parse FIREWOOD_TEST_SEED; must be a u64")),
             )
             .unwrap_or_else(|| thread_rng().gen());
```

firewood/src/proof.rs
```diff
@@ -1,171 +1,225 @@
...
+        let key: Vec<u8> = NibblesIterator::new(key.as_ref()).collect();
+
+        let Some(last_node) = self.0.last() else {
+            return Err(ProofError::Empty);
+        };
+
+        let mut expected_hash = root_hash;
+
+        // TODO danlaine: Is there a better way to do this loop?
+        for i in 0..self.0.len() {
```
rkuris, line 145: at a minimum:
```
for (node, i) in self.0.enumerate() {
```

Even better, use a peekable iterator, then you don't need `i` at all:
```
let mut it = self.0.iter().peekable();
while let Some(node) = it.next()  {
   // code for all elements
    if Some(next) = it.peek() {
        // ... code when there is another element in 'next'
    } else {
        // no more elements code
    }
}
```

rkuris (changes requested):

firewood/src/db.rs
danlaine: The generic is added back in https://github.com/ava-labs/firewood/pull/682/files. I could merge that PR into this one if you want.

firewood/src/merkle.rs
danlaine: Done

firewood/src/hashednode.rs
danlaine: change to use just the bounds

firewood/src/hashednode.rs
danlaine: Changed to `ValueDigestType`

firewood/src/proof.rs
danlaine: using peek now 👍

firewood/src/merkle.rs
danlaine: the seed is printed in the helper function that generates the key-value pairs. Changed it to check for the seed env var.

rkuris (approved):

PR #663, opened 2024-05-24
change `HashPreimage` children type

danlaine:
Is this cleaner? I'm not sure so I'm making a separate PR. I hate checking whether a hash is `default::Default()`. I'd rather see if it's `None`. That's why I think having an array of options might be better than having an option of an array.

@rkuris If you strongly disagree feel free to close.


rkuris (commented):
Seems okay. I think it doesn't read any simpler. What we really need is a non-zero operator on TrieHash which will let us use Option for free, as this change adds a byte to every hash, but that's arguably a different change.

danlaine:
closed in favor of https://github.com/ava-labs/firewood/pull/669

PR #668, opened 2024-06-13
Rkuris/move path iter item

rkuris:


.github/workflows/ci.yaml
```diff
@@ -37,9 +37,9 @@ jobs:
             ${{ runner.os }}-deps-${{ hashFiles('**/Cargo.toml') }}-
             ${{ runner.os }}-deps-
       - name: Check
-        run: cargo check --workspace --tests --examples --benches --all-features
```
rkuris, line 40: This might break the 'log' feature but we aren't using it ... yet.

PR #669, opened 2024-06-24
Consolidate children addresses / hashes

danlaine:
This PR combines the `children` and `child_hash` fields of a branch node into one field

### DIFF
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -310,6 +310,7 @@ impl<'a, K: Iterator<Item = u8> + Clone, V: AsRef<[u8]>> HashPreimage<'a, K, V> 
             .enumerate()
             .filter_map(|(i, child)| match child {
                 Child::AddressWithHash(_, hash) => Some((i, hash)),
+                Child::Address(_) => unreachable!("all children should have hashes"),
                 _ => None,
             });
 
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -220,7 +220,7 @@ impl<T: WriteLinearStore> HashedNodeStore<T> {
             .expect("index is a nibble")
         {
             Some((_, None)) => true,
-            None => true, // TODO danlaine: is this right?
+            None => unreachable!("parent must have node as a child"),
             _ => false,
         };
 
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -356,7 +356,7 @@ fn write_hash_preimage<H: HasUpdate>(node: &Node, path_prefix: &Path, buf: &mut 
             let mut child_hashes: [TrieHash; BranchNode::MAX_CHILDREN] = Default::default();
 
             for (i, child_addr) in branch.children.iter().enumerate() {
-                child_hashes[i] = match child_addr {
+                *child_hashes.get_mut(i).expect("arrays are same size") = match child_addr {
                     Some((_, Some(hash))) => hash.clone(),
                     _ => Default::default(),
                 };
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -129,6 +129,10 @@ impl<T: WriteLinearStore> HashedNodeStore<T> {
                         continue;
                     };
 
+                    if child_hash.is_some() {
+                        continue;
+                    }
+
                     // we found a child that needs hashing, so hash it
                     let mut child = self.take_node(*child_addr)?;
                     let original_length = path_prefix.len();
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -90,13 +90,15 @@ impl<T: ReadLinearStore> HashedNodeStore<T> {
                 })
                 .cloned();
             #[cfg(not(nightly))]
-            Ok(*self.root_hash.get_or_init(|| {
+            let hash = self.root_hash.get_or_init(|| {
                 let node = self
                     .nodestore
                     .read_node(addr)
                     .expect("TODO: use get_or_try_init once it's available");
+
                 hash_node(&node, &Path(Default::default()))
-            }))
+            });
+            Ok(hash.clone())
         } else {
             Ok(TrieHash::default())
         }
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -124,22 +124,26 @@ impl<T: WriteLinearStore> HashedNodeStore<T> {
                 // We found a branch, so find all the children that need hashing
                 let mut modified = false;
 
-                for (nibble, &child_addr, ref mut child_hash) in b
-                    .children
-                    .iter()
-                    .enumerate()
-                    .filter_map(|(i, c)| c.as_ref().map(|(addr, hash)| (i, addr, hash)))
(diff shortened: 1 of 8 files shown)


firewood/src/hashednode.rs
```diff
@@ -81,15 +81,15 @@ impl<T: ReadLinearStore> HashedNodeStore<T> {
     pub fn root_hash(&self) -> Result<TrieHash, Error> {
         debug_assert!(self.modified.is_empty());
         if let Some(addr) = self.nodestore.root_address() {
-            #[cfg(feature = "nightly")]
+            #[cfg(nightly)]
```
danlaine, line 84: Don't know why linting now fails without this change @rkuris ?

storage/src/node/branch.rs
```diff
@@ -50,45 +54,34 @@ impl BranchNode {
...
-            .expect("child_index is in bounds")
-    }
-
-    /// Returns the address of the child at the given index.
+    /// Returns the address of the child at the given index, and the hash,
+    /// which is None if we don't know the hash.
     /// None if there is no child at that index.
     /// Panics if `child_index` >= [BranchNode::MAX_CHILDREN].
-    pub fn child(&self, child_index: u8) -> Option<&LinearAddress> {
+    pub fn child(&self, child_index: u8) -> Option<(&LinearAddress, Option<&TrieHash>)> {
```
richardpringle, line 61: Would it be better to just return `Option<(LinearAddress, Option<TrieHash>)>`? 

`LinearAddress` is `Copy`, I'm assuming that `TrieHash` is also `Copy`.

firewood/src/hashednode.rs
```diff
@@ -308,28 +308,28 @@ impl HasUpdate for Vec<u8> {
...
+        let children_iter = self
+            .children
+            .iter()
+            .enumerate()
+            .filter_map(|(i, child_option)| match child_option {
+                Some((_, Some(hash))) => Some((i, hash)),
+                _ => None,
+            });
+
+        let num_children = children_iter.clone().count() as u64;
```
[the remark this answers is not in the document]
danlaine: > Yikes. Do you really need to do this?

We need the hash preimage format to be the same as merkledb, so I think we do.

I think the choice we have is either:
1. Iterate over `children_iter` twice by cloning it. This is the current behavior. Note that this iterates at most 16 times.
2. Iterate over `children_iter` once. For each child, increment a counter and write the child to a vec. Then, write the number of children to `buf` and then write the children vec to `buf`. Note we need to maintain this serialization format for compatibility with merkledb.

I figured (1) is probably more performant. Do you think that's right? Am I missing some other alternative?

> When is the buf a hash vs a Vec?

We write the preimage to a hasher when we're actually hashing a node. We write the preimage to a vec when verifying proofs.

firewood/src/merkle.rs
```diff
@@ -286,50 +286,60 @@ impl<T: ReadLinearStore> Merkle<T> {
...
+            Child::Address(addr) => (addr, None),
+            Child::AddressWithHash(addr, hash) => (addr, Some(hash)),
+        };
+
+        let node = self.read_node(*addr)?;
+
         match &*node {
             Node::Branch(b) => {
-                write!(writer, "  {addr}[label=\"@{addr}")?;
+                write!(writer, "  {addr}[label=\"@{addr} hash={hash:?}")?;
```
rkuris, line 303: Seems like if `Child` implements `Debug` we can move all this logic there

storage/src/node/branch.rs
```diff
@@ -6,6 +6,18 @@ use serde::{Deserialize, Serialize};
...
+pub enum Child {
+    #[default]
+    /// There is no child at this index.
+    None,
+    /// We know the child's address but not its hash.
+    Address(LinearAddress),
+    /// We know the child's address and hash.
+    AddressWithHash(LinearAddress, TrieHash),
+}
+
```
rkuris, line 20: If you do this, you can eliminate a lot of match statements elsewhere:
```suggestion
impl Child {
  fn address(&self) -> Option<LinearAddress> {
    match self {
        None => None,
        Address(addr) => Some(addr),
        AddressWithHash(addr, _) => Some(addr),
    }
  }
```

storage/src/trie_hash.rs
```diff
@@ -8,7 +8,7 @@ use sha2::digest::{generic_array::GenericArray, typenum};
 
 /// A hash value inside a merkle trie
 /// We use the same type as returned by sha2 here to avoid copies
-#[derive(PartialEq, Eq, Clone, Default)]
+#[derive(PartialEq, Eq, Clone, Default, Copy)]
```
rkuris, line 11: I don't think we want this. Hashes are big, and I'd rather be explicit about where they are being copied rather than telling the compiler we don't care about the cost of copying hashes.

firewood/src/merkle.rs
```diff
@@ -643,10 +640,9 @@ impl<T: WriteLinearStore> Merkle<T> {
                     //
                     // Note in the after diagram `child` may be the only child of `branch`.
                     let branch = BranchNode {
-                        children: branch.children,
+                        children: branch.children.clone(),
                         partial_path: branch.partial_path.clone(),
                         value: None,
-                        child_hashes: branch.child_hashes.clone(),
```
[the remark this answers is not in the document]
danlaine: An array of `TrieHash` which is not `Copy`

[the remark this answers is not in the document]
danlaine: I think because TrieHash isn't `Copy`

firewood/src/merkle.rs
```diff
@@ -386,10 +390,9 @@ impl<T: WriteLinearStore> Merkle<T> {
                     partial_path: Path::from(old_root_partial_path),
                 }),
                 Node::Branch(old_root) => Node::Branch(Box::new(BranchNode {
-                    children: old_root.children,
+                    children: old_root.children.clone(),
```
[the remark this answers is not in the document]
danlaine: > Was the type Copy before because the address is Copy?

Yes

> It doesn't look like old_root is used after this, can't you avoid the borrow at the start of this match statement?

I don't think we can move out of `old_root` (the first declaration) because it's an Arc

firewood/src/merkle.rs
danlaine: Done

firewood/src/hashednode.rs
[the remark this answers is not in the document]
danlaine: > In the case where it's a Vec, you can insert a 0 then modify with the proper count after, avoiding the clone.

Ah I missed this before. Yeah I think this is doable when `buf` is a vec. But we do sometimes want to write into a hasher rather than a vec. Not sure how to square that circle.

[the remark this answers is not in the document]
rkuris: I think it should stay as is. Cloning the iterator is cheap -- it just clones the state information.

firewood/src/hashednode.rs
rkuris: Note that this fails the stable clippy, since it checks for the invalid "nightly" cfg. Perhaps not a huge deal but we'll have to fix this again later probably.

rkuris (approved):

PR #671, opened 2024-06-28
Fix `root_hash` conditional compilation

danlaine:


firewood/src/hashednode.rs
```diff
@@ -80,29 +80,29 @@ impl<T: ReadLinearStore> HashedNodeStore<T> {
...
-                        .expect("TODO: use get_or_try_init once it's available");
-                    hash_node(&node, &Path(Default::default()))
-                })
-                .clone())
-        } else {
-            Ok(TrieHash::default())
-        }
+        let Some(addr) = self.nodestore.root_address() else {
+            return Ok(TrieHash::default());
+        };
```
rkuris, line 85: nit:
```suggestion
            Ok(TrieHash::default());
        } else {
            ...
        }
```
You can probably also factor out the `Ok`.

rkuris (approved):

firewood/src/hashednode.rs
danlaine: I think it's a bit cleaner as is so going to merge it. Lmk if you feel strongly otherwise

PR #678, opened 2024-07-10
refactor `HashedNodeStore.hash`

danlaine:
This fixes a regression. I also think this simplifies the code. 

`added` is a better name than `modified`.

The updated `hash` implementation now handles changed node sizes, which it couldn't before. This is sort of a hack because we shouldn't ever _need_ to increase the size allocated to a branch node _once we know the number of children_. That is, a branch node with more `AddressWithHash` children has a larger serialized representation than the same node with only `Address` children. This can be addressed in a follow up PR.

### DIFF
--- a/firewood/src/hashednode.rs
+++ b/firewood/src/hashednode.rs
@@ -5,7 +5,7 @@ use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use std::io::Error;
 use std::iter::{self, once};
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
 
 use storage::{Child, TrieHash, UpdateError};
 use storage::{LinearAddress, NodeStore};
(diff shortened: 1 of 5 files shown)


firewood/src/hashednode.rs
```diff
@@ -111,69 +96,81 @@ impl<T: ReadLinearStore> HashedNodeStore<T> {
...
         node_addr: LinearAddress,
-        node: &mut Node,
         path_prefix: &mut Path,
-    ) -> Result<TrieHash, Error> {
-        match node {
-            Node::Branch(b) => {
-                // We found a branch, so find all the children that need hashing
-                let mut modified = false;
+    ) -> Result<(TrieHash, LinearAddress), Error> {
+        let modified = self.added.remove(&node_addr);
```
rkuris, line 106: nit: I think this was easier to read with the external function that isolated the magic about into_inner.

rkuris (approved):

PR #709, opened 2024-08-22
Code cleanups

rkuris:
- Added a type alias for `Arc<NodeStore<Committed, FileBacked>>>1` which represents a committed version on disk and is used a lot in the revision manager
- Added a type alias for `Arc<NodeStore<ImmutableProposed, FileBacked>>` which represents a proposed revision, mostly on disk, but new nodes are in memory
- Added a unit test for attempting to commit a cloned proposal

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -25,6 +25,9 @@ pub struct RevisionManagerConfig {
     node_cache_size: NonZero<usize>,
 }
 
+type CommittedRevision = Arc<NodeStore<Committed, FileBacked>>;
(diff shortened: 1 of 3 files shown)


firewood/src/manager.rs
```diff
@@ -25,6 +25,9 @@ pub struct RevisionManagerConfig {
     node_cache_size: NonZero<usize>,
 }
 
+type CommittedRevision = Arc<NodeStore<Committed, FileBacked>>;
+type ProposedRevision = Arc<NodeStore<ImmutableProposal, FileBacked>>;
```
richardpringle, line 29: 👍

PR #712, opened 2024-08-26
Relax constaints around traits for `Arc`s

richardpringle:
Found some time to try this out. I think it's functionally the same (tests pass) and might even compile to the same assembly, but this code is more generic which should make it easier to change in the future

### DIFF
--- a/storage/src/nodestore.rs
+++ b/storage/src/nodestore.rs
@@ -50,6 +50,7 @@ use std::io::{Error, ErrorKind, Write};
 use std::iter::once;
 use std::mem::offset_of;
 use std::num::NonZeroU64;
+use std::ops::Deref;
 use std::sync::Arc;
 
 use crate::hashednode::hash_node;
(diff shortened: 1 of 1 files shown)


storage/src/nodestore.rs
```diff
@@ -587,56 +588,46 @@ pub trait HashedNodeReader: TrieReader {
...
-impl RootReader for &NodeStore<Committed, FileBacked> {
+impl<T> RootReader for T
+where
+    T: Deref,
+    T::Target: RootReader,
+{
     fn root_node(&self) -> Option<Arc<Node>> {
-        self.header
-            .root_address
-            .map(|addr| self.read_node_from_disk(addr).unwrap())
```
richardpringle, line 630: This might be wrong. If that's the case, you can just close this PR. If it's wrong though, that means you've found a bit of a hack for specialization which is *a little* (not a lot) problematic for maintainability IMO. Usually specialization should come in the form of newtype wrappers, where you would end up with something like this:
```rust
node_store.into_specialized().root_node()
```
Where the specific implementations for that one function would differ. Then again, this might be okay? I would have to spend more time with it to figure that out.

rkuris: Discussed in chat. It's fine, although we should consider removing the whole NodeReader trait since reading nodes is always done the same way now.

rkuris (approved):

PR #718, opened 2024-09-06
Use a boxed slice instead of a vec

rkuris:
Proposal keys and values can never change size, so we should make these boxes instead of vecs (avoids the capacity part)


samliok (approved):
good catch! every bit counts :)

PR #720, opened 2024-09-15
Rustify some code, removing a TODO

rkuris:


firewood/src/proof.rs
```diff
@@ -71,15 +71,19 @@ impl Hashable for ProofNode {
...
+                Some(Child::Node(_)) => unreachable!("TODO make unreachable"),
+                Some(Child::AddressWithHash(_, hash)) => Some(hash.clone()),
+            });
+            std::array::from_fn(|_| {
+                iter.next()
+                    .expect("there will always be the correct number of values")
+            })
+        } else {
+            Default::default()
+        };
```
richardpringle, line 86: I would write the whole `From` implementation like this:
```rust
impl From<PathIterItem> for ProofNode {
    fn from(item: PathIterItem) -> Self {
        let PathIterItem {
            key_nibbles: key,
            node,
            ..
        } = item;

        let (value, child_hashes) = match node.as_ref() {
            storage::Node::Branch(branch_node) => {
                let value = branch_node.value.clone();
                let mut iter = branch_node.children.iter().map(|child| {
                    child.as_ref().map(|child| match child {
                        Child::AddressWithHash(_, hash) => *hash,
                        Child::Node(_) => {
                            unreachable!("cannot create `ProofNode` from in-memory `Node`")
                        }
                    })
                });

                (
                    value,
                    std::array::from_fn(move |_| iter.next().expect("always the same size")),
                )
            }
            storage::Node::Leaf(leaf_node) => (Some(leaf_node.value.clone()), Default::default()),
        };

        let value_digest = value.map(ValueDigest::Value);

        Self {
            key,
            value_digest,
            child_hashes,
        }
    }
}
```

You need to derive `Copy` for `storage::TrieHash` for this to work, but you should do that anyway. There are a bunch of places where you could remove a call to `clone`. I'm not sure that you'll get an actual speed up as I would hope that the compiler is smart enough not to perform any allocations on `clone`.

I've also removed the `to_vec` call on value since it's already a boxed slice. 

You could further reduce allocations if you used `bytes::Bytes` for the `value` as I'm guessing it's rare that this changes. 

Is it ever the case the children array contains both a mix of `Child::Node` and `ChildAddressWithHash`? If not, I would recommend wrapping the array in an enum:
```rust
enum Children {
    WithHashes([(LinearAddress, TrieHash); SIZE]),
    Mixed([Child; SIZE]),
}
```

Then you could just use the [`map`](https://doc.rust-lang.org/stable/std/primitive.array.html#method.map) function for the `WithHashes` case since the `[(LinearAddress, TrieHash); SIZE]` type is `Copy`.

richardpringle (commented):
Left a comment with effectively the same changes as the one you have with a couple of other suggestions.

demosdemon:
This is ultra stale and no longer relevant.

PR #721, opened 2024-09-15
Use blake3 instead of sha256

rkuris:
Left sha256 as an option


rkuris:
Moving this to draft. We really want pluggable hashing.

PR #746, opened 2024-11-01
Add feature branch_factor_256

rkuris:
Support branch factor of 256. Removes all assumptions that branches contain 16 children, reimplements NibblesIterator for the larger branch factor to just return the bytes.

Removes some tests that don't make sense with 256 brandchs


rkuris:
Closes #747

PR #764, opened 2024-12-21
Improve fwdctl dump

zdf1230:
### Purpose

This change is to address issue #348.

### Detail

This added flags such as `--start-key`, `--stop-key`, `--max-key-count` and `--output-format` with `csv`, `json` as options and `stdout` as default behavior.

### Test

Tested locally & unit tests added


fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+        u8_to_string(value)
+    };
+    (key_str, value_str)
+}
+
+async fn handle_next_key(next_key: KeyFromStream) {
+    match next_key {
+        Some(Ok((key, _))) => {
+            println!(
+                "Next key is {0}, resume with \"--start-key={0}\".",
```
rkuris, line 150: This could be ugly for non-printable keys, which is the norm. Keys are often hashes, which contain lots of unprintable characters.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+        Some(Err(e)) => {
+            eprintln!("Error occurred while fetching the next key: {}", e);
+        }
+        None => {
+            println!("There is no next key. Data dump completed.");
+        }
+    }
+}
+
+trait OutputHandler {
```
rkuris, line 163: 🏆

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+    fn handle_record(&mut self, key: &[u8], value: &[u8]);
```
rkuris, line 164: This must handle errors better here rather than ignore them. This is likely to be important for things like running out of disk space when dumping the output file.

With the way the code is now, if you run out of disk space, no errors will be reported in some cases.

```suggestion
    fn handle_record(&mut self, key: &[u8], value: &[u8]) -> Result<(), std::io::Error>
```

Flush needs this too.

You can choose to use a `dyn std::error::Error` or just assume we only handle IO errors more gracefully.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+    writer: csv::Writer<File>,
+    hex: bool,
+}
+
+impl OutputHandler for CsvOutputHandler {
+    fn handle_record(&mut self, key: &[u8], value: &[u8]) {
+        let (key_str, value_str) = key_value_to_string(key, value, self.hex);
+        self.writer
+            .write_record(&[key_str, value_str])
+            .expect("Failed to write CSV record");
```
rkuris, line 178: You can map_error to an IO error or panic. A quick look at the possible errors here are IO or deserialization errors.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+
+struct JsonOutputHandler {
+    writer: BufWriter<File>,
+    hex: bool,
+    is_first: bool,
+}
+
+impl JsonOutputHandler {
+    fn new(mut writer: BufWriter<File>, hex: bool) -> JsonOutputHandler {
+        let _ = writer.write("{\n".as_bytes());
```
rkuris, line 194:
```suggestion
    writer.write(b"{\n")
```

There are a number of other places you can avoid calling `as_bytes()`

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+}
+
+impl OutputHandler for JsonOutputHandler {
+    fn handle_record(&mut self, key: &[u8], value: &[u8]) {
+        let (key_str, value_str) = key_value_to_string(key, value, self.hex);
+        if self.is_first {
+            let _ = self
+                .writer
+                .write(format!("  \"{}\": \"{}\"", key_str, value_str).as_bytes());
+            self.is_first = false;
```
rkuris, line 210: write!(self.writer, r#  "{key_str}": "{value_str}"#)

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+        } else {
+            let _ = self
+                .writer
+                .write(format!(",\n  \"{}\": \"{}\"", key_str, value_str).as_bytes());
```
rkuris, line 214: This code is nearly identical to the other case, so I would probably separate the `is_first` case to before printing:

```rust
if self.is_first {
    self.writer.write(b",\n")?;
    self.is_first = true;
}
```

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+        }
+        None => {
+            println!("There is no next key. Data dump completed.");
+        }
+    }
+}
+
+trait OutputHandler {
+    fn handle_record(&mut self, key: &[u8], value: &[u8]);
+    fn flush(&mut self);
```
rkuris, line 165: I debated implementing `flush` as a `Drop` for those implementations that require it, but that limits your error handling to panic.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+    fn flush(&mut self) {}
+}
+
+fn create_output_handler(
+    opts: &Options,
+) -> Result<Box<dyn OutputHandler + Send + Sync>, Box<dyn Error>> {
+    let hex = opts.hex;
+    match opts.output_format.as_str() {
+        "csv" => {
+            let file = File::create("dump.csv")?;
```
rkuris, line 244: This should not be hardcoded. Just write to stdout, or provide a way to specify the filename.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+    #[arg(
+        short = 'o',
+        long,
+        required = false,
+        value_name = "OUTPUT_FORMAT",
+        value_parser = ["csv", "json", "stdout"],
+        default_value = "stdout",
+        help = "Output format of database dump, default to stdout. CSV and JSON formats are available."
+    )]
+    pub output_format: String,
```
rkuris, line 74: Can you use an enum here and let clap handle the errors?

fwdctl/tests/cli.rs
```diff
@@ -210,6 +210,189 @@ fn fwdctl_dump() -> Result<()> {
     Ok(())
 }
 
+#[test]
+#[serial]
```
rkuris, line 214: 🏆

rkuris (changes requested):
This looks really good! A few things we need before landing:

- Error handling when writing is very important here. We can't drop IO errors, or we risk having a bad dump with no warning
- Continuation has to support hash key values, not just text. If the next key is binary, the lossy UTF-8 inputs won't work correctly. Maybe we need `start-key-hex`?
- Small bug in the stop key

Otherwise, this looks fantastic! Great test coverage!

The remaining comments are mostly style preferences and are not mandatory.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+    let mut output_handler = create_output_handler(opts).expect("Error creating output handler");
+
+    while let Some(item) = stream.next().await {
+        match item {
+            Ok((key, value)) => {
+                output_handler.handle_record(&key, &value);
+
+                key_count += 1;
+
+                if key == stop_key || key_count_exceeded(opts.max_key_count, key_count) {
```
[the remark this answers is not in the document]
zdf1230: Got it. This makes sense, stop key might not always exist.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
+
+    /// The output format of database dump.
+    /// Possible Values: ["csv", "json", "stdout"].
+    /// Defaults to "stdout"
+    #[arg(
+        short = 'o',
+        long,
+        required = false,
+        value_name = "OUTPUT_FORMAT",
+        value_parser = ["csv", "json", "stdout"],
```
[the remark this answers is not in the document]
zdf1230: Yes, stdout is the default behavior which would print outputs into stdout.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,230 @@ pub struct Options {
...
     }
+    output_handler.flush();
+
     Ok(())
 }
 
-fn u8_to_string(data: &[u8]) -> Cow<'_, str> {
-    String::from_utf8_lossy(data)
+const fn key_count_exceeded(max: Option<u32>, key_count: u32) -> bool {
+    matches!(max, Some(max) if key_count >= max)
```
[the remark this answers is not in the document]
zdf1230: agree, your version is easier to read

fwdctl/src/dump.rs
zdf1230: I think when declaring `value_parser` in clap, it would work just like using an Enum.

fwdctl/src/dump.rs
zdf1230: Very much agreed, adding error handling for this

fwdctl/src/dump.rs
zdf1230: changed here a little bit.

fwdctl/src/dump.rs
zdf1230: added error handling for this method as well, thanks for the advice.

fwdctl/src/dump.rs
zdf1230: Added a output-file-name option for opening up the choices and also print the name of the file in terminal.

fwdctl/src/dump.rs
zdf1230: added supported for start-key-hex and stop-key-hex

zdf1230:
> This looks really good! A few things we need before landing:
> 
> * Error handling when writing is very important here. We can't drop IO errors, or we risk having a bad dump with no warning
> * Continuation has to support hash key values, not just text. If the next key is binary, the lossy UTF-8 inputs won't work correctly. Maybe we need `start-key-hex`?
> * Small bug in the stop key
> 
> Otherwise, this looks fantastic! Great test coverage!
> 
> The remaining comments are mostly style preferences and are not mandatory.

Thanks for reviewing the code! I pushed some commits to improve. It should be ready to be reviewed now.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
+            println!("Dumping to {}", file_name);
+            let file = File::create(file_name)?;
+            Ok(Box::new(JsonOutputHandler {
+                writer: BufWriter::new(file),
+                hex,
+                is_first: true,
+            }))
+        }
+        "stdout" => Ok(Box::new(StdoutOutputHandler { hex })),
+        _ => Err("Unknown output format".into()),
```
rkuris, line 308: Isn't this `unreachable!()`?

rkuris (changes requested):
A few readability and code reduction changes. It was surprising to see an empty `stop_key` means there isn't one, when an empty `stop_key` is marginally valid. If a script generates an empty `stop_key` I would expect it to not output any data.

There were also some areas where performance could be improved, particularly for non-hex keys and values.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
+    pub stop_key: Option<Key>,
+
+    /// The key to start dumping from (if no key is provided, start from the beginning) in hex format.
+    /// Defaults to None.
+    #[arg(
+        long,
+        required = false,
+        conflicts_with = "start_key",
+        value_name = "START_KEY_HEX",
+        value_parser = key_parser,
```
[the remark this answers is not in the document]
zdf1230: > This should be `key_parser_text`, something with a different implementation. This avoids a lot of work you're doing to figure out what the key should be.
> 
> ```
> fn key_parser_hex(s: &str) -> Result<Box<[u8]>, std::io::Error> {
>     hex::decode(s)
>         .map(Vec::into_boxed_slice)
>         .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidInput, e))
> }
> ```

Good idea! forgot this clap feature which converted it back and forth.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
+    /// Output format must be set when the file name is set.
+    #[arg(
+        short = 'f',
+        long,
+        requires = "output_format",
+        value_name = "OUTPUT_FILE_NAME",
+        default_value = "dump",
+        help = "Output file name of database dump, default to dump. Output format must be set when the file name is set."
+    )]
+    pub output_file_name: String,
```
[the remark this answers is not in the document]
zdf1230: Yep, PathBuf works better here, forgive me that I am new to Rust.

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
         println!("Database is empty");
         return Ok(());
     };
     let latest_rev = db.revision(latest_hash).await?;
-    let start_key = opts.start_key.clone().unwrap_or(Box::new([]));
+
+    let start_key = opts
+        .start_key_hex
+        .clone()
+        .and_then(|start_key_hex| parse_hex_key(start_key_hex).ok())
```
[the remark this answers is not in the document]
zdf1230: Agree. changed it to be easier to read. Still using 2 times of clone, I am not so sure how clone the final result works.

I was trying to use the minimal clone approach, but I think this would still have readability issue.

```
let start_key = opts
    .start_key
    .as_ref()
    .or(opts.start_key_hex.as_ref())
    .cloned()
    .unwrap_or_default();
```

Chose the version below which is easier to read IMO.
```
let start_key = opts
        .start_key
        .clone()
        .or(opts.start_key_hex.clone())
        .unwrap_or_default();
```

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
+    let mut output_handler = create_output_handler(opts).expect("Error creating output handler");
+
+    while let Some(item) = stream.next().await {
+        match item {
+            Ok((key, value)) => {
+                output_handler.handle_record(&key, &value)?;
+
+                key_count += 1;
+
+                if (!stop_key.is_empty() && key >= stop_key)
```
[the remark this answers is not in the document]
zdf1230: Make sense use None to represent no stop_key

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
 
-fn u8_to_string(data: &[u8]) -> Cow<'_, str> {
-    String::from_utf8_lossy(data)
+fn key_count_exceeded(max: Option<u32>, key_count: u32) -> bool {
+    max.map(|max| key_count >= max).unwrap_or(false)
+}
+
+fn u8_to_string(data: &[u8]) -> String {
+    String::from_utf8_lossy(data).to_string()
 }
```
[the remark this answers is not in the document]
zdf1230: true, changed for avoid allocating memory

fwdctl/src/dump.rs
```diff
@@ -24,50 +28,283 @@ pub struct Options {
...
+
+fn parse_hex_key(hex_str: Box<[u8]>) -> Result<Key, Box<dyn Error>> {
+    hex::decode(String::from_utf8_lossy(&hex_str).as_bytes())
+        .map(|decoded| decoded.into_boxed_slice())
+        .map_err(|e| e.into())
+}
+
+// Helper function to convert key and value to a string
+fn key_value_to_string(key: &[u8], value: &[u8], hex: bool) -> (String, String) {
```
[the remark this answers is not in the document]
zdf1230: yep, keep as String

fwdctl/src/dump.rs
```diff
@@ -24,50 +30,277 @@ pub struct Options {
...
         return Ok(());
     };
     let latest_rev = db.revision(latest_hash).await?;
-    let start_key = opts.start_key.clone().unwrap_or(Box::new([]));
+
+    let start_key = opts
+        .start_key
+        .clone()
+        .or(opts.start_key_hex.clone())
+        .unwrap_or_default();
```
rkuris, line 134: nit: the clones here don't seem to be required
```suggestion
let start_key: Box<[u8]> = opts
        .start_key
        .or(opts.start_key_hex)
        .unwrap_or_default();
```

rkuris (approved):

fwdctl/src/dump.rs
```diff
@@ -24,50 +30,277 @@ pub struct Options {
...
+    let mut output_handler = create_output_handler(opts).expect("Error creating output handler");
+
+    while let Some(item) = stream.next().await {
+        match item {
+            Ok((key, value)) => {
+                output_handler.handle_record(&key, &value)?;
+
+                key_count += 1;
+
+                if (stop_key.as_ref().is_some_and(|stop_key| key >= *stop_key))
```
rkuris, line 148: nit: preference is:
```suggestion
                if stop_key.map(|stop_key| key >= stop_key).unwrap_or(false)
```

rkuris (approved):

rkuris:
This repository requires that you sign your commits. Can you set up a GPG key and sign them? Once that's done, I'll merge this.

Some instructions on how to sign your old commits can be found here: https://superuser.com/questions/397149/can-you-gpg-sign-old-commits#

fwdctl/src/dump.rs
rkuris: Well, not quite, but you could use a &Box<[u8]> and just have start_key point to the right thing. You'd have to set up an empty thing to point to for the default that outlives it though, which is a lot of work just to avoid a single clone at the start of the program.

zdf1230: yeah, I tried and found out it is not easy

fwdctl/src/dump.rs
zdf1230: Well, this is in a loop so move value is prohibited here.
I have to do something like this `stop_key.clone().map(|stop_key| key >= stop_key).unwrap_or(false)` to make it work.

That is the reason I chose to use ref here. And I found `is_some_and ` is quite straight forward, so keeping it as is. Thanks for the suggestion though

zdf1230:
Done. I think all the commits are showed up as verified. Please let me know if there is anything I can do. Thanks!

zdf1230:
Looks it the change was hanging there waiting for merging status for a long time. And I could be related to https://www.githubstatus.com/incidents/78wybrzyv0wf

Get a small change to refresh the commits and re-sign all the commits once again. (Looks like rebase feature on Github page cannot sign the commit for me)

PR #778, opened 2025-02-08
io_uring implementation

rkuris:
Only enabled on linux, seems to run faster even in small test cases


rkuris:
System calls reduced to 1 call per 16 writes. Further improvements could be:
 - Serialize using thread pool
 - Increase the count to 32
 - Optimize the preallocated buffers for serialization based on the node type and/or data length

Issue #358, opened 2023-11-18
Cache management: don't blow out cache with range proofs


rkuris:
Consider using a weighted LRU cache, such as https://docs.rs/clru/latest/clru/ or having an alternate cow cache for iterating (probably better).

rkuris:
Related to #789

demosdemon:
Reopen/create new issue once a more specific idea is around

PR #781, opened 2025-02-09
cache node reads + use cache with lfu admit, lru evict

darioush:
related to: https://github.com/ava-labs/firewood/issues/358

Uses mini_moka cache crate:
- allows concurrent access (explicit locks not needed)
- has lfu admission + lru eviction, should work for range proof reads and normal reads as range reads would not be admitted
- adding caching for node reads

(originally tried moka but mini_moka seems fine and has fewer dependencies)

### DIFF
--- a/storage/src/nodestore.rs
+++ b/storage/src/nodestore.rs
@@ -225,6 +225,14 @@ impl<T: ReadInMemoryNode, S: ReadableStorage> NodeStore<T, S> {
         let node = Node::from_reader(area_stream)?;
         Ok(node.into())
     }
+
+    /// Put a [Node] into the cache at the given address.
+    fn cache_node(&self, addr: LinearAddress, node: Arc<Node>) -> Result<Arc<Node>, Error> {
+        self.storage
+            .write_cached_nodes(once((&addr, &node)))
+            .expect("failed to write cached nodes");
+        Ok(node)
+    }
 }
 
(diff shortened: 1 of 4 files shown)


storage/src/nodestore.rs
```diff
@@ -225,6 +225,12 @@ impl<T: ReadInMemoryNode, S: ReadableStorage> NodeStore<T, S> {
         let node = Node::from_reader(area_stream)?;
         Ok(node.into())
     }
+
+    /// Put a [Node] into the cache at the given address.
+    fn cache_node(&self, addr: LinearAddress, node: Arc<Node>) -> Result<Arc<Node>, Error> {
+        self.storage.write_cached_nodes(once((&addr, &node)))?;
```
rkuris, line 231: Using `once` here is likely a performance hit. It has to set up an iterator and then iterate over them in `write_cached_nodes`. Since all it's really doing is calling `self.cache.insert` once, I'd suggest another method on the storage interface that just writes a single node.

The code was structured this way because it was more efficient to grab the lock once, but since that isn't the case any more and you can't write a block of data to the cache using the same lock, you could consider completely removing `write_cached_nodes` and just have a `write_cached_node` method everywhere.

My suspicion is that the locking that has to happen inside moka is complex and we win here because we only do the lock once.

darioush:
going to close this for now since we took an alternative approach.

PR #789, opened 2025-02-11
Allow for caching of reads

rkuris:
This adds a parameter to the open database builder that allows for caching reads. We haven't been caching reads and have had good performance but some use cases may prefer to cache reads.


darioush (approved):
nice, can we pass this through a command line option in benchmarking as well?

rkuris:
> nice, can we pass this through a command line option in benchmarking as well?

Yes, I'll do that under a different PR.

PR #792, opened 2025-02-13
Add periodic metric reporting to the ffi

rkuris:
Collect metrics and provide a prometheus-compatible web service when you open the database.

By default, this creates a server that listens on port 3000, which is what we're expecting for prometheus, so it just works.

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -17,12 +17,12 @@ type Firewood struct {
 
 // openConfig is used to configure the database at open time
 type openConfig struct {
-	path                  string
-	nodeCacheEntries      uintptr
-	revisions             uintptr
-	readCacheStrategy     uint8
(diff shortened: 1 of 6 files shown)


ffi/firewood.go
```diff
@@ -17,11 +17,12 @@ type Firewood struct {
...
-	nodeCacheEntries  uintptr
-	revisions         uintptr
-	readCacheStrategy int8
-	create            bool
+	path                  string
+	nodeCacheEntries      uintptr
+	revisions             uintptr
+	readCacheStrategy     uint8
+	create                bool
+	metrics_interval_secs uint32
```
darioush, line 25: nit: I think the go convention is `metricsIntervalSecs`

darioush (approved):

PR #808, opened 2025-02-27
check against eth hash

darioush:


darioush:
closing, the relevant commits was already added to 807

PR #810, opened 2025-03-04
refactor!(ffi): overhaul Go bindings

ARR4N:
While reading the code to understand the wrapping I thought I'd bring it in line with more idiomatic Go. That ballooned into a broader refactoring:

### General

1. Both `fwd_create_db()` and `fwd_open_db()` accept the same struct as a single argument. This (a) simplifies the Go code by allowing reuse and (b) provides a means of assessing correctness at the call site by confirming that variables are assigned to their respective fields.
2. Management of memory pinning is encapsulated within a `valueFactory`, the constructor of which returns a `cleanup()` function to unpin. This is a common Go idiom and dropping the cleanup would be very obvious as the underscore in `values, _ := newValueFactory()` is a red flag to Gophers.
3. All uses of a returned `C.struct_Value` followed the same pattern of extracting a `[]byte` and then freeing, so this is abstracted, which simplifies usage sites.

### Testing

1. Added CI step to run Go test.
2. Ensured that at least `GODEBUG=cgocheck=1` is enabled to confirm that memory passed to C is pinned, but note that CI runs with `GOEXPERIMENT=cgocheck2` for more comprehensive tests. The latter has to be set at compile time while the former can be overridden in the test setup.
3. The existing `TestInsert100` now runs via both `Batch()` and `Update()` as they're effectively the same method.
4. Where appropriate, I coupled values by reusing the same variables. This is primarily for readability as it makes the test logic "obviously" correct instead of having to cross-reference inputs and outputs. 

### Style

1. All external references will be fully qualified as `firewood.Database` instead of `firewood.Firewood`.
2. Constructors when only one type exists are typically not qualified by name, so it's now just `New()`.
6. A `Config` struct is simpler than the variadic options and also allows a user to access default values; input checks are now performed in the constructor and return an error instead of panicking.
7. Documentation renders properly if you run `godoc` within the module (or if viewed on https://pkg.go.dev in the future).
8. Other stylistic changes can be inferred from the diffs.


ffi/firewood.go
```diff
@@ -1,179 +1,186 @@
...
-	ffi_ops := make([]C.struct_KeyValue, len(ops))
+// Batch applies a batch of updates to the database, returning the hash of the
+// root node after the batch is applied.
+//
+// NOTE that if the `Value` is empty, the respective `Key` will be deleted as a
+// prefix deletion; i.e. all children will be deleted.
+//
+// WARNING: a consequence of prefix deletion is that calling Batch with an empty
+// key and value will delete the entire database.
+func (db *Database) Batch(ops []KeyValue) []byte {
```
ARR4N, line 103: Do we want to return a `[32]byte` instead here?

ffi/firewood.go
```diff
@@ -1,179 +1,186 @@
+// Package firewood provides a Go wrapper around the [Firewood] database.
+//
+// [Firewood]: https://github.com/ava-labs/firewood
 package firewood
 
-// #cgo LDFLAGS: -L../target/release -L/usr/local/lib -lfirewood_ffi -lm
+// #cgo LDFLAGS: -L${SRCDIR}/../target/release -L/usr/local/lib -lfirewood_ffi -lm
```
rkuris, line 6: Please add a comment as to why we need -lm here to deter someone else from thinking about removing it. It's required on linux.

rkuris (commented):
Pleasure to review this code! Please address the very few comments and let's get this merged in.

ffi/firewood.go
```diff
@@ -1,179 +1,186 @@
...
-		return C.struct_Value{0, nil}
+// Get retrieves the value for the given key. It always returns a nil error.
+func (db *Database) Get(key []byte) ([]byte, error) {
+	values, cleanup := newValueFactory()
+	defer cleanup()
+
+	val := C.fwd_get(db.handle, values.from(key))
+	switch buf := extractBytesThenFree(&val); len(buf) {
+	case 0:
+		// TODO(arr4n): discuss with Firewood team if this is necessary.
```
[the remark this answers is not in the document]
ARR4N: Perfect, that's exactly what I meant. The original code differentiated the two, returning `nil` instead of an empty slice, but I now just `return extractBytesThenFree(&val), nil`.

ffi/firewood_test.go
```diff
@@ -1,106 +1,174 @@
...
+		case "cgocheck=0":
+			fmt.Fprint(os.Stderr, "GODEBUG=cgocheck=0; MUST be 1 for Firewood cgo tests")
+			os.Exit(1)
+		}
+	}
+
+	if !hasCgoCheck {
+		debug = append(debug, "cgocheck=1")
+	}
+	os.Setenv("GODEBUG", strings.Join(debug, ","))
```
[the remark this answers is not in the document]
ARR4N: > can we move this inside the "!hasCgoCheck" block, so it only calls Setenv if it changes?

Done

> Is it possible to reuse https://pkg.go.dev/internal/godebug instead of parsing GODEBUG yourself?

Unfortunately not as it's in `/internal/`, which limits its visibility to the Go stdlib.

> since os.Setenv is really dangerous (it's not thread safe, and has a few other problems)

Out of interest, what are the other problems? FWIW at this point in the code there's only 1 goroutine.

ffi/firewood.go
ARR4N: Done

ARR4N:
> Pleasure to review this code!

Thank you :)

rkuris (approved):

PR #825, opened 2025-03-25
test(ffi/tests): basic eth compatibility

darioush:
This pull request introduces a new CI job for testing Ethereum hash compatibility and adds a corresponding Go test file along with its dependencies. The most important changes include the addition of a new job in the CI workflow, the creation of a new test file, and updates to the `go.mod` file to include necessary dependencies.


ffi/tests/go.mod
```diff
@@ -0,0 +1,60 @@
+module github.com/ava-labs/firewood/ffi/tests
+
+go 1.23
+
+toolchain go1.23.6
+
+require (
+	github.com/ava-labs/firewood/ffi/v2 v2.0.0-20250218224920-ad24ac0dbc52
```
rkuris, line 8: How does this work? Why is this always "v2.0.0-20250218224920-ad24ac0dbc52"? Isn't that related to the commit hash?

Or is this then overridden by the `replace` line at the end of the file?

rkuris (approved):

ffi/tests/go.mod
darioush: this gets overridden by the replace at the end of the file. let me see if I can make this sthg else that's like a placeholder.

PR #839, opened 2025-04-12
Framework for change proofs

rkuris:
Adds some additional methods for iterating the trie. Still needed:
 - Some way for the iterator to skip identical branches
 - Complete the traversal algorithm


firewood/src/diff.rs
```diff
@@ -0,0 +1,1744 @@
...
+}
+
+impl fmt::Debug for VisitedPairState {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("VisitedPairState")
+            .field("children_iter_left", &"<iterator>")
+            .field("children_iter_right", &"<iterator>")
+            .finish()
+    }
+}
```
demosdemon: We can use [`derivative`](https://mcarton.github.io/rust-derivative/latest/index.html) to clean up these manually implemented `Debug`s.

rkuris:
Group review comments:
 - Can eliminate the boilerplate code for the debug implementations with a crate @demosdemon knows about
 - Should better document each of the types, especially around understanding how you get in the state. Diagram?
 - The giant match statement is too hard cognitively to see the algorithm, so let's add more helper methods on DiffIterationState

rkuris:
This is a dead branch, replaced by work from @bernard-avalabs

PR #848, opened 2025-04-17
Update prereqs to include cargo and make

aaronbuchwald:
This PR documents additional prerequisites to building Firewood after building on a fresh machine and realizing we depend on them.

I ran into the following dependencies (including `protobuf` which was already documented):

- protobuf
- make
- cargo (obvious for a Rust project)
- C compiler

I wasn't sure the best way to document "depends on some C compiler" that fit into the existing style and it's relatively rare to be missing this, so I skipped it for now.


rkuris (approved):
Could also reference setup-scripts/build-environment.sh in the benchmark directory

aaronbuchwald:
> Could also reference setup-scripts/build-environment.sh in the benchmark directory

Feels like benchmark setup should be separate from pure build instructions. Can make a separate change for benchmarking instructions.

Issue #854, opened 2025-04-21
[ffi] Improve error handling for ffi layer


rkuris:
Currently, errors cause panics. We should figure out a way to map the rust errors into golang errors so that it doesn't just crash if something doesn't work right.

rkuris:
We probably need to return a struct that has (return value, error) where the error is either a pointer to a string or nil.

rkuris:
I think this is done, @alarso16 can you confirm?

rkuris:
Only creating the database can panic at this point

alarso16:
Finished!

PR #857, opened 2025-04-22
Error improvements

rkuris:
Handle errors during database creation and open

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -100,7 +100,7 @@ struct Value fwd_batch(const struct DatabaseHandle *db,
  *
  * # Arguments
  *
- * * `db` - The database handle to close, previously returned from a call to open_db()
+ * * `db` - The database handle to close, previously returned from a call to `open_db()`
  */
 void fwd_close_db(struct DatabaseHandle *db);
 
@@ -149,7 +149,7 @@ const struct DatabaseHandle *fwd_create_db(struct CreateOrOpenArgs args);
 
 /**
  * Drops a proposal from the database.
- * The propopsal's data is now inaccessible, and can be freed by the RevisionManager.
+ * The propopsal's data is now inaccessible, and can be freed by the `RevisionManager`.
  *
  * # Arguments
  *
@@ -176,6 +176,9 @@ struct Value fwd_drop_proposal(const struct DatabaseHandle *db, uint32_t proposa
  * This function is unsafe because it dereferences raw pointers.
  * The caller must ensure that `value` is a valid pointer.
  *
+ * # Panics
+ *
+ * This function panics if `value` is `null`.
  */
 void fwd_free_value(const struct Value *value);
 
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -500,8 +500,7 @@ fn commit(db: *const DatabaseHandle, proposal_id: u32) -> Result<(), String> {
 ///
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn fwd_drop_proposal(db: *const DatabaseHandle, proposal_id: u32) -> Value {
-    drop_proposal(db, proposal_id)
-        .map_or_else(Into::into, Into::into)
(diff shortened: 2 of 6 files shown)


rkuris:
Related to #854

alarso16 (approved):
We could still use this. I think having this sort of "uninformative" error-handling on creation is ok, but we could also create a struct that is essentially a `Value` plus `void *` type pointer if we wanted to return it. I don't think this is necessary

ffi/src/lib.rs
```diff
@@ -185,15 +186,9 @@ impl Value {
...
-impl From<&[u8]> for Value {
-    fn from(data: &[u8]) -> Self {
-        let boxed: Box<[u8]> = data.into();
-        boxed.into()
-    }
-}
-
-impl From<Box<[u8]>> for Value {
-    fn from(data: Box<[u8]>) -> Self {
+impl<T: Into<Box<[u8]>>> From<T> for Value {
```
alarso16, line 189: I don't think we can have this generic `impl` statement now that `Value` is multi-purpose

ffi/src/lib.rs
```diff
@@ -274,13 +269,16 @@ pub struct CreateOrOpenArgs {
 ///
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn fwd_create_db(args: CreateOrOpenArgs) -> *mut Db {
+    let manager_config = match manager_config(args.cache_size, args.revisions, args.strategy) {
```
alarso16, line 272: In theory,  it would be best if we could pass the EVM or AvaGo logger in here, but that sounds really complicated...

ffi/src/lib.rs
```diff
@@ -304,13 +302,16 @@ pub unsafe extern "C" fn fwd_create_db(args: CreateOrOpenArgs) -> *mut Db {
 ///
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn fwd_open_db(args: CreateOrOpenArgs) -> *mut Db {
+    let manager_config = match manager_config(args.cache_size, args.revisions, args.strategy) {
+        Ok(manager_config) => manager_config,
+        Err(_e) => {
```
alarso16, line 307: Why lead with an underscore?

alarso16 (commented):
We could still use this. I think having this sort of "uninformative" error-handling on creation is great for startup

alarso16 (approved):
Adding the warnings made some other issues in the code more obvious

ffi/firewood.h
```diff
@@ -100,7 +100,7 @@ struct Value fwd_batch(const struct DatabaseHandle *db,
  *
  * # Arguments
  *
- * * `db` - The database handle to close, previously returned from a call to open_db()
+ * * `db` - The database handle to close, previously returned from a call to `open_db()`
```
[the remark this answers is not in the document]
rkuris: Definitely a different PR. This doesn't change any interfaces, and I'm not sure I like that change anyway. Discussion on slack.

ffi/firewood.h
```diff
@@ -176,6 +176,9 @@ struct Value fwd_drop_proposal(const struct DatabaseHandle *db, uint32_t proposa
  * This function is unsafe because it dereferences raw pointers.
  * The caller must ensure that `value` is a valid pointer.
  *
+ * # Panics
+ *
+ * This function panics if `value` is `null`.
```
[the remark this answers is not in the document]
rkuris: Yeah I think it's fine, just added the docs.

ffi/src/lib.rs
rkuris: This supresses unused warnings.

ffi/src/lib.rs
rkuris: We shouldn't need to do that here. One thing we *could* do is somehow pass the rust logger messages back to those loggers, but I think that's a bigger project.

ffi/src/lib.rs
rkuris: Removed

PR #860, opened 2025-04-29
feat(ffi): ffi error messages

alarso16:
Addresses issue #854. By wrapping the return values from each function with an additional C string allocated on the heap, we can provide a more detailed error if issues are encountered in Rust, and avoid panicking on many invalid accesses.

This is a breaking change for all consumers of the ffi, as much of the Golang interface now returns an `error`, in addition to the previous return types. For functions that did return an error, the behavior has not changed.

This was tested using existing unit tests. In general, these errors should not be encountered (e.g. IO errors), but testing the case that `db.is_null()` is used to ensure that error cases can be handled. This should still be further tested to ensure there are no memory leaks.


ffi/firewood.go
```diff
@@ -126,10 +127,15 @@ func (db *Database) Batch(ops []KeyValue) []byte {
...
+// Generates error if the error term is nonnull.
+func extractBytesThenFree(v *C.struct_ReturnValue) (buf []byte, err error) {
+	buf = C.GoBytes(unsafe.Pointer(v.data), C.int(v.len))
+	if v.err != nil {
+		errStr := C.GoString(v.err)
+		err = fmt.Errorf("internal firewood error: %s", errStr)
+	}
 	C.fwd_free_value(v)
-	return buf
+	return
```
tsachiherman, line 138: I'm not certain if that's strictly needed, but I would add
```golang
runtime.KeepAlive(v)
```
after the `C.fwd_free_value(v)` in order to ensure that go won't reclaim `v`'s memory during the execution of `fwd_free_value`.

alarso16: Is it better practice to use a `pinner`?

alarso16: Actually, I also think it's safe because we use `C.GoBytes` to ensure that Go doesn't retain the pointer. If it would be better to more explicitly track it, we can add it.

ffi/src/lib.rs
```diff
@@ -162,18 +266,22 @@ impl From<Box<[u8]>> for Value {
 /// This function is unsafe because it dereferences raw pointers.
 /// The caller must ensure that `value` is a valid pointer.
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
-    let value = unsafe { &*value as &Value };
-    if value.len == 0 {
-        return;
+pub unsafe extern "C" fn fwd_free_value(value: *const ReturnValue) {
```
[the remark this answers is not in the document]
alarso16: I don't think we can because we pass it by reference. @rkuris is this correct?

ffi/firewood_test.go
```diff
@@ -95,7 +96,9 @@ func TestGetBadHandle(t *testing.T) {
 	require.Error(t, err)
 
 	err = db.Close()
-	assert.Error(t, err)
+	if !errors.Is(err, dbClosedErr) {
```
tsachiherman, line 99: require.ErrorIs(t, err, dbClosedError)

ffi/firewood.go
```diff
@@ -138,22 +144,32 @@ func (db *Database) Get(key []byte) ([]byte, error) {
...
+	return bytes, nil // Ignore errors.
 }
 
-// Close closes the database and releases all held resources. It always returns
-// nil.
+// Close closes the database and releases all held resources.
+// Returns an error if already closed.
 func (db *Database) Close() error {
+	if db.handle == nil {
+		return errors.New("database already closed")
```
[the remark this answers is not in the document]
alarso16: Fixed! It would be nice if we can do this across the FFI boundary by importing constant strings or something

ffi/src/lib.rs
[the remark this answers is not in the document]
tsachiherman: I was hoping that it would be either here or on the go side. While this is not a showstopper ( since the variable is going out of scope ), it's usually better to sanitize dangling pointers

ffi/firewood.go
tsachiherman: isn't C.GoBytes create a copy of the v.data ? if so, there are no guarantees about v after that point.

alarso16: Ahh I see. I'll add it, even if it's not necessary

tsachiherman (approved):
lgtm

ffi/src/lib.rs
```diff
@@ -130,7 +186,16 @@ pub unsafe extern "C" fn fwd_root_hash(db: *mut Db) -> Value {
...
+        Ok(Some(root)) => root,
+        Ok(None) => {
+            return unsafe { make_err_return("couldn't find root hash for db") };
+        }
+        Err(e) => {
+            return unsafe { make_err_return(&format!("couldn't get root hash: {}", e)) };
+        }
+    };
+
     Value::from(root.as_slice())
```
rkuris, line 199: It's best to avoid using `return` in rust. You can just use the value of the last expression, in this case it's a match expression:

```suggestion
        Ok(Some(root)) => root.into(),
        Ok(None) => "couldn't find root hash for db".into(),
        // I probably would skip adding context here and just use e.into(),
        Err(e) => (&format!("couldn't get root hash: {}", e)).into(),
        }
    }

```

You might even be able to do better than this if you have the right conversions set up. This would need a conversion from Error to Value, and from &str to Value, as shown in the other comments.
```
db.root_hash_sync()?.unwrap_or_else(|| "no root hash".into())

File: ffi/src/lib.rs [hunk 6]
```diff
@@ -164,16 +229,19 @@ impl From<Box<[u8]>> for Value {
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
     let value = unsafe { &*value as &Value };
```
rkuris, line 231: It's not clear to me why this is safe or why we need to do this. Can you add a comment?

ffi/src/lib.rs
```diff
@@ -164,16 +229,19 @@ impl From<Box<[u8]>> for Value {
...
+    if value.len > 0 {
+        let recreated_box = unsafe {
+            Box::from_raw(std::slice::from_raw_parts_mut(
+                value.data as *mut u8,
+                value.len,
+            ))
+        };
+        drop(recreated_box);
+    } else {
+        let cstr = unsafe { CString::from_raw(value.data as *mut c_char) };
```
rkuris, line 242: Keep the unsafe as small as possible:

```suggestion
        let raw_string_data = value.data as *mut c_char;
        let cstr = unsafe (CString::from_raw(raw_string_data);
```

ffi/src/lib.rs
```diff
@@ -302,5 +370,25 @@ fn manager_config(cache_size: usize, revisions: usize, strategy: u8) -> Revision
 /// * `db` - The database handle to close, previously returned from a call to open_db()
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn fwd_close_db(db: *mut Db) {
+    if db.is_null() {
+        return;
```
rkuris, line 374: nit: should we return an error "already closed" maybe?

ffi/src/lib.rs
```diff
@@ -302,5 +370,25 @@ fn manager_config(cache_size: usize, revisions: usize, strategy: u8) -> Revision
...
+    }
     let _ = unsafe { Box::from_raw(db) };
 }
+
+/// Creates a C string from a Rust string.
+/// The caller is responsible for freeing the memory associated with the returned C string.
+/// Returns a null pointer if the string contains a null byte.
+unsafe fn create_cstr(err: &str) -> *const c_char {
```
rkuris, line 382: I am pretty sure this unsafe block is larger than it needs to be (but please comment if I am wrong).

ffi/src/lib.rs
```diff
@@ -302,5 +370,25 @@ fn manager_config(cache_size: usize, revisions: usize, strategy: u8) -> Revision
...
+    match CString::new(String::from(err)) {
+        Ok(cstr) => cstr.into_raw(), // leaks the CString (caller must free)
+        Err(_) => std::ptr::null(),  // contains null byte, invalid C string
+    }
```
rkuris, line 386:
```suggestion
    // convert the string into a c_str. If there are null characters
    // in it, we silently drop the error and return a null pointer
    // TODO: Maybe we should return an allocated "error string could not be rendered" instead?
    CString::new(err.into()).unwrap_or_else(|| std::ptr::null())
```

rkuris (changes requested):

ffi/firewood.go
```diff
@@ -138,22 +149,32 @@ func (db *Database) Get(key []byte) ([]byte, error) {
 	values, cleanup := newValueFactory()
 	defer cleanup()
 	val := C.fwd_get(db.handle, values.from(key))
-	bytes := extractBytesThenFree(&val)
-	if len(bytes) == 0 {
+	bytes, err := extractBytesThenFree(&val)
+
+	// Ignore the error.
+	if err != nil {
```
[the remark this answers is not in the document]
alarso16: Maybe for clarity I will return `bytes, nil`, but I didn't want to change the behavior of the function (which is to never return an error). This may have effects on the benchmarking

ffi/src/lib.rs
```diff
@@ -302,5 +370,25 @@ fn manager_config(cache_size: usize, revisions: usize, strategy: u8) -> Revision
...
+}
+
+unsafe fn make_err_return(err: &str) -> Value {
```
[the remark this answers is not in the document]
rkuris: I also think the `create_cstr` function doesn't need to be unsafe either. Didn't notice that until I reviewed it.

ffi/firewood.go
```diff
@@ -138,22 +149,32 @@ func (db *Database) Get(key []byte) ([]byte, error) {
...
-func (db *Database) Root() []byte {
+func (db *Database) Root() ([]byte, error) {
 	hash := C.fwd_root_hash(db.handle)
-	return extractBytesThenFree(&hash)
+	bytes, err := extractBytesThenFree(&hash)
+	// On error, send empty root hash.
+	if err != nil {
+		bytes = make([]byte, 32)
+	}
+	return bytes, nil // Ignore errors.
```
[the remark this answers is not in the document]
alarso16: The real issue is that there's now way of referencing the exact error. We should find some way of using constant strings for errors in Rust, and Go can identify them.

ffi/src/lib.rs
alarso16: I don't know why - I didn't write it. If I had to guess, it's rescoping and providing ownership

ffi/src/lib.rs
alarso16: This is handled in Go - might be better to just remove altogether

ffi/firewood_test.go
```diff
@@ -79,6 +79,25 @@ func TestGetNonExistent(t *testing.T) {
...
+func TestGetBadHandle(t *testing.T) {
+	db := &Database{handle: nil}
+
+	// This ignores error, but still shouldn't panic.
+	_, _ = db.Get([]byte("non-existent"))
+
+	// We ignore the error, but it shouldn't panic.
+	_, _ = db.Root()
+
+	root, err := db.Update([][]byte{[]byte("key")}, [][]byte{[]byte("value")})
```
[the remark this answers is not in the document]
alarso16: Simplified a little bit

ffi/src/lib.rs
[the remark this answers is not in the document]
alarso16: I left it in another comment, but I don't think you can make `From<&str>`. The compiler warned that "the size is not known at compile time"

ffi/src/lib.rs
alarso16: Added comment

ffi/firewood.go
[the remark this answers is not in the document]
rkuris: SGTM. We should add this info to the source ticket or create a new one.

It might be particularly challenging to make the errors a constant string. They often have additional context data in them. One thing we CAN do is make the "usual" errors more readable, or don't treat them like an error. For example, not found should be "nil, nil" which means it's not there and there is no error. I can't see other errors being handled special in go, so just having them be a string from the API perspective seems fine.

Testing will be a bit tougher here but you can use "contains" or something perhaps if you're testing specific errors, or just make sure it is an error and not worry about the contents of it.

ffi/src/lib.rs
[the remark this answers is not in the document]
rkuris: `From<&str>` does indeed work. Sample program: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a4ccf2a16e3aaede2f0c6b7372ab9dc1

Perhaps you used `From<str>` which will not work, because the size of `str` is not known at compile time.

ffi/src/lib.rs
rkuris: Correct, you cannot, but you can `From<&str>`. `str` by itself has no length. This pattern is the same for several other types in rust `dyn Error` is another example. When you create a reference to something like that, the length information is stored in the fat pointer.

ffi/src/lib.rs
rkuris: Ah yes, you didn't add it, sorry. What I think it's doing is dereferencing the pointer and then making a rust reference out of it. It's unsafe because in theory someone could have a mutable reference to that same object, resulting in UB, or the pointer could be nil, which we aren't checking.

ffi/src/lib.rs
rkuris: SGTM, but we should do it in one place consistently.

ffi/src/lib.rs
```diff
@@ -155,6 +211,16 @@ impl From<Box<[u8]>> for Value {
     }
 }
 
+impl From<String> for Value {
```
rkuris, line 214: Use `From<&str>` (not `From<str>`). You don't need an owned string to convert it to a value, especially since you'll be coping the data out of it anyway.

See See https://doc.rust-lang.org/book/ch20-03-advanced-types.html#dynamically-sized-types-and-the-sized-trait for more info on unsized types, and in particular `str`.

rkuris (changes requested):
There's a little more work to do to make this less go-like and more rusty.

ffi/firewood.go
tsachiherman: The issue is that `v` is a go object, and we're passing a pointer to it to `fwd_free_value`.
The `fwd_free_value` code isn't go code, and therefore, go can't tell if this object is still being used.
Therefore, if a GC takes place *while* the `fwd_free_value` is being executed, it might release that object.
It wouldn't be an issue if the object was passed in by value, though ( but I don't think that we need/want that ).

ffi/firewood.go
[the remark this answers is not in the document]
alarso16: I used contains for now - most of the Go code will have to be thrown out, so we can continue to address this

rkuris (changes requested):

ffi/src/lib.rs
```diff
@@ -43,20 +43,41 @@ impl Display for Value {
...
+            return "db should be non-null".into();
+        }
+    };
+
+    // Find root hash.
+    // Matches `hash` function but we use the TrieHash type here
+    let root = match db.root_hash_sync() {
+        Ok(Some(root)) => root,
+        Ok(None) => {
+            return "unexpected None from db.root_hash_sync".into();
```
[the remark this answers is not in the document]
alarso16: This might be my misunderstanding: Inside `db.root_hash_sync`, there's a map from `Option => Result`, so shouldn't it be impossible for there to be None? This passes unit tests, so this error isn't getting returned right now

ffi/src/lib.rs
```diff
@@ -43,20 +43,41 @@ impl Display for Value {
...
+    let db = match unsafe { db.as_ref() } {
+        Some(db) => db,
+        None => {
+            return "db should be non-null".into();
+        }
+    };
+
+    // Find root hash.
+    // Matches `hash` function but we use the TrieHash type here
+    let root = match db.root_hash_sync() {
```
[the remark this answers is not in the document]
alarso16: I ended up using `ok_or_else` because using the `String::from("constant string")`, and this allocation seems unnecessary in the success case

ffi/src/lib.rs
```diff
@@ -47,34 +47,35 @@ pub unsafe extern "C" fn fwd_get(db: *mut Db, key: Value) -> Value {
...
+            return e.to_string().into();
         }
     };
 
     // Get value associated with key.
     let value = rev.val_sync(key.as_slice());
     match value {
         Ok(Some(value)) => value.into(),
-        Ok(None) => "key not found".to_string().into(),
+        Ok(None) => "key not found".into(), // this is a valid case, handled in Go
```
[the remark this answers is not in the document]
alarso16: Done! that's the default value

ffi/firewood.go
```diff
@@ -130,34 +142,62 @@ func (db *Database) Batch(ops []KeyValue) []byte {
...
+
 	values, cleanup := newValueFactory()
 	defer cleanup()
 	val := C.fwd_get(db.handle, values.from(key))
-	bytes := extractBytesThenFree(&val)
-	if len(bytes) == 0 {
+	bytes, err := extractBytesThenFree(&val)
+
+	// If the root hash or key is not found, return nil.
+	if err != nil && (strings.Contains(err.Error(), rootHashNotFound) || strings.Contains(err.Error(), keyNotFound)) {
```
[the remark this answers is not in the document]
alarso16: They could do that - seemed better to be able to distinguish these cases in the future.

rkuris (approved):

Issue #865, opened 2025-05-06
Uncommitted proposals are never dropped.


alarso16:
Because the `NodeStore` (aka `ProposedRevision`) is stored in an `Arc`, it will not be deallocated until the strong reference count is 0. However, this will never happen in the case that the proposal is not committed. A clone is stored in a vector within the `RevisionManager` which is only removed upon committal. We would instead expect that once this is the only strong reference, the `RevisionManager` would ensure that the memory associated with the uncommitted proposal will be cleaned. Currently, if the user intends to create many `Proposal` structs and only commit some of them, the memory from uncommitted proposals will grow over time.

One solution to this is to check the number of strong references to ALL proposals at commit time. If any of the proposals are no longer being referenced anywhere else (`Arc::strong_count(proposal) == 1`), it can be removed from the list.

alarso16:
Addressed in #866.

PR #866, opened 2025-05-06
fix(src): drop unused revisions

alarso16:
As described in #865, even after a consumer drops an `Arc` to a proposal, the `RevisionManager` maintains a clone of all uncommitted proposals, which will cause the memory throughout runtime to grow, even though no users have access to the data anymore. A test is added to check that once a user-side `Arc<Proposal<'_>>` is dropped, and another proposal is committed, it will be removed from the list, and the `NodeStore` is no longer accessible. This is implemented by checking the number of strong references on commit for all proposals.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -623,15 +623,15 @@ mod test {
         let historical = db.revision(committed).await.unwrap();
         assert_eq!(&*historical.val(b"k1").await.unwrap().unwrap(), b"v1");
 
-        // the first nodestore shouldn't be available to commit anymore
+        // the second proposal shouldn't be available to commit anymore
         assert!(!db.all_hashes().await.unwrap().contains(&p2hash));
 
         // the third proposal should still be contained within the all_hashes list
-        // would be deleted if another proposal was committed and proposal3 was dropped here
         let hash3 = proposal3.root_hash().await.unwrap().unwrap();
         assert!(db.manager.read().unwrap().all_hashes().contains(&hash3));
 
         // moreover, the data from the second and third proposals should still be available
+        // through proposal3
         assert_eq!(&*proposal3.val(b"k2").await.unwrap().unwrap(), b"v2");
         assert_eq!(&*proposal3.val(b"k3").await.unwrap().unwrap(), b"v3");
     }
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -447,8 +447,6 @@ mod test {
     use std::ops::{Deref, DerefMut};
     use std::path::PathBuf;
 
-    use storage::HashedNodeReader;
-
     use crate::db::Db;
     use crate::v2::api::{Db as _, DbView as _, Error, Proposal as _};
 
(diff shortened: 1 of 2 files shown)


firewood/src/manager.rs
```diff
@@ -191,7 +191,9 @@ impl RevisionManager {
 
         // 8. Proposal Cleanup
         // first remove the committing proposal from the list of outstanding proposals
-        self.proposals.retain(|p| !Arc::ptr_eq(&proposal, p));
+        // Any proposals that are only referenced in this list will be freed
+        self.proposals
+            .retain(|p| !Arc::ptr_eq(&proposal, p) && Arc::strong_count(p) > 1);
```
rkuris, line 196: The comment here could be cleaner:
```
// Free proposal that is being committed as well as any proposals no longer
// referenced by anyone else

**Review by rkuris — CHANGES_REQUESTED** [review]
Comments mostly related to testing the functionality outside of the API instead of peeking under the covers at the implementation.

This needs a test for R1->P1->P2->P3 as well, since that edge case is possibly a bigger problem. If P2 is dropped, there should still be a reference to it in P3, so we should check that the data in P2 is still accessible through P3.

File: firewood/src/db.rs [hunk 2]
```diff
@@ -523,6 +525,117 @@ mod test {
...
+        // the second proposal shouldn't be available to commit anymore
+        assert!(!db.all_hashes().await.unwrap().contains(&p2hash));
+
+        // the third proposal should still be contained within the all_hashes list
+        // would be deleted if another proposal was committed and proposal3 was dropped here
+        let hash3 = proposal3
+            .nodestore
+            .root_hash()
+            .unwrap()
+            .expect("should have a root hash");
```
rkuris, line 579: nit: I usually just use expect in production code, not in tests. If the test fails because the proposal3 doesn't have a root hash, we have huge problems anyway :)

firewood/src/db.rs
```diff
@@ -523,6 +525,117 @@ mod test {
...
+            value: b"v1",
+        }];
+        let proposal1 = db.propose(batch1).await.unwrap();
+        assert_eq!(&*proposal1.val(b"k1").await.unwrap().unwrap(), b"v1");
+
+        let batch2 = vec![BatchOp::Put {
+            key: b"k2",
+            value: b"v2",
+        }];
+        let proposal2 = proposal1.clone().propose(batch2).await.unwrap();
```
rkuris, line 602: nit: not sure why the clone was needed here, but it's a test and it doesn't hurt anything.

firewood/src/db.rs
```diff
@@ -523,6 +525,117 @@ mod test {
...
+        let committed = db.root_hash().await.unwrap().unwrap();
+        let historical = db.revision(committed).await.unwrap();
+        assert_eq!(&*historical.val(b"k1").await.unwrap().unwrap(), b"v1");
+
+        // the second proposal shouldn't be available to commit anymore
+        assert!(!db.all_hashes().await.unwrap().contains(&p2hash));
+
+        // the third proposal should still be contained within the all_hashes list
+        let hash3 = proposal3.root_hash().await.unwrap().unwrap();
+        assert!(db.manager.read().unwrap().all_hashes().contains(&hash3));
```
rkuris, line 631:
```suggestion
        assert!(db.all_hashes().await.unwrap().contains(&hash3));
```

rkuris (changes requested):

PR #867, opened 2025-05-08
feat(ffi): proposal creation isolated from committing

rkuris:
This creates a `Proposal` struct in Go, allowing the batching and committing to occur in separate steps for any consumers. Some additional tests are created to ensure that only 1 proposal can be committed and errors are returned if an invalid proposal is attempted to be committed. 

This does not represent the full implementation for proposals, as there are many next steps:
* Creating proposals from other proposals
* Reading values from proposals
* Dropping proposals when no longer needed

Additionally, there are some code-cleanliness and testing still needed:
* Test memory handling for all `Value`s - or even better, find a better way of managing data across the FFI layer without so much magic.
* Move Value implementation to one central location (like another file?)
* Testing inputs (e.g. `KeyValue` structs) to ensure they are valid from Go
* Refactoring batching logic, in both Go and Rust, to reduce duplicate code

### DIFF
--- a/ffi/memory.go
+++ b/ffi/memory.go
@@ -51,6 +51,7 @@ func extractIdThenFree(v *C.struct_Value) (uint32, error) {
 	}
 	// If the value is an error string, it should be freed and an error
 	// returned.
+	// Any valid ID should be non-zero.
 	if v.len == 0 {
 		go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
 		C.fwd_free_value(v)
--- a/ffi/memory.go
+++ b/ffi/memory.go
@@ -15,7 +15,10 @@ import (
 	"unsafe"
 )
 
-var errNilBuffer = errors.New("firewood error: nil buffer returned from cgo")
+var (
+	errNilBuffer = errors.New("firewood error: nil value returned from cgo")
+	errBadValue  = errors.New("firewood error: value from cgo formatted incorrectly")
+)
 
 // KeyValue is a key-value pair.
 type KeyValue struct {
@@ -23,9 +26,20 @@ type KeyValue struct {
 	Value []byte
 }
 
+// extractErrorThenFree converts the cgo `Value` payload to a string, frees the
+// `Value` if an error is returned, and returns the extracted value.
+// If the data is not formatted for a string, and non-null it returns an error.
 func extractErrorThenFree(v *C.struct_Value) error {
 	if v == nil {
 		return errNilBuffer
+	}
+
+	// The length isn't expected to be set in either case.
+	// May indicate a bug.
+	if v.len != 0 {
+		// We should still attempt to free the value.
+		C.fwd_free_value(v)
+		return errBadValue
 	}
 
 	// Expected empty case for Rust's `()`
@@ -35,19 +49,21 @@ func extractErrorThenFree(v *C.struct_Value) error {
 
 	// If the value is an error string, it should be freed and an error
 	// returned.
-	if v.len == 0 {
-		go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
-		C.fwd_free_value(v)
-		// Pin the returned value to prevent it from being garbage collected.
-		runtime.KeepAlive(v)
-		return fmt.Errorf("firewood error: %s", go_str)
-	}
-	return nil
+	go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
+	C.fwd_free_value(v)
+	// Pin the returned value to prevent it from being garbage collected.
+	runtime.KeepAlive(v)
+	return fmt.Errorf("firewood error: %s", go_str)
 }
 
+// extractIdThenFree converts the cgo `Value` payload to a uint32, frees the
+// `Value` if an error is returned, and returns the extracted value.
 func extractIdThenFree(v *C.struct_Value) (uint32, error) {
-	if v == nil || v.len == 0 && v.data == nil {
+	if v == nil {
 		return 0, errNilBuffer
+	}
+	if v.len == 0 && v.data == nil {
+		return 0, errBadValue
 	}
 	// If the value is an error string, it should be freed and an error
 	// returned.
@@ -66,8 +82,11 @@ func extractIdThenFree(v *C.struct_Value) (uint32, error) {
 // the `Value`, and returns the extracted slice.
 // Generates error if the error term is nonnull.
 func extractBytesThenFree(v *C.struct_Value) (buf []byte, err error) {
-	if v == nil || v.data == nil {
+	if v == nil {
 		return nil, errNilBuffer
+	}
+	if v.data == nil {
+		return nil, errBadValue
 	}
 
 	buf = C.GoBytes(unsafe.Pointer(v.data), C.int(v.len))
--- a/ffi/memory.go
+++ b/ffi/memory.go
@@ -1,0 +1,104 @@
+// Package firewood provides a Go wrapper around the [Firewood] database.
+//
+// [Firewood]: https://github.com/ava-labs/firewood
+package firewood
+
+// // Note that -lm is required on Linux but not on Mac.
+// #cgo LDFLAGS: -L${SRCDIR}/../target/release -L/usr/local/lib -lfirewood_ffi -lm
+// #include <stdlib.h>
+// #include "firewood.h"
+import "C"
+import (
+	"errors"
+	"fmt"
+	"runtime"
+	"unsafe"
+)
+
+var errNilBuffer = errors.New("firewood error: nil buffer returned from cgo")
+
+// KeyValue is a key-value pair.
+type KeyValue struct {
+	Key   []byte
+	Value []byte
+}
+
+func extractErrorThenFree(v *C.struct_Value) error {
+	if v == nil {
+		return errNilBuffer
+	}
+
+	// Expected empty case for Rust's `()`
+	if v.data == nil {
+		return nil
+	}
+
+	// If the value is an error string, it should be freed and an error
+	// returned.
+	if v.len == 0 {
+		go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
+		C.fwd_free_value(v)
+		// Pin the returned value to prevent it from being garbage collected.
+		runtime.KeepAlive(v)
+		return fmt.Errorf("firewood error: %s", go_str)
+	}
+	return nil
+}
+
+func extractIdThenFree(v *C.struct_Value) (uint32, error) {
+	if v == nil || v.len == 0 && v.data == nil {
+		return 0, errNilBuffer
+	}
+	// If the value is an error string, it should be freed and an error
+	// returned.
+	if v.len == 0 {
+		go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
+		C.fwd_free_value(v)
+		// Pin the returned value to prevent it from being garbage collected.
+		runtime.KeepAlive(v)
+		return 0, fmt.Errorf("firewood error: %s", go_str)
+	}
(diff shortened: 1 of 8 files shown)


ffi/src/lib.rs
```diff
@@ -96,15 +135,105 @@ pub struct KeyValue {
...
+        .proposals
+        .write()
+        .unwrap()
+        .remove(&proposal_id)
+        .ok_or_else(|| String::from("proposal not found"))?;
+    let hash_val = proposal.root_hash_sync().map_err(|e| e.to_string())?;
+    proposal.commit_sync().map_err(|e| e.to_string())?;
+    Ok(hash_val
+        .map(|h| h.as_slice().to_vec())
+        .unwrap_or_default()
```
rkuris, line 223: This is probably wrong. Not sure what should be returned when everything is deleted.

ffi/src/lib.rs
```diff
@@ -1,18 +1,48 @@
...
 use firewood::manager::{CacheReadStrategy, RevisionManagerConfig};
 
 mod metrics_setup;
 
 use metrics::counter;
 
+type ProposalId = u32;
+
+pub struct DatabaseHandle<'p> {
+    proposals: RwLock<HashMap<ProposalId, Arc<Proposal<'p>>>>,
```
rkuris, line 23: This needs a comment. This code avoids UB by relying on the drop order. Reordering proposals and db would cause problems.

The other choice is to implement `Drop` manually and drain the proposals before dropping the remaining items.

ffi/src/lib.rs
```diff
@@ -1,18 +1,48 @@
...
+    db: Db,
+    next_id: AtomicU32,
```
rkuris, line 25: @alarso16 suggested that maybe this should be a global to avoid possible errors of committing against the wrong database.

ffi/src/lib.rs
```diff
@@ -29,6 +59,15 @@ impl Display for Value {
     }
 }
 
+impl From<u32> for Value {
```
rkuris, line 62: Yuck. This really should be a different type or really have a ton of comments. There is magic in that proposal 0 is impossible, so we use a `len` of 0 to indicate an error, otherwise the proposal id is stuffed into the len.

ffi/src/lib.rs
```diff
@@ -96,15 +135,105 @@ pub struct KeyValue {
...
+///
+/// # Returns
+///
+/// The ID of the proposal, or panics if it cannot be created
+///
+/// # Safety
+///
+/// This function is unsafe because it dereferences raw pointers.
+/// The caller must ensure that `db` is a valid pointer returned by `open_db`
+///
```
rkuris, line 156: also that values is valid too

ffi/src/lib.rs
```diff
@@ -96,15 +135,105 @@ pub struct KeyValue {
...
+
+fn propose(
+    db: *const DatabaseHandle,
+    nkeys: usize,
+    values: *const KeyValue,
+) -> Result<Value, String> {
+    let db = unsafe { db.as_ref() }.ok_or_else(|| String::from("db should be non-null"))?;
+
+    // Create a batch of operations to perform.
+    let mut batch = Vec::with_capacity(nkeys);
```
rkuris, line 174: This is duplicated code. It might be tricky to deduplicate because we want something on the stack here instead of allocating memory.

ffi/src/lib.rs
```diff
@@ -96,15 +135,105 @@ pub struct KeyValue {
...
+    for i in 0..nkeys {
+        let kv = unsafe { values.add(i).as_ref() }
```
rkuris, line 176: nit: we seem to check for nil pointers in most places, but not for values.

ffi/src/lib.rs
```diff
@@ -319,7 +448,7 @@ unsafe fn common_create(
     path: *const std::ffi::c_char,
     metrics_port: u16,
     cfg: DbConfig,
-) -> *mut Db {
+) -> *mut DatabaseHandle<'static> {
```
rkuris, line 451: can this be `*const DatabaseHandle<'_>`?

ffi/src/lib.rs
alarso16: This isn't required for the libevm use case - important note nonetheless to keep in mind. This code has been removed from the current implementation.

ffi/src/lib.rs
alarso16: Comment added!

ffi/src/lib.rs
alarso16: I added several comments in every location someone may get confused (Go and Rust). You're right though - this is, in general a terrible idea. We should find a more permanent solution to this.

ffi/src/lib.rs
alarso16: It is duplicated, but we *could* just remove the `batch` function entirely, and use `propose`, `commit`, and `hash` to replicate behavior. The only compelling reason not to do this is to test that proposals are persisted in the map properly. Although eventually (aka the next PR) will have a `fwd_propose_on_proposal` to create proposals on proposals, so it could be handy there.

ffi/src/lib.rs
alarso16: Added many many argument comments.

ffi/src/lib.rs
alarso16: `.as_ref` does check whether it's a nil pointer for `i=0`, so this is good. I did add clarity to the error message, and errors similar to this should be tested. I think this is out of scope for this PR, but I'll add it to the next steps.

ffi/src/lib.rs
alarso16: It can be `*const`, but the lifetime must be static.

ffi/memory.go
```diff
@@ -0,0 +1,105 @@
...
+	// If the value is an error string, it should be freed and an error
+	// returned.
+	if v.len == 0 {
+		go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
+		C.fwd_free_value(v)
+		// Pin the returned value to prevent it from being garbage collected.
+		runtime.KeepAlive(v)
+		return fmt.Errorf("firewood error: %s", go_str)
+	}
+	return nil
```
rkuris, line 45: If v.len is 0, shouldn't v.data also be nil? Maybe a debug assertion is a good idea here.

ffi/memory.go
```diff
@@ -0,0 +1,105 @@
...
+func newValueFactory() (*valueFactory, func()) {
+	f := new(valueFactory)
+	return f, func() { f.pin.Unpin() }
+}
+
+type valueFactory struct {
+	pin runtime.Pinner
+}
+
+func (f *valueFactory) from(data []byte) C.struct_Value {
```
rkuris, line 98: ooh, this looks mighty rusty 🦀

ffi/src/lib.rs
```diff
@@ -34,6 +69,12 @@ impl Display for Value {
 /// # Arguments
 ///
 /// * `db` - The database handle returned by `open_db`
+/// * `key` - The key to look up, in `Value` form
+///
+/// # Returns
+///
+/// A `Value` containing the root hash of the database.
+/// A `Value` containing {0, "error message"} if the commit failed.
```
rkuris, line 77: not if the commit failed, this is if the get fails

We need to be explicit here about what happens if the key is not found, which isn't really an error, but maybe it is treated as one?

Also we should mention this is getting the value from the most recent commit, not from some proposed revision.

Do we need get() from a proposed revision as well? I didn't see that code?

ffi/src/lib.rs
rkuris: It's also faster to not call the ffi more than once

ffi/src/lib.rs
```diff
@@ -143,15 +199,125 @@ fn batch(db: *mut Db, nkeys: usize, values: *const KeyValue) -> Result<Value, St
...
 /// Don't forget to call `free_value` to free the memory associated with the returned `Value`.
 ///
+/// # Argument
+///
+/// * `db` - The database handle returned by `open_db`
+///
+/// # Returns
+///
+/// A `Value` containing the root hash of the database.
+/// A `Value` containing {0, "error message"} if the commit failed.
```
rkuris, line 313: again, not if the commit failed, something else is probably amiss...

Let's be clear about what happens when you call this on a database with nothing in it

ffi/src/lib.rs
```diff
@@ -211,18 +377,46 @@ impl From<String> for Value {
     }
 }
 
+// WARNING: This should only be called with values >= 1.
```
rkuris, line 380: move this to code

ffi/src/lib.rs
```diff
@@ -211,18 +377,46 @@ impl From<String> for Value {
...
+// In much of the Go code, v.len == 0 is used to indicate a null-terminated string.
+// This may cause a panic!
+impl From<u32> for Value {
+    fn from(v: u32) -> Self {
+        Self {
```
rkuris, line 385: debug_assert if you prefer
```suggestion
        assert_ne!(v, 0);
        Self {
```

ffi/src/lib.rs
```diff
@@ -211,18 +377,46 @@ impl From<String> for Value {
...
 /// The caller must ensure that `value` is a valid pointer.
+///
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
-    // Create a Rust reference to the pointer.
-    // Unsafe because we are dereferencing a raw pointer - possible another thread
-    // has a mutable reference to the same memory.
-    let value = unsafe { &*value as &Value };
+    // Check value is valid.
+    let value = match unsafe { value.as_ref() } {
```
rkuris, line 415: nit: single return for match can probably be rewritten using `map`. maybe:
```rust
value.as_ref().map(|value| {
    if value.len > 0 {
        ...
    } else {
        ...
    }
}
```
although I also wonder if we should panic if you pass a nil pointer

firewood/src/db.rs
```diff
@@ -339,6 +339,13 @@ pub struct Proposal<'p> {
     db: &'p Db,
 }
 
+impl Proposal<'_> {
+    /// Get the root hash of the proposal synchronously
+    pub fn root_hash_sync(&self) -> Result<Option<api::HashKey>, api::Error> {
+        self.nodestore.root_hash().map_err(api::Error::from)
```
rkuris, line 345: Idiomatic, since ? automatically calls `Into::into`
```suggestion
        Ok(self.nodestore.root_hash()?)
```

rkuris (commented):
There are a few incorrect comments that definitely should be fixed. Remainder aren't so important.

I can't approve this, since I am the original author, but I can merge it without a stamp, or you can stamp it when you're ready.

ffi/memory.go
alarso16: I added another error to return when the data is not formatted as expected

ffi/src/lib.rs
alarso16: We'll need that eventually, but I wasn't going to make it right now. I think it's separate from the committing logic.

alarso16: Good catch in the comment though

ffi/src/lib.rs
alarso16: I'll use the real assert. If this happens, it could cause all sorts of issues - might be best to notice it at the source

ffi/src/lib.rs
alarso16: I think panicking is reasonable too - the caller is freeing a null pointer in C, and that would certainly panic

alarso16: Let's panic there, but I also added a check that the data field is not null

alarso16 (approved):

PR #868, opened 2025-05-08
Add workflow to generate and attach static libraries for golang FFI

aaronbuchwald:
This PR adds a GitHub workflow to generate and attach pre-built shared libraries of Firewood, so that the Firewood FFI package can be imported seamlessly in golang projects without requiring a two-step build process.

## Previous Golang Build Process

Prior to this PR, if you want to import Firewood into a golang project, it requires a two-step build process that pollutes every subsequent repo that transitively depends on Firewood. This is because the golang toolchain pulls in the Firewood FFI source code, which includes a CGO directive telling it where to find the Firewood shared library. That requires some shared library to be available on the machine where it's being compiled. The golang toolchain is smart enough to fall back to the C compiler available via `go env CC` when compiling C code, but it does _not_ have a Rust compiler to fall back to.

As a result, prior to this PR, the only way to import and use Firewood is to add it to your `go.mod` and use an explicit `go mod edit -replace` directive to a local path where it can find the shared library (alternatively we could add a CGO directive telling the linker to look in some standard location and check if the user has compiled it locally in that location, but this is quite dicey and less than ideal).

## Updated Build Process

This PR adds a GitHub Action workflow to:
- compile Firewood shared library on our supported architectures (with maxperf enabled, which keeps the binaries below 50MB and avoids [git lfs limitations](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github))
- creates a new branch, downloads the Firewood shared library artifacts from prior step, and adds them to source in `ffi/libs`
- checks out the new branch and tests the golang FFI package on each of the target architectures without compiling Firewood locally to enforce it depends on the attached static libraries

The attached static libraries enables the go toolchain to download the required shared libraries as part of the source code, so that there's no need for a two step build process. This means that you can simply import the Firewood FFI package directly (for the auto-generated tags/commits) and it will work out of the box on the supported architectures.

Un-supported architectures can still use Firewood using the previous `go mod edit -replace` semantics.

## Supported Architectures

Supported architectures with this change are:
- x86_64-unknown-linux-gnu
- aarch64-unknown-linux-gnu
- x86_64-apple-darwin
- aarch64-apple-darwin

## Action Triggers

The newly added GitHub Action runs on:

- tags -> takes vX.X.X and creates a new branch called vX.X.X-ffi-libs
- Manual -> takes `create_branch_name` as input and creates a new branch called `$create_branch_name-ffi-libs`

### Support ethhash

Firewood uses a compiler directive to support the `ethhash` feature for Ethereum hash compatibility. If we want to use this over the FFI, then we need to supply this flag when compiling the shared libraries for the FFI.

Note: this is intentionally left out of this PR becasue the FFI tests currently expect `ethhash` to be turned off.

TODO: create follow up to update FFI tests to use `ethhash` and enable it in the pre-built libraries.

### Guarantee Reproducible Builds

If we want to guarantee bit-for-bit reproducible builds, we may want to use a tool like [Bazel](https://bazel.build/).

This seems a bit heavy weight for a first pass, so planning to leave this out for now, but open to considering it as a hard requirement.

TODO: pin Rust version instead of using rust-stable and set up dependabot for more explicit Rust version updates.

### Improve Behavior for Unsupported Architectures

This PR adds single-step support for go projects that import the Firewood FFI layer, but if running on an architecture that we do not include a pre-built library for, then importing Firewood will produce a compiler error.

We should be able to improve this behavior by using go build constraints: https://pkg.go.dev/cmd/go#hdr-Build_constraints. With build constraints, we can provide a functional definition of firewood when the build constraints are satisfied and an alternative version that does compile, but where the constructor returns an error if the build constraints are not satisifed.

This could be used to make it easier to support Firewood as an optional backend for Coreth/AvalancheGo that will fail if not supported, but not prevent code that uses it from compiling on certain architectures.

### Remove Push Trigger

Removing the push trigger (used for CI testing) after it passed on this commit: https://github.com/ava-labs/firewood/pull/868/commits/a230edfb5e3ef3a29820fc6ddceb060519d2902f.

Removed the push trigger here: https://github.com/ava-labs/firewood/pull/868/commits/728d8cb973460366b3eacb99213e7ca25c78d068


aaronbuchwald:
After I messed up the naming convention, it seems that the GitHub Actions manage to hit a race condition on which commit (my latest commit or the auto-generated commit with the attached static libraries) it checks out:

https://github.com/ava-labs/firewood/actions/runs/14936980889/job/41966697809

Screenshot 2025-05-09 at 16 34 11

This should not be an issue after fixing to always push to a distinct branch.

.github/workflows/attach-static-libs.yaml
```diff
@@ -0,0 +1,136 @@
...
+            echo "target_branch=$target_branch" >> "$GITHUB_OUTPUT"
+          else
+            echo "No valid input or tag found."
+            exit 1
+          fi
+
+      - uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'ci: attach firewood-ffi static libs'
+          branch: ${{ steps.determine_branch.outputs.target_branch }}
```
rkuris, line 109: 😮 super cool

aaronbuchwald:
This is fairly strange:

https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:2:66

Screenshot 2025-05-09 at 17 20 57

The failing run is checking out the wrong branch and then saying that it cannot find the firewood shared library, but the others in the matrix are running an identical job and able to find it.

aaronbuchwald:
Failing in CI on `ubuntu-latest`, which seems to be correctly applying the linking directives we want, but not able to find the shared library:

https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:9

```
GOEXPERIMENT=cgocheck2 go test ./...
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_TERM_COLOR: always
# github.com/ava-labs/firewood/ffi.test
/opt/hostedtoolcache/go/1.2[3](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:3).8/x64/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/gcc -m6[4](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:4) -s -o $WORK/b001/ffi.test -Wl,--export-dynamic-symbol=_cgo_panic -Wl,--export-dynamic-symbol=_cgo_topofstack -Wl,--export-dynamic-symbol=crosscall2 -Wl,--compress-debug-sections=zlib /tmp/go-link-3769068834/go.o /tmp/go-link-3769068834/000000.o /tmp/go-link-3769068834/000001.o /tmp/go-link-3769068834/000002.o /tmp/go-link-3769068834/000003.o /tmp/go-link-3769068834/000004.o /tmp/go-link-3769068834/00000[5](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:5).o /tmp/go-link-3769068834/00000[6](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:7).o /tmp/go-link-3769068834/000007.o /tmp/go-link-3[7](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:8)69068834/000008.o /tmp/go-link-3769068834/000009.o /tmp/go-link-3769068834/000010.o /tmp/go-link-376906[8](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:9)834/000011.o /tmp/go-link-3769068834/000012.o /tmp/go-link-3769068834/000013.o /tmp/go-link-3769068834/000014.o /tmp/go-link-3769068834/000015.o /tmp/go-link-3769068834/000016.o /tmp/go-link-3769068834/000017.o /tmp/go-link-3769068834/000018.o /tmp/go-link-3769068834/000019.o /tmp/go-link-3769068834/000020.o /tmp/go-link-3769068834/000021.o -O2 -g -L/home/runner/work/firewood/firewood/ffi/libs/x86_64-unknown-linux-gnu -lm -L/home/runner/work/firewood/firewood/ffi/../target/release -L/usr/local/lib -lfirewood_ffi -O2 -g -lpthread -O2 -g -lresolv -no-pie
```

Tests are passing on an EC2 instance on the same architecture:

```
~/firewood/ffi$ go version && go test
go version go1.23.6 linux/amd64
PASS
ok  	github.com/ava-labs/firewood/ffi	0.035s
```

aaronbuchwald:
> Failing in CI on `ubuntu-latest`, which seems to be correctly applying the linking directives we want, but not able to find the shared library:
> 
> https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:9
> 
> ```
> GOEXPERIMENT=cgocheck2 go test ./...
>   shell: /usr/bin/bash -e {0}
>   env:
>     CARGO_TERM_COLOR: always
> # github.com/ava-labs/firewood/ffi.test
> /opt/hostedtoolcache/go/1.2[3](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:3).8/x64/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
> /usr/bin/gcc -m6[4](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:4) -s -o $WORK/b001/ffi.test -Wl,--export-dynamic-symbol=_cgo_panic -Wl,--export-dynamic-symbol=_cgo_topofstack -Wl,--export-dynamic-symbol=crosscall2 -Wl,--compress-debug-sections=zlib /tmp/go-link-3769068834/go.o /tmp/go-link-3769068834/000000.o /tmp/go-link-3769068834/000001.o /tmp/go-link-3769068834/000002.o /tmp/go-link-3769068834/000003.o /tmp/go-link-3769068834/000004.o /tmp/go-link-3769068834/00000[5](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:5).o /tmp/go-link-3769068834/00000[6](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:7).o /tmp/go-link-3769068834/000007.o /tmp/go-link-3[7](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:8)69068834/000008.o /tmp/go-link-3769068834/000009.o /tmp/go-link-3769068834/000010.o /tmp/go-link-376906[8](https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782492?pr=868#step:4:9)834/000011.o /tmp/go-link-3769068834/000012.o /tmp/go-link-3769068834/000013.o /tmp/go-link-3769068834/000014.o /tmp/go-link-3769068834/000015.o /tmp/go-link-3769068834/000016.o /tmp/go-link-3769068834/000017.o /tmp/go-link-3769068834/000018.o /tmp/go-link-3769068834/000019.o /tmp/go-link-3769068834/000020.o /tmp/go-link-3769068834/000021.o -O2 -g -L/home/runner/work/firewood/firewood/ffi/libs/x86_64-unknown-linux-gnu -lm -L/home/runner/work/firewood/firewood/ffi/../target/release -L/usr/local/lib -lfirewood_ffi -O2 -g -lpthread -O2 -g -lresolv -no-pie
> ```
> 
> Tests are passing on an EC2 instance on the same architecture:
> 
> ```
> ~/firewood/ffi$ go version && go test
> go version go1.23.6 linux/amd64
> PASS
> ok  	github.com/ava-labs/firewood/ffi	0.035s
> ```

It looks like all the passing tests in CI are reporting as cached, which should not happen: https://github.com/ava-labs/firewood/actions/runs/14937639338/job/41968782491?pr=868#step:4:67

Seems likely that testing is set up incorrectly (possibly due to go setup caching?) and it plays out differently across these environments.

.github/workflows/ci.yaml
```diff
@@ -106,7 +105,10 @@ jobs:
 
   ffi:
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
aaronbuchwald, line 111: Add MacOS testing since we are now providing pre-built binaries for both linux and mac on x86+arm

ffi/firewood.go
```diff
@@ -4,7 +4,15 @@
 package firewood
 
 // // Note that -lm is required on Linux but not on Mac.
-// #cgo LDFLAGS: -L${SRCDIR}/../target/release -L/usr/local/lib -lfirewood_ffi -lm
+// #cgo linux,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-unknown-linux-gnu -lm
+// #cgo linux,arm64 LDFLAGS: -L${SRCDIR}/libs/aarch64-unknown-linux-gnu -lm
+// #cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-apple-darwin
+// #cgo darwin,arm64 LDFLAGS: -L${SRCDIR}/libs/aarch64-apple-darwin
```
aaronbuchwald, line 10: Add pre-built binaries search paths with build constraints.

Add search paths for local build dir afterwards.

These can produce unused path warnings from linkers on different machines. A quick search indicates it is non-trivial to suppress these warnings.

ffi/src/lib.rs
```diff
@@ -432,7 +432,7 @@ impl Display for Value {
         match (self.len, self.data.is_null()) {
             (0, true) => write!(f, "[not found]"),
             (0, false) => write!(f, "[error] {}", unsafe {
-                CStr::from_ptr(self.data as *const i8).to_string_lossy()
+                CStr::from_ptr(self.data as *const c_char).to_string_lossy()
```
aaronbuchwald, line 435: `CStr::from_ptr` takes in `*const c_char`, which is a platform dependent type, so we need to use it directly instead of `*const i8` to compile across different architectures.

Specifically, this fails to compile on Linux/Arm64:

https://github.com/ava-labs/firewood/actions/runs/14982683172/job/42090244425#step:6:41

```
error[E0308]: mismatched types
   --> ffi/src/lib.rs:435:32
    |
435 |                 CStr::from_ptr(self.data as *const i8).to_string_lossy()
    |                 -------------- ^^^^^^^^^^^^^^^^^^^^^^ expected `*const u8`, found `*const i8`
    |                 |
    |                 arguments to this function are incorrect
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`
note: associated function defined here
   --> /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ffi/c_str.rs:264:25

For more information about this error, try `rustc --explain E0308`.
```

alarso16: Good catch! I didn't know this. I was also worried about something like this for `size_t`, but if it didn't cause any compiler errors, it's probably fine.

.github/workflows/attach-static-libs.yaml
```diff
@@ -0,0 +1,170 @@
...
+
+      - uses: actions/checkout@v4
+        with:
+          path: firewood
+
+      - uses: actions/checkout@v4
+        with:
+          repository: aaronbuchwald/firewood-go
+          token: ${{ secrets.FORK_FIREWOOD_GO_GITHUB_TOKEN }}
+          path: firewood-go
```
aaronbuchwald, line 101: Still waiting for approval on a PAT scoped to ava-labs/firewood-go, so I migrated to my own fork so I can create my own scoped token for now.

ffi/firewood.go
```diff
@@ -4,7 +4,15 @@
...
+// // XXX: last search path takes precedence, which means we prioritize
+// // local builds over pre-built and maxperf over release build
+// #cgo LDFLAGS: -L${SRCDIR}/../target/release
```
rkuris, line 13: This is a great idea, we should also include debug in this list. That is, if you have a debug build, it will use that. If not, it will use the release build, and if not, the maxperf build. This would avoid some problems I ran into when I did a "cargo build" and forgot to ask for a release build before retesting ffi.

We should make this very clear in the development workflows.

.github/workflows/attach-static-libs.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Build for ${{ matrix.target }}
+      # TODO: add ethhash feature flag after updating FFI tests
+        run: cargo build --profile maxperf --features logger --target ${{ matrix.target }} -p firewood-ffi
+
+      - name: List compiled target directory
```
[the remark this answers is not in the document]
aaronbuchwald: Ya I can remove it

.github/workflows/attach-static-libs.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Build for ${{ matrix.target }}
+      # TODO: add ethhash feature flag after updating FFI tests
+        run: cargo build --profile maxperf --features logger --target ${{ matrix.target }} -p firewood-ffi
```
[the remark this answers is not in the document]
aaronbuchwald: This has been moved to a separate PR as previously discussed: https://github.com/ava-labs/firewood/pull/891

Happy to block this on completing that if you prefer

.github/workflows/attach-static-libs.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - name: Push static libs to branch
+        working-directory: firewood-go
+        # GITHUB_TOKEN is configured in the last actions/checkout step
+        # to have read/write permissions to the firewood-go repo.
+        run: |
+          git config --global user.name "FirewoodCI"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git checkout -b ${{ steps.determine_branch.outputs.target_branch }}
+          git add .
+          git commit -m "ci: attach firewood static libs"
```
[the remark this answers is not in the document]
aaronbuchwald: Good call

.github/workflows/attach-static-libs.yaml
```diff
@@ -0,0 +1,170 @@
...
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: "ffi/go.mod"
+          cache-dependency-path: "ffi/go.sum"
+      - name: Test Go FFI bindings
+        working-directory: ffi
+        # cgocheck2 is expensive but provides complete pointer checks
+        run: GOEXPERIMENT=cgocheck2 go test ./...
+
+  garbage-collect-branch:
```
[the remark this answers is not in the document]
aaronbuchwald: Updating. The whole workflow typically finishes faster than the rest of CI, so not sure we need to be concerned about this.

I'd opt for running on every PR and change it if it becomes an issue, lmk what you prefer and happy to modify.

ffi/firewood.go
aaronbuchwald: Also as mentioned in the comment - the last search path takes precedence, so the ordering line by line of:

1. Release
2. Maxperf

Means that maxperf takes precedence over release as you requested previously. Happy to modify the ordering if you prefer.

For now, I am adding `debug` to the top of that list going with my understanding of the previous motivation to prioritize using `maxperf` whenever possible.

Maybe you meant that only for the static builds we are pushing for the combination of smaller binaries and performance, but prefer debug to be prioritized when testing locally?

aaronbuchwald:
> Can you describe the workflow where wants to test a change in firewood and needs a new FFI to test with a little better? Both the case where I'm still not fully understanding how someone creates temporary FFI libraries for testing, unless we just plan on them doing it the old way (cargo build --release and use those ffi libs).

Yup, just updated the README: https://github.com/ava-labs/firewood/pull/868/commits/a1611f7bd46715a72129f0b058a955278a502c50#diff-e41bb24cf049888d2b79cd8b03d3c745a48ff81de0b21df86ee381d490a958c5

TLDR; you can develop entirely locally if you like both within Firewood and a dependent project: https://github.com/ava-labs/firewood/blob/a1611f7bd46715a72129f0b058a955278a502c50/ffi/README.md#local-development

If you want to create a temporary FFI library, you can either use a manual workflow to push from any branch in Firewood or create a tag. Both of these will generate a new branch on `firewood-go` that will not be garbage collected.

aaronbuchwald:
Note: building with the attached static libraries may produce an unused paths warning:

```
# github.com/ava-labs/firewood/ffi.test
ld: warning: ignoring duplicate libraries: '-lfirewood_ffi', '-lm'
ld: warning: search path '/Users/aaronbuchwald/go/src/github.com/ava-labs/firewood-go/ffi/../target/debug' not found
ld: warning: search path '/Users/aaronbuchwald/go/src/github.com/ava-labs/firewood-go/ffi/../target/release' not found
ld: warning: search path '/Users/aaronbuchwald/go/src/github.com/ava-labs/firewood-go/ffi/../target/maxperf' not found
ld: warning: search path '/Users/aaronbuchwald/go/src/github.com/ava-labs/firewood-go/ffi/../target/release' not found
ld: warning: search path '/Users/aaronbuchwald/go/src/github.com/ava-labs/firewood-go/ffi/../target/release' not found
ld: warning: search path '/Users/aaronbuchwald/go/src/github.com/ava-labs/firewood-go/ffi/../target/release' not found
PASS
ok      github.com/ava-labs/firewood/ffi        0.388s
```

I am not sure how to both include multiple search paths where we only expect to use one of them and suppress the warnings.

.github/workflows/attach-static-libs.yaml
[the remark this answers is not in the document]
rkuris: Nah, agreed, it's fine as a new PR.

ffi/firewood.go
rkuris: Yes, and just to avoid the stupid mistake of `cargo build && go test .` that I have made several times. I think you also get better stack traces when you're debugging with debug builds.

rkuris:
Three possible approaches to remove the warnings

(1) See if LDFLAGS supports shell escapes, like this:

```suggestion
// #cgo LDFLAGS: $(for dir in "${SRCDIR}/../target/debug" "${SRCDIR}/../target/release ${SRCDIR}/../target/maxperf"; do if [ -d ${dir} ]; then echo -L${dir}; fi; done)
```

(2) Remove all the `cgo LDFLAGS` lines and then when running tests locally, search for them on the command line:
```sh
go test -ldflags "$(for dir in "${SRCDIR}/../target/debug" "${SRCDIR}/../target/release ${SRCDIR}/../target/maxperf"; do if [ -d ${dir} ]; then echo -L${dir}; fi; done)"
```

I think that would let you just leave the location of the binary-release files in there and not have to see warnings during firewood builds. I don't love this as it makes local testing with `go test .` stop working.

(3) Filter out these errors using something like:

```sh
go test . 2>&1 | grep -v 'warning: search path .* not found'
```

.github/workflows/attach-static-libs.yaml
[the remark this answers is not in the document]
rkuris: Yeah let's leave it until it becomes a problem. I suspect that having so many builds means it increases the chance of flaky builds, but maybe not.

ffi/README.md
```diff
@@ -1,13 +1,52 @@
...
+
+to a separate repo [firewood-go](https://github.com/ava-labs/firewood-go) (to avoid including binaries in the Firewood repo).
+
+### Local Development
+
+[firewood.go](./firewood.go) includes CGO directives to include builds in the `target/` directory.
+
+Firewood prioritizes builds in the following order:
+
+1. maxperf
```
rkuris, line 24: Is this correct? Seems like the other way is better (debug before release before maxperf)?

ffi/firewood.go
```diff
@@ -4,7 +4,16 @@
 package firewood
 
 // // Note that -lm is required on Linux but not on Mac.
-// #cgo LDFLAGS: -L${SRCDIR}/../target/release -L/usr/local/lib -lfirewood_ffi -lm
+// #cgo linux,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-unknown-linux-gnu -lm
```
rkuris, line 7: nit: we could remove the comment on line 6 and replace it with:
```
// #cgo darwin LDFLAGS: -lm
```

rkuris (approved):
This looks good enough to push for me. Feel free to work on the stuff I mentioned, which are mostly all nits, address them in a followup, or drop them.

ffi/README.md
aaronbuchwald: I'll switch in a follow up. This is what you requested previously.

PR #873, opened 2025-05-09
test(ffi): Add memory checks for `Value`

alarso16:
This refactors the `extractXXXXThenFree` to be used in all scenarios, so that the caller must explicitly recognize what values are being returned, and all decisions about how the `Value` may be formatted is dictated to a single function in `memory.go`. This also adds a function in `Rust` to test allocations and deallocations to ensure all possible `Value` structs are handled without panicking and with expected behavior.

### DIFF
--- a/ffi/memory_test.go
+++ b/ffi/memory_test.go
@@ -1,0 +1,159 @@
+// Package firewood provides a Go wrapper around the [Firewood] database.
+//
+// [Firewood]: https://github.com/ava-labs/firewood
+package firewood
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// This checks all inputs for `extractErrorThenFree`
+func TestExtractError(t *testing.T) {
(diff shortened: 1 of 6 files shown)


ffi/memory_test.go
```diff
@@ -0,0 +1,68 @@
...
+			name:          "length nonzero",
+			expectedInt:   1,
+			expectedBytes: nil,
+			expectedError: "",
+			value:         newCValueLen(1),
+		},
+		{
+			name:          "data length nonzero",
+			expectedInt:   0,
+			expectedBytes: []byte("test bytes\x00"), // count null pointer created by C
```
alarso16, line 44: This is kind of ugly, but the easiest way to pass data to Rust

alarso16:
These changes, although may prevent bugs, adds unnecessary coupling with the Rust FFI and decreases readability. However, bug fixes and readability improvements are addressed in #875.

PR #874, opened 2025-05-09
doc(ffi): remove private declarations from public docs

alarso16:
Ignores a few unnecessarily documented fields from the docs


ffi/src/lib.rs
```diff
@@ -13,15 +13,23 @@ use std::sync::{Arc, RwLock};
...
 use metrics::counter;
 
+#[doc(hidden)]
+mod metrics_setup;
+
 #[global_allocator]
+#[doc(hidden)]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
+#[doc(hidden)]
```
rkuris, line 25: nit: I would keep this one, and add a doc comment. It's nice to see the type of our proposal IDs, especially since we return them (wrapped in a usize of course, but still).

rkuris (approved):
One small nit, but feel free to merge without fixing it.

PR #875, opened 2025-05-09
fix(ffi): Clarify roles of `Value` extractors

alarso16:
Using the magic `Value` struct to transfer any payload returned by `cgo` is error-prone and relies on the caller to exactly identify which possible values could be returned. Comments are added to clarify when to use each extractor, and all possible `Values` returned from the Rust FFI are attempted to be handled gracefully, and an error would be returned during general testing as long as the unit tests for the Golang interface are thorough

### DIFF
--- a/ffi/memory.go
+++ b/ffi/memory.go
@@ -34,6 +34,9 @@ func extractErrorThenFree(v *C.struct_Value) error {
 		return errNilBuffer
 	}
 
+	// Pin the returned value to prevent it from being garbage collected.
+	defer runtime.KeepAlive(v)
+
 	// The length isn't expected to be set in either case.
 	// May indicate a bug.
 	if v.len != 0 {
@@ -49,11 +52,9 @@ func extractErrorThenFree(v *C.struct_Value) error {
 
 	// If the value is an error string, it should be freed and an error
 	// returned.
-	go_str := C.GoString((*C.char)(unsafe.Pointer(v.data)))
+	errStr := C.GoString((*C.char)(unsafe.Pointer(v.data)))
 	C.fwd_free_value(v)
-	// Pin the returned value to prevent it from being garbage collected.
-	runtime.KeepAlive(v)
-	return fmt.Errorf("firewood error: %s", go_str)
+	return fmt.Errorf("firewood error: %s", errStr)
 }
 
 // extractIdThenFree converts the cgo `Value` payload to a uint32, frees the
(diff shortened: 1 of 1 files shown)


ffi/memory.go
```diff
@@ -26,14 +26,19 @@ type KeyValue struct {
...
+// 2. a non-nil error, indicating an error occurred.
+// This should only be called when the `Value` is expected to only contain an error.
+// Otherwise, an error is returned.
 func extractErrorThenFree(v *C.struct_Value) error {
 	if v == nil {
 		return errNilBuffer
 	}
 
+	// Pin the returned value to prevent it from being garbage collected.
+	defer runtime.KeepAlive(v)
```
rkuris, line 40: Good catch, this is possibly a bug if it's needed.

ffi/memory.go
```diff
@@ -49,53 +54,85 @@ func extractErrorThenFree(v *C.struct_Value) error {
...
+// 2. a zero uint32 and a non-nil error, indicating an error occurred.
+// This should only be called when the `Value` is expected to only contain an error or an ID.
+// Otherwise, an error is returned.
 func extractIdThenFree(v *C.struct_Value) (uint32, error) {
 	if v == nil {
 		return 0, errNilBuffer
 	}
+
+	// Pin the returned value to prevent it from being garbage collected.
+	defer runtime.KeepAlive(v)
```
rkuris, line 73: question only: can't you call this if v is nil? If so, perhaps it should be at the beginning of the function?

Why? Defer ordering is strange and if someone else adds a defer that uses v earlier in the function, the order will be unexpectedly reversed.

Commentary: I hate defer.

ffi/memory.go
```diff
@@ -49,53 +54,85 @@ func extractErrorThenFree(v *C.struct_Value) error {
...
+	if v.data != nil {
+		errStr := C.GoString((*C.char)(unsafe.Pointer(v.data)))
 		C.fwd_free_value(v)
-		// Pin the returned value to prevent it from being garbage collected.
-		runtime.KeepAlive(v)
-		return 0, fmt.Errorf("firewood error: %s", go_str)
+		return 0, fmt.Errorf("firewood error: %s", errStr)
 	}
+
+	// Normal case, length is non-zero and data is nil.
```
rkuris, line 95: I wonder if we can optimize this code some. A check at the top for the most common case might result in faster execution ... `if v.len > 0 && v.data != nil`

ffi/memory.go
```diff
@@ -49,53 +54,85 @@ func extractErrorThenFree(v *C.struct_Value) error {
...
+	// Data non-nil but length is zero indcates an error.
 	if v.len == 0 {
 		errStr := C.GoString((*C.char)(unsafe.Pointer(v.data)))
-		err = fmt.Errorf("firewood error: %s", errStr)
+		C.fwd_free_value(v)
+		return nil, fmt.Errorf("firewood error: %s", errStr)
 	}
+
+	// Normal case, data is non-nil and length is non-zero.
+	buf := C.GoBytes(unsafe.Pointer(v.data), C.int(v.len))
```
rkuris, line 132: again, because golang, we might want to check for the normal case first. In rust this may not matter, as llvm and immutability rules, along with profile guided optimization.

ffi/memory.go
```diff
@@ -49,53 +54,85 @@ func extractErrorThenFree(v *C.struct_Value) error {
...
+	if v.data == nil {
 		return nil, nil
 	}
 
-	buf = C.GoBytes(unsafe.Pointer(v.data), C.int(v.len))
+	// Data non-nil but length is zero indcates an error.
 	if v.len == 0 {
 		errStr := C.GoString((*C.char)(unsafe.Pointer(v.data)))
-		err = fmt.Errorf("firewood error: %s", errStr)
+		C.fwd_free_value(v)
```
rkuris, line 127: Since we always want to call `fwd_free_value` whenever v is not nil (and maybe even if it is) maybe this should be a `defer` (even though I hate defer)

rkuris (approved):
It might not be a bad idea to get @ARR4N to take a look at this too, but I think this is fine. Lots of comments on how I think it could be better, but nothing pressing.

ffi/memory.go
alarso16: We don't need to call `fwd_free_value` unless there was something allocated in Rust (so the `v.data != nil`). I try to avoid calling this - it's just unnecessary

ffi/memory.go
alarso16: I wasn't sure what the expected behavior is if you call `runtime.KeepAlive(nil)` - it turns out it's a no-op. I'll put it at the top

ffi/memory.go
alarso16: done

PR #876, opened 2025-05-12
Disable metrics during ffi unit tests

aaronbuchwald:
This PR disables Firewood metrics collection by setting `MetricsPort=0` https://github.com/ava-labs/firewood/blob/b432ee9920c032fc909f4d2e1b8ac994ab11c861/ffi/src/lib.rs#L577

This avoids test failures on machines where the default port may be occupied.

### DIFF
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -44,6 +44,7 @@ func newTestDatabase(t *testing.T) *Database {
 	t.Helper()
 
 	conf := DefaultConfig()
+	conf.MetricsPort = 0
 	conf.Create = true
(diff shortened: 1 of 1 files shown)


alarso16 (approved):

rkuris (approved):
Seems like selecting a random port in the range (49152, 65535), then verifying nothing is on that port, is a better choice, but this unblocks you, so stamping it.

Could also check the failure on port 3000 and pick a random port if that fails too.

ChatGPT says:
```
func isPortInUse(port int) bool {
	listener, err := net.Listen("tcp", ":"+strconv.Itoa(port))
	if err != nil {
		return true
	}
	defer listener.Close()
	return false
}
```

PR #878, opened 2025-05-12
feat(ffi): Full proposal support

alarso16:
This PR addresses all additional functionality required to support creating proposal trees and dropping unused proposals.

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -148,6 +148,23 @@ struct Value fwd_commit(const struct DatabaseHandle *db, uint32_t proposal_id);
 const struct DatabaseHandle *fwd_create_db(struct CreateOrOpenArgs args);
 
 /**
+ * Drops a proposal from the database.
+ * The propopsal's data will not be freed immediately.
+ *
+ * # Arguments
+ *
+ * * `db` - The database handle returned by `open_db`
+ * * `proposal_id` - The ID of the proposal to drop
+ *
+ * # Safety
+ *
+ * This function is unsafe because it dereferences raw pointers.
+ * The caller must ensure that `db` is a valid pointer returned by `open_db`
+ *
+ */
+void fwd_drop_proposal(const struct DatabaseHandle *db, uint32_t proposal_id);
+
+/**
  * Frees the memory associated with a `Value`.
  *
  * # Arguments
@@ -199,7 +216,7 @@ struct Value fwd_get_from_proposal(const struct DatabaseHandle *db,
  *
  * A `Value` containing the root hash of the database.
  * A `Value` containing {0, "error message"} if the get failed.
- * There is one error case that may be expected to be nil by the caller,
+ * There is one error case that may be expected to be null by the caller,
  * but should be handled externally: The database has no entries - "IO error: Root hash not found"
  * This is expected behavior if the database is empty.
  *
@@ -244,7 +261,7 @@ const struct DatabaseHandle *fwd_open_db(struct CreateOrOpenArgs args);
  *
  * # Returns
  *
- * The new root hash of the database, in Value form.
+ * A `Value` containing {id, null} if creating the proposal succeeded.
  * A `Value` containing {0, "error message"} if creating the proposal failed.
  *
  * # Safety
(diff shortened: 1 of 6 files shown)


ffi/src/lib.rs
```diff
@@ -329,6 +329,81 @@ fn propose_on_db(
...
+    db: *const DatabaseHandle,
+    proposal_id: ProposalId,
+    nkeys: usize,
+    values: *const KeyValue,
+) -> Result<ProposalId, String> {
+    let db = unsafe { db.as_ref() }.ok_or_else(|| String::from("db should be non-null"))?;
+
+    // Get proposal from ID.
+    // We need write access to add the proposal after we create it.
+    let mut proposals = db.proposals.write().unwrap();
```
rkuris, line 380: Curious as to why this unwrap was accepted by clippy. If you want to keep this here, the function documentation needs a Panics section, but it might be better to return an error if the proposal lock is poisoned.

rkuris: This lock is held a really long time, preventing parallel reads from any proposals. It seems like we should demote this to a read lock here and get a new write lock later.

ffi/src/lib.rs
```diff
@@ -366,8 +441,29 @@ fn commit(db: *const DatabaseHandle, proposal_id: u32) -> Result<(), String> {
...
+/// * `proposal_id` - The ID of the proposal to drop
+///
+/// # Safety
+///
+/// This function is unsafe because it dereferences raw pointers.
+/// The caller must ensure that `db` is a valid pointer returned by `open_db`
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn fwd_drop_proposal(db: *const DatabaseHandle, proposal_id: u32) {
+    let db = match unsafe { db.as_ref() } {
```
rkuris, line 459:
```rust
let Some(db) = db.as_ref() else { return }
```
Although it does seem like we should report errors here, such as passing a bad proposal ID or a nil database.

firewood/src/db.rs
```diff
@@ -453,6 +453,42 @@ impl Proposal<'_> {
         let merkle = Merkle::from(self.nodestore.clone());
         merkle.get_value(key.as_ref()).map_err(api::Error::from)
     }
+
+    /// Create a new proposal from the current one synchronously
+    pub fn propose_sync<K: KeyType, V: ValueType>(
```
rkuris, line 458: This is a lot of copy pasta. Let's make a function they both call.

ffi/src/lib.rs
alarso16: No you're right - I should return an error. For the map lock, that's more confusing. I was worried about that, but I didn't want to hold a reference to the proposal after the lock went out of scope, and it seemed weird to arbitrarily scope the lock. Is that the correct way to do it?

ffi/src/lib.rs
alarso16: That seems reasonable too.

firewood/src/db.rs
alarso16: I'm not confident that this works correctly, but I made an attempt.

ffi/src/lib.rs
rkuris: I figured out why we aren't getting warnings. You should avoid unwrap. It's really designed for prototyping. If you are 100% sure there is no way the unwrap can fail, then use `expect`.

In this case, the code wants to know what we should do if some thread dies while holding the proposals lock. I think it's okay to die there too, but we should use expect so you get a message saying "poisoned proposal lock" or something like that.

ffi/firewood.h
```diff
@@ -147,6 +147,23 @@ struct Value fwd_commit(const struct DatabaseHandle *db, uint32_t proposal_id);
  */
 const struct DatabaseHandle *fwd_create_db(struct CreateOrOpenArgs args);
 
+/**
+ * Drops a proposal from the database.
+ * The propopsal's data will not be freed immediately.
```
rkuris, line 152: Why not? I think it will!

ffi/firewood_test.go
```diff
@@ -341,7 +341,311 @@ func TestFakeProposal(t *testing.T) {
...
+	db := newTestDatabase(t)
+
+	// Create a fake proposal with an invalid ID.
+	proposal := &Proposal{
+		handle: db.handle,
+		id:     1, // note that ID 0 is reserved for invalid proposals
+	}
+
+	// Attempt to drop the fake proposal.
+	// This should not be an error, since dropping a proposal is a "no-op".
```
rkuris, line 364: This comment seems wrong. It is an error, as the proposal ID being passed in is not valid.

ffi/firewood_test.go
```diff
@@ -341,7 +341,311 @@ func TestFakeProposal(t *testing.T) {
...
+	err := proposal.Drop()
+	require.Contains(t, err.Error(), "proposal not found", "Drop(fake proposal)")
+
+	// Attempt to get drop again will return an error.
```
rkuris, line 368: I'm not following this test very well.

Why do you need to drop it twice?

ffi/firewood_test.go
```diff
@@ -341,7 +341,311 @@ func TestFakeProposal(t *testing.T) {
...
+
+	// Attempt to commit the dropped proposal.
+	err = proposal.Commit()
+	require.ErrorIs(t, err, errDroppedProposal, "Commit(dropped proposal)")
+
+	// Attempt to get a value from the dropped proposal.
+	_, err = proposal.Get([]byte("non-existent"))
+	require.ErrorIs(t, err, errDroppedProposal, "Get(dropped proposal)")
+
+	// Attempt to "emulate" the proposal to ensure it isn't internally available still.
```
rkuris, line 398: This seems eerily familiar to the prior test.

rkuris (approved):

ffi/firewood.h
alarso16: It won't! So it might be worth clarifying. There will still be 1 reference to the proposal in the `RevisionManager`, and it won't be freed until another proposal is committed, and iterating through proposals to find all those with only one reference. I'll update the comment, since this is not something the user should worry about

ffi/firewood_test.go
alarso16: We expect different errors each time. The first time, we expect "proposal not found" sent from Rust, but the second time, we expect `errDroppedProposal` sent from Go. I don't necessarily "love" that this sends two different errors, but I have it like this because after attempting to `Drop` a proposal, you should not be able to even attempt to call any method with it

alarso16: If you think it's unnecessary, and the second attempt to drop is UB, we can remove this section of the test

rkuris: Definitely it's not UB but testing that we get different errors isn't something that we should verify the behavior of as it might change. As long as it returns *some* error it's probably okay. But this is still okay.

PR #881, opened 2025-05-13
feat(ffi): Support `Get` for historical revisions

alarso16:
This allows retrieving values from previously committed revisions.

Further steps:
- More tests once #878 is merged
- Supporting empty revisions

### DIFF
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -345,3 +345,100 @@ func TestFakeProposal(t *testing.T) {
 	err = proposal.Commit()
 	require.Contains(t, err.Error(), "proposal not found", "Commit(fake proposal)")
 }
+
+// Tests that an empty revision can be retrieved.
+func TestRevision(t *testing.T) {
+	db := newTestDatabase(t)
+
+	// Create a proposal with 10 key-value pairs.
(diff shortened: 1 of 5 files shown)


ffi/firewood_test.go
```diff
@@ -345,3 +345,100 @@ func TestFakeProposal(t *testing.T) {
...
+			vals[j] = valForTest(i*numProposals + j)
+		}
+		root, err := db.Update(keys, vals)
+		require.NoError(t, err, "Update(%d)", i)
+		roots[i] = root
+	}
+
+	// Now that there are 101 roots, the first one should be invalid.
+	revision, err := db.Revision(roots[0])
+	// Note this shouldn't error, because the root is formed correctly.
```
alarso16, line 426: I don't know if this should fail here, since we *could* confirm that the root is lost. However, it doesn't guarantee that the root will be valid at access, so I decided we don't need to check here. Let me know if you disagree.

alarso16: My worry here was that in the case we talked about yesterday, where we implement some `DbView`-like thing, we do expect to remove old revisions from this map somehow. But we more or less decided not to worry about that, so I can just get rid of the test I guess

PR #884, opened 2025-05-14
Pin Rust Version in GitHub Actions

aaronbuchwald:
This PR pins the Rust version used in CI to 1.77.2 and adds GitHub Actions to dependabot, so that we get automated PRs to update it. This is a follow up to https://github.com/ava-labs/firewood/pull/868, so that the pre-built binaries created in CI are added with a precise Rust version instead of stable.

We could look in CI to see what Rust version was actually pulled in and used to generate the binaries, but as discussed it's preferable to use a pinned Rust version and automate updates using Dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,13 @@ updates:
     open-pull-requests-limit: 10
     allow:
       - dependency-type: "all"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "05:00"
+      timezone: "America/Los_Angeles"
+    open-pull-requests-limit: 10
+    allow:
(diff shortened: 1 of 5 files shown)


aaronbuchwald:
Unfortunately, it looks like `dtolnay/rust-toolchain` maintains branches for every rust release instead of creating actual releases. I'm not sure if dependabot will handle this correctly.

It maintains a branch for each rust release version, so that you can specify a rust version as the revision and it will check out that branch:

```
- uses: dtolnay/rust-toolchain@1.86.0
```

However, the recommendation is to use `@stable` and specify the toolchain as an input:

```
      - uses: dtolnay/rust-toolchain@stable
        with:
          toolchain: 1.86.0
```

This means that we'll use the latest stable version of the action itself and pin the toolchain that gets installed.

PR #888, opened 2025-05-14
If ethhash is enabled, return the hash for nil RLP

rkuris:
This calculates and returns the nil RLP instead of None. None is still used internally since we need to know the difference between "there is no revision" and "there is a revision, but it consists of an empty trie".

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -6,6 +6,8 @@ use std::io::Error;
 use std::num::NonZero;
 use std::path::PathBuf;
 use std::sync::Arc;
+#[cfg(feature = "ethhash")]
+use std::sync::OnceLock;
 
 use storage::logger::{trace, trace_enabled, warn};
 use typed_builder::TypedBuilder;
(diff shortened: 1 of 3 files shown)


firewood/src/manager.rs
```diff
@@ -240,6 +258,24 @@ impl RevisionManager {
...
+
+    #[cfg(feature = "ethhash")]
+    #[inline]
+    pub fn empty_trie_hash(&self) -> Option<TrieHash> {
+        // clippy is wrong here. we need to keep the closure since empty_trie_hash
+        // is an instance method that needs self.
+        #[allow(clippy::redundant_closure)]
+        Some(
+            self.empty_hash
+                .get_or_init(|| storage::empty_trie_hash())
```
alarso16, line 275: Why does the closure need self?

storage/src/hashers/ethhash.rs
```diff
@@ -194,6 +194,7 @@ impl<T: Hashable> Preimage for T {
 
             // we've collected all the children in bytes
 
+            #[allow(clippy::let_and_return)]
```
alarso16, line 197: This is the only line changed in this file - why is it necessary?

firewood/src/manager.rs
```diff
@@ -47,6 +49,9 @@ pub(crate) struct RevisionManager {
     proposals: Vec<ProposedRevision>,
     // committing_proposals: VecDeque<Arc<ProposedImmutable>>,
     by_hash: HashMap<TrieHash, CommittedRevision>,
+
+    #[cfg(feature = "ethhash")]
+    empty_hash: OnceLock<TrieHash>,
```
alarso16, line 54: Clever!

alarso16 (approved):

firewood/src/manager.rs
rkuris: That's where the OnceLock is

PR #890, opened 2025-05-14
fix(ffi): Check revision is available

alarso16:
I think expected behavior for this would be to check that the revision is available - although it provides no guarantee that it will be available for the entire length of use, it should fail-fast if already unavailable.

Additionally, `newRevision` shouldn't be exported - this is only required in `db.Revision`, and that's how all consumers should use it.

### DIFF
--- a/ffi/revision.go
+++ b/ffi/revision.go
@@ -14,8 +14,8 @@ import (
 )
 
 var (
-	errRevisionClosed = errors.New("firewood revision already closed")
-	errInvalidRoot    = fmt.Errorf("firewood error: root hash must be %d bytes", RootLength)
+	errRevisionInaccessible = errors.New("firewood revision inaccessible")
+	errInvalidRootLength    = fmt.Errorf("firewood error: root hash must be %d bytes", RootLength)
 )
 
 type Revision struct {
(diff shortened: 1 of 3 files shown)


ffi/revision.go
```diff
@@ -27,14 +27,26 @@ type Revision struct {
...
-		return nil, errInvalidRoot
+		return nil, errInvalidRootLength
+	}
+
+	// Attempt to get any value from the root.
+	// This will verify that the root is valid and accessible.
+	// If the root is not valid, this will return an error.
+	values, cleanup := newValueFactory()
+	defer cleanup()
+	val := C.fwd_get_from_root(handle, values.from(root), values.from([]byte{}))
```
alarso16, line 45: If this should be checked in a less hacky way, we can add a `fwd_has_root` or something, but it seems unnecessary.

ffi/revision.go
```diff
@@ -27,14 +27,26 @@ type Revision struct {
...
+	_, err := extractBytesThenFree(&val)
+	if err != nil {
+		// Any error from this function indicates that the root is inaccessible.
+		return nil, errRevisionInaccessible
```
alarso16, line 49: Another option is to return `nil, nil`, indicating that the root was well-formed, but is inaccessible. I don't think that makes more sense.

rkuris: nit: I prefer `NotFound` over `Inaccessible`. That word implies that there might be some way to access it (permissions or something) instead of "hey this thing isn't here".

rkuris (approved):

ffi/revision.go
alarso16: Fair enough - I'll change it

Issue #887, opened 2025-05-14
Update FFI tests to require ethhash and update attached static libraries to use ethhash


aaronbuchwald:
- Update golang FFI tests to include expected hashes for both with and without `ethhash` support
- Update attached static libraries to enable `ethhash` by default

PR #891, opened 2025-05-14
Update FFI tests to test both default firewood + ethhash

aaronbuchwald:
This PR
- Updates go tests in `ffi/` to use the env variable `TEST_FIREWOOD_ETHHASH=true` to determine the expected hash
- Updates CI to enable `ethhash` where needed
- Updates the attach static libraries job to include `ethhash`

Resolves https://github.com/ava-labs/firewood/issues/887

### DIFF
--- a/.github/workflows/attach-static-libs.yaml
+++ b/.github/workflows/attach-static-libs.yaml
@@ -47,8 +47,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Build for ${{ matrix.target }}
-      # TODO: add ethhash feature flag after updating FFI tests
-        run: cargo build --profile maxperf --features logger --target ${{ matrix.target }} -p firewood-ffi
+        run: cargo build --profile maxperf --features ethhash,logger --target ${{ matrix.target }} -p firewood-ffi
         
       - name: Upload binary
         uses: actions/upload-artifact@v4
@@ -150,7 +149,7 @@ jobs:
       - name: Test Go FFI bindings
         working-directory: ffi
         # cgocheck2 is expensive but provides complete pointer checks
-        run: GOEXPERIMENT=cgocheck2 go test ./...
+        run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_ETHHASH=true go test ./...
 
   remove-if-pr-only:
     runs-on: ubuntu-latest
--- a/.github/workflows/attach-static-libs.yaml
+++ b/.github/workflows/attach-static-libs.yaml
@@ -149,7 +149,7 @@ jobs:
       - name: Test Go FFI bindings
         working-directory: ffi
         # cgocheck2 is expensive but provides complete pointer checks
-        run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_ETHHASH=true go test ./...
+        run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=ethhash go test ./...
 
   remove-if-pr-only:
     runs-on: ubuntu-latest
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -221,7 +221,13 @@ func TestInvariants(t *testing.T) {
 	db := newTestDatabase(t)
 	hash, err := db.Root()
 	require.NoError(t, err, "%T.Root()", db)
-	assert.Equalf(t, make([]byte, 32), hash, "%T.Root() of empty trie")
+	expectedHash := make([]byte, 32) // Default root of an empty trie is zero-ed out.
+	if os.Getenv("TEST_FIREWOOD_ETHHASH") == "true" {
+		// Ethhash uses a special case empty root, so we set it here if enabled.
+		expectedHash, err = hex.DecodeString("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
+		require.NoError(t, err)
+	}
+	assert.Equalf(t, expectedHash, hash, "expected %x, got %x", expectedHash, hash)
 
 	got, err := db.Get([]byte("non-existent"))
 	require.NoError(t, err)
@@ -336,7 +342,13 @@ func TestDeleteAll(t *testing.T) {
 	// Check that the database is empty.
 	hash, err := db.Root()
 	require.NoError(t, err, "%T.Root()", db)
-	assert.Equalf(t, make([]byte, 32), hash, "%T.Root() of empty trie")
+	expectedHash := make([]byte, 32) // Default root of an empty trie is zero-ed out.
+	if os.Getenv("TEST_FIREWOOD_ETHHASH") == "true" {
+		// Ethhash uses a special case empty root, so we set it here if enabled.
+		expectedHash, err = hex.DecodeString("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
+		require.NoError(t, err)
+	}
+	assert.Equalf(t, expectedHash, hash, "%T.Root() of empty trie")
 }
 
 // Tests that a proposal with an invalid ID cannot be committed.
@@ -644,6 +656,7 @@ func TestProposeSameRoot(t *testing.T) {
 
 // Tests that an empty revision can be retrieved.
 func TestRevision(t *testing.T) {
+	t.Skip()
 	db := newTestDatabase(t)
 
 	keys := make([][]byte, 10)
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -217,6 +217,9 @@ func TestRangeDelete(t *testing.T) {
 }
 
 // Tests that the database is empty after creation and doesn't panic.
+//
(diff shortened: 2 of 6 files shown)


ffi/tests/eth_compatibility_test.go
```diff
@@ -37,7 +37,7 @@ func TestInsert(t *testing.T) {
 		key  common.Hash
 	}
 
-	rand.Seed(0)
+	rand := rand.New(rand.NewSource(0))
```
aaronbuchwald, line 40: fix deprecated usage warning

ffi/firewood_test.go
```diff
@@ -168,9 +172,17 @@ func TestInsert100(t *testing.T) {
...
-			// we know the hash starts with 0xf8
-			assert.Equalf(t, byte(0xf8), hash[0], "First byte of %T.Root()", db)
+
+			// Assert the hash is exactly as expected. Test failure indicates a
+			// non-hash compatible change has been made since the string was set.
+			// If that's expected, update the string to fix the test.
+			expectedHashHex := "f858b51ada79c4abeb6566ef1204a453030dba1cca3526d174e2cb3ce2aadc57"
+			if os.Getenv("TEST_FIREWOOD_ETHHASH") == "true" {
+				expectedHashHex = "c25a0076e0337d7c982c3c9dfa445c8088242a0a607f9d9def3762765bcb0fde"
+			}
```
aaronbuchwald, line 182: Inlining seems simplest, can move to improved test vectors as we improve the tests

ffi/firewood_test.go
```diff
@@ -632,6 +656,7 @@ func TestProposeSameRoot(t *testing.T) {
 
 // Tests that an empty revision can be retrieved.
 func TestRevision(t *testing.T) {
+	t.Skip()
```
aaronbuchwald, line 659: This test panics locally with `ethhash` feature enabled:

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:218:5:
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
thread caused non-unwinding panic. aborting.
SIGABRT: abort
PC=0x184a04388 m=0 sigcode=0
signal arrived during cgo execution

goroutine 7 gp=0x140000e4700 m=0 mp=0x101383700 [syscall]:
runtime.cgocall(0x1007351f4, 0x1400006b978)
	/usr/local/go/src/runtime/cgocall.go:167 +0x44 fp=0x1400006b940 sp=0x1400006b900 pc=0x1005eb874
github.com/ava-labs/firewood/ffi._Cfunc_fwd_get_from_root(0x101bb8000, {0x20, 0x14000020380}, {0x0, 0x0})
	_cgo_gotypes.go:267 +0x34 fp=0x1400006b970 sp=0x1400006b940 pc=0x10070b8b4
github.com/ava-labs/firewood/ffi.newRevision.func1(...)
	/Users/aaronbuchwald/code/ava-labs/firewood/ffi/revision.go:45
github.com/ava-labs/firewood/ffi.newRevision(0x101bb8000, {0x14000020380, 0x20, 0x20})
	/Users/aaronbuchwald/code/ava-labs/firewood/ffi/revision.go:45 +0x118 fp=0x1400006ba30 sp=0x1400006b970 pc=0x10070dd58
github.com/ava-labs/firewood/ffi.(*Database).Revision(...)
	/Users/aaronbuchwald/code/ava-labs/firewood/ffi/firewood.go:218
github.com/ava-labs/firewood/ffi.TestRevision(0x140000e4540)
	/Users/aaronbuchwald/code/ava-labs/firewood/ffi/firewood_test.go:680 +0x238 fp=0x1400006bf60 sp=0x1400006ba30 pc=0x10070abb8
testing.tRunner(0x140000e4540, 0x1011f2418)
	/usr/local/go/src/testing/testing.go:1792 +0xe4 fp=0x1400006bfb0 sp=0x1400006bf60 pc=0x10065f824
testing.(*T).Run.gowrap1()
	/usr/local/go/src/testing/testing.go:1851 +0x2c fp=0x1400006bfd0 sp=0x1400006bfb0 pc=0x1006606ac
runtime.goexit({})
```

Looking into it

aaronbuchwald: This is panicking locally with `ethhash` disabled as well, whihc makes me a bit more confused on how this started happening for me within this PR.

aaronbuchwald: Resolving this for now as it occurs on main for me locally as well, so there is something strange to debug, but it's not related to this PR and I'll continue debugging it separately.

.github/workflows/attach-static-libs.yaml
```diff
@@ -47,8 +47,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Build for ${{ matrix.target }}
-      # TODO: add ethhash feature flag after updating FFI tests
-        run: cargo build --profile maxperf --features logger --target ${{ matrix.target }} -p firewood-ffi
+        run: cargo build --profile maxperf --features ethhash,logger --target ${{ matrix.target }} -p firewood-ffi
```
alarso16, line 50: This should maybe go in the README.md for the ffi, as expecgted usage

rkuris (changes requested):
I think this means that if I now just run `cargo build --features ethhash` and `go test ./...` I will get failures, since I will have forgotten to set the required environment variable.

Is there a way to not rely on that, and instead maybe just fetch whatever the empty hash is from firewood early in the test (like right after creating the db) verifying that it's either all zeroes or the magic hash, and then setting a flag in go for that?

ffi/firewood_test.go
rkuris: Can we name these constants somewhere? Having unnamed magics is confusing.

aaronbuchwald:
> I think this means that if I now just run `cargo build --features ethhash` and `go test ./...` I will get failures, since I will have forgotten to set the required environment variable.
> 
> Is there a way to not rely on that, and instead maybe just fetch whatever the empty hash is from firewood early in the test (like right after creating the db) verifying that it's either all zeroes or the magic hash, and then setting a flag in go for that?

This would work in some cases, but not others. For example, one unit test is checking that the root of an empty database is exactly what we expect, so we can't just check that against itself and get a meaingful test.

I'd say the best two options are:
1) Use an env variable and make it very explicit
2) firewood ffi provides a function to fetch whether the feature is enabled, so we can switch on that instead of the env var

.github/workflows/attach-static-libs.yaml
```diff
@@ -47,8 +47,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
```
rkuris, line 48: This will take a lot longer unless you use the correct cache!
```suggestion
with:
          save-if: "false"
          shared-key: "ethhash"
```

aaronbuchwald: Why use `save-if: "false"` this will never write to the cache? I was meaning to ask why this is used throughout and it seems to have something to do with `default-branch-cache.yaml`, but this also uses `save-if: "false"`.

I am reading the documentation for it here, but have not tested explicitly https://github.com/Swatinem/rust-cache

.github/workflows/attach-static-libs.yaml
aaronbuchwald: Good call out. Our handling here is a bit weirder than I'd like to explain, but will add a note for clarity. Currently, `ethhash` is not enabled by default in Firewood because it's an added on feature, but we want it enabled by default for the FFI since we want to use it in coreth.

I'd love it if the Firewood binary was self-describing and could tell me whether this feature is enabled.

.github/workflows/attach-static-libs.yaml
rkuris: Answered in slack. Basically we have two caches, and we want them to contain only the results of the build, not all the subsequent steps including this one.

aaronbuchwald: Discussed offline - `attach-static-libs.yaml` will use its own cache and be separate from the cache used in `ci.yaml`.

This will default to use the `jobid` as the shared key for this action and it does not overlap with the `build` shared key in `ci.yaml`, so this is what we want.

ffi/firewood_test.go
aaronbuchwald: Updated

.github/workflows/attach-static-libs.yaml
aaronbuchwald: Added to the README https://github.com/ava-labs/firewood/pull/891/commits/ae98105060e5135c55d196e43e0916ae890eab7a

ffi/firewood_test.go
```diff
@@ -211,7 +304,11 @@ func TestInvariants(t *testing.T) {
 	db := newTestDatabase(t)
 	hash, err := db.Root()
 	require.NoError(t, err, "%T.Root()", db)
-	require.Equalf(t, make([]byte, 32), hash, "%T.Root() of empty trie", db)
+
+	emptyRootStr := expectedRoots[emptyKey]
```
alarso16, line 308: nit: This could maybe go in a helper, `checkRootHash(key string, hash []byte)`. Either way is fine though

alarso16 (approved):

PR #893, opened 2025-05-15
ci(ffi): add linter

alarso16:
We should have some set of standards for the Go code made in this folder. I "adapted" the `avalanchego` linter to use here. This is my first time messing with CI and linting, so nit reviews are especially helpful.

Any edits outside of the `.yaml` files are nit changes for linting - nothing significant.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -127,6 +127,11 @@ jobs:
         with:
           go-version-file: "ffi/go.mod"
           cache-dependency-path: "ffi/go.sum"
+      - name: Run golanci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          working-directory: ffi
       - name: Test Go FFI bindings
         working-directory: ffi
         # cgocheck2 is expensive but provides complete pointer checks
--- a/ffi/.golangci.yaml
+++ b/ffi/.golangci.yaml
@@ -1,0 +1,170 @@
+# https://golangci-lint.run/usage/configuration/
+run:
+  timeout: 10m
+
+  # If set we pass it to "go list -mod={option}". From "go help modules":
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -127,6 +127,11 @@ jobs:
         with:
           go-version-file: "ffi/go.mod"
           cache-dependency-path: "ffi/go.sum"
+      - name: Run golanci-lint
```
alarso16, line 130: Should this go with the other linting, or with the ffi stuff like it is here?

ffi/.golangci.yaml
```diff
@@ -0,0 +1,170 @@
...
+
+  # Enables skipping of directories:
+  # - vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
+  # Default: true
+  exclude-dirs-use-default: false
+
+linters:
+  disable-all: true
+  enable:
+    - asciicheck
```
alarso16, line 41: Any other linters worth adding?

.github/workflows/ci.yaml
aaronbuchwald: I think it makes sense here the way it's currently set up because we have a separate job for build, lint, and testing that's all separate.

aaronbuchwald:
Looking at the diff from AvalancheGo, what's the reason for:

- adding `dupl`
- adding `errname`

Both seem reasonable, just curious

aaronbuchwald (approved):

alarso16:
> Looking at the diff from AvalancheGo, what's the reason for:
> 
> * adding `dupl`
> * adding `errname`
> 
> Both seem reasonable, just curious

I added `dupl` and `errname` because I like the idea of `dupl`, and I made the `errname` mistake in a previous PR.

rkuris:
No opinion here, but whatever we do on avalanche-go should be done here if possible.

PR #897, opened 2025-05-17
Various github CI fixes

rkuris:
Changes:
 - Stop stripping binaries. We can revisit this if they are still too large for the firewood-go repo.
 - Create a new build type, debug with features ethhash and logger. This allows us to run some previously-missed tests, such as go tests with debug enabled.
 - Add folds to make management of the even-larger yaml file easier, usable in both vscode and vim.
 - Added significant documentation and some labels to help jump to a specific section.
 - Be more thoughtful about the test matrix. Avoided duplicate jobs, such as linting, across all the various builds.


aaronbuchwald:
Folding is a nice touch 🙏

.github/workflows/ci.yaml
```diff
@@ -8,8 +9,42 @@ on:
...
+
+#region Build Matrix
+# The build/test matrix is as follows:
+#
+#            ----------------------Profile & features-------------------------------- 
+#           +---------------------+------------------------+-------------------------+
+#           | debug & no features | debug & ethhash+logger | maxperfb ethhash+logger |
+#           +---------------------+------------------------+-------------------------+
+# CacheKey  | debug-no-features   | debug-ethhash          | maxperf-ethhash         |
+# license   | Yes (A)             | No                     | No                      |
```
aaronbuchwald, line 27: Would it make sense to use a matrix to specify the set of features + release profile throughout?

We can specify arbitrary variables for our own use, so that we don't need to replicate jobs with these two attributes as the only difference

rkuris: I tried that but it was actually less readable. Each job is subtly
different, whether it be the cache key or the command options being run.

Another thought was to generate this ci file from some code, maybe rust or
something, but that also seemed more complicated.

.github/workflows/ci.yaml
```diff
@@ -146,17 +186,154 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: "debug-ethhash"
+      - name: Check
+        run: cargo check --workspace --tests --examples --benches
+      - name: Build Debug (no features)
+        run: cargo build --workspace --tests --examples --benches
```
aaronbuchwald, line 194:
```suggestion
        run: cargo check --features ethhash --workspace --tests --examples --benches
      - name: Build Debug (no features)
        run: cargo build --features ethhash --workspace --tests --examples --benches
```

PR #901, opened 2025-05-21
feat(ffi): auto drop proposals

alarso16:
Instead of relying on the user to remember to drop proposals, they must only ensure that there are no references to the proposal stored, and the Go GC will automatically call a Finalizer function.

My main questions for this:
* Is there any other/better way to test this? The finalizer is run async, so the test is kind of flaky. Adding the timeout seems to have helped, but it could still fail.
* Should we update the go version to use the newer `runtime.AddCleanup` function, which is "less error-prone" according to the [Go docs](https://pkg.go.dev/runtime#SetFinalizer)?

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -169,10 +169,7 @@ func (db *Database) Propose(keys, vals [][]byte) (*Proposal, error) {
 	}
 
 	// The C function will never create an id of 0, unless it is an error.
-	return &Proposal{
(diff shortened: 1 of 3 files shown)


alarso16:
Although this increases ease of use, the Go GC offers no guarantees that the proposal will ever be cleaned, which is insufficient for this use case. A finalizer may be helpful to log accidental drops.

PR #910, opened 2025-05-27
feat(ffi): add proposal root retrieval

alarso16:
This is needed to enable support of Trie wrapping in geth.


ffi/src/lib.rs
```diff
@@ -570,6 +570,57 @@ fn hash(db: &Db) -> Result<Value, String> {
...
+        .map_err(|_| "proposal lock is poisoned")?;
+    let proposal = proposals
+        .get(&proposal_id)
+        .ok_or_else(|| String::from("proposal not found"))?;
+
+    // Get the root hash of the proposal.
+    proposal
+        .root_hash_sync()
+        .map_err(|e| e.to_string())?
+        .ok_or_else(|| "".to_string())
```
alarso16, line 620: It's really weird to me that this is an expected case for proposals, but not for DB. I think this makes more sense.  See [firewood.go](https://github.com/ava-labs/firewood/blob/f5efecc24c1c4ea9218ab7508f005c42cda4f203/ffi/firewood.go#L208), since the "no root" case returns an error from `root_hash_sync`

ffi/proposal.go
```diff
@@ -27,6 +27,23 @@ type Proposal struct {
 	id uint32
 }
 
+// Root retrieves the root hash of the proposal.
+// If the proposal is empty (i.e. no keys in database),
+// it returns nil, nil.
```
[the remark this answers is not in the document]
alarso16: It would make a lot more sense to match the behavior with `db.Root`, but it's not possible currently.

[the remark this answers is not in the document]
alarso16: Fixed with new implementation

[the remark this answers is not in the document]
alarso16: Requires #918

ffi/src/lib.rs
```diff
@@ -449,13 +460,20 @@ fn propose_on_proposal(
     let new_proposal = proposal.propose_sync(batch).map_err(|e| e.to_string())?;
     drop(guard); // Drop the read lock before we get the write lock.
 
+    // Get the root hash of the new proposal.
+    let mut root_hash: Value = match new_proposal.root_hash_sync().map_err(|e| e.to_string())? {
+        Some(root) => Value::from(root.as_slice()),
+        None => String::new().into(),
```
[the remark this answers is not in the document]
alarso16: Should be addressed in #918

ffi/firewood_test.go
```diff
@@ -204,13 +204,19 @@ func kvForTest(i int) KeyValue {
 // 1. By calling [Database.Propose] and then [Proposal.Commit].
 // 2. By calling [Database.Update] directly - no proposal storage is needed.
 func TestInsert100(t *testing.T) {
+	type dbView interface {
+		Get(key []byte) ([]byte, error)
```
alarso16, line 208: @aaronbuchwald I think this conflicts with something you're doing, so I can get rid of this or edit if you like

aaronbuchwald: No need to worry about a conflict here, will be easy to fix when we get to it

ffi/memory.go
```diff
@@ -27,69 +27,72 @@ type KeyValue struct {
...
 	if v == nil {
-		return errNilBuffer
+		return nil, 0, errNilBuffer
 	}
 
 	// Expected empty case for Rust's `()`
 	// Ignores the length.
 	if v.data == nil {
-		return nil
+		return nil, uint32(v.len), nil
```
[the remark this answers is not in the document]
aaronbuchwald: Seems like https://github.com/ava-labs/firewood/blob/main/ffi/src/lib.rs#L579

ffi/proposal.go
```diff
@@ -77,14 +101,15 @@ func (p *Proposal) Propose(keys, vals [][]byte) (*Proposal, error) {
...
-	id, err := extractUintThenFree(&val)
+	bytes, id, err := extractBytesAndErrorThenFree(&val)
 	if err != nil {
 		return nil, err
 	}
 
 	return &Proposal{
 		handle: p.handle,
 		id:     id,
+		root:   bytes,
```
aaronbuchwald: There are two spots we construct a new proposal instance and both come from the same val type, so may make sense to change this to one function that takes in a `val` and the handle and returns a proposal, so that we have one clear constructor rather than two places proposals can be created.

ffi/proposal.go
```diff
@@ -25,6 +25,30 @@ type Proposal struct {
...
+	}
+
+	if p.id == 0 {
+		return nil, errDroppedProposal
+	}
+
+	// If the hash is empty, return the empty root hash.
+	if p.root == nil {
+		return make([]byte, RootLength), nil
+	}
```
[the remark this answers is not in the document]
alarso16: There is an issue #918 for this in the ethhash front, but it might be better to enforce this on "normal" firewood as well.

ffi/src/lib.rs
```diff
@@ -572,13 +590,9 @@ fn hash(db: &Db) -> Result<Value, String> {
 
 /// A value returned by the FFI.
 ///
-/// This is used in several different ways:
-///
-/// - When returning data, the length is the length of the data and the data is a pointer to the data.
-/// - When returning an error, the length is 0 and the data is a null-terminated C-style string.
-/// - When returning an ID, the length is the ID and the data is null.
+/// This is used in several different ways, and should be interpreted based on the
+/// function that returned it.
```
aaronbuchwald: Is this the best way to handle this comment? Maybe it just depends on the context and it is, but not sure as I'm reading it

ffi/memory.go
```diff
@@ -27,107 +27,116 @@ type KeyValue struct {
...
+// 2.    | nil     | non-0 | proposal deleted everything
+// 3.    | non-nil | 0     | error string
+// 4.    | non-nil | non-0 | hash and id
+func hashAndIDFromValue(v *C.struct_Value) ([]byte, uint32, error) {
 	// Pin the returned value to prevent it from being garbage collected.
 	defer runtime.KeepAlive(v)
 
 	if v == nil {
-		return errNilBuffer
+		return nil, 0, errNilBuffer
```
aaronbuchwald, line 45: is this case 1 as well?

ffi/memory.go
```diff
@@ -27,107 +27,116 @@ type KeyValue struct {
...
-// 2. a non-nil error, indicating an error occurred.
-// This should only be called when the `Value` is expected to only contain an error.
-// Otherwise, an error is returned.
-func extractErrorThenFree(v *C.struct_Value) error {
+// hashAndIDFromValue converts the cgo `Value` payload into:
+//
+//	case | data    | len   | meaning
+//
+// 1.    | nil     | 0     | invalid
+// 2.    | nil     | non-0 | proposal deleted everything
```
aaronbuchwald, line 35: to make sure I understand - the proposal ID is non-zero because it's a real proposal, but the root is empty because the trie ends up being empty as well?

ffi/proposal.go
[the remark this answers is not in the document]
aaronbuchwald: Hmm I think #918 is a separate issue.

We have two choices - what to do when there's an empty root in firewood/ethhash mode and separately how to populate the golang proposal struct. My point only applies to the latter ie. if a nil root indicates that it must actually be represented by this zeroed out byte slice, then should we just populate it when creating the proposal instance?

ffi/memory.go
alarso16: This should never happen, since the FFI layer is always returned an instance of the C Value, not a pointer to it. However, I thought it would be best to keep this case to avoid panics, but I can add it to the comment

ffi/memory.go
alarso16: Yes. The proposal ID is created from a counter initialized to 1, so we are guaranteed that it's nonzero.

ffi/proposal.go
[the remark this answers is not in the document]
alarso16: Ahhh ok that makes sense. I'll address that.

ffi/proposal.go
alarso16: Good idea

PR #912, opened 2025-05-28
remove shared key usage

aaronbuchwald:


aaronbuchwald:
Removing "needs" brings the observed CI runtime down to about 6 minutes from a prior observation of about 8 minutes. There is some queuing that takes place, which appears to be a result of hitting the max allocated machines for GitHub (20 linux, 5 mac, 5 windows) [ref](https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits). This is based on observation and the documentation, but I am not sure what our GitHub plan is that may increase this limit. It's also not completely clear whether this is per repo or per organization. If the actual limit is lower or under contention across multiple PRs in this repo or across the org, then it may be more effective to optimize the order of job execution.

At least for this run: https://github.com/ava-labs/firewood/actions/runs/15303653126/job/43050637460, the observed bottleneck seems to be the longest running job (which may vary a bit as there are a few contenders). The slowest jobs appear to be the maxperf build profile and the FFI jobs which include golangci-lint.

PR #920, opened 2025-05-30
Use a Value go type to consolidate

rkuris:
- Changed naming from `extractXThenFree` to `intoX`
- Made all conversions methods on go's `Value`
- Added helpers error(), id(), and bytes() so that once we know which it is, we can call the same code
- Added some godocs for each of the `into` methods


rkuris:
Draft because Austin is working on using parts of this

PR #922, opened 2025-05-30
Windows OS support

rkuris:
Heard that we might need this, as some people want to run avalanchego on windows.

This uses the windows equivalent functions in places where we used unix specific functions.

`grpc-testtool` is no longer compiled at the top level. There are a number of unix-specific calls there, and we're currently not using it anyway.

### DIFF
--- a/ffi/tests/go.mod
+++ b/ffi/tests/go.mod
@@ -63,7 +63,7 @@ require (
 	golang.org/x/exp v0.0.0-20231110203233-9a3e6036ecaa // indirect
 	golang.org/x/sync v0.5.0 // indirect
 	golang.org/x/sys v0.30.0 // indirect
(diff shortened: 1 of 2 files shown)


alarso16 (commented):
I may be misunderstanding something - this means the maxperf allocator isn't available right? We probably need to add a comment about maxperf allocator not being available for windows

rkuris:
> I may be misunderstanding something - this means the maxperf allocator isn't available right? We probably need to add a comment about maxperf allocator not being available for windows

"maxperf" is a cargo profile we made that says "give me the maximum performance on this platform". It does several things, including turning up the optimizer.

On windows, there isn't a jemalloc implementation that is available, so we just use the system default, which is still the best we can do on that platform. The maxperf profile should still work, and the tests should cover that.

Issue #918, opened 2025-05-29
bug: If ethhash is enabled, proposals can return `None` for the root hash


rkuris:
This differs from what the DB stuff does, so we should be consistent. The DB returns an `Err()` in `Result<Option<TrieHash>, Error>`, while the proposal returns `Ok(None)`. Discussion in #910.

Likely, the solution should be (for both DB and proposals):
- Returning a `Result<TrieHash, Error>`, guaranteeing that a Hash will be formed.
- For empty roots: In ethhash, the eth empty root is returned. Otherwise, a 32-byte zero hash is returned.
- An error will only return on disk errors, etc.

PR #925, opened 2025-06-02
fix: Fix empty hash values

rkuris:
When the trie is empty, be consistent about returning the same thing. This fix makes the "same thing" be:

 - If ethhash is enabled, always return the ethereum-compatible empty hash
 - If ethhash is disabled, return None.

If ethhash is disabled, the caller can easily map None to whatever empty hash they want. In hindsight, probably ethhash should return None as well for empty hashes, but that would require the callers to know what the magic hash is for ethhash.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -155,7 +155,7 @@ where
     }
 
     async fn root_hash(&self) -> Result<Option<TrieHash>, api::Error> {
-        Ok(self.manager.read().expect("poisoned lock").root_hash()?)
+        self.root_hash_sync()
     }
 
     async fn all_hashes(&self) -> Result<Vec<TrieHash>, api::Error> {
@@ -254,7 +254,11 @@ impl Db {
 
     /// Synchronously get the root hash of the latest revision.
     pub fn root_hash_sync(&self) -> Result<Option<TrieHash>, api::Error> {
-        Ok(self.manager.read().expect("poisoned lock").root_hash()?)
+        let hash = self.manager.read().expect("poisoned lock").root_hash()?;
+        #[cfg(not(feature = "ethhash"))]
+        return Ok(hash);
+        #[cfg(feature = "ethhash")]
+        return Ok(Some(hash.unwrap_or_else(storage::empty_trie_hash)));
     }
 
     /// Synchronously get a revision from a root hash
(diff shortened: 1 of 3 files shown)


rkuris:
Fixes #918

ffi/firewood.go
```diff
@@ -197,7 +197,7 @@ func (db *Database) Root() ([]byte, error) {
 	bytes, err := bytesFromValue(&hash)
 
 	// If the root hash is not found, return a zeroed slice.
-	if err != nil && strings.Contains(err.Error(), rootHashNotFound) {
+	if err == nil && bytes == nil {
 		bytes = make([]byte, RootLength)
 		err = nil
```
alarso16: nit: `err` is already nil.
```suggestion
```

ffi/src/lib.rs
```diff
@@ -100,10 +100,12 @@ fn get_latest(db: *const DatabaseHandle, key: &Value) -> Result<Value, String> {
...
     // Find root hash.
     // Matches `hash` function but we use the TrieHash type here
-    let root = db
-        .root_hash_sync()
-        .map_err(|e| e.to_string())?
-        .ok_or_else(|| String::from("unexpected None from db.root_hash_sync"))?;
+    let Some(root) = db.root_hash_sync().map_err(|e| e.to_string())? else {
+        return Ok(Value {
+            len: 0,
+            data: std::ptr::null(),
```
alarso16: You could do something like `Ok("".into())` or `Ok(().into())`

ffi/src/lib.rs
```diff
@@ -586,8 +588,16 @@ fn root_hash(db: *const DatabaseHandle) -> Result<Value, String> {
 fn hash(db: &Db) -> Result<Value, String> {
     db.root_hash_sync()
         .map_err(|e| e.to_string())?
-        .ok_or_else(|| String::from("unexpected None from db.root_hash_sync"))
         .map(|root| Value::from(root.as_slice()))
+        .map_or_else(
+            || {
+                Ok(Value {
+                    len: 0,
```
alarso16: See above

ffi/src/lib.rs
alarso16: Or did you avoid it because it's not very readable?

rkuris: I avoided it because I didn't see it. Good callout.

alarso16 (approved):

PR #926, opened 2025-06-03
refactor(ffi): Cleanup unused and duplicate code

alarso16:
Now that we all have the functionality necessary to support coreth, I removed the unused interface, cleaned up the tests a little bit, and removed duplicate code for ffi operations.


aaronbuchwald (approved):

ffi/kvbackend.go
```diff
@@ -1,73 +0,0 @@
-package firewood
```
rkuris: I thought we needed this interface for the loader tests that Darioush had?

rkuris (approved):

ffi/kvbackend.go
alarso16: Fair enough - added back

PR #930, opened 2025-06-04
Create two pkgs for firewood differential testing against eth and merkle hashing

aaronbuchwald:


.github/workflows/ci.yaml
```diff
@@ -208,12 +208,35 @@ jobs:
...
       - name: Test Go FFI bindings
         working-directory: ffi
         # cgocheck2 is expensive but provides complete pointer checks
         run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=ethhash go test ./...
       - name: Test Ethereum hash compatability
-        working-directory: ffi/tests
+        working-directory: ffi/tests/eth
+        run: go test ./...
+
+  firewood-merkle-differential-fuzz:
```
aaronbuchwald, line 223: Add separate job since it requires a different firewood build

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+	dbUpdate:                 "dbUpdate",
+	dbBatch:                  "dbBatch",
+	checkDBHash:              "checkDBHash",
+	createProposalOnProposal: "createProposalOnProposal",
+	createProposalOnDB:       "createProposalOnDB",
+	proposalGet:              "proposalGet",
+	commitProposal:           "commitProposal",
+}
+
+func newTestDatabase(t *testing.T) *firewood.Database {
```
rkuris: nit: naming: `newFirewoodTestDatabase` or `newFwTestDb`

Important because we have two different databases here and I can't easily tell which this is for.

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+	dbFile := filepath.Join(t.TempDir(), "test.db")
+	db, closeDB, err := newDatabase(dbFile)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		require.NoError(t, closeDB())
+	})
+	return db
+}
+
+func newDatabase(dbFile string) (*firewood.Database, func() error, error) {
```
rkuris: nit: naming: `newFirewoodDatabase`

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+	return f, f.Close, nil
+}
+
+type tree struct {
+	require *require.Assertions
+	rand    *rand.Rand
+
+	id       int
+	nextID   int
+	merkleDB merkledb.MerkleDB
```
rkuris, line 79: Why is firewood a pointer but not MerkleDB?

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+	for proposalID := range tr.proposals {
+		proposalIDs = append(proposalIDs, proposalID)
+	}
+	slices.Sort(proposalIDs)
+	selectedProposalID := tr.rand.Intn(len(proposalIDs))
+	return tr.proposals[selectedProposalID]
+}
+
+func (tr *tree) dropAllProposals() {
+	// Drop all firewood proposals out of order (allowed by firewood).
```
rkuris, line 131: ... but not by merkleDB? Curious as to why you mentioned this was allowed.

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+		tr.require.NoError(merkleErr)
+		tr.require.NoError(fwdErr)
+		tr.require.Equal(fwdVal, merkleVal)
+	}
+}
+
+func (tr *tree) dbBatch() {
+	batchSize := tr.rand.Intn(maxBatchSize) + 1 // ensure at least one key-value pair
+	keys, vals := tr.createRandomBatch(batchSize)
+	// Insert the batch of key-value pairs into both databases.
```
rkuris: nit: this comment should be at the function level

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+	if pr == nil {
+		return
+	}
+
+	keys := tr.selectRandomKeys(maxBatchSize)
+	for _, key := range keys {
+		fwdVal, fwdErr := pr.fwdView.Get(key)
+		merkleVal, merkleErr := pr.merkleView.GetValue(context.Background(), key)
+
+		// XXX: handling of non-existent keys diverges between merkleDB and firewood
```
rkuris: same comment and code earlier; can this be refactored into a helper?

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+
+func (tr *tree) commitProposal() {
+	if len(tr.children) == 0 {
+		return // no proposals to commit
+	}
+	tr.commitRandomProposal()
+}
+
+func fuzzTree(t *testing.T, randSource int64, byteSteps []byte) {
+	rand := rand.New(rand.NewSource(randSource))
```
rkuris, line 340: Do we have some way of "remembering" this random value, so we can recreate the same run again if it fails?

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+
+	if len(byteSteps) > 100 {
+		byteSteps = byteSteps[:100] // limit the number of steps to 100
+	}
+
+	tr := newTestTree(t, rand)
+
+	// TODO: replace randomly generated values with bytes from the fuzzer
```
rkuris, line 348: Hmm, maybe the fuzzer will handle the prior comment

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,442 @@
...
+	for _, step := range byteSteps {
+		step = step % maxStep
+		// Make this two lines so debugger displays the stepStr value
```
rkuris, line 351: Curious as to why this is needed, isn't the t.Log of it sufficient?

rkuris (approved):
Bunches of questions and a few nits, otherwise seems okay.

My go-fu is weak so another reviewer is recommended.

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: MerkleDB is an interface

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: When you are running the debugger, you can see all of the variables at any breakpoint.

Separating this into two lines, so that we have a local variable makes it so you can set a breakpoint or watch that variable and more easily step through and see what step is running.

This is very much to make using the debugger as easy as possible. Happy to remove if preferred.

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: Right now the fuzzer provides the random seed, which means that if it inputs random seed 0 we will get a deterministic sequence of events. This ensures that when we re-run a given seed, as long as we always use random values derived from that seed (and in order), then we'll get the same sequence of events.

However, this does not allow the fuzzer to intelligently assign a sequence of inputs. Changing the value of that random seed input will produce a different execution result, but it will not do so in a way that allows the fuzzer to discover more "interesting cases". 

For example, if we took all of the steps and their inputs directly from the fuzzer, then the fuzzer may be able to try useful values like all zeroes for certain key-value pairs, larger key-value pairs, etc. Since it is only able to input the random seed we get a deterministic sequence of inputs, but we do not give the fuzzer as much of a chance to discover interesting cases.

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: Yes, if the golang fuzzer discovers a failure case, it automatically writes it into `testdata/`, so that you can run it again with `go test -run=FuzzTestName/fuzzseed`.

One caveat here is that if you run it within CI, then you must make sure that the automatically saved seeds are saved to a format that we can access. I'd say that can be out of scope for this PR, but happy to update CI to make sure that any failing seeds are pushed as an artifact. Should just need to use `upload-artifact` on failure.

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: MerkleDB does not require explicitly dropping merkle views as they are simply garbage collected, but firewood requires they are explicitly dropped, so that the FFI can tell Firewood when to drop the associated memory.

It's non-obvious to me that it would be safe to drop them out of order given they are stacked on top of each other, so I wanted to put a comment since I drop them in a random order by iterating over the proposal map here (maps provide random iteration in go). Since there's no need to drop merkle proposals at all, I didn't mention MerkleDB here.

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: Going to remove the comments as they should be self-explanatory

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: We could factor even more out more easily if we adopted a View interface, so that we could pass in different view types and execute the same test logic, but for now I will add a helper function to perform the assertion on results from firewood/merkle

ffi/tests/firewood/merkle_compatibility_test.go
alarso16: I think a comment would be nice saying that MerkleDB will use the GC

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+func (tr *tree) createProposalOnProposal() {
+	pr := tr.selectRandomProposal()
+
+	if pr == nil {
+		return
+	}
+
+	batchSize := tr.rand.Intn(maxBatchSize) + 1 // ensure at least one key-value pair
```
alarso16, line 248: Do we need to guarantee there's at least one key-value pair? I know it makes a lot fewer edge cases, but seems like it should be tested

alarso16: Just to note: in coreth, this won't happen. statedb won't attempt to commit the proposal, so we just have to adjust block numbers in the `ProposalContext`.

ffi/tests/firewood/merkle_compatibility_test.go
alarso16: It's a little weird to me, especially since the next step in the debugger will tell you which case you're looking at. If you have a strong preference it's fine

alarso16 (approved):

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: I find it helpful in practice, makes it easy to leave one breakpoint here

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: Currently, firewood returns an error if you attempt to create an empty revision, so I need to either avoid that case or change the check to assert it returns the expected error. I preferred the former here and think it makes more sense to go this way.

I would think we _should_ handle this as opposed to leaving to the caller to guarantee a non-empty batch. For now, I'll add `XXX` to the comment to call attention to it as a case that may change in the future.

ffi/tests/firewood/merkle_compatibility_test.go
aaronbuchwald: There's a comment here where we free the pointers to the merkle views allowing them to be GCed, but I'll rephrase to make this more clear that this is the equivalent of explicitly dropping them: https://github.com/ava-labs/firewood/blob/b3d0b28ef21607dd4466f5c4665ad6579953c01e/ffi/tests/firewood/merkle_compatibility_test.go#L135

aaronbuchwald: Updated to:

```
	// Free pointers at the root to allow garbage collection.
	// MerkleDB does not require explicitly dropping merkle views, so clearing the
	// pointers at the root is sufficient.
```

PR #936, opened 2025-06-05
feat(checker): Firewood checker framework

qusuyan:


qusuyan:
This diff is NOT ready for merge - just here to test with conventional commit requirements

storage/src/checker/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/// Recursively traverse the trie from the given root address.
+async fn traverse_trie(
+    node_store: &NodeStore<Committed, FileBacked>,
+    subtree_root_address: Option<LinearAddress>,
+    visited: &mut LinearAddressRangeSet,
+) -> Result<(), CheckerError> {
+    let Some(root_address) = subtree_root_address else {
+        // empty subtree, do nothing
```
rkuris: It's possible the high water mark is above or below the expected value here

storage/src/checker/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+use range_set::LinearAddressRangeSet;
+
+pub use error::CheckerError;
+
+/// Go through the filebacked storage and check for any inconsistencies. It proceeds in the following steps:
+/// 1. traverse the trie and check the nodes
+/// 2. check the free list
+/// 3. check any bubbles - what are the spaces between trie nodes and free lists?
+// TODO: add merkle hash checks as well
+pub async fn check_node_store(
```
rkuris: This should be a method on a `NodeStore` instead of a standalone function to call, something like:

```rust
impl NodeStore<Committed, FileBacked> {
    fn check() -> Result<(), CheckerError> {
...
```

storage/src/checker/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    node_store: &NodeStore<Committed, FileBacked>,
+) -> Result<(), CheckerError> {
+    let mut visited = LinearAddressRangeSet::new(node_store.storage.get_file_size()?)?;
+
+    // 1. traverse the trie and check the nodes
+    let root_address = node_store.root_address_and_hash()?.map(|(addr, _)| addr);
```
rkuris: Not sure I love getting the address and hash here, since the implementation reads the root node, and then you read it again. Options:
 - Maybe pass a `Option<SharedNode>` down instead. You can get the root one from the `root_node` method on the `NodeStore`.
 - Maybe create a `root_address` method next to `root_address_and_hash` which just returns the root address.

storage/src/checker/range_set.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    pub(super) fn insert_area(
+        &mut self,
+        addr: LinearAddress,
+        size: u64,
+    ) -> Result<(), CheckerError> {
+        let start = addr.get();
+        // end can be larger than the file size: only the node in the stored area is written to disk
+        if start < NodeStoreHeader::SIZE || start > self.size {
```
rkuris: Should this be `start + size > self.size`?

storage/src/checker/range_set.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use btree_range_map::{AnyRange, AsRange, RangeSet};
+use std::num::NonZeroU64;
+use std::ops::Bound;
+
+use crate::nodestore::NodeStoreHeader;
+use crate::{CheckerError, LinearAddress};
+
+pub(super) struct LinearAddressRangeSet {
+    range_set: RangeSet<u64>,
```
rkuris: Would be great if we could use a`RangeSet<LinearAddress>` here. That would save some error checks later and make insertion easier since all addresses are essentially NonZeroU64s.

One snag is you might have to implement the `Measure` trait on LinearAddress/NonZerU64. I bet the crate owner would take a PR to add the non-zero types to the list of those that implement `Measure`...

storage/src/checker/range_set.rs
```diff
@@ -0,0 +1,169 @@
...
+    >;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.range_set.into_iter().map(|range| {
+            let Bound::Included(start) = range.start() else {
+                panic!("start of range is not included, this should not happen");
+            };
+            let Bound::Excluded(end) = range.end() else {
+                panic!("end of range is not excluded, this should not happen");
+            };
```
rkuris: I'm confused why we need these checks and panics. Can't we get an iterator that returns a `RangeFromExcludedToIncluded`? https://docs.rs/btree-range-map/0.7.2/btree_range_map/struct.RangeFromExcludedToIncluded.html seems to have one, not sure why these aren't being returned from `into_iter`.

storage/src/checker/range_set.rs
```diff
@@ -0,0 +1,169 @@
...
+        let mut visited = LinearAddressRangeSet::new(0x1000).unwrap();
+        visited
+            .insert_area(start1_addr, size1)
+            .expect("the given area should be within bounds");
+
+        let error = visited
+            .insert_area(start2_addr, size2)
+            .expect_err("the given area should intersect with the first area");
+
+        let CheckerError::AreaIntersects {
```
rkuris: These are easier to write with assert_matches

https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html

storage/src/linear/filebacked.rs
```diff
@@ -106,6 +106,10 @@ impl FileBacked {
             ring: ring.into(),
         })
     }
+
+    pub(crate) fn get_file_size(&self) -> Result<u64, Error> {
+        Ok(self.fd.metadata()?.len())
```
rkuris: The header has a high-water mark which should be used instead of this, as it is certain to be wrong. Lets say a node is allocated at size 2k but we only write 1.9k of data. The file will be shorter than the nodestore header says it is.
```suggestion
        self.header.size
```

rkuris (changes requested):

storage/src/checker/mod.rs
qusuyan: The problem only exists if we need to access the address from an external file: when we move everything into `nodestore.rs`, we have direct access to `node_store.header.root`.

storage/src/checker/range_set.rs
qusuyan: I think the best way to put it is that I am not sure what's the best interface here. I tried to return something similar to a `StoredArea` but maybe this is not the best thing to do since what we are really interested is the space between visited area. Let me think about this again.

qusuyan: By the way, the checks are there since we only insert ranges with start addr included but end addr excluded. There is no need to cast it to anything else.

storage/src/checker/range_set.rs
qusuyan: This is nightly only

storage/src/checker/mod.rs
qusuyan: Can you elaborate a bit more here? Why is the high water mark a problem during trie traversal?

storage/src/checker/range_set.rs
qusuyan: https://github.com/timothee-haudebourg/range-traits/pull/5

storage/src/checker/mod.rs
rkuris: If the database is empty, then the file size as recorded in the header may still be wrong.

qusuyan: How exactly? Will it be less than the header size? Or will it be larger in which case everything that follows should be in the freelist?

rkuris: These comments assume an empty database.

I *think* (please verify) that a correctly created empty database with no free list has the high water mark at 2048. If there is a free list, then the high water mark must be higher. Let's say the 2k free list has one item at address 2048, then the high water mark should be 4096.

Any *lower* value is a critical corruption waiting to happen, as the next block will be allocated on top of the free list, which will lead to corruption. A higher value means space between the end of the last item on the free list and the value of the high water mark will become a hole on the next allocation.

storage/src/checker/range_set.rs
rkuris: Whenever I see panics, particularly when they are checking some value, my first inclination is to see if the type system can be leveraged to avoid the problem in the first place. Maybe it's not possible in this case.

storage/src/checker/range_set.rs
rkuris: Yes, but you can still do it with:

```rust
assert!(matches!(...))
```

storage/src/checker/mod.rs
qusuyan: That is my understanding as well - currently I only check if the size at the header is at least 2048. It will report an error if that happens. If it is bigger and it has a free list, step 3 (traversing the free lists) will handle that; if it does not have a freelist, the range set will catch it in step 4 (checking the bubbles). 

If there is an area that extends beyond the watermark, the range set will also catch that, so I am not too worried about the watermark being wrong :)

storage/src/checker/range_set.rs
qusuyan: We can avoid panics by taking the value out regardless if its an exclusion bound or an inclusion bound, but I think it is good to have another check to make sure that the code is doing the right thing. I'm not sure what you can do with the type system here since its an enum wrapper around the bound values...

storage/src/checker/range_set.rs
qusuyan: I don't think `matches!` also lets you check the values inside? Part of reason I am doing this now is because I want to take the values in the error type out and check they are correct next.

fwdctl/Cargo.toml
```diff
@@ -6,13 +6,15 @@ rust-version = "1.85.0"
...
 firewood = { version = "0.0.5", path = "../firewood" }
+storage = { version = "0.0.5", path = "../storage" }
 clap = { version = "4.5.0", features = ["cargo", "derive"] }
 env_logger = "0.11.2"
 log = "0.4.20"
 tokio = { version = "1.36.0", features = ["full"] }
 futures-util = "0.3.30"
 hex = "0.4.3"
 csv = "1.3.1"
+rand = "0.9.1"
```
rkuris: rand should only be used for testing, so move this to the dev-dependencies section below

fwdctl/tests/cli.rs
```diff
@@ -539,6 +540,61 @@ fn fwdctl_dump_with_hex() -> Result<()> {
...
+        Command::cargo_bin(PRG)?
+            .arg("insert")
+            .args([key.clone()])
+            .args([value])
+            .args(["--db"])
+            .args([tmpdb::path()])
+            .assert()
+            .success()
+            .stdout(predicate::str::contains(&key));
+    }
```
rkuris: This will be wicked slow I think. It might be smarter to use the firewood API and create a database and insert rows that way.

storage/src/lib.rs
```diff
@@ -73,3 +76,38 @@ pub fn empty_trie_hash() -> TrieHash {
...
         .expect("empty trie hash is 32 bytes")
 }
+
+/// Errors returned by the checker
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum CheckerError {
+    /// The root node was not found
+    #[error("root node not found")]
+    RootNodeNotFound,
```
rkuris: This error should include the error that read_node returned or some other supplemental information

storage/src/lib.rs
```diff
@@ -73,3 +76,38 @@ pub fn empty_trie_hash() -> TrieHash {
...
+
+    /// The file size is not valid
+    #[error("the db size ({0}) is invalid")]
+    InvalidDBSize(u64),
```
rkuris: Consider specifying the correct value for the database size here as well

storage/src/lib.rs
```diff
@@ -73,3 +76,38 @@ pub fn empty_trie_hash() -> TrieHash {
...
+
+    /// The address is out of bounds
+    #[error("stored area at {start} with size {size} is out of bounds")]
+    AreaOutOfBounds {
+        /// Start of the StoredArea
+        start: LinearAddress,
+        /// Size of the StoredArea
+        size: u64,
+    },
```
rkuris, line 110: add the bound

storage/src/lib.rs
```diff
@@ -73,3 +76,38 @@ pub fn empty_trie_hash() -> TrieHash {
...
+
+    /// Stored areas intersect
+    #[error("stored area at {start} with size {size} intersects with another stored area")]
```
rkuris: probably should indicate which stored area it overlaps with

storage/src/nodestore.rs
```diff
@@ -1313,6 +1318,74 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        let mut visited = LinearAddressRangeSet::new(self.header.size)?;
+
+        // 2. traverse the trie and check the nodes
+        let root_address = self.header.root_address;
+        self.traverse_trie(root_address, &mut visited).await?;
+
+        // 3. check the free list - this can happen in parallel with the trie traversal
+
+        // 4. check any bubbles - what are the spaces between trie nodes and free lists?
+        let _ = visited.complement(); // TODO
```
rkuris: nit: naming: maybe `find_missing_space`

storage/src/nodestore.rs
```diff
@@ -1313,6 +1318,74 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        };
+
+        let (_, area_size) = self.area_index_and_size(root_address)?;
+        visited.insert_area(root_address, area_size)?;
+
+        let node = self.read_node(root_address)?;
+
+        match &*node {
+            Node::Branch(branch) => {
+                for child in &branch.children {
```
rkuris: Create a better iterator here, something like (untested):
```rust
for (addr, hash) in branch.children.iter().filter_map(|child| match child {
    None => None,
    Some(Child::AddressWithHash(addr, hash) => Some((addr, hash)),
    _ => unreachable!("All children are persisted"),
})
```

storage/src/range_set.rs
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        if self.range_set.intersects(start..end) {
+            return Err(CheckerError::AreaIntersects { start: addr, size });
+        }
+        self.range_set.insert_range(start..end);
+        Ok(())
+    }
+
+    #[expect(clippy::unwrap_used)]
```
rkuris: where is unwrap being used?

storage/src/nodestore.rs
```diff
@@ -1313,6 +1318,74 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+                    let child_address = child.as_ref().map(|c| {
+                        let Child::AddressWithHash(address, _) = c else {
+                            panic!("the children is not persisted yet, this should not happen");
+                        };
+                        *address
+                    });
+                    Box::pin(self.traverse_trie(child_address, visited)).await?;
```
demosdemon: I'm curious if this really needs to be boxed or if [this](https://doc.rust-lang.org/std/pin/macro.pin.html) would work instead 

```suggestion
                    let fut = self.traverse_trie(child_address, visited);
                    let fut = std::pin::pin!(fut);
                    fut.await?;
```

(the pinned fut may or may not to be `mut`).

storage/src/lib.rs
qusuyan: `read_node` returns an `std::io::Error` so it will become `CheckerError::IO`. This error was used if the `root_address` in header is `None` but now the checker just assumes the database is empty in that case. I will remove this error.

storage/src/lib.rs
qusuyan: This one is hard to do - adjacent stored areas are coalesced during bookkeeping so it is hard to find the specific area it overlaps with.

storage/src/nodestore.rs
qusuyan: I will rename the variables when we finish with all the traversals. Currently this line is there to avoid unused code warnings.

storage/src/range_set.rs
qusuyan: Not any more! I must have removed it in one of my previous commits

storage/src/nodestore.rs
qusuyan: I tried and the compiler still complains about recursive async call, so will keep it as is :)

fwdctl/tests/cli.rs
qusuyan: I wanna keep one large test here as an e2e testing. I will add more thorough tests in the storage crate. Would it cause some additional problems?

rkuris: Answered in person. We also need to make sure the data is repeatable.

firewood/src/v2/api.rs
```diff
@@ -8,7 +8,7 @@ use async_trait::async_trait;
 use futures::Stream;
 use std::fmt::Debug;
 use std::sync::Arc;
-use storage::{FileIoError, TrieHash};
+use storage::{CheckerError, FileIoError, TrieHash};
```
rkuris: This api.rs file is only for errors exposed to the end user, so we should check that it has to be here. I think we can just have an internal error for this.

fwdctl/tests/cli.rs
```diff
@@ -539,6 +539,74 @@ fn fwdctl_dump_with_hex() -> Result<()> {
     Ok(())
 }
 
+#[test]
+#[serial]
+fn fwdctl_check_empty_db() -> Result<()> {
```
rkuris, line 524: just return Result<Box<dyn Error>> and then use ? to map the error for you

fwdctl/tests/cli.rs
```diff
@@ -539,6 +539,74 @@ fn fwdctl_dump_with_hex() -> Result<()> {
...
+    eprintln!("Seed {seed}: to rerun with this data, export FIREWOOD_TEST_SEED={seed}");
+    let mut rng = StdRng::seed_from_u64(seed);
+
+    Command::cargo_bin(PRG)?
+        .arg("create")
+        .arg(tmpdb::path())
+        .assert()
+        .success();
+
+    for _ in 0..100 {
```
rkuris: create (or find) an issue to add import functionality (counterpart to dump). Add a TODO here so that this can be revisited later. Reduce the number of rows so this takes <1 sec

fwdctl/tests/cli.rs
```diff
@@ -539,6 +539,74 @@ fn fwdctl_dump_with_hex() -> Result<()> {
...
+        let key = format!("key_{}", rng.random::<u64>());
+        let value = format!("value_{}", rng.random::<u64>());
+        Command::cargo_bin(PRG)?
```
rkuris, line 568: Lets see if we can eliminate this

storage/src/lib.rs
```diff
@@ -73,3 +77,55 @@ pub fn empty_trie_hash() -> TrieHash {
...
+
+/// Errors related to the size of the db
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum DBSizeError {
+    /// The db size cannot be 0
+    #[error("db size cannot be 0")]
+    Zero,
+    /// The db size cannot be smaller than the file size
+    #[error("db size cannot be smaller than the file size ({0})")]
```
rkuris: You need both the size reported by the OS and the size from the header

storage/src/lib.rs
```diff
@@ -73,3 +77,55 @@ pub fn empty_trie_hash() -> TrieHash {
...
+    IO(#[from] FileIoError),
+}
+
+/// Errors related to the size of the db
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum DBSizeError {
+    /// The db size cannot be 0
+    #[error("db size cannot be 0")]
+    Zero,
```
rkuris: This is the same as SmallerThanHeaderSize

storage/src/linear/filebacked.rs
```diff
@@ -121,6 +121,13 @@ impl FileBacked {
             ring: ring.into(),
         })
     }
+
+    pub(crate) fn get_file_size(&self) -> Result<u64, FileIoError> {
+        self.fd
+            .metadata()
+            .map(|m| m.len())
+            .map_err(|e| self.file_io_error(e, 0, None))
```
rkuris: The context should be "metadata"

storage/src/nodestore.rs
```diff
@@ -46,7 +47,8 @@ use std::sync::Arc;
 use crate::hashednode::hash_node;
 use crate::node::{ByteCounter, Node};
 use crate::{
-    CacheReadStrategy, Child, FileBacked, HashType, Path, ReadableStorage, SharedNode, TrieHash,
+    CacheReadStrategy, CheckerError, Child, DBSizeError, FileBacked, HashType, Path,
```
rkuris: Let's combine these two errors, as they are both CheckerErrors.

storage/src/nodestore.rs
```diff
@@ -1428,6 +1430,68 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
     }
 }
 
+/// NodeStore checker
+impl NodeStore<Committed, FileBacked> {
```
rkuris: I think we can use this:
```suggestion
impl<S: WritableStorage> NodeStore<Committed, S> {
```
or maybe ReadableStorage

storage/src/nodestore.rs
```diff
@@ -1428,6 +1430,68 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+    /// Go through the filebacked storage and check for any inconsistencies. It proceeds in the following steps:
+    /// 1. Check the header
+    /// 2. traverse the trie and check the nodes
+    /// 3. check the free list
+    /// 4. check missed areas - what are the spaces between trie nodes and free lists we have traversed?
+    // TODO: add merkle hash checks as well
+    pub async fn check(&self) -> Result<(), CheckerError> {
+        // 1. Check the header
+        let file_size = self.storage.get_file_size()?;
+        if self.header.size < file_size {
```
rkuris: Can we do something fancier here by mapping the boolean into a Result? Something like:
(self.header.size < file_size).or_else(CheckerError::InvalidDBSize)?

It's okay to close this comment

firewood/src/v2/api.rs
qusuyan: Talked in person: we need it here because `fwdctl` returns an api error. I think if we want to run checker on request by the end user in the future, it can still make sense?

fwdctl/tests/cli.rs
qusuyan: Talked in person: `map_err` doesn't do anything so I have removed it for all test cases in this file :)

storage/src/nodestore.rs
qusuyan: Talked in person - I want to keep the if condition on a separate line from what the error really is.

fwdctl/src/check.rs
```diff
@@ -0,0 +1,42 @@
...
+        value_name = "DB_NAME",
+        default_value_t = String::from("firewood"),
+        help = "Name of the database"
+    )]
+    pub db: String,
+}
+
+pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
+    let db_path = Path::new(&opts.db);
+    let node_cache_size = NonZeroUsize::new(1).expect("this will never crash");
```
rkuris: Found this nifty crate to simplify this to:
```suggestion
    let node_cache_size = nonzero!(1usize);
```
https://docs.rs/nonzero_ext/latest/nonzero_ext/macro.nonzero.html#examples

fwdctl/src/check.rs
```diff
@@ -0,0 +1,42 @@
...
+        db_path.to_path_buf(),
+        node_cache_size,
+        free_list_cache_size,
+        false,
+        CacheReadStrategy::WritesOnly, // cache none since this is a read-only workload - we don't want to cache any nodes since we won't read a node more than once
+    )?);
+
+    let node_store = NodeStore::open(storage)?;
+    node_store.check().await?;
+    Ok(())
```
rkuris: This might work, but up to you if it's cleaner or not.
```suggestion
    node_store.check().await.map_err(Into::into)
```

storage/src/nodestore.rs
```diff
@@ -1428,6 +1430,73 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+
+        if let Node::Branch(branch) = self.read_node(subtree_root_address)?.as_ref() {
+            // this is an internal node, traverse the children
+            let child_addrs = branch.children.iter().filter_map(|child| match child {
+                None => None,
+                Some(Child::AddressWithHash(address, _)) => Some(*address),
+                _ => panic!("the child is not persisted yet, this should not happen"),
+            });
+
+            for child_addr in child_addrs {
```
rkuris:
```suggestion
            for child_addr in branch.children_iter() {
```

storage/src/nodestore.rs
```diff
@@ -1569,3 +1638,115 @@ mod tests {
...
+    ) -> Result<usize, FileIoError> {
+        let node_size = NodeStore::<Arc<ImmutableProposal>, FileBacked>::stored_len(node);
+        let (area_index, node_area) = AREA_SIZES
+            .iter()
+            .enumerate()
+            .find(|(_, size)| **size >= node_size)
+            .unwrap();
+        let mut stored_area_bytes = Vec::with_capacity(*node_area as usize);
+        node.as_bytes(area_index as u8, &mut stored_area_bytes);
+        file_backed.write(offset, &stored_area_bytes)?;
```
rkuris: This is a lot of work, and allocates space that will never get used. probably easier to leverage area_size_to_index.

I'd say the algorithm should be:
 - Call node.as_bytes() to get the length
 - Call filebacked_write(offset, the_bytes)
 - Call area_size_to_index(the_bytes.len())
 - Return AREA_SIZES[whatever_area_size_to_index returned]
 
This will make you immune to changes to the way allocations work in the future.

storage/src/nodestore.rs
```diff
@@ -1569,3 +1638,115 @@ mod tests {
...
+        let ethhash = 1;
+        let header = NodeStoreHeader {
+            version: Version::new(),
+            endian_test: 1,
+            size,
+            free_lists: Default::default(),
+            root_address: Some(root_addr),
+            area_size_hash: area_size_hash().as_slice().try_into().unwrap(),
+            ethhash,
+        };
```
rkuris: Let's use NodeStoreHeader::new() and mutate it, making you immune to changes to the header later. It correctly sets the ethhash value for you. All this code then becomes:

```rust
let mut header = NodeStoreHeader::New();
header.root_address = Some(root_addr);
```

storage/src/nodestore.rs
```diff
@@ -1569,3 +1638,115 @@ mod tests {
...
+        let leaf = Node::Leaf(LeafNode {
+            partial_path: Path::from([0, 1]),
+            value: Box::new([3, 4, 5]),
+        });
+        let leaf_addr = LinearAddress::new(offset).unwrap();
+        let leaf_area = write_new_node(&file_backed, &leaf, offset).unwrap();
+        offset += leaf_area as u64;
+
+        let mut branch_children: [Option<Child>; BranchNode::MAX_CHILDREN] = Default::default();
+        branch_children[1] = Some(Child::AddressWithHash(leaf_addr, HashType::default()));
```
rkuris: Curious as to why this didn't fail the clippy check for indexing. It should have, and you should have needed an `#[allow(clippy::indexing_slicing)]`. Please see if you can figure out why this warning is not enabled, and enable it for the non-test code.

storage/src/nodestore.rs
```diff
@@ -1569,3 +1638,115 @@ mod tests {
...
+        node.as_bytes(area_index as u8, &mut stored_area_bytes);
+        file_backed.write(offset, &stored_area_bytes)?;
+        Ok(*node_area as usize)
+    }
+
+    fn write_header(
+        file_backed: &FileBacked,
+        root_addr: LinearAddress,
+        size: u64,
+    ) -> Result<(), FileIoError> {
```
rkuris: This is just a test helper, so it shouldn't have a return value and should just panic if the write fails. That eliminates a bunch of unwraps at each callsite.

storage/src/range_set.rs
```diff
@@ -0,0 +1,598 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
rkuris, line 1: This needs a top-level comment saying where this originally came from and an overview of the modifications we made from it. It's possible we can't relicense it here.

rkuris (changes requested):

fwdctl/tests/cli.rs
qusuyan: #966

storage/src/nodestore.rs
qusuyan: `children_iter()` gives a iterator over the idx and the hash of children existing in the node. To get the address I need to go back to the list of children with the returned idx anyways. Let me rearrange the code in another way...

storage/src/nodestore.rs
qusuyan: Not exactly following here: why does this code snippet allocate space that will never get used? And doesn't the `area_size_to_index` do something similar to what we are doing here?

But I agree that immunity to future changes is good - let me replace the iterator here with `area_size_to_index` :)

storage/src/nodestore.rs
rkuris: Ah, right, you need something different. Two options:

1. Create a two iterators for BranchNode: `children_hashes` and `children_addresses`, changing existing callers of `children_iter` to `children_hashes`.
2. Change `children_iter` to be a bit more general and return `impl Iterator<Item = (usize, (LinearAddress, &HashType))>` and change the 3 existing callers to do `children_iter().map(|(offset, (addr, hash))| (offset, hash))`

storage/src/nodestore.rs
rkuris: This code preallocates some AREA_SIZE value for `stored_area_size`, but you'll only ever write `node_size` bytes to `stored_area_size`. If you were to fix this code, you should only allocate `node_size` instead of `node_area` on line 1666, but as mentioned you shouldn't fix this code, you should instead use the algorithm I described above anyway.

storage/src/range_set.rs
qusuyan: I might need some enlightenment on copyrights but for this specific file I rewrite entire thing myself and it is completely different from how `btree-range-map` implements this, so we are probably safe here.

storage/src/nodestore.rs
qusuyan: ~~Could this be the reason? https://github.com/rust-lang/rust-clippy/issues/8348~~

Never mind. `storage/Cargo.toml` did not specify any clippy rules at all. Putting the clippy rules in will generate tons of warnings so I will probably put it into a different PR :)

rkuris: I only care about the ones you introduced, so just add `!#[deny(clippy::indexing_slicing)]` to the top of any new files. Otherwise, you're creating more work in https://github.com/ava-labs/firewood/pull/957 which is nearing completion and probably enables those warnings.

storage/src/range_set.rs
rkuris: If it's not derived from anything, then great! It will need a more thorough review on my next pass though.

storage/src/range_set.rs
```diff
@@ -0,0 +1,598 @@
...
+
+use btree_slab::BTreeMap;
+use btree_slab::generic::map::BTreeExt;
+use btree_slab::generic::node::Address;
+use std::ops::Range;
+
+use crate::nodestore::STORAGE_AREA_START;
+use crate::{CheckerError, LinearAddress};
+
+pub struct RangeSet<T> {
```
rkuris: `#[derive(Debug)]` -- please enable full clippy warnings here (even pedantic) and correct any errors. Not commenting on additional likely warnings.

Add `!#[warn(clippy::pedantic)]` to the top of this file.

storage/src/range_set.rs
```diff
@@ -0,0 +1,598 @@
...
+    index: BTreeMap<T, T>, // start --> end
+}
+
+struct CoalescingRanges<'a, T> {
+    prev_consecutive_range: Option<Range<&'a T>>,
+    intersecting_ranges: Vec<Range<&'a T>>,
+    next_consecutive_range: Option<Range<&'a T>>,
+}
+
+#[allow(dead_code)]
```
rkuris, line 24: why is this needed?

rkuris:
Depends on: #957

fwdctl/src/check.rs
```diff
@@ -0,0 +1,41 @@
...
+        required = false,
+        value_name = "DB_NAME",
+        default_value_t = String::from("firewood"),
+        help = "Name of the database"
+    )]
+    pub db: String,
+}
+
+pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
+    let db_path = Path::new(&opts.db);
```
rkuris: why not:
```suggestion
    let db_path = PathBuf::new(&opts.db);
```

fwdctl/src/check.rs
```diff
@@ -0,0 +1,41 @@
...
+    let node_cache_size = nonzero!(1usize);
+    let free_list_cache_size = nonzero!(1usize);
+
+    let storage = Arc::new(FileBacked::new(
+        db_path.to_path_buf(),
+        node_cache_size,
+        free_list_cache_size,
+        false,
+        CacheReadStrategy::WritesOnly, // cache none since this is a read-only workload - we don't want to cache any nodes since we won't read a node more than once
```
rkuris: nit: remove "cache none" from the comment, as that's obvious

fwdctl/src/check.rs
```diff
@@ -0,0 +1,41 @@
...
+    )?);
+
+    let node_store = NodeStore::open(storage)?;
+    node_store.check().await.map_err(Into::into)
```
rkuris: nit:
```suggestion
    let node_store = NodeStore::open(storage)?.check().await.map_err(Into::into)
```

fwdctl/tests/cli.rs
```diff
@@ -534,7 +533,76 @@ fn fwdctl_dump_with_hex() -> Result<()> {
         .stdout(predicate::str::contains("--start-key=c"))
         .stdout(predicate::str::contains("--start-key-hex=63"));
 
-    fwdctl_delete_db().map_err(|e| anyhow!(e))?;
+    fwdctl_delete_db()?;
+
+    Ok(())
```
rkuris: nit: trailing Ok(()):
```suggestion
    fwdctl_delete_db()
```

storage/src/node/branch.rs
```diff
@@ -324,6 +324,18 @@ impl BranchNode {
                 Some(Child::AddressWithHash(_, hash)) => Some((i, hash)),
             })
     }
+
+    /// Returns (index, address) for each child that has a hash set.
+    pub fn children_addresses(&self) -> impl Iterator<Item = (usize, &LinearAddress)> + Clone {
```
rkuris, line 334: These are almost exactly the same, so let's create a helper that returns `(i, Child::AddressWithHash)`

storage/Cargo.toml
```diff
@@ -46,6 +48,8 @@ test-case = "3.3.1"
 criterion = { version = "0.6.0", features = ["async_tokio", "html_reports"] }
 pprof = { version = "0.15.0", features = ["flamegraph"] }
 tempfile = "3.12.0"
+tokio = { version = "1.36.0", features = ["rt", "sync", "macros", "rt-multi-thread"] }
```
rkuris: remove

storage/src/range_set.rs
qusuyan: Mainly because some functions in ranget set are not used by the checker, but I kept them in case we want to put it into a separate general purpose crate.

storage/src/nodestore.rs
```diff
@@ -1721,3 +1786,94 @@ mod tests {
...
+    use nonzero_ext::nonzero;
+    use std::num::NonZeroUsize;
+    use tempfile::NamedTempFile;
+
+    const NODE_CACHE_SIZE: NonZeroUsize = nonzero!(1000usize);
+    const FREE_LIST_CACHE_SIZE: NonZeroUsize = nonzero!(1000usize);
+
+    // Helper function to wrap the node in a StoredArea and write it to the given offset. Returns the area size on success.
+    #[allow(clippy::cast_possible_truncation)]
+    fn write_new_node(file_backed: &FileBacked, node: &Node, offset: u64) -> u64 {
```
rkuris: restructure to reuse code

storage/src/nodestore.rs
```diff
@@ -1721,3 +1786,94 @@ mod tests {
...
+        let node_length = stored_area_bytes.len() as u64;
+        let area_index = area_size_to_index(node_length).unwrap();
+        let area_size = AREA_SIZES[area_index as usize];
+        stored_area_bytes[0] = area_index;
+        file_backed.write(offset, &stored_area_bytes).unwrap();
+        area_size
+    }
+
+    fn write_header(file_backed: &FileBacked, root_addr: LinearAddress, size: u64) {
+        let mut header = NodeStoreHeader::new();
```
rkuris: use `flush_header`

storage/src/nodestore.rs
```diff
@@ -1721,3 +1786,94 @@ mod tests {
...
+        header.size = size;
+        header.root_address = Some(root_addr);
+        let header_bytes = bytemuck::bytes_of(&header);
+        file_backed.write(0, header_bytes).unwrap();
+    }
+
+    #[test]
+    fn test_checker_traverse_correct_trie() {
+        let tf = NamedTempFile::new().unwrap();
```
rkuris: Verify that you're not the only one who needs to do this

storage/src/nodestore.rs
```diff
@@ -1721,3 +1786,94 @@ mod tests {
...
+    fn write_header(file_backed: &FileBacked, root_addr: LinearAddress, size: u64) {
+        let mut header = NodeStoreHeader::new();
+        header.size = size;
+        header.root_address = Some(root_addr);
+        let header_bytes = bytemuck::bytes_of(&header);
+        file_backed.write(0, header_bytes).unwrap();
+    }
+
+    #[test]
+    fn test_checker_traverse_correct_trie() {
```
rkuris: Add comment: use primitive calls here to do a low level check, and reference where the higher level checks will live (or a TODO)

storage/src/nodestore.rs
```diff
@@ -1721,3 +1786,94 @@ mod tests {
...
+        let file_backed = FileBacked::new(
+            tf.path().to_path_buf(),
+            NODE_CACHE_SIZE,
+            FREE_LIST_CACHE_SIZE,
+            true,
+            CacheReadStrategy::All,
+        )
+        .unwrap();
+
+        // set up a basic trie
```
rkuris: needs more comments

storage/src/nodestore.rs
```diff
@@ -1721,3 +1786,94 @@ mod tests {
...
+            value: None,
+            children: root_children,
+        }));
+        let root_addr = LinearAddress::new(offset).unwrap();
+        let root_area = write_new_node(&file_backed, &root, offset);
+        offset += root_area;
+
+        write_header(&file_backed, root_addr, offset);
+
+        // test that the we traversed the entire trie
```
rkuris: comment: add verify that all of the space is accounted for

storage/src/range_set.rs
```diff
@@ -0,0 +1,671 @@
...
+use btree_slab::BTreeMap;
+use btree_slab::generic::map::BTreeExt;
+use btree_slab::generic::node::Address;
+use std::cmp::Ordering;
+use std::ops::Range;
+
+use crate::nodestore::STORAGE_AREA_START;
+use crate::{CheckerError, LinearAddress};
+
+pub struct RangeSet<T> {
```
rkuris: implement debug

rkuris (changes requested):

storage/src/range_set.rs
```diff
@@ -0,0 +1,657 @@
...
+
+use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::ops::Range;
+
+use crate::nodestore::NodeStoreHeader;
+use crate::{CheckerError, LinearAddress};
+
+#[derive(Debug)]
+pub struct RangeSet<T>(BTreeMap<T, T>); // end --> start so that we can look up ranges by its start
```
aaronbuchwald:
```suggestion
pub struct RangeSet<T>(BTreeMap<T, T>); // end --> start so that we can look up ranges by its end
```

qusuyan: Updated the comment with more context - we look up any given ranges by its start to find intersections with what's already in the `RangeSet`

storage/src/checker.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Go through the filebacked storage and check for any inconsistencies. It proceeds in the following steps:
+    /// 1. Check the header
+    /// 2. traverse the trie and check the nodes
+    /// 3. check the free list
+    /// 4. check missed areas - what are the spaces between trie nodes and free lists we have traversed?
+    /// # Errors
+    /// Returns a [`CheckerError`] if the database is inconsistent.
+    // TODO: report all errors, not just the first one
+    // TODO: add merkle hash checks as well
+    pub fn check(&self) -> Result<(), CheckerError> {
```
demosdemon, line 22: High level: this only returns a single error and only the first error we find. Ideally, we should be returning every error we can discover and only abort the entire scan when we find an error we cannot skip past. And even then, we should include all the errors we discovered up until that point. If there is something we can fix, then we will want to do that in bulk instead of (1) running fsck, (2) getting and error, (3) fixing it, and (1) running fsck again until (4) we get an error we can't fix or it's clean.

This could potentially be done using a callback instead of collecting everything into a vec. Have the caller of `check` pass in something that implements a trait that receives each error we discover as we discover it.

qusuyan: We are aware of this issue but at the same time some errors may be detrimental and it might not make sense to continue running the checker. 

My plan for this PR (and the next one) is to get a simplistic checker that catches only the first error it encounters so that we know all the errors that can happen. In future PRs we can decide which errors can be fixed and which ones are detrimental.

demosdemon (approved):

PR #941, opened 2025-06-06
feat: Improve error handling and add sync iterator

rkuris:
While debugging, I couldn't figure out where the error was coming from, so I revamped the error handling to include the filename and line number everywhere an IO error could happen.

Many places that previously used std::io::Error now use FileIoError to include the additional context information.

Also, this change adds support for synchronous iterators for both nodes and key/values, which is needed for synchronous range and change proofs.

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -2,7 +2,6 @@
 // See the file LICENSE.md for licensing terms.
 
 use std::collections::{HashMap, VecDeque};
-use std::io::Error;
 use std::num::NonZero;
 use std::path::PathBuf;
 use std::sync::Arc;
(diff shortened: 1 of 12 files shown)


firewood/src/manager.rs
```diff
@@ -221,7 +225,7 @@ impl RevisionManager {
 
         if trace_enabled() {
             let _merkle = Merkle::from(committed);
-            trace!("{}", _merkle.dump()?);
+            trace!("{}", _merkle.dump().expect("failed to dump merkle"));
```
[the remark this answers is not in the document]
rkuris: This is in trace code, so I'm going to leave it. Seems reasonable to panic if you're tracing and can't write the trace output here.

firewood/src/stream.rs
```diff
@@ -90,23 +92,19 @@ impl<'a, T: TrieReader> MerkleNodeStream<'a, T> {
...
 
-    fn poll_next(
-        mut self: std::pin::Pin<&mut Self>,
-        _cx: &mut std::task::Context<'_>,
-    ) -> Poll<Option<Self::Item>> {
-        // destructuring is necessary here because we need mutable access to `state`
-        // at the same time as immutable access to `merkle`.
+    /// Internal function that handles the core iteration logic shared between Iterator and Stream implementations.
+    /// Returns None when iteration is complete, or Some(Result) with either a node or an error.
+    fn next_internal(&mut self) -> Option<Result<(Key, SharedNode), FileIoError>> {
```
alarso16, line 98: This is a weird return type, why would you do this instead of `Result<Option<>,Error>`

alarso16: I now get that since it's for the iterator, it's to signal that the iterator ended, but I still think it reads a little strange. Is this a common Rust idiom?

alarso16 (approved):
fwict looks good

PR #945, opened 2025-06-09
feat: Return database creation errors through FFI

alarso16:
This returns errors from database creation across the FFI layer rather than panicking in `expect` clauses.

I also wanted to test the case that prompted this (metric port errors), but I couldn't cause it to fail in CI by opening a port with `net.Listen`. If there's a way to consume the port, that would be really helpful


ffi/memory.go
```diff
@@ -146,6 +146,20 @@ func bytesFromValue(v *C.struct_Value) ([]byte, error) {
...
 
+func databaseFromResult(result *C.struct_DatabaseCreationResult) (*C.DatabaseHandle, error) {
+	if result == nil {
+		return nil, errNilBuffer
+	}
+
+	if result.error_str != nil {
+		errStr := C.GoString((*C.char)(unsafe.Pointer(result.error_str)))
+		C.fwd_free_database_result(result)
+		runtime.KeepAlive(result)
```
alarso16, line 157: We could also generalize the `valueFactory` to pin anything, but seemed unnecessary since it's only used once

ffi/src/lib.rs
```diff
@@ -730,6 +730,60 @@ pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
...
+#[derive(Debug)]
+#[repr(C)]
+pub struct DatabaseCreationResult {
+    pub db: *const DatabaseHandle<'static>,
+    pub error_str: *const u8,
+}
+
+impl From<Result<*const DatabaseHandle<'static>, String>> for DatabaseCreationResult {
+    fn from(result: Result<*const DatabaseHandle<'static>, String>) -> Self {
+        match result {
```
alarso16, line 742: This doesn't feel super rusty, but it is super readable

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +16,44 @@ use chrono::{DateTime, Utc};
...
+                    Response::builder()
+                        .status(StatusCode::OK)
+                        .header("Content-Type", "text/plain")
+                        .body(Body::from(recorder.stats()))
+                        .expect("failed to build response")
+                } else {
+                    Response::builder()
+                        .status(StatusCode::METHOD_NOT_ALLOWED)
+                        .body(Body::from("Method not allowed"))
+                        .expect("failed to build response")
```
alarso16: I'm not sure if there's a way to return this string. Any ideas?

ffi/firewood.go
```diff
@@ -103,11 +103,16 @@ func New(filePath string, conf *Config) (*Database, error) {
...
-		db = C.fwd_create_db(args)
+		dbResult = C.fwd_create_db(args)
 	} else {
-		db = C.fwd_open_db(args)
+		dbResult = C.fwd_open_db(args)
+	}
+
+	db, err := databaseFromResult(&dbResult)
+	if err != nil {
+		return nil, err
```
aaronbuchwald, line 115: ❤️

ffi/firewood_test.go
```diff
@@ -144,6 +144,15 @@ func newDatabase(dbFile string) (*Database, func() error, error) {
 	return f, f.Close, nil
 }
 
+func TestOpenNonexistentDatabase(t *testing.T) {
+	r := require.New(t)
+	cfg := DefaultConfig()
+	cfg.Create = false
+	db, err := New(filepath.Join(t.TempDir(), "test.db"), cfg)
+	r.Error(err)
```
aaronbuchwald: Can we use `ErrorIs` or `ErrorContains` here? We should prefer making it as specific as possible

ffi/memory.go
```diff
@@ -146,6 +146,20 @@ func bytesFromValue(v *C.struct_Value) ([]byte, error) {
 	return nil, errBadValue
 }
 
+func databaseFromResult(result *C.struct_DatabaseCreationResult) (*C.DatabaseHandle, error) {
+	if result == nil {
+		return nil, errNilBuffer
```
aaronbuchwald: Could we use a specific error here since this is not a nil buffer?

ffi/memory.go
```diff
@@ -146,6 +146,20 @@ func bytesFromValue(v *C.struct_Value) ([]byte, error) {
...
+	}
+
+	if result.error_str != nil {
+		errStr := C.GoString((*C.char)(unsafe.Pointer(result.error_str)))
+		C.fwd_free_database_result(result)
+		runtime.KeepAlive(result)
+		return nil, fmt.Errorf("firewood error: %s", errStr)
```
aaronbuchwald, line 158: not necessary for this PR - would it make sense to use an array of error codes that can be translated to golang error values here rather than passing the string?

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +16,44 @@ use chrono::{DateTime, Utc};
...
-                    .header("Content-Type", "text/plain")
-                    .body(Body::from(recorder.stats()))
-                    .expect("failed to build response")
-            } else {
-                Response::builder()
-                    .status(StatusCode::METHOD_NOT_ALLOWED)
-                    .body(Body::from("Method not allowed"))
-                    .expect("failed to build response")
-            }
+static INIT_RESULT: OnceLock<Result<(), String>> = OnceLock::new();
```
aaronbuchwald: Do we want this Once block here?

Is this intended to prevent our own code from accidentally setting up the metrics twice for a single firewood db or so that if we intentionally create multiple all of the metrics are ingested in one place?

If the latter, I think we may prefer to separate the metrics into two separate instances?

alarso16: @rkuris I just want to check and understand the propose of the Once - we want multiple Firewood instances (separate database files) to use the same port for metrics?

ffi/src/metrics_setup.rs
aaronbuchwald: This is the callback fn used by the server to handle each request as opposed to part of the constructor, so this function's return type is not `Result<(), String>`

The function parameter doesn't take an error in its return type, so I think `.expect(...)` may be required here. Looking at the docs for `Server::new(...)`, it uses unwrap in its own example, so this seems to be expected usage.

ffi/memory.go
alarso16: Maybe? There's of course a lot of errors that COULD be sent back, so I'm not sure that's intuitive. It would make error handling a lot faster, since we wouldn't have to free the string. I'm not sure it's worth the maintenance effort.

ffi/firewood_test.go
alarso16: I used `ErrorContains` since all of the FFI errors are passed as strings. It's still mildly vague because I think this error could be generally unreliable:

`firewood error: File IO error: failed to fill whole buffer at offset 0 of file '/var/folders/42/7kdskh7x4pg9yly8kr3k67240000gp/T/TestOpenNonexistentDatabase3266225433/001/test.db' header read`

ffi/memory.go
alarso16: Renamed

ffi/src/metrics_setup.rs
alarso16: @rkuris implemented this initially - I only changed the semantics. I agree though - I don't think two firewood databases should necessarily use the same metrics port

alarso16: I just don't know a strong way to enforce that each database has only one metric port

ffi/memory.go
rkuris: This is harder than you think. Most of the rust errors contain context information, so for example Error::HashNotFound contains a provided HashKey. Other errors, such as InvalidRange contain start and end keys. Passing all these back in non-string format would be quite tricky.

If you wanted to implement this, you'd need all the api::Error types to have a corresponding C type. I doubt that is worth it.

ffi/src/lib.rs
rkuris: Looks fine to me! The only alternative is `.map` and `.map_err` so I think this match is cleaner.

ffi/src/lib.rs
```diff
@@ -730,6 +730,60 @@ pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
...
+/// # Safety
+///
+/// This function is unsafe because it dereferences raw pointers.
+/// The caller must ensure that `result` is a valid pointer.
+///
+/// # Panics
+///
+/// This function panics if `result` is `null`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn fwd_free_database_result(result: *const DatabaseCreationResult) {
```
rkuris: nit: naming might be better as`fwd_free_database_error_result` and document that it's only needed if there was an error (but doesn't hurt to call it anyway)

ffi/src/metrics_setup.rs
rkuris: You can't actually. They all share the same metrics. If you open a second database and specify a metrics port on both, it should be an error.

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +16,44 @@ use chrono::{DateTime, Utc};
...
+            metrics::set_global_recorder(recorder.clone())
+                .map_err(|e| format!("failed to set recorder: {e}"))?;
+
+            Server::new(move |request| {
+                if request.method() == "GET" {
+                    Response::builder()
+                        .status(StatusCode::OK)
+                        .header("Content-Type", "text/plain")
+                        .body(Body::from(recorder.stats()))
+                        .expect("failed to build response")
```
rkuris: these will kill the metrics reporter thread, so perhaps we should be a little more graceful here.

ffi/src/metrics_setup.rs
rkuris: You can only initialize metrics once, it is a global thing. The [docs for set_global_reporter](https://docs.rs/metrics/0.24.2/metrics/fn.set_global_recorder.html) say you can only do it once.

If you have two databases open, the metrics will mix.

The way around this is to add something during the prometheus setup to indicate the instance and plunk this into the DB object, then include that each time a metric is emitted. This doesn't exist yet, so if we need multiple database support sometime in the future, we'd have to add that. Probably reasonable to make a ticket for that.

ffi/src/metrics_setup.rs
rkuris: Which string? As stated, this is happening in a different thread, which continues running even after `setup_metrics` has returned.

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +16,44 @@ use chrono::{DateTime, Utc};
...
+                        .body(Body::from("Method not allowed"))
+                        .expect("failed to build response")
+                }
+            })
+            .bind((Ipv4Addr::LOCALHOST, metrics_port))
+            .bind((Ipv6Addr::LOCALHOST, metrics_port))
+            .with_global_timeout(Duration::from_secs(60 * 60))
+            .with_max_concurrent_connections(2)
+            .spawn()
+            .map_err(|e| format!("failed to spawn server: {e}"))?;
```
rkuris: You shouldn't need this, and the caller should be converting the error to a string

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +16,44 @@ use chrono::{DateTime, Utc};
...
+            Ok(())
         })
-        .bind((Ipv4Addr::LOCALHOST, metrics_port))
-        .bind((Ipv6Addr::LOCALHOST, metrics_port))
-        .with_global_timeout(Duration::from_secs(60 * 60))
-        .with_max_concurrent_connections(2)
-        .spawn()
-        .expect("failed to spawn server");
-    });
+        .clone()
```
rkuris: Why was this needed?

ffi/src/metrics_setup.rs
alarso16: What is being "more graceful"? Is there anything else that can be done besides killing the thread?

rkuris: log and ignore the error

ffi/src/metrics_setup.rs
alarso16: `get_or_init` returns a reference

ffi/src/metrics_setup.rs
alarso16: Opened in #955

ffi/src/metrics_setup.rs
alarso16: Where do I log? If I ignore the error, there's still no Response to return in the closure, so I'll still need an `expect` to create the error-filled response

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +17,44 @@ use chrono::{DateTime, Utc};
...
+        registry: Registry::atomic(),
+    };
+    let recorder = TextRecorder {
+        inner: Arc::new(inner),
+    };
+
+    // Any caller should ensure that the metrics system is not already initialized
+    // by explicitly calling with metrics_port == 0.
+    // If multiple databases are running, they should use the same port.
+    if let Err(e) = metrics::set_global_recorder(recorder.clone()) {
```
alarso16: Unfortunately, this cannot really be well tested because of reserved ports in different OSs

ffi/src/metrics_setup.rs
rkuris: Looked a little deeper, this is okay because, according to the docs, body will only fail if it can't convert the request into a valid response due to a broken header. Since we're in control of the headers ourselves, body should always work, so the expect calls are never going happen in our use case.

ffi/src/metrics_setup.rs
rkuris: You can use a port that you lack permission to bind to, or one that is already listening, like the ssh port 22. This has two problems in that it is both a protected port as well as probably being in use.

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +17,44 @@ use chrono::{DateTime, Utc};
...
+        if metrics_port == 0 {
```
rkuris: This must be done *before* calling `set_global_recorder`. Otherwise, if you open two databases, the first one doesn't want metrics and the second one does, it won't work correctly.

rkuris (changes requested):

ffi/src/metrics_setup.rs
```diff
@@ -16,39 +17,44 @@ use chrono::{DateTime, Utc};
...
+    let inner: TextRecorderInner = TextRecorderInner {
+        registry: Registry::atomic(),
+    };
+    let recorder = TextRecorder {
+        inner: Arc::new(inner),
+    };
+
+    // Any caller should ensure that the metrics system is not already initialized
+    // by explicitly calling with metrics_port == 0.
+    // If multiple databases are running, they should use the same port.
```
aaronbuchwald: The expectations of this function seem a bit unclear to me with this change and perhaps should be documented above the function itself as well?

If this is called twice, it's the caller's responsibility to guarantee that the first has a non-zero port and the second specifies port 0?

alarso16: Yes. The other option is to store the previous port in a global lock and check whether it matches to avoid error.

ffi/src/metrics_setup.rs
alarso16: Actually, the caller defines port 0 as "no metrics". We shouldn't have to handle this case technically

alarso16: I still return, but it shouldn't be necessary

ffi/src/metrics_setup.rs
alarso16: Surprisingly, a different error occurs if you run multiple tests at the same time. The metrics binding behavior is illogical to me, and should likely be further investigated

alarso16: I could create a strange almost flaky test that just checks for any error, either that the recorder is already initialized (which happens sometimes for some reason unknown to me) or the permission denied error. It should always pass, but I don't like that the error changes based on the situation

ffi/src/lib.rs
```diff
@@ -730,6 +730,61 @@ pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
...
+/// # Safety
+///
+/// This function is unsafe because it dereferences raw pointers.
+/// The caller must ensure that `result` is a valid pointer.
+///
+/// # Panics
+///
+/// This function panics if `result` is `null`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn fwd_free_database_error_result(result: *const DatabaseCreationResult) {
```
demosdemon: This is undefined behavior. We are mutating what is owned by a const pointer. This should be a mut pointer instead. Or even plainly `&mut DatabaseCreationResult`.

```suggestion
pub unsafe extern "C" fn fwd_free_database_error_result(result: &mut DatabaseCreationResult) {
```

ffi/src/lib.rs
```diff
@@ -730,6 +730,61 @@ pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
     }
 }
 
+/// Struct returned by `fwd_create_db` and `fwd_open_db`
+#[derive(Debug)]
+#[repr(C)]
+pub struct DatabaseCreationResult {
+    pub db: *const DatabaseHandle<'static>,
```
demosdemon, line 736: This doesn't need to be behind a pointer. cbindgen will do the right thing of you do `Option<Box<DatabaseHandle<'static>>>` instead. This applies to all structs fully known to cbindgen and also works on functions. So, instead of `ptr: *mut DatabaseHandle<'static>`, we can do `mut hdl: Box<DatabaseHandle<'static>>` to reduce some unsafe code. It will also handle `&DatabaseHandle<'static>` and `&mut DatabaseHandle<'static>` correctly.

```suggestion
    pub db: Option<Box<DatabaseHandle<'static>>>,
```

ffi/src/lib.rs
```diff
@@ -730,6 +730,61 @@ pub unsafe extern "C" fn fwd_free_value(value: *const Value) {
...
+    pub error_str: *const u8,
```
demosdemon: this should be `*mut c_char` or even better `Option<std::ptr::NonNull<c_char>>`

```sugggestion
    pub error_str: Option<std::ptr::NonNull<c_char>>,
```

While cbindgen won't automatically handle `CString`, it will handle that which allows you to `result.error_str.take()` to clear the option and get a known non-null pointer.

demosdemon (changes requested):

ffi/src/lib.rs
alarso16: Good catch! This occurred somewhere too, so I fixed the UB there as well.

ffi/src/lib.rs
alarso16: We're avoiding `c_char` due to issues compiling across different architectures

alarso16: For the more Rust-y type, see #967

ffi/src/lib.rs
alarso16: This will be address with #967

ffi/src/lib.rs
alarso16: For using the reference, I will address this in #967

ffi/src/metrics_setup.rs
rkuris: nit: It would be good to doc this in the code comments

rkuris (approved):
LGTM after the const fixes @demosdemon called out

ffi/src/metrics_setup.rs
demosdemon: Anything that calls `set_global_recorder` must be tested in an isolated process as that is a call-only-once method. Tests in rust are by default all run in the same process in parallel. There's some tricks like `#[serial]` to add mutexes around certain tests, but that won't help with something that mutates global process state like `set_global_recorder`. 

There's nothing built into stdlib to handle forking tests like that. I avoid this problem however much possible so I'm not familiar with existing third-party tools that can help. For unit tests, it's better to somehow architect in a way to use [`with_local_recorder`](https://docs.rs/metrics/latest/metrics/fn.with_local_recorder.html) instead of relying on the global recorder.

ffi/src/lib.rs
```diff
@@ -793,58 +840,58 @@ pub unsafe extern "C" fn fwd_create_db(args: CreateOrOpenArgs) -> *const Databas
...
+    let path = unsafe { CStr::from_ptr(args.path) };
     #[cfg(unix)]
     let path: &Path = OsStr::from_bytes(path.to_bytes()).as_ref();
     #[cfg(windows)]
     let path: &Path = OsStr::new(path.to_str().expect("path should be valid UTF-8")).as_ref();
-    if metrics_port > 0 {
-        metrics_setup::setup_metrics(metrics_port);
+    if args.metrics_port > 0 {
+        metrics_setup::setup_metrics(args.metrics_port).map_err(|e| e.to_string())?;
     }
```
demosdemon, line 869: On that note, I think we should probably add a follow up to move configuring metrics outside of setting up the database. With this current setup, we can only handle one db per process. If we want to have a unique metrics instance per db, we would need to not use something that relies on global process state. Relying on the user providing `0` as the port for this isn't the best api contract. I would prefer to let the user provide `0` to let the os auto-assign a port and then somehow communicate that port back to the user.

alarso16: Good call out - this does already have an issue #955, since this is a rather large change.

ffi/src/metrics_setup.rs
aaronbuchwald: Why is it not possible to create a second, separate database?

aaronbuchwald (approved):

PR #956, opened 2025-06-11
test: add ethhash fuzz test

aaronbuchwald:


aaronbuchwald:
Note: this PR changes the name of `ethhash-compatibility-go`, which is currently a required job. Will need to override the protection and update the name of the required job after merging.

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -20,225 +20,324 @@ import (
...
+		createAccount: "createAccount",
+		updateAccount: "updateAccount",
+		deleteAccount: "deleteAccount",
+		addStorage:    "addStorage",
+		updateStorage: "updateStorage",
+		deleteStorage: "deleteStorage",
+	}
+)
+
+type tree struct {
```
aaronbuchwald: consider renaming since we are not using a tree of proposals here and simply fuzzing a single view (committed database itself)

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -20,225 +20,324 @@ import (
...
+		tr.pendingFwdKeys = append(tr.pendingFwdKeys, accHash[:])
+		updatedAccountRLP, err := rlp.EncodeToBytes(acc)
+		tr.require.NoError(err)
+		tr.pendingFwdVals = append(tr.pendingFwdVals, updatedAccountRLP)
+	}
+
+	updatedRoot, set, err := tr.accountTrie.Commit(true)
+	tr.require.NoError(err)
+
+	// A no-op change returns a nil set, which will cause merge to panic.
```
aaronbuchwald: Same comment in both places, but this seems reasonable to me as opposed to trying to factor out

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -20,225 +20,324 @@ import (
...
+	tr.pendingFwdVals = append(tr.pendingFwdVals, accountRLP)
+}
+
+func (tr *tree) selectAccount(addrIndex int) (common.Address, common.Hash) {
+	addr := tr.currentAddrs[addrIndex]
+	return addr, crypto.Keccak256Hash(addr[:])
+}
 
-	rand := rand.New(rand.NewSource(0))
+func (tr *tree) updateAccount(addrIndex int) {
```
aaronbuchwald: Could use another pass over the comments

ffi/tests/eth/eth_compatibility_test.go
aaronbuchwald: Renaming to `merkleTriePair`

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -20,225 +20,351 @@ import (
...
 
-		// update firewood db
-		got, err := db.Update(fwKeys, fwVals)
-		require.NoError(t, err)
-		require.Equal(t, next[:], got)
+// deleteStorage selects an account and deletes an existing storage key-value pair
+// note: this may "delete" a key-value pair that doesn't exist if it was previously deleted.
+func (tr *merkleTriePair) deleteStorage(accountIndex int, storageIndexInput uint64) {
+	addr, accHash := tr.selectAccount(accountIndex)
+	storageMap, ok := tr.currentStorage[addr]
```
alarso16: `storageMap` isn't actually used here or in `updateStorage`. It should either not be instantiated or (more likely) updated to reflect the real state

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -20,225 +20,351 @@ import (
...
+	return addr, crypto.Keccak256Hash(addr[:])
+}
+
+// updateAccount selects a random account, increments its nonce, and adds the update
+// to the pending changes for both tries.
+func (tr *merkleTriePair) updateAccount(addrIndex int) {
+	addr, accHash := tr.selectAccount(addrIndex)
+	acc, err := tr.accountTrie.GetAccount(addr)
+	tr.require.NoError(err)
+	acc.Nonce++
```
alarso16, line 181: It might be "fun" to update other things in the account to ensure it doesn't change hashing, since there's some weird bits that are ignored (like the storage root is completely ignored in Firewood, but what if the parsing of the account is wrong?)

ffi/tests/eth/eth_compatibility_test.go
alarso16: Actually, I don't know what the point of the `storageMap` is at all

ffi/tests/eth/eth_compatibility_test.go
aaronbuchwald: Good point, this is just borrowing the diff from before. I'll update this to apply a small diff to all possible fields.

ffi/tests/eth/eth_compatibility_test.go
aaronbuchwald: Nice catch!

aaronbuchwald: Added this thinking I'd want to track storage separate from the two tries, but this is unnecessary.

PR #957, opened 2025-06-11
build: move lints to the workspace

### LINKED ISSUE #951: storage doesn't enforce lint rules
It used to be storage actually had a more aggressive lint ruleset, but it's missing now for some reason, and there are now 54 violations, more if you include the warning we want in #938. 

We should move the lint rules to the workspace level and just add "lints.workspace = true" to each of the members. I don't know if there is a way to check that it is in each member, but that would be a nice addition.

demosdemon:
Previously, the lints were defined in each crate's `Cargo.toml`. This commit moves the lints to the workspace's `Cargo.toml` file. This allows for a more consistent linting configuration across all crates in the workspace and simplifies the management of lint rules. This change also ensures that all crates in the workspace adhere to the same linting standards, improving code quality and maintainability.

New crates added to the workspace via `cargo new` will automatically inherit the linting rules defined in the workspace's `Cargo.toml`. However, manually crafted crates will need to ensure the stanza

```toml
[lints]
workspace = true
```

is included in `Cargo.toml`to benefit from the workspace-wide linting rules.

Caveats:
  - This change adds an excessive amount of warnings because `clippy::pedantic` is very easy to violate and is now enabled workspace wide. This is also why I included pedantic lints with a lower priority, so that any individual lint within pedantic can be separately disabled with an `allow`.
  - Packages cannot override the workspace lints within their `Cargo.toml`. This can be worked around by adding the lint rule change via a crate attribute in the crate's `lib.rs` or `main.rs`. See the `ffi` crate for an example of this workaround.

Resolves #951

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,7 +44,6 @@ jobs:
       - name: Build
         run: cargo build ${{ matrix.profile-args }} --workspace --all-targets
 
-
   no-rust-cache-lint:
     runs-on: ubuntu-latest
     steps:
@@ -58,6 +57,35 @@ jobs:
           strict: true
       - name: Format
         run: cargo fmt -- --check
+
+  rust-lint-pr-comments:
+    # caveat emptor: actions-rs/clippy-check can only write comments on pull requests
+    # not originating from a fork.
+    if: github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - profile-key: debug-no-features
+            profile-args: ""
+          - profile-key: debug-ethhash-logger
+            profile-args: "--features ethhash,logger"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: "false"
+          shared-key: ${{ matrix.profile-key }}
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          name: ${{ matrix.profile-key }}
+          args: ${{ matrix.profile-args }} --workspace --all-targets
 
   rust-lint:
     needs: build
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,9 +40,9 @@ jobs:
         with:
           shared-key: ${{ matrix.profile-key }}
       - name: Check
-        run: cargo check ${{ matrix.profile-args }} --workspace --tests --examples --benches
+        run: cargo check ${{ matrix.profile-args }} --workspace --all-targets
       - name: Build
-        run: cargo build ${{ matrix.profile-args }} --workspace --tests --examples --benches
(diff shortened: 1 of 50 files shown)


Cargo.toml
```diff
@@ -18,3 +18,16 @@ codegen-units = 1
...
+# lower the priority of rust_2024_compatibility to allow overriding the lints it includes
+rust_2024_compatibility = { level = "warn", priority = -1 }
+
+[workspace.lints.clippy]
+unwrap_used = "warn"
+indexing_slicing = "warn"
+explicit_deref_methods = "warn"
+missing_const_for_fn = "warn"
+# lower the priority of pedantic to allow overriding the lints it includes
+pedantic = { level = "warn", priority = -1 }
```
rkuris, line 32: 🆒 🍪

Cargo.toml
```diff
@@ -18,3 +18,16 @@ codegen-units = 1
 lto = "fat"
 debug = false
 inherits = "release"
+
+[workspace.lints.rust]
+unsafe_code = "deny"
+# lower the priority of rust_2024_compatibility to allow overriding the lints it includes
+rust_2024_compatibility = { level = "warn", priority = -1 }
```
rkuris: I think we are already on rust 2024 so this warning can be a warning and cause the builds to fail

.github/workflows/ci.yaml
```diff
@@ -27,8 +27,12 @@ jobs:
             profile-args: ""
           - profile-key: debug-ethhash-logger
             profile-args: "--features ethhash,logger"
+          - profile-key: debug-branch_factor_256-logger
+            profile-args: "--features branch_factor_256,logger"
           - profile-key: maxperf-ethhash-logger
             profile-args: "--profile maxperf --features ethhash,logger"
+          - profile-key: maxperf-branch_factor_256-logger
+            profile-args: "--profile maxperf --features branch_factor_256,logger"
```
rkuris: branch_factor_256 is likely going away, so I would prefer to test without it.

xtask/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "xtask"
+version = "0.1.0"
+edition = "2024"
```
rkuris: This needs all the metadata for publishing, and let's call it firewood-xtask

xtask/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
```
rkuris: xtask needs a README.md that explains what it does and why we need it

rkuris (changes requested):
I didn't walk through the xtask code, but if we intend to keep it, it's going to need more documentation, probably at least a README and a LICENSE file. Your check-in comment about xtask looks like a good start.

I'm not strongly opposed to adding another crate, unless we can find a crate that already does what we need, like maybe `cargo-xtask`.

rkuris:
It also might be a good idea to make xtask changes a separate PR.

demosdemon:
> separate PR

yeah, can do. mainly included as a "show your work" kind of deal for discovering and injecting the bulk suppressions but I'm also a fan of the [xtask](https://github.com/matklad/cargo-xtask) pattern.

firewood/src/merkle.rs
```diff
@@ -369,18 +410,18 @@ impl<T: HashedNodeReader> Merkle<T> {
                     };
 
                     let inserted = seen.insert(child_addr);
-                    if !inserted {
+                    if inserted {
```
rkuris, line 393: nit: not sure I agree with this change, as I'd prefer to handle errors closer to where the problem occurred, but I'm not completely against it

PR #962, opened 2025-06-13
chore: Set up for publishing to crates.io

rkuris:
It's been a while, but I'd like to get crates.io published. These are all the changes required to do that.

Also, the authors have been updated to correctly reflect the authors for the latest release. We'll still need to publish a new revision to get this on crates.io.


rkuris:
> signing binaries (e.g., the static ffi assets)
Created #963 to address this

ffi/Cargo.toml
```diff
@@ -3,13 +3,24 @@ name = "firewood-ffi"
 version = "0.0.5"
 edition = "2024"
 rust-version = "1.85.0"
+authors = [
+     "Aaron Buchwald <aaron.buchwald56@gmail.com>",
+     "Arran Schlosberg <519948+ARR4N@users.noreply.github.com>",
+     "Austin Larson <78000745+alarso16@users.noreply.github.com>",
+     "Darioush Jalali <darioush.jalali@avalabs.org>",
```
alarso16, line 10: nit: I'm not sure what the point of this is, but obv Darioush's email is not valid

rkuris: True, but I don't want to leave him off, and probably don't have permission to add his personal email.

Note that this came from the commit history, so I'm not exposing anything more than what was already there.

rkuris: @aaronbuchwald can you stamp this also please?

demosdemon: fwiw, authors is [a deprecated field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field). There might be better ways to ensure proper recognition of contributions. 

- [Motivation](https://rust-lang.github.io/rfcs/3052-optional-authors-field.html#motivation)
- [History](https://github.com/rust-lang/cargo/issues/15027)

Issue #958, opened 2025-06-12
bug: Incompatible on every version update


alarso16:
Every time a new version is released, there is a [check](https://github.com/ava-labs/firewood/blob/b2141e835ba6c8b6ff3ea3863a583f9b5f137f99/storage/src/nodestore.rs#L308) in the file header comparing versions. If the versions aren't compatible, it panics.

This points out two issues:
- This sort of error on startup shouldn't panic. An error should be returned to the user.
- Not every version is incompatible. We could design a versioning scheme where major versions describe breaking changes, and make an exception for 0->1.

PR #964, opened 2025-06-14
fix(storage)!: Restrict database version identifier

demosdemon:
BREAKING CHANGE: This will make the previous database unreadable; but, this would also happen every time we bumped the patch version.

Problem: Currently, the database version includes both the minor and patch numbers as well as any other potential values in `CARGO_PKG_VERSION`. Therefore, whenever the patch version is bumped, which is expected to happen fairly often, the database becomes incompatible.

Solution: We're truncating the version included in the db to just the major version. Therefore, when we eventually release 1.0.0, then will be a database incompatible upgrade.

Resolves: #958


storage/build.rs
```diff
@@ -0,0 +1,18 @@
...
+// See the file LICENSE.md for licensing terms.
+
+use semver::Version;
+
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+fn main() {
+    let crate_version = Version::parse(VERSION).expect("failed to parse CARGO_PKG_VERSION");
+
+    // Set FIREWOOD_DB_VERSION to the major version of the crate. E.g., if the version
```
rkuris: :trophy: on approach

Let's move this comment to the top of the function or as a doc comment for the whole file so I don't have to read the function to figure out what it does.
```suggestion
    // Put FIREWOOD_DB_VERSION into the build environment
    // this is the major version of semver release
```

storage/build.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use semver::Version;
+
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+fn main() {
+    let crate_version = Version::parse(VERSION).expect("failed to parse CARGO_PKG_VERSION");
```
rkuris: We need to say what the bad string was, where it came from and why it's a problem.

```suggestion
    let crate_version = Version::parse(VERSION).expect(format!("The version '{}' in Cargo.toml does not match the semver spec", VERSION));
```

storage/src/nodestore.rs
```diff
@@ -694,14 +694,34 @@ impl Version {
...
+
+        // compile-time assertion that the version string is not too long.
+        const {
+            assert!(
+                VERSION_STR.len() <= Self::SIZE as usize,
+                "Version string too long"
+            );
+        }
+
+        // statically extend the version string to the full size of the Version struct
```
rkuris: format strings are magic, and can be fixed size with padding. Suggest using format! to get this string padded with the bytes you want.

rkuris (commented):
I would prefer to keep compatibility at this point, so maybe we can add a special case for the current version, which can be removed later.

storage/src/nodestore.rs
demosdemon: That doesn't work in const expressions. The advantage here is the string is constructed at compile time and won't allocate at runtime.

storage/build.rs
demosdemon: As mentioned offline, I will add a comment here. 

In a well-formed environment, it should be impossible for cargo to provide a value that does not parse as the value provided was previously a valid `semver::Version`:

https://github.com/rust-lang/cargo/blob/1d71929823e99bb7175098adefdedd928006c321/src/cargo/core/compiler/compilation.rs#L367

https://github.com/rust-lang/cargo/blob/1d71929823e99bb7175098adefdedd928006c321/src/cargo/core/package.rs#L151-L154

After seeing that code, I could also switch to using `CARGO_PKG_VERSION_MAJOR` directly.

storage/src/nodestore.rs
rkuris: Since this is during DB open, I don't mind an allocation here.

rkuris (changes requested):
Discussed in-person. We need to support old versions (actually would be best for any 0.0.x to be grandfathered in, since this code for version checks is really new.

rkuris (changes requested):
Discussed in-person. Would also really like to see us store the full version that created this database and then do the checks to see if we know how to open this particular version. This would allow us to change the rules later as to what we can and can't be compatible with, and might help us identify databases that could possibly need some special handling.

storage/src/nodestore.rs
```diff
@@ -691,18 +642,99 @@ struct Version {
...
-        let mut version_bytes: [u8; Self::SIZE as usize] = Default::default();
-        let version = env!("CARGO_PKG_VERSION");
-        let _ = version_bytes
-            .as_mut_slice()
-            .write_all(format!("firewood {}", version).as_bytes());
-        Self {
-            bytes: version_bytes,
+        // Note that with this magic token of 9 bytes, we can store a version string of
+        // up to 7 bytes. If we always include the major, minor, and patch versions,
+        // then no more than two of three can be 2 digits long.
```
rkuris, line 728: what a mess. I wonder if we can shorten "firewood" to "fwd" going forward. Sounds like a different PR for someone else to address later though.

storage/src/nodestore.rs
```diff
@@ -755,6 +787,60 @@ impl NodeStoreHeader {
         }
     }
 
+    fn validate(&self) -> Result<(), Error> {
+        self.version.validate()?;
+        self.validate_endian_test()?;
+        self.validate_area_size_hash()?;
+        self.validate_ethhash()?;
+        Ok(())
```
rkuris: So much better!

nit: probably would have spaced this out and separated it with debug! messages for readability:

```
    debug!("Validating the version number {}", self.version);
    self.version.validate()?;
    
    debug!("Checking endianness");
    self.validate_endian_test()?;

    ...
```

storage/src/nodestore.rs
```diff
@@ -755,6 +787,60 @@ impl NodeStoreHeader {
...
+            Ok(())
+        } else {
+            Err(Error::new(
+                ErrorKind::InvalidData,
+                "Database cannot be opened due to difference in area size hash",
+            ))
+        }
+    }
+
+    #[cfg(not(feature = "ethhash"))]
```
rkuris, line 836: nit: almost feels like a macro or something would have been better here

rkuris (approved):
nit: AI could have generated a lot more thorough tests for the smaller helpers in seconds.

This looks fine to me otherwise.

storage/src/nodestore.rs
demosdemon: added the logging, used trace to match everything else in the module but switched the one I had in `Version::validate` to debug since it seems more important than a trace line. I didn't want to try to print the version before the `str::from_utf8`. I updated the error message from that to include the bytes in hex.

storage/src/nodestore.rs
demosdemon: opened #970 for that

storage/src/nodestore.rs
demosdemon: macro to check for ethhash or macro to reduce the boilerplate of the validate methods?

For checking ethhash, there's the `cfg!` macro but that would make the code a less clear. 

```rust
fn validate_ethhash(&self) -> Result<(), Error> {
    if self.ethhash == u64::from(cfg!(feature = "ethhash")) {
        Ok(())
    } else {
        Err(Error::new(
            ErrorKind::InvalidData,
            if cfg!(feature = "ethhash") {
                "Database cannot be opened as it was created without ethhash enabled"
            } else {
                "Database cannot be opened as it was created with ethhash enabled"
            },
        ))
    }
}

```

For a custom macro, there's a huge maintainability cost to benefit ratio that doesn't make it worth it.

Issue #831, opened 2025-03-27
Deserialization of a node in a freed area probably produces the wrong error


rkuris:
The problem here is that we currently deserialize `StoredArea<Area<Node, FreeArea>>` but only when it's expted to be a freed area. If there happens to be a node on a free list, bad things are likely to happen since we'll deserialize the node using serde.

The right fix is to remove serde everywhere.

We're not using it in many places any more, and it's clearly slower. Although it's in `Cargo.toml` for the firewood subdirectory, it isn't referenced anywhere in that directory. Only storage is currently using it, and it's only being used to serialize some stuff related to free space management, such as this one and another for `AreaIndex`.

This would also remove a lot of the benchmarks related to serde serialization timings, which show serde is almost always slower.

Bonus points if you can figure out why manual serialization of a full node is a tiny bit slower than serde.
```
     Running benches/serializer.rs (/Users/rkuris/open-source/firewood/target/release/deps/serializer-6c232d65bd4e75a9)
leaf/serde              time:   [69.045 ns 69.287 ns 69.553 ns]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
leaf/manual             time:   [39.281 ns 39.360 ns 39.450 ns]
Found 9 outliers among 100 measurements (9.00%)
  9 (9.00%) high mild

has_value/serde         time:   [239.35 ns 243.83 ns 247.95 ns]
has_value/manual        time:   [80.964 ns 81.715 ns 83.037 ns]
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

1_child/serde           time:   [153.20 ns 154.21 ns 155.28 ns]
1_child/manual          time:   [73.337 ns 73.438 ns 73.569 ns]
Found 14 outliers among 100 measurements (14.00%)
  7 (7.00%) high mild
  7 (7.00%) high severe

2_child/serde           time:   [246.32 ns 251.62 ns 256.86 ns]
2_child/manual          time:   [106.19 ns 106.60 ns 107.05 ns]
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

16_child/serde          time:   [483.69 ns 494.48 ns 504.73 ns]
16_child/manual         time:   [502.68 ns 505.68 ns 509.86 ns]
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
```

rkuris:
[...]

PR #965, opened 2025-06-14
perf: remove some unecessary allocs during serialization

demosdemon:
In #831, it was mentioned that the benchmark for serialization showed we
were slower than serde for a full node. I noticed when walking though
the code that we were allocating a fixed sized array in a tight loop.
This alloc was also unecessary as the `TrieHash` already implemented
`AsRef<[u8]>`.

Removing this alloc and copying directly from the existing hash
significantly improved the benchmarks. There is a small and significant
regression in the leaf benchmark; however, in aggregate the performance
overall has improved.

https://gist.github.com/demosdemon/fa1258e96c8b09851b18e84c45005a28

### DIFF
--- a/triehash/Cargo.toml
+++ b/triehash/Cargo.toml
@@ -8,8 +8,8 @@ license = "MIT OR Apache-2.0"
 edition = "2024"
 
 [dependencies]
-hash-db = { version = "0.16.0", default-features = false }
-rlp = { version = "0.6", default-features = false }
+hash-db = "0.16.0"
+rlp = "0.6"
 
 [dev-dependencies]
 criterion = "0.6.0"
@@ -18,13 +18,6 @@ ethereum-types = { version = "0.15.1" }
 tiny-keccak = { version = "2.0", features = ["keccak"] }
 trie-standardmap = "0.16.0"
 hex-literal = "1.0.0"
-
-[features]
-default = ["std"]
-std = [
-	"hash-db/std",
-	"rlp/std",
-]
 
 [[bench]]
 name = "triehash"
--- a/triehash/src/lib.rs
+++ b/triehash/src/lib.rs
@@ -10,25 +10,9 @@
 //!
 //! This module should be used to generate trie root hash.
 
-#![cfg_attr(not(feature = "std"), no_std)]
-
-#[cfg(not(feature = "std"))]
-extern crate alloc;
-
-#[cfg(feature = "std")]
-mod rstd {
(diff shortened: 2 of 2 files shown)


storage/src/node/mod.rs
```diff
@@ -223,12 +223,7 @@ impl Node {
     pub fn as_bytes<T: ExtendableBytes>(&self, prefix: u8, encoded: &mut T) {
         match self {
             Node::Branch(b) => {
-                let child_iter = b
-                    .children
-                    .iter()
-                    .enumerate()
-                    .filter_map(|(offset, child)| child.as_ref().map(|c| (offset, c)));
-                let childcount = child_iter.clone().count();
+                let childcount = b.children.iter().filter(|c| c.is_some()).count();
```
rkuris: Surprised to see the `clone` was slower than creating the iterator twice. Using `as_ref` is actually the preferred way to do this, as this will not advance the iterator and can be used to fetch the count.

storage/src/node/mod.rs
```diff
@@ -272,24 +267,27 @@ impl Node {
...
+                            hash.write_to_vec(encoded);
                         } else {
                             panic!(
                                 "attempt to serialize to persist a branch with a child that is not an AddressWithHash"
                             );
                         }
                     }
                 } else {
-                    for (position, child) in child_iter {
+                    for (position, child) in b.children.iter().enumerate() {
```
rkuris: Also surprised on this one. Are you sure that unrolling the filter from the iterator is faster than checking it yourself?

rkuris:
The after code still shows serde roughly 2-3x slower than the manual encoding. Presumably this makes manual encoding faster, but it's not clear why.

storage/src/node/mod.rs
demosdemon: `as_ref` just takes mutable reference to the iterator so you can use other methods without taking ownership. Any advancing will persist.

However, this ultimately generated only a negligible change in performance at the expense of more code after compilation and can be reverted.

rkuris: Discussed in person. It's actually by_ref and according to the docs, you can't use consumers and expect it not to advance the iterator.

However, these changes shouldn't be the perf win, it's just the vec allocation we can avoid.

rkuris (changes requested):
Discussed in person, this PR should be trimmed down to just the Vec allocation change.

rkuris (approved):

Issue #967, opened 2025-06-16
chore: [ffi]: Use the full power of cbindgen


alarso16:
A lot of the pointer leaking and type conversions being done manually right now in `ffi/src/lib.rs` is unnecessary, as pointed out in #945 reviews. There should be many changes throughout the file, including:
- Changing `*mut def` to `Option<T> def` or something similar (see the [cbindgen docs](https://github.com/mozilla/cbindgen/blob/master/docs.md#supported-types)
- Improve struct initialization to remove manual `as_ptr` calls to leak the pointer.

alarso16:
Closed with #969. Further work could potentially be done to autogen Go code.

PR #969, opened 2025-06-17
style: use cbindgen to convert to pointers

alarso16:
We use `cbindgen` to manage a lot of the FFI layer, but we are underutilizing it (#967), and can make our Rust-side FFI code much more idiomatic without any changes to the Go side.

Every call that used to take `*const T` or something similar now takes `&T<'_>` or `Option<&T<'_>>`, except freeing the database in `fwd_close_db`, since we would immediately cast to a raw pointer anyway


ffi/src/lib.rs
```diff
@@ -355,7 +339,7 @@ fn batch(
 pub unsafe extern "C" fn fwd_propose_on_db(
     db: *const DatabaseHandle,
```
alarso16: This should be changed as well, but there is a lifetime error with the proposal outliving the database handle

ffi/src/lib.rs
```diff
@@ -152,13 +150,10 @@ pub unsafe extern "C" fn fwd_get_from_proposal(
 /// Internal call for `fwd_get_from_proposal` to remove error handling from the C API
 #[doc(hidden)]
 fn get_from_proposal(
-    db: *const DatabaseHandle,
+    db: &DatabaseHandle<'_>,
     id: ProposalId,
     key: &Value,
 ) -> Result<Value, String> {
-    // Check db is valid.
-    let db = unsafe { db.as_ref() }.ok_or_else(|| String::from("db should be non-null"))?;
```
demosdemon, line 162: Thinking about this, if we want to retain null-safety checks we can do `Option<&DatabaseHandle<'_>>` instead.

It may make sense on some of the methods to protect from bugs in go.

ffi/src/lib.rs
demosdemon: ah, that would be because we're passing in two items using the elided lifetime and rust can't figure out which one owns the lifetime within `DatabaseHandle` and it's certain that it's not both `db` and `values`.

Explicitly defining the lifetimes should work:

```rust
pub unsafe extern "C" fn fwd_propose_on_db<'p>(
    db: &'p DatabaseHandle<'p>,
    nkeys: usize,
    values: &KeyValue,
) -> Value {
// ...
}
```

If that doesn't work, can you paste the full compiler error?

ffi/src/lib.rs
alarso16: Which methods? If we want to rely on c-style guarantees, I think assuming that `db` is nonnull is ok.

demosdemon (approved):

ffi/firewood.h
```diff
@@ -30,7 +30,7 @@ typedef struct DatabaseHandle DatabaseHandle;
  */
 typedef struct Value {
   size_t len;
-  const uint8_t *data;
+  uint8_t *data;
```
rkuris, line 33: Why can't these be const any more? Seems to build with it back in there.

ffi/firewood.h
```diff
@@ -45,7 +45,7 @@ typedef struct KeyValue {
  * Struct returned by `fwd_create_db` and `fwd_open_db`
  */
 typedef struct DatabaseCreationResult {
-  const struct DatabaseHandle *db;
+  struct DatabaseHandle *db;
```
rkuris, line 48: same

ffi/src/lib.rs
rkuris: Does cbindgen panic automatically if nil is passed in for a reference, or is it UB? If it's UB, we have to add the `Option` as @demosdemon suggests. Otherwise, it's too easy to write this golang code in some test and have it drive you into a wild goose chase when it fails in some unknown way:

```go
db, _ = fwd_open_database(...)
db.fwd_get_from_proposal(...)
```

rkuris (changes requested):
Please answer questions before merging.

ffi/src/lib.rs
alarso16: I believe it is UB, but I'm not confident even that `Option` is sufficient. It may need to be an `Option<NonNull<T>>`

rkuris: `Option<T>` is sufficient. It is guaranteed that if something is an option, then a zero pointer is the representation of the `None` value. I am okay with changing all of these to Options and then do this:

```rust
let db = db.ok_or_else(|| String::from("db should be non-null"))?;
```

Wish you could just do this:

```rust
let db = db.ok_or("db should be non-null")?;
```

But you can't because you need a String, and we don't want construct it unless the value is nil.

demosdemon: Yes, if in go we passed in null where we only used `&_` in rust, it would be UB and not panic. 

```rust
let db = db.ok_or("db should be non-null")?;
```

should work because `?` implicitly calls `.into()` on the `ControlFlow::Break(T)` value. and `String: From<str>` so that works.

ffi/firewood.h
demosdemon: these are autogenerated by cbindgen. they're not const because we're giving up the whole struct. Rust no longer has ownership and this matches. When we drop the value, we pass back in a mutable reference. When we drop the value, this cannot be const otherwise that's UB.

ffi/src/lib.rs
alarso16: It did work! I changed this everywhere

ffi/firewood.h
```diff
@@ -192,6 +192,7 @@ struct Value fwd_drop_proposal(const struct DatabaseHandle *db, uint32_t proposa
  * # Panics
  *
  * This function panics if `result` is `null`.
+ *
```
alarso16, line 195: These diffs were just to match the rest of the documentation.

ffi/src/lib.rs
```diff
@@ -608,42 +588,35 @@ fn hash(db: &Db) -> Result<Value, String> {
...
 pub struct Value {
     pub len: usize,
-    pub data: *const u8,
+    pub data: Option<std::ptr::NonNull<u8>>,
 }
 
 impl Display for Value {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        match (self.len, self.data.is_null()) {
+        match (self.len, self.data.is_none()) {
```
rkuris: make this more rusty by matching on self.data instead of using is_none()

rkuris (approved):

Issue #974, opened 2025-06-18
ffi: cgo warning 'not found' search paths


aaronbuchwald:
Compiling Firewood results in the following 'not found' search paths because we populate relative paths to support using the FFI package with local builds placed in `./target/<build-profile/...`

```bash
go test                         
# github.com/ava-labs/firewood/ffi.test
ld: warning: ignoring duplicate libraries: '-lfirewood_ffi', '-lm'
ld: warning: search path '/Users/aaronbuchwald/code/ava-labs/firewood/ffi/libs/aarch64-apple-darwin' not found
ld: warning: search path '/Users/aaronbuchwald/code/ava-labs/firewood/ffi/../target/release' not found
ld: warning: search path '/Users/aaronbuchwald/code/ava-labs/firewood/ffi/../target/maxperf' not found
ld: warning: search path '/Users/aaronbuchwald/code/ava-labs/firewood/ffi/../target/release' not found
ld: warning: search path '/Users/aaronbuchwald/code/ava-labs/firewood/ffi/../target/release' not found
ld: warning: search path '/Users/aaronbuchwald/code/ava-labs/firewood/ffi/../target/release' not found
PASS
ok      github.com/ava-labs/firewood/ffi        0.644s
```

These search paths are passed in [here](https://github.com/ava-labs/firewood/blob/bb49db8584fd96c434bdc9825bf0d1193a70f4a9/ffi/firewood.go#L12)

We should be able to use golang build tags to restrict, which cgo directives are used ie. use `./target` or use the path that is expected to be populated within `firewood-go`, but that puts it up to the user (not desired) rather than all working seamlessly once firewood has been compiled locally.

aaronbuchwald:
A good middle ground may be to do the following:

Firewood uses the local `./target` search paths and contains these warnings when compiled locally. If there's another way to suppress these warnings we can adopt that here.

Before pushing to `firewood-go` we can remove the local search paths, which are never expected to be used when we have attached the static libraries to the source code.

aaronbuchwald:
Closed by #978 - this will still produce "search path not found" errors for local builds as it currently prefers these warnings over requiring an exact build profile match.

PR #978, opened 2025-06-18
feat(ffi): add go generate switch between enabled cgo blocks

aaronbuchwald:
📳 📳 📳 
Adds a go generate script to parse the `firewood.go` file and use an environment variable to switch, which block of CGO directives is enabled.

Adds that command into attach-static-libs GH Action job during push, so that firewood-go skips the local build directory search paths.
📳 📳 📳

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -3,17 +3,21 @@
 // [Firewood]: https://github.com/ava-labs/firewood
 package ffi
 
+//go:generate go run generate_cgo.go
+
 // // Note that -lm is required on Linux but not on Mac.
-// #cgo linux,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-unknown-linux-gnu -lm
-// #cgo linux,arm64 LDFLAGS: -L${SRCDIR}/libs/aarch64-unknown-linux-gnu -lm
-// #cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-apple-darwin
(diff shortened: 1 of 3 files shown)


aaronbuchwald:
Fixes https://github.com/ava-labs/firewood/issues/974

ffi/firewood.go
```diff
@@ -3,17 +3,21 @@
...
+//go:generate go run generate_cgo.go
+
 // // Note that -lm is required on Linux but not on Mac.
-// #cgo linux,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-unknown-linux-gnu -lm
-// #cgo linux,arm64 LDFLAGS: -L${SRCDIR}/libs/aarch64-unknown-linux-gnu -lm
-// #cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-apple-darwin
-// #cgo darwin,arm64 LDFLAGS: -L${SRCDIR}/libs/aarch64-apple-darwin
-// // XXX: last search path takes precedence, which means we prioritize
-// // local builds over pre-built and maxperf over release build
+// // FIREWOOD_CGO_BEGIN_STATIC_LIBS
```
aaronbuchwald, line 9: If desired, we could move as much as possible into the source code, so that `generate_cgo.go` ie. we can define the default value here as well, so that the default is clear when looking at the source code directly as opposed to needing to open the other file.

rkuris (approved):

PR #979, opened 2025-06-19
chore: Simplify generate_cgo script

rkuris:
- Simplify CGO block processing with state machine approach
- Correct isCGODirective to ensure each line has an actual directive
- Remove duplicate CGO LDFLAGS from memory.go, proposal.go, revision.go
- Avoid writing the file if it is unchanged
- Cleaner error reporting
- Improved documentation


rkuris:
I think we could also eliminate the warnings for dev builds too if we just search for what is available and only enable the appropriate cgo directives. It might even be smarter for the generator to always replace a marker with the first found library location.

ffi/firewood.go
```diff
@@ -7,17 +7,17 @@ package ffi
...
 // // #cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/libs/x86_64-apple-darwin
 // // #cgo darwin,arm64 LDFLAGS: -L${SRCDIR}/libs/aarch64-apple-darwin
 // // FIREWOOD_CGO_END_STATIC_LIBS
 // // FIREWOOD_CGO_BEGIN_LOCAL_LIBS
 // #cgo LDFLAGS: -L${SRCDIR}/../target/debug
 // #cgo LDFLAGS: -L${SRCDIR}/../target/release
 // #cgo LDFLAGS: -L${SRCDIR}/../target/maxperf
-// #cgo LDFLAGS: -L/usr/local/lib -lfirewood_ffi
 // // FIREWOOD_CGO_END_LOCAL_LIBS
+// #cgo LDFLAGS: -lfirewood_ffi -lm
```
aaronbuchwald, line 20: If we are moving the `-lm` suffix here, then we should remove or move the comment above stating it's only needed on linux

ffi/generate_cgo.go
```diff
@@ -1,46 +1,73 @@
...
+// FIREWOOD_CGO_BEGIN_<FIREWOOD_LD_MODE>
+// cgo line 1
+// ...
+// cgo line n
+// FIREWOOD_CGO_END_<FIREWOOD_LD_MODE>
+//
+// FIREWOOD_LD_MODE is an environment variable that decides which blocks are activated.
+// The default value for FIREWOOD_LD_MODE is "LOCAL_LIBS" for local development.
+// When building production static libraries, FIREWOOD_LD_MODE is set to "STATIC_LIBS"
+// in the github actions workflow.
```
aaronbuchwald, line 18: It seems this comment is out of date as the env variable `FIREWOOD_LD_MODE` is no longer set in the GitHub Actions?

ffi/generate_cgo.go
```diff
@@ -1,46 +1,73 @@
...
+			path := filepath.Join("../target/", profile, "libfirewood_ffi.a")
+			if _, err := os.Stat(path); err == nil {
+				mode = "LOCAL_LIBS"
+				break
+			}
+		}
+	}
+	return mode
+}
+
```
aaronbuchwald: it looks like this comment and code is doing something more complicated than the top-level comment describes above.

At the top of the file, it states we simply default to local libs, but we do something slightly more complicated by first checking if they exist.

We should make sure these align and imho I'd prefer to avoid checking for local libs to infer what to do and simply make it explicit.

I have frequently had to clean the cache to switch between different builds / sets of feature flags, so it's not uncommon for the build dir to be empty (which would cause this to infer it should use static libs over local) even during a normal local dev workflow. This shouldn't matter much because in local development, I'll never need to run this, but I'd still prefer we simplify this and avoid inferring what to use from the build dir.

If we instead wanted to go this route: https://github.com/ava-labs/firewood/pull/979#issuecomment-2988702051 and use `go generate` to specify an exact build to use, then I could see this being more useful. I don't think the local dev build warnings are much of an issue since this will be used as a library anyways.

ffi/generate_cgo.go
aaronbuchwald: Looks like this is relying on how we infer whether to use a local build based on of it's populated.

aaronbuchwald (approved):

ffi/generate_cgo.go
```diff
@@ -51,27 +53,20 @@ func main() {
 	fmt.Printf("Successfully switched CGO directives to %s mode\n", mode)
 }
 
-// Get the FIREWOOD_LD_MODE environment variable. If it's not set, look for local libs,
-// and if found, set it to "LOCAL_LIBS". Otherwise, set it to "STATIC_LIBS".
-func localOrStaticLibs() string {
+// getFirewoodLdMode returns the FIREWOOD_LD_MODE environment variable.
+// Defaults to "LOCAL_LIBS".
+func getFirewoodLdMode() string {
 	mode, ok := os.LookupEnv("FIREWOOD_LD_MODE")
```
rkuris, line 65: nit: we should probably check that `FIREWOOD_LD_MODE` is one of the two possible values, I can imagine if there's a typo `FIREWOOD_LD_MODE=STATICLIBS` might catch someone off guard.

ffi/firewood.go
rkuris: We were including it everywhere before, and we are still now. It doesn't actually hurt anything except maybe making the linker take another few nanos.

ffi/generate_cgo.go
rkuris: This was switched back, and so the comment is correct again :)

ffi/generate_cgo.go
rkuris: Thanks for fixing this @aaronbuchwald 👍

Issue #940, opened 2025-06-06
chore: metrics registration


rkuris:
We've been really lax on registering metrics, which means that the description of the metric is not available in prometheus. This isn't hard to do but it requires a little more boilerplate on each metric.

The recommended approach is something like this (untested):

Old:
```rust
        counter!("firewood.delete_node", "index" => index_name(area_size_index)).increment(1);
```

New:
```rust
static METRIC_DELETE_NODE: LazyLock<IntCounter> = LazyLock::new(|| {
    register_int_counter_vec!("delete_node", "Number of nodes deleted by index", &["index"])
        .expect("metric registration failure")
});

METRIC_DELETE_NODE.with_label_values(&[index_name(area_size_index)]).inc();
```

Might be better to use a macro.

PR #980, opened 2025-06-20
fix: create metrics registration macros

KushnerykPavel:
resolve https://github.com/ava-labs/firewood/issues/940

### DIFF
--- a/storage/src/macros.rs
+++ b/storage/src/macros.rs
@@ -1,0 +1,22 @@
+
+/// Registers and increments a Prometheus counter with description and labels.
+/// Usage:
+///   firewood_metric!(NAME, DESC, LABELS, [label_values], inc_value)
+///   firewood_metric!(NAME, DESC, [], [], inc_value) for no labels
+#[macro_export]
+macro_rules! firewood_metric {
+    // With labels
+    ($name:expr, $desc:expr, [$($label:expr),*], [$($value:expr),*], $inc:expr) => {{
+        static METRIC: std::sync::LazyLock<prometheus::IntCounterVec> = std::sync::LazyLock::new(|| {
+            prometheus::register_int_counter_vec!($name, $desc, &[$($label),*]).expect("metric registration failure")
+        });
+        METRIC.with_label_values(&[$($value),*]).inc_by($inc);
+    }};
+    // No labels
+    ($name:expr, $desc:expr, [], [], $inc:expr) => {{
+        static METRIC: std::sync::LazyLock<prometheus::IntCounter> = std::sync::LazyLock::new(|| {
+            prometheus::register_int_counter!($name, $desc).expect("metric registration failure")
+        });
+        METRIC.inc_by($inc);
+    }};
+}
--- a/storage/src/nodestore.rs
+++ b/storage/src/nodestore.rs
@@ -47,6 +47,7 @@ use crate::hashednode::hash_node;
 use crate::node::{ByteCounter, Node};
 use crate::{
     CacheReadStrategy, Child, FileBacked, HashType, Path, ReadableStorage, SharedNode, TrieHash,
+    firewood_metric,
 };
 
 use super::linear::WritableStorage;
@@ -565,11 +566,20 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
                 // Update the free list to point to the next free block.
                 *free_stored_area_addr = free_head.next_free_block;
             }
-
-            counter!("firewood.space.reused", "index" => index_name(index as u8))
-                .increment(AREA_SIZES[index]);
-            counter!("firewood.space.wasted", "index" => index_name(index as u8))
-                .increment(AREA_SIZES[index] - n);
+            firewood_metric!(
+                "firewood_space_reused",
+                "Bytes reused from free list by index",
+                ["index"],
+                [index_name(index as u8)],
+                AREA_SIZES[index]
+            );
+            firewood_metric!(
+                "firewood_space_wasted",
+                "Bytes wasted from free list by index",
+                ["index"],
+                [index_name(index as u8)],
+                AREA_SIZES[index] - n
+            );
 
             // Return the address of the newly allocated block.
             trace!(
@@ -580,8 +590,13 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
         }
 
         trace!("No free blocks of sufficient size {index_wanted} found");
-        counter!("firewood.space.from_end", "index" => index_name(index_wanted as u8))
-            .increment(AREA_SIZES[index_wanted as usize]);
+        firewood_metric!(
+                "firewood_space_from_end",
+                "Space allocated from end of nodestore",
+                ["index"],
+                [index_name(index_wanted as u8)],
+                AREA_SIZES[index_wanted as usize]
+            );
         Ok(None)
     }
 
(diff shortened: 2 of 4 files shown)


rkuris (changes requested):
The stack looks like this:
```
(executable) -> firewood -> storage
            \-> prometheus-specific (or datadog or whatever)
```
This fix adds a dependency on a prometheus-specific library in storage, which we don't want, since the executable might want to use datadog or something better.

KushnerykPavel:
Hello, @rkuris ! I have made changes. Could you please review it?

storage/src/macros.rs
```diff
@@ -0,0 +1,29 @@
+
```
demosdemon: please copy the license header here.

storage/src/macros.rs
```diff
@@ -0,0 +1,29 @@
...
+///
+/// Usage:
+///   firewood_metric!("metric_name", "description", value_to_increment_by)
+///   firewood_metric!("metric_name", "description", value_to_increment_by, "label" => "value")
+#[macro_export]
+macro_rules! firewood_metric {
+    // With labels
+    ($name:expr, $desc:expr, $inc:expr, $($labels:tt)+) => {
+        {
+            static METRIC: std::sync::LazyLock<()> = std::sync::LazyLock::new(|| {
```
demosdemon: I like the use of `LazyLock` to get these added only once at runtime.

Let's switch to using `std::sync::Once` since you're not actually initializing a variable. [`call_once`](https://doc.rust-lang.org/stable/std/sync/struct.Once.html#method.call_once) does exactly the same thing (`LazyLock` uses `Once` [itself](https://doc.rust-lang.org/stable/src/std/sync/lazy_lock.rs.html#67)) and we can use it directly here since we don't need to retain the ZST or even a pointer to it.

storage/src/macros.rs
```diff
@@ -0,0 +1,29 @@
...
+                metrics::describe_counter!($name, $desc);
+            });
+            metrics::counter!($name, $($labels)+).increment($inc);
```
demosdemon: Let's leave off the call to `increment($inc)` and have this just be a wrapper around `metrics::describe_counter` and `metrics::counter` and leave calling `.increment` to the caller, like the metrics macros do. 

That way we don't force everyone to use increment when it's not appropriate, such as timestamps/timedeltas where `.absolute` is more appropriate.

demosdemon (changes requested):

storage/src/macros.rs
```diff
@@ -0,0 +1,29 @@
+
+/// Macro to register and use a metric with description and labels.
+/// This macro is a wrapper around the `metrics` crate's `counter!` and `describe_counter!`
+/// macros. It ensures that the description is registered only once.
+///
+/// Usage:
+///   firewood_metric!("metric_name", "description", value_to_increment_by)
+///   firewood_metric!("metric_name", "description", value_to_increment_by, "label" => "value")
+#[macro_export]
+macro_rules! firewood_metric {
```
demosdemon: Rename this to `firewood_counter`, as we're only constructing a counter. `metrics` also has gauges and histograms. Historically, `counter` is used even when not correct and there's examples of this even in our own code.

Lets add equivalent wrappers around `metrics::histogram` (you can include gauge as well, but it's not used yet).

`firewood.delete_node` is a counter, but the rest appear to be more appropriate as [Histograms](https://docs.rs/metrics/latest/metrics/index.html#histograms).

storage/src/macros.rs
demosdemon: also include a module doc comment

e.g.,
```rust
//! Macros for registering metrics on first use.
```

rkuris:
Hey @KushnerykPavel please click the "re-request review" button near the top of this page when you're ready for me to have another look. It will then bother me daily until I review it. Thank you!

storage/src/macros.rs
KushnerykPavel: Am I need to add it in this MR ?

rkuris:
There are still some linter errors in this PR, and now some conflicts.

storage/src/nodestore.rs
```diff
@@ -1031,6 +1051,7 @@ pub struct ImmutableProposal {
 impl ImmutableProposal {
     /// Returns true if the parent of this proposal is committed and has the given hash.
     #[must_use]
+    #[allow(clippy::needless_pass_by_value)]
```
rkuris: please use `expect` instead of `allow`. I think this lint was disabled, so rebasing on main may mean you can delete this line.

storage/src/nodestore.rs
```diff
@@ -539,20 +540,31 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
...
             trace!("Allocating from free list: addr: {address:?}, size: {index}");
             return Ok(Some((address, index as AreaIndex)));
         }
 
         trace!("No free blocks of sufficient size {index_wanted} found");
-        counter!("firewood.space.from_end", "index" => index_name(index_wanted as usize))
-            .increment(AREA_SIZES[index_wanted as usize]);
+        firewood_counter!(
+            "firewood.space.from_end",
+            "Space allocated from end of nodestore",
```
rkuris:
```suggestion
            "Bytes allocated from end of nodestore",
```

storage/src/nodestore.rs
```diff
@@ -616,10 +628,18 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 
         let (area_size_index, _) = self.area_index_and_size(addr)?;
         trace!("Deleting node at {addr:?} of size {area_size_index}");
-        counter!("firewood.delete_node", "index" => index_name(area_size_index as usize))
-            .increment(1);
-        counter!("firewood.space.freed", "index" => index_name(area_size_index as usize))
-            .increment(AREA_SIZES[area_size_index as usize]);
+        firewood_counter!(
+            "firewood.delete_node",
+            "Node deleted from nodestore",
```
rkuris:
```suggestion
            "Nodes deleted",
```

storage/src/nodestore.rs
```diff
@@ -616,10 +628,18 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+            "index" => index_name(area_size_index as usize)
+        )
+        .increment(1);
+        firewood_counter!(
+            "firewood.space.freed",
+            "Space freed in nodestore",
```
rkuris:
```suggestion
            "Bytes freed in nodestore",
```

rkuris (changes requested):
This looks very close!

Please move these macros to the relatively-new firewood-macros so they can be used by both storage and firewood, and double check that all our counter! instances in both project now use firewood_counter!

KushnerykPavel:
> This looks very close!
> 
> Please move these macros to the relatively-new firewood-macros so they can be used by both storage and firewood, and double check that all our counter! instances in both project now use firewood_counter!

As I see firewood-macros create as a proc-macro crate, so it can export only procedural macros not macro like  firewood_counter! Would you like me to create the new shared macro crate and update all usages?

rkuris:
> > This looks very close!
> > Please move these macros to the relatively-new firewood-macros so they can be used by both storage and firewood, and double check that all our counter! instances in both project now use firewood_counter!
> 
> As I see firewood-macros create as a proc-macro crate, so it can export only procedural macros not macro like firewood_counter! Would you like me to create the new shared macro crate and update all usages?

Ah, you're right. Leave them where they are for now.

KushnerykPavel:
@rkuris Hello! So can we merge MR if you do not have any comments ?

KushnerykPavel:
@rkuris @demosdemon Hello! Can we merge it, or may I change sometheing for it will be merged ?

PR #986, opened 2025-06-23
Cleanup Nodestore Public Interface

qusuyan:
The `kind` and `storage` fields of `Nodestore` are public so that `firewood` can read the root hash - but it is already stored in memory when the `Nodestore` is created. This diff converts these two fields back to private.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -417,11 +417,7 @@ impl<T: HashedNodeReader> Merkle<T> {
     pub(crate) fn dump(&self) -> Result<String, Error> {
         let mut result = String::new();
         writeln!(result, "digraph Merkle {{\n  rankdir=LR;").map_err(Error::other)?;
-        if let Some((root_addr, root_hash)) = self
-            .nodestore
-            .root_address_and_hash()
-            .expect("failed to get root address and hash")
-        {
+        if let Some((root_addr, root_hash)) = self.nodestore.root_address_and_hash() {
             writeln!(result, " root -> {root_addr}")
                 .map_err(Error::other)
                 .map_err(|e| FileIoError::new(e, None, 0, None))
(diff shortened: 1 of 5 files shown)


storage/src/nodestore.rs
```diff
@@ -906,13 +896,11 @@ struct FreeArea {
 
 /// Reads from an immutable (i.e. already hashed) merkle trie.
```
qusuyan, line 897: `HashedNodeReader` is implemented for `MutableProposal` as well for dumping

firewood/src/merkle.rs
```diff
@@ -417,11 +417,11 @@ impl<T: HashedNodeReader> Merkle<T> {
...
-        if let Some((root_addr, root_hash)) = self
-            .nodestore
-            .root_address_and_hash()
-            .expect("failed to get root address and hash")
-        {
+        if let Some(root_addr) = self.nodestore.root_address() {
+            let root_hash = self
+                .nodestore
+                .root_hash()
+                .expect("root hash should be present");
```
rkuris:
```suggestion
        if let (Some(root_addr), Some(root_hash)) = (self.nodestore.root_address(), self.nodestore.root_hash()) {
```

firewood/src/merkle.rs
```diff
@@ -2036,29 +2036,30 @@ mod tests {
...
             items.sort();
 
-            let mut merkle = merkle_build_test(items.clone())?;
+            let mut merkle = merkle_build_test::<Vec<u8>, Box<[u8]>>(vec![])?;
+            let mut immutable_merkle: Merkle<NodeStore<Arc<ImmutableProposal>, _>> =
+                merkle.try_into().unwrap();
 
             let mut hashes = Vec::new();
 
             for (k, v) in &items {
```
rkuris: change to use fold

rkuris (approved):
This diff also fixes up one of the earliest firewood merkle tests, which was quite broken.

PR #989, opened 2025-06-24
perf(metrics): Add some metrics around propose and commit times

rkuris:
Adds:

 - firewood.proposal.create for number of created proposals
 - firewood.proposal.create_ms for number of ms spent creating proposals
 - firewood.proposal.commit for number of commits
 - firewood.proposal.commit_ms for number of ms spent commmiting proposals

All metrics are tagged with "success=false" or "success=true" depending on whether the operation succeeded or failed.

To do this easily and start the process of consolidating metrics, created a few metrics macros, start_metrics and finish_metrics, with appropriate documentation.


firewood/src/metrics_macros.rs
```diff
@@ -0,0 +1,83 @@
...
+pub use {finish_metrics, start_metrics};
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn metrics_macros() {
+        // This test verifies that the macros work together properly
+        let start = start_metrics!();
+
+        let result: Result<(), &str> = Ok(());
```
demosdemon: The two step macro feels error-prone. A single macro that accepts an `expr` or `block` (which is also an `expr`) collect the timing without relying on the author to get it right.

```rust
macro_rules! measure_block {
    ($metric_name:expr => $body:expr) => {{
        let start = coarsetime::Instant::now();
        // `expr` is evaluated here
        let result_value = $body; 
        // elapsed should be calculated before anything else
        let elapsed = start.elapsed().as_millis();

        let success_label = if result_value.is_err() { "false" } else { "true" };
        // ... etc

        result_value
    }};
}

let result = measure_block!("test.metric" => {
    Ok::<(), &str>::()
});
```

The `$body` can still contain `awaits` if needed when the macro is invoked within an async block/fn.

In essence, this macro was most of the way there.

---

This is also only walltime. For async futures, it's useful to measure busy vs wait which can be done by wrapping the poll method on the future (e.g., [Instrumented](https://docs.rs/tracing/latest/tracing/instrument/struct.Instrumented.html)). This is something we should be interested w.r.t. i/o.

In which case, I might recommend redesigning the measurement utility to be two different functions that accept `Fn() -> Result<_, _>` or `AsyncFn() -> Result<_, _>` so that busy/wait time can be collected for futures. 

For `Fn`, you could technically measure when the thread parks and I'd be willing to go down that route too if we wanted to.

firewood/src/lib.rs
```diff
@@ -8,7 +8,7 @@
 //! store trie nodes on-disk. Unlike most of state management approaches in the field,
 //! it is not built on top of a generic KV store such as LevelDB/RocksDB. Firewood, like a
 //! B+-tree based database, directly uses the trie structure as the index on-disk. Thus,
-//! there is no additional “emulation” of the logical trie to flatten out the data structure
+//! there is no additional "emulation" of the logical trie to flatten out the data structure
```
alarso16, line 11: What's with this diff?

demosdemon: looks like unicode smart quotes to ascii quotes

firewood/src/metrics_macros.rs
```diff
@@ -0,0 +1,83 @@
...
+macro_rules! finish_metrics {
+    ($metric_name:expr, $result:expr, $start:expr) => {{
+        let result_value = $result;
+        let success_label = if result_value.is_err() { "false" } else { "true" };
+
+        // Increment count counter (base name)
+        metrics::counter!($metric_name, "success" => success_label).increment(1);
+
+        // Increment timing counter (base name + "_ms")
+        metrics::counter!(concat!($metric_name, "_ms"), "success" => success_label).increment($start.elapsed().as_millis());
```
demosdemon: This should probably be a [`Gauge`](https://docs.rs/metrics/latest/metrics/index.html#gauges) or `Histogram` and not a `Counter`.

alarso16: Why would it be a Gauge? It can only go up

demosdemon: This is regarding the elapsed time. The elapsed time can go up or down arbitrarily and the elapsed time for each invocation is not connected to the previous invocation. A histogram sounds more appropriate than a gauge even. A gauge is more appropriate for a single value that can go up or down, like the number of nodes or depth.

firewood/src/metrics_macros.rs
rkuris: I think we can add busy vs wait for async once we have a better async story. Right now, we mostly use firewood in sync mode anyway. The framework here could definitely be used as a starting point for better async statistics when we are ready for them.

firewood/src/lib.rs
rkuris: It showed up as a lint warning in nightly, so I fixed it. No idea how the smart quotes got there in the first place.

firewood/src/metrics_macros.rs
rkuris: This is always increasing, since we use increment(). We use grafana to get the accumulated metrics. This pattern is pervasive through all our golang code and existing rust code, and it works fairly well. You can divide the milliseconds by the counter to get the average time per call.

What you don't get are histogram data, outliers, and p50, p99, etc, which is expensive to collect. If we want those, in theory we could change the macros to emit some histograms.

firewood/src/metrics_macros.rs
rkuris: Switched to proc macros. We might want some additional proc macros later anyway, so having a crate for them seemed reasonable.

firewood-macros/Cargo.toml
alarso16, line 1: Why put these in another crate?

firewood/src/db.rs
```diff
@@ -446,6 +446,7 @@ impl Proposal<'_> {
         Ok(self.create_proposal(batch)?.into())
     }
 
+    #[crate::metrics("firewood.proposal.create", "database proposal creation")]
```
alarso16, line 449: Whoa this looks clean

firewood-macros/Cargo.toml
rkuris: You can't put proc macros in the same crate, unfortunately.

firewood/src/db.rs
rkuris: Agreed. I like this approach because I hate looking at metrics code when you're reading a function, and this puts it all out of sight.

firewood-macros/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    };
+
+    quote! {
+        #(#fn_attrs)*
+        #fn_vis #fn_sig {
+            #registration_code
+
+            let __metrics_start = coarsetime::Instant::now();
+
+            let __metrics_result = (|| #fn_block)();
```
demosdemon: wrapping this around an IILE is not necessary. the body will be evaluated like normal if you wrap in braces. The IILE also makes it impossible to use in async functions. (The braces aren't necessary as they're already included in the tokens, but this is explicit without breaking anything).

```suggestion
            let __metrics_result = { #fn_block };
```

firewood-macros/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+use syn::parse::{Parse, ParseStream};
+use syn::{ItemFn, Lit, ReturnType, Token, parse_macro_input};
+
+/// Arguments for the metrics macro
+struct MetricsArgs {
+    name: String,
+    description: Option<String>,
+}
+
+impl Parse for MetricsArgs {
```
demosdemon, line 17: I recommend [darling](https://docs.rs/darling/latest/darling/) in the future before this gets too wildly

demosdemon (approved):
comments are just nits and can be followups

firewood-macros/src/lib.rs
rkuris: Will look into this for a followup!

PR #998, opened 2025-06-26
feat(checker): add freelist traversal

qusuyan:


storage/src/checker.rs
```diff
@@ -41,6 +41,12 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         }
 
         // 3. check the free list - this can happen in parallel with the trie traversal
+        let freelists = self.get_free_lists();
+        for (area_idx, head_addr) in freelists.iter().enumerate() {
+            #[allow(clippy::cast_possible_truncation)]
+            let area_size = Self::size_from_area_index(area_idx as u8);
+            self.check_freelist(area_size, *head_addr, &mut visited)?;
```
rkuris: nit: prefer calling out the conversion rather than the clippy warning:
```suggestion
            let area_idx = u8::try_from(n).expect("freelists index always fit in u8");
            let area_size = Self::size_from_area_index(area_idx);
            self.check_freelist(area_size, *head_addr, &mut visited)?;
```
otherwise a comment

storage/src/checker.rs
```diff
@@ -132,12 +168,83 @@ mod test {
...
         let complement = visited.complement();
         assert_eq!(complement.into_iter().collect::<Vec<_>>(), vec![]);
     }
+
+    #[test]
+    fn test_traverse_correct_freelist() {
+        use rand::rngs::StdRng;
+        use rand::{Rng, SeedableRng, rng};
+
+        let seed = std::env::var("FIREWOOD_STORAGE_TEST_SEED")
```
rkuris: nit: really wish this were more centralized instead of copied everywhere.

storage/src/nodestore.rs
```diff
@@ -323,6 +323,39 @@ impl<T: ReadInMemoryNode, S: ReadableStorage> NodeStore<T, S> {
...
+impl<T, S: ReadableStorage> NodeStore<T, S> {
+    // Read a [FreeArea] from the provided [LinearAddress]
+    // `addr` is the address of a StoredArea in the ReadableStorage.
+    pub(crate) fn read_free_area_with_area_index_from_disk(
+        &self,
+        addr: LinearAddress,
+    ) -> Result<(FreeArea, AreaIndex), FileIoError> {
+        let free_area_addr = addr.get();
+        let free_head_stream = self.storage.stream_from(free_area_addr)?;
+        let free_head: StoredArea<Area<Node, FreeArea>> = serializer()
```
rkuris: We need to centralize this code, especially since it is changing (serde is being removed). Please see if you can call something central, and you're going to conflict with @demosdemon

storage/src/nodestore.rs
```diff
@@ -323,6 +323,39 @@ impl<T: ReadInMemoryNode, S: ReadableStorage> NodeStore<T, S> {
     }
 }
 
+impl<T, S: ReadableStorage> NodeStore<T, S> {
+    // Read a [FreeArea] from the provided [LinearAddress]
+    // `addr` is the address of a StoredArea in the ReadableStorage.
+    pub(crate) fn read_free_area_with_area_index_from_disk(
```
rkuris: nit: naming: it doesn't necessary ready from disk, it reads from the ReadableStorage, which could be in memory.

Since the method operates on NodeStore<T, S> anyway and S has to be ReadableStorage, it should be assumed that "read_free_area_and_index" is enough. I'd also maybe argue that it's obvious it's reading FreeArea and AreaIndex from the return values, so maybe it should just be "read_free".

Also add a comment that says it expects the caller to pass the address of a free area.

storage/src/nodestore.rs
```diff
@@ -323,6 +323,39 @@ impl<T: ReadInMemoryNode, S: ReadableStorage> NodeStore<T, S> {
...
+        let free_head: StoredArea<Area<Node, FreeArea>> = serializer()
+            .deserialize_from(free_head_stream)
+            .map_err(|e| {
+                self.storage.file_io_error(
+                    Error::new(ErrorKind::InvalidData, e),
+                    free_area_addr,
+                    Some("read_free_area_with_area_index_from_disk".to_string()),
+                )
+            })?;
+        let StoredArea {
```
rkuris: This is great code, but a bit tricky. Centralizing this code would make this a smaller diff.

storage/src/nodestore.rs
```diff
@@ -506,49 +539,30 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
                 .file_io_error(e, 0, Some("allocate_from_freed".to_string()))
         })?;
 
-        if let Some((index, free_stored_area_addr)) = self
+        if let Some(index) = self
```
rkuris: please move this to a different PR

storage/src/nodestore.rs
```diff
@@ -1588,17 +1620,37 @@ pub(crate) mod nodestore_test_utils {
         AREA_SIZES[area_size_index as usize]
     }
 
+    // Helper function to write the NodeStoreHeader
     pub(crate) fn write_header<S: WritableStorage>(
         nodestore: &NodeStore<Committed, S>,
-        root_addr: LinearAddress,
         size: u64,
+        root_addr: Option<LinearAddress>,
```
rkuris, line 1746: This is really odd, how did it work before when root_addr was None.

storage/src/checker.rs
qusuyan: Maybe it should go into a separate crate? It is used in `firewood`, `storage`, and `fwdctl`

storage/src/nodestore.rs
qusuyan: This was the test utils I introduced in the previous PR so it was only used on trie with data.

storage/src/checker.rs
```diff
@@ -41,6 +43,12 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         }
 
         // 3. check the free list - this can happen in parallel with the trie traversal
+        let freelists = self.get_free_lists();
+        for (idx, head_addr) in freelists.iter().enumerate() {
+            let area_idx = u8::try_from(idx).expect("area index should always fit in u8");
+            let area_size = Self::size_from_area_index(area_idx);
+            self.check_freelist(area_size, *head_addr, &mut visited)?;
```
rkuris: check_freelist has side effects as it takes a mut. Therefore, it's not just checking the freelist, so it is likely named incorrectly. I'll confirm that when I review the new function, but it probably needs a new name.

storage/src/checker.rs
```diff
@@ -66,16 +74,45 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 
         Ok(())
     }
+
+    /// Traverse all the free areas in the freelist
```
rkuris, line 73: clue 2 also points to a name that could be better, and in fact there's a better suggestion right there: `traverse_freelist`. So we aren't just checking `FreeList` we're actually traversing all the entries. Still doesn't explain `mut`.

storage/src/checker.rs
```diff
@@ -66,16 +74,45 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+    fn check_freelist(
+        &self,
+        freelist_area_size: u64,
+        head_addr: Option<LinearAddress>,
+        visited: &mut LinearAddressRangeSet,
+    ) -> Result<(), CheckerError> {
+        let mut cur_free_area_addr = head_addr;
+        while let Some(free_area_addr) = cur_free_area_addr {
```
rkuris:
```suggestion
        mut head_addr: Option<LinearAddress>,
        visited: &mut LinearAddressRangeSet,
    ) -> Result<(), CheckerError> {
        while let Some(free_area_addr) = head_addr {
```

storage/src/checker.rs
```diff
@@ -66,16 +74,45 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+            visited.insert_area(free_area_addr, freelist_area_size)?;
+            let (free_head, read_index) = self.read_free_area(free_area_addr)?;
```
rkuris: nit: rename
```suggestion
            let (free_head, stored_area_index) = self.read_free_area(free_area_addr)?;
```

storage/src/checker.rs
```diff
@@ -66,16 +74,45 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+            let free_area_size = Self::size_from_area_index(read_index);
```
rkuris: nit: rename
```suggestion
            let stored_free_area_size = Self::size_from_area_index(read_index);
```

rkuris (changes requested):
I am not crazy about the separation of concerns here, so please change this code to work like this.

Create a function with this signature:

```rust
fn visit_freelists<V>(&self, mut visitfn: V) -> Result<(), CheckerError>
where
    V: FnMut(LinearAddress, u64) -> Result<(), CheckerError>,
```

This new function should:
1. Call `get_free_lists`, iterate, and enumerate
2. walk the free list, and call `visitfn` for each free list item

Please rename `get_free_lists` to `freeelists`; see https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

Also a good read:
https://randycoulman.com/blog/2017/01/17/avoid-get-and-set/

Maybe someday there will be a lint for it:
https://github.com/rust-lang/rust-clippy/issues/1673

storage/src/checker.rs
```diff
@@ -66,16 +69,27 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 
         Ok(())
     }
+
+    /// Traverse all the free areas in the freelist
+    fn traverse_freelist(&self, visited: &mut LinearAddressRangeSet) -> Result<(), CheckerError> {
```
rkuris: naming: This function isn't traversing the freelist, it's adding to visited, so maybe `visit_freelist`

storage/src/checker.rs
```diff
@@ -66,16 +69,27 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        for free_area in self.freelist_iter() {
```
rkuris: This should be on the freelist type, not in NodeStore

storage/src/checker.rs
```diff
@@ -66,16 +69,27 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+            let (addr, area_size) = free_area?;
```
rkuris: area_index

storage/src/checker.rs
```diff
@@ -132,12 +146,51 @@ mod test {
         high_watermark += root_area;
 
         // write the header
-        write_header(&nodestore, root_addr, high_watermark);
+        write_header(
```
rkuris: test_write_header

storage/src/checker.rs
```diff
@@ -132,12 +146,51 @@ mod test {
...
+            &mut nodestore,
+            high_watermark,
+            Some(root_addr),
+            Default::default(),
```
rkuris: nit: FreeLists::default()

storage/src/checker.rs
```diff
@@ -132,12 +146,51 @@ mod test {
...
 
         // verify that all of the space is accounted for - since there is no free area
         let mut visited = LinearAddressRangeSet::new(high_watermark).unwrap();
         nodestore.traverse_trie(root_addr, &mut visited).unwrap();
         let complement = visited.complement();
         assert_eq!(complement.into_iter().collect::<Vec<_>>(), vec![]);
     }
+
+    #[test]
+    fn test_traverse_correct_freelist() {
```
rkuris: nit: traverse_correct_freelist

storage/src/checker.rs
```diff
@@ -132,12 +146,51 @@ mod test {
...
+        use rand::Rng;
+
+        let mut rng = crate::test_utils::seeded_rng();
+
+        let memstore = MemStore::new(vec![]);
+        let mut nodestore = NodeStore::new_empty_committed(memstore.into()).unwrap();
+
+        // write free areas
+        let mut high_watermark = NodeStoreHeader::SIZE;
+        let mut free_list: FreeLists = Default::default();
```
rkuris:
```suggestion
        let mut free_lists = Default::default();
```

storage/src/checker.rs
```diff
@@ -132,12 +146,51 @@ mod test {
...
+        for (area_index, area_size) in area_sizes().iter().enumerate() {
+            let mut next_free_block = None;
+            let num_free_areas = rng.random_range(0..4);
+            for _ in 0..num_free_areas {
```
rkuris, line 179: Can you use an existing method to add to a freelist?

storage/src/lib.rs
```diff
@@ -122,7 +122,41 @@ pub enum CheckerError {
         intersection: Vec<Range<LinearAddress>>,
     },
 
+    /// Freelist area size does not match
+    #[error(
+        "Free area {address} has miss matching size: has {size} bytes but is in freelist {freelist_size}"
```
rkuris:
```suggestion
        "Free area {address} has {size} bytes but is in freelist {freelist_size}"
```
Should this be: "Free area {address} of size {size} is in freelist {freelist_index} but should be in {actual_index}"

storage/src/nodestore.rs
```diff
@@ -891,10 +871,45 @@ impl NodeStoreHeader {
 /// A [`FreeArea`] is stored at the start of the area that contained a node that
 /// has been freed.
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Deserialize)]
-struct FreeArea {
+pub(crate) struct FreeArea {
     next_free_block: Option<LinearAddress>,
 }
 
+impl FreeArea {
+    pub(crate) const fn next_free_block(self) -> Option<LinearAddress> {
```
rkuris, line 879: nit: pub, not pub(crate)

storage/src/nodestore.rs
```diff
@@ -891,10 +871,45 @@ impl NodeStoreHeader {
 /// A [`FreeArea`] is stored at the start of the area that contained a node that
 /// has been freed.
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Deserialize)]
-struct FreeArea {
+pub(crate) struct FreeArea {
```
rkuris, line 874: Is this needed?

storage/src/nodestore.rs
```diff
@@ -1555,21 +1570,120 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
 }
 
+pub struct FreeListIterator<'a, S: ReadableStorage> {
+    storage: &'a S,
+    freelist: std::iter::Enumerate<std::slice::Iter<'a, Option<LinearAddress>>>,
+    current_area_index: Option<AreaIndex>,
+    next_addr: Option<LinearAddress>,
+}
+
+#[allow(dead_code)]
```
rkuris: do we need this?

storage/src/nodestore.rs
```diff
@@ -1555,21 +1570,120 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+    pub(crate) fn new(storage: &'a S, freelist: &'a FreeLists) -> Self {
+        Self {
+            storage,
+            freelist: freelist.iter().enumerate(),
+            current_area_index: None,
+            next_addr: None,
+        }
+    }
+
+    pub(crate) fn start_from(storage: &'a S, next_addr: LinearAddress) -> Self {
```
rkuris: #[cfg(test)]

storage/src/nodestore.rs
```diff
@@ -1555,21 +1570,120 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+    type Item = Result<(LinearAddress, AreaIndex), FileIoError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // find the next address to read from
+        let next_addr = match self.next_addr {
+            // if set, then we are in the middle of a freelist, use the set address directly
+            Some(next_addr) => next_addr,
+            // if not set, then we are at the end of the current freelist, find the next free list head
+            None => loop {
+                if let Some((area_index, free_list_head)) = self.freelist.next() {
```
rkuris: rewrite with `std::iter::find`

storage/src/checker.rs
```diff
@@ -60,31 +63,43 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
             }
         }
 
         Ok(())
     }
+
+    /// Traverse all the free areas in the freelist
+    fn visit_freelist(&self, visited: &mut LinearAddressRangeSet) -> Result<(), CheckerError> {
+        for free_area in self.freelists_iter() {
+            let (addr, area_index) = free_area?;
```
rkuris, line 76:
```suggestion
        for (addr, area_index) in self.freelists_iter()? {
```

storage/src/nodestore.rs
```diff
@@ -1601,13 +1616,95 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
     }
 }
 
+pub struct FreeListIterator<'a, S: ReadableStorage> {
```
rkuris, line 1619: pub(crate)? Otherwise, this is exposed to the outside interface. This might be what we want, but let's be clearer about what is exposed and what isn't.

storage/src/nodestore.rs
```diff
@@ -1601,13 +1616,95 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        }
+
+        // update the next address to the next free block
+        self.next_addr = free_head.next_free_block();
+        Some(Ok(next_addr))
+    }
+}
+
+impl<T, S: ReadableStorage> NodeStore<T, S> {
+    #[allow(dead_code)]
```
rkuris, line 1673:
```suggestion
    #![allow(dead_code, reason = "Explain why here")] 

```

storage/src/nodestore.rs
```diff
@@ -1616,9 +1713,17 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
 pub(crate) mod nodestore_test_utils {
     use super::*;
 
+    pub(crate) const fn area_sizes() -> &'static [u64] {
```
rkuris, line 1716: why did you go this route instead of just making AREA_SIZES `pub(crate)`? It's already an immutable static.

storage/src/nodestore.rs
```diff
@@ -1616,9 +1713,17 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
+        &AREA_SIZES
+    }
+
+    pub(crate) fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
+        super::area_size_to_index(n)
+    }
```
rkuris, line 1722: This looks like a visibility problem. Let's look closer at our 1:1

storage/src/nodestore.rs
rkuris: Much clearer with the renaming!

qusuyan:
Will reopen after the iterator is pushed #1015

PR #1004, opened 2025-06-28
perf: Cache the latest view

### LINKED ISSUE #1003: perf: Improve EVM contention
The interface from coreth now pretty much exclusively uses view-sync. During bootstrapping and general block execution, it looks like this:

1. Repeatedly call get_from_root with the same revision hash but with different keys
2. Create a proposal
3. Asynchronously commit the proposal, then return to step 1 with the hash from what we are committing

Right now, we contend on steps 1 and 3, because `view_sync` called from `get_from_root` in step 1 locks the revision manager and so does `commit` in step 3. This either slows down the proposal or slows down the commit, but there's no reason we shouldn't be able to do these in parallel.

The naive way to fix this would be to cache a few revisions. This won't quite help though, since it's likely the commit from step 3 will start before the first read in step 1. A slightly better way would be to save the proposal in a cache at the start of the commit, but update this cache when the commit finishes so we're now referencing the committed revision.

rkuris:
Because of the usage pattern for coreth, there is contention between committing and getting values. Coreth attempts to commit the latest version asynchronously while fetching values from it.

This is fixed by adding a single cached revision. In the normal case, there will be many reads on some revision, followed by creating a proposal, followed by an asynchronous commit. During this commit, there will be additional reads, so we save the view just before committing.

Fixes #1003

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -235,11 +235,13 @@ fn get_from_root(
     let value = match cached_view.as_ref() {
         // found the cached view, use it
         Some((root_hash, view)) if root_hash == &requested_root => {
+            counter!("firewood.ffi.cached_view.hit").increment(1);
             view.val_sync_bytes(key.as_slice())
         }
         // If what was there didn't match the requested root, we need a new view, so we
         // update the cache
         _ => {
+            counter!("firewood.ffi.cached_view.miss").increment(1);
             let rev = view_sync_from_root(db, root)?;
             let result = rev.val_sync_bytes(key.as_slice());
             *cached_view = Some((requested_root.clone(), rev));


ffi/src/lib.rs
```diff
@@ -215,17 +230,42 @@ fn get_from_root(
...
-    // Get the revision associated with the root hash.
+    let requested_root = root.as_slice().try_into()?;
+    let mut cached_view = db.cached_view.lock().expect("cached_view lock is poisoned");
+    let value = match cached_view.as_ref() {
+        // found the cached view, use it
+        Some((root_hash, view)) if root_hash == &requested_root => {
+            view.val_sync_bytes(key.as_slice())
+        }
+        // found a cached view, but it's for a different root, so we need to get a new view
+        Some(_) => {
```
alarso16: These two cases are handled the same - can we combine them?

ffi/src/lib.rs
```diff
@@ -883,9 +942,26 @@ fn manager_config(
...
+/// This function panics if:
+/// * `db` is `None` (null pointer)
+/// * A lock is poisoned
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_close_db(db: *mut DatabaseHandle) {
-    let _ = unsafe { Box::from_raw(db) };
+pub unsafe extern "C" fn fwd_close_db(db: Option<&mut DatabaseHandle>) {
+    let db_handle = db.expect("db should be non-null");
+
+    // Explicitly clear the downstream items. Drop will do these in order, so this
```
alarso16, line 950: Why must we drop the map first?

alarso16 (commented):
This cached proposal seems reasonable, but I worry about this causing performance issues for API calls. Also, long down the road it seems like we would like to be able to parallelize reads

ffi/src/lib.rs
```diff
@@ -52,6 +53,10 @@ pub struct DatabaseHandle<'p> {
     // Keep proposals first, as they must be dropped before the database handle is dropped due to lifetime
     // issues.
     proposals: RwLock<HashMap<ProposalId, Arc<Proposal<'p>>>>,
+
+    /// A single cached view to improve performance of reads while committing
+    cached_view: Mutex<Option<(HashKey, Box<dyn DbViewSyncBytes>)>>,
```
demosdemon, line 58: This pattern smells like [dashmap](https://github.com/xacrimon/dashmap) might be really useful. It splits up concurrent map access into multiple locks partitioned by hashes.

ffi/src/lib.rs
demosdemon: This is an attempt to defend against the fact that `DatabaseHandle` is an "illegal" self-referential struct.

https://github.com/ava-labs/firewood/blob/5a4b85acb2a4f6d25933d4c811f84c0c57a6485f/ffi/src/lib.rs#L45-L57

`proposals` has a lifetime bound to the `db` field. Which is not legal in safe rust. We make it work by passing the struct through unsafe barriers which erases the lifetime provenance. It's technically UB to interact with `&mut DatabaseHandle` without first going through a `Pin`. But, pin is not a compiler construct, just a SDK pattern to indicate you cannot move the value behind the reference.

ffi/src/lib.rs
```diff
@@ -509,7 +549,26 @@ fn commit(db: Option<&DatabaseHandle<'_>>, proposal_id: u32) -> Result<(), Strin
         .map_err(|_| "proposal lock is poisoned")?
         .remove(&proposal_id)
         .ok_or("proposal not found")?;
-    proposal.commit_sync().map_err(|e| e.to_string())
+
+    // Get the proposal hash and cache the view. We never cache an empty proposal.
```
alarso16, line 548: What do you mean "we never cache an empty proposal"? We definitely should cache reading to a proposal with no changes, but I also don't know why that would matter

ffi/src/lib.rs
demosdemon: +1, clippy usually has a lint for this that might be squashed because of the comments.

demosdemon:
https://github.com/ava-labs/firewood/blob/a483d5827b27b7d5847e5d5a5bfcb5bdb555211c/ffi/src/lib.rs#L494-L502

Why do we keep the lock acquired for the entire time `proposal.propose_sync` is running only to immediately drop it and acquire the write lock again later. Can't we just clone the arc and then drop the lock? `proposal` here is a reference to an arc bound to the guard. cloning the arc would release the bound on guard.

```rust
    // Get proposal from ID.
    let proposal = db
        .proposals
        .write()
        .expect("failed to acquire write lock on proposals")
        .get(&proposal_id)
        .ok_or("proposal not found")?
        .clone();
    let new_proposal = proposal.propose_sync(batch).map_err(|e| e.to_string())?;
```

or more expliclity
```rust
    let new_proposal = {
        // Get proposal from ID.
        db.proposals
            .write()
            .expect("failed to acquire write lock on proposals")
            .get(&proposal_id)
            .ok_or("proposal not found")?
            .clone()
    } // force lock to be released
    .propose_sync(batch)
    .map_err(|e| e.to_string())?;
```

ffi/src/lib.rs
```diff
@@ -52,6 +53,10 @@ pub struct DatabaseHandle<'p> {
     // Keep proposals first, as they must be dropped before the database handle is dropped due to lifetime
     // issues.
     proposals: RwLock<HashMap<ProposalId, Arc<Proposal<'p>>>>,
```
demosdemon, line 55: because the value of the hashmap is an `Arc<_>`, there is no need for this to be a `RwLock`. A regular `Mutex` where you always clone the `Arc` on get will be more performant. See [a](https://users.rust-lang.org/t/rwlock-slow-to-acquire-read-lock-even-when-uncontended/100259).

Wrapping it in a newtype might help:

```rust
struct ArcMap<K, V>(Mutex<HashMap<K, Arc<V>>>);

impl ArcMap<K, V> {
    fn get(&self, key: &impl Borrow<K>) -> Option<Arc<V>> { self.0.lock().expect("to not be poisoned").get(key).map(Arc::clone) /* or .cloned() */ }
    fn insert(&self, key: K, val: Arc<V>) -> Option<(K, Arc<V>)> { todo!() }
}
```

Though, dashmap will still be better as it uses distributed locks.

alarso16:
> https://github.com/ava-labs/firewood/blob/a483d5827b27b7d5847e5d5a5bfcb5bdb555211c/ffi/src/lib.rs#L494-L502
> 
> Why do we keep the lock acquired for the entire time `proposal.propose_sync` is running only to immediately drop it and acquire the write lock again later. Can't we just clone the arc and then drop the lock? `proposal` here is a reference to an arc bound to the guard. cloning the arc would release the bound on guard.

This might actually be a bigger issue (i.e. a race). What happens if we we attempt to commit the base proposal while we are making the new proposal? The commit might fail because the reference count is too high. We should probably hold the write lock for the entirety of this call

ffi/src/lib.rs
rkuris: I mean a proposal that has a root hash of None.

ffi/src/lib.rs
rkuris: Let's avoid this. There are usually only one or two proposals at any given time and we win nothing by spanning across these.

rkuris:
> Why do we keep the lock acquired for the entire time `proposal.propose_sync` is running only to immediately drop it and acquire the write lock again later. Can't we just clone the arc and then drop the lock? `proposal` here is a reference to an arc bound to the guard. cloning the arc would release the bound on guard.

We probably shouldn't need to do this, but that's a different fix.

ffi/src/lib.rs
rkuris: Except there will probably only be 1 or 2 of these at any given time.

rkuris:
> This might actually be a bigger issue (i.e. a race). What happens if we we attempt to commit the base proposal while we are making the new proposal? The commit might fail because the reference count is too high. We should probably hold the write lock for the entirety of this call

I don't understand what you mean by "reference count is too high" which can cause a commit to fail. I think you might be referring to the fact that if you have a cloned `Proposal` you aren't allowed to commit it. You can, however, have a proposla based on a cloned `NodeStore` and still commit the proposal.

There are a bunch of possibilities as to when P1 commits during P2's creation. Without races:

1) P2 is fully created before P1 commits. In this case, P2 will be reparented to R1, P1 will be unreferenced (assuming there is no other proposal, say P3, that references P1) and will be cleaned up toward the end of the commit.
2) P1 fully commits before P2 starts. This case is generally impossible: if you have a clone of P1, you can't commit P1, so there is no way to create P2.

During the creation of P2, the first thing we do is clone the parent P1 `NodeStore` and build a new `NodeStore` with it as a parent, which increases the reference count of P1's `NodeStore`. If P1 commits while we're referencing that, everything is fine -- some nodes will be read from P1's `NodeStore` and this will stick around until we're done with P2. Interestingly, we don't lock the manager until the proposal is fully created, so all the merkle operations happen without a lock and this isn't a problem.

At the very end of the merkle operations, we convert the `NodeStore` into a `NodeStore<Arc<ImmutableProposal>, _>. This does mean that the proposal might have missed it's opportunity to be reparented if the commit happened during the merkle operations. This is okay, because the commit check compares hashes.

One thing that might be a problem here is the case where there are two revisions with different parents with the same hash. That is, let's say you have R1->P1->P2->P3 and R1 and P2 have the same hash, and you choose to commit P3. I can create an issue for that to investigate that further, but that's not something that should happen in coreth anyway.

ffi/src/lib.rs
```diff
@@ -215,17 +230,35 @@ fn get_from_root(
...
+    let requested_root = root.as_slice().try_into()?;
+    let mut cached_view = db.cached_view.lock().expect("cached_view lock is poisoned");
+    let value = match cached_view.as_ref() {
+        // found the cached view, use it
+        Some((root_hash, view)) if root_hash == &requested_root => {
+            view.val_sync_bytes(key.as_slice())
+        }
+        // If what was there didn't match the requested root, we need a new view, so we
+        // update the cache
+        _ => {
```
alarso16, line 243: This may cause issues later with EVM performance if APIs are reading from a revision parallel to execution, causing a high cache miss rate. This only makes a difference in the case that there a proposal is also being committed, so we can delay this for a further performance improvement.

ffi/src/lib.rs
```diff
@@ -509,7 +542,26 @@ fn commit(db: Option<&DatabaseHandle<'_>>, proposal_id: u32) -> Result<(), Strin
...
+        let mut guard = db.cached_view.lock().expect("cached_view lock is poisoned");
+        match db.view_sync(proposal_hash.clone()) {
+            Ok(view) => *guard = Some((proposal_hash, view)),
+            Err(_) => *guard = None, // Clear cache on error
+        }
+        drop(guard);
+    }
+
+    // Commit the proposal
+    let result = proposal.commit_sync().map_err(|e| e.to_string());
```
alarso16, line 561: One further optimization we may want to have is being able to have clearly defined behavior for proposing and committing simultaneously - e.g what if another thread is attempting to create a proposal on this proposal? see #1007

ffi/src/lib.rs
alarso16: One option to avoid this may be to store all revisions accessible in a map similar to proposals, and explicitly calling some `drop` method when they are no longer needed. This requires further management for consumers to clear old data, but would avoid lock contention for the case described above (see #1008 )

alarso16 (approved):
I think this is ok for now. We likely should have some way to avoid contention over the cache for concurrent readers (e.g. API and EVM execution), and this PR does highlight some interesting inefficiencies of our current interface to be improved.

rkuris:
> https://github.com/ava-labs/firewood/blob/a483d5827b27b7d5847e5d5a5bfcb5bdb555211c/ffi/src/lib.rs#L494-L502
> 
> Why do we keep the lock acquired for the entire time `proposal.propose_sync` is running only to immediately drop it and acquire the write lock again later. Can't we just clone the arc and then drop the lock? `proposal` here is a reference to an arc bound to the guard. cloning the arc would release the bound on guard.


If we clone the Arc, then the commit will fail due to the commit check requiring that a proposal not be cloned, rather than just waiting. For now, we prefer the wait method, but it would be even better if these could run in parallel. A few new tickets got created today to address these (notably #1007 )

demosdemon:
Spoke offline with Ron to understand the use of `Arc::into_inner` during commit. I have longer thoughts that I will collect later. But it ultimately feels like a wart we will need to correct.

w.r.t. this PR, I would like to add metrics for cache hit vs miss now before we ship and then review the `db` lifetime and `Arc` around proposal separately.

demosdemon (approved):

rkuris:
> Why do we keep the lock acquired for the entire time `proposal.propose_sync` is running only to immediately drop it and acquire the write lock again later. Can't we just clone the arc and then drop the lock? `proposal` here is a reference to an arc bound to the guard. cloning the arc would release the bound on guard.

We can do this once we merge #1010

rkuris:
> Spoke offline with Ron to understand the use of `Arc::into_inner` during commit. I have longer thoughts that I will collect later. But it ultimately feels like a wart we will need to correct.

Addressed in #1010

Issue #999, opened 2025-06-26
IO Error in io_uring


alarso16:
Due to handling while write nodes in the io_uring case (see `storage/src/nodestore.rs/flush_nodes`), any IO error that occurs may not be propagated. Example: if some nodes cannot be written to disk, the kernel may reschedule them, meaning that no error is returned when checked, but not all nodes were written.

PR #1005, opened 2025-06-29
fix: Report IO errors

rkuris:
Previously, if an error occurred while flushing nodes, it was silently ignored during io-uring processing. To reproduce this, I created a tiny filesystem using an AWS instance running ubuntu:

    sudo mkdir /mnt/tmpfs
    sudo mount -t tmpfs -o size=5k,mode=1777,noatime tmpfs /mnt/tmpfs

I then ran a few fwdctl commands, which passed on main:

    cargo run --bin fwdctl --features logger -- create /mnt/tmpfs/test.db
    target/debug/fwdctl insert --db /mnt/tmpfs/test.db a "$(tr -d '\n' < /etc/passwd)"
    target/debug/fwdctl insert --db /mnt/tmpfs/test.db b "$(tr -d '\n' < /etc/passwd)"
    target/debug/fwdctl insert --db /mnt/tmpfs/test.db c "$(tr -d '\n' < /etc/passwd)"

When running this after these fixes, the following is produced from the final insert, which fills the disk:

    Error: FileIO(FileIoError { inner: Os { code: 28, kind: StorageFull, message: "No space left on device" }, filename: Some("/mnt/tmpfs/test.db"), offset: 8544, context: Some("write failure") })

This fix restructures the pinned buffer pool to add an offset, which is only used for error reporting. This offset is also used as a flag indicating the entry is in use rather than using a bool.

### DIFF
--- a/storage/src/nodestore.rs
+++ b/storage/src/nodestore.rs
@@ -1318,35 +1318,50 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
     #[fastrace::trace(short_name = true)]
     #[cfg(feature = "io-uring")]
     pub fn flush_nodes(&self) -> Result<(), FileIoError> {
+        use std::pin::Pin;
+
+        #[derive(Clone, Debug)]
+        struct PinnedBufferEntry {
+            pinned_buffer: Pin<Box<[u8]>>,
+            offset: Option<u64>,
+        }
+
         const RINGSIZE: usize = FileBacked::RINGSIZE as usize;
 
         let flush_start = Instant::now();
 
         let mut ring = self.storage.ring.lock().expect("poisoned lock");
-        let mut saved_pinned_buffers = vec![(false, std::pin::Pin::new(Box::default())); RINGSIZE];
+        let mut saved_pinned_buffers = vec![
+            PinnedBufferEntry {
+                pinned_buffer: Pin::new(Box::new([0; 0])),
+                offset: None,
+            };
+            RINGSIZE
+        ];
         for (&addr, &(area_size_index, ref node)) in &self.kind.new {
             let mut serialized = Vec::with_capacity(100); // TODO: better size? we can guess branches are larger
             node.as_bytes(area_size_index, &mut serialized);
             let mut serialized = serialized.into_boxed_slice();
             loop {
                 // Find the first available write buffer, enumerate to get the position for marking it completed
-                if let Some((pos, (busy, found))) = saved_pinned_buffers
+                if let Some((pos, pbe)) = saved_pinned_buffers
                     .iter_mut()
                     .enumerate()
-                    .find(|(_, (busy, _))| !*busy)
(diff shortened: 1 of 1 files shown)


rkuris:
Fixes #999

rkuris:
This wasn't a problem on mac, since it doesn't use io-uring:
```sh
mkdir ~/mnt
sudo mount_tmpfs -s 81920 ~/mnt
sudo chown $USER ~/mnt
cargo run --bin fwdctl --features logger -- create ~/mnt/test.db
target/debug/fwdctl insert --db ~/mnt/test.db a "$(tr -d '\n' < /etc/passwd)"
target/debug/fwdctl insert --db ~/mnt/test.db b "$(tr -d '\n' < /etc/passwd)"
target/debug/fwdctl insert --db ~/mnt/test.db c "$(tr -d '\n' < /etc/passwd)"
target/debug/fwdctl insert --db ~/mnt/test.db d "$(tr -d '\n' < /etc/passwd)"
```
results in:
```
Error: FileIO(FileIoError { inner: Os { code: 28, kind: StorageFull, message: "No space left on device" }, filename: Some("/Users/rkuris/mnt/test.db"), offset: 68064, context: Some("write") })
```

storage/src/nodestore.rs
```diff
@@ -1318,35 +1318,85 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
+            RINGSIZE
+        ];
         for (&addr, &(area_size_index, ref node)) in &self.kind.new {
             let mut serialized = Vec::with_capacity(100); // TODO: better size? we can guess branches are larger
             node.as_bytes(area_size_index, &mut serialized);
             let mut serialized = serialized.into_boxed_slice();
             loop {
                 // Find the first available write buffer, enumerate to get the position for marking it completed
-                if let Some((pos, (busy, found))) = saved_pinned_buffers
+                if let Some((pos, pbe)) = saved_pinned_buffers
```
alarso16, line 1382: Rust question: do you need two sets of parentheses to unwrap the tuple?

demosdemon: This is because `Some` only holds one value but the value is a tuple of two items. The inner parens destructures the tuple whereas the outer `Some(_)` is a whole unit also being destructured.

storage/src/nodestore.rs
```diff
@@ -1318,35 +1318,85 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
+                let pbe = saved_pinned_buffers
+                    .get_mut(item)
+                    .expect("should be an index into the array");
+
+                if entry.result()
+                    != pbe
+                        .pinned_buffer
+                        .len()
+                        .try_into()
+                        .expect("buffer should be small enough")
```
demosdemon, line 1346: this is a bit annoying. I'm going to suppress this lint. usize -> u64 and u64 -> usize does not truncate on any supported platform.

demosdemon (approved):

PR #1010, opened 2025-07-01
perf: Allow cloned proposals

rkuris:
We were using the Arc as a way to know if a proposal could be committed. Instead, we now allow proposals to be cloned, and just check an AtomicBool to verify whether we can commit the proposal or not.

This will allow for some additional optimizations in the ffi layer.


alarso16 (approved):
It seems reasonable to me

demosdemon (approved):

PR #1012, opened 2025-07-01
chore: suppress clippy::cast_possible_truncation across the workspace

demosdemon:
As noted in the comment in `firewood/src/lib.rs`, this is to avoid the unecessary try_into calls for `u64 -> usize`.


demosdemon:
https://github.com/rust-lang/rust-clippy/issues/15197 for the hopeful case of splitting the lint out

PR #1013, opened 2025-07-02
ci: Check for metrics changes

rkuris:
If someone changes metrics, and doesn't change the dashboard, let's emit a warning and add a comment to the PR to ensure we keep the dashboard updated.


.github/workflows/metrics-check.yaml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          script: |
+            const { execSync } = require('child_process');
+            
+            try {
+              // Get the base commit for comparison
+              const baseCommit = context.payload.pull_request.base.sha;
+              
+              // Check for metric changes
+              const metricChanges = execSync(`git diff ${baseCommit}..HEAD --unified=0 | grep -E "(metric!|#\\[metrics\\])" || true`, { encoding: 'utf8' });
```
demosdemon: doesn't match the expression above. `counter!` instead of `metric!` but should probably also check for `histogram!` since we are about to add one.

this should try to only search for `.rs` files as you triggered on yourself.

--- 
Suggestion: Might also be fun to try to use `tree-sitter` for this (e.g., https://github.com/BrianHicks/tree-grepper) but obviously grep's just fine.

rkuris:

Seems like we might get spammed since this runs on every commit, but since someone might decide to add metrics later, I think this is a necessary evil.

Another option is to add some keyword in the PR title to suppress this check, but I think that's too much magic.

demosdemon (approved):

PR #1015, opened 2025-07-02
feat(nodestore): add functionalities to iterate the free list

qusuyan:


storage/src/nodestore.rs
```diff
@@ -1677,7 +1756,22 @@ pub(crate) mod nodestore_test_utils {
         AREA_SIZES[area_size_index as usize]
     }
 
-    pub(crate) fn write_header<S: WritableStorage>(
+    pub(crate) fn test_write_free_area<S: WritableStorage>(
+        nodestore: &NodeStore<Committed, S>,
```
rkuris, line 1760: nit: probably not strictly necessary, since it's just an Arc. If you had a pedantic warning, I think those are suppressed now.

PR #1016, opened 2025-07-02
refactor(ffi)!: split starting metrics exporter from db startup

### LINKED ISSUE #984: [ffi] split registering metrics exporter and opening db into 2 methods
See also #955

RodrigoVillar:
This PR closes #984 by splitting startup of the metrics exporter into its own ffi method (`fwd_start_metrics`). 

## What

I removed the metrics logic from `common_create` and put it into `fwd_start_metrics`. I also removed the metrics field from `CreateOrOpenArgs`.

Given that starting the metrics exporter is now a separate function call, I also removed the special case of not starting the metrics exporter for when `metrics_port = 0`.

## Testing

Added a unit test for starting up the metrics server and querying it.


ffi/firewood_test.go
```diff
@@ -959,3 +962,37 @@ func TestGetFromRoot(t *testing.T) {
 	_, err = db.GetFromRoot(nonExistentRoot, []byte("key"))
 	r.Error(err, "GetFromRoot with non-existent root should return error")
 }
+
+func TestStartMetrics(t *testing.T) {
```
alarso16, line 966: Any chance you could make a test that you can't open a different metrics port later? You shouldn't even have to instantiate database. This might not be super trivial - if you can't, that's ok

ffi/src/lib.rs
```diff
@@ -823,6 +823,18 @@ pub unsafe extern "C" fn fwd_free_database_error_result(
     // Note: we don't free the db pointer as it's managed by the caller
 }
 
+/// Start metrics exporter for this process
+///
+/// # Arguments
```
alarso16, line 828: You should add the qualitative return type (which I believe is the same as `fwd_commit`, if you wanted to copy and paste)

ffi/firewood.go
```diff
@@ -126,6 +123,13 @@ func New(filePath string, conf *Config) (*Database, error) {
 	return &Database{handle: db}, nil
 }
 
+// Starts metrics exporter for this process.
+// Returns an error if the metrics exporter was unable to start.
```
```suggestion
// Returns an error if the metrics exporter was unable to start or already started.
// This function should only be called once per process.
```

alarso16 (changes requested):

ffi/firewood_test.go
RodrigoVillar: Discussed offline - adding this test is difficult since we would need to spawn a separate process. Ideally, we should have a `StopMetrics` function or something similar to terminate exporters spawned in different tests, but this is outside the scope of this PR.

ffi/src/lib.rs
RodrigoVillar: Done: 73daf6e

ffi/src/metrics_setup.rs
```diff
@@ -18,17 +18,9 @@ use metrics::Key;
 use metrics_util::registry::{AtomicStorage, Registry};
 
 /// Sets up a metrics server over a specified port.
-/// If `metrics_port` is set to 0, it will not initialize the metrics system.
 /// This happens on a per-process basis, meaning that the metrics system
 /// cannot be initialized if it has already been set up in the same process.
-/// Any caller should ensure that the metrics system is not already initialized
-/// by explicitly calling with `metrics_port` == 0..
 pub(crate) fn setup_metrics(metrics_port: u16) -> Result<(), Box<dyn Error>> {
```
rkuris, line 23: Interesting to note that 0 might now be a valid metrics port, which lets the OS decide which port to use. It would be a super cool enhancement to have this return `Result<u16, Box<dyn Error>>` with the configured metrics port in it.

rkuris (approved):
This is a breaking change and therefore we should be in the practice of annotating it. One way to do that is to change the title of the PR to "refactor!(ffi)". Other ways are documented here:

https://www.conventionalcommits.org/en/v1.0.0/

I realize that coreth is the only consumer at the moment, so approving anyway.

demosdemon (approved):

PR #1017, opened 2025-07-02
feat: enable a configurable free list cache in the FFI

alarso16:
This allows a configurable free list cache for all consumers of the FFI

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -827,12 +827,12 @@ pub unsafe extern "C" fn fwd_free_database_error_result(
 ///
 /// * `path` - The path to the database file, which will be truncated if passed to `fwd_create_db()`
 ///   otherwise should exist if passed to `fwd_open_db()`.
-/// * `cache_size` - The size of the node cache, panics if <= 0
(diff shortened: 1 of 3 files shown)


ffi/src/lib.rs
```diff
@@ -828,11 +828,17 @@ pub unsafe extern "C" fn fwd_free_database_error_result(
 /// * `path` - The path to the database file, which will be truncated if passed to `fwd_create_db()`
 ///   otherwise should exist if passed to `fwd_open_db()`.
 /// * `cache_size` - The size of the node cache, panics if <= 0
-/// * `revisions` - The maximum number of revisions to keep; firewood currently requires this to be at least 2
+/// * `free_list_cache_size` - The size of the free list cache, panics if <= 0
```
demosdemon: nit: I don't see a panic, just an error return

demosdemon (approved):

ffi/src/lib.rs
alarso16: True, we just never updated the other comments. I fixed them while I was near

PR #1021, opened 2025-07-02
fix: logger macros causing linting warnings

demosdemon:
Previously, the logger macros were empty when logging was disabled. This caused no code to be generated which resulted in warnings from clippy around dead code.

This change bypasses those warnings by generating code that lexically uses the parameters passed in however at runtime will never be executed and will be elided by the compiler when optimizations are enabled and logging is disabled.

### DIFF
--- a/storage/src/logger.rs
+++ b/storage/src/logger.rs
@@ -23,7 +23,14 @@ mod noop_logger {
     #[macro_export]
     /// A noop logger, when the logger feature is disabled
     macro_rules! noop {
-        ($($arg:tt)+) => {};
+        ($($arg:tt)+) => {
+            if $crate::logger::trace_enabled() {
+                // This is a no-op. `trace_enabled` is always false. However, this
+                // branch exists to satisfy lints that complain about unused variables
+                // passed into the macro.
+                let _ = format!($($arg)+);
+            }
+        };
     }
 
     pub use noop as debug;
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -1,13 +1,6 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-#![cfg_attr(
-    feature = "ethhash",
-    expect(
-        clippy::used_underscore_binding,
-        reason = "Found 1 occurrences after enabling the lint."
-    )
-)]
 #![expect(
     clippy::cast_precision_loss,
     reason = "Found 2 occurrences after enabling the lint."
@@ -24,7 +17,7 @@ use std::sync::Arc;
 #[cfg(feature = "ethhash")]
 use std::sync::OnceLock;
 
-use firewood_storage::logger::{trace, trace_enabled, warn};
+use firewood_storage::logger::{trace, warn};
 use metrics::gauge;
 use typed_builder::TypedBuilder;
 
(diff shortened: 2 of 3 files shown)


storage/src/logger.rs
```diff
@@ -23,7 +23,14 @@ mod noop_logger {
     #[macro_export]
     /// A noop logger, when the logger feature is disabled
     macro_rules! noop {
-        ($($arg:tt)+) => {};
+        ($($arg:tt)+) => {
+            if $crate::logger::trace_enabled() {
+                // This is a no-op. `trace_enabled` is always false. However, this
+                // branch exists to satisfy lints that complain about unused variables
+                // passed into the macro.
+                let _ = format!($($arg)+);
```
rkuris, line 37: Does this get completely optimized out? I'm concerned because if you say:
```
let _ = format!("update encoded value {:02X?}", hex::encode(&bytes))
```
that might call hex::encode, which we don't want. In the extreme case, imagine if it called a function that has a mutable reference; the compiler then can't be sure the function has no side effects and therefore must call the function, even when tracing is disabled.

demosdemon: Yes, this gets completely optimized out. It's important to note that macros work lexicographically; they expand tokens, not expressions.

First, inspect the generated code using `cargo-expand`:

```rust
        if crate::logger::trace_enabled() {
            if self.key().size_hint().0 == 64 {
                trace!("SIZE WAS 64 {}", hex::encode(&collector));
            } else {
                trace!(
                    "SIZE WAS {1} {0}",
                    hex::encode(&collector),
                    self.key().size_hint().0
                );
            }
        }

```

Is expanded into:

```rust
                if crate::logger::trace_enabled() {
                    if self.key().size_hint().0 == 64 {
                        if crate::logger::trace_enabled() {
                            let _ = ::alloc::__export::must_use({
                                ::alloc::fmt::format(
                                    format_args!("SIZE WAS 64 {0}", hex::encode(&collector)),
                                )
                            });
                        }
                    } else {
                        if crate::logger::trace_enabled() {
                            let _ = ::alloc::__export::must_use({
                                ::alloc::fmt::format(
                                    format_args!(
                                        "SIZE WAS {1} {0}",
                                        hex::encode(&collector),
                                        self.key().size_hint().0,
                                    ),
                                )
                            });
                        }
                    }
                }
```

The format expression is pushed down into the `crate::logger::trace_enabled()` branch. The expression cannot even evaluate unless that function returns true. Which, if `logger` is not enabled, is defined as:

```
    /// `trace_enabled` for a noop logger is always false
    #[inline]
    #[must_use]
    pub const fn trace_enabled() -> bool {
        false
    }
```

It is both `const` and `#[inline]` meaning the compiler will shake out the dead branch when optimizing.

rkuris: Ah yes. The trick here is the extra call to `trace_enabled` so now every logging macro expansion checks `trace_enabled` first. We call `trace_enabled` even for debug, error, etc, which is what confused me.

Why doesn't `noop` just expand to `if false {` instead of having to call `trace_enabled`?

demosdemon: > Why doesn't `noop` just expand to `if false {` instead of having to call `trace_enabled`?

That is a good question. We need to actually call `trace_enabled` because `clippy` _doesn't_ evaluate the const function. Therefore, it sees a function call and is not able to detect a dead branch. `if false { ... }` would not achieve the desired goal suppressing dead code warnings because clippy is smart enough to see the dead branch and ignore anything within it (and generate an additional lint for the trouble).

demosdemon: Turns out, this isn't true if done within macros, only outside. So, I can switch to `if false` and update the comment to be a little clearer.

rkuris (approved):

PR #1022, opened 2025-07-02
chore: remove `#[allow]`s no longer needed

demosdemon:
This cleans up a lot of the remaining `#[allow]` attributes in the code base. Only one remains, `firewood-macros/tests/compile_pass/with_attributes.rs` because I did not want to alter the tests of the macro.


firewood/src/merkle.rs
```diff
@@ -235,12 +242,12 @@ impl<T: TrieReader> Merkle<T> {
         PathIterator::new(&self.nodestore, key)
     }
 
-    #[allow(dead_code)]
+    #[cfg(test)]
```
rkuris, line 245: This one (and the next two) are about to get used in real code in an upcoming commit. We could change these to `expect` so they don't get forgotten.

firewood/src/merkle.rs
```diff
@@ -421,9 +428,7 @@ impl<T: HashedNodeReader> Merkle<T> {
                 .map_err(|e| FileIoError::new(e, None, 0, None))
                 .map_err(Error::other)?;
             let mut seen = HashSet::new();
-            // If ethhash is off, root_hash.into() is already the correct type
-            // so we disable the warning here
-            #[allow(clippy::useless_conversion)]
+            #[cfg_attr(not(feature = "ethhash"), expect(clippy::useless_conversion))]
```
rkuris: let's add the comment back, since it's not obvious why it's only needed for ethhash without it.

firewood/src/merkle.rs
```diff
@@ -2095,7 +2099,6 @@ mod tests {
     }
 
     // #[test]
```
rkuris, line 2105: These tests were left over since the rewrite and have never been re-enabled. They need evaluation to see if they add value (that is, they test something that we might not catch with one of the other tests) and enabled if they do, or deleted if they don't.

I doubt there is a ticket for this. If not, let's create one.

firewood/src/proof.rs
```diff
@@ -209,7 +209,7 @@ impl<T: Hashable> Proof<T> {
             return Err(ProofError::Empty);
         };
 
-        #[allow(clippy::useless_conversion)]
+        #[cfg_attr(not(feature = "ethhash"), expect(clippy::useless_conversion))]
```
rkuris: A better way to fix this is to conditionally compile this let statement to not even do the into() when ethhash is enabled.

rkuris (changes requested):
The tests in tests/common seem to be good integration tests, so they should be reactivated instead of deleted. The API changed somewhat after the rewrite so they may need some tweaking.

I'd suggest leaving these around for now.

firewood/src/proof.rs
demosdemon: Yeah, I did that but it was rather ugly. I think the new approach is a good middle ground.

PR #1026, opened 2025-07-03
feat(checker): traverse free lists

qusuyan:


storage/src/checker.rs
```diff
@@ -60,31 +64,54 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
 
         Ok(())
     }
+
+    /// Traverse all the free areas in the freelist
+    fn visit_freelist(&self, visited: &mut LinearAddressRangeSet) -> Result<(), CheckerError> {
+        for free_area in self.free_list_iter_inner(0) {
+            let (addr, area_index, free_list_id) = free_area?;
+            let free_list_id =
+                AreaIndex::try_from(free_list_id).expect("area index will not exceed u8");
```
rkuris: This conversion might be better do to in the iterator. Any consumer of this will want the AreaIndex, not a "free_list_id".

storage/src/checker.rs
```diff
@@ -60,31 +64,54 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
             }
         }
 
         Ok(())
     }
+
+    /// Traverse all the free areas in the freelist
+    fn visit_freelist(&self, visited: &mut LinearAddressRangeSet) -> Result<(), CheckerError> {
+        for free_area in self.free_list_iter_inner(0) {
+            let (addr, area_index, free_list_id) = free_area?;
```
rkuris: naming: I can't tell by looking at this which of these is on-disk and which is the one that indiciates the freelist.

storage/src/lib.rs
```diff
@@ -129,6 +129,21 @@ pub enum CheckerError {
...
+    #[error(
+        "Free area {address} of size {size} is found in free list {free_list} but it should be in freelist {expected_free_list}"
+    )]
+    FreelistAreaSizeMismatch {
+        /// Address of the free area
+        address: LinearAddress,
+        /// Actual size of the free area
+        size: u64,
+        /// Free list on which the area is stored
+        free_list: u8,
```
rkuris: nit: actual_free_list

rkuris (approved):
Approved with some optional renames.

storage/src/checker.rs
qusuyan: I guess not too important but only the checker will use free_list_id and it will be ignored in most cases. I am keeping it here just to avoid this conversion with `try_from` in most cases

PR #1027, opened 2025-07-03
perf: Break up the RevisionManager lock

rkuris:
Historical and by_hash members might get some benefit from RwLock, but proposals are small enough and locked less frequently so a mutex is a better choice.

Fearless concurrency.


firewood/src/manager.rs
```diff
@@ -277,6 +309,8 @@ impl RevisionManager {
 
     pub fn current_revision(&self) -> CommittedRevision {
         self.historical
+            .read()
+            .expect("poisoned lock")
```
demosdemon, line 313: If we ever want to get rid of these panics for poisoned locks, we can switch to [parking_lot](https://docs.rs/parking_lot) which doesn't do poisoning. Poisoning only happens when a thread panics while a lock is acquired. If we are strict about not panicking, then we don't need the poisoning feature and its extra overhead.

rkuris: Created #1030

PR #1029, opened 2025-07-03
perf(metrics): Metrics were allocating memory

rkuris:
This is a quick workaround to avoid allocating memory for these counters.

### DIFF
--- a/storage/src/linear/filebacked.rs
+++ b/storage/src/linear/filebacked.rs
@@ -31,14 +31,28 @@ use std::os::unix::fs::FileExt;
 #[cfg(windows)]
 use std::os::windows::fs::FileExt;
 use std::path::PathBuf;
-use std::sync::Mutex;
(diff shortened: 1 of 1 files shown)


demosdemon (commented):
really, how much memory are we allocating here? Most recorders should be returning a similarly cached object.

E.g.,

https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-prometheus/src/recorder.rs#L293-L303

Which uses maps to monotonize these things:

https://github.com/metrics-rs/metrics/blob/main/metrics-util/src/registry/mod.rs#L49-L58

we are similarly do that too:

https://github.com/ava-labs/firewood/blob/main/ffi/src/metrics_setup.rs#L156-L184

rkuris:
Actually this isn't true. We're using this expansion: https://docs.rs/metrics/latest/src/metrics/macros.rs.html#35 which doesn't actually allocate anything.

PR #1033, opened 2025-07-03
feat(checker): identify leaked areas

qusuyan:


storage/src/checker.rs
```diff
@@ -46,7 +51,18 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
         // 4. check missed areas - what are the spaces between trie nodes and free lists we have traversed?
-        let _ = visited.complement(); // TODO
+        let leaked_areas = visited
+            .complement()
+            .into_iter()
+            .map(|range| self.find_leaked_areas(range))
+            .collect::<Result<Vec<_>, CheckerError>>()?
+            .into_iter()
+            .flatten()
+            .collect::<Vec<_>>();
```
rkuris: Looks more like try_fold than map/collect/errorcheck/iterate/flatten, something like:
```suggestion
           let leaked_areas  = visited
                .complement()
                .into_iter()
                .try_fold(Vec::new(), |mut acc, range| {
                    acc.extend(self.find_leaked_areas(range)?);
                    Ok::<_, CheckerError>(acc)
                })?;
```

storage/src/checker.rs
```diff
@@ -87,17 +103,64 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        loop {
+            let (area_size_index, area_size) = self.area_index_and_size(current_addr)?;
+            leaked.push((current_addr, area_size_index));
+
+            // advance to the next area
+            let Some(next_addr) = current_addr.checked_add(area_size) else {
+                return Err(CheckerError::AreaOutOfBounds {
+                    start: current_addr,
+                    size: area_size,
+                    bounds: LinearAddress::new(NodeStoreHeader::SIZE)
```
rkuris: Nonzero macros would make this easier to read

storage/src/checker.rs
```diff
@@ -87,17 +103,64 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+                        .expect("this constant is non-zero")
+                        ..LinearAddress::new(self.size())
+                            .expect("size is not zero due to previous check"),
```
rkuris: what previous check?

storage/src/checker.rs
```diff
@@ -87,17 +103,64 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         }
         Ok(())
     }
+
+    fn find_leaked_areas(
```
rkuris: this needs a block comment

storage/src/checker.rs
```diff
@@ -87,17 +103,64 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        leaked_range: Range<LinearAddress>,
+    ) -> Result<Vec<(LinearAddress, AreaIndex)>, CheckerError> {
+        let mut leaked = Vec::new();
+        let mut current_addr = leaked_range.start;
+        loop {
+            let (area_size_index, area_size) = self.area_index_and_size(current_addr)?;
+            leaked.push((current_addr, area_size_index));
+
+            // advance to the next area
+            let Some(next_addr) = current_addr.checked_add(area_size) else {
```
rkuris: nit: I prefer ok_or_else here, to focus on the computation instead of the error:
```
     let next_addr = current_addr.checked_add(area_size).ok_or_else(||
        Err(CheckerError::AreaOutOfBounds {
                    start: current_addr,
                    size: area_size,
                    bounds: LinearAddress::new(NodeStoreHeader::SIZE))?;

File: storage/src/checker.rs [hunk 6]
```diff
@@ -208,4 +271,68 @@ mod test {
...
+
+        // check the leaked areas
+        let leaked_areas = leaked
+            .into_iter()
+            .map(|range| nodestore.find_leaked_areas(range))
+            .collect::<Result<Vec<_>, CheckerError>>()
+            .unwrap()
+            .into_iter()
+            .flatten()
+            .collect::<HashMap<_, _>>();
```
rkuris: This looks really familiar. Maybe break this iterator out into a helper?

I realize this is in a test but you can call the non-test code from there.

storage/src/lib.rs
```diff
@@ -146,6 +146,13 @@ pub enum CheckerError {
         expected_free_list: AreaIndex,
     },
 
+    /// Found leaked areas
+    #[error("Found leaked areas: {leaked:?}")]
```
rkuris: This might be a massive string. Suggest using `leaked.get(0..SOME_SANE_MAXIMUM)` to avoid that

rkuris (changes requested):

storage/src/checker.rs
qusuyan: I meant we checked the size when `LinearAddressRangeSet` is created, but seems we might need to restructure this code better...

storage/src/lib.rs
```diff
@@ -94,6 +94,29 @@ pub fn empty_trie_hash() -> TrieHash {
...
+            .take(MAX_LEAKED_AREAS_PRINT)
+            .map(|(addr, size)| format!("{{start: {addr}, size: {size}}}"))
+            .collect::<Vec<_>>()
+            .join(", ");
+        if let Some(hidden_count) = self.0.len().checked_sub(MAX_LEAKED_AREAS_PRINT) {
+            write!(f, "[{content}, ... ({hidden_count} more hidden)]")?;
+        } else {
+            write!(f, "[{content}]")?;
+        }
+        Ok(())
```
demosdemon: this unnecessarily allocates intermediate string and a vector
```suggestion
        let mut content = self.0.iter();
        let Some((addr, size)) = content.next() else {
            // unexpected, but if there are no leaked areas then we print an empty list
            return write!(f, "[]");
        };

        // write the first element
        write!(f, "[{{start: {addr}, size: {size}}}")?;
        let mut count = MAX_LEAKED_AREAS_PRINT - 1;
        for (addr, size) in content {
            if count == 0 {
                return write!(
                    f,
                    ", ... ({}) more hidden)]",
                    self.0.len().saturating_sub(MAX_LEAKED_AREAS_PRINT)
                );
            }

            // write the next element with preceding comma
            write!(f, ", {{start: {addr}, size: {size}}}")?;

            #[expect(
                clippy::arithmetic_side_effects,
                reason = "count is always greater than zero"
            )]
            {
                count -= 1;
            }
        }

        // less than (or equal to) MAX_LEAKED_AREAS_PRINT elements, so we can close the list
        write!(f, "]")
```

storage/src/checker.rs
rkuris: Discussed offline. Use the size that has already been converted.

storage/src/iter.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
qusuyan, line 1: This file is included by @demosdemon to implement `Debug` for `LeakedAreas` without allocating new strings

demosdemon (approved):

qusuyan:
TODO: It is possible that the area size stored on disk is corrupted and it does not correspond to any valid area size.

storage/src/checker.rs
```diff
@@ -46,7 +51,13 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         self.visit_freelist(&mut visited)?;
 
         // 4. check missed areas - what are the spaces between trie nodes and free lists we have traversed?
-        let _ = visited.complement(); // TODO
+        let leaked_areas = self.find_all_leaked_areas(visited.complement())?;
+
+        if !leaked_areas.is_empty() {
+            // warning here since there are multiple revisions on disk but only the latest one is traversed
+            // TODO: change it to an error once we have a way to traverse all the revisions
+            warn!("Found leaked areas: {:?}", LeakedAreas(leaked_areas));
```
rkuris, line 62: If there are 5000 leaked areas (this is not unreasonable) this print line will be incredibly long. It's possible for there are 100k of them.

storage/src/checker.rs
```diff
@@ -87,17 +98,93 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+    /// We assume that all space within `leaked_range` are leaked and the stored areas are contiguous.
+    /// Returns error if the last leaked area extends beyond the end of the range.
+    fn find_leaked_areas(
+        &self,
+        leaked_range: Range<LinearAddress>,
+    ) -> Result<Vec<(LinearAddress, AreaIndex)>, CheckerError> {
+        let mut leaked = Vec::new();
+        let mut current_addr = leaked_range.start;
+        loop {
+            let (area_size_index, area_size) = self.area_index_and_size(current_addr)?;
```
rkuris: This is too strict. Let's assume the freed area contains all zeros and 0xff as examples. For the zero case, you'll be iterating over 16 byte chunks for a very long time. For 0xff, you'll be erroring out and not returning any values.

storage/src/checker.rs
```diff
@@ -87,17 +98,93 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+                let leaked_areas = self.find_leaked_areas(range)?;
+                leaked.extend(leaked_areas);
+                Ok::<_, CheckerError>(leaked)
+            })
+    }
+
+    /// Find all stored areas within `leaked_range`.
+    /// We assume that all space within `leaked_range` are leaked and the stored areas are contiguous.
+    /// Returns error if the last leaked area extends beyond the end of the range.
+    fn find_leaked_areas(
```
rkuris: naming: leaked_range_to_areas

storage/src/checker.rs
```diff
@@ -87,17 +98,93 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        let mut current_addr = leaked_range.start;
+        loop {
+            let (area_size_index, area_size) = self.area_index_and_size(current_addr)?;
+            leaked.push((current_addr, area_size_index));
+
+            // advance to the next area
+            let next_addr =
+                current_addr
+                    .checked_add(area_size)
+                    .ok_or(CheckerError::LeakedAreaTooLarge {
```
rkuris: This should be impossible so use expect

qusuyan:
Splitting this diff into multiple parts

PR #1035, opened 2025-07-05
feat(ffi): Add gauges to metrics reporter

rkuris:
Gauges weren't being reported via the homegrown HTTP endpoint we used for ffi, now they are. This is particularly important since we now have a few useful gauges in firewood.

The surrounding HTTP structure might be going away, but the underlying `stats` call is likely to stay, so this enhancement is probably still useful.

Also took the opportunity to reduce memory allocations here, and write directly to strings instead of converting back and forth between string/bytes/string with all the utf8 conversions.

Histograms are still not supported, but we don't have any, and aren't likely to add any anytime soon.

### DIFF
--- a/ffi/src/metrics_setup.rs
+++ b/ffi/src/metrics_setup.rs
@@ -1,6 +1,7 @@
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::error::Error;
-use std::io::Write;
+use std::fmt::Write;
 use std::net::Ipv6Addr;
 use std::ops::Deref;
 use std::sync::Arc;
@@ -63,7 +64,7 @@ struct TextRecorder {
 
 impl TextRecorder {
     fn stats(&self) -> String {
-        let mut output = Vec::new();
+        let mut output = String::new();
         let systemtime_now = SystemTime::now();
         let utc_now: DateTime<Utc> = systemtime_now.into();
         let epoch_duration = systemtime_now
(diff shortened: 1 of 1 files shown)


ffi/src/metrics_setup.rs
```diff
@@ -63,7 +66,7 @@ struct TextRecorder {
 
 impl TextRecorder {
     fn stats(&self) -> String {
-        let mut output = Vec::new();
+        let mut output = String::new();
```
demosdemon, line 69: I would go a step further and use thread-local storage to reuse the same buffer every time because this is going to be called a lot and will generally output the same length of data. Reusing the string would amortize the growing penalty.

ffi/src/metrics_setup.rs
```diff
@@ -75,40 +78,82 @@ impl TextRecorder {
             .saturating_add(u64::from(epoch_duration.subsec_millis()));
         writeln!(output, "# {utc_now}").unwrap();
 
+        let help_guard = self.inner.help.lock().expect("poisoned lock");
+
         let counters = self.registry.get_counter_handles();
-        let mut seen = HashSet::new();
+        let mut seen_counters = HashSet::new();
```
demosdemon, line 84: Same for the hashset, it can be in TLS to re-use the allocations it makes.

ffi/src/metrics_setup.rs
```diff
@@ -75,40 +78,82 @@ impl TextRecorder {
...
+            self.write_labels(&mut output, key.labels());
+
+            writeln!(output, " {} {}", counter.load(Ordering::Relaxed), epoch_ms)
                 .expect("write error");
+        }
+
+        // Get gauge handles: Uses self.registry.get_gauge_handles() to retrieve all registered gauges
+        let gauges = self.registry.get_gauge_handles();
+        // Track seen gauges: Uses a separate HashSet to avoid duplicate # TYPE declarations
+        let mut seen_gauges = HashSet::new();
```
demosdemon, line 104: this could reuse the previous hashset if it were cleared first

demosdemon: The comment "Uses a separate HashSet to avoid duplicate # TYPE declarations" doesn't make sense because you do not check `seen_counters`.

Re-using the same hashset without clearing it would dedupe counter and gauge descriptions.

ffi/src/metrics_setup.rs
rkuris: Not sure this is much of an improvement:
 - The total allocated size is pretty small and can vary between runs
 - We only call this method every few seconds (5 seconds I think)
 - I already think I overoptimized this, and there's nothing to indicate that this is a performance drain.

So while yes we can do better, I think this is better than it was. Making it TLS or reusing the memory could be a future low priority task.

ffi/src/metrics_setup.rs
rkuris: Prometheus does not support gauges and counters with the same name. It might be a good idea to check the `seen_counters` but I don't know what we should do. Silently dropping it is worse, and a runtime panic is also terrible. Suggestions welcome.

But yeah the comment is wrong. I'll fix that.

rkuris: I went with "panic in debug mode and do nothing in release mode".

PR #1040, opened 2025-07-07
chore: update firewood in grpc-testtool

demosdemon:
I noticed in this action output an error from the incorrect firewood version:

https://github.com/ava-labs/firewood/actions/runs/16127166800/job/45506937221#step:7:13


rkuris:
Cult of the Dead Code says "delete grpc-testtool". We kept this around because it might be handy reference for x/sync, but otherwise is unused.

rkuris (approved):

PR #1041, opened 2025-07-07
feat(delayed-persist): Part 1: Roots may be in mem

rkuris:
Introduces a new type, `MaybePersistedNode`, which is a node which is either a SharedNode or a LinearAddress. This new type will allow writes to occur while readers are still reading this revision using an `ArcSwap`.

Note that this does not actually put any additional nodes in memory, it just sets things up so they could be in memory.

Additional changes needed for this feature to be complete are:

 - [ ] BranchNode children can now be Node, LinearAddress or a MaybePersistedNode
 - [ ] When converting a `MutableProposal` to an `ImmutableProposal`, don't actually allocate space, just create `SharedNode`s from the `Node`s.
 - [ ] Remove `NodeStoreHeader` from `NodeStore`. The header should move into the `RevisionManager`.
 - [ ] Allocate new nodes from the recently-freed nodes from an expired revision, then from the freelist. This can be done by maintaining a local freelist from the nodes deleted in the expiring revision and falling back to the actual freelist. Deleted nodes that are not reused must still be pushed onto the freelist.


storage/src/node/memdisk.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// This method changes the internal state of the `MemDiskNode` from `Mem` to `Disk`,
+    /// effectively moving the node from memory to the specified disk location.
+    ///
+    /// This is done atomically using the `ArcSwap` mechanism.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `LinearAddress` where the node should be stored on disk
+    pub fn allocate(&self, addr: LinearAddress) {
```
rkuris: `persist` or `persist_at` is a better name I think, so code would be `node.persist_at(addr)` which seems more readable than `node.allocate(addr)`, especially since the allocation at addr happens elsewhere.

rkuris:
The goal of this set of diffs is to separate persistence from both proposals and committed revisions. When proposing, we currently allocate space and modify the freelist. When committing, we currently flush all the modified nodes to disk.

This diff is step 1. The only thing this diff does is allow for the root node (and only the root node) of proposals and revisions to be kept in memory and, using a lockless type, allow them to be persisted even while being read by other threads. It doesn't do that, since they are currently always created as an allocated node, but that will change on future diffs.

In other words, it still always marks the node as persisted by assigning the address at propose time and using the persisted variant of the new `MemDiskNode` type when created, doesn't allow for branches to have their changes in memory, and doesn't defer the allocation to commit time. So this means this diff doesn't actually make any improvements. It might have a performance impact and it might be good to measure it at this point though.

storage/src/nodestore.rs
```diff
@@ -1569,6 +1578,7 @@ impl<S: ReadableStorage> TryFrom<NodeStore<MutableProposal, S>>
             deleted: immutable_proposal.deleted,
             parent: immutable_proposal.parent,
             root_hash: Some(root_hash.into_triehash()),
+            root: Some(root_addr.into()),
```
qusuyan, line 1581: I wonder if this works...
```suggestion
            root
```
Then you will directly have the root node in memory and don't need to fetch it from the address on next access

storage/src/node/memdisk.rs
```diff
@@ -0,0 +1,75 @@
...
+use arc_swap::ArcSwap;
+
+use crate::{FileIoError, LinearAddress, NodeReader, SharedNode};
+
+/// A node that is either in memory or on disk.
+///
+/// In-memory nodes can be moved to disk. This structure allows that to happen
+/// atomically.
+#[derive(Debug, Clone)]
+pub struct MemDiskNode(Arc<ArcSwap<MemDiskEnum>>);
```
qusuyan: `MemDisk` sounds like you are using ramdisk for storage. Maybe rename to `SharedNodeRef`?

storage/src/node/memdisk.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// This method changes the internal state of the `MemDiskNode` from `Mem` to `Disk`,
+    /// indicating that the node has been written to the specified disk location.
+    ///
+    /// This is done atomically using the `ArcSwap` mechanism.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The `LinearAddress` where the node has been persisted on disk
+    pub fn persist_at(&self, addr: LinearAddress) {
```
qusuyan: Sounds like this function is really called when the node is evicted from cache instead of when it gets persisted. Maybe rename?

rkuris: No, it's called after the. node is written to disk.

storage/src/node/memdisk.rs
rkuris: I don't like `MemDiskNode` either, but `SharedNodeRef` is even more obscure.

I'll try `MaybePersisted`.

storage/src/node/memdisk.rs
qusuyan: Oh because after you persist it you can read it from the cache of linear storage using its address right? Then I think `MaybeAllocated` may be a better name than `MemDiskNode`.

storage/src/nodestore.rs
rkuris: It doesn't. The only thing we have here is the address of the root, not the root node.

qusuyan: Don't we have it from this line? https://github.com/ava-labs/firewood/pull/1041/files/65b9200e82e1c3b64d696c83f4efa198ca4b93c3#diff-5c0709f9ddf7f6e5050a693caf97e718e19a9ddd86ac438a3c3f764060353e4bR1558

qusuyan: Never mind I see it gets consumed by `hash_helper`

qusuyan: I think this is good for now but maybe in a future diff `hash_helper` will only compute the hash and not allocate the space and return an address :)

storage/src/node/memdisk.rs
rkuris: It's not just allocated, it's persisted. The states here are:
 - Node in memory (SharedNode)
 - Node on disk (LinearAddress)
 
 Allocation is the verb that describes part of persisting a node. First you allocate space, then you write (persist) it. If it has a LinearAddress, and I'm done with all my changes, the only thing that will have LinearAddresses are persisted. Right now we get a LinearAddress at allocation time, which is at propose time, but that's all moving.

qusuyan (approved):

storage/src/node/persist.rs
```diff
@@ -0,0 +1,79 @@
...
+use arc_swap::ArcSwap;
+
+use crate::{FileIoError, LinearAddress, NodeReader, SharedNode};
+
+/// A node that is either in memory or on disk.
+///
+/// In-memory nodes can be moved to disk. This structure allows that to happen
+/// atomically.
+#[derive(Debug, Clone)]
+pub struct MaybePersistedNode(Arc<ArcSwap<MaybePersisted>>);
```
demosdemon: Might be good to expand the doc here to explain the layout:

```suggestion
/// A node that is either in memory or on disk.
///
/// In-memory nodes can be moved to disk. This structure allows that to happen
/// atomically.
///
/// `MaybePersistedNode` owns a reference counted pointer to an atomically swapable
/// pointer. The atomically swapable pointer points to a reference counted pointer
/// to the enum of either an un-persisted but committed (or committing) node or the
/// linear address of a persisted node.
///
/// ```text
/// |--------------------|--------------------------------------------------------|
/// | MaybePersistedNode | Newtype wrapper around the Arc<ArcSwap<_>> allowing    |
/// |                    | it to be shared and updated without breaking existing  |
/// |                    | references.                                            |
/// |--------------------|--------------------------------------------------------|
/// | Arc                | ARC pointer to an updatable pointer.                   |
/// |--------------------|--------------------------------------------------------|
/// | ArcSwap            | Shared updatable pointer. Updates done by replacing    |
/// |                    | the pointer atomically.                                |
/// |--------------------|--------------------------------------------------------|
/// | Arc                | ARC pointer to the underlying node. Uses a separate    |
/// |                    | ARC to maintain the memory for the node                |
/// |                    | independently of the shared `MaybePersisted` node.     |
/// |                    | This is also a constraint of ArcSwap as only values    |
/// |                    | <=64bits can be atomically swapped on most platforms.  |
/// |--------------------|--------------------------------------------------------|
/// | MaybePersisted     | Enum wrapper of either kind of node.                   |
/// |--------------------|--------------------------------------------------------|
/// | Unpersisted or     | The shared node for unpersisted nodes or the linear    |
/// |    Persisted       | address of a persisted node.                           |
/// |--------------------|--------------------------------------------------------|
/// ```
///
/// Traversing these pointers does incur a penalty. To resolve the data, the
/// program must traverse 6 references:
///
/// - Reference to the top-level `MaybePersistedNode` owned by another node or
///   or proposal.
/// - Reference to Arc referenced by `MaybePersistedNode` (note: most of the time
///   the compiler will elide this specific dereference because we are using a
///   single element tuple struct, aka a newtype struct, therefore memory layouts
///   are transparent).
/// - Reference to the ArcSwap owned by the Arc
/// - Reference to the Arc owned by the ArcSwap -- this is also an atomic load
///   which incurs a separate small penalty
/// - Reference to the MaybePersisted
/// - Evaluation of the MaybePersisted enum discriminant
/// - And so on for the Unpersisted or Persisted variants
///
/// When an Unpersisted node is Persisted, a new Arc is created to the new
/// `MaybePersisted::Persisted` variant and the `ArcSwap` is updated atomically.
/// Subsequent accesses to the `MaybePersistedNode` will then see the `Persisted`
/// node.
///
/// Any thread that has a unique clone of the `Arc<ArcSwap<_>>` either via the
/// `MaybePersistedNode` or independently will see the change. However, any thread
/// that has a unique clone of `Arc<MaybePersisted>` _will not_ see the change.
#[derive(Debug, Clone)]
pub struct MaybePersistedNode(Arc<ArcSwap<MaybePersisted>>);
```

demosdemon (approved):

storage/src/node/persist.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+/// The internal state of a `MaybePersistedNode`.
+///
+/// This enum represents the two possible states of a `MaybePersisted`:
+/// - `Unpersisted(SharedNode)`: The node is currently in memory
+/// - `Persisted(LinearAddress)`: The node is currently on disk at the specified address
+#[derive(Debug)]
+pub enum MaybePersisted {
```
demosdemon: We can drop `pub` from this. Currently don't see a need for it.

storage/src/node/persist.rs
rkuris: Agreed. This is a fairly complicated type and someone might want to disect it later.

I made the explanations a little more brief, used the table markdown syntax instead of ascii tables, and added links to local types. Links to private types were left out.

PR #1044, opened 2025-07-08
feat!: export firewood metrics

RodrigoVillar:
This PR enables Firewood to export metrics so that clients such as Coreth can aggregate those metrics into its own.

## How
Splits the Firewood API into the following:

```go
// Start recorder
func StartMetrics()

// Start recorder and exporter
func StartMetricsWithExporter()

// Gather latest metrics
func GatherMetrics() (string, error)
```

`GatherMetrics()` is used by to create a custom Prometheus gatherer that can used by Coreth's Multigatherer.

## Testing

Extended the existing metrics test to account for `Gatherer`.


ffi/firewood.go
alarso16, line 1: imo this should go in a different file `metrics.go` or something, just for clarity that it's disjoint from initialization of the database.

ffi/firewood.go
```diff
@@ -129,14 +129,35 @@ func New(filePath string, conf *Config) (*Database, error) {
...
 
-// Starts metrics exporter for this process.
-// Returns an error if the metrics exporter was unable to start or already started.
-// This function should only be called once per process.
-func StartMetrics(metricsPort uint16) error {
-	result := C.fwd_start_metrics(C.uint16_t(metricsPort))
+// Starts global recorder for metrics.
+// Returns an error if the global recorder was already initialized.
+// This function should only be called once per process and cannot be called
+// alongside StartMetricsWithExporter.
```
rkuris: nit
```suggestion
// This function only needs to be called once.
// An error is returned if this method is called a second time, or if it is called after StartMetricsWithExporter.
```

ffi/firewood.go
```diff
@@ -129,14 +129,35 @@ func New(filePath string, conf *Config) (*Database, error) {
...
+// Returns an error if the global recorder was already initialized or if the
+// metrics exporter failed to start.
+// This function should only be called once per process and cannot be called
+// alongside StartMetrics.
+func StartMetricsWithExporter(metricsPort uint16) error {
+	result := C.fwd_start_metrics_with_exporter(C.uint16_t(metricsPort))
+	return errorFromValue(&result)
+}
+
+// Collect metrics from global recorder
```
rkuris:
```suggestion
// Collect metrics from global recorder
// This method must be called after StartMetrics
```
Is it true that this will still work even if there is a reporter?

ffi/firewood_test.go
```diff
@@ -995,4 +998,31 @@ func TestStartMetrics(t *testing.T) {
 
 	// Check that batch op was recorded
 	r.Contains(string(body), "firewood_ffi_batch 1")
+
+	g := Gatherer{}
+	metricsFamily, err := g.Gather()
+	r.NoError(err)
+
+	expectedMetrics := map[string]dto.MetricType{
```
rkuris: nifty 🏆

ffi/src/lib.rs
```diff
@@ -821,23 +821,50 @@ pub unsafe extern "C" fn fwd_free_database_error_result(
     // Note: we don't free the db pointer as it's managed by the caller
 }
 
-/// Start metrics exporter for this process
+/// Start metrics recorder for this process.
 ///
-/// # Arguments
+/// * `metrics_port` - the port where metrics will be exposed at
```
rkuris: remove this line

ffi/src/metrics_setup.rs
```diff
@@ -18,25 +19,36 @@ use chrono::{DateTime, Utc};
...
         inner: Arc::new(inner),
-    };
-    metrics::set_global_recorder(recorder.clone())?;
+    }
+});
+
+/// Starts metrics recorder.
+/// This happens on a per-process basis, meaning that the metrics system cannot
+/// be initialized if it has already been set up in the same process.
+pub(crate) fn setup_metrics() -> Result<(), Box<dyn Error>> {
```
rkuris: `pub(crate)` is not needed here since the module is not exported anyway in lib.rs. If `pub(crate)` is required, then `pub` should be sufficient, but you might be able to remove it altogether.

Same comment a few lines down. I don't know why the code was originally`pub(crate)`.

ffi/src/metrics_setup.rs
```diff
@@ -18,25 +19,36 @@ use chrono::{DateTime, Utc};
...
+    metrics::set_global_recorder(RECORDER.clone())?;
```
rkuris: Should this be done in recorder's initialization?

ffi/src/metrics_setup.rs
```diff
@@ -53,6 +65,11 @@ pub(crate) fn setup_metrics(metrics_port: u16) -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
+/// Returns the latest metrics for this process.
+pub(crate) fn gather_metrics() -> String {
+    RECORDER.stats()
```
rkuris: This has the wrong behavior if nobody called setup_metrics, because the global recorder has not been initialized.

Possible solutions:
(1) Remove the requirement to call setup_metrics. Allow someone to call setup_metrics_recorder as long as they do it before calling gather_metrics. If not, it is implicitly initialized through the LazyLock, which also registers the global recorder.
(2) Have gather_metrics return an error when RECORDER is not initialized. You can do this easily with OnceLock instead of LazyLock:
```
    let Some(recorder) = RECORDER.get() else { /* error */ }
    Ok(recorder.stats())
```

ffi/gatherer.go
```diff
@@ -0,0 +1,35 @@
...
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+var _ prometheus.Gatherer = (*Gatherer)(nil)
+
+type Gatherer struct{}
```
alarso16: It's strange to me that creating this struct isn't tied to initializing metrics. A function like `StartGatherer` that returns the struct and starts the metrics feels more intuitive (but feel free to disagree)

ffi/firewood.go
RodrigoVillar: Moved metrics functions + `Gatherer` to `metrics.go`. Also moved metrics test to `metrics_test.go`: ecdf934

ffi/src/metrics_setup.rs
RodrigoVillar: Done: 4dd8274

ffi/src/lib.rs
RodrigoVillar: Done: 805e548

ffi/firewood.go
RodrigoVillar: Done: ff34e5b

ffi/src/metrics_setup.rs
RodrigoVillar: Chose the second option of returning an error if `RECORDER` is not initialized: 1be2cd8

ffi/src/metrics_setup.rs
RodrigoVillar: Moved setting the global recorder to the once lock: 1be2cd8

ffi/firewood.go
RodrigoVillar: Update documentation to note that this function will error if the global recorder has not been initialized: 1f57bc6

ffi/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	return lst, nil
+}
+
+// Starts global recorder for metrics.
+// Returns an error if the global recorder was already initialized.
+// This function only needs to be called once.
+// An error is returned if this method is called a second time, or if it is
+// called after StartMetricsWithExporter.
```
rkuris: This says the same thing twice, same with StartMetricsWithExporter.

ffi/src/lib.rs
```diff
@@ -862,7 +860,7 @@ pub extern "C" fn fwd_start_metrics_with_exporter(metrics_port: u16) -> Value {
 /// A `Value` containing {0, "error message"} if unable to get the latest metrics.
 #[unsafe(no_mangle)]
 pub extern "C" fn fwd_gather() -> Value {
-    Value::from(metrics_setup::gather_metrics().as_bytes())
+    metrics_setup::gather_metrics().map_or_else(Into::into, |s| Value::from(s.as_bytes()))
```
rkuris: This is a little awkward, but it seems that this was the pattern previously. Would have preferred adding `use metrics_setup::{gather_metrics, gather_metrics_with_exporter}` earlier in the file, and then:
```suggestion
    gather_metrics().map_or_else(Into::into, |s| Value::from(s.as_bytes()))
```

ffi/gatherer.go
RodrigoVillar: I think `StartGatherer` would make sense, but this comes with the drawback of complicating the Firewood Go API; that is, we can only call one of the following:

```go
StartMetrics()
StartMetricsWithExporter()
StartGatherer()
```

I could live with `StartMetrics` and `StartMetricsWithExporter` since the latter is an extension of the former, but I could see how users could call `StartMetrics` and then proceeding to call `StartGatherer`.

rkuris (approved):
Approved with some minor tweaks

ffi/firewood.go
rkuris: Looks like you left the original comment there too, so now it says the same thing twice

ffi/metrics.go
RodrigoVillar: Removed duplicate documentation here: fbe67d8

ffi/src/lib.rs
alarso16: You could also do `|s| s.as_bytes().into()`, right?

ffi/src/lib.rs
```diff
@@ -821,23 +821,48 @@ pub unsafe extern "C" fn fwd_free_database_error_result(
...
 /// * `metrics_port` - the port where metrics will be exposed at
 ///
 /// # Returns
 ///
-/// A `Value` containing {0, null} if the metrics exporter successfully started.
-/// A `Value` containing {0, "error message"} if the metrics exporter failed to start.
+/// A `Value` containing {0, null} if the metrics recorder was initialized and
+/// the exporter was started.
+/// A `Value` containing {0, "error message"} if the metrics recorder was
+/// already initialized or the exporter failed to start.
```
alarso16: Could anything else happen? (e.g. metrics port taken). It might be better to just say "if an error occurs". Same above

ffi/src/lib.rs
```diff
@@ -821,23 +821,48 @@ pub unsafe extern "C" fn fwd_free_database_error_result(
...
+    metrics_setup::setup_metrics_with_exporter(metrics_port)
         .map_err(|e| e.to_string())
         .map_or_else(Into::into, Into::into)
 }
 
+/// Gather latest metrics for this process.
+///
+/// # Returns
+///
+/// A `Value` containing the latest metrics for this process.
```
alarso16: I assume this is `(len, bytes)`, but it should be clarified

alarso16 (approved):
A couple nits, but looks good to me

ffi/src/lib.rs
RodrigoVillar: Done: 6d18154

ffi/src/lib.rs
RodrigoVillar: Done: a9cad5a

ffi/src/lib.rs
RodrigoVillar: Great point, modified to use `into()` instead of `from()`: 8c2c33a

PR #1046, opened 2025-07-08
feat(checker): check for misaligned stored areas

qusuyan:
This PR adds checks for stored area misalignment. We only focus on areas that are not leaked (areas tracked in the trie or in the free list). As long as the areas not leaked are aligned, leaked areas will be aligned. 

When a misaligned area is identified, we return not only the address of this area but also who points to that area directly. If this area is allocated for a trie node, the root or the parent trie node will be returned. If this area is free, the free list head or the parent of this free area is returned. We return the parent because, if we were to fix the misalignment issue (by migrating to another area or remove it from the free list), the parent pointer needs to be updated accordingly. 

As a side effect, `FreeListIterator` will return the parent of the free list item. This design can help checker fix various issues it sees when traversing the free list but will add some overhead if it is used for other purposes.

### DIFF
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -22,10 +22,10 @@
 
 use crate::linear::FileIoError;
 use crate::logger::trace;
+use crate::node::branch::{ReadSerializable, Serializable};
 use crate::nodestore::is_aligned;
-use bincode::{DefaultOptions, Options as _};
+use integer_encoding::VarIntReader;
 use metrics::counter;
-use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::io::{Error, ErrorKind, Read};
 use std::iter::FusedIterator;
@@ -33,10 +33,15 @@ use std::num::NonZeroU64;
 use std::sync::Arc;
 
 use crate::node::persist::MaybePersistedNode;
-use crate::node::{ByteCounter, Node};
+use crate::node::{ByteCounter, ExtendableBytes, Node};
 use crate::{CacheReadStrategy, FreeListParent, ReadableStorage, SharedNode, TrieHash};
 
 use crate::linear::WritableStorage;
+
+/// Returns the maximum size needed to encode a `VarInt`.
+const fn var_int_max_size<VI>() -> usize {
+    const { (size_of::<VI>() * 8 + 7) / 7 }
+}
 
 /// [`NodeStore`] divides the linear store into blocks of different sizes.
 /// [`AREA_SIZES`] is every valid block size.
@@ -65,10 +70,6 @@ pub const AREA_SIZES: [u64; 23] = [
     1024 << 13,
     1024 << 14,
 ];
-
-pub fn serializer() -> impl bincode::Options {
-    DefaultOptions::new().with_varint_encoding()
-}
 
 pub fn area_size_hash() -> TrieHash {
     let mut hasher = Sha256::new();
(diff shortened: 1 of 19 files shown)


storage/src/checker.rs
```diff
@@ -73,7 +88,13 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
     /// Traverse all the free areas in the freelist
     fn visit_freelist(&self, visited: &mut LinearAddressRangeSet) -> Result<(), CheckerError> {
         for free_area in self.free_list_iter_inner(0) {
-            let (addr, stored_area_index, free_list_id) = free_area?;
+            let (addr, stored_area_index, free_list_id, parent_ptr) = free_area?;
+            if addr.get() % 16 != 0 {
```
rkuris: Since you're doing almost exactly the same thing multiple times in the same file, might be a good idea to make this a helper, perhaps `check_address_alignment(addr: LinearAddress, parent: ParentPtr) -> Result<(), CheckerError>` then replace all these calls with:

```
check_address_alignment(addr, ParentPtr::FreeList(parent_ptr))?;
```

The inner function then does the get, the modulus with AREA_SIZES[0], and construction of the error.

storage/src/checker.rs
```diff
@@ -39,6 +39,12 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         // 2. traverse the trie and check the nodes
         if let Some(root_address) = self.root_address() {
             // the database is not empty, traverse the trie
+            if root_address.get() % 16 != 0 {
```
rkuris: Use AREA_SIZES[0] instead of 16. For completeness, you may want to write a test in the file that defines AREA_SIZES that verifies that all AREA_SIZES are multiples of the first one, something simple like this (untested):
```
assert_eq!(
        AREA_SIZES.first(),
        AREA_SIZES.iter().reduce(|lo, hi| if hi % lo == 0 { lo } else { hi })
    );
```

storage/src/nodestore.rs
```diff
@@ -1686,16 +1687,25 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
-        Self { storage, next_addr }
+    pub(crate) const fn new(
+        storage: &'a S,
+        next_addr: Option<LinearAddress>,
+        parent_ptr: FreeListParentPtr,
+    ) -> Self {
+        Self {
+            storage,
+            next_addr,
+            parent_ptr,
```
rkuris: I'm not sure I love having the iterator know what context it is running in. Things in the iterator should only iterate. The caller should have the context and map in the missing bits.

storage/src/nodestore.rs
```diff
@@ -1711,8 +1721,12 @@ impl<S: ReadableStorage> Iterator for FreeListIterator<'_, S> {
         };
 
         // update the next address to the next free block
```
rkuris: misplaced comment, with little value. The code it is attempting to comment is self documenting, so you can probably nuke this comment.

Bad comment examples:
```
// set x to 5
let x = 5;
// call the overflow method and return errors
overflow(x)?;
```

storage/src/nodestore.rs
```diff
@@ -1711,8 +1721,12 @@ impl<S: ReadableStorage> Iterator for FreeListIterator<'_, S> {
...
+        let parent_ptr = std::mem::replace(
```
rkuris: Take a look at PrevPeekable https://docs.rs/prev-iter/latest/prev_iter/struct.PrevPeekable.html

You might be able to use this or at least implement it the same way. I think parent_ptr is too heavy and names ending in `_ptr` usually do not add any value (what is wrong with parent?)

storage/src/nodestore.rs
```diff
@@ -1727,23 +1741,33 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
...
     }
 
     // pub(crate) since checker will use this to verify that the free areas are in the correct free list
     // Since this is a low-level iterator, we avoid safe conversion to AreaIndex for performance
     pub(crate) fn free_list_iter_inner(
         &self,
         start_area_index: AreaIndex,
-    ) -> impl Iterator<Item = Result<(LinearAddress, AreaIndex, AreaIndex), FileIoError>> {
+    ) -> impl Iterator<
+        Item = Result<(LinearAddress, AreaIndex, AreaIndex, FreeListParentPtr), FileIoError>,
```
rkuris: This is way too big for a tuple, it should be named. Do we really need all these return values?

storage/src/nodestore.rs
```diff
@@ -2000,16 +2024,30 @@ mod test_free_list_iterator {
...
         );
+        let (next_addr, returned_area_index, parent_ptr) = free_list_iter.next().unwrap().unwrap();
+        assert_eq!(next_addr, LinearAddress::new(start).unwrap());
+        assert_eq!(returned_area_index, area_index);
+        assert!(matches!(
+            parent_ptr,
+            FreeListParentPtr::FreeListHead(free_list_id) if free_list_id == area_index
+        ));
 
+        let mut parent_addr = next_addr;
```
rkuris: split this test up into two test methods with an initializer. You might have more than 2 tests here.

alternatively, a block comment at the beginning of each test describing what you're doing would make the review easier and improve readability. But it's better to split it up.

rkuris (changes requested):

storage/src/nodestore.rs
```diff
@@ -130,9 +131,9 @@ pub type AreaIndex = u8;
 
 // TODO danlaine: have type for index in AREA_SIZES
 // Implement try_into() for it.
-const NUM_AREA_SIZES: usize = AREA_SIZES.len();
-const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
-const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
+pub(crate) const NUM_AREA_SIZES: usize = AREA_SIZES.len();
+pub(crate) const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
+pub(crate) const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
```
demosdemon: suggestion: it appears that these would not need to be exported if you could implement methods on `LinearAddress` (such as `const fn is_aligned(&self) -> bool`) suggesting we should convert the type alias into a newtype struct wrapping the NonZeroU64 that we can add methods to. However, this is out of scope for this change and should be a change of its own.

storage/src/lib.rs
```diff
@@ -150,6 +176,17 @@ pub enum CheckerError {
         expected_free_list: AreaIndex,
     },
 
+    /// The start address of a stored area is not a multiple of 16
+    #[error(
+        "The start address of a stored area is not a multiple of 16: {address} (parent: {parent_ptr:?})"
```
rkuris: The hardcoded 16 should be computed and inserted here when we change the minimum aligment (if ever):
```suggestion
        "The start address of a stored area is not a multiple of {alignment}: {address} (parent: {parent_ptr:?})"
```

storage/src/lib.rs
```diff
@@ -98,6 +97,33 @@ pub fn empty_trie_hash() -> TrieHash {
...
+}
+
+/// This enum encapsulates what points to the stored area allocated for a trie node.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TrieNodeParent {
+    /// The stored area is the root of the trie, so the header points to it
+    Root,
+    /// The stored area is not the root of the trie, so a parent trie node points to it
+    Parent(LinearAddress, usize),
+}
```
rkuris, line 117:
```suggestion
type TrieNodeParent = Option<(LinearAddress, usize)>;
```
You can use `TrieNodeParent::None` in the root case, and `TrieNodeParent::Some((..., ...))` in the other.

storage/src/nodestore/mod.rs
```diff
@@ -54,7 +54,7 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 
 // Re-export types from alloc module
-pub use alloc::{AreaIndex, LinearAddress};
+pub use alloc::{AreaIndex, FreeLists, LinearAddress, MIN_AREA_SIZE};
```
rkuris: Do we need to export this? What's wrong with `AREA_SIZES.first().expect("at least one area size")`?

Smaller surface areas are good.

storage/src/nodestore/alloc.rs
```diff
@@ -545,61 +566,120 @@ impl<S: ReadableStorage> Iterator for FreeListIterator<'_, S> {
...
+        loop {
+            if let Some(next) = next_fn(&mut self.free_list_iter) {
+                // the current free list is not exhausted, return the next free area
+                return Some(next);
+            }
+
+            self.current_free_list_id = self
+                .current_free_list_id
+                .checked_add(1)
+                .expect("free list id overflow is not possible");
```
rkuris: This might be okay, but I think it's not properly expressing what's going on.

The point is not that free list id is overflowing, it's the underlying u8. What you're allowing here is setting `current_free_list_id` equal to something that isn't a valid AREA_SIZE. This only works because you later use it with `get` and handle the overflow.

Recommend adding a helper, say `next_free_list_id` which handles all the details and returns an Option<AreaIndex>.

If you prefer to stick with this approach, I'm okay with adding a bunch of comments explaining what you're doing.

I'm not sure why you aren't just maintaining an enumerated iterator over the freelist either.

rkuris:
Some duplicated work in #1062 here. Probably should take the alignment debug check fixes from that PR and move them here.

storage/src/lib.rs
qusuyan: Does it really make a difference? I like to keep things explicit here.

storage/src/nodestore/mod.rs
qusuyan: I should check cursor generated code more carefully here :/

storage/src/lib.rs
rkuris: I think it's clearer if you say "TrieNodeParent is None" or "there is no TrieNodeParent" instead of "TrieNodeParent is Root". Also, `Option` has a lot of nifty methods that you can use in the implementations, further improving readability with less verbosity.

rkuris (approved):
Not much left to fix

PR #1047, opened 2025-07-08
feat(delayed-persist): Part 2.2: Branch Children

rkuris:
Branch children might also not be persisted.

This also splits up hash_helper which got too long for clippy.

- [x] Roots may not be persisted (Part 1: #1041)
- [x] BranchNode children can now be Node, LinearAddress or a
MaybePersistedNode (Part 2: #1045 and this PR #1047)
- [ ] When converting a `MutableProposal` to an `ImmutableProposal`,
don't actually allocate space, just create `SharedNode`s from the
`Node`s.
- [ ] Remove `NodeStoreHeader` from `NodeStore`. The header should move
into the `RevisionManager`.
- [ ] Allocate new nodes from the recently-freed nodes from an expired
revision, then from the freelist. This can be done by maintaining a
local freelist from the nodes deleted in the expiring revision and
falling back to the actual freelist. Deleted nodes that are not reused
must still be pushed onto the freelist.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -2,10 +2,6 @@
 // See the file LICENSE.md for licensing terms.
 
 #![expect(
-    clippy::match_same_arms,
-    reason = "Found 1 occurrences after enabling the lint."
-)]
-#![expect(
     clippy::missing_errors_doc,
     reason = "Found 6 occurrences after enabling the lint."
 )]
@@ -392,14 +388,11 @@ impl<T: HashedNodeReader> Merkle<T> {
                 for (childidx, child) in b.children.iter().enumerate() {
                     let (child_addr, child_hash) = match child {
                         None => continue,
-                        Some(Child::Node(_)) => continue, // TODO
-                        Some(Child::AddressWithHash(addr, hash)) => (*addr, Some(hash)),
-                        Some(Child::MaybePersisted(maybe_persisted, hash)) => {
-                            // For MaybePersisted, we need the address if it's persisted
-                            match maybe_persisted.as_linear_address() {
-                                Some(addr) => (addr, Some(hash)),
-                                None => continue, // Skip if not persisted
-                            }
+                        Some(node) => {
+                            let (Some(addr), hash) = (node.persisted_address(), node.hash()) else {
+                                continue;
+                            };
+                            (addr, hash)
                         }
                     };
 
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -142,6 +142,10 @@ fn get_helper<T: TrieReader>(
                         let child = nodestore.read_node(*addr)?;
                         get_helper(nodestore, &child, remaining_key)
                     }
+                    Some(Child::MaybePersisted(maybe_persisted, _)) => {
+                        let child = maybe_persisted.as_shared_node(nodestore)?;
+                        get_helper(nodestore, &child, remaining_key)
+                    }
                 },
             }
         }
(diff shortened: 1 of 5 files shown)


storage/src/nodestore.rs
```diff
@@ -1229,15 +1286,20 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
...
-            // everything already hashed
+            // 6. everything already hashed
 
             for (nibble, child) in b.children.iter_mut().enumerate() {
                 // if this is already hashed, we're done
-                if matches!(child, Some(Child::AddressWithHash(_, _))) {
+                if matches!(
+                    child,
+                    Some(Child::AddressWithHash(_, _) | Child::MaybePersisted(_, _))
+                ) {
```
demosdemon: Instead of the `matches!` expression, it feels like this could be more clear if you do something like `if child.hash().is_some()` where `hash` is a method on `Child` that destructures the enum down to `Option<HashType>`.

storage/src/nodestore.rs
```diff
@@ -1246,8 +1308,16 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
                 // the branch and hash it. This has the side effect of dropping the [Child::Node]
                 // that was allocated. This is fine because we're about to replace it with a
                 // [Child::AddressWithHash].
-                let Some(Child::Node(child_node)) = std::mem::take(child) else {
-                    continue;
+                let child_node = match std::mem::take(child) {
+                    Some(Child::Node(node)) => node,
+                    Some(Child::MaybePersisted(maybe_persisted, _)) => {
```
demosdemon: When would this branch ever be hit? The previous if check should continue if `child` is a `MaybePersisted`.

storage/src/nodestore.rs
```diff
@@ -1246,8 +1308,16 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
...
+                        // For MaybePersisted, we need to get the node to hash it
+                        match maybe_persisted.as_shared_node(self) {
+                            Ok(shared_node) => shared_node.deref().clone(),
+                            Err(_) => continue, // Skip if we can't read the node
+                        }
+                    }
+                    _ => continue,
                 };
```
demosdemon, line 1303: Reviewing my two comments again, this entire thing feels like it should be a single match on the child and do separate things depending on the variant. The if branch above is redundantly matching what this is also matching.

firewood/src/merkle.rs
```diff
@@ -142,6 +142,10 @@ fn get_helper<T: TrieReader>(
                         let child = nodestore.read_node(*addr)?;
                         get_helper(nodestore, &child, remaining_key)
                     }
+                    Some(Child::MaybePersisted(maybe_persisted, _)) => {
+                        let child = maybe_persisted.as_shared_node(nodestore)?;
+                        get_helper(nodestore, &child, remaining_key)
+                    }
```
demosdemon, line 144: This is redundant. Feels like something should be implemented on the child to yield a node then call get_helper only once.

firewood/src/merkle.rs
```diff
@@ -390,6 +394,13 @@ impl<T: HashedNodeReader> Merkle<T> {
                         None => continue,
                         Some(Child::Node(_)) => continue, // TODO
                         Some(Child::AddressWithHash(addr, hash)) => (*addr, Some(hash)),
+                        Some(Child::MaybePersisted(maybe_persisted, hash)) => {
+                            // For MaybePersisted, we need the address if it's persisted
```
demosdemon: Same, this redundant. Feels like we need a method to yield `Result<(LinearAddress, HashType, _>` on child.

storage/src/nodestore.rs
rkuris: Great suggestion!

firewood/src/merkle.rs
rkuris: Yes, but this wasn't trivial. The problem is that in the latter two cases, we have `Arc<Node>` but in the first case we just have a `Node`. I don't want to wrap the `Child::Node` case into an Arc. Suggestions welcome.

storage/src/nodestore.rs
rkuris: Yeah it was definitely wrong. Good catch.

demosdemon (approved):

PR #1048, opened 2025-07-09
chore: Split nodestore into functional areas

rkuris:
A nodestore and its related types evolved over time becoming more complex. This change splits up the functionality into different areas to make it easier to find specific parts of the nodestore logic.
    
Doc comments about the layout are extensive and are not repeated here.

Other than moving things around, some additional changes are:
 - `UpdateError` was dead code. It was used in early development of
   merkle's insert but is no longer used.
 - Accessors for various members of NodeStoreHeader instead of
   referencing them directly. This makes auditing their use a little
   easier, and improves scoping and readability.


storage/src/nodestore/mod.rs
```diff
@@ -0,0 +1,1008 @@
...
+    pub(crate) const fn size(&self) -> u64 {
+        self.header.size()
+    }
+
+    pub(crate) fn physical_size(&self) -> Result<u64, FileIoError> {
+        self.storage.size()
+    }
+}
+
+pub(crate) struct FreeListIterator<'a, S: ReadableStorage> {
```
qusuyan: Maybe move `FreeListIterator` to `alloc.rs` too

demosdemon (approved):

storage/src/nodestore/mod.rs
demosdemon: After looking at #1046, we probably need a separate `iter.rs` for iterating.

PR #1052, opened 2025-07-10
feat(checker): add error to report finding leaked areas

qusuyan:
This is the first part of this diff: #1033


storage/src/iter.rs
```diff
@@ -0,0 +1,108 @@
...
+            (Some(_), None) => {
+                // we exhausted our limit, but there are still items left including the item
+                // we just removed from the iterator.
+                let n = iter.count().saturating_add(1); // +1 for the current item
+                write!(writer, "{sep}... ({n} more hidden)")?;
+                break;
+            }
+            (None, _) => break,
+        }
+    }
```
rkuris: Avoid counting yourself, let the iterator do that for you. And why not make limit just a regular usize and handle the zero case, especially since you seem to handle the empty iterator case.

I didn't spend more time on this but there are some join methods that might also do what you want here.

```suggestion
   let mut iter = iter.into_iter();
    let Some(first) = iter.next() else {
        return Ok(());
    };
    if limit == 0 {
        return Ok(());
    }
    write!(writer, "{first}")?;
    let mut limiter = iter.by_ref().take(limit);
    while let Some(item) = limiter.next() {
        write!(writer, "{sep}{item}")?;
    }
    let remaining = iter.count();
    if remaining > 0 {
        write!(writer, "{sep} ... and {} more", remaining)?;
    }
```

storage/src/iter.rs
```diff
@@ -0,0 +1,108 @@
...
+    limit: Option<NonZeroUsize>,
+) -> std::fmt::Result {
+    if let Some(limit) = limit {
+        write_iter_with_limit(writer, iter, sep, limit)
+    } else {
+        write_iter_no_limit(writer, iter, sep)
+    }
+}
+
+fn write_iter_no_limit(
```
rkuris: This code seems like it should be callable by the limit function on a `by_ref` iterator.

storage/src/iter.rs
```diff
@@ -0,0 +1,108 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    #![expect(clippy::unwrap_used)]
+
+    use super::*;
+
+    #[test]
```
rkuris: use the `test_case` crate here

storage/src/lib.rs
```diff
@@ -98,6 +99,32 @@ pub fn empty_trie_hash() -> TrieHash {
...
 }
 
+const MAX_LEAKED_AREAS_PRINT: usize = 10;
+
+/// A collection of leaked areas
+pub struct LeakedAreas(Vec<(LinearAddress, AreaIndex)>);
+
+impl std::fmt::Debug for LeakedAreas {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        struct DisplayArea<'a>(&'a (LinearAddress, AreaIndex));
```
rkuris: Isn't this a `LeakedArea`? I think this needs a type.

storage/src/iter.rs
demosdemon: FYI, I wrote this on the other PR to avoid allocating intermediate vecs and strings.

What do you mean by using a `by_ref` iterator? Can you please elaborate?

storage/src/iter.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// # Returns
+/// A `std::fmt::Result` indicating success or failure of the write operation.
+pub(crate) fn write_iter(
+    writer: &mut (impl std::fmt::Write + ?Sized),
+    iter: impl IntoIterator<Item: std::fmt::Display>,
+    sep: impl std::fmt::Display,
+    limit: Option<usize>,
+) -> std::fmt::Result {
+    match limit {
```
qusuyan, line 28: @demosdemon I think this is what @rkuris is looking for - you mentioned you added this write iterator in another PR?

storage/src/iter.rs
```diff
@@ -0,0 +1,78 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! Internal utilities for working with iterators.
+
+/// Writes an iterator of displayable items to a writer, separated by a specified separator.
```
rkuris:
```suggestion
/// Writes a limited number of items to a writer, separated by a specified separator.
```

storage/src/iter.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// # Arguments
+/// - `writer`: The writer to which the items will be written.
+/// - `iter`: An iterator of items that implement `std::fmt::Display`.
+/// - `sep`: A separator that will be used between items.
+/// - `limit`: An optional limit on the number of items to write.
+///
+/// # Returns
+/// A `std::fmt::Result` indicating success or failure of the write operation.
+pub(crate) fn write_iter(
```
rkuris: Maybe `write_limited_with_sep`? Generally functions that end in `iter` return iterators, which this does not.

storage/src/iter.rs
```diff
@@ -0,0 +1,78 @@
...
+
+//! Internal utilities for working with iterators.
+
+/// Writes an iterator of displayable items to a writer, separated by a specified separator.
+///
+/// - If `limit` is `Some(0)`, it will only write the number of hidden items.
+/// - If `limit` is `Some(n)`, it will write at most `n` items, followed by a message
+///   indicating how many more items were not written.
+/// - If `limit` is `None`, it will write all items without any limit.
+///
```
rkuris, line 13: Add a caution about this function not working with iterators that do not terminate like `once(1).cycle()`

storage/src/iter.rs
```diff
@@ -0,0 +1,78 @@
...
+            if hidden_count > 0 {
+                write!(writer, "{sep}... ({hidden_count} more hidden)")?;
+            }
+            Ok(())
+        }
+        None => write_all_iter(writer, iter, &sep),
+    }
+}
+
+fn write_all_iter(
```
rkuris: This could still use a block comment. A better name is a good idea too, as it doesn't return an iterator.

rkuris (approved):
approved with nits

PR #1053, opened 2025-07-11
chore: update `golangci-lint`

RodrigoVillar:
While looking into Taskfiles for `ffi`, I noticed that we're still using `golangci-lint v1`  while AvalancheGo is already using `golangci-lint v2`. This PR upgrades the version of `golangci-lint` used in `ffi` to `v2`.

To maintain synchrony between the Firewood and AvalancheGo linter rules, an additional CI job (`expected-golangci-yaml-diff`) ensures that the difference between the two `.golangci.yaml` files is equal to the contents of `golangci_yaml_expected_changes.txt`.

## What

- Copied over current AvalancheGo `golangci.yml` file and removing AvalancheGo-specific rules. 
- Bumped version of `golangci-lint` action used in CI.
- Added `expected-golangci-yaml-diff` CI job
- Linted files

## How This Was Tested

Ran `golangci-lint` locally + CI.


rkuris (commented):
Questions:

Is there any way to share the linter configuration between these two projects?
(if no) Can you provide a pointer to where this is being done in avalanchego so I can compare?
(if yes) let's do that instead!

ffi/.golangci.yaml
```diff
@@ -73,98 +58,148 @@ linters:
...
+    errorlint:
+      # Check for plain type assertions and type switches.
+      asserts: false
+      # Check for plain error comparisons.
+      comparison: false
+    forbidigo:
+      # Forbid the following identifiers (list of regexp).
+      forbid:
+        - pattern: require\.Error$(# ErrorIs should be used instead)?
+        - pattern: require\.ErrorContains$(# ErrorIs should be used instead)?
```
alarso16: We need this one, because our errors are translated from Rust to Go. `ErrorContains` (and provide a string) works really well for this.

RodrigoVillar:
@rkuris I don't believe there's a way to share linter configs between the two projects.

Linter rules in AvalancheGo: https://github.com/ava-labs/avalanchego/blob/master/.golangci.yml

ffi/.golangci.yaml
alarso16: Also, in general we prefer:
```
r := require.New(t)
r.NoError(..)
```
Which is why the linter did catch this i think

RodrigoVillar: Hmm interesting - currently, we actually enforce this rule:

https://github.com/ava-labs/firewood/blob/ee5d1d8ee292ca346fe6ceb33af5f4149cb8a317/ffi/.golangci.yaml#L106

However, since we only use `r` as you mentioned, the regexes never get matched.

RodrigoVillar: Since we didn't enforce the `require` forbids, going to remove them from this PR: a92cbfe

rkuris:
> @rkuris I don't believe there's a way to share linter configs between the two projects.
> 
> Linter rules in AvalancheGo: https://github.com/ava-labs/avalanchego/blob/master/.golangci.yml

I'm worried about them getting out of sync again. Possible solutions:

1 - Generate the golangci.yml file by fetching it from avalanchego's main and tweaking it before running the ci rule
2 - Add a new ci rule that verifies that it only contains the expected changes
3 - Use build.rs in the ffi directory to fetch avalanchego's rules and update them if necessary.

These are in order of preference.

RodrigoVillar:
> > @rkuris I don't believe there's a way to share linter configs between the two projects.
> > Linter rules in AvalancheGo: https://github.com/ava-labs/avalanchego/blob/master/.golangci.yml
> 
> I'm worried about them getting out of sync again. Possible solutions:
> 
> 1 - Generate the golangci.yml file by fetching it from avalanchego's main and tweaking it before running the ci rule 2 - Add a new ci rule that verifies that it only contains the expected changes 3 - Use build.rs in the ffi directory to fetch avalanchego's rules and update them if necessary.
> 
> These are in order of preference.

Went with the second option. The expected diff between the Firewood and AvalancheGo `.golangci.yml` files is now stored in `golangci_yaml_expected_changes.txt` and is verified by the `expected-golangci-yaml-diff` CI job: b7d7197

ffi/scripts/verify_golangci_yaml_changes.sh
```diff
@@ -0,0 +1,12 @@
...
+
+curl -o /tmp/upstream.yml https://raw.githubusercontent.com/ava-labs/avalanchego/refs/heads/master/.golangci.yml
+    
+# Generate diff
+diff /tmp/upstream.yml .golangci.yaml > /tmp/diff.txt || true
+
+# Compare with expected diff
+if ! diff /tmp/diff.txt .golangci_yaml_expected_changes.txt; then
+    echo ".golangci.yaml has unexpected changes from AvalancheGo"
+    exit 1
```
demosdemon: Not sure I'm a fan of this as it introduces a period of time where our CI _will_ fail whenever the upstream file changes regardless of the changes that were made. This will add unnecessary friction to PRs.

This script should only run when the `.golangci.yaml` in this repo is being updated so that only then we validate the changes against the expected diff.

ffi/.golangci_yaml_expected_changes.txt
demosdemon, line 1: move this to the `.github` directory.

demosdemon (changes requested):

ffi/scripts/verify_golangci_yaml_changes.sh
rkuris: It should just run on push to main. I catch those after the fact and add issues later.

ffi/.golangci_yaml_expected_changes.txt
rkuris: +1 since only github cares about this

.github/workflows/ci.yaml
```diff
@@ -304,3 +304,12 @@ jobs:
       - name: Check for stale dependencies
         run: cargo machete --with-metadata
 
+  expected-golangci-yaml-diff:
```
rkuris: This should be done only when pushing to main (that is, the PR is complete) reducing friction for someone working on something different.

I catch these as it sends me an email on workflow failures on main.

gh-pages works like this if you need an example. It has a test branch name in case you want to verify it too.

RodrigoVillar: Done: 4bdc9a9

ffi/.golangci_yaml_expected_changes.txt
RodrigoVillar: Done: 468839b

ffi/scripts/verify_golangci_yaml_changes.sh
RodrigoVillar: This will now run on pushes to `main`: 4bdc9a9

.github/scripts/verify_golangci_yaml_changes.sh
```diff
@@ -1,12 +1,17 @@
 #!/usr/bin/env bash
 
+if ! [[ "$0" =~ scripts/verify_golangci_yaml_changes.sh ]]; then
+  echo "must be run from .github"
+  exit 255
+fi
```
rkuris: nit: prefer fixing it rather than making the person running it do that:
```sh
cd "$(dirname "$(realpath "$0")")"/..
```

.github/scripts/verify_golangci_yaml_changes.sh
```diff
@@ -1,12 +1,17 @@
...
-    
+
 # Generate diff
-diff /tmp/upstream.yml .golangci.yaml > /tmp/diff.txt || true
+diff /tmp/upstream.yml ../ffi/.golangci.yaml > /tmp/diff.txt || true
 
 # Compare with expected diff
 if ! diff /tmp/diff.txt .golangci_yaml_expected_changes.txt; then
-    echo ".golangci.yaml has unexpected changes from AvalancheGo"
+    echo "ffi/.golangci.yaml has unexpected changes from AvalancheGo"
```
rkuris, line 17: nit: might be good to show the commands to generate a new one

I'll probably be the one that has to fix it and it isn't likely to happen very often, I'm sure to forget

.github/scripts/verify_golangci_yaml_changes.sh
demosdemon: Yeah, the expectation should that it's run from project root and not within a subdir.

RodrigoVillar: Done: 454f9e5

.github/scripts/verify_golangci_yaml_changes.sh
RodrigoVillar: Added documentation for generating diff: 094aa7e

ffi/.golangci.yaml
```diff
@@ -1,7 +1,6 @@
 # https://golangci-lint.run/usage/configuration/
+version: "2"
 run:
-  timeout: 10m
```
alarso16, line 3: Why remove the timeout?

demosdemon (approved):

PR #1055, opened 2025-07-12
feat(delayed-persist): dump unpersisted nodestore

rkuris:
dump has to work with unpersisted nodes, as there are a bunch of tests that assume an immutable proposal can be dumped. This was useful anyway while debugging some tests in the next commit, and small enough to commit separately.

### DIFF
--- a/storage/src/nodestore/mod.rs
+++ b/storage/src/nodestore/mod.rs
@@ -330,12 +330,23 @@ where
     fn root_node(&self) -> Option<SharedNode> {
         self.deref().root_node()
     }
+    fn root_as_maybe_persisted_node(&self) -> Option<MaybePersistedNode> {
+        self.deref().root_as_maybe_persisted_node()
+    }
 }
 
 /// Reads the root of a merkle trie.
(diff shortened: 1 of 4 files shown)


storage/src/nodestore/mod.rs
```diff
@@ -330,12 +330,23 @@ where
...
+/// The root may be None if the trie is empty.
 pub trait RootReader {
     /// Returns the root of the trie.
+    /// Callers that just need the node at the root should use this function.
     fn root_node(&self) -> Option<SharedNode>;
+
+    /// Returns the root of the trie as a `MaybePersistedNode`.
+    /// Callers that might want to modify the root or know how it is stored
+    /// should use this function.
+    fn root_as_maybe_persisted_node(&self) -> Option<MaybePersistedNode>;
```
rkuris, line 349: In a future diff, this will just be called `root` but I want it verbose during the rest of the development of this feature as it helped identify areas still to changed.

demosdemon (approved):

PR #1059, opened 2025-07-14
Qusuyan/leak area error

qusuyan:
Add functions to split leaked ranges into areas that can be added to free lists. This concludes this diff: #1033.


storage/src/checker.rs
```diff
@@ -91,6 +97,95 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+        }
+
+        // we assume that all areas are aligned to `MIN_AREA_SIZE`, in which case leaked ranges can always be split into free areas perfectly
+        debug_assert!(current_addr == leaked_range.end);
+        leaked
+    }
+
+    // Find the area index and size of the stored area at the given address if the area is valid.
+    // TODO: there should be a way to read stored area directly instead of try reading as a free area then as a node
+    fn find_stored_area_index_and_size(
```
rkuris: nit: naming: read_freed_area

This would have been easier to name if we had a type FreedArea or something, but I think that's too intrusive now.

storage/src/nodestore/alloc.rs
```diff
@@ -213,22 +213,21 @@ impl FreeArea {
         storage: &S,
         address: LinearAddress,
     ) -> Result<(Self, AreaIndex), FileIoError> {
-        let free_area_addr = address.get();
-        let stored_area_stream = storage.stream_from(free_area_addr)?;
+        let mut area_stream = storage.stream_from(address.get())?;
```
rkuris: nit: the temporary makes debugging and improves readability.

storage/src/nodestore/alloc.rs
```diff
@@ -213,22 +213,21 @@ impl FreeArea {
...
         let stored_area: StoredArea<Area<Node, FreeArea>> = serializer()
-            .deserialize_from(stored_area_stream)
+            .deserialize_from(&mut area_stream)
```
rkuris: Why borrow here instead of move?

storage/src/nodestore/mod.rs
```diff
@@ -472,6 +472,12 @@ pub struct NodeStore<T, S> {
     storage: Arc<S>,
 }
 
+impl<T, S> NodeStore<T, S> {
+    pub(crate) fn storage(&self) -> &S {
+        self.storage.as_ref()
+    }
+}
+
```
rkuris: nit: the surface area seems to be getting bigger here. Can this be better abstracted into something like `fn read_freed_area` on NodeStore instead of doing it here? Then it could live in `alloc.rs`.

rkuris (approved):
Nothing but nits, nice job

storage/src/nodestore/alloc.rs
qusuyan: absolutely not - let me fix that

storage/src/checker.rs
qusuyan: I renamed it to `read_leaked_area` since it may be free or allocated.

PR #1060, opened 2025-07-14
feat(deferred-allocate): UnpersistedIterator

rkuris:
An UnpersistedIterator iterates through a NodeStore looking for unpersisted nodes. Unpersisted nodes are  only of the type MaybeUnpersisted that are not yet persisted.

Note that it's impossible to have a persisted parent for a MaybeUnpersistedNode, so it's relatively easy   to find the unpersisted ones.

There are tests that verify that the nodes are returned in the order that they will need persisting (that is, depth first).


storage/src/node/branch.rs
```diff
@@ -113,6 +135,26 @@ impl Child {
...
+    /// - [`Child::Node`] variants (unpersisted nodes)
+    /// - [`Child::MaybePersisted`] variants that are not yet persisted
+    #[must_use]
+    pub fn persist_info(&self) -> Option<(LinearAddress, &HashType)> {
+        match self {
+            Child::AddressWithHash(addr, hash) => Some((*addr, hash)),
+            Child::MaybePersisted(maybe_persisted, hash) => {
+                maybe_persisted.as_linear_address().map(|addr| (addr, hash))
+            }
+            Child::Node(_) => None,
```
qusuyan, line 153: When do you use `Node` now? My understanding is that it indicates the node is persisted but is loaded in memory, in which case it should behave like a persisted node?

storage/src/nodestore/persist.rs
```diff
@@ -74,6 +80,117 @@ impl<T, S: WritableStorage> NodeStore<T, S> {
...
+                let shared_node = next_child
+                    .as_shared_node(self.store)
+                    .expect("in memory, so IO is impossible");
+
+                if shared_node.is_leaf() {
+                    return Some(next_child);
+                }
+
+                // It's a branch, so we need to get its children
+                if let Some(branch) = shared_node.as_branch() {
```
qusuyan: why not using `match`?

storage/src/nodestore/persist.rs
```diff
@@ -247,3 +364,207 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
+        linear::memory::MemStore,
+        node::{BranchNode, LeafNode, Node},
+        nodestore::MutableProposal,
+    };
+
+    /// Helper to create a test node store with a specific root
+    fn create_test_store_with_root(root: Node) -> NodeStore<MutableProposal, MemStore> {
+        let mem_store = MemStore::new(vec![]).into();
+        let mut store = NodeStore::new_empty_proposal(mem_store);
+        store.mut_root().replace(root);
```
qusuyan, line 383: nit: why is this function `mut_root()` instead of `root_mut()`?

storage/src/nodestore/persist.rs
```diff
@@ -247,3 +364,207 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
+        assert_eq!(nodes.len(), 4);
+
+        // The branch should be last (depth-first)
+        assert_eq!(&*nodes[3], &branch);
+
+        // Children should come first - verify all expected leaf nodes are present
+        let children_nodes = &nodes[0..3];
+        assert!(children_nodes.iter().any(|n| **n == leaves[0]));
+        assert!(children_nodes.iter().any(|n| **n == leaves[1]));
+        assert!(children_nodes.iter().any(|n| **n == leaves[2]));
```
qusuyan, line 492: Does the order of children matter? Seems to me it does

storage/src/nodestore/persist.rs
```diff
@@ -74,6 +80,117 @@ impl<T, S: WritableStorage> NodeStore<T, S> {
...
+                if let Some(branch) = shared_node.as_branch() {
+                    // Create an iterator over unpersisted children
+                    let unpersisted_children: Vec<MaybePersistedNode> = branch
+                        .children
+                        .iter()
+                        .filter_map(|child_opt| {
+                            child_opt
+                                .as_ref()
+                                .and_then(|child| child.unpersisted().cloned())
+                        })
```
qusuyan, line 168: may be add a `rev()` to make sure that childrens are iterated in increasing order.

storage/src/nodestore/persist.rs
```diff
@@ -247,3 +364,207 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
+    fn test_nested_branches() {
+        let leaves = [
+            create_leaf(&[1], &[100]),
+            create_leaf(&[2], &[200]),
+            create_leaf(&[3], &[255]),
+        ];
+
+        // Create a nested structure: root -> branch1 -> leaf[0]
+        //                                -> leaf[1]
+        //                                -> branch2 -> leaf[2]
```
qusuyan, line 505: The description does not seem to match with test. 
```suggestion
        // Create a nested structure: root -> leaf[0]
        //                                -> leaf[1]
        //                                -> inner_ branch -> leaf[2]
        //                                -> persisted_node
```

storage/src/node/branch.rs
demosdemon: I noticed this as well while touching serialization logic. It appears `Child::Node` is mostly if not completely used by `firewood` and not `firewood-storage` and I'm wondering if including `Child::Node` as part of storage is a violation of separation of concerns.

rkuris: `Child::Node` is used only in mutable proposals. This is the use case where we need branches and children to have a generic type, but it doesn't do that today. In this context, `Child::Node` is technically invalid, so this could have said `unreachable!()` to better doc it.

We don't want an `Arc<Node>` for mutable proposals just yet because it is fairly normal to make multiple mutations to the same node while creating the proposal. Once all the operations are complete, all `Node` children are replaced with `(LinearAddress, HashType)` children. A future diff will change it to create `MaybePersisted` children to defer the allocation.

storage/src/nodestore/persist.rs
rkuris: The child order doesn't actually matter. In fact ,this could be done in parallel in a future PR.

storage/src/node/branch.rs
demosdemon: Perhaps we should go further into the mutable vs immutable nodes vs immutable but not persisted split and carry that through the child representation in branches as well: 
- mutable nodes can reference immutable nodes (both persisted and non-persisted) as well as other mutable nodes
- persisted immutable nodes can only reference other persisted immutable nodes
- non-persisted immutable nodes can reference both persisted and non-persisted immutable nodes (no mutable nodes)
- Reading from storage can only yield persisted immutable nodes
- Storage layer converts mutable nodes to immutable non-persisted which are transparently updated to immutable persisted when appropriate. 
  - These maybe-persisted nodes can only be referenced by other non-persisted nodes as well as mutable nodes. 
  - When converting to a persisted node, the internal repr can only reference other persisted nodes (i.e., other non-persisted references must have been upgraded).

Big downside is that we would need to duplicate all datastructures from branch upward.

storage/src/nodestore/persist.rs
rkuris: Yeah it should be. I can do that as a different PR.

storage/src/node/branch.rs
```diff
@@ -103,6 +103,28 @@ impl Child {
...
+            maybe_persisted.unpersisted()
+        } else {
+            None
+        }
+    }
+
+    /// Return a mutable reference to the unpersisted node if the child is an
+    /// unpersisted [`Child::MaybePersisted`] variant, otherwise None.
+    #[must_use]
+    pub fn unpersisted_mut(&mut self) -> Option<&mut MaybePersistedNode> {
```
demosdemon: What is the usecase for mutable access to this? `MaybePersistedNode` is just a wrapper around `Arc`. You shouldn't need mutable access as we're only swapping the `ArcSwap` pointer (which doesn't require `&mut`).

storage/src/nodestore/persist.rs
rkuris: No, it does not, and arguably I don't want the test to care, as the order could change.

storage/src/nodestore/persist.rs
rkuris: Because there are only two cases, branch or leaf, I only have code for branches, and I want a branch type.

storage/src/node/branch.rs
rkuris: We wouldn't need to duplicate any data structures if we used some generics. If we had `BranchNode<Mutable>` and `BranchNode<Immutable>` which contained an array of `Child<Mutable>` or `Child<Immutable>` respectively, we could fix this, but that would be under a different PR as well, and maybe isn't worth it right now.

storage/src/nodestore/persist.rs
qusuyan: I think that is the exact case you wanna use a `match` clause
```suggestion
                match shared_node {
                       Node::Leaf(_) => return Some(next_child),
                       Node::Branch(branch) => { ... }
                }
```
Right now it makes me wonder if there is a third case

storage/src/node/branch.rs
rkuris: Good catch, didn't need it as you mentioned.

storage/src/nodestore/persist.rs
```diff
@@ -74,6 +80,117 @@ impl<T, S: WritableStorage> NodeStore<T, S> {
...
+                        .collect();
+
+                    // Push new child iterator to the stack
+                    if !unpersisted_children.is_empty() {
+                        self.child_iter_stack
+                            .push(Box::new(unpersisted_children.into_iter()));
+                    }
+                }
+
+                self.stack.push(next_child); // visit this node after the children
```
qusuyan: For the same reason I think this line should go inside the if clause
```suggestion
                     self.stack.push(next_child); // visit this node after the children
                }
```

storage/src/nodestore/persist.rs
rkuris: `match` created yet another level of indentation, but I switched to `if let else`. IMO slightly less readable in both the `match` and `if let else` cases, but not terrible with a comment.

PR #1061, opened 2025-07-15
Rkuris/defer allocations

rkuris:


storage/src/checker/mod.rs
```diff
@@ -43,19 +43,35 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
-                opt.hash_check,
-            )?;
+            if let Some(root_address) = root.as_linear_address() {
+                self.check_area_aligned(
+                    root_address,
+                    StoredAreaParent::TrieNode(TrieNodeParent::Root),
+                )?;
+                // For hash checking, we need the root hash if it exists
+                #[cfg_attr(not(feature = "ethhash"), expect(clippy::useless_conversion))]
+                let root_hash = self.root_hash().map(HashType::from);
```
demosdemon: there's the [into_hash_type](https://github.com/ava-labs/firewood/blob/37a217639d005743eb2c0bf426d26ca39f46caaa/firewood/src/proof.rs#L212) method for this

storage/src/nodestore/persist.rs
```diff
@@ -282,9 +340,12 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
             RINGSIZE
         ];
 
-        for (&addr, &(area_size_index, ref node)) in &self.kind.new {
+        // Process each unpersisted node
+        for node in unpersisted_nodes {
+            let shared_node = node.as_shared_node(self).expect("in memory, so no IO");
```
demosdemon, line 356: can you explain this `expect` a bit? Is this because the node is `MaybePersisted::Unpersisted`? If so, then perhaps `UnPersistedNodeIterator` should yield `SharedNode`s and not `MaybePersisted` so that structurally this expect isn't necessary.

storage/src/nodestore/mod.rs
```diff
@@ -827,41 +830,8 @@ mod tests {
...
-
-        let computed_length =
-            NodeStore::<std::sync::Arc<ImmutableProposal>, MemStore>::stored_len(&node);
-
-        let mut serialized = Vec::new();
-        node.as_bytes(0, &mut serialized);
-        assert_eq!(serialized.len() as u64, computed_length);
-    }
     #[test]
+    #[ignore = "https://github.com/firewood-io/firewood/issues/1054"]
```
demosdemon: 404, repo is under `ava-labs`.

storage/src/nodestore/mod.rs
```diff
@@ -681,23 +617,92 @@ where
...
+        &self,
+        addr: LinearAddress,
+    ) -> Result<(AreaIndex, u64), FileIoError> {
+        let mut area_stream = self.storage.stream_from(addr.get())?;
+
+        let index: AreaIndex = area_stream.read_byte().map_err(|e| {
+            self.storage.file_io_error(
+                Error::new(ErrorKind::InvalidData, e),
+                addr.get(),
+                Some("deserialize".to_string()),
```
demosdemon: nit: or even more context than this

```suggestion
                Some("area_index_and_size".to_string()),
```

storage/src/nodestore/mod.rs
```diff
@@ -681,23 +617,92 @@ where
...
+            )
+        })?;
+
+        let size = *AREA_SIZES
+            .get(index as usize)
+            .ok_or(self.storage.file_io_error(
+                Error::other(format!("Invalid area size index {index}")),
+                addr.get(),
+                None,
```
demosdemon: nit:

```suggestion
                Some("area_index_and_size".to_string()),
```

demosdemon (commented):
nothing jumped out at me except nits. will make another pass later

storage/src/nodestore/persist.rs
rkuris: I wish it could. It needs the `MaybePersisted` entry because it needs to call `persist_at` on it.

storage/src/nodestore/mod.rs
rkuris: I'm adding it, but this error will probably stutter, since the `Error::other` variant already says it in this case.

firewood/src/manager.rs
```diff
@@ -145,21 +145,10 @@ impl RevisionManager {
     ///    It only contains the address of the nodes that are deleted, which should be very small.
     /// 3. Revision reaping. If more than the maximum number of revisions are kept in memory, the
     ///    oldest revision is reaped.
-    /// 4. Set last committed revision.
+    /// 4. Persist to disk. This includes flushing everything to disk.
+    /// 5. Set last committed revision.
```
qusuyan, line 149: Will the flushing overwrite the previously committed revision?

firewood/src/proof.rs
```diff
@@ -106,11 +106,11 @@ impl Hashable for ProofNode {
         })
     }
 
-    fn children(&self) -> impl Iterator<Item = (usize, &HashType)> + Clone {
+    fn children(&self) -> impl Iterator<Item = (usize, HashType)> + Clone {
         self.child_hashes
             .iter()
             .enumerate()
-            .filter_map(|(i, hash)| hash.as_ref().map(|h| (i, h)))
+            .filter_map(|(i, hash)| hash.as_ref().map(|h| (i, h.clone())))
```
qusuyan, line 114: Any particular reason we want to return the 32-byte hash instead of an 8-byte pointer?

storage/src/checker/mod.rs
```diff
@@ -42,19 +43,34 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+                if let Some(root_hash) = root_hash {
+                    self.visit_trie(
+                        root_address,
+                        root_hash,
+                        Path::new(),
+                        &mut visited,
+                        opt.hash_check,
+                    )?;
+                } else {
+                    // No root hash available, visit without hash checking
```
qusuyan: Why would there ever be a trie without root hash? I think it is fine to compute the root hash before traversing for completeness, but we can also change this parameter to `Option<HashType>` so that we still do hash checks on children nodes.

firewood/src/proof.rs
qusuyan: Nvm I see it is a trait implementation

storage/src/linear/filebacked.rs
```diff
@@ -212,13 +212,13 @@ impl WritableStorage for FileBacked {
         }
     }
 
-    fn write_cached_nodes<'a>(
+    fn write_cached_nodes(
         &self,
-        nodes: impl Iterator<Item = (LinearAddress, &'a SharedNode)>,
+        nodes: impl Iterator<Item = (NonZeroU64, SharedNode)>,
```
qusuyan: Why are we changing back to `NonZeroU64` here? I thought the `LinearAddress` PR should get rid of all of them?

storage/src/nodestore/mod.rs
```diff
@@ -685,6 +613,83 @@ where
...
+    pub fn read_node_from_disk(
+        &self,
+        addr: LinearAddress,
+        mode: &'static str,
+    ) -> Result<SharedNode, FileIoError> {
+        if let Some(node) = self.storage.read_cached_node(addr, mode) {
+            return Ok(node);
+        }
+
+        debug_assert!(addr.get() % 8 == 0);
```
qusuyan: nit: use `is_aligned()` here

storage/src/nodestore/persist.rs
```diff
@@ -208,21 +206,71 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
+        let unpersisted_iter = UnPersistedNodeIterator::new(self);
+
+        let mut header = self.header;
+        let mut allocator = NodeAllocator::new(self.storage.as_ref(), &mut header);
+        for node in unpersisted_iter {
+            let shared_node = node.as_shared_node(self).expect("in memory, so no IO");
+            let mut serialized = Vec::new();
+            shared_node.as_bytes(0, &mut serialized);
+
+            let (persisted_address, _) = allocator.allocate_node(serialized.as_slice())?;
```
qusuyan: When do you write the `AreaIndex` in? I don't see it anywhere...

storage/src/checker/mod.rs
rkuris: It was less clean than this was I think.

firewood/src/manager.rs
rkuris: No, it will not. New nodes are allocated from the free list.

qusuyan (commented):
I think I need to go through the code again for better understanding...

storage/src/checker/mod.rs
rkuris: Yes, good point. This was way more complicated than it needs to be. We will always have a hash and a root node when the database is non-empty.

Currently, all `Committed` nodestores are also persisted, so I added an error if you attempt to run it on an unpersisted Committed nodestore.

This is likely to get cleaned up when we defer persisting nodestores.

storage/src/linear/filebacked.rs
rkuris: Another good catch. This could be a lot simpler.

storage/src/nodestore/persist.rs
rkuris: Another good catch. This was a bug, which should have been caught by the checker. Is it?

storage/src/nodestore/mod.rs
```diff
@@ -685,6 +613,83 @@ where
...
+    pub fn read_node_from_disk(
+        &self,
+        addr: LinearAddress,
+        mode: &'static str,
+    ) -> Result<SharedNode, FileIoError> {
+        if let Some(node) = self.storage.read_cached_node(addr, mode) {
+            return Ok(node);
+        }
+
+        debug_assert!(addr.is_aligned());
```
demosdemon, line 632: really unrelated to your change, but since I saw it elsewhere:

this panics when branch factor 256 is enabled. obv low priority but I didn't linger enough to figure out why.

storage/src/nodestore/persist.rs
```diff
@@ -186,7 +184,7 @@ impl<N: NodeReader + RootReader> Iterator for UnPersistedNodeIterator<'_, N> {
     }
 }
 
-impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
+impl NodeStore<Committed, FileBacked> {
```
demosdemon, line 187: Since you're working on this, I need a method that is generic over the storage types so I can use it with the memstore. In my branch, I did this:

```rust

impl<S: WritableStorage> NodeStore<Arc<ImmutableProposal>, S> {
    /// Persist all the nodes of a proposal to storage.
    ///
    /// # Errors
    ///
    /// Returns a [`FileIoError`] if any node cannot be written to storage.
    fn flush_nodes_generic(&self) -> Result<(), FileIoError> {
        let flush_start = Instant::now();

        for (addr, (area_size_index, node)) in &self.kind.new {
            let mut stored_area_bytes = Vec::new();
            node.as_bytes(*area_size_index, &mut stored_area_bytes);
            self.storage
                .write(addr.get(), stored_area_bytes.as_slice())?;
        }

        self.storage
            .write_cached_nodes(self.kind.new.iter().map(|(&addr, (_, node))| (addr, node)))?;
        let flush_time = flush_start.elapsed().as_millis();
        counter!("firewood.flush_nodes").increment(flush_time);

        Ok(())
    }
}

impl NodeStore<Arc<ImmutableProposal>, MemStore> {
    /// Persist all the nodes of a proposal to storage.
    ///
    /// # Errors
    ///
    /// Returns a [`FileIoError`] if any node cannot be written to storage.
    #[fastrace::trace(short_name = true)]
    pub fn flush_nodes(&self) -> Result<(), FileIoError> {
        self.flush_nodes_generic()
    }
}

impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
    /// Persist all the nodes of a proposal to storage.
    ///
    /// # Errors
    ///
    /// Returns a [`FileIoError`] if any node cannot be written to storage.
    #[fastrace::trace(short_name = true)]
    #[cfg(not(feature = "io-uring"))]
    pub fn flush_nodes(&self) -> Result<(), FileIoError> {
        self.flush_nodes_generic()
    }
}
```

Same for `as_committed`:

```rust

impl<S: WritableStorage> NodeStore<Arc<ImmutableProposal>, S> {
    /// Return a Committed version of this proposal, which doesn't have any modified nodes.
    /// This function is used during commit.
    #[must_use]
    pub fn as_committed(&self) -> NodeStore<Committed, S> {
        NodeStore {
            header: self.header,
            kind: Committed {
                deleted: self.kind.deleted.clone(),
                root_hash: self.kind.root_hash.clone(),
                root: self.kind.root.clone(),
            },
            storage: self.storage.clone(),
        }
    }
}
```

But, you might have a better option.

Having these be implemented only for `FileStore` does not make sense in a lot of places.

---

It's also a bit of violating separation of concerns to have the io-uring code here and not in inside the file store module.

storage/src/nodestore/persist.rs
```diff
@@ -281,6 +330,18 @@ impl NodeStore<Arc<ImmutableProposal>, FileBacked> {
...
         let flush_start = Instant::now();
 
+        let mut header = self.header;
+        let mut node_allocator = NodeAllocator::new(self.storage.as_ref(), &mut header);
+
+        // Collect all unpersisted nodes first to avoid mutating self while iterating
+        let unpersisted_nodes: Vec<MaybePersistedNode> = {
+            let unpersisted_iter = UnPersistedNodeIterator::new(self);
+            unpersisted_iter.collect()
+        };
```
demosdemon, line 340: bit weird to do this in two lines.
```suggestion
        // Collect all unpersisted nodes first to avoid mutating self while iterating
        let unpersisted_nodes = UnPersistedNodeIterator::new(self).collect::<Vec<_>>();
```

demosdemon (approved):

PR #1062, opened 2025-07-15
Generated newtypes for LinearAddress.

Salonii02:
Refactor : Use LinearAddress newtype for trie addresses.

- Replaced all usages of the NonZeroU64 type alias for addresses with a LinearAddress newtype struct, enabling method implementations and improved type safety.
- Updated all code, tests, and benchmarks to construct and use LinearAddress where an address is required.
- Added accessor methods to LinearAddress for ergonomic and type-safe access to the inner value.
- Switched serialization and deserialization of nodestore headers and freelists from bytemuck to bincode/serde, as the newtype and option types are not Pod.
- Removed all bytemuck derives and usages for non-Pod types.
- Updated all relevant code paths to use bincode for reading and writing headers and freelists.

### DIFF
--- a/storage/src/linear/filebacked.rs
+++ b/storage/src/linear/filebacked.rs
@@ -218,7 +218,7 @@ impl WritableStorage for FileBacked {
     ) -> Result<(), FileIoError> {
         let mut guard = self.cache.lock().expect("poisoned lock");
         for (addr, node) in nodes {
-            guard.put(*addr, node.clone());
+            guard.put(LinearAddress::new((*addr).get()).unwrap(), node.clone());
         }
         Ok(())
     }
--- a/storage/src/node/persist.rs
+++ b/storage/src/node/persist.rs
@@ -160,6 +160,7 @@ mod test {
             Option::from(&maybe_persisted_node)
         );
 
+        let addr = LinearAddress::new(addr.get()).unwrap();
         maybe_persisted_node.persist_at(addr);
         assert!(maybe_persisted_node.as_shared_node(&store).is_err());
         assert_eq!(Some(addr), Option::from(&maybe_persisted_node));
@@ -168,7 +169,8 @@ mod test {
 
     #[test]
     fn test_from_linear_address() {
-        let addr = nonzero!(1024u64);
+        let raw_addr = nonzero!(1024u64);
+        let addr = LinearAddress::new(raw_addr.get()).unwrap();
         let maybe_persisted_node = MaybePersistedNode::from(addr);
         assert_eq!(Some(addr), Option::from(&maybe_persisted_node));
     }
@@ -196,6 +198,7 @@ mod test {
 
         // Persist the original
         let addr = nonzero!(4096u64);
+        let addr = LinearAddress::new(addr.get()).unwrap();
         original.persist_at(addr);
 
         // Both original and clone should now be persisted since they share the same ArcSwap
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -30,6 +30,7 @@ use std::io::{Error, ErrorKind, Read};
 use std::iter::FusedIterator;
 use std::num::NonZeroU64;
 use std::sync::Arc;
+use std::fmt;
 
 use crate::node::persist::MaybePersistedNode;
 use crate::node::{ByteCounter, Node};
@@ -111,9 +112,9 @@ pub const fn index_name(index: usize) -> &'static str {
 /// This is not usize because we can store this as a single byte
 pub type AreaIndex = u8;
 
-pub const NUM_AREA_SIZES: usize = AREA_SIZES.len();
-pub const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
-pub const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
+const NUM_AREA_SIZES: usize = AREA_SIZES.len();
+const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
+const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
 
 #[inline]
 pub fn new_area_index(n: usize) -> AreaIndex {
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
         })
 }
 
-/// Objects cannot be stored at the zero address, so a [`LinearAddress`] is guaranteed not
-/// to be zero. This reserved zero can be used as a [None] value for some use cases. In particular,
-/// branches can use `Option<LinearAddress>` which is the same size as a [`LinearAddress`]
-pub type LinearAddress = NonZeroU64;
+// pub type LinearAddress = NonZeroU64;
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[repr(transparent)]
+pub struct LinearAddress(NonZeroU64);
+
+impl LinearAddress {
+    /// Create a new LinearAddress, returns None if value is zero.
+    pub fn new(addr: u64) -> Option<Self> {
+        NonZeroU64::new(addr).map(LinearAddress)
+    }
+
+    /// Get the underlying address as u64.
+    pub fn get(&self) -> u64 {
+        self.0.get()
+    }
+
+    /// Check if the address is 8-byte aligned.
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
+    }
+
+    pub fn checked_add(&self, offset: u64) -> Option<Self> {
+        self.get().checked_add(offset).and_then(LinearAddress::new)
+    }
+
+    pub fn max_area_size() -> u64 {
+        AREA_SIZES[AREA_SIZES.len() - 1]
+    }
+    pub fn min_area_size() -> u64 {
+        AREA_SIZES[0]
+    }
+    pub fn num_area_sizes() -> usize {
+        AREA_SIZES.len()
+    }
+
+    /// Returns a reference to the inner NonZeroU64
+    pub fn as_nonzero(&self) -> &NonZeroU64 {
+        &self.0
+    }
+}
+
+impl fmt::Display for LinearAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.get())
+    }
+}
+
+impl From<LinearAddress> for u64 {
+    fn from(addr: LinearAddress) -> Self {
+        addr.get()
+    }
+}
 
 /// Each [`StoredArea`] contains an [Area] which is either a [Node] or a [`FreeArea`].
 #[repr(u8)]
@@ -187,7 +236,8 @@ impl<T> StoredArea<T> {
     }
 }
 
-pub type FreeLists = [Option<LinearAddress>; NUM_AREA_SIZES];
+/// FreeLists is an array of Option<LinearAddress> for each area size.
+pub type FreeLists = [Option<LinearAddress>; AREA_SIZES.len()];
 
 /// A [`FreeArea`] is stored at the start of the area that contained a node that
 /// has been freed.
@@ -289,7 +339,7 @@ impl<T: ReadInMemoryNode, S: ReadableStorage> NodeStore<T, S> {
             return Ok(node);
         }
 
-        debug_assert!(addr.get() % 8 == 0);
+        debug_assert!(addr.is_aligned());
 
         // saturating because there is no way we can be reading at u64::MAX
         // and this will fail very soon afterwards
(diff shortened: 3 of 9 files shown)


storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+
+    /// Check if the address is 8-byte aligned.
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
+    }
+
+    pub fn checked_add(&self, offset: u64) -> Option<Self> {
+        self.get().checked_add(offset).and_then(LinearAddress::new)
+    }
+
```
rkuris, line 173: Move to `impl CheckedAdd`
https://docs.rs/num/latest/num/trait.CheckedAdd.html

storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    pub fn max_area_size() -> u64 {
+        AREA_SIZES[AREA_SIZES.len() - 1]
+    }
+    pub fn min_area_size() -> u64 {
+        AREA_SIZES[0]
+    }
+    pub fn num_area_sizes() -> usize {
+        AREA_SIZES.len()
+    }
```
rkuris, line 182: These don't have anything to do with LinearAddress so they shouldn't be methods on LinearAddress. We probably need another newtype for AREA_SIZES, but I like that this PR is only about LinearAddresses, so let's move these back and do a different PR for AREA_SIZES.

storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+
+    /// Returns a reference to the inner NonZeroU64
+    pub fn as_nonzero(&self) -> &NonZeroU64 {
+        &self.0
+    }
```
rkuris, line 187: `impl Deref`
https://doc.rust-lang.org/std/ops/trait.Deref.html

storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+
+    /// Get the underlying address as u64.
+    pub fn get(&self) -> u64 {
+        self.0.get()
+    }
+
+    /// Check if the address is 8-byte aligned.
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
+    }
```
rkuris, line 168: This probably belongs in AREA_SIZES as well. It's the only method that is an intersection between the two.

Aligned means it is a multiple of `AREA_SIZES.first()`, not 8.

storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+
+    pub fn checked_add(&self, offset: u64) -> Option<Self> {
+        self.get().checked_add(offset).and_then(LinearAddress::new)
+    }
+
+    pub fn max_area_size() -> u64 {
+        AREA_SIZES[AREA_SIZES.len() - 1]
+    }
+    pub fn min_area_size() -> u64 {
+        AREA_SIZES[0]
```
rkuris, line 178: Not sure why this didn't hit the linter. We generally prohibit index dereferencing, and prefer the more explicit version of this.
```suggestion
        AREA_SIZES.first().expect("always one area size")
```

storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
+    }
+
+    pub fn checked_add(&self, offset: u64) -> Option<Self> {
+        self.get().checked_add(offset).and_then(LinearAddress::new)
+    }
+
+    pub fn max_area_size() -> u64 {
+        AREA_SIZES[AREA_SIZES.len() - 1]
```
rkuris, line 175:
```suggestion
        AREA_SIZES.last().expect("always one AREA_SIZE")
```
See next comment

storage/src/nodestore/header.rs
```diff
@@ -293,6 +294,16 @@ impl NodeStoreHeader {
             ))
         }
     }
+
+    /// Helper to get free list as LinearAddress
+    pub fn free_list_linear(&self, idx: usize) -> Option<LinearAddress> {
+        self.free_lists[idx]
```
rkuris, line 300: This could panic. Maybe something like:

```suggestion
        self.free_lists.get(idx).flatten()
```

storage/src/nodestore/header.rs
```diff
@@ -293,6 +294,16 @@ impl NodeStoreHeader {
...
+    }
+
+    /// Helper to set free list as LinearAddress
+    pub fn set_free_list_linear(&mut self, idx: usize, addr: Option<LinearAddress>) {
+        self.free_lists[idx] = addr;
```
rkuris, line 305: Also could panic.
```suggestion
        *self.free_lists.get_mut(idx).expect("invalid index") = addr;
```

storage/src/nodestore/header.rs
```diff
@@ -325,12 +336,12 @@ mod tests {
...
-        let mut header = NodeStoreHeader::new();
         let mut header_stream = node_store.storage.stream_from(0).unwrap();
-        let header_bytes = bytemuck::bytes_of_mut(&mut header);
-        header_stream.read_exact(header_bytes).unwrap();
+        let mut header_bytes = vec![0u8; bincode::serialized_size(&NodeStoreHeader::new()).unwrap() as usize];
+        header_stream.read_exact(&mut header_bytes).unwrap();
+        let header: NodeStoreHeader = bincode::deserialize(&header_bytes).unwrap();
         assert_eq!(header.version, Version::new());
-        let empty_free_list: super::FreeLists = Default::default();
+        let empty_free_list: FreeLists = Default::default();
```
rkuris, line 344: We are moving away from serde, so we should stick with bytemuck.

storage/src/nodestore/mod.rs
```diff
@@ -101,12 +101,16 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
+        stream.read_exact(&mut header_bytes)
             .map_err(|e| storage.file_io_error(e, 0, Some("header read".to_string())))?;
-
+       
+        let header: NodeStoreHeader = bincode::deserialize(&header_bytes)
+            .map_err(|e| storage.file_io_error(
+                std::io::Error::new(std::io::ErrorKind::InvalidData, e),
+                0,
+                Some("header read".to_string())
+            ))?;
```
rkuris, line 113: Same as above. We want to eliminate the serde dependency at this layer.

storage/src/nodestore/mod.rs
```diff
@@ -273,9 +277,9 @@ impl<S: WritableStorage> NodeStore<MutableProposal, S> {
     /// Panics if the header cannot be written.
     pub fn new_empty_proposal(storage: Arc<S>) -> Self {
         let header = NodeStoreHeader::new();
-        let header_bytes = bytemuck::bytes_of(&header);
+        let header_bytes = bincode::serialize(&header).expect("failed to serialize header");
```
rkuris, line 280: Same

storage/src/linear/filebacked.rs
```diff
@@ -218,7 +218,7 @@ impl WritableStorage for FileBacked {
     ) -> Result<(), FileIoError> {
         let mut guard = self.cache.lock().expect("poisoned lock");
         for (addr, node) in nodes {
-            guard.put(*addr, node.clone());
+            guard.put(LinearAddress::new((*addr).get()).unwrap(), node.clone());
```
rkuris, line 221: Not sure why this got complicated, we're just moving the addr, and it implements Copy.

storage/src/nodestore/alloc.rs
Salonii02: For third point remove the bytemuck->serde changes : 
Explored several ways to keep using bytemuck (including newtype wrappers and manual trait impls), but ran into trait errors (Pod, NoUninit, AnyBitPattern) due to NonZeroU64/LinearAddress. Making it work would have required unsafe code or major structural changes, so switched to serde/bincode for safe serialization, updating all related logic and tests.
If you have any alternative ideas or know a way to safely use zero-copy with these types, I’d love to hear your suggestions!

storage/src/nodestore/alloc.rs
qusuyan: moved to #1046

storage/src/node/persist.rs
```diff
@@ -168,7 +169,8 @@ mod test {
 
     #[test]
     fn test_from_linear_address() {
-        let addr = nonzero!(1024u64);
+        let raw_addr = nonzero!(1024u64);
+        let addr = LinearAddress::new(raw_addr.get()).unwrap();
```
qusuyan, line 173: this should just be `LinearAddress::new(1024u64)`

storage/src/nodestore/alloc.rs
qusuyan: why can't this just be 
```suggestion
        MAX_AREA_SIZE
```

storage/src/nodestore/alloc.rs
qusuyan: If things were working with `LinearAddress = NonZeroU64`, `impl Deref` may resolve most of the issues here.

rkuris: You'll need to add the following to allow these to be bytemuckable:
```
#[expect(unsafe_code)]
unsafe impl bytemuck::ZeroableInOption for LinearAddress {}
#[expect(unsafe_code)]
unsafe impl bytemuck::PodInOption for LinearAddress {}
```
Along with some explanation as to why these are okay.

storage/src/nodestore/alloc.rs
```diff
@@ -145,10 +146,58 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
         })
 }
 
-/// Objects cannot be stored at the zero address, so a [`LinearAddress`] is guaranteed not
-/// to be zero. This reserved zero can be used as a [None] value for some use cases. In particular,
-/// branches can use `Option<LinearAddress>` which is the same size as a [`LinearAddress`]
-pub type LinearAddress = NonZeroU64;
+// pub type LinearAddress = NonZeroU64;
```
rkuris, line 149: This type needs some documentation:
```suggestion
/// A newtype wrapper around an offset into the database file for a node or free list item.
/// Only the header is stored at location 0, so we use None as Option<LInearAddress> to
/// represent that a child is not present in a branch.
```

storage/src/nodestore/alloc.rs
rkuris: Thanks @qusuyan -- @Salonii02 please rebase off that PR

Apologies for the shuffling around, there is a lot of active development here!

storage/src/nodestore/alloc.rs
rkuris: Note that checked_add is in the num_traits crate, so you'll need to import that.

storage/src/nodestore/alloc.rs
rkuris: There's also MIN_AREA_SIZE

storage/src/nodestore/header.rs
```diff
@@ -224,22 +225,22 @@ impl NodeStoreHeader {
     }
 
     /// Get the free lists
-    pub const fn free_lists(&self) -> &FreeLists {
+    pub fn free_lists(&self) -> &FreeLists {
```
rkuris, line 228: Why were these removed? They didn't seem to cause any problems when I replaced them.

storage/src/nodestore/mod.rs
```diff
@@ -779,7 +783,7 @@ mod tests {
         value: Some(vec![9, 10, 11].into_boxed_slice()),
         children: from_fn(|i| {
             if i == 15 {
-                Some(Child::AddressWithHash(nonzero!(1u64), std::array::from_fn::<u8, 32, _>(|i| i as u8).into()))
+                Some(Child::AddressWithHash(LinearAddress::new(1).unwrap(), std::array::from_fn::<u8, 32, _>(|i| i as u8).into()))
```
rkuris, line 786: The nonzero! macro was really handy here, as it keeps the code focused on what we're doing. You can probably create a macro here for this, something like this:

```
macro_rules! linear_address {
    ($addr:expr) => {
        LinearAddress::new(nonzero!($addr))
    };
}
```
Then you end up with this much cleaner code at the callsite:

```suggestion
                Some(Child::AddressWithHash(linear_address!(1), std::array::from_fn::<u8, 32, _>(|i| i as u8).into()))
```

It might be a little trickier; you'll have to dig into the nonzero! macro to see exactly what types it expects.

storage/src/linear/filebacked.rs
rkuris: Would like this to work
```suggestion
            guard.put(guard.put(addr.into(), node.clone()););
```

To get this to work, you'd need to `impl From<NonZeroU64> for LinearAddress` and change the loop to `for (&addr, node)`.

rkuris:
With as much as we have going on in these sections of the code, it's going to be challenging to find a window where this code can merge cleanly. I'd suggest not rebasing until you address the above issues. I'm happy to help rebase it at that point.

demosdemon:
included as part of https://github.com/ava-labs/firewood/commit/f2876b291ed892b732bac270253ff1262b5b717e

PR #1063, opened 2025-07-15
feat(checker): add hash checks

qusuyan:


storage/src/checker.rs
```diff
@@ -170,11 +206,101 @@ mod test {
...
+            .visit_trie(root_addr, root_hash, Path::new(), &mut visited, true)
+            .unwrap();
         let complement = visited.complement();
         assert_eq!(complement.into_iter().collect::<Vec<_>>(), vec![]);
     }
 
+    #[test]
+    // This test creates a trie with a wrong hash and checks that the checker detects it.
+    // the trie structure is the same as in the previous test.
+    fn checker_traverse_trie_with_wrong_hash() {
```
rkuris: this is really painful. Can't we reuse another existing trie and just tweak one of the hashes? I'd like something more like this, which can be shared with other tests:

```suggestion
    fn checker_traverse_trie_with_wrong_hash() {
        let mut valid_nodestore = generate_valid_test_nodestore();
        let child = &mut valid_nodestore.mut_root().as_branch().children[1];
        // ...
        *child = Some(Child::AddressWithHash(leaf_addr, HashType::default()));
        // ...
    }
        
```

Even if this is the only test that uses this clean nodestore, it helps make this test more focused and readble.

fwdctl/src/check.rs
```diff
@@ -36,5 +45,7 @@ pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
         CacheReadStrategy::WritesOnly, // we scan the database once - no need to cache anything
     )?);
 
-    NodeStore::open(storage)?.check().map_err(Into::into)
+    NodeStore::open(storage)?
+        .check(opts.hash_check)
```
rkuris: nit: Generally better to send `opts` because then you can just reference `hash_check` instead of a named bool.

Things to read:
https://medium.com/web-tech-journals/why-you-should-avoid-pass-boolean-arguments-to-functions-and-what-to-do-instead-1288e23f1f8e
https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

I once had a reviewer who never allowed bools: shoutout to @klnusbaum

fwdctl/src/check.rs
```diff
@@ -21,6 +21,15 @@ pub struct Options {
         help = "Name of the database"
     )]
     pub db: String,
+
+    /// Whether to perform hash check
+    #[arg(
+        long,
+        required = false,
+        default_value_t = false,
```
rkuris, line 29: why not true? Is it much slower?

storage/src/checker.rs
```diff
@@ -4,12 +4,13 @@
...
 use crate::{
-    CheckerError, Committed, HashedNodeReader, LinearAddress, Node, NodeReader, NodeStore,
-    WritableStorage,
+    CheckerError, Committed, HashType, HashedNodeReader, LinearAddress, Node, NodeReader,
+    NodeStore, Path, WritableStorage, hash_node,
 };
 
 /// [`NodeStore`] checker
 // TODO: S needs to be writeable if we ask checker to fix the issues
+#[allow(clippy::result_large_err)]
```
rkuris: Use "expect" instead of allow since this is expected. If someone fixes it, they will then remove it.

nit: maybe box some stuff.

rkuris (changes requested):

fwdctl/src/check.rs
qusuyan: Haven't timed it yet, would be great if you can point me to a real sized image for benchmarking.

storage/src/checker.rs
qusuyan: I'm gonna box it in another diff - just don't want to put it here.

fwdctl/src/check.rs
qusuyan: I don't want to send `opts` into the storage crate, that is a real violation of separation of concerns. Any better alternatives?

rkuris: You need a CheckOpts. The alternative is two methods, but it's likely other options will pop up (such as repair options).

fwdctl/src/check.rs
rkuris: The fastest way to resolve this is to create one. Just use the `benchmark create` command to make one. There are specific instructions in the benchmark/README.md

rkuris (approved):
nothing but nits and followups.

PR #1064, opened 2025-07-15
feat!: auto open or create with truncate

### LINKED ISSUE #1039: Improve FFI db open simplicity
Right now, the caller decides whether a file should be "opened" or "created" on database open, which is typically inconvenient for the caller, who just expects "open if it exists, otherwise create". This would be more intuitive for FFI consumers.

alarso16:
It's cumbersome (and more prone to TOCTOU issues) for users to validate whether a file exists, and if it does, set a flag. This should be handled internally. The `truncate` config now more closely resembles it's actual meaning. If the file already exists, it will be wiped and re-created.

There are unit tests in Rust and in Go for this.

Close #1039

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -870,6 +870,7 @@ pub extern "C" fn fwd_gather() -> Value {
 /// * `revisions` - The maximum number of revisions to keep; firewood currently requires this to be at least 2.
 /// * `strategy` - The cache read strategy to use, 0 for writes only,
 ///   1 for branch reads, and 2 for all reads.
+/// * `truncate` - Whether to truncate the database file if it exists.
 ///   Returns an error if the value is not 0, 1, or 2.
 #[repr(C)]
 pub struct CreateOrOpenArgs {
@@ -878,29 +879,7 @@ pub struct CreateOrOpenArgs {
     free_list_cache_size: usize,
     revisions: usize,
     strategy: u8,
-}
-
-/// Create a database with the given cache size and maximum number of revisions, as well
-/// as a specific cache strategy
-///
-/// # Arguments
-///
-/// See `CreateOrOpenArgs`.
-///
-/// # Returns
-///
-/// A database handle, or panics if it cannot be created
-///
-/// # Safety
-///
-/// This function uses raw pointers so it is unsafe.
-/// It is the caller's responsibility to ensure that path is a valid pointer to a null-terminated string.
-/// The caller must also ensure that the cache size is greater than 0 and that the number of revisions is at least 2.
-/// The caller must call `close` to free the memory associated with the returned database handle.
-///
-#[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_create_db(args: CreateOrOpenArgs) -> DatabaseCreationResult {
-    unsafe { common_create(&args, true) }.into()
+    truncate: bool,
 }
 
 /// Open a database with the given cache size and maximum number of revisions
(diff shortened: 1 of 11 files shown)


firewood/src/db.rs
```diff
@@ -543,11 +543,15 @@ mod test {
         proposal.commit().await.unwrap();
         println!("{:?}", db.root_hash().await.unwrap().unwrap());
 
-        let db = db.reopen().await;
+        let db = db.reopen(false).await;
```
rkuris: nit: this is test code so a nit, but I hate unadorned bools, since the reader has no idea what this is doing.

firewood/src/manager.rs
```diff
@@ -84,14 +84,15 @@ impl RevisionManager {
         truncate: bool,
         config: RevisionManagerConfig,
     ) -> Result<Self, FileIoError> {
-        let storage = Arc::new(FileBacked::new(
+        let (fb, created) = FileBacked::new(
```
rkuris: Rather than indicating whether or not it was created from new here, what about if we just handle the EOF read from `NodeStore::open()` and call `new_empty_committed` from `NodeStore::open` if the header got an EOF because the file was empty?

storage/src/linear/filebacked.rs
```diff
@@ -92,19 +92,39 @@ impl FileBacked {
...
+            }
+            Err(e) => {
+                return Err(FileIoError {
+                    inner: e,
+                    filename: Some(path.clone()),
+                    offset: 0,
+                    context: Some("file open".to_string()),
+                });
+            }
+        };
```
rkuris: I don't love opening the file twice. I think you should just open it and return an empty file as the clue that things needs initialization.

storage/src/linear/filebacked.rs
```diff
@@ -92,19 +92,39 @@ impl FileBacked {
...
-            .open(&path)
-            .map_err(|inner| FileIoError {
-                inner,
-                filename: Some(path.clone()),
-                offset: 0,
-                context: Some("file open".to_string()),
-            })?;
+            .create_new(true)
+            .open(&path);
+        let fd = match fd {
```
rkuris: `created` temporarily has a possibly-incorrect value (true) above, instead do something like this:
```suggestion
        let (fd, created) = match fd {
```

This is common in golang but often leads to bugs, especially on early returns.

firewood/src/db.rs
alarso16: Split into several functions. Not sure if this is better, but definitely fixes your issue

ffi/src/lib.rs
```diff
@@ -922,14 +901,14 @@ pub unsafe extern "C" fn fwd_create_db(args: CreateOrOpenArgs) -> DatabaseCreati
...
 pub unsafe extern "C" fn fwd_open_db(args: CreateOrOpenArgs) -> DatabaseCreationResult {
-    unsafe { common_create(&args, false) }.into()
+    unsafe { create(&args) }.into()
 }
 
-/// Internal call for `fwd_create_db` and `fwd_open_db` to remove error handling from the C API
+/// Internal call for `fwd_open_db` to remove error handling from the C API
 #[doc(hidden)]
-unsafe fn common_create(args: &CreateOrOpenArgs, create_file: bool) -> Result<Db, String> {
+unsafe fn create(args: &CreateOrOpenArgs) -> Result<Db, String> {
```
rkuris: naming: open_db?

firewood/src/db.rs
```diff
@@ -729,9 +734,15 @@ mod test {
                 .collect()
         }
         async fn reopen(self) -> Self {
+            self.reopen_truncate(false).await
+        }
+        async fn replace(self) -> Self {
+            self.reopen_truncate(true).await
+        }
+        async fn reopen_truncate(self, truncate: bool) -> Self {
```
rkuris: nit: there's still an unadorned bool, but it's trivial to remove it now.

It's still just a nit so feel free to ignore.
```suggestion
            self.reopen_with_config(DbConfig::default()).await
        }
        async fn replace(self) -> Self {
              self.reopen_with_config(DbConfig::builder().truncate(true).build()).await
        }
        async fn reopen_with_config(self, dbconfig: DbConfig) -> Self {
```

rkuris (approved):
Nothing but naming suggestions and nits.

Issue #1057, opened 2025-07-12
Serde derived serializer of `Area::Free` does not write 0xff for a free node


demosdemon:
The serde derived serializer for `Area<T, U>` does not serialize `0xff` for the free node.

See the auto-generated code when running `cargo expand` (cargo install cargo-expand first):

https://gist.github.com/demosdemon/255715a6e32d978cb2ffc02658d5a5a5

On deserializing, it matches on the literal 0 and 1 for node and free respectively:
https://gist.github.com/demosdemon/255715a6e32d978cb2ffc02658d5a5a5#file-expand-rs-L55-L66

On serializing, it writes the literal 0 or 1 for node and free respectively:
https://gist.github.com/demosdemon/255715a6e32d978cb2ffc02658d5a5a5#file-expand-rs-L205-L222

See the bincode serializer for its behavior:
https://github.com/bincode-org/bincode/blob/v1.3.3/src/ser/mod.rs#L206-L218

bincode writes the literal value of `variant_index` as provided by the serde serializer which is 0u32 and 1u32 for node and free respectively.

The following code confirms this:

```rust
fn main() {
    let node = StoredArea::new(
        12,
        Area::<Node, _>::Free(FreeArea::new(LinearAddress::new(42))),
    );
    let mut buf = Vec::new();
    DefaultOptions::new()
        .with_fixint_encoding()
        .serialize_into(&mut buf, &node)
        .unwrap();
    println!("Serialized StoredArea: {buf:#04x?}");

    buf.clear();
    DefaultOptions::new()
        .with_varint_encoding()
        .serialize_into(&mut buf, &node)
        .unwrap();
    println!("VarInt Serialized StoredArea: {buf:#04x?}");
}
```

```text
Serialized StoredArea: [
    0x0c, // 12 -- area_index
    0x01, // LE u32 begin - note the literal 0x01
    0x00,
    0x00,
    0x00, // LE u32 end
    0x01, // Indicates `Some(_)` from `Option<LinearAddress>` because 
          // serde does not handle the niche optimization of `Option<NonZero<_>>`
    0x2a, // LinearAddress(LE u64) start
    0x00,
    0x00,
    0x00,
    0x00,
    0x00,
    0x00,
    0x00, // LE u64 end
]
VarInt Serialized StoredArea: [
    0x0c, // 12
    0x01, // 1
    0x01, // 1
    0x2a, // 42
]
```

demosdemon:
[...]

Issue #1056, opened 2025-07-12
Node deserialization fails when the last seven bits of the partial path length is all ones (when `(pp_len & 0x7F) == 0x7F`)


demosdemon:
https://github.com/ava-labs/firewood/blob/ee5d1d8ee292ca346fe6ceb33af5f4149cb8a317/storage/src/node/mod.rs#L326

https://github.com/ava-labs/firewood/blob/ee5d1d8ee292ca346fe6ceb33af5f4149cb8a317/storage/src/node/mod.rs#L355-L358

The `set_partial_path_len` parameter to `LeafFirstByte::new` is `l.partial_path.0.len() as u8`. This is equivalent to `(l.partial_path.0.len() & 0x7f) as u8`. Because first we truncate the integer to only a u8 and then `bitflags` will additionally truncate the value to 7 bits.

The below output is a verification of this:

```rust
fn main() {
    for b in 0..=u8::MAX {
        let first_byte = LeafFirstByte::new(1, b);
        let b = first_byte.0;
        let is_leaf = first_byte.is_leaf();
        let pp_len = first_byte.partial_path_length();
        println!(
            "{b:#04x} ({b:#010b}): is_leaf={is_leaf:01b}, partial_path_length={pp_len:#09b} ({pp_len:#04x})",
        );
    }
}
```

```text
0x01 (0b00000001): is_leaf=1, partial_path_length=0b0000000 (0x00)
0x03 (0b00000011): is_leaf=1, partial_path_length=0b0000001 (0x01)
0x05 (0b00000101): is_leaf=1, partial_path_length=0b0000010 (0x02)
0x07 (0b00000111): is_leaf=1, partial_path_length=0b0000011 (0x03)
0x09 (0b00001001): is_leaf=1, partial_path_length=0b0000100 (0x04)
0x0b (0b00001011): is_leaf=1, partial_path_length=0b0000101 (0x05)
0x0d (0b00001101): is_leaf=1, partial_path_length=0b0000110 (0x06)
0x0f (0b00001111): is_leaf=1, partial_path_length=0b0000111 (0x07)
0x11 (0b00010001): is_leaf=1, partial_path_length=0b0001000 (0x08)
0x13 (0b00010011): is_leaf=1, partial_path_length=0b0001001 (0x09)
0x15 (0b00010101): is_leaf=1, partial_path_length=0b0001010 (0x0a)
0x17 (0b00010111): is_leaf=1, partial_path_length=0b0001011 (0x0b)
[...]

PR #1066, opened 2025-07-15
fix: remove dependency on serde

demosdemon:
The way we used serde when serializing free areas but not other nodes was inconsistent and caused an issue were we were expecting data to be encoded one way but in actuality it was encoded another way.

This commit removes the dependency on serde for serialization in the storage crate, specifically in the `FreeArea` serialization. This ensures that all serialization is done manually, which is consistent and avoids the issues we were facing with serde's encoding.

Fixes: #831 #1057


demosdemon:
This PR was stripped down to the minimum.

Fix for #1056 will be in a separate PR and an additional PR as well for other experimental serialization changes.

storage/src/nodestore/alloc.rs
```diff
@@ -150,52 +146,107 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    ///     0x00,
+    ///     0x00,
+    ///     0x00,
+    ///     0x00,
+    ///     0x00,
+    ///     0x00, // LE u64 end
+    /// ]
+    /// ```
+    fn from_reader<R: Read>(mut reader: R) -> std::io::Result<Self> {
+        match reader.read_byte()? {
```
rkuris, line 211: I suspect you will get much better performance if you do some buffering here. These translate directly into system calls read() with length 1.

Perhaps read 12 bytes, which is guaranteed to be safe, and then iterate over those, rather than repeatedly asking the reader for more data.

I am fairly sure this will be substantially faster, and is impossible to spread across multiple disk blocks so it should be almost as fast as reading 1 byte.

An alternative would be to set up a BufferedReader but I think the overhead there is too high.

storage/src/nodestore/alloc.rs
```diff
@@ -490,19 +542,9 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
-        let stored_area = StoredArea::new(area_size_index, area);
-
-        let stored_area_bytes = serializer().serialize(&stored_area).map_err(|e| {
-            self.storage.file_io_error(
-                Error::new(ErrorKind::InvalidData, e),
-                addr.get(),
-                Some("delete_node".to_string()),
-            )
-        })?;
+        let mut stored_area_bytes = Vec::with_capacity(16);
```
rkuris: Can we use `std::mem::size_of<T>` here instead of magic 16?

If not, then let's make it a `const` somewhere to call out what 16 represents.

storage/src/nodestore/alloc.rs
```diff
@@ -626,9 +667,8 @@ pub mod test_utils {
         area_size_index: AreaIndex,
         offset: u64,
     ) {
-        let area: Area<Node, FreeArea> = Area::Free(FreeArea::new(next_free_block));
-        let stored_area = StoredArea::new(area_size_index, area);
-        let stored_area_bytes = serializer().serialize(&stored_area).unwrap();
+        let mut stored_area_bytes = Vec::with_capacity(16);
```
rkuris: nit: for a test, I'd just go with `vec![]` but maybe this is good practice :)

storage/src/nodestore/alloc.rs
```diff
@@ -666,12 +706,49 @@ mod test_free_list_iterator {
...
+        FreeArea::new(next_free_block).as_bytes(area_size_index, &mut stored_area_bytes);
         storage.write(offset, &stored_area_bytes).unwrap();
     }
 
+    #[test]
+    fn test_free_list_decoding_old_format() {
```
rkuris: #[test_case] crate is really nice here:
```suggestion
    #[test_case(&[0x01, 0x01, 0x01, 0x2a], 1, 42), "old format"])
    #[test_case(&[0x02, 0x01, 0x00], 2, 0), "none"])
    #[test_case(&[0x03, 0xff, 0x2b], 3, 43, "new format"]) 
    fn test_free_list_decoding_old_format(data: &[u8], area: u8, addr: u64) {
```

Then, in the function, I'd consider:
 - Calling from_reader and as_bytes and verifying you get the same byte array
 - verify that the area_size matches
 - verify that linear address matches

Probably some other options are possible too.

storage/src/nodestore/alloc.rs
```diff
@@ -666,12 +706,49 @@ mod test_free_list_iterator {
...
+        let mut reader = &data[..];
+
+        let area_index = reader.read_byte().unwrap();
+        assert_eq!(area_index, 12);
+
+        let free_area: FreeArea = reader.next_value().unwrap();
+        assert_eq!(
+            free_area.next_free_block(),
+            Some(LinearAddress::new(42).unwrap())
+        );
```
rkuris: This seems like this reimplementing the code in `from_reader`. Is there something you're testing here that won't get caught by running this through `from_reader`?

rkuris (changes requested):
Overall looks good, but there is likely to be a perf problem when deserializing the free list items. I don' t know if this worked correctly before; it seems like we need some timings to verify that we're as fast or faster than before.

Alternatively, let's create a ticket to perf test this against the old implementation. I would prefer not to find this slower when loading a large database as there are too many other variables.

demosdemon:
> I don' t know if this worked correctly before; it seems like we need some timings to verify that we're as fast or faster than before.

This is no different than before, which performed short reads of exactly one byte and other small chunks and is also what you did for `node`:

https://github.com/ava-labs/firewood/blob/da2c55dff172dc142c7d1d39c127f8310298fe5b/storage/src/node/mod.rs#L352-L353

storage/src/nodestore/alloc.rs
demosdemon: 16 is not significant. I can make a constant for this and size it correctly. It's really `size_of<u64>() + size_of<u8>()`.

storage/src/nodestore/alloc.rs
demosdemon: The bytes includes the extra `area_index` byte which is not parsed in `FreeArea::from_reader` but is parsed in `FreeArea::from_storage`. 

I will clarify.

storage/src/nodestore/alloc.rs
demosdemon: If these should use buffered readers, that should really but a problem the storage provider solves and is out of scope for this change. Also, as mentioned, none of the other deserialization logic uses buffered readers so doing so here would be also out of scope.

storage/src/nodestore/alloc.rs
demosdemon: > Probably some other options are possible too

like what?

> Calling from_reader and as_bytes and verifying you get the same byte array

That won't work for this as we're testing deserializing the old format. won't get the same result. And, the other test in this file already does roundtrip serialization of the free node.

I can switch to using test_case for the test verifying the parsed area_size and linear address though.

storage/src/nodestore/alloc.rs
rkuris: I think we can just do one read and then create a cursor on that as a reader.
The alternative is to write some benchmarks. I just want to avoid debugging this when we're loading a big db.

storage/src/nodestore/alloc.rs
```diff
@@ -263,32 +268,29 @@ impl FreeArea {
...
-        encoded.push(prefix);
+    pub fn as_bytes<T: ExtendableBytes>(self, area_index: AreaIndex, encoded: &mut T) {
+        const RESERVE_SIZE: usize = size_of::<u8>() + var_int_max_size::<u64>();
+
+        encoded.reserve(RESERVE_SIZE);
+        encoded.push(area_index);
         self.write_to(encoded);
     }
+
+    fn from_storage_reader(mut reader: impl Read) -> std::io::Result<(Self, AreaIndex)> {
```
rkuris, line 289: nit: In most places in the code, this tuple is `(AreaIndex, LinearAddress)`. I wish we had a type, maybe FreeListEntry. @qusuyan runs into this a lot too.

demosdemon: This matches the `from_storage` return value right above this method which was unchanged from before. We do have a type for this, `StoredArea`, but it's now unused and removed in this PR. Changing the return type of `from_storage` is fine but would touch a lot of irrelevant things for this change.

storage/src/nodestore/alloc.rs
demosdemon: As I pointed out, this is no different that what `Node` is doing. Doing it differently for FreeArea but not the others *in this PR* does not make sense.

https://github.com/ava-labs/firewood/blob/da2c55dff172dc142c7d1d39c127f8310298fe5b/storage/src/node/mod.rs#L351-L382

rkuris: Agreed. I ran some tests and it's a tad faster now anyway. Serde isn't free.

rkuris (approved):

PR #1070, opened 2025-07-16
feat: export logs

RodrigoVillar:


ffi/firewood.go
```diff
@@ -108,11 +111,16 @@ func New(filePath string, conf *Config) (*Database, error) {
 		cache_size:           C.size_t(conf.NodeCacheEntries),
 		free_list_cache_size: C.size_t(conf.FreeListCacheEntries),
 		revisions:            C.size_t(conf.Revisions),
+		enable_logs:          C.bool(conf.EnableLogs),
+		logs_dir:             C.CString(conf.LogsDir),
+		filter_level:         C.CString(conf.FilterLevel),
```
RodrigoVillar: Q: should these be grouped in a subconfig within `Config`?

ffi/src/lib.rs
```diff
@@ -937,9 +945,35 @@ unsafe fn common_create(args: &CreateOrOpenArgs, create_file: bool) -> Result<Db
...
         )?)
         .build();
-    #[cfg(feature = "logger")]
-    let _ = env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
-        .try_init();
+
+    if args.enable_logs {
+        let logs_dir = unsafe { CStr::from_ptr(args.logs_dir) }
+            .to_str()
+            .map(|v| if v.is_empty() { "/tmp/log" } else { v })
```
RodrigoVillar: Q: is setting a default value this way idiomatic in Rust?

ffi/Cargo.toml
```diff
@@ -20,19 +20,19 @@ rust-version.workspace = true
...
+log = "0.4.27"
 coarsetime = { workspace = true }
-env_logger = { workspace = true, optional = true}
+env_logger = { workspace = true}
 
 [target.'cfg(unix)'.dependencies]
 tikv-jemallocator = "0.6.0"
 
 [features]
-logger = ["dep:env_logger", "firewood/logger"]
```
RodrigoVillar: I removed the `logger` feature as I thought that Go clients of Firewood should be able to customize logging via config values rather than build Firewood with the `logger` feature. However,  Go clients shouldn't have to worry about this since `firewood-go-ethhash` should enable `logging` by default?

RodrigoVillar (commented):
Questions for reviewers, any reply would be greatly appreciated!

ffi/firewood.go
alarso16: I'm not sure. If there's a sub-config, then we may not need the `enable_logs` flag, since you could just not set the sub-config and it should be obvious (unless you don't use a pointer?). I'm not sure what's idiomatic here

ffi/firewood.go
```diff
@@ -108,11 +111,16 @@ func New(filePath string, conf *Config) (*Database, error) {
...
 		strategy:             C.uint8_t(conf.ReadCacheStrategy),
 	}
 	// Defer freeing the C string allocated to the heap on the other side
 	// of the FFI boundary.
 	defer C.free(unsafe.Pointer(args.path))
+	defer C.free(unsafe.Pointer(args.logs_dir))
```
alarso16: I don't love this. What if we don't want logs? are we still passing a 1 byte string of "\0" to Rust? Seems like sending nil would be more explicit

ffi/metrics_test.go
```diff
@@ -19,9 +21,19 @@ func TestMetrics(t *testing.T) {
...
 
-	db := newTestDatabase(t)
+	// test params
+	var (
+		logDir      = "/tmp/logs"
+		metricsPort = uint16(3000)
+	)
+
+	config := DefaultConfig()
+	config.Create = true
```
alarso16:
```suggestion
```
This is gonna conflict with the change I just made. You can just remove this line.

ffi/src/lib.rs
```diff
@@ -868,6 +869,10 @@ pub extern "C" fn fwd_gather() -> Value {
 /// * `cache_size` - The size of the node cache, returns an error if <= 0
 /// * `free_list_cache_size` - The size of the free list cache, returns an error if <= 0
 /// * `revisions` - The maximum number of revisions to keep; firewood currently requires this to be at least 2.
+/// * `enable_logs` - Whether to enable logs for this process.
+/// * `logs_dir` - The directory where logs for this process are stored. By
+///   default, this is set to /tmp/logs.
```
alarso16: Is this standard for most database? If so, that's ok - I'm just curious.

ffi/src/lib.rs
```diff
@@ -937,9 +945,35 @@ unsafe fn common_create(args: &CreateOrOpenArgs, create_file: bool) -> Result<Db
...
+            .map_err(|e| e.to_string())?;
+
+        env_logger::Builder::new()
+            .filter_level(level)
+            .target(env_logger::Target::Pipe(Box::new(
+                std::fs::OpenOptions::new()
+                    .create(true)
+                    .write(true)
+                    .truncate(true)
+                    .open(PathBuf::from(logs_dir).join("firewood.log"))
```
alarso16: I don't know if this should be hard-coded as path. What if we run multiple firewood instances, e.g. C-Chain and P-Chain, or the atomic trie? We should probably be able to specify a file name. There's more changes that need to happen first for this, but seems nice to not have to refactor too many things when we want to operate multiple databases

ffi/src/lib.rs
```diff
@@ -937,9 +945,35 @@ unsafe fn common_create(args: &CreateOrOpenArgs, create_file: bool) -> Result<Db
...
+            .parse::<log::LevelFilter>()
+            .map_err(|e| e.to_string())?;
+
+        env_logger::Builder::new()
+            .filter_level(level)
+            .target(env_logger::Target::Pipe(Box::new(
+                std::fs::OpenOptions::new()
+                    .create(true)
+                    .write(true)
+                    .truncate(true)
```
alarso16: Is it common practice to truncate? This is fine, I'm just curious.

ffi/src/lib.rs
rkuris: Path handling is really hard. Although this would work, we probably should address the following:
 - A non-UTF8 path is valid, but a &str with an invalid UTF8 path is not, so converting the path with `to_str` is technically incorrect. Your code returned an error when it wasn't UTF-8.
 - The default value contains an OS-specific separator
 
Also a nit: the default name could be a lot better, showing what program created it and marking it as text.

Here's my stab (but could use testing):
```suggestion
        let cstr = unsafe { CStr::from_ptr(args.logs_dir) };
        let osstr = OsStr::from_bytes(cstr.to_bytes());
        let logs_dir = (!osstr.is_empty())
            .then(|| PathBuf::from(osstr))
            .unwrap_or_else(|| {
                [std::env::temp_dir(), "firewood-log.txt".into()]
                    .iter()
                    .collect()
            });
```

This makes logs_dir into a `PathBuf` which is the right type for the calls later.

The only other question here is "what happens if cstr is nil?"

ffi/Cargo.toml
alarso16: I think it still has to be a compile time feature, otherwise it will be optimized away. I could definitely be wrong though

alarso16: I also disagree with enabling logs by default. Like all of our tests shouldn't create logs probably (besides your logging test)

alarso16 (changes requested):
I think this is overall a good way to go about it. I left some comments that may be poorly informed, and I definitely think the behavior for the config should be documented somewhere. Maybe a different PR with that and the metrics changes in the README would be ok

ffi/Cargo.toml
demosdemon: by removing the feature, he also enabled the compile-flag for it in storage. This is ok *for the ffi* library. However, ideally go consumers would build their own version of the ffi library with the flag enabled or disabled. And if disabled, the FFI should act accordingly.

But, because right now avalanchego is the only consumer of the golang ffi library, I feel it's fine to have the feature compiled in by default but not necessarily enabled by default. The FFI API caller should still request that it be enabled. 

There is a non-zero performance hit for having it compiled in but disabled. Thus the desire to have it also be a compile-time flag.

ffi/src/lib.rs
```diff
@@ -937,9 +945,35 @@ unsafe fn common_create(args: &CreateOrOpenArgs, create_file: bool) -> Result<Db
...
+        let logs_dir = unsafe { CStr::from_ptr(args.logs_dir) }
+            .to_str()
+            .map(|v| if v.is_empty() { "/tmp/log" } else { v })
+            .map_err(|e| e.to_string())?;
+
+        std::fs::create_dir_all(logs_dir).map_err(|e| e.to_string())?;
+
+        let level = unsafe { CStr::from_ptr(args.filter_level) }
+            .to_str()
+            .map(|v| if v.is_empty() { "info" } else { v })
```
rkuris: Move the map_err up one, and break this chain up for readability and debuggability.

Yes, you can make these huge combinators, but that doesn't mean you should.

ffi/metrics_test.go
RodrigoVillar: Done: 32fbd50

ffi/firewood.go
RodrigoVillar: Went with using a subconfig here: b12e5e9

ffi/src/lib.rs
RodrigoVillar: Done: 899730b

ffi/src/lib.rs
RodrigoVillar: Clients now pass in the full path (i.e. directory + file name) to Firewood: 2497c66

ffi/firewood.go
RodrigoVillar: Modified the `LogConfig` to make as few pointer allocs as necessary:  3508b05

ffi/metrics_test.go
```diff
@@ -19,9 +20,19 @@ func TestMetrics(t *testing.T) {
 	r := require.New(t)
 	ctx := context.Background()
 
-	db := newTestDatabase(t)
+	// test params
+	var (
+		logPath     = "/tmp/logs/firewood.log"
```
alarso16: nit: Might be better to use `t.TempDir()` just to ensure it's cleaned up afterward, and so you know that it's a fresh file.

ffi/src/lib.rs
```diff
@@ -916,9 +930,45 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
             args.strategy,
         )?)
         .build();
-    #[cfg(feature = "logger")]
-    let _ = env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
-        .try_init();
+
+    if !args.log_args.is_null() {
```
alarso16: nit: this is a lot of lines, any chance we want a helper?

ffi/src/lib.rs
rkuris: No. Most of the scrapers keep track of the end of the file, and on restart generally expect the old data to be there. You can actually lose data this way. If you generate enough logs on startup to pass the old "current offset" the tailer won't understand what happened in some obscure cases.

ffi/src/lib.rs
rkuris: Seems like the caller should specify the file, not the directory.

ffi/src/lib.rs
```diff
@@ -937,9 +945,35 @@ unsafe fn common_create(args: &CreateOrOpenArgs, create_file: bool) -> Result<Db
...
+            .filter_level(level)
+            .target(env_logger::Target::Pipe(Box::new(
+                std::fs::OpenOptions::new()
+                    .create(true)
+                    .write(true)
+                    .truncate(true)
+                    .open(PathBuf::from(logs_dir).join("firewood.log"))
+                    .map_err(|e| e.to_string())?,
+            )))
+            .try_init()
```
rkuris: https://docs.rs/env_logger/latest/env_logger/fn.try_init.html#errors

This means you can't open more than one database with logs.

Logs, like metrics, are not per-database, they are per-process.

ffi/src/lib.rs
RodrigoVillar: Removed truncate here: 4f3b80b

ffi/src/lib.rs
RodrigoVillar: Done: 2497c66e98d547a82cd8e9f933712a5e830f41c4

ffi/src/lib.rs
RodrigoVillar: Moved all logging logic to a helper function: 45ead16

ffi/metrics_test.go
RodrigoVillar: Great catch - done: 706df61

ffi/src/lib.rs
```diff
@@ -928,6 +938,60 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+
+#[doc(hidden)]
+fn start_logs(log_args: &LogArgs) -> Result<(), String> {
+    let cstr = unsafe { CStr::from_ptr(log_args.path) };
+    let osstr = OsStr::from_bytes(cstr.to_bytes());
+    let log_path = if osstr.is_empty() {
+        Path::join(&std::env::temp_dir(), "firewood-log.txt")
+    } else {
+        PathBuf::from(osstr)
+    };
```
demosdemon: This must check that `logs_args.path` is not null. If it is null, `CStr::from_ptr` here causes undefined behavior. 

Be sure to read the `safety` documentation for any unsafe method: https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_ptr

Best practice is to always comment above the unsafe code with `// SAFETY: ...` explaining why this is safe w.r.t. to the constraints of the unsafe call. However, FFI libraries tend not to do that because there's so much. But, upholding the unsafe requirements is still necessary.

ffi/src/lib.rs
```diff
@@ -928,6 +938,60 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+        .ok_or("failed to get log directory")?;
+    std::fs::create_dir_all(log_dir).map_err(|e| e.to_string())?;
+
+    let cstr = unsafe { CStr::from_ptr(log_args.filter_level) };
+    let level_str = OsStr::from_bytes(cstr.to_bytes())
+        .to_str()
+        .unwrap_or("info");
+    let level = level_str
+        .parse::<log::LevelFilter>()
+        .map_err(|e| e.to_string())?;
```
demosdemon: Same here, you must first check that `filter_level` is not null before calling `CStr::from_ptr` otherwise this will cause undefined behavior.

demosdemon (changes requested):

ffi/src/lib.rs
RodrigoVillar: Will do; I think we'll also need a similar check for `open_db()` as well?: https://github.com/ava-labs/firewood/blob/d8235ba35f5bc7043f8f7005ca2b56abc20a0198/ffi/src/lib.rs#L923

demosdemon: Yep! good find. Since we do this alot, can you move that into a method that converts the cstr pointer into an `Option<PathBuf>` and reuse it for your stuff? That way we also get the windows handling done (windows isn't a priority).

RodrigoVillar: Done: 31e6b5f5ca4f86105e24bd694fa64b9cc9e72a19

ffi/src/lib.rs
RodrigoVillar: Done: 31e6b5f5ca4f86105e24bd694fa64b9cc9e72a19

ffi/src/lib.rs
RodrigoVillar: Split enabling logs similar to what we have for metrics: aa383a3

ffi/src/lib.rs
RodrigoVillar: Refactor path handling here: 31e6b5f5ca4f86105e24bd694fa64b9cc9e72a19; this code also handles the `nil` case.

ffi/src/lib.rs
RodrigoVillar: If a user doesn't specify a file path for logs, then I assume they probably don't care about logs since the contents of `/tmp` are ephemeral. Current default is `/tmp/firewood-log.txt`.

demosdemon (approved):

ffi/Cargo.toml
```diff
@@ -20,19 +20,19 @@ rust-version.workspace = true
 crate-type = ["staticlib"]
 
 [dependencies]
-firewood.workspace = true
+firewood = { workspace = true, features = ["logger"] }
```
rkuris: nit: would be great if this could be decided by the builder instead of forcing it on

ffi/Cargo.toml
```diff
@@ -20,19 +20,19 @@ rust-version.workspace = true
...
 metrics = { workspace = true }
 metrics-util = { workspace = true }
 chrono = "0.4.41"
 oxhttp = "0.3.1"
+log = "0.4.27"
```
rkuris: nit: log is used by two other crates so probably belongs in the workspace now.

ffi/Cargo.toml
```diff
@@ -20,19 +20,19 @@ rust-version.workspace = true
...
 coarsetime = { workspace = true }
-env_logger = { workspace = true, optional = true}
+env_logger = { workspace = true}
```
rkuris: nit
```suggestion
env_logger = { workspace = true }
```

ffi/metrics_test.go
```diff
@@ -74,4 +86,9 @@ func TestMetrics(t *testing.T) {
 		r.NotNil(d)
 		r.Equal(v, *d.Type)
 	}
+
+	// logs should be non-empty if logging with trace filter level
```
rkuris: 💯 sure hope so

ffi/src/lib.rs
```diff
@@ -916,18 +916,88 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
-    let _ = env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
-        .try_init();
-
-    let path = unsafe { CStr::from_ptr(args.path) };
-    #[cfg(unix)]
-    let path: &Path = OsStr::from_bytes(path.to_bytes()).as_ref();
-    #[cfg(windows)]
-    let path: &Path = OsStr::new(path.to_str().expect("path should be valid UTF-8")).as_ref();
+
+    let path = to_path(args.path).ok_or("failed to provide path")?;
```
rkuris:
```suggestion
    let path = to_path(args.path).ok_or("database path is nil or empty")?;
```
Seems like if `to_path` returned a `Result<PathBuf, String>` you could do:
```
    let path = to_path(args.path).map_err(|e| format!("database path: {e}")?;
```
which could render "database path: nil" or "database path: empty"

ffi/src/lib.rs
```diff
@@ -916,18 +916,88 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+/// # Arguments
+///
+/// See `LogsArgs`.
+///
+/// # Returns
+///
+/// A `Value` containing {0, null} if the global logger was initialized.
+/// A `Value` containing {0, "error message"} if an error occurs.
+#[unsafe(no_mangle)]
+pub extern "C" fn fwd_start_logs(args: &LogArgs) -> Value {
```
rkuris: We've been defensive here and used an option here for a nil check. Otherwise, we get undefined behavior if someone gives us a nil:
```suggestion
pub extern "C" fn fwd_start_logs(args: Option<&LogArgs>) -> Value {
```

ffi/src/lib.rs
```diff
@@ -916,18 +916,88 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+    start_logs(args).map_or_else(Into::into, Into::into)
+}
+
+#[doc(hidden)]
+fn start_logs(log_args: &LogArgs) -> Result<(), String> {
+    let log_path =
+        to_path(log_args.path).unwrap_or(Path::join(&std::env::temp_dir(), "firewood-log.txt"));
+
```
rkuris, line 962: ⭐ ⭐ ⭐ ⭐ ⭐ 
so much cleaner than the earlier version 👍

ffi/src/lib.rs
```diff
@@ -916,18 +916,88 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+    let log_dir = Path::new(&log_path)
+        .parent()
+        .ok_or("failed to get log directory")?;
```
rkuris:
```suggestion
    let log_dir = log_path
        .parent()
        .unwrap_or_else(||Path::new("/"));
```
Reasons:
1 - it's fine to have a borrowed log_dir rather than an owned one. This prevents one memory allocation
2 - The docs say you get `None` only if the path is empty (which you checked earlier) or if it ends in the root directory.

Seems like having your log in the root directory is odd enough, so if you'd rather error saying "we don't support logs in the root directory" that would be okay too.

ffi/src/lib.rs
```diff
@@ -916,18 +916,88 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+        .target(env_logger::Target::Pipe(Box::new(
+            std::fs::OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(false)
+                .open(log_path)
+                .map_err(|e| e.to_string())?,
+        )))
+        .try_init()
+        .map_err(|e| e.to_string())?;
```
rkuris: nit: split this, it's a bit long. Maybe `let file = OpenOptions::new().create()...?` and then: `env_logger::Builder::new()...Box::new(file))...`

also add some additional `use` directives to shorten these paths:
```
use env_logger::Target::Pipe;
use std::fs::OpenOptions;
```
I think it's okay to keep env_logger::Builder as it shows you're setting up the env_logger.

ffi/Cargo.toml
RodrigoVillar: Since this'll affect other crates, I decided to put this in an issue to tackle after this PR: #1087

ffi/src/lib.rs
demosdemon: Not sure what the difference between empty or null is. For go, they are interchangeable in most contexts.

ffi/src/lib.rs
demosdemon: would  (`.` or empty) be better than root. unless the process is running as root, writing to root is almost guaranteed to fail.

RodrigoVillar: Went with defaulting to `.`: 02dc9cc

ffi/src/lib.rs
RodrigoVillar: Done: 14ba64e

ffi/src/lib.rs
RodrigoVillar: Great idea; done: a28a780

ffi/src/lib.rs
RodrigoVillar: Done: ba0edf2b92f92d96298c8269bef827d1744365f2

ffi/Cargo.toml
RodrigoVillar: Added back the `logger` feature here: 2961f2488bef6e9bb9e8c6f77cc3f006ac4ada25

ffi/src/lib.rs
```diff
@@ -13,13 +13,20 @@
     )
 )]
 
+#[cfg(feature = "logger")]
+use env_logger::Target::Pipe;
+#[cfg(feature = "logger")]
+use std::fs::OpenOptions;
+#[cfg(feature = "logger")]
+use std::path::Path;
```
demosdemon: nit/tip, since these are only used within the `start_logs` where the feature is enabled, you can move the imports to inside the function (at the top). This way you can remove the cfg attributes and keep the conditional imports closer to where they are used.

ffi/src/lib.rs
```diff
@@ -916,18 +923,96 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+/// Helper function to convert C String pointers to a path
+/// Returns None if the pointer is null or if the string is empty
+#[doc(hidden)]
+fn to_path(cstr: *const std::ffi::c_char) -> Option<PathBuf> {
+    if cstr.is_null() {
+        return None;
+    }
+
+    let cstr = unsafe { CStr::from_ptr(cstr) };
+    let osstr = OsStr::from_bytes(cstr.to_bytes());
```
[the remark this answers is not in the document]
demosdemon: moved to issue, okay recommendation

ffi/src/lib.rs
```diff
@@ -916,18 +923,96 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+            .map_err(|e| e.to_string())?
+    };
+    let level = level_str
+        .parse::<log::LevelFilter>()
+        .map_err(|e| e.to_string())?;
+
+    let file = OpenOptions::new()
+        .create(true)
+        .write(true)
+        .truncate(false)
```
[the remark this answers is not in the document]
demosdemon: intended

ffi/src/lib.rs
RodrigoVillar: Done: c8b665e

ffi/src/lib.rs
```diff
@@ -916,18 +916,100 @@ unsafe fn open_db(args: &CreateOrOpenArgs) -> Result<Db, String> {
...
+
+#[cfg(feature = "logger")]
+#[doc(hidden)]
+fn start_logs(log_args: &LogArgs) -> Result<(), String> {
+    use env_logger::Target::Pipe;
+    use std::fs::OpenOptions;
+    use std::path::Path;
+
+    let log_path =
+        to_path(log_args.path).unwrap_or(Path::join(&std::env::temp_dir(), "firewood-log.txt"));
```
alarso16, line 961: Isn't this platform dependent and can be overridden? I think using the temp directory is more logical than insisting on `/tmp`, but that should be documented

alarso16 (approved):

ffi/src/lib.rs
demosdemon: Yeah, this is [platform specific](https://doc.rust-lang.org/std/env/fn.temp_dir.html#platform-specific-behavior). I'm not confident on what the best thing to do is. Ideally, the caller always provides the path and the supposes whatever risk is caused by using that path. Writing to a deterministic path could pose a security risk. However, for Windows and macOS, this is "safe-enough" as the paths provided by the OS are owned by only the executing user. Linux doesn't give that guarantee.

rkuris: The safe bit here is that we are not letting the end-user specify the filename, so you can't set it up to be "/etdc/passwd" or soemthing.

Issue #1075, opened 2025-07-16
merkle::tests::test_root_hash_reversed_deletions generates duplicate keys and fails


demosdemon:
logs https://gist.github.com/demosdemon/3c11b0649d5cb42c352a34f3bec16ef2
FIREWOOD_TEST_SEED=13284556855046797830

demosdemon:
failure reproduces on v0.0.7 tag as well (with the same tree hashes) so not because of any recent changes.

demosdemon:
4f1fdc88d11788cb8be23aa5f5bdc128a3399969 is the first bad commit
commit 4f1fdc88d11788cb8be23aa5f5bdc128a3399969
Author: Suyan Qu <36519575+qusuyan@users.noreply.github.com>
Date:   Wed Jun 25 10:05:33 2025 -0400

    Cleanup Nodestore Public Interface (#986)

    The `kind` and `storage` fields of `Nodestore` are public so that
    `firewood` can compute the root hash - but it is already stored in memory
    when the `Nodestore` is created. This diff converts these two fields
    back to private.

 firewood/src/db.rs       |   8 ++--
 firewood/src/manager.rs  |  27 ++++--------
 firewood/src/merkle.rs   | 105 +++++++++++++++++++++++++++--------------------
 firewood/src/stream.rs   |   9 ++--
 storage/src/nodestore.rs |  68 +++++++++++++-----------------
 5 files changed, 107 insertions(+), 110 deletions(-)
bisect found first bad commit

demosdemon:
commit is just the introduction of the test so unsure if issue is also before

rkuris:
Yeah the test is flaky/broken. There is a duplicate key being generated:
  inserting [1, 0, 135] [138, 191, 89, 141, 78, 152, 116, 242]
  inserting [1, 0, 135] [162, 125, 174, 53, 71, 197, 50, 166]

PR #1076, opened 2025-07-16
fix: root_hash_reversed_deletions duplicate keys

rkuris:
This test inserts a bunch of keys and then verifies that the resulting hashes are the same if we delete them in the reverse order. However, the test is generating a duplicate key, which will not reproduce the original hash set.

Changed the test to remove any duplicate keys found. It's okay if the data is the same.

Now passes with the errant key from #1075

Also fixed the merkle dump generator, which I used to debug this. A bracket was missing so graphviz was failing to parse the file. Also shortened the hashes as they made the nodes look much bigger, and added a new line. See attached merkle dump for the new format.


rkuris:
graphviz (14)

Updated graphviz output sample
```text
digraph Merkle {
  rankdir=LR;
 root -> 4256
  4256[label="4256
 H=c2841b pp=00"]
  4256 -> 2720[label="0"]
  2720[label="2720
 H=0b86a6"]
  2720 -> 2240[label="0"]
  2240[label="2240
 H=515143 pp=0"]
  2240 -> 2080[label="0"]
  2080[label="2080
 H=ffc1c8 pp=00000100fd0d val=5b44eb..." shape=rect]
  2240 -> 2112[label="1"]
  2112[label="2112
 H=76dca5 pp=000000e822 val=6cea1e..." shape=rect]
  2720 -> 2528[label="1"]
  2528[label="2528
 H=f44f56 pp=010"]
  2528 -> 2464[label="0"]
  2464[label="2464
 H=6f9847 pp=0000dc4e13 val=250ebe..." shape=rect]
  2528 -> 2496[label="13"]
  2496[label="2496
 H=4edb8d pp=a785af val=98fe9e..." shape=rect]
  4256 -> 2816[label="5"]
  2816[label="2816
 H=a5b53e pp=1 val=edb0ff..." shape=rect]
  4256 -> 2928[label="15"]
  2928[label="2928
 H=28d21b pp=d27 val=13ffdc..." shape=rect]
}
```

demosdemon (approved):

Issue #960, opened 2025-06-12
ci does not build triehash with --no-default-features


demosdemon:
Caused #959 to go unnoticed

demosdemon:
og issue is fixed, but leaving the devops task open as it is good practice to ensure `--no-default-features` is included in our CI and builds.

PR #1081, opened 2025-07-17
ci: run CI with --no-default-features

demosdemon:
Fixes: #960


rkuris (approved):
I'm a bit concerned about adding another cache. We may already be hitting the cache size limits. Assuming that isn't a problem here, and approving anyway.

If we find that it is a problem, we should consider moving this to a main check and/or not using a cache.

PR #1085, opened 2025-07-17
feat(checker): checker errors include reference to parent

qusuyan:
This is the first step towards fixing issues reported by the checker. We will include the parent reference in the error so that we can update the parent to remove the area with issue.

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -26,6 +26,13 @@ pub struct CheckOpt {
     pub progress_bar: Option<ProgressBar>,
 }
 
+struct SubTrieMetadata {
+    root_address: LinearAddress,
+    root_hash: HashType,
+    parent: TrieNodeParent,
+    path_prefix: Path,
+}
+
 /// [`NodeStore`] checker
 // TODO: S needs to be writeable if we ask checker to fix the issues
 #[expect(clippy::result_large_err)]
@@ -65,8 +72,6 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 self.visit_trie(
                     root_address,
                     root_hash.into_hash_type(),
-                    TrieNodeParent::Root,
-                    Path::new(),
                     &mut visited,
                     opt.progress_bar.as_ref(),
                     opt.hash_check,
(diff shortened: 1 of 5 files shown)


storage/src/checker/mod.rs
```diff
@@ -2,7 +2,7 @@
 // See the file LICENSE.md for licensing terms.
 
 mod range_set;
-use range_set::LinearAddressRangeSet;
+pub use range_set::LinearAddressRangeSet;
```
rkuris: This widens the storage interface significantly, and this PR does not have references to `LinearAddressRangeSet` outside the crate. Was this intended, and if so, why?

storage/src/checker/mod.rs
```diff
@@ -77,30 +72,37 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
         subtree_root_hash: HashType,
+        parent: TrieNodeParent,
         path_prefix: Path,
         visited: &mut LinearAddressRangeSet,
         hash_check: bool,
     ) -> Result<(), CheckerError> {
+        // check that address is aligned
+        self.check_area_aligned(
+            subtree_root_address,
+            StoredAreaParent::TrieNode(parent.clone()),
```
rkuris: Assuming you are using either vscode or cursor, please set the rust-analyzer's check command to clippy to avoid checking these in next time:

Screenshot 2025-07-17 at 12 52 49 PM

storage/src/lib.rs
```diff
@@ -141,13 +141,15 @@ pub enum CheckerError {
 
     /// Hash mismatch for a node
     #[error(
-        "Hash mismatch for node {partial_path:?} at address {address}: parent stored {parent_stored_hash}, computed {computed_hash}"
+        "Hash mismatch for node {partial_path:?} at address {address}: parent ({parent:?}) stored {parent_stored_hash}, computed {computed_hash}"
```
rkuris: This looks pretty ugly I think. It's going to say "parent TrieNodeParent{Parent{LinearAddress{123}, 42}}}" or something like that.

I also think the partial path is not valuable at this point either.

rkuris (changes requested):

storage/src/checker/mod.rs
qusuyan: I have been looking for this trick for so long lol

storage/src/checker/mod.rs
qusuyan: Currently `CheckerError::AreaLeaks` returns this. I am not too worried about the scope issue since the only things we exposed to external crates are `fmt()` (`std::fmt::Display`) and `into_iter()`. All other methods and members of `LinearAddressRangeSet` are not exposed.

storage/src/lib.rs
qusuyan: I think partial path does add value here - if a node is polluted, it will have one conflict with its child and one with its parent. The partial path allows us to correlate these two together easily.

storage/src/lib.rs
```diff
@@ -118,14 +118,41 @@ pub enum TrieNodeParent {
...
     FreeListHead(AreaIndex),
     /// The stored area is not the head of the free list, so a previous free area points to it
     PrevFreeArea(LinearAddress),
 }
 
+impl std::fmt::Display for StoredAreaParent {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            StoredAreaParent::TrieNode(trie_parent) => write!(f, "{trie_parent}"),
+            StoredAreaParent::FreeList(free_list_parent) => write!(f, "{free_list_parent}"),
```
rkuris: Avoid write, as you lose a lot of the other attributes being passed down to the formatter.

If someone does `{sap:5}` your code will not respect the limit
```suggestion
            StoredAreaParent::TrieNode(trie_parent) => std::fmt::Display::fmt(trie_parent, f),
            StoredAreaParent::FreeList(free_list_parent) => std::fmt::Display::fmt(free_list_parent, f),
```

Also please implement LowerHex. I just fixed a bug in some test because this was missing.

storage/src/lib.rs
```diff
@@ -118,14 +118,41 @@ pub enum TrieNodeParent {
...
+            TrieNodeParent::Parent(addr, index) => write!(f, "TrieNode@{addr:#x}[{index}]"),
+        }
+    }
+}
+
+impl std::fmt::Display for FreeListParent {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            FreeListParent::FreeListHead(index) => write!(f, "FreeLists[{index}]"),
+            FreeListParent::PrevFreeArea(addr) => write!(f, "FreeArea@{addr:#x}"),
```
rkuris: Delegate, and implement LowerHex (possibly also UpperHex).

storage/src/lib.rs
rkuris: I'm not following.

The full path as nibbles to this node might be useful, since it's unique, but not the partial path. Are you thinking that the partial path is the same as this? It's not, it's just the common components below where we are, and it's only nibbles.

Imagine a data set of AAA.ZZZ roughly 50% of all keys are populated. Your partial path of a corrupted branch node is 0x5. What does that tell you? Nothing. There are probably a handful of branches with that partial path. It's the first half of a large percentage of any of the parts of the keys. Instead, imagine hashes, and typically we see a single nibble in a branch, which could be anywhere in the trie.

The only way to correlate children is if they are in the same node, and partial paths are never going to be unique.

qusuyan: My bad - naming here is horrible

This is the full path, let me change it real quick.

storage/src/node/path.rs
```diff
@@ -40,6 +40,16 @@ impl Debug for Path {
     }
 }
 
+impl LowerHex for Path {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        write!(f, "0x")?;
```
demosdemon: This is nitpicky and a tad esoteric, but the prefix should only be included when the alternate form is used [(for parity with stdlib)](https://doc.rust-lang.org/stable/std/fmt/index.html#:~:text=This%20flag%20indicates%20that%20the%20%E2%80%9Calternate%E2%80%9D%20form%20of%20printing%20should%20be%20used.%20The%20alternate%20forms%20are%3A).

The difference is:

```rust
write!(f, "{path:x}") // should not print the prefix
```

vs

```rust
write!(f, "{path:#x}") // should print the prefix
```

Where the hash indicates to use the alternate form.

```suggestion
        if f.alternate() {
            write!(f, "0x")?;
        }
```

---

This change would also allow us to chain multiple partial paths into one large one, for print only, without actually concatenating them into a larger buffer.

storage/src/lib.rs
```diff
@@ -118,14 +118,50 @@ pub enum TrieNodeParent {
...
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
 pub enum FreeListParent {
     /// The stored area is the head of the free list, so the header points to it
     FreeListHead(AreaIndex),
     /// The stored area is not the head of the free list, so a previous free area points to it
     PrevFreeArea(LinearAddress),
 }
 
+impl std::fmt::LowerHex for StoredAreaParent {
```
rkuris: `use std::fmt` here, then use `impl fmt::LowerHex`. This removes a lot of `std::fmt` stutter below too.

storage/src/node/path.rs
```diff
@@ -40,6 +40,16 @@ impl Debug for Path {
...
+        for nib in &self.0 {
+            write!(f, "{:x}", *nib)?;
+        }
```
rkuris: I would probably delegate. Alignment, fill, and precision are not handled, and probably should be.

I'm okay with leaving a TODO on whatever it is you aren't implementing yet.
```suggestion
    let mut iter = self.0.iter();
        let Some(first) = iter.next() else {
            return Ok(())
        };
        fmt::LowerHex::fmt(first, f)?;
        while let Some(next) = iter.next() {
            write!(f, "{:x}", next)?;
        }
        Ok(())
```

storage/src/node/path.rs
```diff
@@ -40,6 +40,16 @@ impl Debug for Path {
...
+        Ok(())
+    }
+}
+
```
rkuris, line 60: This needs some tests!

rkuris (changes requested):

storage/src/node/path.rs
qusuyan: I don't think we will ever want to print nibbles with precisions or `0x` prefix but will leave a todo here if there is a real need.

storage/src/node/path.rs
```diff
@@ -40,6 +40,23 @@ impl Debug for Path {
...
+impl LowerHex for Path {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        if self.0.is_empty() {
+            write!(f, "[]")
+        } else {
+            if f.alternate() {
+                write!(f, "0x")?;
+            }
+            for nib in &self.0 {
+                // TODO: delegate instead of calling write! directly
```
demosdemon: this is a case where you _don't_ want to delegate to the `<u8 as LowerHex>::fmt` because it would intersperse the `0x` prefixes.

PR #1086, opened 2025-07-17
chore: Implement newtype for LInearAddress

rkuris:
Took over the diff from @Salonii02 <devkss2000@gmail.com>, see PR#1062 or the commit history for their original commit and my changes.

Basically the same as PR#1062 with these changes:
  - Warning cleanups
  - serde->bytemuck
  - array dereference
  - more liberal use of into()
  - Implement LowerHex and delegate Display handling
  
I'm not happy with where LinearAddress lives, it probably belongs one level up, but it can move later. This adds a lot of value.

### DIFF
--- a/storage/Cargo.toml
+++ b/storage/Cargo.toml
@@ -39,7 +39,9 @@ sha3 = { version = "0.10.8", optional = true }
 bytes = { version = "1.10.1", optional = true }
 thiserror = { workspace = true }
 semver = "1.0.26"
+
 nonzero_ext = "0.3.0"
+num-traits = "0.2.19"
 
 [dev-dependencies]
 rand = { workspace = true }
--- a/storage/Cargo.toml
+++ b/storage/Cargo.toml
@@ -39,7 +39,6 @@ sha3 = { version = "0.10.8", optional = true }
 bytes = { version = "1.10.1", optional = true }
 thiserror = { workspace = true }
 semver = "1.0.26"
-
 nonzero_ext = "0.3.0"
 num-traits = "0.2.19"
 
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -212,7 +212,8 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 Ordering::Less => {
                     // continue to the next area
                     leaked.push((current_addr, area_index));
-                    current_addr = next_addr;
+                    current_addr =
+                        LinearAddress::new(next_addr.get()).expect("next_addr is non-zero");
                 }
             }
         }
@@ -224,9 +225,10 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 let next_addr = current_addr
                     .checked_add(*area_size)
                     .expect("address overflow is impossible");
-                if next_addr <= leaked_range.end {
+                if next_addr <= *leaked_range.end {
                     leaked.push((current_addr, area_index as AreaIndex));
-                    current_addr = next_addr;
+                    current_addr =
+                        LinearAddress::new(next_addr.get()).expect("next_addr is non-zero");
                 } else {
                     break;
                 }
@@ -243,6 +245,8 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 mod test {
     #![expect(clippy::unwrap_used)]
     #![expect(clippy::indexing_slicing)]
+
+    use nonzero_ext::nonzero;
 
     use super::*;
     use crate::linear::memory::MemStore;
@@ -320,7 +324,6 @@ mod test {
         )
     }
 
-    use nonzero_ext::nonzero;
     use std::collections::HashMap;
 
     #[test]
@@ -470,13 +473,13 @@ mod test {
             } else {
                 // we are not freeing this area, free the aggregated areas before this one
                 if let Some((start, end)) = area_to_free {
-                    leaked.push(start..end);
+                    leaked.push(start..end.into());
                     area_to_free = None;
                 }
             }
         }
         if let Some((start, end)) = area_to_free {
-            leaked.push(start..end);
+            leaked.push(start..end.into());
         }
 
         // check the leaked areas
@@ -513,7 +516,7 @@ mod test {
         test_write_zeroed_area(&nodestore, leaked_range_size, NodeStoreHeader::SIZE);
 
         // check the leaked areas
-        let leaked_range = nonzero!(NodeStoreHeader::SIZE)
+        let leaked_range = nonzero!(NodeStoreHeader::SIZE).into()
             ..LinearAddress::new(
                 NodeStoreHeader::SIZE
                     .checked_add(leaked_range_size)
@@ -566,7 +569,7 @@ mod test {
 
         // check the leaked areas
         let leaked_range =
-            nonzero!(NodeStoreHeader::SIZE)..LinearAddress::new(high_watermark).unwrap();
+            nonzero!(NodeStoreHeader::SIZE).into()..LinearAddress::new(high_watermark).unwrap();
         let (leaked_areas_offsets, leaked_area_size_indices): (Vec<LinearAddress>, Vec<AreaIndex>) =
             nodestore
                 .split_range_into_leaked_areas(leaked_range)
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -6,7 +6,6 @@ use range_set::LinearAddressRangeSet;
 
 use crate::logger::warn;
 use crate::nodestore::alloc::{AREA_SIZES, AreaIndex, FreeAreaWithMetadata};
-use crate::nodestore::is_aligned;
 use crate::{
     CheckerError, Committed, HashType, HashedNodeReader, LinearAddress, Node, NodeReader,
     NodeStore, Path, StoredAreaParent, TrieNodeParent, WritableStorage, hash_node,
@@ -154,7 +153,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         address: LinearAddress,
         parent_ptr: StoredAreaParent,
     ) -> Result<(), CheckerError> {
-        if !is_aligned(address) {
+        if !address.is_aligned() {
             return Err(CheckerError::AreaMisaligned {
                 address,
                 parent_ptr,
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -194,7 +194,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
             };
 
             let next_addr = current_addr
-                .checked_add(area_size)
+                .advance(area_size)
                 .expect("address overflow is impossible");
             match next_addr.cmp(&leaked_range.end) {
                 Ordering::Equal => {
@@ -212,8 +212,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 Ordering::Less => {
                     // continue to the next area
                     leaked.push((current_addr, area_index));
-                    current_addr =
-                        LinearAddress::new(next_addr.get()).expect("next_addr is non-zero");
+                    current_addr = next_addr;
                 }
             }
         }
@@ -223,12 +222,11 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         for (area_index, area_size) in AREA_SIZES.iter().enumerate().rev() {
             loop {
                 let next_addr = current_addr
-                    .checked_add(*area_size)
+                    .advance(*area_size)
                     .expect("address overflow is impossible");
-                if next_addr <= *leaked_range.end {
+                if next_addr <= leaked_range.end {
                     leaked.push((current_addr, area_index as AreaIndex));
-                    current_addr =
-                        LinearAddress::new(next_addr.get()).expect("next_addr is non-zero");
+                    current_addr = next_addr;
                 } else {
                     break;
                 }
--- a/storage/src/checker/range_set.rs
+++ b/storage/src/checker/range_set.rs
@@ -221,9 +221,7 @@ pub(super) struct LinearAddressRangeSet {
 
 #[expect(clippy::result_large_err)]
 impl LinearAddressRangeSet {
-    fn node_store_addr_start() -> LinearAddress {
-        LinearAddress::new(NodeStoreHeader::SIZE).expect("NodeStoreHeader::SIZE is non-zero")
-    }
+    const NODE_STORE_START_ADDR: LinearAddress = LinearAddress::new(NodeStoreHeader::SIZE).unwrap();
 
     pub(super) fn new(db_size: u64) -> Result<Self, CheckerError> {
         if db_size < NodeStoreHeader::SIZE {
@@ -256,13 +254,13 @@ impl LinearAddressRangeSet {
             .ok_or(CheckerError::AreaOutOfBounds {
                 start,
                 size,
-                bounds: Self::node_store_addr_start()..self.max_addr,
+                bounds: Self::NODE_STORE_START_ADDR..self.max_addr,
             })?; // This can only happen due to overflow
-        if addr < Self::node_store_addr_start() || end > *self.max_addr {
+        if addr < Self::NODE_STORE_START_ADDR || end > *self.max_addr {
             return Err(CheckerError::AreaOutOfBounds {
                 start: addr,
                 size,
-                bounds: Self::node_store_addr_start()..self.max_addr,
+                bounds: Self::NODE_STORE_START_ADDR..self.max_addr,
             });
         }
 
@@ -281,7 +279,7 @@ impl LinearAddressRangeSet {
     pub(super) fn complement(&self) -> Self {
         let complement_set = self
             .range_set
-            .complement(&Self::node_store_addr_start(), &self.max_addr);
+            .complement(&Self::NODE_STORE_START_ADDR, &self.max_addr);
 
         Self {
             range_set: complement_set,
@@ -659,7 +657,7 @@ mod test_linear_address_range_set {
         let size2 = 1024;
         let db_size = 0x2000;
 
-        let db_begin = LinearAddressRangeSet::node_store_addr_start();
+        let db_begin = LinearAddressRangeSet::NODE_STORE_START_ADDR;
         let start1_addr = LinearAddress::new(start1).unwrap();
         let end1_addr = LinearAddress::new(start1 + size1).unwrap();
         let start2_addr = LinearAddress::new(start2).unwrap();
@@ -697,7 +695,7 @@ mod test_linear_address_range_set {
 
     #[test]
     fn test_complement_with_empty() {
-        let db_size = LinearAddressRangeSet::node_store_addr_start();
+        let db_size = LinearAddressRangeSet::NODE_STORE_START_ADDR;
         let visited = LinearAddressRangeSet::new(db_size.get()).unwrap();
         let complement = visited.complement().into_iter().collect::<Vec<_>>();
         assert_eq!(complement, vec![]);
--- a/storage/src/checker/range_set.rs
+++ b/storage/src/checker/range_set.rs
@@ -221,7 +221,9 @@ pub(super) struct LinearAddressRangeSet {
 
 #[expect(clippy::result_large_err)]
 impl LinearAddressRangeSet {
-    const NODE_STORE_ADDR_START: LinearAddress = LinearAddress::new(NodeStoreHeader::SIZE).unwrap();
+    fn node_store_addr_start() -> LinearAddress {
+        LinearAddress::new(NodeStoreHeader::SIZE).unwrap()
+    }
 
     pub(super) fn new(db_size: u64) -> Result<Self, CheckerError> {
         if db_size < NodeStoreHeader::SIZE {
@@ -239,7 +241,7 @@ impl LinearAddressRangeSet {
 
         Ok(Self {
             range_set: RangeSet::new(),
-            max_addr, // STORAGE_AREA_START..U64::MAX
+            max_addr: max_addr, // STORAGE_AREA_START..U64::MAX
         })
     }
 
@@ -254,13 +256,13 @@ impl LinearAddressRangeSet {
             .ok_or(CheckerError::AreaOutOfBounds {
                 start,
                 size,
-                bounds: Self::NODE_STORE_ADDR_START..self.max_addr,
+                bounds: Self::node_store_addr_start()..self.max_addr,
             })?; // This can only happen due to overflow
-        if addr < Self::NODE_STORE_ADDR_START || end > self.max_addr {
+        if addr < Self::node_store_addr_start() || end > self.max_addr {
             return Err(CheckerError::AreaOutOfBounds {
                 start: addr,
                 size,
-                bounds: Self::NODE_STORE_ADDR_START..self.max_addr,
+                bounds: Self::node_store_addr_start()..self.max_addr,
             });
         }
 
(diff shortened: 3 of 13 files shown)


storage/Cargo.toml
```diff
@@ -39,7 +39,9 @@ sha3 = { version = "0.10.8", optional = true }
 bytes = { version = "1.10.1", optional = true }
 thiserror = { workspace = true }
 semver = "1.0.26"
+
```
demosdemon: nit: no space.

storage/src/checker/range_set.rs
```diff
@@ -221,7 +221,9 @@ pub(super) struct LinearAddressRangeSet {
 
 #[expect(clippy::result_large_err)]
 impl LinearAddressRangeSet {
-    const NODE_STORE_ADDR_START: LinearAddress = LinearAddress::new(NodeStoreHeader::SIZE).unwrap();
```
demosdemon, line 224: This should still be a const value. There should be no need to push this into a function.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+pub struct LinearAddress(NonZeroU64);
+
+#[expect(unsafe_code)]
+unsafe impl bytemuck::ZeroableInOption for LinearAddress {}
+#[expect(unsafe_code)]
+unsafe impl bytemuck::PodInOption for LinearAddress {}
+
+impl LinearAddress {
+    /// Create a new `LinearAddress`, returns None if value is zero.
+    pub fn new(addr: u64) -> Option<Self> {
```
demosdemon: This should be const. `map` isn't const so you have to use a match expr.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    }
+
+    /// Returns a reference to the inner `NonZeroU64`
+    #[must_use]
+    pub const fn as_nonzero(&self) -> &NonZeroU64 {
+        &self.0
+    }
+}
+
+impl std::ops::Add for LinearAddress {
```
demosdemon, line 227: We should have const versions of these methods so we can do arithmetic in const context. (`const fn checked_add(self, rhs: u64) -> Option<Self> {...}`)

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        self.checked_add(other.get())
+            .map(LinearAddress)
+            .expect("non zero result")
+    }
+}
+
+impl num_traits::CheckedAdd for LinearAddress {
```
demosdemon, line 236: Traits like these aren't able to be const (yet) so it's better to have regular implementations for const functions like this and then call those for the trait implementation.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
+pub struct LinearAddress(NonZeroU64);
+
+#[expect(unsafe_code)]
+unsafe impl bytemuck::ZeroableInOption for LinearAddress {}
+#[expect(unsafe_code)]
+unsafe impl bytemuck::PodInOption for LinearAddress {}
+
+impl LinearAddress {
```
demosdemon, line 163: Everything that is one line should be marked inline.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    #[must_use]
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
+    }
+
+    /// Returns the maximum area size available for allocation.
+    #[allow(clippy::missing_panics_doc)]
+    #[must_use]
+    pub const fn max_area_size() -> u64 {
+        *AREA_SIZES.last().expect("AREA_SIZES is never empty")
```
demosdemon: Wrap this in `const {...}` and the clippy lint should go away. This could also be saved as a const value.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    }
+
+    /// Returns the minimum area size available for allocation.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
+    pub const fn min_area_size() -> u64 {
+        *AREA_SIZES.first().expect("AREA_SIZES is never empty")
```
demosdemon: Same here

storage/src/nodestore/alloc.rs
demosdemon: also recommend a const method to construct from a nonzero number.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,125 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    }
+
+    /// Check if the address is 8-byte aligned.
+    #[must_use]
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
+    }
+
+    /// Returns the maximum area size available for allocation.
+    #[allow(clippy::missing_panics_doc)]
```
demosdemon: Also, don't use allow if you can't get the lint to go away. Expect instead.

storage/src/nodestore/alloc.rs
rkuris: This is somewhat tricky because many of these trait methods are generated so they can take compatible types. For example, this works
```
        let next_address = LinearAddress::new(1).unwrap().checked_add(1u64);
```

It's quite rare that you'll actually be adding two LinearAddresses together.

A good solution to this would be to write an `advance` method that advances a LinearAddress using checked_add. I don't see use cases where you'd want to subtract coming up very often, so I implemented this as a const method, which addresses the concern here I think.

storage/src/nodestore/alloc.rs
rkuris: I wrote a new_unchecked but it seemed wrong, so I left it out. We can do it as a followup if you feel strongly about it.

storage/src/nodestore/alloc.rs
rkuris: See the next comment. Solved with an `advance` method.

storage/src/checker/mod.rs
```diff
@@ -212,7 +212,8 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 Ordering::Less => {
                     // continue to the next area
                     leaked.push((current_addr, area_index));
-                    current_addr = next_addr;
+                    current_addr =
+                        LinearAddress::new(next_addr.get()).expect("next_addr is non-zero");
```
qusuyan: Why do you need to do this? Is `next_addr` not `LinearAddress`?

storage/src/checker/mod.rs
```diff
@@ -224,9 +225,10 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 let next_addr = current_addr
                     .checked_add(*area_size)
                     .expect("address overflow is impossible");
-                if next_addr <= leaked_range.end {
+                if next_addr <= *leaked_range.end {
                     leaked.push((current_addr, area_index as AreaIndex));
-                    current_addr = next_addr;
+                    current_addr =
+                        LinearAddress::new(next_addr.get()).expect("next_addr is non-zero");
```
qusuyan: Same here

storage/src/checker/range_set.rs
```diff
@@ -254,17 +254,19 @@ impl LinearAddressRangeSet {
...
                 start: addr,
                 size,
-                bounds: Self::NODE_STORE_ADDR_START..self.max_addr,
+                bounds: Self::NODE_STORE_START_ADDR..self.max_addr,
             });
         }
 
-        if let Err(intersection) = self.range_set.insert_disjoint_range(start..end) {
+        if let Err(intersection) = self.range_set.insert_disjoint_range(
+            start..LinearAddress::new(end.get()).expect("end address is non-zero"),
```
qusuyan: Same here

storage/src/node/persist.rs
```diff
@@ -199,7 +200,8 @@ mod test {
 
     #[test]
     fn test_from_linear_address() {
-        let addr = nonzero!(1024u64);
+        let raw_addr = nonzero!(1024u64);
+        let addr = LinearAddress::new(raw_addr.get()).unwrap();
```
qusuyan: Why not just
```suggestion
        let addr = LinearAddress::new(1024u64).unwrap();
```

storage/src/node/persist.rs
```diff
@@ -227,6 +229,7 @@ mod test {
 
         // Persist the original
         let addr = nonzero!(4096u64);
+        let addr = LinearAddress::new(addr.get()).unwrap();
```
qusuyan: Same here
```suggestion
        let addr = LinearAddress::new(4096u64).unwrap();
```

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+impl LinearAddress {
+    /// Create a new `LinearAddress`, returns None if value is zero.
+    #[expect(clippy::missing_panics_doc)]
+    #[inline]
+    #[must_use]
+    pub const fn new(addr: u64) -> Option<Self> {
+        match addr {
+            0 => None,
+            _ => Some(LinearAddress(NonZeroU64::new(addr).expect("non zero"))),
+        }
```
qusuyan: Can we just do this?
```suggestion
        NonZeroU64::new(addr).map(Self)
```

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    }
```
demosdemon: expect is unnecessary here:

```suggestion
    pub const fn new(addr: u64) -> Option<Self> {
        match NonZeroU64::new(addr) {
            Some(addr) => Some(LinearAddress(addr)),
            None => None,
        }
    }
```

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    #[must_use]
+    pub const fn get(&self) -> u64 {
+        self.0.get()
+    }
+
+    /// Check if the address is 8-byte aligned.
+    #[inline]
+    #[must_use]
+    pub const fn is_aligned(&self) -> bool {
+        self.0.get() % 8 == 0
```
qusuyan:
```suggestion
        self.0.get() % Self::min_area_size() == 0
```

storage/src/nodestore/alloc.rs
demosdemon: Making `new` const solves most of the problem as I specifically want to be able to construct values in const context as that means we have no need to check for `Option` or zero at runtime.

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+        const { *AREA_SIZES.first().unwrap() }
+    }
 
-pub type FreeLists = [Option<LinearAddress>; NUM_AREA_SIZES];
+    /// Returns the number of different area sizes available.
+    #[inline]
+    #[must_use]
+    pub const fn num_area_sizes() -> usize {
+        const { AREA_SIZES.len() }
+    }
```
qusuyan: Why do these functions need to be in `impl LinearAddress`? And why can't we just return `MIN_AREA_SIZE`, `MAX_AREA_SIZE`, and `NUM_AREA_SIZES`?

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+            // overflowed
+            None => None,
+
+            // It is impossible to add a non-zero positive number to a u64 and get 0 without
+            // overflowing, so we don't check for that here, and panic instead.
+            Some(sum) => Some(LinearAddress(
+                NonZeroU64::new(sum).expect("sum cannot be zero"),
+            )),
+        }
+    }
```
demosdemon: use [checked_add](https://doc.rust-lang.org/stable/std/num/type.NonZeroU64.html#method.checked_add) on the nonzero and not the u64. It will return a `Some(NonZero<_>)`
```suggestion
        match self.0.checked_add(n) {
            // overflowed
            None => None,
            Some(sum) => Some(LinearAddress(sum)),
        }
    }
```

qusuyan: Isn't this a wrapper for `checked_add`?
```suggestion
    pub const fn advance(self, n: u64) -> Option<Self> {
        self.0.checked_add(n).map(Self)
    }
```

storage/src/checker/mod.rs
demosdemon:
```
LinearAddress::new(next_addr.get()).expect(...)
```

This should be an anti-pattern. Why is `next_addr` not a `LinearAddress`. The point of the newtype should be to minimize panic points.

storage/src/checker/mod.rs
demosdemon: same here, `next_addr` should already be a `LinearAddress` by this point and we should not need to call expect.

storage/src/nodestore/mod.rs
```diff
@@ -93,7 +93,7 @@ use super::linear::WritableStorage;
 
 #[inline]
 pub(crate) const fn is_aligned(addr: LinearAddress) -> bool {
-    addr.get() % alloc::MIN_AREA_SIZE == 0
+    addr.is_aligned()
 }
```
qusuyan: We probably don't need this function any more, just call `addr.is_aligned()` should be sufficient

storage/src/checker/range_set.rs
demosdemon: `end` should already be a `LinearAddress` here.

qusuyan (changes requested):

storage/src/nodestore/mod.rs
```diff
@@ -832,7 +832,7 @@ mod tests {
         value: Some(vec![9, 10, 11].into_boxed_slice()),
         children: from_fn(|i| {
             if i == 15 {
-                Some(Child::AddressWithHash(nonzero!(1u64), std::array::from_fn::<u8, 32, _>(|i| i as u8).into()))
+                Some(Child::AddressWithHash(LinearAddress::new(1).unwrap(), std::array::from_fn::<u8, 32, _>(|i| i as u8).into()))
```
demosdemon, line 830: we could potentially create our own macro

```rust
#[macro_export]
macro_rules! linear_addr {
    ($v:literal) => {
        const { $crate::nodestore::alloc::LinearAddress::new($v).unwrap() }
    };
}
```

Which will evaluate the literal at compile time. E.g.,

```
error[E0080]: evaluation of `test_utils::seeded_rng::{constant#0}` failed
   --> storage/src/lib.rs:63:17
    |
63  |         const { $crate::nodestore::alloc::LinearAddress::new($v).unwrap() }
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation panicked: called `Option::unwrap()` on a `None` value
...
250 |         let addr = linear_addr!(0);
    |                    --------------- in this macro invocation
    |
    = note: this error originates in the macro `linear_addr` (in Nightly builds, run with -Z macro-backtrace for more info)

```

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+        match addr {
+            0 => None,
+            _ => Some(LinearAddress(NonZeroU64::new(addr).expect("non zero"))),
+        }
+    }
+
+    /// Get the underlying address as u64.
+    #[inline]
+    #[must_use]
+    pub const fn get(&self) -> u64 {
```
demosdemon: `LinearAddress` is `Copy` and only 8 bytes. It's more efficient to always use `(self)` and not `(&self)`. There is a clippy lint for this and I'm unsure why it is not triggering: https://rust-lang.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,158 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
         })
 }
 
-/// Objects cannot be stored at the zero address, so a [`LinearAddress`] is guaranteed not
-/// to be zero. This reserved zero can be used as a [None] value for some use cases. In particular,
-/// branches can use `Option<LinearAddress>` which is the same size as a [`LinearAddress`]
-pub type LinearAddress = NonZeroU64;
+// pub type LinearAddress = NonZeroU64;
```
demosdemon: nit: delete

storage/src/node/persist.rs
demosdemon: This is why having the ability to construct a linearaddress from an existing non-zero is necessary. this call to unwrap should be unnecessary.

demosdemon (changes requested):

storage/src/nodestore/alloc.rs
demosdemon: `map` is not `const` so we cannot use it here, but the expect is not necessary

storage/src/nodestore/alloc.rs
demosdemon: yes, having associated functions that return const values and not associated constants feels weird.

storage/src/linear/filebacked.rs
```diff
@@ -217,8 +217,8 @@ impl WritableStorage for FileBacked {
         nodes: impl Iterator<Item = (&'a std::num::NonZero<u64>, &'a SharedNode)>,
```
demosdemon: should this trait use `LinearAddress`es instead of a plain nonzero?

storage/src/nodestore/alloc.rs
```diff
@@ -164,7 +311,7 @@ pub struct FreeArea {
 impl Serializable for FreeArea {
     fn write_to<W: crate::node::ExtendableBytes>(&self, vec: &mut W) {
         vec.push(0xff); // 0xff indicates a free area
-        vec.extend_var_int(self.next_free_block.map_or(0, LinearAddress::get));
+        vec.extend_var_int(self.next_free_block.map_or(0, |addr| addr.get()));
```
demosdemon: this change should be unnecessary once the signature for `get` is `(self)->u64;`

storage/src/nodestore/alloc.rs
demosdemon: Ah, I realize what it is. These do not belong on `LinearAddress`. They belong to the future `AreaIndex` newtype. Adding them here feels like a wart.

storage/src/nodestore/alloc.rs
rkuris: Clippy whines so I'm leaving it. The alternative is to suppress clippy.

storage/src/nodestore/alloc.rs
rkuris: I noticed that at some point but forgot to get back to it

storage/src/nodestore/alloc.rs
rkuris: That doesn't work because this is a const fn.

storage/src/nodestore/mod.rs
rkuris: Yes, this is a good idea. I don't know why the nonzero crate doesn't do that.

storage/src/nodestore/alloc.rs
rkuris: Created issue for resolution.

storage/src/nodestore/alloc.rs
rkuris: Clippy seems to disagree. `unwrap` is allowed in constexpr. It's not sufficient to use it in a constfn, and this can't be a constexpr.

storage/src/nodestore/mod.rs
rkuris: I'm going to skip this for now. I think I'm okay with `nonzero!(...).into()` for these use cases. Let's create an issue if you disagree.

storage/src/nodestore/alloc.rs
rkuris: Clippy seems to disagree

storage/src/nodestore/alloc.rs
demosdemon: what do you mean?

```rust
    /// Create a new `LinearAddress`, returns None if value is zero.
    #[inline]
    #[must_use]
    pub const fn new(addr: u64) -> Option<Self> {
        match NonZeroU64::new(addr) {
            Some(addr) => Some(LinearAddress(addr)),
            None => None,
        }
    }
```

This compiles without warning or the need to call `expect`

demosdemon (approved):

storage/src/nodestore/alloc.rs
```diff
@@ -113,9 +114,9 @@ pub const fn index_name(index: usize) -> &'static str {
 /// This is not usize because we can store this as a single byte
 pub type AreaIndex = u8;
 
-pub const NUM_AREA_SIZES: usize = AREA_SIZES.len();
-pub const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
-pub const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
+const NUM_AREA_SIZES: usize = AREA_SIZES.len();
+const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
+const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
```
qusuyan, line 119: We already have these constants defined here...

storage/src/nodestore/alloc.rs
```diff
@@ -147,12 +148,143 @@ pub fn area_size_to_index(n: u64) -> Result<AreaIndex, Error> {
...
+    #[must_use]
+    pub const fn is_aligned(self) -> bool {
+        self.0.get() % (Self::MIN_AREA_SIZE) == 0
+    }
+
+    /// The maximum area size available for allocation.
+    pub const MAX_AREA_SIZE: u64 = *AREA_SIZES.last().unwrap();
+
+    /// The minimum area size available for allocation.
+    pub const MIN_AREA_SIZE: u64 = *AREA_SIZES.first().unwrap();
```
qusuyan, line 192: Why do we need to define them here again?

storage/src/nodestore/alloc.rs
rkuris: Lets fix this with all the other issues with not having a newtype for an area size.

storage/src/nodestore/alloc.rs
rkuris: See above.

PR #1092, opened 2025-07-18
chore: skeleton of ffi proofs

alarso16:
This is a skeleton of what the interface might look like for range/change proofs. Rather than saying what's here, let me say what is missing:

- Config: Both the client and server should be able to set some config options, probably just request size limits
- An interface between x/sync and this (proposed interface in ava-labs/avalanchego#4064)
  - This would have to include some error parsing, probably something like `errors.As()` using the exported error types to determine whether to quit the syncer on some fatal error or simply re-request the data
  - It would also definitely have to convert between avalanchego types (e.g. `ids.ID => []byte`)
- Any implementation in Rust. The functions are only stubs
- Any decent comments describing what things mean on the Rust-side.

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -162,12 +162,14 @@ struct ProofResponse fwd_commit_change_proof(const struct DatabaseHandle *db,
                                              struct Value start_root,
                                              struct Value end_root,
                                              struct Value start_key,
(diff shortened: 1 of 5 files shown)


alarso16:
This is no longer the expected interface - closing in favor of a different design

PR #1097, opened 2025-07-20
chore: Add propose-on-propose test

rkuris:
During debugging, the ffi test TestProposeFromProposal failed, even though all the firewood tests passed. We shouldn't be using ffi to test some functionality in firewood, so I ported the test into the firewood repo.


alarso16 (approved):
This looks really clean. One test I really enjoyed making in Go was the deep propose test. I made N proposals, and committed them sequentially. You expect the same output as making a proposal, and committing it, N times. This might be a good addition

demosdemon (approved):

PR #1112, opened 2025-07-22
chore: enforce licensing in `ffi`

RodrigoVillar:


RodrigoVillar:
Closing this since we can just use https://github.com/ava-labs/firewood/blob/main/.github/check-license-headers.yaml. However, the `./scripts/lint.sh` might be useful in the future.

PR #1115, opened 2025-07-22
test: Port TestDeepPropose from go->rust

### LINKED ISSUE #1102: New Test: deep propose test
This looks really clean. One test I really enjoyed making in Go was the deep propose test. I made N proposals, and committed them sequentially. You expect the same output as making a proposal, and committing it, N times. This might be a good addition

_Originally posted by @alarso16 in https://github.com/ava-labs/firewood/pull/1097#pullrequestreview-3038828508_

rkuris:
Changed to test with 2 keys and values and 100 proposals, which seems like it exercises more of the code with less data.

Fixes #1102

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -833,6 +833,76 @@ mod test {
         }
     }
 
+    #[tokio::test]
+    async fn test_deep_propose() {
+        const NUM_KEYS: usize = 2;
+        const NUM_PROPOSALS: usize = 100;
+
+        let db = testdb().await;
+
+        // create NUM_KEYS * NUM_PROPOSALS keys and values
+        let (keys, vals): (Vec<_>, Vec<_>) = (0..NUM_KEYS * NUM_PROPOSALS)
+            .map(|i| {
+                (
+                    format!("key{i}").into_bytes(),
+                    Box::from(format!("value{i}").as_bytes()),
+                )
+            })
+            .unzip();
+
+        // create batches of NUM_KEYS keys and values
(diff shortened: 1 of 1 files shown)


firewood/src/db.rs
```diff
@@ -772,6 +772,76 @@ mod test {
...
+            .zip(vals.chunks(NUM_KEYS))
+            .map(|(k, v)| {
+                k.iter()
+                    .zip(v.iter())
+                    .map(|(k, v)| BatchOp::Put { key: k, value: v })
+                    .collect()
+            })
+            .collect();
+
+        // better me correct
```
alarso16: What is this comment supposed to mean?

firewood/src/db.rs
```diff
@@ -772,6 +772,76 @@ mod test {
...
+        }
+
+        // get the last committed revision
+        let committed = db
+            .revision(db.root_hash().await.unwrap().unwrap())
+            .await
+            .unwrap();
+
+        // check that all the keys and values are still present
+        for (k, v) in keys.iter().zip(vals.iter()) {
```
alarso16, line 911: Is it worth checking that the roots match?

firewood/src/db.rs
rkuris: It's a test to see if anyone is reading the code...
Just kidding. It was a typo.
```suggestion
        // better be correct
```

firewood/src/db.rs
rkuris: Can't hurt.

PR #1116, opened 2025-07-23
Rkuris/track unwritten nodes

rkuris:


storage/src/nodestore/mod.rs
```diff
@@ -350,11 +354,27 @@ pub trait RootReader {
...
+impl Clone for Committed {
+    fn clone(&self) -> Self {
+        Self {
+            deleted: self.deleted.clone(),
+            root_hash: self.root_hash.clone(),
+            root: self.root.clone(),
+            unwritten_nodes: AtomicUsize::new(
+                self.unwritten_nodes
+                    .load(std::sync::atomic::Ordering::Relaxed),
+            ),
```
demosdemon, line 375: Confirming that this is intentional the the cloned committed is using a new atomic integer and isn't re-using the integer from the previous cloned. The diverging committed nodes will have different unwritten_nodes counts.

storage/src/nodestore/mod.rs
```diff
@@ -532,20 +576,28 @@ impl<S: ReadableStorage> TryFrom<NodeStore<MutableProposal, S>>
...
+        let (root, root_hash, unwritten_count) =
+            nodestore.hash_helper(root, &mut Path::new(), None)?;
         #[cfg(not(feature = "ethhash"))]
-        let (root, root_hash) =
+        let (root, root_hash, unwritten_count) =
             NodeStore::<MutableProposal, S>::hash_helper(root, &mut Path::new())?;
 
         let immutable_proposal =
             Arc::into_inner(nodestore.kind).expect("no other references to the proposal");
+        // Use ManuallyDrop to prevent Drop from running since we're replacing the proposal
```
demosdemon, line 591: can you expand the comment to explain the implications of _not_ using `ManuallyDrop` here?

storage/src/nodestore/mod.rs
rkuris: Good catch, it's not correct. I'll remove this code as it's dead anyway; we never clone Committed.

storage/src/nodestore/persist.rs
```diff
@@ -456,7 +456,7 @@ impl NodeStore<Committed, FileBacked> {
 
                 // Decrement gauge for writes that have actually completed
                 if completed_writes > 0 {
-                    #[expect(clippy::cast_possible_truncation)]
+                    #[expect(clippy::cast_possible_truncation, clippy::cast_precision_loss)]
```
demosdemon: not understanding the lint problems here. some targets are triggering it and some aren't.

demosdemon: let's switch to allow and I'll research this next week (https://github.com/ava-labs/firewood/issues/1135)

PR #1118, opened 2025-07-23
test: add fuzz testing for checker, with fixes

### LINKED ISSUE #1110: Enhance differential fuzz tests to run the checker after completing
The checker should be used after running the differential fuzz tests, both in ethhash and non-ethhash cases.

rkuris:
Fixes #1110

Passes consistently with the following fixes:
 - The partial path for the branch was not added when computing the child hashes
 - Always render the path, not the partial path before this node

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -7,10 +7,8 @@
 )]
 
 use crate::merkle::Merkle;
-use crate::proof::{Proof, ProofNode};
-use crate::range_proof::RangeProof;
 use crate::stream::MerkleKeyValueStream;
-use crate::v2::api::{self, KeyType, ValueType};
+use crate::v2::api::{self, FrozenProof, FrozenRangeProof, KeyType, ValueType};
 pub use crate::v2::api::{Batch, BatchOp};
 
 use crate::manager::{RevisionManager, RevisionManagerConfig};
@@ -108,10 +106,7 @@ impl api::DbView for HistoricalRev {
         Ok(merkle.get_value(key.as_ref())?)
     }
 
-    async fn single_key_proof<K: api::KeyType>(
-        &self,
-        key: K,
-    ) -> Result<Proof<ProofNode>, api::Error> {
+    async fn single_key_proof<K: api::KeyType>(&self, key: K) -> Result<FrozenProof, api::Error> {
         let merkle = Merkle::from(self);
         merkle.prove(key.as_ref()).map_err(api::Error::from)
     }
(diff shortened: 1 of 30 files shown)


storage/src/checker/mod.rs
```diff
@@ -374,12 +377,12 @@ mod test {
         err,
         CheckerError::HashMismatch {
             address,
-            partial_path,
+            path,
```
rkuris, line 387: Debugging this would have been easier if we constructed the error and used assert_eq! on them, as the debug output would highlight the difference.

firewood/src/db.rs
```diff
@@ -767,6 +778,48 @@ mod test {
...
+                let value: [u8; 32] = rng.borrow_mut().random();
+                batch.push(BatchOp::Put { key, value });
+                trace!("batch: {batch:?}");
+                batch
+            });
+            let proposal = db.propose(batch).await.unwrap();
+            proposal.commit().await.unwrap();
+
+            // check the database for consistency, sometimes checking the hashes
+            let hash_check = rng.borrow_mut().random();
```
qusuyan, line 815: I think we should just always check the hashes - I don't expect the checker to do any more when hash_check is false

storage/src/checker/mod.rs
```diff
@@ -111,8 +112,10 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         if hash_check {
             let hash = hash_node(&node, &path_prefix);
             if hash != subtree_root_hash {
+                let mut path = path_prefix.clone();
```
qusuyan, line 122: we probably don't need the extra clone()

storage/src/checker/mod.rs
```diff
@@ -278,7 +281,7 @@ mod test {
             value: Box::new([3, 4, 5]),
         });
         let leaf_addr = LinearAddress::new(high_watermark).unwrap();
-        let leaf_hash = hash_node(&leaf, &Path::from_nibbles_iterator([0u8, 1].into_iter()));
+        let leaf_hash = hash_node(&leaf, &Path::from_nibbles_iterator([0u8, 0, 1].into_iter()));
```
qusuyan, line 291: Seems that I have constantly mistaken what partial path really means. Thanks for pointing this out!

firewood/src/db.rs
rkuris: Let's check both paths by disabling it sometimes. It might run a little faster too, giving us more fuzzing.

storage/src/checker/mod.rs
```diff
@@ -256,41 +259,44 @@ mod test {
...
     // -------------------------
+    // | partial path: []      |
     // |     |  X  |  X  | ... |    Root node
     // -------------------------
     //    |
     //    V
     // -------------------------
-    // |  X  |     |  X  | ... |    Branch node
+    // | partial path: [9]     |
+    // |  X  |     |  X  | ... |    Branch node (path: 0x09)
```
rkuris: Why are we changing this in this PR?

This feels really random. Is that the intention?

Seems like having something more orthogonal would make understanding this structure a bit easier going forward.

PR #1119, opened 2025-07-23
chore: refactor verifying value digests

demosdemon:
In #1117, I have a need to reuse this verification method. I have refactored it out into separate steps to make that easier.

### DIFF
--- a/storage/src/hashednode.rs
+++ b/storage/src/hashednode.rs
@@ -6,14 +6,13 @@
     reason = "Found 1 occurrences after enabling the lint."
 )]
 
+use crate::{BranchNode, HashType, LeafNode, Node, Path};
+use sha2::{Digest, Sha256};
+use smallvec::SmallVec;
 use std::{
     iter::{self},
     ops::Deref,
 };
-
-use smallvec::SmallVec;
-
-use crate::{BranchNode, HashType, LeafNode, Node, Path};
 
 /// Returns the hash of `node`, which is at the given `path_prefix`.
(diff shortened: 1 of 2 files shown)


storage/src/hashednode.rs
```diff
@@ -113,6 +112,23 @@ impl<T> Deref for ValueDigest<T> {
     }
 }
 
+impl<T: AsRef<[u8]>> ValueDigest<T> {
```
rkuris, line 115: nit: feels like this could be a PartialEq implementation

rkuris (approved):
This code is much better than the old code.

nit: move verify closer to it's caller

storage/src/hashednode.rs
demosdemon: Possibly, but that `Sha256::digest` is expensive so I'd rather not hide that under an operator.

demosdemon:
> This code is much better than the old code.
> 
> nit: move verify closer to it's caller

I'd actually like to put `verify_opt_value_digest` in `storage` with the `verify` method, but the api error type is only in firewood and I didn't want to carry an error representation over just yet. That may happen later during cleanup.

rkuris:
> I'd actually like to put `verify_opt_value_digest` in `storage` with the `verify` method, but the api error type is only in firewood and I didn't want to carry an error representation over just yet. That may happen later during cleanup.

I've fought hard to keep storage as small as possible. Merkle is not there, and I don't think proof belongs there either.

PR #1123, opened 2025-07-23
feat: add `Children<T>` type alias

demosdemon:
This commit introduces a new type alias `Children<T>` which is a fixed-size array of `Option<T>`. This is convenient for all of the places where we represent the children of a branch node in the trie.

Also included is a method to construct the array in const context, allowing for easy creation of `Children<T>` instances.


storage/src/hashers/ethhash.rs
```diff
@@ -170,31 +176,17 @@ impl<T: Hashable> Preimage for T {
...
+                };
             }
 
-            if let Some(digest) = self.value_digest() {
-                if is_account {
-                    rlp.append_empty_data();
-                } else {
-                    rlp.append(&*digest);
-                }
+            if !is_account && let Some(digest) = self.value_digest() {
```
rkuris: let chain is a very recent addition (1.88 I think) which may affect our MSRV.

storage/src/hashers/merkledb.rs
```diff
@@ -35,12 +35,20 @@ impl<T: Hashable> Preimage for T {
     fn write(&self, buf: &mut impl HasUpdate) {
         let children = self.children();
 
-        let num_children = children.clone().count() as u64;
+        // This will use auto-vectorization to sum the non-zero children
+        let num_children = children
+            .iter()
+            .map(Option::is_some)
+            .map(u64::from)
+            .sum::<u64>();
```
rkuris: why not:
```suggestion
            .filter(Option::is_some)
            .count() as u64
```

storage/src/node/branch.rs
```diff
@@ -411,15 +420,47 @@ impl BranchNode {
...
+            match child {
+                None => {}
+                Some(Child::Node(_)) => unreachable!("TODO make unreachable"),
+                Some(Child::AddressWithHash(address, _)) => _ = slot.replace(*address),
+                Some(Child::MaybePersisted(maybe_persisted, _)) => {
+                    // For MaybePersisted, we need the address if it's persisted
+                    if let Some(addr) = maybe_persisted.as_linear_address() {
+                        slot.replace(addr);
+                    } else {
+                        // should we panic here?
```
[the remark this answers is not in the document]
rkuris: Thanks copilot. We should at least doc what the behavior is for unpersisted children.

storage/src/hashers/ethhash.rs
demosdemon: Let chaining is part of [2024 edition](https://doc.rust-lang.org/beta/edition-guide/rust-2024/let-chains.html) (+1.85).

storage/src/hashers/merkledb.rs
demosdemon: Using the filter does not take advantage of auto-vectorization which the compiler can do if the inputs are all a static sized vector.

Even in debug mode, the compiler optimizes one differently than the other (see [godbolt](https://rust.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:26,endLineNumber:10,positionColumn:26,positionLineNumber:10,selectionStartColumn:26,selectionStartLineNumber:10,startColumn:26,startLineNumber:10),source:'type+Children%3CT%3E+%3D+%5BOption%3CT%3E%3B+16%5D%3B%0A%0Apub+fn+children_len_v1(c:+%26Children%3C%5Bu64%3B+2%5D%3E)+-%3E+usize+%7B%0A++++c.iter().filter(%7Cv%7C+v.is_some()).count()%0A%7D%0A%0Apub+fn+children_len_v2(c:+%26Children%3C%5Bu64%3B+2%5D%3E)+-%3E+usize+%7B%0A++++c.iter()%0A++++++++.map(Option::is_some)%0A++++++++.map(usize::from)%0A++++++++.sum::%3Cusize%3E()%0A%7D%0A'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1880,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'',overrides:!((name:edition,value:'2024')),selection:(endColumn:1,endLineNumber:624,positionColumn:1,positionLineNumber:624,selectionStartColumn:1,selectionStartLineNumber:583,startColumn:1,startLineNumber:583),source:1),l:'5',n:'0',o:'+rustc+1.88.0+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4):

```rust
type Children<T> = [Option<T>; 16];

pub fn children_len_v1(c: &Children<[u64; 2]>) -> usize {
    c.iter().filter(|v| v.is_some()).count()
}

pub fn children_len_v2(c: &Children<[u64; 2]>) -> usize {
    c.iter()
        .map(Option::is_some)
        .map(usize::from)
        .sum::<usize>()
}
```

```text

example::children_len_v1::hb07441eb8d9afdfc:
        push    rax
        mov     qword ptr [rsp], rdi
        mov     esi, 16
        call    qword ptr [rip + core::slice::<impl [T]>::iter::hc4c01f1a8bd6469a@GOTPCREL]
        mov     rdi, rax
        mov     rsi, rdx
        call    qword ptr [rip + core::iter::traits::iterator::Iterator::filter::h2db63312c004ae16@GOTPCREL]
        mov     rdi, rax
        mov     rsi, rdx
        call    qword ptr [rip + <core::iter::adapters::filter::Filter<I,P> as core::iter::traits::iterator::Iterator>::count::h78eb81182a69c198@GOTPCREL]
        pop     rcx
        ret

example::children_len_v1::{{closure}}::hc5058959d1490b3e:
        sub     rsp, 24
        mov     qword ptr [rsp + 8], rdi
        mov     qword ptr [rsp + 16], rsi
        mov     rdi, qword ptr [rsi]
        call    qword ptr [rip + core::option::Option<T>::is_some::h771f3ff5d4044778@GOTPCREL]
        and     al, 1
        add     rsp, 24
        ret

example::children_len_v2::hc589cadf7465d79a:
        push    rax
        mov     qword ptr [rsp], rdi
        mov     esi, 16
        call    qword ptr [rip + core::slice::<impl [T]>::iter::hc4c01f1a8bd6469a@GOTPCREL]
        mov     rdi, rax
        mov     rsi, rdx
        call    qword ptr [rip + core::iter::traits::iterator::Iterator::map::h0788c100c8261235@GOTPCREL]
        mov     rdi, rax
        mov     rsi, rdx
        call    qword ptr [rip + core::iter::traits::iterator::Iterator::map::h52ba7c822e44c1a8@GOTPCREL]
        mov     rdi, rax
        mov     rsi, rdx
        call    qword ptr [rip + core::iter::traits::iterator::Iterator::sum::hbe81f184f0348e76@GOTPCREL]
        pop     rcx
        ret
```

But, it turns out that it's irrelevant with optimizations enables as rustc is smart enough

```rust
example::children_len_v1::hb07441eb8d9afdfc:
        mov     rax, qword ptr [rdi + 24]
        add     rax, qword ptr [rdi]
        add     rax, qword ptr [rdi + 48]
        add     rax, qword ptr [rdi + 72]
        add     rax, qword ptr [rdi + 96]
        add     rax, qword ptr [rdi + 120]
        add     rax, qword ptr [rdi + 144]
        add     rax, qword ptr [rdi + 168]
        add     rax, qword ptr [rdi + 192]
        add     rax, qword ptr [rdi + 216]
        add     rax, qword ptr [rdi + 240]
        add     rax, qword ptr [rdi + 264]
        add     rax, qword ptr [rdi + 288]
        add     rax, qword ptr [rdi + 312]
        add     rax, qword ptr [rdi + 336]
        add     rax, qword ptr [rdi + 360]
        ret

example::children_len_v2::hc589cadf7465d79a:
        mov     rax, qword ptr [rdi + 24]
        add     rax, qword ptr [rdi]
        add     rax, qword ptr [rdi + 48]
        add     rax, qword ptr [rdi + 72]
        add     rax, qword ptr [rdi + 96]
        add     rax, qword ptr [rdi + 120]
        add     rax, qword ptr [rdi + 144]
        add     rax, qword ptr [rdi + 168]
        add     rax, qword ptr [rdi + 192]
        add     rax, qword ptr [rdi + 216]
        add     rax, qword ptr [rdi + 240]
        add     rax, qword ptr [rdi + 264]
        add     rax, qword ptr [rdi + 288]
        add     rax, qword ptr [rdi + 312]
        add     rax, qword ptr [rdi + 336]
        add     rax, qword ptr [rdi + 360]
        ret
```

storage/src/node/branch.rs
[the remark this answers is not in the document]
demosdemon: Yeah, I was unsure of what to do here. The previous implmentation panicked for `Child::Node(_)` and I left that but also silently ignored when the maybe persisted doesn't have an address. Context matters too. During serialization, it should be an error to try to use `children_addresses` when there's no address yet. But some other contexts might not care.

storage/src/hashers/ethhash.rs
demosdemon: you're right, this specific feature is new to 1.88. For future reference, I never noticed because this is gated under the ethhash feature and I missed that when I did my test with 1.85.


```console
$ cargo +1.85.1 clippy --workspace --all-targets -F ethhash,logger
    Checking firewood-storage v0.0.9 (/Users/brandon.leblanc/src/firewood-pull-requests/storage)
   Compiling firewood-fwdctl v0.0.9 (/Users/brandon.leblanc/src/firewood-pull-requests/fwdctl)
   Compiling firewood-ffi v0.0.9 (/Users/brandon.leblanc/src/firewood-pull-requests/ffi)
error[E0658]: `let` expressions in this position are unstable
   --> storage/src/hashers/ethhash.rs:183:31
    |
183 |             if !is_account && let Some(digest) = self.value_digest() {
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `firewood-storage` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `firewood-storage` (lib test) due to 1 previous error
```

storage/src/hashers/ethhash.rs
```diff
@@ -170,31 +171,23 @@ impl<T: Hashable> Preimage for T {
...
-                    rlp.append_empty_data();
-                } else {
-                    rlp.append(&*digest);
-                }
+            // For branch nodes, we need to append the value as the 17th element in the RLP list.
+            // However, account nodes (depth 32) handle values differently - they don't store
+            // the value directly in the branch node, but rather in the account structure itself.
+            // This is because account nodes have special handling where the storage root hash
+            // gets replaced in the account data structure during serialization.
+            let digest = (!is_account).then(|| self.value_digest()).flatten();
```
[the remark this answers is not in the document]
demosdemon: Open to other comments here. I disagree with the bot but I'm also very familiar with the syntax.

rkuris (approved):

Issue #1108, opened 2025-07-22
Checker found error in an existing image


qusuyan:
The checker finds a hash mismatch error on a supposedly good firewood image. Some information about the "erroneous" node:
```
Path Prefix: 0x00002782
Parent Stored Hash: Hash(fad487c098e147fdd7cd0e427823b343f7b4e041dea946a782980ec7f4d27a65)
Computed Hash: Hash(0f52edaf6bd16fdfab9eab1a75ca0e9a0665a5754fb2c1e6c0c5758343b067f3)
Leaf Node Content: Leaf([Leaf 9 0 d e c d 9 5 4 8 b 6 2 a 8 d 6 0 3 4 5 a 9 8 8 3 8 6 f c 8 4 b a 6 b c 9 5 4 8 4 0 0 8 f 6 3 6 2 f 9 3 1 6 0 e f 3 e 5 6 3  94c0fecbc9f3c8f6d3fd304cd030e65692abf7946a])
``` 
The checker reports this error within 1s of running. It is not clear if the image is wrong or the checker has a bug.

rkuris:
On the surface, this appears to be a checker bug. The root hash calculation is correct.

rkuris:
It is indeed a checker bug. For ethhash, we need to use `hash_helper`'s logic before calling `hash_node`. They are pretty similar when ethhash is disabled.

rkuris:
---- db::test::fuzz_checker stdout ----
Seed 10768121300286458725: to rerun with this data, export FIREWOOD_TEST_SEED=10768121300286458725

thread 'db::test::fuzz_checker' panicked at firewood/src/db.rs:822:25:
error: Hash mismatch for node 9 5 c  at address 15488: parent stored ad7835f2822dfabf901683324cce535136ca8c419b634d6f9c86d9f581993206, computed 2c16db2985d077e62a29e68f31dab4207844d869f1adf9e908eaf389cb76b04f

![Image](https://github.com/user-attachments/assets/4ae207ba-a8fb-44eb-82d0-99c43ca1c281)

rkuris:
The actual key for the failed node is 95c plus the node's partial path of 9924b9ba5e9f7a6fa4b03100eb15caf00077d5002c483c8761983701c4fd, so it's clear that the checker's output does not contain the partial path.

rkuris:
I think we can close this now, since the issue has been isolated to the checker, and there are tasks to get ethhash working with the checker in the single-child case.

PR #1127, opened 2025-07-23
feat(checker): disable buggy ethhash checker

qusuyan:
Current checker has bug with ethhash #1108. Disable for now.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -14,7 +14,8 @@ pub use crate::v2::api::{Batch, BatchOp};
 use crate::manager::{RevisionManager, RevisionManagerConfig};
 use async_trait::async_trait;
 use firewood_storage::{
-    Committed, FileBacked, FileIoError, HashedNodeReader, ImmutableProposal, NodeStore, TrieHash,
+    CheckOpt, Committed, FileBacked, FileIoError, HashedNodeReader, ImmutableProposal, NodeStore,
+    TrieHash,
 };
 use metrics::{counter, describe_counter};
 use std::io::Write;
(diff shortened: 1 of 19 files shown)


storage/src/checker/mod.rs
```diff
@@ -37,6 +37,10 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
     /// Panics if the header has too many free lists, which can never happen since freelists have a fixed size.
     // TODO: report all errors, not just the first one
     pub fn check(&self, opt: CheckOpt) -> Result<(), CheckerError> {
+        if cfg!(feature = "ethhash") {
+            unimplemented!("ethhash is not supported yet");
+        }
```
demosdemon, line 42: This shouldn't panic here as it will cause the tests to fail. broken tests should be annotated with `#[ignore]` and/or this function should not be compiled in for ethhash.

qusuyan: Let me ignore the tests for ethhash - this is temporary as we are currently working on a fix for ethhash. It may take a while so we just want to block it for now.

storage/src/nodestore/alloc.rs
```diff
@@ -776,19 +776,22 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
 }
 
 #[cfg(test)]
-#[expect(clippy::unwrap_used, clippy::indexing_slicing)]
+#[expect(clippy::unwrap_used)]
 pub mod test_utils {
     use super::*;
 
+    #[cfg(not(feature = "ethhash"))]
```
demosdemon: It is preferred to `#[ignore]` instead of suppressing compilation as it will cause weird issues with other lints. (not here but on the tests themselves, I picked a bad line to make this suggestion)

```suggestion
    #[cfg_attr(feature = "ethhash", ignore = "https://github.com/ava-labs/firewood/issues/1108")]
```

firewood/src/db.rs
```diff
@@ -779,6 +779,7 @@ mod test {
     }
 
     #[tokio::test]
+    #[cfg(not(feature = "ethhash"))]
```
demosdemon:
```suggestion
    #[cfg_attr(
        feature = "ethhash",
        ignore = "https://github.com/ava-labs/firewood/issues/1108"
    )]
```

storage/src/nodestore/alloc.rs
qusuyan: that is new - I dont know you can ignore only when certain features are enabled

demosdemon (approved):
🚀

PR #1130, opened 2025-07-23
fix(checker): fix checker with ethhash

qusuyan:
With ethhash there is a special case checker did not handle: when an account has only one child. This diff aims to fix that bug and (hopefully) refactor `hash_helper` for ethhash for better understanding.

### DIFF
--- a/storage/src/nodestore/hash.rs
+++ b/storage/src/nodestore/hash.rs
@@ -10,20 +10,11 @@ use crate::hashednode::hash_node;
 use crate::linear::FileIoError;
 use crate::logger::trace;
 use crate::node::Node;
-use crate::{Child, HashType, Path, ReadableStorage, SharedNode};
+use crate::{Child, HashType, MaybePersistedNode, NodeStore, Path, ReadableStorage, SharedNode};
 
-#[cfg(feature = "ethhash")]
 use super::NodeReader;
 #[cfg(feature = "ethhash")]
-use crate::node::persist::MaybePersistedNode;
-#[cfg(feature = "ethhash")]
 use std::ops::Deref;
-
-use std::collections::HashMap;
-use std::sync::Arc;
-
-use super::alloc::LinearAddress;
-use super::{ImmutableProposal, NodeStore};
 
 /// Classified children for ethereum hash processing
 #[cfg(feature = "ethhash")]
@@ -32,7 +23,10 @@ pub(super) struct ClassifiedChildren<'a> {
     pub(super) hashed: Vec<(usize, (MaybePersistedNode, &'a mut HashType))>,
 }
 
-impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
+impl<T, S: ReadableStorage> NodeStore<T, S>
+where
+    NodeStore<T, S>: NodeReader,
+{
     /// Helper function to classify children for ethereum hash processing
     /// We have some special cases based on the number of children
     /// and whether they are hashed or unhashed, so we need to classify them.
@@ -62,7 +56,7 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
                             acc.hashed.push((idx, (maybe_persisted_node, h)));
                         } else {
                             // If not persisted, we need to get the node to hash it
-                            if let Ok(node) = maybe_persisted.as_shared_node(self) {
+                            if let Ok(node) = maybe_persisted.as_shared_node(&self) {
                                 acc.unhashed.push((idx, node.deref().clone()));
                             }
                         }
@@ -76,13 +70,12 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
     /// Hashes `node`, which is at the given `path_prefix`, and its children recursively.
     /// Returns the hashed node and its hash.
     pub(super) fn hash_helper(
-        &mut self,
+        #[cfg(feature = "ethhash")] &self,
         mut node: Node,
         path_prefix: &mut Path,
-        new_nodes: &mut HashMap<LinearAddress, (u8, SharedNode)>,
         #[cfg(feature = "ethhash")] fake_root_extra_nibble: Option<u8>,
-    ) -> Result<(LinearAddress, HashType), FileIoError> {
-        // If this is a branch, find all unhashed children and recursively call hash_helper on them.
+    ) -> Result<(MaybePersistedNode, HashType), FileIoError> {
+        // If this is a branch, find all unhashed children and recursively hash them.
         trace!("hashing {node:?} at {path_prefix:?}");
         if let Node::Branch(ref mut b) = node {
             // special case code for ethereum hashes at the account level
@@ -143,7 +136,7 @@ impl<S: ReadableStorage> NodeStore<Arc<ImmutableProposal>, S> {
                 // If this is empty or already hashed, we're done
                 // Empty matches None, and non-Node types match Some(None) here, so we want
                 // Some(Some(node))
-                let Some(Some(child_node)) = child.as_mut().map(|child| child.as_mut_node()) else {
+                let Some(child_node) = child.as_mut().and_then(|child| child.as_mut_node()) else {
                     continue;
                 };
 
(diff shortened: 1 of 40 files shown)


storage/src/nodestore/hash.rs
```diff
@@ -93,27 +95,26 @@ where
...
                 if hashed.len() == 1 {
                     // we were left with one hashed node that must be rehashed
-                    let invalidated_node = hashed.first_mut().expect("hashed is not empty");
+                    let (invalidated_node_idx, (invalidated_node, invalidated_hash)) =
+                        hashed.first_mut().expect("hashed is not empty");
                     // Extract the address from the MaybePersistedNode
-                    let addr = invalidated_node
-                        .1
-                        .0
+                    let addr: crate::LinearAddress = invalidated_node
```
demosdemon, line 101: nit: type hint should be redundant.

storage/src/nodestore/hash.rs
```diff
@@ -93,27 +95,26 @@ where
                 trace!("hashed {hashed:?} unhashed {unhashed:?}");
                 if hashed.len() == 1 {
                     // we were left with one hashed node that must be rehashed
-                    let invalidated_node = hashed.first_mut().expect("hashed is not empty");
+                    let (invalidated_node_idx, (invalidated_node, invalidated_hash)) =
+                        hashed.first_mut().expect("hashed is not empty");
```
demosdemon, line 99: we can avoid this expect.

```rust
match &mut hashed[..] {
    [] => todo!("empty"),
    [(invalidated_node_idx, (invalidated_node, invalidated_hash))] => todo!("single element"),
    [_, ..] => todo!("more than one element"),
}
```

example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=95cec1e1961bcb015cf08c5dc4318553

storage/src/nodestore/hash.rs
```diff
@@ -93,27 +95,26 @@ where
...
-                        path_prefix.0.push(invalidated_node.0 as u8);
+                        path_prefix.0.push(*invalidated_node_idx as u8);
                     }
                     let hash = hash_node(&hashable_node, path_prefix);
                     path_prefix.0.truncate(original_length);
-                    *invalidated_node.1.1 = hash;
+                    **invalidated_hash = hash;
                 }
                 // handle the single-child case for an account special below
                 if hashed.is_empty() && unhashed.len() == 1 {
```
demosdemon, line 120: same could be done here

```rust
match (hashed, unhashed) {
    ([], &[(only, _]) => Some(only),
    _ => None,
}
```

storage/src/nodestore/hash.rs
```diff
@@ -93,27 +95,26 @@ where
...
-                    let invalidated_node = hashed.first_mut().expect("hashed is not empty");
+                    let (invalidated_node_idx, (invalidated_node, invalidated_hash)) =
+                        hashed.first_mut().expect("hashed is not empty");
                     // Extract the address from the MaybePersistedNode
-                    let addr = invalidated_node
-                        .1
-                        .0
+                    let addr: crate::LinearAddress = invalidated_node
                         .as_linear_address()
                         .expect("hashed node should be persisted");
```
demosdemon, line 103: ~this should probably return an error and not panic~

rather, it looks like we could put the LA in the vector when we first discover it so we don't have to unwrap it again later.

storage/src/nodestore/hash.rs
```diff
@@ -192,4 +193,24 @@ where
...
+        node: &Node,
+        path_prefix: &Path,
+        have_peers: bool,
+    ) -> HashType {
+        if path_prefix.0.len() == 65 && !have_peers {
+            // This is the special case when this node is the only child of an account
+            //  - 64 nibbles for account + 1 nibble for its position in account branch node
+            let mut fake_root = node.clone();
+            fake_root.update_partial_path(Path::from_nibbles_iterator(
+                std::iter::once(*path_prefix.0.last().expect("path_prefix not empty"))
```
demosdemon: `Option<_>` implements IntoIterator that yields zero or one elements. We can use that here to avoid this expect:

```rust
path_prefix.0.last().copied().into_iter().chain(fake_root.partial_path().0.iter().copied())
```

Coincidentally, this is exactly how `once` is implemented:

https://doc.rust-lang.org/stable/src/core/iter/sources/once.rs.html#56-58

storage/src/nodestore/hash.rs
```diff
@@ -58,9 +58,10 @@ where
                             acc.hashed.push((idx, (maybe_persisted_node, h)));
                         } else {
                             // If not persisted, we need to get the node to hash it
-                            if let Ok(node) = maybe_persisted.as_shared_node(&self) {
-                                acc.unhashed.push((idx, node.deref().clone()));
-                            }
+                            let node = maybe_persisted
+                                .as_shared_node(&self)
+                                .expect("will never fail for unpersisted nodes");
+                            acc.unhashed.push((idx, node.deref().clone()));
```
demosdemon, line 64: this is a case that `MaybePersisted` should help us avoid. MaybePersisted might need a method that lets you destructure to either the linear address or the shared node so that you can do either/or without trying to panic unecessarily.

```rust
match maybe_persisted.something() {
    MaybePersisted::Unpersisted(shared_node) => todo!(),
    MaybePersisted::Persisted(linear_address) => todo!(),
}
```

littering the code with panics when we can avoid it is bad.

demosdemon: although, on second look, this appears to be the pattern:

```rust
if let Ok(addr) = maybe_persisted. as_shared_node(&self) {
    acc.unhashed.push((idx, node.deref().clone())); // why are we deep cloning the node and not the arc of the node?
} else {
    acc.hashed.push((idx, (maybe_persisted_node.clone(), h)));
}
```

Curious why we are keeping the maybe_persisted around for just the addr but this would avoid the panic _here_ but not later.

-- 

but I did also just notice that we are de-referencing the arc around the node and then cloning it. IIUC, that will do a deep clone of the node and not the arc around the node.

demosdemon:
although, now that I've reviewed the entire file and not just what's in the diff context, I am very concerned with this:

https://github.com/ava-labs/firewood/blob/653af24be5d8be09e506d067f40a270508555707/storage/src/nodestore/hash.rs#L168

We shouldn't be replacing the child with a new MaybePersistedNode (unless it was never one to begin with).  Instead, we should be calling `persist_at` on the existing `MaybePersistedNode`.

https://github.com/ava-labs/firewood/blob/08383cf25325cdbe727213d2848629bdf0adcf2b/storage/src/node/persist.rs#L124-L136

storage/src/nodestore/hash.rs
qusuyan: Seems that most comments related to this diff are small changes I made to `hash_helper` that I made when trying to understand what it is doing. I am planning to reorganize it with a different PR #1137 and will address the comments there.

storage/src/nodestore/hash.rs
qusuyan: Addressed here: https://github.com/ava-labs/firewood/pull/1137/files#diff-4af05feae47a13c859269dcdb6d6a9d88d14f935b6f593e9515bd294b1271ec1R94

storage/src/nodestore/hash.rs
qusuyan: Addressed here: https://github.com/ava-labs/firewood/pull/1137/files#diff-4af05feae47a13c859269dcdb6d6a9d88d14f935b6f593e9515bd294b1271ec1R99

storage/src/nodestore/hash.rs
qusuyan: We no longer need this: https://github.com/ava-labs/firewood/pull/1137/files#diff-4af05feae47a13c859269dcdb6d6a9d88d14f935b6f593e9515bd294b1271ec1R120

storage/src/nodestore/hash.rs
qusuyan: Good call - addressed here: http://github.com/ava-labs/firewood/pull/1137/files#diff-4af05feae47a13c859269dcdb6d6a9d88d14f935b6f593e9515bd294b1271ec1R28

storage/src/nodestore/hash.rs
qusuyan: The changed ethhash should resolve most of copying :)

PR #1131, opened 2025-07-23
chore: Checker test cleanups

rkuris:
- Make each nibble and byte different for easier identification
 - Add a partial path to the root node
 - Return TestTrie structure for all the metadata for the test trie
 - Handle structural changes to test data more gracefully
 - Implement PartialEq on CheckerError
 - Replace ascii gen_test with mermaid diagram

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -268,57 +268,58 @@ mod test {
     use crate::nodestore::alloc::{AREA_SIZES, FreeLists};
     use crate::{BranchNode, Child, LeafNode, NodeStore, Path, hash_node};
 
-    // set up a basic trie:
-    // -------------------------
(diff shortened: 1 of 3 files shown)


demosdemon (approved):

storage/src/checker/mod.rs
```diff
@@ -268,57 +268,58 @@ mod test {
...
+    struct TestTrie {
+        nodes: Vec<(Node, LinearAddress)>,
+        high_watermark: u64,
+        root_address: LinearAddress,
+        root_hash: HashType,
+    }
+
+    /// Generate a test trie with the following structure:
+    ///
+    #[cfg_attr(doc, aquamarine)]
```
demosdemon, line 281: this is great. I didn't realize there was an attr to render the mermaid diagrams

PR #1132, opened 2025-07-23
ci: add a job to invoke ingored tests

demosdemon:
It's understandable that this might be a controversial change. However, I believe that having a job to invoke the ignored tests is beneficial so that we can still run them and ensure they are still broken. In the future, we may even want to ensure this job fails so that we can detect when a test is no longer broken and can be removed from the ignore list.

Currently, this will run on every pull request and merge to main, but we can adjust this as needed.


demosdemon:
nvm, after seeing the behavior with this enabled; I am not a fan. I will rethink.

PR #1134, opened 2025-07-24
feat: make NodeStore more generic

demosdemon:
This commit enhances the `NodeStore` structure to be more generic, allowing it to work with different storage backends. In a future change, I have tests that execute a write-commit-read loop using `MemStore`. However, this does not work without this change as `NodeStore` currently has many methods that are specific to the `FileStore` implementation.

Passing in an `Arc` is not required as we're not cloning the arc. This makes it easier to use the `NodeStore` without needing to wrap it in an `Arc` every time.

A `fork` method is also added to `NodeStore` for easier cloning of the store into a new mutable instance.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -1,19 +1,6 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-#![expect(
-    clippy::missing_errors_doc,
-    reason = "Found 6 occurrences after enabling the lint."
-)]
-#![expect(
-    clippy::missing_panics_doc,
-    reason = "Found 1 occurrences after enabling the lint."
-)]
-#![expect(
-    clippy::too_many_lines,
-    reason = "Found 1 occurrences after enabling the lint."
-)]
-
 use crate::proof::{Proof, ProofError, ProofNode};
 #[cfg(test)]
 use crate::range_proof::RangeProof;
@@ -26,7 +13,7 @@ use crate::v2::api::{self, FrozenRangeProof};
 use firewood_storage::{
     BranchNode, Child, FileIoError, HashType, Hashable, HashedNodeReader, ImmutableProposal,
     IntoHashType, LeafNode, MaybePersistedNode, MutableProposal, NibblesIterator, Node, NodeStore,
-    Path, ReadableStorage, SharedNode, TrieReader, ValueDigest,
+    Parentable, Path, ReadableStorage, SharedNode, TrieReader, ValueDigest,
 };
 #[cfg(test)]
 use futures::{StreamExt, TryStreamExt};
@@ -178,6 +165,10 @@ impl<T: TrieReader> Merkle<T> {
 
     /// Returns a proof that the given key has a certain value,
     /// or that the key isn't in the trie.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the trie is empty or an error occurs while reading from storage.
     pub fn prove(&self, key: &[u8]) -> Result<FrozenProof, ProofError> {
         let Some(root) = self.root() else {
             return Err(ProofError::Empty);
@@ -216,6 +207,7 @@ impl<T: TrieReader> Merkle<T> {
     }
 
     /// Verify a proof that a key has a certain value, or that the key isn't in the trie.
+    #[expect(clippy::missing_errors_doc)]
     pub fn verify_range_proof<V: AsRef<[u8]>>(
         &self,
         _proof: &Proof<impl Hashable>,
@@ -442,6 +434,17 @@ impl<T: HashedNodeReader> Merkle<T> {
     }
 }
 
+impl<F: Parentable, S: ReadableStorage> Merkle<NodeStore<F, S>> {
+    /// Forks the current Merkle trie into a new mutable proposal.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the nodestore cannot be created. See [`NodeStore::new`].
+    pub fn fork(&self) -> Result<Merkle<NodeStore<MutableProposal, S>>, FileIoError> {
+        NodeStore::new(&self.nodestore).map(Into::into)
+    }
+}
+
 impl<S: ReadableStorage> TryFrom<Merkle<NodeStore<MutableProposal, S>>>
     for Merkle<NodeStore<Arc<ImmutableProposal>, S>>
 {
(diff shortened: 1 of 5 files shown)


firewood/src/merkle.rs
```diff
@@ -216,6 +207,7 @@ impl<T: TrieReader> Merkle<T> {
     }
 
     /// Verify a proof that a key has a certain value, or that the key isn't in the trie.
+    #[expect(clippy::missing_errors_doc)]
```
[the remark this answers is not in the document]
demosdemon: doc is added in a future pr with the implementation

demosdemon:
## Code Review for PR #1134: feat: make NodeStore more generic

### Overview
This PR enhances the `NodeStore` structure to be more generic by:
- Removing the requirement to pass `Arc` references to `NodeStore::new()`
- Adding support for different storage backends (not just `FileStore`)
- Adding a `fork()` method to `Merkle` for easier cloning
- Implementing generic storage support for commit operations

### ✅ Strengths
- **Better API design**: Removing the `Arc` requirement simplifies the API and makes it more flexible
- **Improved testability**: Enables write-commit-read loops with `MemStore` for testing
- **Clean implementation**: Changes are focused and well-structured
- **Good test coverage**: Added tests for new functionality and edge cases

### 🔍 Code Quality Analysis

#### Architecture & Design
- The change from `Arc<NodeStore<F, S>>` to `&NodeStore<F, S>` is a good design decision that reduces unnecessary Arc wrapping
- The `fork()` method provides a clean abstraction for creating mutable proposals
- Generic storage support is properly implemented with appropriate trait bounds

#### Performance Considerations
- No performance regressions expected; removing Arc requirement may slightly improve performance
- The `io-uring` feature gate handling with downcast is a clever solution for specialization

#### Security & Safety
- Maintains Rust's safety guarantees
- No unsafe code introduced
- Proper error handling throughout

### 📝 Specific Suggestions

1. **Documentation**: While Copilot suggested adding error docs for `verify_range_proof`, the author indicates this will be done in a future PR with the implementation - this is acceptable.

2. **Consider renaming**: The `fork()` method name might be slightly ambiguous. Consider `create_mutable_proposal()` or similar for clarity.

3. **Test organization**: The new test `test_as_committed_with_generic_storage` is comprehensive but quite long. Consider breaking it into smaller, focused tests.

### ⚠️ Minor Issues

1. **Clippy suppressions**: The PR adds several `#[expect(clippy::...)]` annotations. While justified, ensure these are tracked for future cleanup.

2. **Feature gate complexity**: The `downcast_to_file_backed` hack for io-uring is understandable given Rust limitations, but should be documented as technical debt for when specialization stabilizes.

--- 

added since Ron likes the summaries

firewood/src/merkle.rs
```diff
@@ -653,6 +661,7 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
     /// Removes the value associated with the given `key` from the subtrie rooted at `node`.
     /// Returns the new root of the subtrie and the value that was removed, if any.
     /// Each element of `key` is 1 nibble.
+    #[expect(clippy::too_many_lines)]
```
rkuris, line 664: I wonder why this was needed? It doesn't seem like you added any lines.

storage/src/nodestore/persist.rs
```diff
@@ -245,7 +276,7 @@ impl NodeStore<Committed, FileBacked> {
     }
 }
 
-impl NodeStore<Committed, FileBacked> {
+impl<S: WritableStorage + 'static> NodeStore<Committed, S> {
```
rkuris, line 279: Question only: Why was 'static required here? I don't think it matters because S will never be a reference, but it seems like it shouldn't be required.

storage/src/nodestore/persist.rs
```diff
@@ -640,4 +685,94 @@ mod tests {
         assert!(leaf3_pos < inner_branch_pos);
         assert!(inner_branch_pos < root_pos);
     }
+
+    #[test]
+    fn test_as_committed_with_generic_storage() {
```
rkuris: nit: `test_into_committed_with_generic_storage`?

firewood/src/merkle.rs
demosdemon: I did this when I removed the top-level suppressions. I've made it a point to anytime I touch a file to remove and fix them if it didn't cause too much clutter in the diff.

storage/src/nodestore/persist.rs
demosdemon: This is a requirement in order to use the [`Any`](https://doc.rust-lang.org/stable/std/any/trait.Any.html) trait which requires that the type be `'static` for it to auto-implement `Any`.

Note that this doesn't mean `'static` lifetime. This is a `'static` type. They have similar purposes and a lot of similar constraints but are different in important ways that specifically apply here because I'm using the type system to downcast a dyn pointer to a fixed pointer.

Rust only generates TypeIds for static types and static types are types that outlive the execution of the program. This excludes things like anonymous closures and anonymous futures or generic types the include non-static types. Another way of interpreting `'static` types is that they are self-sufficient and do not depend on the existence of another non-'static type.

The `unsafe` way also requires this because we cannot access the `TypeId::of::<T>` unless `T: 'static`. The unsafe way would require us to still check `TypeId` before casting:

```rust
unsafe {
    if TypeId::of::<A>() == TypeId::of::<B>>() { 
        return Some(std::mem::transmute::<&mut A, &mut B>(self)); 
    }
}
todo!()
```

coincidentally, this is similar to how `dyn Any` implements it

```rust
    pub fn downcast_mut<T: Any>(&mut self) -> Option<&mut T> {
        if self.is::<T>() {
            // SAFETY: just checked whether we are pointing to the correct type, and we can rely on
            // that check for memory safety because we have implemented Any for all types; no other
            // impls can exist as they would conflict with our impl.
            unsafe { Some(self.downcast_mut_unchecked()) }
        } else {
            None
        }
    }

    pub unsafe fn downcast_mut_unchecked<T: Any>(&mut self) -> &mut T {
        debug_assert!(self.is::<T>());
        // SAFETY: caller guarantees that T is the correct type
        unsafe { &mut *(self as *mut dyn Any as *mut T) }
    }

    pub fn is<T: Any>(&self) -> bool {
        // Get `TypeId` of the type this function is instantiated with.
        let t = TypeId::of::<T>();

        // Get `TypeId` of the type in the trait object (`self`).
        let concrete = self.type_id();

        // Compare both `TypeId`s on equality.
        t == concrete
    }
```

PR #1136, opened 2025-07-24
feat: Update RangeProof structure

demosdemon:
While implementing the range proof verification, I realized that having an Option over a byte range is not ideal to work with. This meant that I needed to do `proof.is_none_or(|p| p.is_empty())` every time I wanted to use say a proof was empty. It was easier to use `proof.is_empty()` everywhere. The inverse was also true where when I wanted to use a proof, I first needed to destructure it and then check if it was empty. 

```rust
let proof = match proof { Some(proof) if !proof.is_empty() => Some(proof), None };
```

Was in many places in my code.

Also updated is the method signature and documentation for `verify_range_proof` to match how it will be used in the future.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -330,6 +330,8 @@ impl<T: TrieReader> Merkle<T> {
     ///   - If `Some(n)`, at most n key-value pairs will be included
     ///   - If `None`, all key-value pairs in the range will be included
     ///   - Useful for paginating through large ranges
+    ///   - **NOTE**: avalanchego's limit is based on the entire packet size and not the
+    ///     number of key-value pairs. Currently, we only limit by the number of pairs.
     ///
     /// # Returns
     ///
--- a/firewood/src/range_proof.rs
+++ b/firewood/src/range_proof.rs
@@ -1,45 +1,6 @@
 // Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-//! Range proof implementation for Merkle tries.
-//!
-//! This module provides the [`RangeProof`] type, which enables efficient cryptographic
-//! proofs for contiguous ranges of key-value pairs within a Merkle trie. Range proofs
-//! are essential for blockchain state synchronization, allowing nodes to verify that
-//! they have received a complete and authentic subset of the trie without downloading
-//! the entire data structure.
-//!
-//! # Overview
-//!
-//! A range proof consists of three components:
-//! 1. **Start proof**: Proves the lower boundary of the range
-//! 2. **End proof**: Proves the upper boundary of the range
-//! 3. **Key-value pairs**: The actual data within the proven range
-//!
-//! Together, these components allow a verifier to confirm that:
-//! - The provided key-value pairs are authentic members of the trie
-//! - No keys have been omitted from within the specified range
-//! - The trie has the expected root hash
-//!
-//! # Use Cases
-//!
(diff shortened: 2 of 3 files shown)


firewood/src/merkle.rs
```diff
@@ -206,16 +205,87 @@ impl<T: TrieReader> Merkle<T> {
...
+    /// 3. **Hash verification**: Compute the root hash of the constructed proposal
+    ///    - The computed hash must match the provided `root_hash` exactly
+    ///    - Any mismatch indicates an invalid or tampered proof
+    ///
+    /// # Errors
+    ///
+    /// * [`api::Error::ProofError`] - The proof structure is malformed or inconsistent
+    /// * [`api::Error::InvalidRange`] - The proof boundaries don't match the requested range
+    /// * [`api::Error::IncorrectRootHash`] - The computed root hash doesn't match the expected hash
+    /// * [`api::Error`] - Other errors during proposal construction or verification
```
rkuris, line 260: nit: Not all `api::Error` types can be returned here I don't think. We currently don't do any I/O when validating a range proof, but in theory that could change

firewood/src/merkle.rs
```diff
@@ -240,6 +310,72 @@ impl<T: TrieReader> Merkle<T> {
...
+    ///
+    /// * `start_key` - The optional lower bound of the range (inclusive).
+    ///   - If `Some(key)`, the proof will include all keys >= this key
+    ///   - If `None`, the proof starts from the beginning of the trie
+    ///
+    /// * `end_key` - The optional upper bound of the range (inclusive).
+    ///   - If `Some(key)`, the proof will include all keys <= this key
+    ///   - If `None`, the proof extends to the end of the trie
+    ///
+    /// * `limit` - Optional maximum number of key-value pairs to include in the proof.
```
rkuris, line 329: FYI: The go code supported a maximum network packet size. It was very pedantic on making sure it was the largest proof that would fit into a network packet. IMO we don't need to support this, but we might want to call it out.

firewood/src/range_proof.rs
```diff
@@ -1,16 +1,141 @@
...
+//!     key_values,   // Vec of (key, value) pairs in the range
+//! );
+//! ```
 
-/// A range proof proves that a given set of key-value pairs
-/// are in the trie with a given root hash.
+use crate::proof::{Proof, ProofCollection};
+
+/// A range proof is a cryptographic proof that demonstrates a contiguous set of key-value pairs
+/// exists within a Merkle trie with a given root hash.
```
rkuris, line 7: This seems really repetitive. A "range proof" is defined a bit differently in the top level README. A link there is probably a good idea here, and we should choose the best definition for the README.md.

firewood/src/range_proof.rs
```diff
@@ -1,16 +1,141 @@
...
+    ///   These are the consecutive entries from the trie that fall within the boundaries
+    ///   established by the start and end proofs. The keys should be in lexicographic
+    ///   order as they appear in the trie. May be empty if proving the absence of keys
+    ///   in a range.
+    ///
+    /// # Example Use Cases
+    ///
+    /// - Proving keys "alice" through "charlie" exist with their values
+    /// - Proving no keys exist between "aardvark" and "apple"
+    /// - Proving all keys starting with a specific prefix
```
rkuris: Feels misplaced. Why is this important for the new method?

Maybe some examples if you want to keep this.

rkuris (approved):
Approved with a few minor suggestions

PR #1137, opened 2025-07-24
chore: reorganize hash_helper for ethhash

qusuyan:


qusuyan:
This PR depends on #1130

qusuyan:
fuzz checker failed with seed 6461963400337942007

storage/src/nodestore/hash.rs
```diff
@@ -170,49 +155,36 @@ where
...
-        have_peers: bool,
+        path_prefix: &mut Path,
+        account_child_no_peers: bool,
     ) -> HashType {
-        if path_prefix.0.len() == 65 && !have_peers {
+        if path_prefix.0.len() == 65 && account_child_no_peers {
             // This is the special case when this node is the only child of an account
             //  - 64 nibbles for account + 1 nibble for its position in account branch node
             let mut fake_root = node.clone();
+            let extra_nibble = path_prefix.0.pop().expect("path_prefix not empty");
```
qusuyan, line 234: It seems that we do not need to explicitly remove the extra nibble and `hash_node` will only take the first 32 nibbles. @rkuris can you confirm?

demosdemon:
We appear to have an opportunity to use a drop guard to ensure the path is properly reset to the correct length preventing any bugs from accidentally forgetting to truncate:

```rust
struct PathGuard<'a> {
    path: &'a mut Path,
    original_len: usize,
}
impl<'a> PathGuard<'a> {
    pub fn new(path: &'a mut Path) -> Self {
        Self {
            original_len: path.len(),
            path,
        }
    }

    // note: cannot use `Self` as it's bound to `'a` and we need a new PathGuard bound to `'_`.
    pub fn fork(&mut self) -> PathGuard<'_> {
        PathGuard::new(self.path)
    }
}
impl Drop for PathGuard<'_> {
    fn drop(&mut self) {
        self.path.0.truncate(self.original_len);
    }
}
impl std::ops::Deref for PathGuard<'_> {
    type Target = Path;
        self.path
    }
}
impl std::ops::DerefMut for PathGuard<'_> {
    fn deref_mut(&self) -> &mut Self::Target {
        self.path
    }
}
```

complete example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6f106d75a0119d7783469810682ca91c

storage/src/nodestore/hash.rs
```diff
@@ -148,71 +176,52 @@ where
...
-                #[cfg(not(feature = "ethhash"))]
-                path_prefix.0.push(nibble as u8);
-
-                #[cfg(feature = "ethhash")]
-                let (child_node, child_hash) =
-                    self.hash_helper(child_node, path_prefix, make_fake_root)?;
-                #[cfg(not(feature = "ethhash"))]
-                let (child_node, child_hash) = Self::hash_helper(child_node, path_prefix)?;
+                let (child_node, child_hash) = {
+                    let mut path_guard = PathGuard::new(path_prefix);
```
demosdemon: don't need to change unless you're coming back for something else: good rule of thumb when using something like this is to shadow the original variable (i.e., also name this `path_prefix`) so it's impossible to use the wrong one. Although, rust should also make that impossible due to borrowing rules.

storage/src/nodestore/hash.rs
```diff
@@ -148,71 +176,52 @@ where
...
     }
 
     #[cfg(feature = "ethhash")]
+    /// This function computes the ethhash of a single node assuming all its children are hashed.
+    /// The function appends to `path_prefix` and then truncate it back to the original length - we only reuse the memory space to avoid allocations
     pub(crate) fn compute_node_ethhash(
         node: &Node,
-        path_prefix: &Path,
-        have_peers: bool,
+        path_prefix: &mut Path,
```
demosdemon, line 227: another minor recommendation is to type this function signature to also require an owned path guard. This way you require through the type system a forked guard and you guarantee it will be dropped before return. Additionally, owned `PathGuard<'_>` could be threaded through all of the hash calculation logic so that forking and dropping is prevalent though the entire process in a typed way.

```
mut path_prefix: PathGuard<'_>,
```

the `fork` method on the demo is useful to facilitate this

storage/src/checker/mod.rs
```diff
@@ -34,7 +34,7 @@ struct SubTrieMetadata {
     parent: TrieNodeParent,
     path_prefix: Path,
     #[cfg(feature = "ethhash")]
-    has_peers: bool,
+    has_no_peers: bool,
```
demosdemon: nit: naming is so hard, `is_peerless` is a bit better but I don't have a strong opinion. `has_no_x` is a bit clunky.

Alternatively, passing in the peer count directly would remove the `bool` which is typically a confusing type to begin with.

A third alternative would be to give this an enum so what `true`/`false` means isn't even a question.

storage/src/nodestore/hash.rs
qusuyan: I renamed it to `child_path_prefix` to match up with `child_node` so it should make things easier to read :)

storage/src/nodestore/hash.rs
```diff
@@ -33,35 +78,32 @@ where
...
-                            acc.hashed.push((idx, (maybe_persisted_node, h)));
-                        } else {
-                            // If not persisted, we need to get the node to hash it
-                            let node = maybe_persisted
-                                .as_shared_node(&self)
-                                .expect("will never fail for unpersisted nodes");
-                            acc.unhashed.push((idx, node.deref().clone()));
+                        match maybe_persisted.as_linear_address() {
+                            Some(addr) => acc.hashed.push((idx, (addr, h))),
+                            None => acc.num_unhashed += 1,
```
rkuris: This code seems broken to me. If it has an address, that does not mean it has or doesn't have a hash. All variants of MaybePersisted are already hashed.

A `Child::MaybeePersisted` has both a `MaybePersistedNode` and a `HashType`, so it is already hashed.

rkuris (changes requested):

storage/src/nodestore/hash.rs
qusuyan: Does that mean the current `hash_helper` in main is broken?

storage/src/node/branch.rs
```diff
@@ -91,6 +94,23 @@ pub(crate) trait ReadSerializable: Read {
...
+
+#[cfg(feature = "ethhash")]
+impl std::fmt::Debug for NodeRefWithHashMut<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "NodeRefWithHashMut {{ node_ref: {:?}, hash: {:?} }}",
+            **self.node_ref, self.hash
+        )
+    }
```
rkuris: You can avoid this by creating a new trait:
```
#[cfg(feature = "ethhash")]
trait NodeRef: std::ops::Deref<Target = Node> + std::fmt::Debug {}
```

Then, change `node_ref` to `Box<dyn NodeRef + 'a>`

storage/src/checker/mod.rs
```diff
@@ -534,7 +543,7 @@ mod test {
 
     /// Generate a test trie with the following structure:
     ///
-    #[cfg_attr(doc, aquamarine::aquamarine)]
+    #[cfg_attr(doc, aquamarine)]
```
rkuris: Why did this change? Seems to cause a checker dependency warning, and I'm guessing the diagram is not rendered correctly.

storage/src/nodestore/hash.rs
qusuyan: Nope - `SmallVec::with_capacity(32)` allocates space but it can continue to grow (just like `Vec`)

storage/src/node/branch.rs
qusuyan: It doesn't seem to work cuz `&Node` is not `std::fmt::Debug` :(

rkuris: I don't understand what you mean by this. I made the change and pushed it to https://github.com/ava-labs/firewood/tree/rkuris/qusuyan/hash-helper-reorg and it compiles fine. I didn't run all the tests though.

You can see the exact diff I applied here: https://github.com/ava-labs/firewood/commit/ab31227c1e22931b76a24e2eeb98b4c961eef2fa

rkuris (changes requested):
I don't like this change as it adds another allocation in the critical path.

storage/src/nodestore/hash.rs
```diff
@@ -33,99 +79,95 @@ where
...
+                        Self::compute_node_ethhash(child_node, &mut path_guard, num_children)
+                    };
+                    **child_hash = hash;
                 }
+
+                num_children
             } else {
-                // not a single child
-                None
+                // not an account branch - does not matter what we return here
```
rkuris, line 169: 🤔

I bet we can structure this code so this statement is not reachable, or let's wrap this in an Option for readability.

storage/src/nodestore/hash.rs
```diff
@@ -33,99 +79,95 @@ where
...
-                    Some(unhashed.last().expect("only one").0 as u8)
-                } else {
-                    None
+                } = self.ethhash_classify_children(&mut b.children)?;
+                trace!("hashed {hashed:?} unhashed {num_unhashed:?}");
+                #[expect(
+                    clippy::arithmetic_side_effects,
+                    reason = "hashed and unhashed can have at most 16 elements"
+                )]
+                let num_children = hashed.len() + num_unhashed;
```
rkuris, line 155: nit: just use wrapping_add, which is free. The comment can stay as to why if you want.
```suggestion
                // this sum can never exceed 16
                let num_children = hashed.len().wrapping_add(num_unhashed);
```

rkuris (commented):
This code needs to run the eth compatible hash tests, or bootstrap from genesis and verify hashes. I'm not 100% confident all this restructuring is correct.

Would have preferred this split into PathGuard changes first, which I really like and is much more straightforward, and the rest of the code, which could be breaking compatibility with ethereum's hash.

qusuyan:
This PR will be split into smaller PRs. First one here: #1202

demosdemon:
closing because this is stale, has conflicts, and likely needs to be redone considering those conflicts

Issue #1146, opened 2025-07-26
FreeArea::from_reader() backward compatibility


qusuyan:
`FreeArea::from_reader()` is not backward compatible. The case with two 0x01 bytes is broken on some images, likely because we changed how to encode `LinearAddress`. 

The on disk data (from hexdump)
```
                  [] - second 0x01 indicating this is a Some(_)
                     [ ---------- ] - next FreeArea
6e276da280 0103 fd01 a2e0 276d 006e 0000 090d 0003
             [] - AreaIndex
           [] - first 0x01 to indicate an old version
```
is supposed to be a stored area with `AreaIndex` 3 and next free area address `0x6e276da2e0`. The actual decoded result is `0xda8b07d` (229159037 in decimal). 
```rust
let raw_bytes = [0xfdu8, 0xe0, 0xa2, 0x6d, 0x27, 0x6e, 0x00];
let mut reader = std::io::Cursor::new(&raw_bytes);
let addr: u64 = reader.read_varint().unwrap();
println!("{addr:#x}");  // outputs 0xda8b07d
```
When running checker on this image, it reports the following error:
```
Error: InternalError(IO(FileIoError { inner: Custom { kind: InvalidData, error: "Invalid FreeArea marker, expected 0xFF (or 0x01 for old format), found 0xa" }, filename: Some("../good_state"), offset: 229159037, context: Some("FreeArea::from_storage") }))
```

demosdemon:
from my notes

### Free Areas

Prior to my changes, a free area was encoded using serde and bincode using the
`StoredArea<Area<Node, FreeArea>>` structure (with `FreeArea` remaining):

```rust
type LinearAddress = NonZeroU64;

struct StoredArea<T> {
    area_size_index: AreaIndex,
    area: T,
}

enum Area<T, U> {
    Node(T),
    Free(U) = 255,
}


struct FreeArea {
    next_free_block: Option<LinearAddress>,
}
```

There was intention for `Area<T, U>` to be serialized as `0xff` when an `Area::Free`,
however, the autogenerated serde code looked something like this ([full code]).

```rust
impl<T: Serialize> Serialize for StoredArea<T> {
  fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
    let mut s = s.serialize_struct("StoredArea", 2);
    s.serialize_field("area_size_index", &self.area_size_index)?;
    s.serialize_field("area", &self.area)?;
    s.end()
  }
}

[...]

PR #1147, opened 2025-07-26
fix: Fix broken deserialization of old FreeArea format

demosdemon:
As noted in the issue, we were failing to deserialize FreeAreas when they were encoded in the old serde format. This was because I did not correctly implement deserializing the varints. Bincode's format is different from the format `integer-encoding` uses thus the problem when deserializing.

This fixes the issue by implementing the correct deserialization logic for FreeArea. Now, it can correctly read the old format.

Fixes: #1146

### DIFF
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -354,12 +354,13 @@ impl Serializable for FreeArea {
                     }
                     0x01 => {
                         // encoded `Some(_)` as 1 with the data following
-                        let addr = LinearAddress::new(reader.read_varint()?).ok_or_else(|| {
-                            Error::new(
-                                ErrorKind::InvalidData,
-                                "Option::<LinearAddress> was Some(0) which is invalid",
-                            )
-                        })?;
+                        let addr = LinearAddress::new(read_bincode_varint_u64_le(&mut reader)?)
+                            .ok_or_else(|| {
+                                Error::new(
+                                    ErrorKind::InvalidData,
+                                    "Option::<LinearAddress> was Some(0) which is invalid",
+                                )
+                            })?;
                         Ok(Self {
                             next_free_block: Some(addr),
                         })
@@ -379,7 +380,7 @@ impl Serializable for FreeArea {
             first_byte => Err(Error::new(
                 ErrorKind::InvalidData,
                 format!(
-                    "Invalid FreeArea marker, expected 0xFF (or 0x01 for old format), found {first_byte:#02x}"
+                    "Invalid FreeArea marker, expected 0xFF (or 0x01 for old format), found {first_byte:#04x}"
                 ),
             )),
         }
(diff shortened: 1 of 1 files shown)


storage/src/nodestore/alloc.rs
```diff
@@ -379,7 +380,7 @@ impl Serializable for FreeArea {
             first_byte => Err(Error::new(
                 ErrorKind::InvalidData,
                 format!(
-                    "Invalid FreeArea marker, expected 0xFF (or 0x01 for old format), found {first_byte:#02x}"
+                    "Invalid FreeArea marker, expected 0xFF (or 0x01 for old format), found {first_byte:#04x}"
```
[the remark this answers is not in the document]
demosdemon: bot is wrong. 2 bytes are used for the prefix. so 4 is `0x00`.

storage/src/nodestore/alloc.rs
```diff
@@ -768,6 +769,41 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
...
+fn read_bincode_varint_u64_le(reader: &mut impl Read) -> std::io::Result<u64> {
+    // See https://github.com/ava-labs/firewood/issues/1146 for full details.
+    // emulate this behavior: https://github.com/bincode-org/bincode/blob/c44b5e364e7084cdbabf9f94b63a3c7f32b8fb68/src/config/int.rs#L241-L258
+
+    const SINGLE_BYTE_MAX: u8 = 250;
+    const U16_BYTE: u8 = 251;
+    const U32_BYTE: u8 = 252;
+    const U64_BYTE: u8 = 253;
+
+    match reader.read_byte()? {
```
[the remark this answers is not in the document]
demosdemon: bot is wrong, we have an extension trait

storage/src/nodestore/alloc.rs
```diff
@@ -768,6 +769,41 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
...
+        byte @ 0..=SINGLE_BYTE_MAX => Ok(u64::from(byte)),
```
qusuyan, line 782: TIL: `@` is valid syntax in rust

qusuyan (approved):

storage/src/nodestore/alloc.rs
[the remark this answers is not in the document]
rkuris: Quite annoying. Bot should realize this compiles at least, and even mentions that a way to fix this is to use a trait extension.

storage/src/nodestore/alloc.rs
demosdemon: yep, in pattern matches it allows you to capture the entire pattern in a variable. Sometimes a pattern can match multiple values, such as here with the range.

You could also do something like `var @ (Some(_) | None) => todo!(),` in a weird case I have seen sometimes where you might have been previously matching on a specific some value. This is commonly seen inside custom `PartialOrd::partial_cmp` implementations (but not `Ord::cmp`).

E.g.,

```rust
match PartialCmp::partial_cmp(&left, &right) {
    Some(Ordering::Equal) => todo!(),
    var @ (Some(_) | None) => return var,
}
```

This _could_ be implemented by removing the `@ ...` because you are matching on the remainder; but, perhaps in a more explicit case it is more useful to do something like below where you explicitly match on every variant so you can intentionally cause a compiler error here if a new variant were added.

```rust
match PartialCmp::partial_cmp(&left, &right) {
    Some(Ordering::Equal) => todo!(),
    var @ (Some(Ordering::Less | Ordering::Greater) | None) => return var,
}
```

demosdemon: this also works anywhere there is a pattern.

```rust
match PartialCmp::partial_cmp(&left, &right) {
    Some(cmp @ (Ordering::Less | Ordering::Greater)) => todo!(),
    var @ (Some(Ordering::Equal) | None) => return var,
}
```

qusuyan: Never mind I see the difference

PR #1148, opened 2025-07-28
Adding initial support for Node<T>

bernard-avalabs:
As part of the goal of supporting concurrent inserts into our Merkle trie, we will need the root Node to have separate locks for each entry in its children array. Eventually, this should hopefully allow different threads have exclusive access to different branches in the trie. 

The first step towards this effort is to add generics to Node as we only want the root node to need these locks. This PR adds generics support for most of the functions that use Node. There are some places where a generic version of Node is not used. For example, we do not return a Node<T> in insert_helper, remove_helper, and remove_prefix_helper. This is because it is not clear whether they need to return Node<T> in the common case if the goal is to only have the root use a specialized Node type. Similarly, as_bytes and from_reader also do not use/create Node<T> for a similar reason. They will be changed in a future PR if it turns it that using/returning is necessary.

There are also some parts of the code where additional clones are now performed. There may be ways to refactor the code such that these clones are not necessary. I would appreciate any comments there.

Finally, there are a number of Rust-related items that I'm uncertain about. For example, I'm not sure why adding the extra where clauses in insert_helper/remove_helper/remove_prefix_helper is necessary for the compiler to not complain. I would appreciate any feedback on ways to Rustify the code more.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -495,7 +495,6 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
     /// Map `key` to `value` into the subtrie rooted at `node`.
     /// Each element of `key` is 1 nibble.
     /// Returns the new root of the subtrie.
-    /// Note that the returned root is always a Node<Option<Child>>.
     pub fn insert_helper<T: ChildOption + Default>(
         &mut self,
         mut node: Node<T>,
--- a/firewood/src/stream.rs
+++ b/firewood/src/stream.rs
@@ -116,7 +116,6 @@ impl<'a, T: TrieReader> MerkleNodeStream<'a, T> {
                     match iter_node {
                         IterationNode::Unvisited { key, node } => {
                             match &*node {
-                                //Node::Root(_) => {todo!();}
                                 Node::Leaf(_) => {}
                                 Node::Branch(branch) => {
                                     // `node` is a branch node. Visit its children next.
@@ -252,7 +251,6 @@ fn get_iterator_intial_state<T: TrieReader>(
                 return Ok(NodeStreamState::Iterating { iter_stack });
             }
             Ordering::Equal => match &*node {
-                //Node::Root(_) => { todo!() }
                 Node::Leaf(_) => {
                     iter_stack.push(IterationNode::Unvisited {
                         key: matched_key_nibbles.clone().into_boxed_slice(),
--- a/storage/src/hashednode.rs
+++ b/storage/src/hashednode.rs
@@ -36,11 +36,6 @@ pub fn hash_node(node: &Node<Option<Child>>, path_prefix: &Path) -> HashType {
             }
             .into()
         }
-        /*
-        Node::Root(_) => {
-            todo!();
-        }
-        */
         Node::Leaf(node) => NodeAndPrefix {
             node,
             prefix: path_prefix,
--- a/storage/src/node/mod.rs
+++ b/storage/src/node/mod.rs
@@ -22,7 +22,6 @@
     reason = "Found 1 occurrences after enabling the lint."
 )]
 
-#[cfg(not(feature = "branch_factor_256"))]
 use crate::node::branch::ReadSerializable;
 use crate::{HashType, LinearAddress, Path, SharedNode};
 use bitfield::bitfield;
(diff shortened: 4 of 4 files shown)


storage/src/node/branch.rs
```diff
@@ -326,15 +326,30 @@ pub struct BranchNode {
...
     /// `child_address` is None if we don't know the child's hash.
-    pub children: [Option<Child>; Self::MAX_CHILDREN],
+    pub children: [T; BranchConstants::MAX_CHILDREN],
 }
 
-impl Debug for BranchNode {
+/// Contants used in Branch. Moved out into its own struct as they are
+/// independent of the generic parameter for Branch.
+#[derive(Debug)]
+pub struct BranchConstants;
```
demosdemon, line 335: I'm generally not a fan of the "constants" type pattern. I prefer to keep the constants close to their owner or where they are used. In this case, `BranchNode` is the appropriate owner because it has the set number of children.

However, in https://github.com/ava-labs/firewood/commit/984123d83b2fc675538e37551edac5a54b2f84f5 (which has caused some gnarly conflicts with your changes here), I created a type alias for the `Children<T>` array. If instead we had a `Children<T>` [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html), we could move the associated constants to that structure.

storage/src/node/branch.rs
```diff
@@ -326,15 +326,30 @@ pub struct BranchNode {
...
+    pub const MAX_CHILDREN: usize = 16;
+}
+
+impl<T: ChildOption> Debug for BranchNode<T> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         write!(f, "[BranchNode")?;
         write!(f, r#" path="{:?}""#, self.partial_path)?;
 
         for (i, c) in self.children.iter().enumerate() {
+            let c: &Option<Child> = c.as_child_option();
```
rkuris, line 352: Seems like the type should be inferred here.
```suggestion
            let c = c.as_child_option();
```

storage/src/node/branch.rs
rkuris: I agree with @demosdemon here. I think the next step is to merge in his changes which add some generics to BranchNode already.

firewood/src/proof.rs
```diff
@@ -86,7 +85,7 @@ pub struct ProofNode {
     /// Otherwise, the node's value or the hash of its value.
     pub value_digest: Option<ValueDigest<Box<[u8]>>>,
     /// The hash of each child, or None if the child does not exist.
-    pub child_hashes: [Option<HashType>; BranchNode::MAX_CHILDREN],
+    pub child_hashes: [Option<HashType>; BranchConstants::MAX_CHILDREN],
```
rkuris, line 88: Pro tip: make a separate PR that pulls out these BranchConstants.

I don't think they should be separated from BranchNode without good cause though.

rkuris (changes requested):
It's not clear to me how you're going to return a reference to a locked `Option<Child>` without running into lifetime issues with the trait the way you've set it up. Have you tried to implement this trait with an underlying mutex?

More specifically, if you have a child protected by a mutex, you can't return a reference to the inner part of that unless you're holding the mutex lock, and the caller doesn't have anything but a reference to the `Option<Child>`.

rkuris:
Additional cleanups:

 - This code doesn't compile with ethhash enabled.
 - There are many clippy warnings. You can tell your IDE to run clippy instead of just `cargo check` to avoid checking in code with warnings.

firewood/src/merkle.rs
```diff
@@ -554,7 +560,7 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
                 match node {
                     Node::Branch(ref mut branch) => {
                         #[expect(clippy::indexing_slicing)]
-                        let child = match std::mem::take(&mut branch.children[child_index as usize])
+                        let child: Node<Option<Child>> = match std::mem::take(&mut branch.children[child_index as usize]).as_child_option()
```
rkuris, line 563: You probably want a `take` method on T:

```suggestion
                        let child: Node<Option<Child>> = match &mut branch.children[child_index as usize].take()
```

The implementation is something like:

```
fn take(&smut self) -> Option<Child> {
    std::mem::take(self)
}
```

This is probably doable with a lock too, but I think you'll have trouble because you probably need to keep holding the lock after the take.

rkuris (changes requested):

PR #1149, opened 2025-07-28
feat(checker): collect basic statistics while checking the db image

qusuyan:


storage/src/nodestore/mod.rs
```diff
@@ -684,6 +704,19 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
 
         Ok((index, size))
     }
+
+    pub(crate) fn physical_size(&self) -> Result<u64, FileIoError> {
+        self.storage.size()
+    }
+
+    pub(crate) fn file_io_error(
```
demosdemon, line 774: nit: annotate this with `#[cold]` or remove the method altogether since its usage is inconsistent

storage/src/nodestore/mod.rs
```diff
@@ -651,9 +638,42 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
...
+            .into();
+        let length =
+            area_stream
+                .offset()
+                .checked_sub(offset_before)
+                .ok_or(self.storage.file_io_error(
+                    Error::other("Reader offset went backwards"),
+                    actual_addr,
+                    Some("read_node_with_num_bytes_from_disk".to_string()),
+                ))?;
```
demosdemon: `ok_or` is not lazy. This will construct an error before actually doing the subtraction which is wasteful as this actually cannot happen unless the underlying reader is broken (`Read` provides no mechanism for seeking backwards).

```suggestion
                .ok_or_else(|| self.file_io_error(
                    Error::other("Reader offset went backwards"),
                    actual_addr,
                    Some("read_node_with_num_bytes_from_disk".to_string()),
                ))?;
```

storage/src/checker/mod.rs
```diff
@@ -98,7 +143,12 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         let _leaked_areas = self.split_all_leaked_ranges(leaked_ranges, opt.progress_bar.as_ref());
         // TODO: add leaked areas to the free list
 
-        Ok(())
+        Ok(CheckerReport {
+            high_watermark: db_size,
+            physical_bytes: self.physical_size()?,
```
demosdemon: can we check `physical_bytes` early and handle the potential error early. Worst case scenario is that there's a random I/O error when we try to stat the db that causes everything we've collected to be thrown away.

storage/src/nodestore/mod.rs
```diff
@@ -651,9 +638,42 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
...
 
+    pub(crate) fn read_node_with_num_bytes_from_disk(
+        &self,
+        addr: LinearAddress,
+    ) -> Result<(SharedNode, u64), FileIoError> {
+        debug_assert!(addr.is_aligned());
+
+        // saturating because there is no way we can be reading at u64::MAX
+        // and this will fail very soon afterwards
+        let actual_addr = addr.get().saturating_add(1); // skip the length byte
```
demosdemon, line 713: theoretically we should read the area index byte and then use [`Read::take(area_size)`](https://doc.rust-lang.org/stable/std/io/trait.Read.html#method.take) on the reader to ensure we never read past the end of the area.

~We should also read to to ensure it matches the expected area index from the loop. If it differs, then this area is invalid.~ (`area_index_and_size` already did this, but the `take` suggestion remains)

demosdemon (changes requested):

storage/src/nodestore/mod.rs
demosdemon: but, noticing `area_index_and_size` acquires a stream to read the area index byte; then perhaps we could acquire the stream only once and skip this addr addition altogether.

storage/src/nodestore/mod.rs
qusuyan: I cannot access `nodestore.storage` from `checker` for scoping reasons, so kind of need this function for error handling :(.

storage/src/checker/mod.rs
qusuyan: @rkuris do you know if certain filesystems will always fail to read the file size? Should we have a fallback value (e.g., 0) on error?

storage/src/nodestore/mod.rs
qusuyan: I did not change the code logic here and changing to acquiring the stream only once seems like a big change. We should address it in another PR if needed. 

Although looking at the code again, we do not check for area size when reading data from disk most of the time. The area size is mainly for freelist management rather than reading data, so the proposed change probably does not make too much sense. Adding area size checks will again be a separate big PR.

fwdctl/src/check.rs
```diff
@@ -55,10 +55,38 @@ pub(super) async fn run(opts: &Options) -> Result<(), api::Error> {
...
-        .map_err(|e| api::Error::InternalError(Box::new(e)))
+        .map_err(|e| api::Error::InternalError(Box::new(e)))?;
+
+    print_checker_report(report);
+
+    Ok(())
+}
+
+#[expect(clippy::cast_precision_loss)]
+fn print_checker_report(report: CheckerReport) {
```
bernard-avalabs, line 71: Some additional data that would be helpful:

- Alignment-related stats. For example, the number of areas that span more than one 4K block.
- Distribution of area sizes, including the areas on the free lists.
- Distribution of distances between branch nodes and their children. This may help us determine if there are opportunities for prefetching
- This may not be possible since we don't have older revision roots, but it would be interesting to know size distribution of revisions (where size is a measure of the amount of new/updated data for a revision)

qusuyan: Talked with Bernard today: 

* I added this one to this PR since it is easy to collect. 
* We have already collected the distribution of area sizes. 
* Distribution of distance between parent and child nodes can be broadly considered as spatial locality. This requires a bit more thinking. I have added it to the checker doc and will collect this stat in a future PR. 
* This one is impossible since we only keep track of one revision on disk currently. In addition, since we only keep a fixed number of most recent revisions on disk, the difference in trie size might be not very noticeable. It should be more reasonable to collect the change set size online.

demosdemon (approved):

Issue #1138, opened 2025-07-24
Race condition when committing and reading in FFI


alarso16:
Go code which causes race:
```go
root, _ := proposal.Root()
go func () {
     proposal.Commit()
}()

// Do some other stuff

val, err := db.GetFromRoot(root, key)
```

This was discovered in a [flaky test](https://github.com/ava-labs/coreth/actions/runs/16507375551/job/46681367179?pr=1065) in coreth. The observed behavior is that any of the keys in memory (but I think it's ok on disk) will not be found if happening during commit. This is of course a very likely candidate for a race in the first place, so to fix this, we should do some of these things:

- [x] Add some sort of fuzzed concurrency testing in the FFI with `-race` enabled
- [ ] Add some sort of fuzzed concurrency testing in Rust with whatever race detector is available
- [x] Better diagnose where in the Commit process that the memory nodestore is unavailable.
- [x] Fix this issue (likely somewhere around the atomic swap of node stores?

rkuris:
This code seems to fail, using v0.0.9
```go
package main

import (
	"strconv"
	"fmt"
	"reflect"
	"sync"
	fw "github.com/ava-labs/firewood-go-ethhash/ffi"
)

func main() {
	config := fw.DefaultConfig();
	db, err := fw.New("test.db", config)
	if err != nil {
		panic(err)
	}
	var mu sync.Mutex

	for i := range 100 {
		key :=   [][]byte{[]byte("key" + strconv.Itoa(i))}
		value := [][]byte{[]byte("value" + strconv.Itoa(i))}
		prop, err := db.Propose(key, value)
		if err != nil {
			panic(err)
		}

		got, err :=  prop.Get(key[0])
		if err != nil {
			panic(err)
		}
		if !reflect.DeepEqual(got, value[0]) {
			fmt.Println("%v %v", got, value[0])
			panic(value)
		}

		go func() {
			// avoid a panic by preventing multiple commits
			mu.Lock()
			err = prop.Commit()
			mu.Unlock()
			if err != nil {
				panic(err)
			}
		}()
		got, err =  prop.Get(key[0])
		if err != nil {
			panic(err)
		}
		if !reflect.DeepEqual(got, value[0]) {
			fmt.Println("%v %v", got, value[0])
			panic(got)
		}
		fmt.Println(i)
	}
	db.Close()
}
```

```
0
1
2
3
4
5
6
[...]

PR #1153, opened 2025-07-29
test: Add parallel test

alarso16:
This test fails and should be addressed. Do not merge until the bug is resolved.

Proves #1138


alarso16:
Added to #1180

Issue #1009, opened 2025-07-01
FFI Iterator


alarso16:
Iterating across the FFI boundary is extremely slow, so it was initially determined to not be included. However, it may be necessary for some coreth API functionality (as well as being generally helpful for testing). This could still be implemented relatively simply by amortizing the FFI calls to batch the key-value pairs to relatively few calls. This will still require some clever memory management to avoid copying too much data (idea: allocate a pool in Go?) and to enable 2D arrays in C to be translated into slices of slices in Go. Additionally, freeing the iterator may be difficult if using a literal Rust iterator. 

At minimum, this must (in ethhash):
- Provide an iterator for accounts (values associated with keys of length 32)
- Given an account hash (length 32), iterate through all storage values associated with this account (all values of length 64 with keys prefixed with the account hash)

rkuris:
@alarso16 do we still need this?

alarso16:
Yes we do! In coreth, we will be returning errors on some APIs until we add this (see ava-labs/coreth#1059). The way these APIs are implemented can be different than the way they are currently, for Firewood specifically. For example, maybe using the ChangeProof api with slightly changed semantics might work well.

JuanLeon2:
These PRs implement: #1255 #1256 #1257 #1236

PR #1154, opened 2025-07-30
feat(ffi-iterator): Initial Implementation of FFI iterator

AminR443:
This draft is a step towards implementing the FFI iterator mentioned in #1009. The implementation is kept simple as possible for now without any batching and directly fetches the next item of the iterator. 

Steps on the list:
- [x] Implement a simple iterator in Go with `Next()`, `Key()`, `Value()`, and `Err()`, with proper Rust counterparts `fwd_iter_latest` and `fwd_iter_next`. 
- [ ] Implement `IterOnRoot` for creating an iterator associated with specified root, for a revision or an existing proposal.
- [ ] Explore batching for reduced overhead
- [ ] Optimize memory management for batching (avoiding large copies)
- [ ] Proper resource deallocation
- [ ] More tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@ compose-dev.yaml
 # ignore test databases
 *_db
 
+# ignore env and local
+.env
+.local_ws
+
 #### Below sections are auto-generated ####
 
 # Created by https://www.toptal.com/developers/gitignore/api/rust,visualstudiocode,vim,macos
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -27,7 +27,7 @@ use firewood::db::{BatchOp as DbBatchOp, Db, DbConfig, DbViewSync, DbViewSyncByt
 use firewood::manager::{CacheReadStrategy, RevisionManagerConfig};
 
 use firewood::merkle::Merkle;
-use firewood::stream::{MerkleKeyValueStream, NodeStreamState};
+use firewood::stream::{InternalStreamState, MerkleKeyValueStream};
 use firewood::v2::api::HashKey;
 use metrics::counter;
 
@@ -75,7 +75,7 @@ pub struct DatabaseHandle<'p> {
 #[derive(Debug)]
 pub struct IteratorHandle {
     /// Internal state of the iterator
-    state: Mutex<Box<Option<NodeStreamState>>>,
+    state: Mutex<Box<Option<InternalStreamState>>>,
 }
 
 impl From<Db> for DatabaseHandle<'_> {
@@ -188,7 +188,7 @@ fn iter_latest(db: Option<&DatabaseHandle<'_>>, key: &Value) -> Result<IteratorH
     // Find root hash.
     // Matches `hash` function but we use the TrieHash type here
     let Some(root) = db.root_hash_sync().map_err(|e| e.to_string())? else {
-        todo!()
+        return Ok(IteratorHandle{ state: Mutex::new(Box::new(None)) });
     };
 
     // Find revision associated with root.
@@ -259,10 +259,10 @@ fn iter_next(
     let mut stream =
         MerkleKeyValueStream::from_internal_state(merkle.nodestore(), inner_state);
     let next = stream.next_sync();
+    **state = Some(stream.internal_state());
 
     if let Some(next) = next {
         let (k, v) = next.map_err(|e| e.to_string())?;
-        **state = Some(stream.internal_state());
         let key: Value = k.into();
         let value: Value = v.as_slice().into();
         Ok(KeyValue { key, value })
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -188,7 +188,9 @@ fn iter_latest(db: Option<&DatabaseHandle<'_>>, key: &Value) -> Result<IteratorH
     // Find root hash.
     // Matches `hash` function but we use the TrieHash type here
     let Some(root) = db.root_hash_sync().map_err(|e| e.to_string())? else {
-        return Ok(IteratorHandle{ state: Mutex::new(Box::new(None)) });
+        return Ok(IteratorHandle {
+            state: Mutex::new(Box::new(None)),
+        });
     };
 
     // Find revision associated with root.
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -7,6 +7,10 @@ compose-dev.yaml
 # ignore test databases
 *_db
 
+# ignore env and local
+.env
+.local_ws
+
```
qusuyan, line 13: What are these for? Please put changes to `.gitignore` in a separate PR.

ffi/src/lib.rs
```diff
@@ -145,6 +155,127 @@ fn get_latest(db: Option<&DatabaseHandle<'_>>, key: &Value) -> Result<Value, Str
...
+        .state
+        .lock()
+        .expect("iterator_handle lock is poisoned");
+    let inner_state = state.take();
+    let Some(inner_state) = inner_state else {
+        return Ok(KeyValue::default());
+    };
+    let mut stream = MerkleKeyValueStream::from_internal_state(merkle.nodestore(), inner_state);
+    let next = stream.next_sync();
+    **state = Some(stream.internal_state());
```
qusuyan, line 267: nit: seems that you can release the lock early here
```suggestion
    let next = {
      let mut state = iterator_handle
          .state
          .lock()
          .expect("iterator_handle lock is poisoned");
      let inner_state = state.take();
      let Some(inner_state) = inner_state else {
          return Ok(KeyValue::default());
      };
      let mut stream = MerkleKeyValueStream::from_internal_state(merkle.nodestore(), inner_state);
      let next = stream.next_sync();
      **state = Some(stream.internal_state());
      next
    }
```

ffi/src/lib.rs
```diff
@@ -714,6 +846,15 @@ impl From<String> for Value {
...
 }
 
+impl From<String> for KeyValue {
+    fn from(s: String) -> Self {
+        KeyValue {
+            key: Value::default(),
+            value: s.into(),
+        }
+    }
+}
```
qusuyan, line 856: Why do you ever need to convert a `String` to `KeyValue` pairs? This seem very error prone since you might accidentally convert an error string into a valid key-value pair.

firewood/src/merkle.rs
```diff
@@ -157,8 +156,8 @@ impl<T: TrieReader> Merkle<T> {
         self.nodestore.root_node()
     }
 
-    #[cfg(test)]
-    pub(crate) const fn nodestore(&self) -> &T {
+    /// retrieve the underlying nodestore
+    pub const fn nodestore(&self) -> &T {
```
qusuyan, line 160: This method should not be made public since it will expose a lot of things to the external caller. This increases attack surface and violates separation of concern. Consider creating `key_value_iter_from_internal_state` in `Merkle<T>` instead. 

Then you should also put `#[cfg(test)]` back.

firewood/src/stream.rs
```diff
@@ -74,9 +77,14 @@ pub struct MerkleNodeStream<'a, T> {
...
 
+#[derive(Debug)]
+#[doc(hidden)]
+/// Internal state of the stream, for use in FFI
+pub struct InternalStreamState(NodeStreamState);
+
 impl From<Key> for NodeStreamState {
     fn from(key: Key) -> Self {
-        Self::StartFromKey(key)
+        StartFromKey(key)
```
qusuyan, line 87: This line change leads to more confusions.

firewood/src/stream.rs
```diff
@@ -9,6 +9,8 @@
 use crate::merkle::{Key, Value};
 use crate::v2::api;
 
+use crate::stream::NodeStreamState::StartFromKey;
```
qusuyan, line 12: Why do you need to import an enum defined in this file (line 62)?

firewood/src/stream.rs
```diff
@@ -353,6 +365,58 @@ impl<'a, T: TrieReader> MerkleKeyValueStream<'a, T> {
...
     }
+
+    /// Construct a [`MerkleKeyValueStream`] from prior internal state
+    pub fn from_internal_state(merkle: &'a T, state: InternalStreamState) -> Self {
+        Self {
+            state: MerkleKeyValueStreamState::Initialized {
+                node_iter: MerkleNodeStream::from(merkle, state.0),
+            },
+            merkle,
+        }
```
qusuyan, line 376: I don't think you need `MerkleNodeStream::from`. 
```suggestion
        Self {
            state: MerkleKeyValueStreamState::Initialized {
                node_iter: MerkleNodeStream {
                    merkle,
                    state: state.0
                }
            },
            merkle,
        }
```

firewood/src/stream.rs
```diff
@@ -353,6 +365,58 @@ impl<'a, T: TrieReader> MerkleKeyValueStream<'a, T> {
...
+    pub fn internal_state(self) -> InternalStreamState {
+        let node_state = match self.state {
+            MerkleKeyValueStreamState::Initialized { node_iter } => node_iter.state,
+            MerkleKeyValueStreamState::_Uninitialized(k) => StartFromKey(k),
+        };
+        InternalStreamState(node_state)
+    }
+
+    /// gets the next value synchronously
+    // TODO: Maybe Just impl Iterator? Caller would need to explicitly use Stream::next, Iteartor::next
```
qusuyan, line 390: Change `impl Stream for MerkleKeyValueStream` to use this function instead of reimplementing the same logic again.

qusuyan (changes requested):
This is a huge PR. Consider splitting it into two (e.g., one makes all changes needed in `firewood` crate and one implementing the ffi stuff).

firewood/src/stream.rs
qusuyan: See `impl Stream for MerkleNodeStream` for an example.

ffi/src/lib.rs
AminR443: We need a way in `fwd_iter_next` which returns `KeyValue` to indicate error status. For now I considered `KeyValue` struct with an empty key to indicate an error, with the error message encoded in `value`. Any better alternative?

firewood/src/merkle.rs
AminR443: This is more complicated than it seems and had to handle it this way for now. As this is a WIP, I'm exploring better alternatives, as was needed with the next steps.

firewood/src/stream.rs
AminR443: mistake from IDE auto import, will fix.

AminR443:
@qusuyan Thanks for the feedback, I will apply the changes. However, I don't think splitting this PR into two, one targeting `firewood` and another one for `ffi` would be a good idea, as it's against natural order of changes I make. I would suggest a merge after iterator is implemented completely (step 1 and 2 on the list). The later changes are scoped to FFI and won't need changes on firewood itself. 

@rkuris what do you think?

qusuyan:
> @qusuyan Thanks for the feedback, I will apply the changes. However, I don't think splitting this PR into two, one targeting `firewood` and another one for `ffi` would be a good idea, as it's against natural order of changes I make. I would suggest a merge after iterator is implemented completely (step 1 and 2 on the list). The later changes are scoped to FFI and won't need changes on firewood itself.
> 
> @rkuris what do you think?

I am ok with however you want to split it, but currently this PR is too big. By Ron's standard, he will not review a PR with 500+ lines of code change. If you think their is a more intuitive way to split the PR, feel free to do so :).

AminR443:
Thanks, Good to know the standard. I'll work a way around it to split and keep it under 500.

PR #1155, opened 2025-07-30
feat(checker): add EthKeySize error

qusuyan:
With ethhash, all values should correspond to a key with 32 or 64 bytes (64 or 128 nibbles).

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -150,14 +150,26 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 
         // read the node and iterate over the children if branch node
         let node = self.read_node(subtrie_root_address)?;
+        let mut current_node_path = path_prefix.clone();
+        current_node_path.0.extend_from_slice(node.partial_path());
+        #[cfg(feature = "ethhash")]
+        if node.value().is_some()
+            && current_node_path.0.len() != 32
+            && current_node_path.0.len() != 64
+        {
+            return Err(CheckerError::EthKeySize {
+                key: current_node_path,
+                address: subtrie_root_address,
+                parent,
+            });
+        }
         if let Node::Branch(branch) = node.as_ref() {
             // this is an internal node, traverse the children
             #[cfg(feature = "ethhash")]
             let num_children = branch.children_iter().count();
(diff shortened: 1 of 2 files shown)


storage/src/lib.rs
```diff
@@ -261,6 +261,20 @@ pub enum CheckerError {
     #[error("The checker can only check persisted nodestores")]
     UnpersistedRoot,
 
+    #[cfg(feature = "ethhash")]
+    #[error(
+        "The node {key:?} at {address:#x} (parent: {parent:#x}) has a value but its path is not 32 or 64 bytes long"
```
rkuris: Is this key the debug representation of Path or the LowerHex representation? I think we want the latter.

storage/src/checker/mod.rs
```diff
@@ -150,14 +153,23 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 
         // read the node and iterate over the children if branch node
         let node = self.read_node(subtrie_root_address)?;
+        let mut current_node_path = path_prefix.clone();
+        current_node_path.0.extend_from_slice(node.partial_path());
+        #[cfg(feature = "ethhash")]
+        if node.value().is_some() && !VALID_ETH_KEY_SIZES.contains(&current_node_path.0.len()) {
```
rkuris: FYI, this produced surprisingly good assembly: https://godbolt.org/z/3jcacrrEn

storage/src/checker/mod.rs
```diff
@@ -150,14 +153,23 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 
         // read the node and iterate over the children if branch node
         let node = self.read_node(subtrie_root_address)?;
+        let mut current_node_path = path_prefix.clone();
+        current_node_path.0.extend_from_slice(node.partial_path());
+        #[cfg(feature = "ethhash")]
```
rkuris: While we're here, if the nibble length is odd, there can't be a value either, regardless of ethhash.

rkuris (approved):
Approved, but please address the comments.

storage/src/checker/mod.rs
qusuyan: Ahha because we are using `const blabla = [64, 128]` right?

PR #1156, opened 2025-08-01
feat: Add lock to the deleted vector in MutableProposal

bernard-avalabs:
This PR adds a lock to the deleted vector in MutableProsposal. This allows read_for_update in MutableProposal to take a &self instead of a &mut self, which in turn allows insert_helper, remove_helper, and remove_prefix_helper to take a &self instead of a &mut self.

The purpose of this change is to allow other threads (worker threads) to call insert_helper concurrently in certain cases. That change will be implemented in a later PR.

In a discussion with Ron, we considered two options for allowing concurrent access to the Merkle trie. The first is to create Merkle tries for each child node from the root, and the second was to allow concurrent updates to the MutableProposal. There are some potential performance advantages to the first option, but the second one (as implemented in this PR) will likely require fewer interface changes to implement.

### DIFF
--- a/storage/src/nodestore/mod.rs
+++ b/storage/src/nodestore/mod.rs
@@ -81,7 +81,7 @@ pub use header::NodeStoreHeader;
 /// ```
 use std::mem::take;
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use crate::hashednode::hash_node;
 use crate::node::Node;
@@ -229,7 +229,7 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
         };
         let kind = MutableProposal {
             root,
-            deleted,
+            deleted: Mutex::new(Some(deleted)),
             parent: parent.kind.as_nodestore_parent(),
         };
         Ok(NodeStore {
@@ -240,9 +240,18 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
     }
 
     /// Marks the node at `addr` as deleted in this proposal.
-    pub fn delete_node(&mut self, node: MaybePersistedNode) {
+    /// 
+    /// # Panics
+    ///
+    /// Will panic if the deleted array is None
+    pub fn delete_node(&self, node: MaybePersistedNode) {
         trace!("Pending delete at {node:?}");
-        self.kind.deleted.push(node);
+        self.kind.deleted
+            .lock()
+            .expect("failed lock acquire")
+            .as_mut()
+            .expect("none option")
+            .push(node);
     }
 
     /// Reads a node for update, marking it as deleted in this proposal.
@@ -252,7 +261,7 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
     /// # Errors
     ///
     /// Returns a [`FileIoError`] if the node cannot be read.
-    pub fn read_for_update(&mut self, node: MaybePersistedNode) -> Result<Node, FileIoError> {
+    pub fn read_for_update(&self, node: MaybePersistedNode) -> Result<Node, FileIoError> {
         let arc_wrapped_node = node.as_shared_node(self)?;
         self.delete_node(node);
         Ok((*arc_wrapped_node).clone())
@@ -281,7 +290,7 @@ impl<S: WritableStorage> NodeStore<MutableProposal, S> {
             header,
             kind: MutableProposal {
                 root: None,
-                deleted: Vec::default(),
+                deleted: Mutex::new(Some(Vec::default())),
                 parent: NodeStoreParent::Committed(None),
             },
             storage,
@@ -438,7 +447,7 @@ pub struct MutableProposal {
     /// The root of the trie in this proposal.
     root: Option<Node>,
     /// Nodes that have been deleted in this proposal.
-    deleted: Vec<MaybePersistedNode>,
+    deleted: Mutex<Option<Vec<MaybePersistedNode>>>,
     parent: NodeStoreParent,
 }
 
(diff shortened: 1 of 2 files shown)


storage/src/nodestore/mod.rs
```diff
@@ -438,7 +448,7 @@ pub struct MutableProposal {
     /// The root of the trie in this proposal.
     root: Option<Node>,
     /// Nodes that have been deleted in this proposal.
-    deleted: Vec<MaybePersistedNode>,
+    deleted: Mutex<Option<Vec<MaybePersistedNode>>>,
     parent: NodeStoreParent,
```
qusuyan, line 450: The `Option<>` here seem redundant - if no node has been deleted, we simply have an empty `Vec<>`. It also allows you to remove an extra `expect()` every time you lock it.

storage/src/nodestore/mod.rs
```diff
@@ -240,9 +240,19 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
...
+    /// Will panic if deleted is None
+    pub fn delete_node(&self, node: MaybePersistedNode) {
         trace!("Pending delete at {node:?}");
-        self.kind.deleted.push(node);
+        self.kind
+            .deleted
+            .lock()
+            .expect("lock acquire failed")
+            .as_mut()
+            .expect("none option")
```
qusuyan: This can be removed after the unnecessary `Option<>` is removed

storage/src/nodestore/mod.rs
```diff
@@ -516,7 +526,13 @@ impl<S: ReadableStorage> TryFrom<NodeStore<MutableProposal, S>>
         let mut nodestore = NodeStore {
             header,
             kind: Arc::new(ImmutableProposal {
-                deleted: kind.deleted.into(),
+                deleted: kind
+                    .deleted
+                    .lock()
+                    .expect("lock acquire failed")
+                    .take()
+                    .expect("none option")
```
qusuyan: Same with this `expect()`

qusuyan (changes requested):

storage/src/nodestore/mod.rs
bernard-avalabs: Thanks! I've removed the redundant Option<>.

storage/src/nodestore/mod.rs
bernard-avalabs: I've replaced this with an into_inner on the mutex.

storage/src/nodestore/mod.rs
bernard-avalabs: Removed.

storage/src/nodestore/mod.rs
qusuyan: Good call

qusuyan:
This looks good to me. Regarding how to handle lock errors @rkuris and @demosdemon may have some more suggestions. It would be good if they can do another pass. 

Another slight thing is that I don't know if `Mutex<Vec<>>` is the correct data structure to use here. It seems that we can benefit from a concurrent vector/set built on atomic primitives rather than locking. Also, since it looks like the critical sections are very small. If we do not expect high concurrency, spin locks may deliver better performance.

demosdemon:
> Another slight thing is that I don't know if `Mutex<Vec<>>` is the correct data structure to use here.

TBH, you will typically get a response from me along the lines of "locks are never the right answer." I love lock free designs because they tackle concurrency problems by ownership instead of by reference. And, the trie data-structure conveniently supports the "by ownership" model by allowing each node to also be a complete trie. So, multi-threading should be "divide and conquer" based.

bernard-avalabs:
> > Another slight thing is that I don't know if `Mutex<Vec<>>` is the correct data structure to use here.
> 
> TBH, you will typically get a response from me along the lines of "locks are never the right answer." I love lock free designs because they tackle concurrency problems by ownership instead of by reference. And, the trie data-structure conveniently supports the "by ownership" model by allowing each node to also be a complete trie. So, multi-threading should be "divide and conquer" based.

The approach that I'm planning to take is largely based on "divide and conquer". For insert cases where the root node will not change, the main thread will pass ownership of a child node to one of the worker threads. The worker thread will then perform the remaining insert operation along that branch. At the end of a batch (or once we receive an insert request that will change the root), the updated child nodes are collected from the workers and used to update the root node.

For simplicity, I've decided to allow the worker threads to use the same nodestore as the only mutable operation that they perform on the nodestore is to add an entry to the "deleted" vector. Given the small critical section, I don't expect this lock to have a large impact on performance. Of course I won't know for sure until I evaluate it once I've completed the prototype. My preference is to keep the changes small so I can more quickly prototype the parallel insert and test its performance impact. As part of the test/refine cycle, I'm happy to revisit other designs including ones that eliminate the shared nodestore.

With that said, I'm still new to the codebase and to Rust, and I'm open to suggestions and feedback especially if you think I've overlooked something important.

bernard-avalabs:
> Another slight thing is that I don't know if `Mutex<Vec<>>` is the correct data structure to use here. It seems that we can benefit from a concurrent vector/set built on atomic primitives rather than locking. Also, since it looks like the critical sections are very small. If we do not expect high concurrency, spin locks may deliver better performance.

This is a great suggestion. I think it is worth exploring once the prototype is complete and we can evaluate the performance of the different options.

rkuris (commented):
The right answer here is to have each thread collect the set of deleted nodes and insert them into the array with a single lock when the batch is finished rather than creating the risk of contention here.

Rephrasing, each thread should maintain it's own delete list and they should be combined at the end. Since the thread already has to tell it's parent where it's new root is, it could also pass back a vector of deleted nodes.

PR #1158, opened 2025-08-01
feat(ffi): refactor FFI layer for improved modularity and consistency

demosdemon:
- Split monolithic `lib.`rs into focused modules: `handle`, `proposal`, `arc_cache`, `logging`, and `value` submodules.
- Replace global proposal tracking with encapsulated `ProposalHandle` that retains database reference. This extra reference is only necessary to do the cached memory view while we are committing the view.
- Added `ArcCache` to hold the cached memory view. Maybe this goes away in the future.
- Restructure value types into separate modules (`hash_key`, `owned`, `results`)
- Add comprehensive error handling with `CResult` types and panic recovery
- Remove undocumented unsafe blocks clippy exception
- Update Go bindings to use new handle-based API
- Add Grafana dashboard configuration for metrics (to silence the CI bot)

This refactor improves code organization, type safety, and maintainability while preserving the existing FFI API surface.

I would like to break this down into smaller pieces, but at this point I don't know how much I can break it down without making it an interim mess.

---

1. #1174
2. #1158

### DIFF
--- a/benchmark/Grafana-dashboard.json
+++ b/benchmark/Grafana-dashboard.json
@@ -952,6 +952,375 @@
       ],
       "title": "Unwritten Nodes (In Memory)",
       "type": "timeseries"
+    },
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 39
+      },
+      "id": 14,
+      "panels": [],
+      "title": "FFI",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisBorderShow": false,
+            "axisCenteredZero": false,
+            "axisColorMode": "text",
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "barAlignment": 0,
+            "barWidthFactor": 0.6,
+            "drawStyle": "line",
+            "fillOpacity": 0,
+            "gradientMode": "none",
+            "hideFrom": {
+              "legend": false,
+              "tooltip": false,
+              "viz": false
+            },
+            "insertNulls": false,
+            "lineInterpolation": "linear",
+            "lineWidth": 1,
+            "pointSize": 5,
+            "scaleDistribution": {
+              "type": "linear"
+            },
+            "showPoints": "auto",
+            "spanNulls": false,
+            "stacking": {
+              "group": "A",
+              "mode": "none"
+            },
+            "thresholdsStyle": {
+              "mode": "off"
+            }
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 0,
+        "y": 40
+      },
+      "id": 15,
+      "options": {
+        "legend": {
+          "calcs": [
+            "min",
+            "mean",
+            "max"
+          ],
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "hideZeros": false,
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "pluginVersion": "11.5.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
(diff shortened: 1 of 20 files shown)


ffi/memory.go
```diff
@@ -102,135 +124,314 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
...
+// referenced. If the C.OwnedBytes is nil, it returns nil.
+func fromOwnedBytes(owned C.OwnedBytes) *RustOwnedBytes {
+	if owned.ptr == nil {
+		return nil
 	}
 
-	// Case 1
-	if v.len == 0 && v.data == nil {
+	rustBytes := &RustOwnedBytes{owned: owned}
+	runtime.SetFinalizer(rustBytes, func(b *RustOwnedBytes) {
```
rkuris, line 287: I am not sure I love this. Previous experience with the finalizer has shown it to be slow both when setting and resetting it. If we do this for every allocated thing, it might be a big performance loss.

Let's chat about this more in person, or we can run some golang benchmarks to see how much impact this change has.

demosdemon: ack on Finalizers. I can rework to use cleanup handlers. I'm not particularly a fan of that but we are bound to go1.23. If and when avalanchego upgrades to 1.24+, we can switch to [Cleanup](https://pkg.go.dev/runtime#AddCleanup) which is much better with performance and reliability.

But, also we might not use this as much as you think because of my other PR (#1174). The finalizers here are only used for bytes allocated by Rust returned to Go. I acknowledge that this might be a bit too much overhead for individual values (especially because I do copy and free before returning those to the caller) but I would really like to keep them for the handles.

benchmark/Grafana-dashboard.json
alarso16, line 1: I don't think this change belongs in this PR (not clear to me what it is anyway)

ffi/firewood.go
```diff
@@ -108,22 +104,16 @@ func New(filePath string, conf *Config) (*Database, error) {
...
 	defer pinner.Unpin()
 
-	args := C.struct_CreateOrOpenArgs{
+	args := C.struct_DatabaseHandleArgs{
 		path:                 newBorrowedBytes([]byte(filePath), &pinner),
 		cache_size:           C.size_t(conf.NodeCacheEntries),
 		free_list_cache_size: C.size_t(conf.FreeListCacheEntries),
 		revisions:            C.size_t(conf.Revisions),
-		strategy:             C.uint8_t(conf.ReadCacheStrategy),
+		strategy:             C.CacheReadStrategy(conf.ReadCacheStrategy),
```
alarso16, line 112: What happens here if it's not a valid `C.CacheReadStrategy`? E.g. it's 7. does this panic?

ffi/firewood.go
```diff
@@ -194,21 +179,23 @@ func (db *Database) GetFromRoot(root, key []byte) ([]byte, error) {
...
 	}
 
 	var pinner runtime.Pinner
 	defer pinner.Unpin()
 
-	val := C.fwd_get_from_root(
+	return fromValueResult(C.fwd_get_from_root(
 		db.handle,
-		newBorrowedBytes(root, &pinner),
+		rootKey.toC(),
```
alarso16, line 196: This is relatively unreadable to me (looks kinda Rust-y). `toC` has really strange capitalizaiton

ffi/memory.go
```diff
@@ -6,28 +6,50 @@
...
+	}
+
+	return (*RustOwnedBytes)(e).string()
+}
+
+type HashKey [32]byte
+
+func NewHashKey(b []byte) (HashKey, error) {
+	if len(b) != 32 {
+		return HashKey{}, fmt.Errorf("hash key must be 32 bytes, got %d", len(b))
```
alarso16, line 42: nit: move this to a testable error, like `fmt.Errorf("%w, got length %d", errBadKeyLength)`

ffi/memory.go
```diff
@@ -6,28 +6,50 @@
...
+	if e == nil {
+		return errUnknown.Error()
+	}
+
+	return (*RustOwnedBytes)(e).string()
+}
+
+type HashKey [32]byte
+
+func NewHashKey(b []byte) (HashKey, error) {
```
alarso16, line 40: Should this be exported? It should probably only be used internally, right?

alarso16 (commented):
Haven't finished yet, but a few notes

ffi/firewood.go
demosdemon: this method checks for `invalidCacheStrategy` before reaching this point, but it would do nothing (here) if we did that and cause UB on the rust side

benchmark/Grafana-dashboard.json
demosdemon: this is because I touch the counter! and the CI bot would create a comment on the PR every time I updated if I didn't touch this file.

ffi/memory.go
demosdemon: No reason it can't be used externally so callers have a typed key but I can fix that here.

demosdemon:
Please also review #1174 before this. I apologize for publishing both at the same time.

PR #1159, opened 2025-08-01
chore: require license header for ffi code

demosdemon:


ffi/cbindgen.toml
```diff
@@ -93,7 +97,7 @@ rename_variants = "None"
 # deprecated = "DEPRECATED_ENUM"
 # deprecated_with_note = "DEPRECATED_ENUM_WITH_NOTE"
 add_sentinel = false
-prefix_with_name = false
+prefix_with_name = true
```
demosdemon, line 100: this change doesn't affect anything now but does affect a future PR. I didn't mean to include it here but also don't want to spatch out this one line.

PR #1160, opened 2025-08-01
feat(ffi): Remove the Arc wrapper around Proposal

demosdemon:
Ron mentioned that having the proposals be cloneable posed a problem where we wanted to assert that a proposal was no longer in scope once committed. After some digging, I realized
that this was unnecessary. 

There was the `ProposalBase` enum that we used to build a layerable in-memory database off an existing view (like the EmptyDb). However, neither are in use and created confusion with the other Proposal object. So, they were removed as part of unwrapping the Arc around proposals.

A nit change also was to give names to all of the lifetimes in the proposal and db traits. Naming the lifetimes makes it easier to understand their purpose (though, I avoided the FFI package to save myself some rebase pains).


rkuris (approved):
⭐

PR #1161, opened 2025-08-01
feat(fwdctl): Add support for dump formats

rkuris:
Found myself needing to dump dot graphs during debugging. This was too hard, since the API for generating these wasn't exposed to fwdctl. Added it to fwdctl, as well as refactoring the dump methods to use writers instead of returning a string.

### DIFF
--- a/firewood/src/merkle.rs
+++ b/firewood/src/merkle.rs
@@ -16,7 +16,7 @@ use firewood_storage::{
 use futures::{StreamExt, TryStreamExt};
 use metrics::counter;
 use std::collections::HashSet;
-use std::fmt::{Debug, Write};
+use std::fmt::Debug;
 use std::future::ready;
 use std::io::Error;
 use std::iter::once;
@@ -31,6 +31,8 @@ pub type Value = Box<[u8]>;
 
 // convert a set of nibbles into a printable string
 // panics if there is a non-nibble byte in the set
+// TODO: This is more efficient than the LowerHex implementation for Path
+// so combine them.
 #[cfg(not(feature = "branch_factor_256"))]
 fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
     nib.into_iter()
@@ -51,27 +53,27 @@ fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
 macro_rules! write_attributes {
     ($writer:ident, $node:expr, $value:expr) => {
         if !$node.partial_path.0.is_empty() {
-            write!(
-                $writer,
-                " pp={}",
-                nibbles_formatter($node.partial_path.0.clone())
-            )
-            .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
+            let pp_str = format!(" pp={}", nibbles_formatter($node.partial_path.0.clone()));
+            $writer
+                .write_all(pp_str.as_bytes())
+                .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
         }
         if !$value.is_empty() {
             match std::str::from_utf8($value) {
                 Ok(string) if string.chars().all(char::is_alphanumeric) => {
-                    write!($writer, " val={}", string)
+                    let val_str = format!(" val={}", string);
+                    $writer
+                        .write_all(val_str.as_bytes())
                         .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
                 }
                 _ => {
                     let hex = hex::encode($value);
+                    let val_str = format!(" val={:.6}", hex);
+                    $writer
+                        .write_all(val_str.as_bytes())
+                        .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
                     if hex.len() > 6 {
-                        write!($writer, " val={:.6}...", hex).map_err(|e| {
-                            FileIoError::from_generic_no_file(e, "write attributes")
-                        })?;
-                    } else {
-                        write!($writer, " val={}", hex).map_err(|e| {
+                        $writer.write_all(b"...").map_err(|e| {
                             FileIoError::from_generic_no_file(e, "write attributes")
                         })?;
                     }
@@ -478,12 +480,12 @@ impl<T: TrieReader> Merkle<T> {
 
 impl<T: HashedNodeReader> Merkle<T> {
     /// Dump a node, recursively, to a dot file
-    pub(crate) fn dump_node(
+    pub(crate) fn dump_node<W: std::io::Write + ?Sized>(
         &self,
         node: &MaybePersistedNode,
         hash: Option<&HashType>,
         seen: &mut HashSet<String>,
-        writer: &mut dyn Write,
+        writer: &mut W,
     ) -> Result<(), FileIoError> {
         writeln!(writer, "  {node}[label=\"{node}")
             .map_err(Error::other)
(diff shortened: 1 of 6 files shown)


qusuyan:
omg copilot PR comments

firewood/src/merkle.rs
```diff
@@ -51,27 +53,27 @@ fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
...
                 Ok(string) if string.chars().all(char::is_alphanumeric) => {
-                    write!($writer, " val={}", string)
+                    let val_str = format!(" val={}", string);
+                    $writer
+                        .write_all(val_str.as_bytes())
                         .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
                 }
                 _ => {
                     let hex = hex::encode($value);
+                    let val_str = format!(" val={:.6}", hex);
```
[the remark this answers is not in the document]
rkuris: https://doc.rust-lang.org/std/fmt/#precision
 > For non-numeric types, this can be considered a “maximum width”. If the resulting string is longer than this width, then it is truncated down to this many characters and that truncated value is emitted with proper fill, alignment and width if those parameters are set.

firewood/src/merkle.rs
```diff
@@ -478,12 +480,12 @@ impl<T: TrieReader> Merkle<T> {
...
-    pub(crate) fn dump_node(
+    pub(crate) fn dump_node<W: std::io::Write + ?Sized>(
         &self,
         node: &MaybePersistedNode,
         hash: Option<&HashType>,
         seen: &mut HashSet<String>,
-        writer: &mut dyn Write,
+        writer: &mut W,
     ) -> Result<(), FileIoError> {
         writeln!(writer, "  {node}[label=\"{node}")
```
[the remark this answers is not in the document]
demosdemon: incorrect. `writeln!` works lexicographically (i.e., not with the type system) with anything with a `write_fmt` implementation in scope and does not care where it comes from.

fwdctl/src/dump.rs
```diff
@@ -306,6 +309,15 @@ fn create_output_handler(
...
         }
         "stdout" => Ok(Box::new(StdoutOutputHandler { hex })),
+        "dot" => {
+            println!("Dumping to {}", file_name.display());
+            let file = File::create(file_name)?;
+            let mut writer = BufWriter::new(file);
+            // For dot format, we generate the output immediately since it doesn't use streaming
+            db.dump_sync(&mut writer)?;
+            // Return a no-op handler since the work is already done
+            Ok(Box::new(StdoutOutputHandler { hex: false }))
```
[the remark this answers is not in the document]
demosdemon: actually helpful comment.

The comment doesn't match the behavior. Returning a stream here is equivalent to the `"stdout"` branch, which allows the normal dump process to proceed. The comment about a no-op isn't true.

fwdctl/src/dump.rs
```diff
@@ -91,16 +91,16 @@ pub struct Options {
...
     /// The output format of database dump.
-    /// Possible Values: ["csv", "json", "stdout"].
+    /// Possible Values: ["csv", "json", "stdout", "dot"].
     /// Defaults to "stdout"
     #[arg(
         short = 'o',
         long,
         required = false,
         value_name = "OUTPUT_FORMAT",
-        value_parser = ["csv", "json", "stdout"],
```
demosdemon, line 101: clap has a macro for this

```
#[derive(Debug, clap::ValueEnum)]
enum OutputFormat {
  Csv,
  Json,
  Stdout,
  Dot,
}
```

and replace this `value_parser` with `value_enum`:

```
#[arg(value_enum, ...)]
pub output_format: OutputFormat,
```

rkuris: Added as issue for later

firewood/src/merkle.rs
```diff
@@ -31,6 +31,8 @@ pub type Value = Box<[u8]>;
 
 // convert a set of nibbles into a printable string
 // panics if there is a non-nibble byte in the set
+// TODO: This is more efficient than the LowerHex implementation for Path
```
demosdemon: "more efficient" is debatable. This allocates a string whereas the lowerhex writes to whatever the formatter is wrapping, which could be a string buffer or could be directly to a writer. If this is constructed only to immediately write to another buffer, then this is not more efficient.

firewood/src/merkle.rs
```diff
@@ -51,27 +53,27 @@ fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
 macro_rules! write_attributes {
     ($writer:ident, $node:expr, $value:expr) => {
         if !$node.partial_path.0.is_empty() {
-            write!(
-                $writer,
-                " pp={}",
-                nibbles_formatter($node.partial_path.0.clone())
-            )
-            .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
+            let pp_str = format!(" pp={}", nibbles_formatter($node.partial_path.0.clone()));
```
demosdemon: example; here: this is less efficient than lowerhex because you're allocating a string only to immediately copy it into to another string and then drop it.

firewood/src/merkle.rs
```diff
@@ -51,27 +53,27 @@ fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
...
+            $writer
+                .write_all(pp_str.as_bytes())
```
demosdemon: and then the temporary `pp_str` it's copied to the writer and then immediately dropped.

we could completely obviate the need for two heap buffers by writing directly to the writer:

```
write!($writer, " pp={:x}", $node.partial_path)
.map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
```

firewood/src/merkle.rs
```diff
@@ -51,27 +53,27 @@ fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
...
+                .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
         }
         if !$value.is_empty() {
             match std::str::from_utf8($value) {
                 Ok(string) if string.chars().all(char::is_alphanumeric) => {
-                    write!($writer, " val={}", string)
+                    let val_str = format!(" val={}", string);
+                    $writer
+                        .write_all(val_str.as_bytes())
```
demosdemon: this is unnecessary. you're constructing a heap string only to write it to the buffer when before it was correctly writing both directly to the writer.

firewood/src/merkle.rs
```diff
@@ -51,27 +53,27 @@ fn nibbles_formatter<X: IntoIterator<Item = u8>>(nib: X) -> String {
...
+                        .map_err(|e| FileIoError::from_generic_no_file(e, "write attributes"))?;
                     if hex.len() > 6 {
-                        write!($writer, " val={:.6}...", hex).map_err(|e| {
-                            FileIoError::from_generic_no_file(e, "write attributes")
-                        })?;
-                    } else {
-                        write!($writer, " val={}", hex).map_err(|e| {
+                        $writer.write_all(b"...").map_err(|e| {
```
demosdemon: Same here. Why are we allocating on the heap only to copy it to a buffer and then drop it?

rkuris: Agreed, this all needs a makeover. I'll update the other task I created to include these comments as well.

firewood/src/merkle.rs
rkuris: Moved to 1163

firewood/src/merkle.rs
rkuris: Moved to 1163

firewood/src/merkle.rs
demosdemon: moving it to an issue is unnecessary because this PR is the change that's making it worse. It should be fixed here.

demosdemon (changes requested):

firewood/src/merkle.rs
rkuris: Unmoved, and fixed here.

PR #1166, opened 2025-08-01
ci: update .golangci.yaml

### LINKED ISSUE #1164: ci: golangci-lint config out of date
https://github.com/ava-labs/firewood/actions/runs/16685941334/job/47235257503

demosdemon:
I also got bored and updated the script. This allows us to get more useful output when there is a change; e.g.,:

```console
$ .github/scripts/verify_golangci_yaml_changes.sh
'ffi/.golangci.yaml' has unexpected changes from AvalancheGo.
View the upstream changes at: https://github.com/ava-labs/avalanchego/commits/master/.golangci.yml and apply them if necessary.

Current changes:
--- .github/.golangci.yaml      2025-08-02 10:04:10
+++ ffi/.golangci.yaml  2025-08-02 10:01:55
@@ -12,7 +12,7 @@
   #
   # Allowed values: readonly|vendor|mod
   # By default, it isn't set.
-  modules-download-mode: mod
+  modules-download-mode: readonly
 
   # Include non-test files tagged as test-only.
   # Context: https://github.com/ava-labs/avalanchego/pull/3173
```

Additionally, we can run `apply` to apply our patch against the upstream file and refresh the patch file:

```console
$ .github/scripts/verify_golangci_yaml_changes.sh apply
patching file '.github/.golangci.yaml'
Successfully applied the patch from .github/.golangci.yaml.patch to ffi/.golangci.yaml
Updated expected changes in .github/.golangci.yaml.patch
```

fixes #1164


ffi/.golangci.yaml
```diff
@@ -36,7 +36,6 @@ linters:
     - bodyclose
     - copyloopvar
     - depguard
-    - dupword
```
demosdemon, line 39: discussion around this change is here: https://github.com/ava-labs/avalanchego/pull/4146/files#r2248670142

.github/scripts/verify_golangci_yaml_changes.sh
```diff
@@ -1,19 +1,135 @@
...
+
+    # refresh the patch so `check` apply later can use it
+    @update
+}
+
+function @check() {
+    @apply-patch-to-upstream
+
+    local patch
+    patch=$(mktemp)
```
rkuris, line 95: Needs quotes:

```suggestion
patch="$(mktemp)"
```

.github/scripts/verify_golangci_yaml_changes.sh
```diff
@@ -1,19 +1,135 @@
...
+function @apply-patch-to-upstream() {
+    if ! patch -t "$upstream_yaml" "$expected_patch"; then
+        @die 4 "Failed to apply the patch from $expected_patch. Please review the changes manually."
+    fi
+}
+
+function @apply() {
+    local backup
+    backup=$(mktemp)
+    trap 'rm -f "$backup"' EXIT
```
rkuris, line 76: This looks incorrect. The single quotes will prevent $backup from being expanded.

rkuris (approved):

rkuris:
I found a few problems, there are probably more issues that can't be caught by review easily.

Wish we could write this in rust...

.github/scripts/verify_golangci_yaml_changes.sh
demosdemon: This is correct. The trap string is expanded `eval` style on trigger.

.github/scripts/verify_golangci_yaml_changes.sh
demosdemon: Quotes are not needed when doing a variable assignment.

demosdemon:
> I found a few problems, there are probably more issues that can't be caught by review easily.
> 
> Wish we could write this in rust...

These "issues" (that are not issues) are easily verifiable with [shellcheck](https://www.shellcheck.net/).

PR #1169, opened 2025-08-02
feat: remove Default impl on HashType

demosdemon:
We currently have two "default" values for the TrieHash. In some cases I
was using `unwrap_or_default` when I should have been using the empty
RLP hash. This commit removes the `Default` impl on `HashType` and
replaces those instances with `HashType::empty()`.

Additionally, I have pre-calculated the empty hash value so it doesn't
need to be done at runtime.

---

1. #1169
2. #1170
3. #1171
4. #1172
5. #1173
6. #1174
7. #1158


storage/Cargo.toml
```diff
@@ -52,6 +52,8 @@ sha3 = { version = "0.10.8", optional = true }
 criterion = { workspace = true, features = ["async_tokio", "html_reports"] }
 pprof = { workspace = true, features = ["flamegraph"] }
 rand.workspace = true
+rlp = "0.6.1"
+sha3 = "0.10.8"
```
rkuris: I don't see where these are getting used, and I'd much prefer not having rlp or sha3 in storage.

storage/src/node/branch.rs
```diff
@@ -206,6 +206,37 @@ mod ethhash {
...
+        ///     sha3::Keccak256::digest(rlp::NULL_RLP)
+        ///             .as_slice()
+        ///             .try_into()
+        ///             .expect("empty trie hash is 32 bytes"),
+        /// )
+        /// ```
+        #[must_use]
+        pub fn empty_rlp_hash() -> Self {
+            TrieHash::empty_rlp_hash().into()
+        }
```
rkuris: It's unfortunate that there's no way to keep RLP out of storage. I realize this is just a constant so it's not terrible but it would be nice if storage couldn't care less about what an empty hash is, and it's always `empty` if ethhash is disabled.

This leaves you with a coding error of someone calling `empty_rlp_hash` in firewood when `ethhash` is disabled.

storage/src/trie_hash.rs
```diff
@@ -7,11 +7,63 @@ use sha2::digest::generic_array::GenericArray;
 use sha2::digest::typenum;
 use std::fmt::{self, Debug, Display, Formatter};
 
+const EMPTY_HASH: [u8; TRIE_HASH_LEN] = [0; TRIE_HASH_LEN];
+
+const EMPTY_RLP_HASH: [u8; TRIE_HASH_LEN] = [
+    // "56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"
+    0x56, 0xe8, 0x1f, 0x17, 0x1b, 0xcc, 0x55, 0xa6, 0xff, 0x83, 0x45, 0xe6, 0x92, 0xc0, 0xf8, 0x6e,
+    0x5b, 0x48, 0xe0, 0x1b, 0x99, 0x6c, 0xad, 0xc0, 0x01, 0x62, 0x2f, 0xb5, 0xe3, 0x63, 0xb4, 0x21,
+];
```
rkuris: In theory, it's possible for someone to need a version of firewood that uses something other than RLP encoding an empty value as their root hash.

rkuris (changes requested):
Not opposed to removing the Default value since you're spot on about it being a footgun, but I don't love bleeding so much knowledge into storage about what represents an empty hash.

Maybe we should be using a different hash type outside of storage or something.

storage/Cargo.toml
demosdemon: They are used in testing to ensure the pre-computed hash is correct.

storage/src/trie_hash.rs
demosdemon: The function we previously had that computes the hash a runtime also wouldn't provide that.

demosdemon:
> Not opposed to removing the Default value since you're spot on about it being a footgun, but I don't love bleeding so much knowledge into storage about what represents an empty hash.
> 
> Maybe we should be using a different hash type outside of storage or something.

`empty_rlp_hash` is only used in `firewood` and `firewood-ffi`. I can move it out of storage and leave the default implementation removed in favor of the explicit `empty` function.

storage/src/node/branch.rs
alarso16: By coding error, you mean that it compiles but is incorrect, right?

This might be helpful to start moving away from compile-time decision about hashing mode

storage/src/trie_hash.rs
rkuris: True, but it would have been trivial to implement. All you would have to do is put the empty hash into the revision manager and you're done.

firewood/src/manager.rs
```diff
@@ -61,9 +59,6 @@ pub(crate) struct RevisionManager {
     proposals: Mutex<Vec<ProposedRevision>>,
     // committing_proposals: VecDeque<Arc<ProposedImmutable>>,
     by_hash: RwLock<HashMap<TrieHash, CommittedRevision>>,
-
-    #[cfg(feature = "ethhash")]
-    empty_hash: OnceLock<TrieHash>,
```
alarso16, line 66: Getting rid of this is a win for sure

firewood/src/v2/api.rs
```diff
@@ -154,6 +154,10 @@ pub enum Error {
     /// Revision not found
     #[error("revision not found")]
     RevisionNotFound,
+
+    /// An invalid root hash was provided
+    #[error("provided root hash is not a valid hash key")]
```
alarso16: nit: how is a root hash provided? In the eyes of any user, the root hash is retrieved, and only exposed through the function that returns this error.
```suggestion
    #[error("root hash found is not a valid hash key")]
```

firewood/src/db.rs
```diff
@@ -242,11 +239,8 @@ impl Db {
             proposals: counter!("firewood.proposals"),
         });
         describe_counter!("firewood.proposals", "Number of proposals created");
-        let manager = RevisionManager::new(
-            db_path.as_ref().to_path_buf(),
-            cfg.truncate,
-            cfg.manager.clone(),
-        )?;
+        let manager =
+            RevisionManager::new(db_path.as_ref().to_path_buf(), cfg.truncate, cfg.manager)?;
```
alarso16, line 245: Just curious - how come you don't need to clone the manager config?

alarso16: Separately (I'll create an issue if you agree), I think that `truncate` and `path` could go into the config. I know that truncate originally was used to distinguish between creating and opening a database, but that behavior has been changed

storage/src/trie_hash.rs
alarso16: Just a note: all of the "special" hashes in `eth` are listed [here](https://github.com/ava-labs/libevm/blob/b028cff60da9e4f5da857848d73463cf80d9ed83/core/types/hashes.go#L27). Notably, during the FFI work, there's also the empty code hash (which is different)

firewood/src/v2/api.rs
demosdemon: this error message is from doing `<&[u8]>::try_into()` for `TrieHash`. They're provided by the FFI caller.

firewood/src/db.rs
demosdemon: I added `Copy` to the options which trigger clippy forcing me to remove the clones. This leaked in from another change but was small enough that I didn't notice.

demosdemon: > Separately (I'll create an issue if you agree), I think that `truncate` and `path` could go into the config. I know that truncate originally was used to distinguish between creating and opening a database, but that behavior has been changed

Sure, I never noticed it but I agree that having the one extra field is out of place after those changes. However, I kind of like having path be a separate input parameter as it forces it to be provided. I have also seen builder patterns use that as a way of finalizing the builder into the built object by requiring the required parameters in that method.

So something like this:

```rust
impl RevisionManagerConfig {
    pub fn open(&self, path: impl AsRef<Path>) -> Result<Db, Error> { todo!(); }
}
```

ffi/src/lib.rs
```diff
@@ -238,7 +238,7 @@ fn get_from_root(
     key: &Value,
 ) -> Result<Value, String> {
     let db = db.ok_or("db should be non-null")?;
-    let requested_root = root.as_slice().try_into()?;
+    let requested_root = HashKey::try_from(root.as_slice()).map_err(|e| e.to_string())?;
```
demosdemon, line 241: fyi: this looks a bit weird now but it'll look better in a future PR that handles results a bit better

firewood/src/v2/api.rs
```diff
@@ -154,6 +197,10 @@ pub enum Error {
...
```
rkuris: Error should be more explicit: "not a valid hash key" doesn't say what is wrong and how to fix it.
```suggestion
    #[error("provided root hash is not a 32 byte hash")]
```

Also wondering if this is valid for any hash, not just root hashes.

nit: could provide how long it was: "Expected a hash of 32 bytes, got 31 bytes" is a dead giveaway as to what the problem is to the caller.

firewood/src/v2/api.rs
```diff
@@ -322,3 +369,28 @@ pub trait Proposal: DbView + Send + Sync {
...
     ) -> Result<Self::Proposal, Error>;
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[cfg(feature = "ethhash")]
+    fn test_default_root_hash() {
```
rkuris: nit: naming:
```suggestion
    fn test_eth_root_hash_equals_empty_rlp() {
```
Or a comment explaining why (prefer the name change though)

PR #1171, opened 2025-08-02
feat: remove `Batch` type alias

demosdemon:
The `Batch<K, V>` type alias has been removed in favor of using `IntoIterator<Item = BatchOp<K, V>>` in its place. The functions were already generic over `K` and `V`. Therefore, making them generic over the iterator type makes sense as well.

This also prevents unecessary allocations as we are no longer forced to collect into a vector unecessarily. Being able to iteratively build batch operations allow other range related improvements to be embedded in the iterator passed into `propose...`.

`BatchOp<K, V>` is constructed from `KeyValuePair` implementations and a default implementation is provided for `(K, V)` tuples. The behavior of using `DeleteRange` when the value is empty is preserved from FFI.

Also included is a refactor of the benchmarks to use the preferred batch iterator style. This is also used later on to benchmark future changes in range operations.

---

1. #1169
2. #1170
3. #1171
4. #1172
5. #1173
6. #1174
7. #1158

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -8,10 +8,11 @@
 
 use crate::merkle::{Merkle, Value};
 use crate::stream::MerkleKeyValueStream;
+pub use crate::v2::api::BatchOp;
 use crate::v2::api::{
-    self, FrozenProof, FrozenRangeProof, HashKey, KeyType, OptionalHashKeyExt, ValueType,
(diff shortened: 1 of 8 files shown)


firewood/src/db.rs
```diff
@@ -789,54 +785,82 @@ mod test {
         }
     }
 
+    struct Fuse<I>(Option<I>);
```
rkuris: This needs some doc to provide for better reuse

rkuris (approved):

PR #1175, opened 2025-08-04
feat(ffi-iterator): Part 1: Implement rust sync iterator

AminR443:
This PR is a step towards implementing the FFI iterator mentioned in #1009 and is an updated split of #1154 which is easy to review and also provides a better abstraction than previous one.

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -402,6 +402,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 Ok(free_area) => free_area,
                 Err(e) => {
                     errors.push(CheckerError::IO(e));
+                    free_list_iter.move_to_next_free_list();
                     continue;
                 }
             };
@@ -409,6 +410,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
             // check that the area is aligned
             if let Err(e) = self.check_area_aligned(addr, StoredAreaParent::FreeList(parent)) {
                 errors.push(e);
+                free_list_iter.move_to_next_free_list();
                 continue;
             }
 
@@ -422,6 +424,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                     expected_free_list: area_index,
                     parent,
                 });
+                free_list_iter.move_to_next_free_list();
                 continue;
             }
 
@@ -429,8 +432,10 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
             if let Err(e) = visited.insert_area(addr, area_size, StoredAreaParent::FreeList(parent))
             {
                 errors.push(e);
+                free_list_iter.move_to_next_free_list();
                 continue;
             }
+            update_progress_bar(progress_bar, visited);
 
             // collect the free list stats
             {
@@ -440,14 +445,13 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                 // collect the multi-page area count
                 if page_start(addr)
                     != page_start(
-                        addr.advance(area_size)
+                        addr.advance(area_size.saturating_sub(1))   // subtract 1 to get the last byte of the area
                             .expect("impossible since we checked in visited.insert_area()"),
                     )
                 {
                     multi_page_area_count = multi_page_area_count.saturating_add(1);
                 }
             }
-            update_progress_bar(progress_bar, visited);
         }
 
         (
@@ -582,7 +586,7 @@ mod test {
         test_write_free_area, test_write_header, test_write_new_node, test_write_zeroed_area,
     };
     use crate::nodestore::alloc::{AREA_SIZES, FreeLists};
-    use crate::{BranchNode, Child, LeafNode, NodeStore, Path, hash_node};
+    use crate::{BranchNode, Child, FreeListParent, LeafNode, NodeStore, Path, hash_node};
 
     #[derive(Debug)]
     struct TestTrie {
@@ -804,7 +808,7 @@ mod test {
                 );
                 next_free_block = Some(LinearAddress::new(high_watermark).unwrap());
                 let start_addr = LinearAddress::new(high_watermark).unwrap();
-                let end_addr = start_addr.advance(*area_size).unwrap();
+                let end_addr = start_addr.advance(*area_size - 1).unwrap();
                 if page_start(start_addr) != page_start(end_addr) {
                     multi_page_area_count = multi_page_area_count.saturating_add(1);
                 }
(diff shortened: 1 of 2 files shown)


firewood/src/db.rs
```diff
@@ -72,6 +89,44 @@ impl DbViewSyncBytes for Arc<HistoricalRev> {
...
         Ok(value)
     }
+
+    fn iterate(
+        &self,
+        state: IterationState,
+    ) -> Result<(Option<(Key, Value)>, IterationState), Error> {
+        if let IterationState::Done = state {
+            return Ok((None, state));
+        }
```
rkuris: This check belongs where the unreachable statement is in iterate_internal, otherwise you're always checking it twice, once here and again in the sub function.

firewood/src/db.rs
```diff
@@ -80,6 +135,17 @@ impl DbViewSyncBytes for Proposal<'_> {
         let value = merkle.get_value(key)?;
         Ok(value)
     }
+
+    fn iterate(
```
rkuris, line 134: This implements DbViewSyncBytes on both nodestores and a higher level object, a proposal, that contains a nodestore. You should be able to use a more generic implementation that operates on `NodeStore<T, S> where T: Parentable + RootReader + NodeReader`. The operation on `Proposal` should just be a method that dereferences the nodestore and calls `iterate` on that.

I think this would avoid needing to create an `iterate_internal` as well, since you'd then only have one `iterate` call.

rkuris (changes requested):
This is a great start! It needs more tests. Consider:

 - test_empty: verify that nothing is found on an empty nodestore
 - test_one_key: start at beginning, get one key, check for None on second, and None again on third. Perform same test starting at the key. Then test again starting after the key, which should get None (and None if you call it again).
 - test_multi_key: You have this I think (needs review still)
 - test_start_from: key that exists, key before all other keys, key between keys, keys after all keys
 - test_fuzz (optional, but not a terrible idea)

firewood/src/db.rs
AminR443: Done. Added following implementation which also abstracts away the duplication in `val_sync_bytes`.
```rust
impl<T, S> DbViewSyncBytes for NodeStore<T, S>
where
    NodeStore<T, S>: TrieReader,
    T: std::fmt::Debug,
    S: ReadableStorage,
```

firewood/src/db.rs
AminR443: Fixed.

AminR443:
Both changes are applied in the new commits. All the tests that you mentioned, except the fuzzing one, are implemented and working. I will look into fuzzing tomorrow. 

Also I noticed that we can avoid some duplication in db tests and make them cleaner. If nobody else is focused on them, I may work on them later after finishing the current stuff.

rkuris:
> Both changes are applied in the new commits. All the tests that you mentioned, except the fuzzing one, are implemented and working. I will look into fuzzing tomorrow.

Don't forget to hit the "re-request review" button when you are ready for the next review.

firewood/src/db.rs
```diff
@@ -66,27 +83,72 @@ impl<T: DbViewSyncBytes> DbViewSync for T {
...
+                }
+            }
+            IterationState::ContinueFromState(state) => {
+                MerkleKeyValueStream::from_internal_state(merkle.nodestore(), state)
+            }
+            IterationState::Done => {
+                return Ok((None, state));
+            }
+        };
+        let kv = it.next_internal();
```
qusuyan, line 118: Creating an iterator and call `next_internal()` once every call seem pretty wasteful. Can we reuse the iterator across multiple runs or directly return the kv pair without creating an iterator?

AminR443: Unfortunately not, this is the best way I came up with current code structure. The reason is we can't pass MerkleKeyValueStream to the FFI because of lifetime bound of the merkle reference it holds. So instead, as Ron suggested, we pass the state of the iterator to FFI and recreate the stream from internal state whenever next is called, this has some overhead but shouldn't be significant.

AminR443: Also this should be mostly used from FFI side and it will be extended to support batching as well, we can explore optimizations there.

rkuris: > Unfortunately not, this is the best way I came up with current code structure. The reason is we can't pass MerkleKeyValueStream to the FFI because of lifetime bound of the merkle reference it holds. So instead, as Ron suggested, we pass the state of the iterator to FFI and recreate the stream from internal state whenever next is called, this has some overhead but shouldn't be significant. 
> 

We shouldn't need to recreate the stream every time. We solved the lifetime problem with proposals which have a lifetime that depends on the database.

We just need a map of existing iterators and ensure they all get closed out when closing the database.

AminR443: But the point is, stream has a lifetime dependency on merkle, not the database directly. If I understand correctly, to do what you're suggesting, either we have to hold a merkle, similar to how we hold a db, or restructure the code so it's possible without doing that. What do you think is best to do with the current state of the PR?

firewood/src/db.rs
```diff
@@ -47,6 +47,17 @@ impl std::fmt::Debug for DbMetrics {
     }
 }
 
+/// State of an iteration on a revision, view, or proposal
+#[derive(Debug)]
+pub enum IterationState {
```
rkuris, line 52: IterationState should not be part of the public API.

firewood/src/db.rs
```diff
@@ -684,6 +748,105 @@ mod test {
...
+        let state = assert_iterator_next_done(&nodestore, IterationState::StartFromKey(None));
+        assert_iterator_next_done(&nodestore, state);
+    }
+
+    /// Tests if iterate works correctly with one key
+    #[tokio::test]
+    async fn test_iterate_single() {
+        let db = testdb().await;
+
+        let batch = batch_data(1, 0, 0);
```
rkuris, line 764: It's impossible from this vantage to tell what `batch_data` might do with these three numbers. Let's name this `create_kvs_from_range` and pass it a `range`, and call it like this:
```
    let batch = generate_test_keys(0..1)
```

You have one caller that wants to generate values that are shifted. I'd probably do this by creating a new offset batch, maybe like this:
```
    let batch1 = generate_test_keys(0..10);
    let batch2 = increase_values(&batch1, 10);
```
Another way to do this would be with an extension trait, but I think that's overkill for a test.

firewood/src/db.rs
```diff
@@ -684,6 +748,105 @@ mod test {
...
+        let root_hash = propose_on_testdb(&db, batch, true).await;
+        let view = db.view_sync(root_hash).unwrap();
+        let state = assert_iterator_next_is(
```
rkuris, line 767: This test should look something like this:
```
    let iterator = view.Iterate(None);
    assert_eq!(iterator.next(),
        Some((batch[0].key(), batch[0].value())));
    assert_eq!(iterator.next(), None);
```

firewood/src/db.rs
```diff
@@ -684,6 +748,105 @@ mod test {
...
+    #[test_case(0; "key before all keys")]
+    #[test_case(1; "first key")]
+    #[test_case(3; "key in middle")]
+    #[test_case(7; "last key")]
+    #[test_case(8; "key after last key")]
+    #[tokio::test]
+    async fn test_iterate_start_from(from: u32) {
+        let db = testdb().await;
+
+        let batch = batch_data(6, 1, 1);
```
rkuris, line 791: `let batch = generate_test_keys(1..=6)` is probably clearer here.

firewood/src/db.rs
```diff
@@ -684,6 +748,105 @@ mod test {
...
+        assert_iterator_next_done(
+            view.as_ref(),
+            IterationState::StartFromKey(Some(b"key1".as_slice().into())),
+        );
+    }
+
+    /// Tests if iterate works correctly starting from different keys
+    #[test_case(0; "key before all keys")]
+    #[test_case(1; "first key")]
+    #[test_case(3; "key in middle")]
```
rkuris, line 784: this key isn't in the middle, it's an existing key, very similar to last key. You need a test that looks for something that does not exist in between two keys such as if the keys are (1, 3) and you start at 2.

firewood/src/db.rs
```diff
@@ -883,6 +1121,57 @@ mod test {
...
+                key: format!("key{}", key_start.saturating_add(i)).into_bytes(),
+                value: format!("value{}", value_start.saturating_add(i)).into_bytes(),
+            })
+            .collect()
+    }
+
+    async fn propose_on_testdb(
+        db: &TestDb,
+        batch: Batch<Vec<u8>, Vec<u8>>,
+        commit: bool,
```
rkuris, line 1165: This bool hurts readability. If you pass `true`, this isn't just doing what the method implies, it actually creates a new revision.

If you want a method that proposes and commits, consider making one, and just using `propose` directly from db when you just want a proposal.

PR #1176, opened 2025-08-04
feat(checker): checker returns all errors found in the report

qusuyan:
Beforehand, checker returns an error whenever it encounters a problem. With this PR, it will return all errors it can find:

1. If a node in the trie is problematic, it will not traverse any deeper from this node, but checker will return multiple errors with nodes that do not have ancestry relationships. 
2. If a free area is problematic, it will continue traversing and the iterator controls when to stop. 
3. No error will ever be reported when visiting leaked areas.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -331,10 +331,7 @@ impl Db {
     }
 
     /// Check the database for consistency
-    pub async fn check(
-        &self,
-        opt: CheckOpt,
-    ) -> Result<CheckerReport, firewood_storage::CheckerError> {
+    pub async fn check(&self, opt: CheckOpt) -> CheckerReport {
         let latest_rev_nodestore = self.manager.current_revision();
         latest_rev_nodestore.check(opt)
     }
@@ -799,15 +796,15 @@ mod test {
 
             // check the database for consistency, sometimes checking the hashes
             let hash_check = rng.borrow_mut().random();
-            if let Err(e) = db
+            let report = db
                 .check(CheckOpt {
                     hash_check,
                     progress_bar: None,
                 })
-                .await
-            {
+                .await;
+            if !report.errors.is_empty() {
                 db.dump(&mut std::io::stdout()).await.unwrap();
-                panic!("error: {e}");
+                panic!("error: {:?}", report.errors);
             }
         }
     }
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -144,23 +144,19 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
             .root_as_maybe_persisted_node()
             .and_then(|node| self.root_hash().map(|root_hash| (node, root_hash)))
             .and_then(|(root, root_hash)| {
-                match root
-                    .as_linear_address()
-                    .ok_or(vec![CheckerError::UnpersistedRoot])
-                    .and_then(|root_address| {
-                        self.visit_trie(
-                            root_address,
-                            root_hash.into_hash_type(),
-                            &mut visited,
-                            opt.progress_bar.as_ref(),
-                            opt.hash_check,
-                        )
-                    }) {
-                    Ok(trie_stats) => Some(trie_stats),
-                    Err(e) => {
-                        errors.extend(e);
-                        None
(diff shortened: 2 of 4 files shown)


qusuyan:
NOT READY YET - currently it will not report any statistics on the merkle trie if there are errors.

firewood/src/db.rs
```diff
@@ -799,15 +796,15 @@ mod test {
...
+            let report = db
                 .check(CheckOpt {
                     hash_check,
                     progress_bar: None,
                 })
-                .await
-            {
+                .await;
+            if !report.errors.is_empty() {
                 db.dump(&mut std::io::stdout()).await.unwrap();
```
rkuris, line 806: I doubt we still want or need this. There is an option in fwdctl now that can do this.

storage/src/checker/mod.rs
```diff
@@ -115,46 +117,57 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
+            .unwrap_or_default();
 
         // 3. check the free list - this can happen in parallel with the trie traversal
         if let Some(progress_bar) = &opt.progress_bar {
             progress_bar.set_message("Traversing free lists...");
         }
-        let free_list_stats = self.visit_freelist(&mut visited, opt.progress_bar.as_ref())?;
+        let (free_list_stats, free_list_traverse_errors) =
+            self.visit_freelist(&mut visited, opt.progress_bar.as_ref());
+        errors.extend(free_list_traverse_errors);
```
rkuris, line 170: nit: since this is the only caller, seems like we should just pass `&mut free_list_traverse_errors` to `visit_freelist` instead of allocating a new array and extending errors later.

You might have it this way to support more parallelism, in which case you can ignore this comment.

storage/src/checker/mod.rs
```diff
@@ -167,12 +180,21 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
...
         // TODO: add leaked areas to the free list
 
-        Ok(CheckerReport {
+        let physical_bytes = match self.physical_size() {
+            Ok(physical_bytes) => physical_bytes,
+            Err(e) => {
+                errors.push(e.into());
+                0
+            }
+        };
```
rkuris, line 189: I thought we weren't going to do any physical size checks? This is bad if we use a raw device.

storage/src/checker/mod.rs
```diff
@@ -392,10 +446,14 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
             }
             update_progress_bar(progress_bar, visited);
```
rkuris, line 450: this gets missed if we `continue` due to errors, which means the progress bar won't get updated when you get a bunch of errors.

rkuris (approved):

firewood/src/db.rs
qusuyan: What are you referring to specifically? We don't want the fuzz test anymore?

storage/src/checker/mod.rs
qusuyan: We are not doing any physical size checks. This is purely for collecting statistics.

storage/src/checker/mod.rs
qusuyan: This will not happen - the progress bar records, out of the entire db size, how much we have visited. If we encountered an error, we probably will ignore this area and recollect it when handling leaked areas. 

Speaking of which, I will check again when visited areas should be inserted so that invalid spaces are properly ignored.

storage/src/checker/mod.rs
qusuyan: I kind of want `visit_trie` and `visit_freelist` to have the same return type `(Stats, Vec<CheckerError>)`. The checker is not performance critical so gonna leave it here for now.

firewood/src/db.rs
qusuyan: Oh regarding `dump`? The reason we had it here in the first place was not because of dump but because we want to fuzz testing and fuzz testing in `fwdctl` is REALLY SLOW. I will skip this comment in this PR. If you have a different opinion do let me know and I can address it in a separate PR.

PR #1177, opened 2025-08-05
chore: allow FreeListIterator to skip to next free list

qusuyan:
For the checker: if the free area is invalid, its next pointer may also contain garbage data, in which case we should stop iterating through the current free list and skip to the next one.

### DIFF
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -771,6 +771,21 @@ impl<'a, S: ReadableStorage> FreeListsIterator<'a, S> {
             }
         }
     }
+
+    // TODO: this function will be used by the checker to ignore free areas pointed by an invalid parent free area
+    #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


storage/src/nodestore/alloc.rs
```diff
@@ -771,6 +771,21 @@ impl<'a, S: ReadableStorage> FreeListsIterator<'a, S> {
             }
         }
     }
+
+    // TODO: this function will be used by the checker to ignore free areas pointed by an invalid parent free area
+    #[allow(dead_code)]
```
demosdemon, line 776: nit: use expect so you don't forget to remove this later.

You could also use it now where you naturally transition to the next size so you use the same method for that purpose.

qusuyan: My next PR is almost ready so let me see if I can fit this method into the natural transition there :)

PR #1180, opened 2025-08-06
fix: Fix race around reading nodes during commit

rkuris:
We called persist_at before the node was actually persisted to disk in the io-uring case. What we really want is to keep all the nodes around in memory while they are being flushed, and then once they are added to the general cache, we can mark them as fully persisted.
      
This is a lot better since now nodes will be used directly while they are being flushed instead of having a disk address that either hasn't been written to disk or at least isn't yet in the cache.


ffi/firewood_test.go
```diff
@@ -984,3 +987,74 @@ func TestGetFromRoot(t *testing.T) {
...
+			got, err := db.GetFromRoot(root, key)
+			if err != nil {
+				return fmt.Errorf("reader %d, iteration %d: GetFromRoot error: %w", readerID, j, err)
+			}
+			if !bytes.Equal(got, val) {
+				return fmt.Errorf("reader %d, iteration %d: expected %q, got %q", readerID, j, val, got)
+			}
+
+			// Add small delays and yield to increase race chances
+			if j%100 == 0 {
```
alarso16, line 1025: I'm not really sure if this is necessary, but it won't hurt

ffi/firewood_test.go
```diff
@@ -984,3 +987,74 @@ func TestGetFromRoot(t *testing.T) {
...
+
+	// Use multiple goroutines to stress test concurrent access
+	const numReaders = 10
+	results := make(chan error, numReaders)
+	finish := make(chan struct{})
+	wg := sync.WaitGroup{}
+
+	readLots := func(readerID int) error {
+		for j := 0; ; j++ {
+			t.Log("Reader", readerID, "iteration", j)
```
alarso16: Maybe delete this log, it's super super verbose - I only used it to ensure I was actually reading during a commit

ffi/firewood_test.go
```diff
@@ -984,3 +987,74 @@ func TestGetFromRoot(t *testing.T) {
...
+	db := newTestDatabase(t)
+
+	key, val := []byte("key"), []byte("value")
+	allKeys, allVals := kvForTest(10000) // large number to encourage contention
+	allKeys = append(allKeys, key)
+	allVals = append(allVals, val)
+
+	// Create a proposal
+	p, err := db.Propose(allKeys, allVals)
+	r.NoError(err, "Propose single key-value pair")
```
alarso16: nit: this message isn't correct - oops

alarso16 (commented):
The Go test was written a little haphazardly

PR #1183, opened 2025-08-06
feat(checker): add error when node data is bigger than area size

qusuyan:
When reading a node from an area, we simply read from the address of the area without checking if the read will span beyond the given area. This PR fills in the gap.

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -261,6 +261,16 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         let (area_index, area_size) = self.area_index_and_size(subtrie_root_address)?;
         let (node, node_bytes) = self.read_node_with_num_bytes_from_disk(subtrie_root_address)?;
 
+        // check if the node fits in the area
+        if node_bytes > area_size {
+            return Err(vec![CheckerError::NodeLargerThanArea {
+                area_start: subtrie_root_address,
+                area_size,
+                node_bytes,
+                parent,
+            }]);
+        }
+
         // if the node has a value, check that the key is valid
         let mut current_path_prefix = path_prefix.clone();
         current_path_prefix.0.extend_from_slice(node.partial_path());
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -114,7 +114,6 @@ struct SubTrieMetadata {
 }
 
 /// [`NodeStore`] checker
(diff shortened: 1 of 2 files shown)


storage/src/checker/mod.rs
```diff
@@ -261,6 +259,16 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         let (area_index, area_size) = self.area_index_and_size(subtrie_root_address)?;
         let (node, node_bytes) = self.read_node_with_num_bytes_from_disk(subtrie_root_address)?;
 
+        // check if the node fits in the area
+        if node_bytes > area_size {
```
rkuris: This feels like an off by one error but might be okay. Is it possible to remove the impossible below and just figure out what the minimum AreaIndex is first?

rkuris (approved):
Curious also if we check the size byte and verify it matches what should be there

storage/src/checker/mod.rs
qusuyan: This is not off by one but a pretty significant difference - see #1184

storage/src/checker/mod.rs
```diff
@@ -261,6 +259,16 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
         let (area_index, area_size) = self.area_index_and_size(subtrie_root_address)?;
         let (node, node_bytes) = self.read_node_with_num_bytes_from_disk(subtrie_root_address)?;
 
+        // check if the node fits in the area, equal is not allowed due to 1-byte area size index
```
rkuris, line 262: Why isn't `read_node_with_num_bytes_from_disk` including the extra byte? Are there any callers that aren't going to add 1?

Seems like it might be better to pass in the maximum allowed size to `read_node_with_num_bytes_from_disk` so it can stop once it hits the end of the area and return an error, rather than returning the number of bytes and checking. Are there any callers that don't know how big the area is supposed to be?

rkuris (approved):

storage/src/checker/mod.rs
qusuyan: They all should except they are not reading this one extra byte - see `storage/src/nodestore/mod.rs:713`, we can always get area size from the skipped byte but none is doing that. The reason I am not adding one in the return value here is that, when computing the minimum storage space needed (as part of collecting statistics in the checker), I don't want to include the extra byte since stored area index is firewood specific storage metadata.

PR #1189, opened 2025-08-07
fix(fwdctl)!: db path consistency + no auto-create

rkuris:
I found that a small typo in the database creates a new one. This isn't the expected behavior, and is a regression from the ffi changes that always create a database.

In addition, cleaned up the database path option so it's exactly the same across all subcommands.

- Use consistent `--db` flag across all commands instead of positional arguments
- Set `create_if_missing=false` for most commands (except `create`)
- Refactor tests to use cleaner `.arg()` syntax for single arguments
- Compute the test path one time for all tests
- Remove unnecessary string conversions

### DIFF
--- a/fwdctl/src/main.rs
+++ b/fwdctl/src/main.rs
@@ -3,6 +3,8 @@
 
 #![doc = include_str!("../README.md")]
 
+use std::path::PathBuf;
+
 use clap::{Parser, Subcommand};
 use firewood::v2::api;
 
@@ -14,6 +16,20 @@ pub mod get;
 pub mod graph;
 pub mod insert;
 pub mod root;
+
+#[derive(Clone, Debug, Parser)]
+pub struct DatabasePath {
+    /// The database path. Defaults to firewood.db
+    #[arg(
+        long = "db",
+        short = 'd',
+        required = false,
+        value_name = "DB_NAME",
+        default_value = default_db_path().into_os_string(),
+        help = "Name of the database"
+    )]
+    pub dbpath: PathBuf,
+}
 
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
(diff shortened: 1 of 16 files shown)


fwdctl/src/main.rs
```diff
@@ -15,6 +17,20 @@ pub mod graph;
...
 
+#[derive(Clone, Debug, Parser)]
+pub struct DatabasePath {
+    /// The database path. Defaults to firewood.db
+    #[arg(
+        long = "db",
+        short = 'd',
+        required = false,
+        value_name = "DB_NAME",
+        default_value = default_db_path().into_os_string(),
```
demosdemon: nit: 

```suggestion
        default_value_os_t = default_db_path(),
```

fwdctl/src/main.rs
```diff
@@ -15,6 +17,20 @@ pub mod graph;
 pub mod insert;
 pub mod root;
 
+#[derive(Clone, Debug, Parser)]
+pub struct DatabasePath {
```
rkuris, line 21: Maybe this could go in the top level so it's used by every subcommand

demosdemon: there's the `#[arg(global = true)]` [option](https://docs.rs/clap/latest/clap/struct.Arg.html#method.global) to allow it to be seen by subcommands as well but I've not had great luck with it in the past

alarso16 (commented):
I think this is a great way to address both options!

fwdctl/src/main.rs
```diff
@@ -15,6 +17,20 @@ pub mod graph;
...
+    /// The database path. Defaults to firewood.db
+    #[arg(
+        long = "db",
+        short = 'd',
+        required = false,
+        value_name = "DB_NAME",
+        default_missing_value_os = default_db_path(),
```
demosdemon: `default_value_os_t` is appropriate, not `default_missing_value_os`.

Not only does it not compile, but `default_missing_value_os` is for if `--db` is provided but with no argument. In that case, the `default_missing_value_os` value would be the argument. But, not providing `--db` would not generate a default value.

demosdemon (approved):

PR #1191, opened 2025-08-07
feat(checker): checker truncate erroneous free lists

qusuyan:
This is the second PR for the checker to fix the DB image. This PR does two things:

1. The checker now takes a `&mut self` instead of `&self`. This is required to update the header freelists. 
2. For each error with a valid parent reference, the checker will truncate the free list, effectively truncate the erroneous free area from the list.


qusuyan:
This PR is not yet ready: I need to write test cases to ensure that fixes are correct.

PR #1192, opened 2025-08-08
test: ban `rand::rng()` and provide an env seeded alternative

demosdemon:


storage/src/test_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// # Panics
+    ///
+    /// Panics if the environment variable is present but contains invalid UTF-8
+    /// or does not parse as a valid `u64`.
+    #[track_caller]
+    #[must_use]
+    pub fn from_env() -> Option<Self> {
+        let s = std::env::var_os(Self::ENV)?
+            .into_string()
+            .unwrap_or_else(|_| panic!("{} must be a valid UTF-8 string", Self::ENV));
```
rkuris: `unwrap_or_else` `panic` is exactly what `expect` does with less noise

"FIREWOOD_TEST_SEED must be a valid UTF-8 string" is a tautology. If it won't convert to a UTF-8 string, you can be pretty sure it isn't a u64 either, so let's tell them how to fix it:

```suggestion
.expect(concat!(Self::ENV, " must contain a valid u64"))
```

Could also be written with other combinators like `and_then` to consolidate the error message.

storage/src/test_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+        Some(Self::new(s.parse().unwrap_or_else(|_| {
+            panic!("{} must be a valid u64", Self::ENV)
+        })))
+    }
+
+    /// Creates a new `SeededRng` with a random seed generated by the OS.
+    pub fn from_random() -> Self {
+        let seed = rand::rngs::OsRng.unwrap_err().next_u64();
+        eprintln!(
```
rkuris: This message used to always be rendered, even when explicitly set, and that was quite helpful. If I set the variable, I want the value to show up in the crash report.

storage/src/test_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+        SeededRng::next_u64(self)
+    }
+
+    #[inline]
+    fn fill_bytes(&mut self, dst: &mut [u8]) {
+        SeededRng::fill_bytes(self, dst);
+    }
+}
+
+impl rand::RngCore for &SeededRng {
```
rkuris, line 153: Not sure why, but I usually see this written as `impl<T: SeededRng> rand::RngCore for &T` but what you have might be better.

Do we really need both thouugh? Deref might work fine if you just have this one.

storage/src/test_utils.rs
demosdemon: `expect` has the extra baggage of requiring the err type also implement display, which here it does not because `into_string` returns the `OsString` as the err variant.

However, sure `FIREWOOD_TEST_SEED` must be an ascii string which is more strict that utf-8. But the difference is that it must be a string and not 8 bytes encoded to an integer.

storage/src/test_utils.rs
demosdemon: `impl<T: SeededRng> rand::RngCore for &T` works if *we* provided the `SeededRng` trait or the `RngCore` trait. But we provide neither (`SeededRng` also isn't a trait, the `rand` trait is `SeedableRng`). This normally isn't a thing because `RngCore`'s methods are all mutable. But here, we implement it for a reference of `SeededRng`, which is a strictly different type, because mutable references are acquired through the borrowed cell.

This is used elsewhere, particularly in the tests that need to consume the RngCore type (anywhere that uses [`sample_iter`](https://docs.rs/rand/0.9.2/rand/trait.Rng.html#method.sample_iter)).

storage/src/test_utils.rs
demosdemon: oh, no, expect requires debug. But, `concat` does not work with the constant

```
error: expected a literal
  --> storage/src/test_utils.rs:64:29
   |
64 |             .expect(concat!(Self::ENV, " must be a valid UTF-8 string"));
   |                             ^^^^^^^^^
   |
   = note: only literals (like `"foo"`, `-42` and `3.14`) can be passed to `concat!()`
```

rkuris: I didn't have any trouble with expect on an OsString: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d87c7cec49b0854a2d5d6b19f43542c5

PR #1193, opened 2025-08-08
chore: AreaIndex newtype

rkuris:


storage/src/nodestore/alloc.rs
```diff
@@ -121,41 +121,125 @@ pub const fn index_name(index: usize) -> &'static str {
...
+    fn from(index: usize) -> Self {
+        assert!(
+            (index < Self::NUM_AREA_SIZES),
+            "Area index out of bounds: {} >= {}",
+            index,
+            Self::NUM_AREA_SIZES
+        );
+        AreaIndex(index as u8)
+    }
+}
```
demosdemon: Definitely not a fan of the `From` implementations that can panic. That is what `TryFrom` is for. Hiding a panic in the from implementation makes it really east for callers to do the wrong thing. If I were someone new to this code, why would I use `AreaIndex::new` over `56_u8.into()`?

Callers should be forced to use the method that returns a `None` or otherwise acknowledges that the cast is unsafe and unchecked.

In many cases, it appears the conversion could be obviated with a bit more care on how we get the `u8` values and perhaps preventing even needing to have the u8 value.

storage/src/nodestore/alloc.rs
```diff
@@ -87,7 +87,7 @@ pub fn area_size_hash() -> TrieHash {
 #[must_use]
 pub const fn size_from_area_index(index: AreaIndex) -> u64 {
     #[expect(clippy::indexing_slicing)]
-    AREA_SIZES[index as usize]
+    AREA_SIZES[index.as_usize()]
```
demosdemon: We can go a step further an turn `AREA_SIZES` into a newtype as well that implements `Index<AreaIndex>`. If not up for that in this change, `AREA_SIZES` should probably be an associated constant of `AreaIndex`.

storage/src/nodestore/alloc.rs
```diff
@@ -121,41 +121,125 @@ pub const fn index_name(index: usize) -> &'static str {
 }
 
 /// The type of an index into the [`AREA_SIZES`] array
-/// This is not usize because we can store this as a single byte
-pub type AreaIndex = u8;
+/// This is not usize because we store this as a single byte
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
+pub struct AreaIndex(u8);
```
demosdemon: This should be pushed down into a module to hide `.0` from everything in this module.

storage/src/checker/mod.rs
```diff
@@ -573,7 +571,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
                     .advance(*area_size)
                     .expect("address overflow is impossible");
                 if next_addr <= *leaked_range.end {
-                    leaked.push((current_addr, area_index as AreaIndex));
+                    leaked.push((current_addr, AreaIndex::from(area_index)));
```
demosdemon: casting `area_index` into `AreaIndex` looks like the type should have already been an `AreaIndex`. Perhaps we need an iterator that yields typed `AreaIndex` instead of forcing callers to make this cast a lot.

storage/src/nodestore/alloc.rs
```diff
@@ -121,41 +121,125 @@ pub const fn index_name(index: usize) -> &'static str {
...
 
-const NUM_AREA_SIZES: usize = AREA_SIZES.len();
-const MIN_AREA_SIZE: u64 = AREA_SIZES[0];
-const MAX_AREA_SIZE: u64 = AREA_SIZES[NUM_AREA_SIZES - 1];
+impl AreaIndex {
```
demosdemon: This type should have a safe const method to get the area size from `self` and ideally we don't expose `AREA_SIZES` at all. There's many instances of `AREA_SIZES[area_size_index.as_usize()]` that could be `area_size_index.size()`.

storage/src/macros.rs
```diff
@@ -66,3 +66,25 @@ macro_rules! firewood_gauge {
...
+/// The macro will panic at compile time if the value is negative or >= `NUM_AREA_SIZES`.
+macro_rules! area_index {
+    ($v:expr) => {
+        const {
+            if $v >= $crate::nodestore::alloc::AreaIndex::NUM_AREA_SIZES {
+                panic!("Constant area index out of bounds");
+            } else {
+                $crate::nodestore::alloc::AreaIndex::from_u8_unchecked($v as u8)
+            }
+        }
```
demosdemon: love the macro but it doesn't need to check the `NUM_AREA_SIZES` const or call the unchecked method:

```rust
const {
    match $crate::nodestore::alloc::AreaIndex::new($v) {
        Some(v) => v,
        None => panic!("Constant area index out of bounds"),
    }
}
```

This does require the input `$v` to resolve to a `u8`. We could add `($v & 0xff) as u8` if we wanted to hide that from the caller. But, I don't feel it's necessary in most cases.

storage/src/checker/mod.rs
```diff
@@ -972,10 +999,10 @@ mod test {
             test_write_free_area(
                 &nodestore,
                 None,
-                area_size_index as AreaIndex,
+                AreaIndex::from(area_size_index),
```
demosdemon: spot where an iterator of typed AreaIndexes would be useful

storage/src/nodestore/alloc.rs
```diff
@@ -407,14 +491,17 @@ impl FreeArea {
...
         self.write_to(encoded);
     }
 
     fn from_storage_reader(mut reader: impl Read) -> std::io::Result<(Self, AreaIndex)> {
         let area_index = reader.read_byte()?;
         let free_area = reader.next_value()?;
-        Ok((free_area, area_index))
+        Ok((
+            free_area,
+            AreaIndex::new(area_index).expect("area_index should be valid"),
```
demosdemon: should return an error here, not panic. Or, you could add a method to the extension trait that provides `read_byte` to also provide `read_area_index`.

storage/src/nodestore/alloc.rs
```diff
@@ -445,16 +532,17 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
...
-        let index: AreaIndex = area_stream.read_byte().map_err(|e| {
+        let index: AreaIndex = AreaIndex::new(area_stream.read_byte().map_err(|e| {
             self.storage.file_io_error(
                 Error::new(ErrorKind::InvalidData, e),
                 addr.get(),
                 Some("deserialize".to_string()),
             )
-        })?;
+        })?)
+        .expect("area index should be valid");
```
demosdemon: also should return an error instead of panicking.

storage/src/nodestore/alloc.rs
```diff
@@ -518,26 +606,26 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
 
             // Return the address of the newly allocated block.
             trace!("Allocating from free list: addr: {address:?}, size: {index}");
-            return Ok(Some((address, index as AreaIndex)));
+            return Ok(Some((address, AreaIndex::from(index))));
```
demosdemon: `index` should probably be an `AreaIndex` before here but is also looks like we assert that it is equal to `read_index` which is already typed.

storage/src/nodestore/alloc.rs
```diff
@@ -590,28 +678,28 @@ impl<S: WritableStorage> NodeAllocator<'_, S> {
         firewood_counter!(
             "firewood.delete_node",
             "Nodes deleted",
-            "index" => index_name(area_size_index as usize)
+            "index" => index_name(area_size_index.as_usize())
```
demosdemon: looks like this `index_name` method should be on the typed `AreaIndex`.

storage/src/nodestore/mod.rs
```diff
@@ -742,22 +742,28 @@ impl<T, S: ReadableStorage> NodeStore<T, S> {
...
-        let index: AreaIndex = area_stream.read_byte().map_err(|e| {
+        let index: AreaIndex = AreaIndex::new(area_stream.read_byte().map_err(|e| {
             self.storage.file_io_error(
                 Error::new(ErrorKind::InvalidData, e),
                 addr.get(),
                 Some("area_index_and_size".to_string()),
             )
-        })?;
+        })?)
+        .expect("area index should be valid");
```
demosdemon: shouldn't need to panic in a method that can return an error

storage/src/nodestore/alloc.rs
```diff
@@ -840,17 +928,17 @@ pub mod test_utils {
         let mut encoded_node = Vec::new();
         node.as_bytes(0, &mut encoded_node);
         let encoded_node_len = encoded_node.len() as u64;
-        let area_size_index = area_size_to_index(encoded_node_len).unwrap();
+        let area_size_index = AreaIndex::from_size(encoded_node_len).unwrap();
         let mut stored_area_bytes = Vec::new();
-        node.as_bytes(area_size_index, &mut stored_area_bytes);
+        node.as_bytes(area_size_index.get(), &mut stored_area_bytes);
```
demosdemon: feels like `as_bytes` should receive `AreaIndex` as a parameter instead of the `u8`.

storage/src/nodestore/alloc.rs
```diff
@@ -121,41 +121,125 @@ pub const fn index_name(index: usize) -> &'static str {
...
+
+impl From<AreaIndex> for usize {
+    fn from(area_index: AreaIndex) -> Self {
+        area_index.as_usize()
+    }
+}
+
+impl fmt::Display for AreaIndex {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        std::fmt::Display::fmt(&self.get(), f)
```
demosdemon: curious if display here should include the `index_name` (maybe if `f.alternate()` is set?)

demosdemon (changes requested):

storage/src/nodestore/alloc.rs
rkuris: Yes, but that's arguably a different PR, and is a lot easlier.

storage/src/nodestore/alloc.rs
rkuris: Debug could, but probably not display.

storage/src/nodestore/alloc.rs
qusuyan: wait I just addressed it

storage/src/nodestore/alloc.rs
```diff
@@ -20,262 +20,29 @@
...
-
-impl From<NonZeroU64> for LinearAddress {
-    fn from(addr: NonZeroU64) -> Self {
-        LinearAddress(addr)
-    }
-}
-
 /// `FreeLists` is an array of `Option<LinearAddress>` for each area size.
-pub type FreeLists = [Option<LinearAddress>; AREA_SIZES.len()];
+pub type FreeLists = [Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES];
```
demosdemon, line 45: Looks like `FreeLists` is a candidate for our next newtype! It can have a implementation similar to `Children<T>` and also implement `Index<AreaIndex>`.

storage/src/nodestore/primitives.rs
```diff
@@ -0,0 +1,353 @@
...
+    1024 << 8,
+    1024 << 9,
+    1024 << 10,
+    1024 << 11,
+    1024 << 12,
+    1024 << 13,
+    1024 << 14,
+];
+
+pub fn area_size_iter() -> impl DoubleEndedIterator<Item = (AreaIndex, u64)> {
```
demosdemon, line 47: This is perhaps "less elegant" than providing a named iterator. But, this is really good enough for most cases. And in the event we really do need a named type, we can expand it to `std::iter::Map<std::iter::Enumerate<std::slice::Iter<&'static, u64>>, fn((usize, &'static u64)) -> (AreaIndex, u64)>` as a type alias.

storage/src/nodestore/alloc.rs
```diff
@@ -497,7 +247,7 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
                 *free_stored_area_addr = free_head;
             } else {
                 let (free_head, read_index) = FreeArea::from_storage(self.storage, address)?;
-                debug_assert_eq!(read_index as usize, index);
+                debug_assert_eq!(read_index.as_usize(), index.as_usize());
```
demosdemon: nit: should be able to compare directly.

demosdemon (approved):
Looks like the remaining instances of `try_from` might go away with #1194 since it's iterator would also be able to guarantee the indexes are `AreaIndex`. So, I'll let them go today.

storage/src/nodestore/primitives.rs
qusuyan: I will add a TODO here for named iterators. If things really come up I can make the change in a different PR :)

PR #1195, opened 2025-08-08
ci: automatically label PRs from external contributors

demosdemon:
This checks if the pull request is from a fork and not whether or not the PR is from a member of the org; there is no easy way for us to do that via the API.


rkuris (approved):
nifty

PR #1196, opened 2025-08-11
feat: Parallel insert into Merkle trie

bernard-avalabs:
Added support for performing parallel inserts to a Merkle trie. This PR consists of 3 main changes:

1. Added `read_for_update_ext_deleted` and `append_deleted` to nodestore, which allows tracking deleted nodes in an external vector. The `insert` and `insert_helper` functions in Merkle have also be updated to use these functions, which allows `insert_helper` to take a &self instead of a &mut self, allowing it to be used by the worker threads.
2. Added `insert_parallel` and `insert_parallel_helper` that send insert operations that can't change the root node to workers in the the worker pool to perform. For the remaining insert operations, `insert_parallel` falls back to calling `insert`. This can only be done by first calling `wait` to block until all of the outstanding operations are complete.
3. Added WorkerPool and MerkleParallel. WorkerPool creates a thread for each child of the root node. Communication is handled by sending different MerkleOps and Results to channels shared by the parent thread and the workers. MerkleParallel is a wrapper around Merkle and WorkerPool to provide a clean interface for the caller to use. 

Testing: `test_parallel_insert` inserts a batch of random keys in parallel, verifies the keys exist in the trie, inserts a second batch of keys, and then verifies both batches are in the trie. It also illustrates how to use the `MerkleParallel` interface.

This PR should have minimal impact on the performance of normal inserts. I have not yet tested the performance impact on batch inserts, but I am hopeful that it will deliver appreciable improvements especially in cases where we can perform parallel disk reads to fetch nodes needed to complete the insert operations.


firewood/src/merkle.rs
```diff
@@ -110,6 +112,188 @@ fn get_helper<T: TrieReader>(
...
+#[derive(Debug)]
+/// Parallel compatible wrapper for a Merkle trie
+pub struct MerkleParallel<T, S> {
+    merkle_arc: Option<Arc<T>>,
+    root_node: Option<Node>,
+    worker_pool: WorkerPool<S>,
+}
+
+#[derive(Debug)]
+/// Error type that can return from calling parallel insert and its related functions.
```
rkuris, line 124: nit: not clear if these are errors from the main thread or the worker threads.

Most of these errors are really just an internal error and should never happen, so it might make more sense to bundle them all up (except perhaps for the IO error).

firewood/src/merkle.rs
```diff
@@ -110,6 +112,188 @@ fn get_helper<T: TrieReader>(
...
+
+impl<S: ReadableStorage + 'static> MerkleParallel<Merkle<NodeStore<MutableProposal, S>>, S> {
+    /// Constructor for `MerkleParallel`.
+    #[must_use]
+    pub fn new(mut m: Merkle<NodeStore<MutableProposal, S>>) -> Self {
+        let root_node = std::mem::take(m.nodestore.mut_root());
+        let merkle_arc = Arc::new(m);
+        let worker_pool = WorkerPool::new(merkle_arc.clone());
+        MerkleParallel {
+            merkle_arc: Some(merkle_arc),
```
rkuris, line 161: Not clear why this is an option. Under what conditions do we want `merkle_arc` to be `None`?

firewood/src/merkle.rs
```diff
@@ -110,6 +112,188 @@ fn get_helper<T: TrieReader>(
...
+    /// Parallel insert into the Merkle trie.
+    ///
+    /// ## Errors
+    ///
+    /// Can return a `ParallelInsertError` which could be due to a `FileIoError` from calling `insert_helper`
+    /// or `insert`, a channel error from sending to a worker, or a None merkle, which means the merkle
+    /// structure has been cleared without re-setting it to a new merkle structure.
+    pub fn insert(&mut self, key: &[u8], value: Value) -> Result<(), ParallelInsertError<S>> {
+        let Some(merkle_arc) = self.merkle_arc.take() else {
+            return Err(ParallelInsertError::MerkleUnset);
```
rkuris, line 176: same as prior comment. Just make this an Arc instead of an `Option<Arc<T>>`.

rkuris (changes requested):
I see the approach you're taking, but it's not exactly what I thought you would do.

You're basically sending an arc-wrapped merkle across threads and preventing the merkle from getting mutated during insertion.

I assumed you were going to create merkle objects with their root set to the child root and just perform the insert with the first nibble removed on each child. This way, you're reusing the regular merkle insert routines, rtather than having your own that don't own the merkle.

I think there will be a lot less code re-use with this approach.

bernard-avalabs:
> I assumed you were going to create merkle objects with their root set to the child root and just perform the insert with the first nibble removed on each child. This way, you're reusing the regular merkle insert routines, rtather than having your own that don't own the merkle.
> 
> I think there will be a lot less code re-use with this approach.

Most of the work for insert is done in `insert_helper`, and `insert_helper` was already nicely designed to operate on an arbitrary sub-trie rooted at a specified node. My approach just calls `insert_helper` from a worker once it has identified that the insert cannot modify the root. I'm not sure if there is a lot less code re-use by calling `insert_helper` directly on a sub-trie instead of creating a Merkle object on the sub-trie. 

Both approaches will still need additional code to inspect the root node and its children array to determine if an insert can modify the root (this is what `insert_parallel_helper` does in my approach, which is only called once on each insert). Similarly, both approaches will need to wait for the workers to complete any outstanding operations before they can fallback to perform a non-parallel insert. 

Perhaps some of the code in the InsertData arm of the worker thread can be slightly simplified, but I think it would still require similar code to determine whether or not a Merkle object has already been created for the sub-trie.

firewood/src/merkle.rs
bernard-avalabs: The functions `wait` and `wait_params` call into_inner on a merkle_arc to convert it back to a normal merkle. This serves two purposes: 1) it allows the main thread to perform mutable calls on merkle and 2) it allows MerkleParallel to wrap around a different merkle afterwards by calling `update_merkle`, which enables the worker pool to be reused.

After calling `wait` or `wait_params`, MerkleParallel should no longer have a reference to a merkle_arc. My understanding is that, if it is not an Option, then I would need to replace self.merkle_arc with something else. I'm not sure what that would be after essentially releasing the underlying merkle from MerkleParallel.

Issue #1194, opened 2025-08-08
Add a `FreeLists` newtype


demosdemon:
Looks like `FreeLists` is a candidate for our next newtype! It can have a implementation similar to `Children<T>` and also implement `Index<AreaIndex>`.

_Originally posted by @demosdemon in https://github.com/ava-labs/firewood/pull/1193#discussion_r2264021850_

KushnerykPavel:
I'll work on it

PR #1197, opened 2025-08-11
fix: add FreeLists type

KushnerykPavel:
resolve https://github.com/ava-labs/firewood/issues/1194

### DIFF
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -41,8 +41,109 @@ const fn var_int_max_size<VI>() -> usize {
     const { (size_of::<VI>() * 8 + 7) / 7 }
 }
 
-/// `FreeLists` is an array of `Option<LinearAddress>` for each area size.
-pub type FreeLists = [Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES];
+/// `FreeLists` is a wrapper around an array of `Option<LinearAddress>` for each area size.
+/// It provides safe indexing with `AreaIndex` and similar functionality to `Children<T>`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct FreeLists([Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES]);
+
+#[expect(unsafe_code)]
+// SAFETY: FreeLists is a  wrapper around an array of Option<LinearAddress>
+// which is already Pod and Zeroable, so FreeLists is also Pod and Zeroable
+unsafe impl bytemuck::Zeroable for FreeLists {}
+#[expect(unsafe_code)]
+// SAFETY: FreeLists is a wrapper around an array of Option<LinearAddress>
+// which is already Pod and Zeroable, so FreeLists is also Pod and Zeroable
+unsafe impl bytemuck::Pod for FreeLists {}
+
+impl Default for FreeLists {
+    fn default() -> Self {
+        Self([None; AreaIndex::NUM_AREA_SIZES])
+    }
+}
+
+impl std::ops::Index<AreaIndex> for FreeLists {
+    type Output = Option<LinearAddress>;
+
+    fn index(&self, index: AreaIndex) -> &Self::Output {
+        &self.0[index.as_usize()]
+    }
+}
+
+impl std::ops::IndexMut<AreaIndex> for FreeLists {
+    fn index_mut(&mut self, index: AreaIndex) -> &mut Self::Output {
+        &mut self.0[index.as_usize()]
+    }
+}
+
+impl std::ops::Index<usize> for FreeLists {
+    type Output = Option<LinearAddress>;
+
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.0[index]
+    }
+}
+
+impl std::ops::IndexMut<usize> for FreeLists {
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        &mut self.0[index]
+    }
+}
+
+impl FreeLists {
+    /// Create a new `FreeLists` with all entries set to `None`.
+    pub const fn new() -> Self {
+        Self([None; AreaIndex::NUM_AREA_SIZES])
+    }
+
+    /// Get a reference to the underlying array.
+    pub const fn as_array(&self) -> &[Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES] {
+        &self.0
+    }
+
+    /// Get a mutable reference to the underlying array.
+    pub fn as_array_mut(&mut self) -> &mut [Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES] {
+        &mut self.0
+    }
+
+    /// Get an iterator over the free lists.
+    pub fn iter(&self) -> std::slice::Iter<'_, Option<LinearAddress>> {
+        self.0.iter()
+    }
+
+    /// Get a mutable iterator over the free lists.
+    pub fn iter_mut(&mut self) -> std::slice::IterMut<'_, Option<LinearAddress>> {
+        self.0.iter_mut()
+    }
+
+    /// Get the length of the free lists array.
+    pub const fn len(&self) -> usize {
+        AreaIndex::NUM_AREA_SIZES
+    }
+
+    /// Check if the free lists array is empty (always false).
+    pub const fn is_empty(&self) -> bool {
+        false
+    }
+}
+
+impl<'a> IntoIterator for &'a FreeLists {
+    type Item = &'a Option<LinearAddress>;
+    type IntoIter = std::slice::Iter<'a, Option<LinearAddress>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+impl<'a> IntoIterator for &'a mut FreeLists {
+    type Item = &'a mut Option<LinearAddress>;
+    type IntoIter = std::slice::IterMut<'a, Option<LinearAddress>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter_mut()
+    }
+}
 
 /// A [`FreeArea`] is stored at the start of the area that contained a node that
 /// has been freed.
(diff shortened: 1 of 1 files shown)


storage/src/nodestore/alloc.rs
```diff
@@ -41,8 +41,113 @@ const fn var_int_max_size<VI>() -> usize {
...
+#[expect(unsafe_code)]
+// SAFETY: FreeLists is a wrapper around an array of Option<LinearAddress>
+// which is already Pod and Zeroable, so FreeLists is also Pod and Zeroable
+unsafe impl bytemuck::Pod for FreeLists {}
+
+impl Default for FreeLists {
+    fn default() -> Self {
+        Self([None; AreaIndex::NUM_AREA_SIZES])
+    }
+}
```
rkuris: All these can be done without unsafe using derive. Use `use bytemuck_derive::{Pod, Zeroable};` then:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Pod, Zeroable, Defauilt)]
#[repr(transparent)]
pub struct FreeLists([Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES]);
```

storage/src/nodestore/alloc.rs
```diff
@@ -41,8 +41,113 @@ const fn var_int_max_size<VI>() -> usize {
...
+
+impl std::ops::Index<AreaIndex> for FreeLists {
+    type Output = Option<LinearAddress>;
+
+    fn index(&self, index: AreaIndex) -> &Self::Output {
+        self.0
+            .get(index.as_usize())
+            .expect("AreaIndex is guaranteed to be within bounds")
```
rkuris: It's impossible for the index to be out of range, unless someone uses unsafe to construct an AreaSize, but the compiler cannot check this. This is a candidate for `get_unchecked`:
```suggestion
        unsafe { self.0.get_unchecked(index.as_usize()) }
```

storage/src/nodestore/alloc.rs
```diff
@@ -41,8 +41,113 @@ const fn var_int_max_size<VI>() -> usize {
...
+    fn index(&self, index: usize) -> &Self::Output {
+        self.0.get(index).expect("index out of bounds")
+    }
+}
+
+impl std::ops::IndexMut<usize> for FreeLists {
+    fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+        self.0.get_mut(index).expect("index out of bounds")
+    }
+}
```
rkuris: Remove all these. All the callers either (a) are already calling `AreaIndex::as_usize` or hardcode a constant 0, which is `AreaIndex::MIN`.

storage/src/nodestore/alloc.rs
```diff
@@ -41,8 +41,113 @@ const fn var_int_max_size<VI>() -> usize {
...
+
+impl FreeLists {
+    /// Create a new `FreeLists` with all entries set to `None`.
+    pub const fn new() -> Self {
+        Self([None; AreaIndex::NUM_AREA_SIZES])
+    }
+
```
rkuris: We don't need `new` since we have `Default` implemented.

storage/src/nodestore/alloc.rs
```diff
@@ -41,8 +41,113 @@ const fn var_int_max_size<VI>() -> usize {
...
+    /// Get a mutable iterator over the free lists.
+    pub fn iter_mut(&mut self) -> std::slice::IterMut<'_, Option<LinearAddress>> {
+        self.0.iter_mut()
+    }
+
+    /// Get the length of the free lists array.
+    pub const fn len(&self) -> usize {
+        AreaIndex::NUM_AREA_SIZES
+    }
+
```
rkuris: This appears to only be used in tests, so either (a) remove it and have the callers use AreaIndex::NUM_AREA_SIZES; or (b) mark this method `#[cfg(test)]`. (a) is better.

storage/src/nodestore/alloc.rs
```diff
@@ -41,8 +41,113 @@ const fn var_int_max_size<VI>() -> usize {
...
+    /// Check if the free lists array is empty (always false).
+    pub const fn is_empty(&self) -> bool {
+        false
+    }
```
rkuris: This looks suspect. I think it's only used in tests, and the caller just has an assert that it's empty. I think the intent is to compare it to the default value.

Two things to do: #[cfg(test)] and compare self to FreeLists::default()

storage/src/nodestore/alloc.rs
```diff
@@ -952,4 +1056,48 @@ mod tests {
...
+        // Test indexing with usize (backward compatibility)
+        free_lists[0] = None;
+        assert_eq!(free_lists[0], None);
+
+        // Test iterator functionality
+        let mut count = 0;
+        for _ in &free_lists {
+            count += 1;
+        }
+        assert_eq!(count, AreaIndex::NUM_AREA_SIZES);
```
rkuris:
```suggestion
        assert_eq!(free_lists.iter().count(), AreaIndex::NUM_AREA_SIZES);
```

storage/src/nodestore/alloc.rs
```diff
@@ -952,4 +1056,48 @@ mod tests {
...
+
+        // Test len and is_empty
+        assert_eq!(free_lists.len(), AreaIndex::NUM_AREA_SIZES);
+        assert!(!free_lists.is_empty());
```
rkuris: These don't add much value, so please remove them.

storage/src/nodestore/alloc.rs
```diff
@@ -952,4 +1056,48 @@ mod tests {
...
+
+        // Test default
+        let default_free_lists = FreeLists::default();
+        assert_eq!(default_free_lists.len(), AreaIndex::NUM_AREA_SIZES);
+        for item in &default_free_lists {
+            assert_eq!(item, &None);
+        }
```
rkuris, line 980:
```suggestion
       assert_eq!(default_free_lists.iter().find(|item| item.is_some()), None);
```

storage/src/nodestore/primitives.rs
```diff
@@ -278,10 +278,10 @@ impl LinearAddress {
     }
 
     /// The maximum area size available for allocation.
-    pub const MAX_AREA_SIZE: u64 = *AREA_SIZES.last().unwrap();
+    pub const MAX_AREA_SIZE: u64 = *AREA_SIZES.last().expect("AREA_SIZES is not empty");
```
rkuris: We don't need this change, since it's impossible to hit this case. When evaluating a constexpr, it's okay to use unwrap().

storage/src/nodestore/primitives.rs
```diff
@@ -278,10 +278,10 @@ impl LinearAddress {
...
 
     /// The minimum area size available for allocation.
-    pub const MIN_AREA_SIZE: u64 = *AREA_SIZES.first().unwrap();
+    pub const MIN_AREA_SIZE: u64 = *AREA_SIZES.first().expect("AREA_SIZES is not empty");
```
rkuris: Same here

storage/src/nodestore/alloc.rs
KushnerykPavel: As I see I can't use unsafe code according to #![deny(unsafe_code)] directive in storage/src/lib.rs . Any other comments have fixed! Thank you very much for review!

storage/src/nodestore/alloc.rs
```diff
@@ -28,21 +28,64 @@ use crate::node::branch::{ReadSerializable, Serializable};
...
+impl FreeLists {
+    /// Get a reference to the underlying array.
+    pub const fn as_array(&self) -> &[Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES] {
+        &self.0
+    }
+
+    /// Get a mutable reference to the underlying array.
+    pub const fn as_array_mut(&mut self) -> &mut [Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES] {
+        &mut self.0
+    }
```
demosdemon: Not a fan of these as array methods. Although these are const methods, I'd prefer if we implemented [`Index<AreaIndex>`](https://doc.rust-lang.org/stable/std/ops/trait.Index.html) and [`IndexMut<AreaIndex>`](https://doc.rust-lang.org/stable/std/ops/trait.IndexMut.html) so that we can ensure that the array is only ever indexed with the `ArrayIndex` newtype.

Additionally, if we need to have a method that exposes the inner array  (which I would like to avoid), it should also include the constant size instead of erasing the size and returning a slice. `&[T; N]` and `&[T]` are two different types. 

```rust
    assert_eq!(size_of::<&[u8; 42]>(), size_of::<usize>());
    assert_eq!(size_of::<&[u8]>(), 2 * size_of::<usize>());
```

For the `size_of` reason alone, we should prefer to include the constant size.

storage/src/nodestore/alloc.rs
```diff
@@ -28,21 +28,64 @@ use crate::node::branch::{ReadSerializable, Serializable};
...
+}
+
+impl<'a> IntoIterator for &'a mut FreeLists {
+    type Item = &'a mut Option<LinearAddress>;
+    type IntoIter = std::slice::IterMut<'a, Option<LinearAddress>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter_mut()
+    }
+}
```
demosdemon: We also need an `IntoIterator for FreeLists` implementation.

---

I would also like a filtered version of all three iterators that yields `(AreaIndex, LinearAddress)` instead of `Option<LinearAddress>`. This filtered iterator should be equivalent to 

```rust
free_lists
    .into_iter()
    .enumerate()
    .filter_map(|(slot, value)| value.map((
        // SAFETY: `slot` is guaranteed to be a valid AreaIndex.
        unsafe { AreaIndex::new_unchecked(slot as u8) }, 
        value,
     )))
```

Where we yield a typed AreaIndex instead of the `usize`.

storage/src/nodestore/alloc.rs
```diff
@@ -28,21 +28,64 @@ use crate::node::branch::{ReadSerializable, Serializable};
...
     const { (size_of::<VI>() * 8 + 7) / 7 }
 }
 
-/// `FreeLists` is an array of `Option<LinearAddress>` for each area size.
-pub type FreeLists = [Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES];
+/// `FreeLists` is a wrapper around an array of `Option<LinearAddress>` for each area size.
+/// It provides safe indexing with `AreaIndex` and similar functionality to `Children<T>`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Pod, Zeroable, Default)]
+#[repr(transparent)]
+pub struct FreeLists([Option<LinearAddress>; AreaIndex::NUM_AREA_SIZES]);
```
demosdemon: Accidentally deleted this comment: since this is growing and we're adding additional support types, lets move this into a dedicated module. This will also hide the inner value from code not in the same module forcing callers to use our type-safe APIs.

rkuris:
This hasn't had any activity for a while, and is now quite out of date. I'm switching this to draft.

I think this work is still important to reduce tech debt.

rkuris:
Closed as stale. This work should be started anew to address https://github.com/ava-labs/firewood/issues/1194

PR #1199, opened 2025-08-11
feat(checker): add checker fix template

qusuyan:


qusuyan:
> I see you did the fix I was gonna ask for. Did I publish that comment? If so, then I didn't mean to delete it 😅

LOOOL I did a fix Ron asked for yesterday during our 1:1, maybe you are just asking for the same thing xD

PR #1205, opened 2025-08-13
fix: remove extra nibble from path prefix when computing ethhash at account level

qusuyan:
This is part 2 of #1137. Critically, when computing the hash for the special case, we need to remove the extra nibble from the path prefix, which will be appended in the end. For whatever reason checker works correctly with and without this fix :/


rkuris:
We need to run a full bootstrap with these changes to avoid breaking any invariant. This is very sensitive code and difficult to debug once deployed.

rkuris:
This looks like a "make the code better" fix. It's parent was closed as stale, so we are closing this one as stale as well.

PR #1207, opened 2025-08-13
chore: update ClassifiedChildren to return only what's needed

qusuyan:
This is part 3 of #1137


qusuyan:
Depends on #1205

RodrigoVillar:
The parent PR was closed as stale, so we are closing this one as stale as well.

PR #1208, opened 2025-08-14
feat(ffi-iterator): Part 1: Initial Implementation of FFI iterator (Rust side)

AminR443:
This PR is a step towards implementing the FFI iterator mentioned in #1009. The implementation is kept simple as possible for now without any batching and directly fetches the next item of the iterator. 

Steps on the list:
- [x] Implement FFI functions (`iter_on_root`, `iter_on_proposal`, `iter_next`)  in Rust, and also necessary changes in `firewood` crate for `MerkleKeyValueStream` to work with `Arc<NodeStore<..>>` as well
- [ ] Implement a simple iterator in Go with `Next()`, `Key()`, `Value()`, and `Err()`. (Already done, to be included in next PR)
- [ ] Explore batching for reduced overhead
- [ ] Optimize memory management for batching (avoiding large copies)
- [ ] Proper resource deallocation
- [ ] More tests

#1154 and #1175 are also closed because this work handles the same goal efficiently.

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -36,18 +36,14 @@ use std::ops::Deref;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::{Arc, Mutex, RwLock};
 
-use firewood::db::{
-    BatchOp as DbBatchOp, Db, DbConfig, DbViewSync as _, DbViewSync, DbViewSyncBytes, Proposal,
-};
+use firewood::db::{Db, DbConfig, DbViewSync as _, DbViewSyncBytes, Proposal};
 use firewood::manager::{CacheReadStrategy, RevisionManagerConfig};
 
 pub use crate::value::*;
 use firewood::merkle;
-use firewood::merkle::Merkle;
 use firewood::stream::MerkleKeyValueStream;
-use firewood::v2::api;
-use firewood::v2::api::{DbView, HashKey, KeyValuePairIter};
-use firewood_storage::{Committed, FileBacked, ImmutableProposal, NodeStore, TrieReader};
+use firewood::v2::api::{HashKey, KeyValuePairIter};
+use firewood_storage::{Committed, FileBacked, ImmutableProposal, NodeStore};
 use metrics::counter;
 
 #[cfg(unix)]
@@ -85,6 +81,8 @@ pub struct DatabaseHandle<'p> {
     // Order of fields are important, streams must be dropped before proposals, and proposals
     // should be dropped before database handle to ensure lifetime orders
     streams: RwLock<HashMap<IteratorId, MerkleIterator<'p>>>,
+    /// Maps proposal IDs to the iterators that depend on them
+    proposal_iterators: RwLock<HashMap<ProposalId, Vec<IteratorId>>>,
     /// List of outstanding proposals, by ID
     proposals: RwLock<HashMap<ProposalId, Proposal<'p>>>,
     /// A single cached view to improve performance of reads while committing
@@ -101,12 +99,17 @@ enum MerkleIterator<'v> {
 }
 
 impl Iterator for MerkleIterator<'_> {
-    type Item = Result<(merkle::Key, merkle::Value), api::Error>;
+    type Item = Result<(merkle::Key, merkle::Value), String>;
+
     fn next(&mut self) -> Option<Self::Item> {
         match self {
-            MerkleIterator::View(iter) => iter.next_sync(),
-            MerkleIterator::Proposal(iter) => iter.next_sync(),
-            MerkleIterator::Dropped => None,
+            MerkleIterator::View(iter) => iter.next_sync().map(|r| r.map_err(|e| e.to_string())),
+            MerkleIterator::Proposal(iter) => {
+                iter.next_sync().map(|r| r.map_err(|e| e.to_string()))
+            }
+            MerkleIterator::Dropped => Some(Err(
+                "iterator is dropped and no longer pointing to a valid nodestore".to_string(),
+            )),
         }
     }
 }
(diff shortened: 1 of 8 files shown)


ffi/src/lib.rs
```diff
@@ -34,55 +34,93 @@ use std::ffi::{CStr, CString, c_char};
...
 ///
 /// These handles are passed to the other FFI functions.
 ///
 #[derive(Debug)]
 pub struct DatabaseHandle<'p> {
-    /// List of oustanding proposals, by ID
-    // Keep proposals first, as they must be dropped before the database handle is dropped due to lifetime
-    // issues.
+    /// List of iterators, by ID
+    // Order of fields are important, streams must be dropped before proposals, and proposals
```
qusuyan, line 88: Sounds like `impl Drop` is what you need (instead of relying on the order of fields in the struct)

ffi/src/lib.rs
```diff
@@ -34,55 +34,93 @@ use std::ffi::{CStr, CString, c_char};
...
+
+impl Iterator for MerkleIterator<'_> {
+    type Item = Result<(merkle::Key, merkle::Value), String>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
+            MerkleIterator::View(iter) => iter.next_sync().map(|r| r.map_err(|e| e.to_string())),
+            MerkleIterator::Proposal(iter) => {
+                iter.next_sync().map(|r| r.map_err(|e| e.to_string()))
+            }
```
qusuyan: nit: 
```suggestion
            MerkleIterator::View(iter) | MerkleIterator::Proposal(iter) => {
                iter.next_sync().map(|r| r.map_err(|e| e.to_string()))
            }
```

ffi/src/lib.rs
```diff
@@ -34,55 +34,93 @@ use std::ffi::{CStr, CString, c_char};
...
     cached_view: Mutex<Option<(HashKey, Box<dyn DbViewSyncBytes>)>>,
-
     /// The database
     db: Db,
 }
 
+#[derive(Debug)]
+enum MerkleIterator<'v> {
+    View(MerkleKeyValueStream<'v, NodeStore<Committed, FileBacked>>),
+    Proposal(MerkleKeyValueStream<'v, NodeStore<Arc<ImmutableProposal>, FileBacked>>),
```
qusuyan: Does the storage need to be `FileBacked`? Maybe consider using generics here.

qusuyan: Also, does it matter to distinguish between `View` and `Proposal`? If not there is a `Parentable` trait that both `Arc<ImmutableProposal>` and `Committed` implements.

ffi/src/lib.rs
```diff
@@ -560,6 +765,24 @@ pub unsafe extern "C" fn fwd_drop_proposal(
...
+            .read()
+            .map_err(|_| "proposal iterators lock is poisoned")?;
+        guard.get(&proposal_id).cloned()
+    };
+
+    if let Some(iter_ids) = iterators {
+        let mut guard = db.streams.write().map_err(|_| "streams lock is poisoned")?;
+        for id in iter_ids {
+            let _ = guard.remove(&id);
+            guard.insert(id, MerkleIterator::Dropped);
```
qusuyan: nit: `remove()` here seem redundant

firewood/src/stream.rs
```diff
@@ -67,11 +65,45 @@ enum NodeStreamState {
     },
 }
 
+/// A reference to the merkle, could be borrowed, or behind an Arc
+#[derive(Debug)]
+pub enum MerkleRef<'a, T> {
+    /// reference with borrowing the merkle
+    Borrowed(&'a T),
+    /// referencing the merkle via Arc
+    Owned(Arc<T>),
```
qusuyan, line 74: Why do you need an `Owned` alternative here? `MerkleRef` always have the same lifetime as `Borrowed(&'a T)`. If you need an owned alternative (like lock guards), you may need to create a different type instead of adding a type to the enum.

qusuyan (changes requested):

ffi/src/lib.rs
AminR443: Yes, but didn't want to change how it was done originally in this PR.

ffi/src/lib.rs
AminR443: those two `iter`s don't have the same type, so this is not possible. I will check your comment about generics, and if that seems to work, then yes, this would be possible after refactoring.

ffi/src/lib.rs
AminR443: Good points, will check.

firewood/src/stream.rs
AminR443: This is the most straightforward approach to allow the streams to work with `Arc<T>` as well, while maintaining compatibility with `&'a T`. In case of `MerkleRef::Owned`, we're simply not using the lifetime parameter which is no harm I suppose. How would you suggest solving this without an enum and with a type instead?

qusuyan: You can always get an `&'a T` from `Arc<T>` using `Arc::as_ref()`. This is probably what you really need.

AminR443: That's the tricky part, for this to work, you would need to have that arc that you got the reference from to be in scope. You can get the reference but you can not pass that along to the other functions and structs that require it. That reference is passed down several times, so it's kinda inevitable to own and hold the arc.

AminR443: Marking this as resolved, let me know if you have any better way of handling this.

AminR443:
@qusuyan Thanks for the feedback! The interface looks much cleaner now. You were right about the enum, I was focused on the lifetime issues that I overlooked the fact that all of them implement stream/iterator.

I've wrapped the stream with another struct that implements Iterator for now. Direct implementation was possible but would've needed changing all the `stream.next()` calls, which felt beyond this PR's scope.

@demosdemon I know you're working on the async removal. Happy to help if needed,  just let me know and I'll handle the stream changes in a separate PR.

ffi/src/lib.rs
qusuyan: Can you put a TODO here in case we forget?

PR #1209, opened 2025-08-14
chore: Moved merkle.rs to mod.rs in merkle directory

bernard-avalabs:
Planning to extend the merkle trie in separate files. First step is to move merkle.rs to mod.rs in the merkle directory.


firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,252 @@
...
+#[derive(Debug)]
+enum Response {
+    // return the new root of the subtrie at the given nibble
+    //Root(u8, Option<Node>),
+    Root(Option<Node>),
+    Error(FileIoError),
+}
+
+
+static THREADPOOL: OnceLock<ThreadPool> = OnceLock::new();
```
qusuyan, line 56: I'm not sure if this is the right approach here - I personally would prefer just having an async function so that the caller can decide the runtime to use and can share thread with other code instead of spawning new threads.

bernard-avalabs: Sorry, this was not meant to be part of this PR. I'll close this PR for now.

Issue #798, opened 2025-02-20
Unify sync/async code


rkuris:
We can use the [maybe_async crate](https://docs.rs/maybe-async/latest/maybe_async/) to avoid code duplication for many database operations, and produce either a synchronous or async runtime.

Currently we're using sync primitives at the bottom of the stack but we could easily switch this to async to better support web services that use firewood as a backend.

PR #1219, opened 2025-08-20
feat(async-removal): Phase 4 - Make `DbView` synchronous

demosdemon:
This also removes all final `async` and `stream` usages in the code, with the  exception of `firewood-benchmark` which must use tokio for handling the prometheus metrics listener.

This finalizes the `async` removal process.

# Why `ArcDynDbView`?

You may wonder why I replaced `Box<dyn DbViewSyncBytes>` with `Arc<dyn DybDbView ...>`. This is because in both cases where it was used, we already had an Arc object that the compiler could coerce into our `Arc<dyn ...>` variant without adding an additional box into the mix. This removes one layer of pointer indirection.

Fixes: #798


firewood/src/manager.rs
```diff
@@ -280,7 +277,7 @@ impl RevisionManager {
                 provided: root_hash,
             })?;
 
-        Ok(Box::new(proposal))
+        Ok(proposal)
```
demosdemon, line 280: This and the `Ok(committed)` above are the crux of the `ArcDynDbView` changes that allow us to return the arc here directly instead of wrapping it around a separate heap allocation.

Cargo.toml
```diff
@@ -76,7 +76,6 @@ sha2 = "0.10.9"
 smallvec = "1.15.1"
 test-case = "3.3.1"
 thiserror = "2.0.12"
-tokio = "1.46.1"
```
rkuris, line 79: 🌟

firewood/src/iter.rs
```diff
@@ -68,138 +63,119 @@ enum NodeStreamState {
...
     /// Returns None when iteration is complete, or Some(Result) with either a node or an error.
     fn next_internal(&mut self) -> Option<Result<(Key, SharedNode), FileIoError>> {
-        let Self { state, merkle } = &mut *self;
-
-        match state {
-            NodeStreamState::StartFromKey(key) => {
-                match get_iterator_intial_state(*merkle, key) {
-                    Ok(state) => self.state = state,
-                    Err(e) => return Some(Err(e)),
+        'outer: loop {
```
rkuris, line 93: nit: the whole point of next_internal was so we could share the code whether it was sync or async. We probably can get rid of it now, but I'm not sure it's worth doing rn.

firewood/src/iter.rs
```diff
@@ -1474,23 +1348,19 @@ mod tests {
...
+        assert_iterator_is_exhausted(iter);
     }
 
-    async fn check_stream_is_done<S>(mut stream: S)
-    where
-        S: FusedStream + Unpin,
-    {
-        assert!(stream.next().await.is_none());
-        assert!(stream.is_terminated());
+    fn assert_iterator_is_exhausted<I: Iterator>(mut iter: I) {
```
rkuris: nit:
```suggestion
    fn assert_iterator_is_exhausted<I: FusedIterator>(mut iter: I) {
```

This now only works for fused iterators.

rkuris (approved):
I'm a little ambivalent about `try_extend`. This seems cleanly implemented but only has one use case, which might make it more readable to do it inline.

PR #1221, opened 2025-08-20
feat(ffi-refactor): finish ffi refactor

demosdemon:


ffi/firewood_test.go
```diff
@@ -513,21 +513,7 @@ func TestDropProposal(t *testing.T) {
...
-	}
-
-	// Check all operations on the fake proposal.
-	_, err = proposal.Get([]byte("non-existent"))
-	r.Contains(err.Error(), "proposal not found", "Get(fake proposal)")
-	_, err = proposal.Propose([][]byte{[]byte("key")}, [][]byte{[]byte("value")})
-	r.Contains(err.Error(), "proposal not found", "Propose(fake proposal)")
-	err = proposal.Commit()
-	r.Contains(err.Error(), "proposal not found", "Commit(fake proposal)")
+	r.NoError(err, "Root of dropped proposal should still be accessible")
```
rkuris, line 516: Why is this? Seems like if it's dropped, that revision should no longer be available.

demosdemon: This was more for the commit case and not drop case where we could be able to read the hash after committing. 

This isn't a requirement either but it is a side-effect of storing the hash on an inline struct separate from the handle (which allows us to fully skip the ffi layer when grabbing the hash).

ffi/firewood_test.go
```diff
@@ -513,21 +513,7 @@ func TestDropProposal(t *testing.T) {
 	_, err = proposal.Get([]byte("non-existent"))
 	r.ErrorIs(err, errDroppedProposal)
 	_, err = proposal.Root()
-	r.ErrorIs(err, errDroppedProposal)
-
-	// Attempt to "emulate" the proposal to ensure it isn't internally available still.
-	proposal = &Proposal{
```
alarso16, line 519: This was a garbage test (sorry about that), but it is still probably worth it ensuring that the data can't be retrieved (and dropped completely). Maybe use `GetFromRoot` to ensure that the keys are no longer accessible

ffi/memory.go
```diff
@@ -138,6 +135,9 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
 
 // Close releases the memory associated with the Database.
 //
+// This is not safe to call while there are any outstanding Proposals. All proposals
```
alarso16: I thought it was? If this is not the case, we have a memory leak in coreth/subnet-evm

ffi/firewood_test.go
demosdemon: this won't be possible as we have no way of minting a token. the token is a pointer to memory now. saw your edit, we can check for the keys.

ffi/memory.go
demosdemon: This is true as of this PR because the proposal handle also contains a reference to the same database handle:

https://github.com/ava-labs/firewood/blob/29cff9f0cee0ac65b5b945c0d1a79e4791f34755/ffi/src/proposal.rs#L10-L17

and before we kept all of the proposals on the handle behind a lock. We can stop using the lock by using direct pointers but that requires us to forget about the proposal and give it up to go.

I see that'll require a change here:

https://github.com/ava-labs/coreth/blob/215620be0590fe3c775529c5ad7fe08b44091a22/triedb/firewood/database.go#L385

demosdemon:
I'm gonna move this back to draft and get the coreth change ready first.

ffi/memory.go
demosdemon: Thanks @alarso16, the change needed in coreth was added in https://github.com/ava-labs/coreth/commit/91e0d7c05ffcaaf380044c1a1065e4123c9a0548

rkuris:
Closes #1272

demosdemon:
My [coreth PR](https://github.com/ava-labs/coreth/pull/1173) will be merged soon. This is r4r again.

rkuris (approved):

PR #1222, opened 2025-08-20
feat(ffi-refactor): refactor cached view

demosdemon:
Because of earlier PRs, we can take advantage of the arc and eagerly drop the lock on the cache afer loading it.

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -1,0 +1,44 @@
+use firewood::v2::api::{self, ArcDynDbView, HashKey};
+use metrics::counter;
+
+use crate::DatabaseHandle;
+
+impl DatabaseHandle<'_> {
+    pub(crate) fn get_root(&self, root: HashKey) -> Result<ArcDynDbView, api::Error> {
+        // construct outside of the closure so that when the closure is dropped
+        // without executing, it triggers the hit counter
+        let token = CachedViewHitOnDrop;
+        self.cached_view.get_or_try_insert_with(root, move |key| {
+            token.miss();
+            self.db.view(HashKey::clone(key))
+        })
+    }
+}
+
+/// A RAII metrics helper that tracks cache hits and misses for database views.
+///
+/// This type uses the drop pattern to automatically record cache metrics:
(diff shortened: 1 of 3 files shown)


ffi/src/handle.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl DatabaseHandle<'_> {
+    pub(crate) fn get_root(&self, root: HashKey) -> Result<ArcDynDbView, api::Error> {
+        // construct outside of the closure so that when the closure is dropped
+        // without executing, it triggers the hit counter
+        let token = CachedViewHitOnDrop;
+        self.cached_view.get_or_try_insert_with(root, move |key| {
+            token.miss();
+            self.db.view(HashKey::clone(key))
+        })
```
rkuris: nit: this seems unnecessarily complicated, and separates the key names from the function that increments them.

Perhaps the metric management belongs with the ArcCache, since it does seem that any future caller will want a metric around hit rates anyway.

The other choice is a mutable or a second closure to call on hit.

```suggestion
        let mut metric = "firewood.ffi.cached_view.hit";
        self.cached_view.get_or_try_insert_with(root, move |key| {
            metric = "firewood.ffi.cached_view.miss";
            self.db.view(HashKey::clone(key))
        })
        counter!(metric).increment(1);
```

or
```suggestion
        self.cached_view.get_or_try_insert_with(root, move |key| {
            counter!("firewood.ffi.cached_view.miss").increment(1);
            self.db.view(HashKey::clone(key))
        }, |_key| counter!("firewood.ffi.cached_view.hit").increment(1))
```

rkuris (approved):

ffi/src/handle.rs
demosdemon: I agree that it is perhaps a bit more complicated than usual but I disagree with:

> and separates the key names from the function that increments them.

Moving the cache keys into arccache would move them further away from the relevant code. I chose the RAII token specifically to keep the counters closer to the cache loading operation and I'm not sure either of your suggestions are cleaner.

alarso16 (approved):

PR #1174, opened 2025-08-03
feat(ffi): add BorrowedBytes type

demosdemon:

PR #1223, opened 2025-08-20
feat(ffi-refactor): add OwnedSlice and OwnedBytes (2/8)

demosdemon:
This is similar to the BorrowedBytes from #1174 but for bytes returns by Rust back to Go.

This does not actually use the new type yet as there are too many places that need to be updated. However, this introduces the new type and it's free method so that we can start using it in the future.

Depends On: #1222


ffi/memory.go
```diff
@@ -103,6 +103,111 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
...
+	if b == nil {
+		return nil
+	}
+
+	this := *b               // copy so we can reset before freeing,
+	b.owned = C.OwnedBytes{} // reset to clear the pointer
+
+	if this.owned.ptr == nil {
+		return nil
+	}
```
rkuris: This feels racey. If two threads call Free() on the same object at the same time, we're in trouble, but there isn't a comment about that and it doesn't appear to be handled correctly.

rkuris (approved):
Race condition should be addressed before merging.

ffi/memory.go
demosdemon: It's been a while since I've worked with this level of go but I was under the impression that copying the pointer to a local variable is the best way to handle the potential race condition.

We copy the shared variable to a local variable and indiscriminately set the shared variable to zero. Then, we check the local variable for nilness. If the local variable is not nil, then we are the only ones with the non-nil pointer because we cleared the shared pointer.

demosdemon: There's also the atomics module. I can use that to atomically swap the pointer out.

ffi/memory.go
```diff
@@ -103,6 +103,111 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
...
+}
+
+// intoError converts the RustOwnedBytes into an error. This is used for methods
+// that return a RustOwnedBytes as an error type.
+//
+// If the RustOwnedBytes is nil or has already been freed, it returns nil.
+// Otherwise, the bytes will be copied into Go memory and converted into an
+// error. The original RustOwnedBytes will be freed after this operation.
+func (b *RustOwnedBytes) intoError() error {
+	if b == nil {
```
alarso16: This isn't possible - if you tried to call a method on a nil pointer, it should panic

ffi/memory.go
```diff
@@ -103,6 +103,111 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
...
+	}
+
+	return C.GoBytes(unsafe.Pointer(b.owned.ptr), C.int(b.owned.len))
+}
+
+// fromOwnedBytes creates a RustOwnedBytes from a C.OwnedBytes.
+//
+// The caller is responsible for calling Free() on the returned RustOwnedBytes
+// when it is no longer needed otherwise memory will leak.
+func fromOwnedBytes(owned C.OwnedBytes) *RustOwnedBytes {
```
alarso16: mighty rusty, maybe call it `newOwnedBytes`

alarso16: Also, I think either everything or nothing should be exported. Since the FFI code is small (and we probably don't want to expose this to consumers), it shouldn't be exported

ffi/memory.go
demosdemon: That's not true. Go does allow you to call methods on the nil pointer. It is the responsibility of the receiver function to verify that the pointer is not nil.

https://go.dev/tour/methods/12

ffi/memory.go
alarso16: Yeah atomically swapping is probably the way to go (pun intended)

ffi/memory.go
demosdemon: I originally exported the `RustOwnedBytes` with the intent that it could be returned to callers instead of copying the memory into go. However, I went with copying for now because I didn't want to make interface breaking changes. It would be better in the long term to use `BorrowedBytes()` instead of `CopiedBytes()`.

E.g., @AminR443 should probably use `BorrowedBytes()` for the iterators to keep them zero copy.

ffi/memory.go
alarso16: Whoa, weird - I had no idea. It doesn't seem common to check this, but you can I guess... I rescind my comment

ffi/memory.go
```diff
@@ -103,6 +103,111 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
...
+	if owned.ptr == nil {
```
alarso16: I would personally vote that this never returns nil, even if owned.ptr is nil, because then you do not have to check that `b == nil` for each function. These checks are essentially the same, so just to reduce edge cases. If somebody creates the struct themselves, that's on them :)

ffi/memory.go
alarso16: Added a global comment - I think the check is unnecessary since we should enforce that it's not nil by the constructor, but it's nbd - we can keep

ffi/memory.go
demosdemon: On second thought, I don't think it is worth handling concurrent freeing here and it should probably be the responsibility of the thing that is freeing it. So, I will document and change accordingly.

Safely handling concurrent frees then exposes the question of what should we do about someone trying to call `intoError`, or `CopiedBytes` or `BorrowedBytes` concurrently as well. And I've resolved that this is a problem better solved by saying that while it is safe to call free multiple times (i.e., with the nil pointer) it is not safe to call it concurrently while the pointer is not nil and especially while there are outstanding `BorrowedBytes()` slices.

ffi/src/value/owned.rs
```diff
@@ -0,0 +1,161 @@
...
+///
+/// C callers must free this memory using the respective FFI function for the
+/// concrete type (but not using the `free` function from the C standard library).
+#[derive(Debug)]
+#[repr(C)]
+pub struct OwnedSlice<T> {
+    ptr: Option<NonNull<T>>,
+    len: usize,
+}
+
```
AminR443, line 19: What would be the disadvantage of having the same design on go side as well? I understand that we can't pass generics through FFI boundary but what I mean is having ownedSlice[T] on go side, with some constraints for T. I think this would provide a richer FFI base for future, and also could help reduce the boilerplate. For our iterator for example, I'll be passing `OwnedSlice<KeyValuePair>` back to go.

Most of my go experience predates the generics and I haven't had the chance to work with them, so I may be missing sth here, would be interesting if possible. I may try to experiment with it later.

ffi/memory.go
```diff
@@ -22,13 +22,46 @@ var (
...
+	//
+	// It is safe to call this method multiple times. Subsequent calls will
+	// do nothing if the data has already been freed (or was never set).
+	//
+	// However, it is not safe to call this method concurrently from multiple
+	// goroutines. It is also not safe to call this method while there are
+	// outstanding references to the slice returned by BorrowedBytes. Any
+	// existing slices will become invalid and may cause undefined behavior
+	// if used after the Free call.
+	Free() error
```
AminR443, line 60: Would it be possible to use something like [`runtime.AddCleanup`](https://pkg.go.dev/runtime#AddCleanup) for automatic cleanup when go's gc processes the struct holding the `C.Owned*` in case user hasn't called `Free()`?

ffi/src/value/owned.rs
demosdemon: generics don't work over the ffi barrier. we need concrete types on the `extern "C"` functions.

ffi/memory.go
demosdemon: cleanups are go 1.24. We can't use them here cause we're stuck on 1.23 right now. Finalizers are expensive a.f.

ffi/src/value/owned.rs
demosdemon: maybe I misunderstood your question. There's nothing stopping us from using `OwnedSlice<KeyValuePair>`. Go can safely cast that into a `[]C.KeyValuePair` slice. However, we cannot use generics on the ffi functions themselves. we need concrete types on the C layer and could implement this as a generic struct over `[]T` 
-- although that doesn't really work because a generic `T` needs to be a pointer in go. This means `ownedSlice[byte]` and `ownedSlice[C.KeyValuePair]` are invalid because the generic type is not a pointer type.

PR #1224, opened 2025-08-21
feat(ffi-result): Introduce VoidResult and panic handlers

demosdemon:


ffi/memory.go
```diff
@@ -103,6 +103,30 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
...
+		return nil
+	}
+
+	ptr := db.handle
+	if ptr == nil {
+		return nil
+	}
+
+	// Clear the handle before freeing to prevent double free
+	db.handle = nil
```
alarso16: This is racy if you expect multiple `Close` parallely. You could use `sync.Once` to ensure that this is only called once, or an atomic swap, or just say that this isn't parallel safe

ffi/memory.go
```diff
@@ -103,6 +103,30 @@ func newKeyValuePairs(keys, vals [][]byte, pinner Pinner) (C.BorrowedKeyValuePai
 	return newBorrowedKeyValuePairs(pairs, pinner), nil
 }
 
+// Close releases the memory associated with the Database.
+//
+// This is safe to call if the pointer is nil, in which case it does nothing. The
+// pointer will be set to nil after freeing to prevent double free.
+func (db *Database) Close() error {
+	if db == nil {
```
alarso16: I think it's reasonable to assume that it isn't nil, especially since it must be provided by our constructor

ffi/memory.go
```diff
@@ -208,6 +233,22 @@ func fromOwnedBytes(owned C.OwnedBytes) *RustOwnedBytes {
 	return rustBytes
 }
 
+// fromVoidResult converts a C.VoidResult to an error.
+//
+// It return nil if the result is Ok, otherwise it return an error.
+func fromVoidResult(result C.VoidResult) error {
```
alarso16: Very rust-y...
```suggestion
func getVoidResultError(result C.VoidResult) error {
```

ffi/memory.go
```diff
@@ -307,7 +356,9 @@ func bytesFromValue(v *C.struct_Value) ([]byte, error) {
 	// Case 4
 	if v.len != 0 && v.data != nil {
 		buf := C.GoBytes(unsafe.Pointer(v.data), C.int(v.len))
-		C.fwd_free_value(v)
+		if err := fromVoidResult(C.fwd_free_value(v)); err != nil {
+			return nil, fmt.Errorf("unexpected error while freeing value: %w", err)
```
alarso16: nit: you have this string a lot, you could put something like this. Whatever you like - both are ok, since you're not going to be testing this specifically
```suggestion
var unexpectedFreeError = errors.New("unexpected error while freeing value")
...
			return nil, fmt.Errorf("%w: %w", unexpectedFreeError, err)
```

alarso16 (approved):
Loose opinions, feel free to ignore

PR #1225, opened 2025-08-21
feat(ffi-refactor): Refactor Db opening to use new Result structure (4/8)

demosdemon:
Depends On: #1224


ffi/firewood.go
```diff
@@ -74,16 +74,14 @@ func DefaultConfig() *Config {
...
-type CacheStrategy uint8
+type CacheStrategy C.CacheReadStrategy
 
 const (
-	OnlyCacheWrites CacheStrategy = iota
-	CacheBranchReads
-	CacheAllReads
+	OnlyCacheWrites  CacheStrategy = C.CacheReadStrategy_WritesOnly
+	CacheBranchReads CacheStrategy = C.CacheReadStrategy_BranchReads
+	CacheAllReads    CacheStrategy = C.CacheReadStrategy_All
```
rkuris: nit: Seems strange that we have two completely different naming schemes here. Why can't we use `CacheWritesOnly` `CacheBranchReads` and `CacheAll` to line up with the C defs?

ffi/firewood.go
```diff
@@ -74,16 +74,14 @@ func DefaultConfig() *Config {
...
 
-	// invalidCacheStrategy MUST be the final value in the iota block to make it
-	// the smallest value greater than all valid values.
-	invalidCacheStrategy
+	invalidCacheStrategy = CacheAllReads + 1
```
rkuris: This seems risky, as it assumes `C.CacheReadStrategy_All` is greater than all the others. Can we use math.MaxInt64 or maybe -1?

ffi/src/handle.rs
```diff
@@ -1,12 +1,155 @@
...
+    /// Cache both write operations and branch node reads. This provides
+    /// better performance for tree traversal operations while keeping
+    /// memory usage moderate.
+    BranchReads = 1,
+
+    /// Cache all read and write operations. This provides the best performance
+    /// but uses the most memory as it caches leaf nodes and values in addition
+    /// to branch nodes.
+    All = 2,
+}
```
rkuris: Here might be another good place to put the Invalid variant.

ffi/src/handle.rs
```diff
@@ -1,12 +1,155 @@
...
+    /// Whether to truncate the database file if it exists.
+    pub truncate: bool,
+}
+
+impl DatabaseHandleArgs<'_> {
+    const fn cache_size(&self) -> NonZeroUsize {
+        match NonZeroUsize::new(self.cache_size) {
+            Some(size) => size,
+            None => DEFAULT_CACHE_SIZE,
+        }
```
rkuris:
```suggestion
        NonZeroUsize::new(self.cache_size).unwrap_or(DEFAULT_CACHE_SIZE)
```

ffi/src/handle.rs
```diff
@@ -1,12 +1,155 @@
...
+    }
+
+    const fn free_list_cache_size(&self) -> NonZeroUsize {
+        match NonZeroUsize::new(self.free_list_cache_size) {
+            Some(size) => size,
+            None => DEFAULT_FREE_LIST_CACHE_SIZE,
+        }
```
rkuris: unwrap_or is cleaner IMO

ffi/src/handle.rs
```diff
@@ -1,12 +1,155 @@
...
+    }
+
+    const fn revisions(&self) -> usize {
+        if self.revisions == 0 {
+            DEFAULT_REVISIONS
+        } else {
+            // if 1, building the revision manager will fail, but that's on the caller
```
rkuris: Is it? Then why are we handling 0?

Maybe we just change the earlier if to `if self.revisions < 2`

rkuris (approved):
A few comments, otherwise LGTM

ffi/firewood.go
demosdemon: One is a go enum and the other is the C enum. We could expose the `C.CacheReadStrategy` directly but that's not how things are typically done with go.

ffi/src/handle.rs
demosdemon: Adding the invalid variant here is not appropriate. That mixes rust and go. Now rust has to handle the invalid variant case when it was previously not necessary.

ffi/src/handle.rs
demosdemon: `unwrap_or` is not yet const safe. the match is required for const contexts.

ffi/src/handle.rs
demosdemon: `unwrap_or` is not yet const safe. the match is required for const contexts.

ffi/src/handle.rs
demosdemon: we are handling 0 because the "default" value for everything is zero in go. So, zero wrt to go is also unset. passing in a `1` is explicitly wrong so that should be an error.

ffi/firewood.go
demosdemon: This behavior is exactly the same as before which also assumed this. We want it to be more than the good value because of this check:

```go
	if conf.ReadCacheStrategy >= invalidCacheStrategy {
		return nil, fmt.Errorf("invalid %T (%[1]d)", conf.ReadCacheStrategy)
	}
```

These are unsigned integers, so the only valid range is 0..=255.

PR #1226, opened 2025-08-21
feat(ffi-refactor): refactor how hash values are returned (5/8)

demosdemon:
This way we avoid allocating just to return the hash value.

A 32-byte hash value is the same size as 4 words and the allocated buffer would be returning 2 words. If we skip the heap allocation, we can return all 4 words on the stack directly and avoid the heap allocation (on the rust side). The Go side later allocates for the slice, but this could be removed later if the interface was build around a fixed-size 32 byte array.

Depends On: #1225

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -20,9 +20,9 @@
 typedef struct DatabaseHandle DatabaseHandle;
 
 /**
- * A root database hash key, used in FFI functions that provide or return
- * a root hash. This type requires no allocation and can be copied freely
- * and droppped without any additional overhead.
+ * A database hash key, used in FFI functions that require hashes.
+ * This type requires no allocation and can be copied freely and
+ * dropped without any additional overhead.
  *
  * This is useful because it is the same size as 4 words which is equivalent
  * to 2 heap-allocated slices (pointer + length each), or 1.5 vectors (which
--- a/ffi/src/value/hash_key.rs
+++ b/ffi/src/value/hash_key.rs
@@ -3,9 +3,9 @@
 
 use std::fmt;
 
-/// A root database hash key, used in FFI functions that provide or return
-/// a root hash. This type requires no allocation and can be copied freely
-/// and droppped without any additional overhead.
+/// A database hash key, used in FFI functions that require hashes.
+/// This type requires no allocation and can be copied freely and
+/// dropped without any additional overhead.
 ///
 /// This is useful because it is the same size as 4 words which is equivalent
 /// to 2 heap-allocated slices (pointer + length each), or 1.5 vectors (which


ffi/firewood.h
```diff
@@ -20,24 +20,78 @@
...
- * * A byte slice containing data.
- *
- * For more details on how the data may be stored, refer to the function signature
- * that returned it or the `From` implementations.
- *
- * The data stored in this struct (if `data` is not null) must be manually freed
- * by the caller using `fwd_free_value`.
+ * A root database hash key, used in FFI functions that provide or return
+ * a root hash. This type requires no allocation and can be copied freely
+ * and droppped without any additional overhead.
```
rkuris: Arguably it's any hash key, although currently in the FFI we only use it as the root database hash key. That could change, and this comment could go stale.

Also a "ppp" typo in dropped.
```suggestion
 * A database hash key, used in FFI functions that require hashes.
 * This type requires no allocation and can be copied freely and
 * dropped without any additional overhead.
```

ffi/firewood.h
```diff
@@ -20,24 +20,78 @@
...
+
+/**
+ * A result type returned from FFI functions return the database root hash. This
+ * may or may not be after a mutation.
+ */
+typedef enum HashResult_Tag {
+  /**
+   * The caller provided a null pointer to a database handle.
+   */
+  HashResult_NullHandlePointer,
```
rkuris, line 60: This seems pretty uncommon. Why can't we just make this another HashResult_Err?

ffi/firewood.h
```diff
@@ -20,24 +20,78 @@
...
+/**
+ * A type alias for a rust-owned byte slice.
+ */
+typedef struct OwnedSlice_u8 OwnedBytes;
+
+/**
+ * A result type returned from FFI functions return the database root hash. This
+ * may or may not be after a mutation.
+ */
+typedef enum HashResult_Tag {
```
rkuris, line 56: This is likely to be a full 8 bytes. I'm not sure what can be done to change that, but it seems quite wasteful.

In straight C, you'd have to add a `__attribute__` to this.

Can this be done differently so that it's always just a byte?

rkuris (changes requested):
Concerned about discriminant size.

ffi/firewood.h
demosdemon: Ideally all errors would eventually propagate as an enum instead of allocating a string, but I started with the NPE because it's the most common variant. The generated asm code surrounding handling the NPE also be more efficient as it will go from zero input to zero output.

ffi/firewood.h
demosdemon: where does your concern come from? the enum discriminants will never make it onto the heap. and in terms of a 64-bit system, 8 bytes is one word. operating on complete words will be more efficient than anything smaller (unless auto vectorization kicks in somehow).

afaik, this is not controllable as it's baked into cbindgen but I can do some research.

rkuris:
> Concerned about discriminant size.

While this matters, it (a) probably doesn't matter much, and (b) forces some alignment issues we probably don't want.

ffi/firewood.h
rkuris: Don't worry about it, I think it's okay based on arm's strange alignment issues. This might matter if size mattered, but I think it doesn't. We aren't likely to pass large arrays of these across the ffi boundary.

rkuris (approved):

Issue #1008, opened 2025-07-01
Require the FFI to explicitly track revisions


alarso16:
Rather unintuitively, in concurrent contexts, there is no way to guarantee that revision won't be removed from the revision manager during use, since Rust cannot accurately count the number of references in use. This is currently handled relatively gracefully by returning an error upon the next use, but it would be ideal if the FFI supported holding a revision in use, even if it could be evicted due to age.

demosdemon:
@alarso16, am I doing what you asked here in #1158?

demosdemon:
Ah, after rethinking, the answer is sort of but no. But, I have set up the framework for doing so. Instead of returning a pointer to a proposal, we can return a pointer to a nodestore.

PR #1227, opened 2025-08-21
feat(ffi-refactor): refactor revision to use database handle

demosdemon:
Depends On: #1226

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -224,5 +224,17 @@ func (db *Database) Root() ([]byte, error) {
 
 // Revision returns a historical revision of the database.
 func (db *Database) Revision(root []byte) (*Revision, error) {
-	return newRevision(db.handle, root)
+	if root == nil || len(root) != RootLength {
+		return nil, errInvalidRootLength
+	}
+
+	// Attempt to get any value from the root.
+	// This will verify that the root is valid and accessible.
+	// If the root is not valid, this will return an error.
+	_, err := db.GetFromRoot(root, []byte{})
+	if err != nil {
+		return nil, err
+	}
+
+	return &Revision{database: db, root: root}, nil
 }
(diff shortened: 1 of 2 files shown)


ffi/firewood.go
```diff
@@ -222,5 +222,17 @@ func (db *Database) Root() ([]byte, error) {
...
+	if root == nil || len(root) != RootLength {
+		return nil, errInvalidRootLength
+	}
+
+	// Attempt to get any value from the root.
+	// This will verify that the root is valid and accessible.
+	// If the root is not valid, this will return an error.
+	_, err := db.GetFromRoot(root, []byte{})
+	if err != nil {
+		return nil, err
```
rkuris, line 234: pre-existing bug, but this is racy. The revision might expire right after this check.

ffi/firewood_test.go
```diff
@@ -839,7 +839,7 @@ func TestInvalidRevision(t *testing.T) {
 	validRoot := []byte("counting 32 bytes to make a hash")
 	r.Len(validRoot, 32, "valid root")
 	_, err = db.Revision(validRoot)
-	r.ErrorIs(err, errRevisionNotFound, "Revision(valid root)")
+	r.ErrorContains(err, "Revision for Some(636f756e74696e6720333220627974657320746f206d616b6520612068617368) not found", "Revision(valid root)")
```
demosdemon, line 842: this is a temporary change that is reverted in the next PR

rkuris: IMO the best way to handle this is add a TODO comment, just in case we abandon the next PR or redo it. You remove the TODO in the next PR along with fixing it.

rkuris (approved):

ffi/firewood.go
demosdemon: Yeah, I realized that earlier. I plan to fix that when I get around to #1008

ffi/firewood_test.go
alarso16: Our linter (lazily) attempts to reject this in the first place. As long as it's removed later

alarso16 (approved):

PR #1228, opened 2025-08-21
feat(ffi-refactor): Add `ValueResult` type (7/8)

demosdemon:
This is the penultimate pull-request in the series to refactor the Firewood FFI api. This adds a structures `ValueResult` type for functions that return a byte array back to the caller.

Depends On: #1227


ffi/firewood.go
```diff
@@ -168,15 +166,12 @@ func (db *Database) Get(key []byte) ([]byte, error) {
 	var pinner runtime.Pinner
 	defer pinner.Unpin()
 
-	val := C.fwd_get_latest(db.handle, newBorrowedBytes(key, &pinner))
-	bytes, err := bytesFromValue(&val)
-
-	// If the root hash is not found, return nil.
-	if err != nil && strings.Contains(err.Error(), rootHashNotFound) {
+	val, err := getValueFromValueResult(C.fwd_get_latest(db.handle, newBorrowedBytes(key, &pinner)))
+	if errors.Is(err, errRevisionNotFound) {
```
alarso16, line 172: Maybe add an explanation (better than what I wrote) explaining why this would happen. Explicitly something similar to this:
```suggestion
    // The revision won't be found if the database is empty.
    // This is valid, but should be treated as a non-existent key
	if errors.Is(err, errRevisionNotFound) {
```

alarso16 (approved):

PR #1233, opened 2025-08-22
perf: optimize hashing by avoiding unnecessary clone operations

demosdemon:
I did not notice this the first time when I touched this, but, we were unecessarily cloning the data to be hashed. This change avoid that.


demosdemon:
closing cause I misread. we're only cloning the hash not the entire node. I do like the `map` syntax on the array so I'll probably come back to clean this up later.

PR #1239, opened 2025-08-26
chore: rename FuzzTree

rkuris:
Confusing since there is also one in coreth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,28 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.0.12] - 2025-08-26
+
+### 🚀 Features
+
+- *(async-removal)* Phase 3 - make `Db` trait sync ([#1213](https://github.com/ava-labs/firewood/pull/1213))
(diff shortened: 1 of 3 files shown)


demosdemon (approved):

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -295,7 +295,7 @@ func (tr *merkleTriePair) deleteStorage(accountIndex int, storageIndexInput uint
 	tr.pendingFwdVals = append(tr.pendingFwdVals, []byte{})
 }
 
-func FuzzTree(f *testing.F) {
+func FuzzFirewoodTree(f *testing.F) {
```
StephenButtolph, line 298: Not to be pedantic, but seeing as we are in the firewood repo, when would we ever be fuzzing a non-firewood tree?

alarso16: Maybe `FuzzEthHash`? The coreth one actually tests the trie implementation

PR #1237, opened 2025-08-25
feat(checker): print report using template for better readability

qusuyan:
```
Errors (1):
	Found leaked areas: Linear Address Range Set:
	Address Range: [0x800, 0x900)
	Address Range: [0xa00, 0xb00)
	Address Range: [0xb60, 0xc60)
	Address Range: [0xdc0, 0xec0)
	Address Range: [0x1120, 0x12a0)
	Address Range: [0x13a0, 0x1620)
	Address Range: [0x1720, 0x17a0)
	Address Range: [0x1aa0, 0x20a0)
	Address Range: [0x27a0, 0x28a0)
	Address Range: [0x29a0, 0x2ea0)
	... (76480 more hidden)


Basic Stats:
    Firewood Image Size / High Watermark (high_watermark): 394,533,632
    Total Key-Value Count (kv_count): 586,687
    Total Key-Value Bytes (kv_bytes): 46,859,598

Trie Stats:
    Branching Factor Distribution: 1: 98, 2: 114,082, 3: 33,897, 4: 12,560, 5: 6,810, 6: 4,994, 7: 4,381, 8: 4,359, 9: 4,311, 10: 3,917, 11: 3,779, 12: 3,162, 13: 2,448, 14: 1,516, 15: 941, 16: 3,452
    Depth Distribution: 4: 23,285, 5: 36,333, 6: 16,921, 7: 42,667, 8: 119,550, 9: 197,709, 10: 139,367, 11: 9,045, 12: 40

Branch Area Stats:
    Total Branch Data Bytes (branch_bytes): 33,759,976
    Total Branch Area Count (branch_area_count): 204,707
    Total Branch Area Bytes (branch_area_bytes): 97,750,016
    Branch Area Distribution: 16: 0, 32: 0, 64: 0, 96: 43,100, 128: 16,525, 256: 24,453, 512: 37,967, 768: 73,624, 1,024: 9,038, 2,048: 0, 4,096: 0, 8,192: 0, 16,384: 0, 32,768: 0, 65,536: 0, 131,072: 0, 262,144: 0, 524,288: 0, 1,048,576: 0, 2,097,152: 0, 4,194,304: 0, 8,388,608: 0, 16,777,216: 0
    Branches that Can Fit Into Smaller Area (low_occupancy_branch_area): 118,365 (57.82%)

Leaf Area Stats:
    Total Leaf Data Bytes (leaf_bytes): 47,874,086
    Total Leaf Area Count (leaf_area_count): 584,917
[...]

PR #1240, opened 2025-08-26
feat(checker): fix free lists when the erroneous area is the head

qusuyan:
Here instead of fixing things in place, we create a new revision. This can also benefit future fixes for the trie.


qusuyan:
I made a mistake and this PR branches from another PR. Waiting for #1237 before merging into main.

storage/src/nodestore/mod.rs
```diff
@@ -832,6 +839,15 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
 
+// Helper functions for the fixer
+impl<T: Parentable, S: WritableStorage> NodeStore<T, S> {
+    pub(crate) fn new_proposal(&self) -> NodeStore<MutableProposal, S> {
+        NodeStore::<MutableProposal, S>::new(self).unwrap_or_else(|_| {
+            NodeStore::<MutableProposal, S>::new_empty_proposal(self.storage.clone())
+        })
+    }
+}
+
```
demosdemon: Because this is only used by the checker, this impl block should be moved to the `checker/mod.rs` which will allow you to remove the visibility modifier. This method specifically feels dirty to expose to anyone except the checker.

storage/src/nodestore/mod.rs
```diff
@@ -812,7 +816,10 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
 }
 
 // Helper functions for the checker
-impl<S: ReadableStorage> NodeStore<Committed, S> {
+impl<T, S: ReadableStorage> NodeStore<T, S>
```
demosdemon, line 820: The same could be done for this impl block.

storage/src/nodestore/mod.rs
qusuyan: The reason I put it here is because I need visibility to `self.storage`, which I don't believe I have access to from `checker/mod.rs`. Maybe we should move the entire `checker` to inside the `nodestore` mod?

demosdemon: looks like `new_empty_proposal` should be decorated with `#[cfg(any(test, feature = "test_utils"))]` because 

- it is currently only used in tests. 
- it panics because it attempts to overwrite the nodestore header and doesn't handle the potential error 
- it overwrites the nodestore header, this is probably not safe for you to do: 

https://github.com/ava-labs/firewood/blob/4fda6b12a396a2ba230aab9e7064f15308d42994/storage/src/nodestore/mod.rs#L280-L285

Let's take a look at why `NodeStore::new` can fail:

https://github.com/ava-labs/firewood/blob/4fda6b12a396a2ba230aab9e7064f15308d42994/storage/src/nodestore/mod.rs#L231-L231

It will fail if it cannot read the root node. I don't see any other failure case which means you probably don't need to do anything _but_ call `new` and let the i/o error propagate.

storage/src/nodestore/mod.rs
qusuyan: Same problem here - I do not have access to `header` or `storage` from `checker/mod.rs`, which are needed for checker. Is there a better walkaround?

demosdemon (approved):

PR #1242, opened 2025-08-26
chore: upgrade to rust 1.89

rkuris:
Needed to get File::lock in an upcoming PR

### DIFF
--- a/.github/check-license-headers.yaml
+++ b/.github/check-license-headers.yaml
@@ -27,6 +27,7 @@
       "**/go.mod",
       "**/go.mod",
       "**/cbindgen.toml",
+      "rust-toolchain.toml",
     ],
     "license": "./.github/license-header.txt"
   },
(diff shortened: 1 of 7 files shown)


demosdemon:
let's also create a [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) file: 

```toml
[toolchain]
channel = "1.89"
```

This will cause cargo to always use whatever is defined there unless overridden with `rustup override` or `RUSTUP_TOOLCHAIN` environment variable. This will also automatically download 1.89 if it's not installed, which is typically a good thing because not everyone runs `rustup update` religiously.

We could also use this to pin to a specific nightly version if we ever needed to. 

It's a touch annoying that it's defined in so many places but I'd rather hook into the tools wherever relevant.

demosdemon:
nvm about `rust-toolchain.toml`, good experiment though. setting rust-toolchain.toml separate from the ci action causes the cache to act weird. So, this would be a bigger change than I'm ok with just tacking on like that.

PR #1244, opened 2025-08-27
fix: Add an advisory lock

rkuris:
Use `File::try_lock` and produce an error if it's already opened by someone else.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -268,7 +268,7 @@ impl<'db> api::Proposal for Proposal<'db> {
     }
 
     fn commit(self) -> Result<(), api::Error> {
-        Ok(self.db.manager.commit(self.nodestore.clone())?)
+        Ok(self.db.manager.commit(self.nodestore)?)
     }
 }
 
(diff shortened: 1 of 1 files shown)


storage/src/linear/filebacked.rs
```diff
@@ -64,6 +64,19 @@ impl std::fmt::Debug for FileBacked {
 }
 
 impl FileBacked {
+    /// Acquire an advisory lock on the underlying file to prevent multiple processes
+    /// from accessing it simultaneously
+    pub fn lock(&self) -> Result<(), FileIoError> {
+        self.fd.try_lock().map_err(|e| {
```
demosdemon, line 70: is `e` not already an `std::io::Error`?

rkuris: Nope. It's one of these: https://doc.rust-lang.org/beta/std/fs/enum.TryLockError.html

demosdemon: yeah, [it is](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.lock) so wrapping it in another io error is unnecessary.

demosdemon: OH, this is try_lock and not lock. sorry.

demosdemon:
oh heh. lock is causing the reopen test to fail because we keep a revision around which keeps the file open.

https://github.com/ava-labs/firewood/blob/f11ef562d479e646ea996f1043ba22b2e020ca27/firewood/src/db.rs#L408

can drop the rev before the replace call

PR #1253, opened 2025-09-03
feat(ffi-proofs): Stub interfaces for FFI to interact with proofs.

demosdemon:
This contains a stubbed api for the FFI library to build range and change proofs as well as verify and commit. No methods in the FFI library itself are implemented except for the free handlers.

### DIFF
--- a/ffi/maybe.go
+++ b/ffi/maybe.go
@@ -12,6 +12,14 @@ import (
 )
 
 // Maybe is an interface that represents an optional value of type T.
+//
+// Maybe is a drop-in replacement for the Maybe type defined in avalanchego/utils/maybe.
+// This interface is used to avoid importing avalanchego packages into the ffi
+// package, which would create a circular dependency.
+//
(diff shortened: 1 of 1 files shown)


demosdemon:
> do we need `maybe`?

The empty key is technically valid so I didn't want to restrict its use. The interface is compatible with avalanchego's maybe and should work as-is: https://github.com/ava-labs/avalanchego/blob/66ca7dc0ce354ff4c4ebdc8b377e099ba91f2e2b/utils/maybe/maybe.go#L40-L48

ffi/maybe.go
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+package ffi
```
rkuris, line 4: Let's add a reference to the avalanchego implementation.

While I'd love to use that, we can't create.a circular reference. Perhaps we can create an issue in avalanchego to move these into a shareable subproject, and we can depend on that.

rkuris (approved):

alarso16 (approved):

PR #1255, opened 2025-09-03
feat(ffi-iterator): Implementation of Iterator in Rust (1/4)

AminR443:
This PR is a step towards implementing the FFI iterator mentioned in #1009. 

This implement FFI functions (`fwd_iter_on_root`, `fwd_iter_on_proposal`, `fwd_iter_next`)  in Rust, and also necessary changes in `firewood` crate for `MerkleKeyValueStream` to work with `Arc<NodeStore<..>>` as well.


ffi/src/handle.rs
```diff
@@ -176,6 +178,24 @@ impl DatabaseHandle {
         Ok(root_hash)
     }
 
+    /// Creates an iterator on the revision with specified root starting from the given key.
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if the iterator could not be created.
```
rkuris: Tautological comment, please remove, or enumerate the errors.

ffi/src/handle.rs
```diff
@@ -176,6 +178,24 @@ impl DatabaseHandle {
...
+    pub fn iter_on_root(
```
rkuris: This seems a little strange. Why aren't we passing in the revision object directly, and instead creating a new one?

ffi/src/handle.rs
```diff
@@ -176,6 +178,24 @@ impl DatabaseHandle {
...
+        &self,
+        root: HashKey,
+        first_key: Option<&[u8]>,
+    ) -> Result<CreateIteratorResult<'_>, api::Error> {
+        let rev = self.db.revision(root)?;
```
rkuris: This will only work for historical revisions, not proposals. This should probably take a `ArcDynDbView`.

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
...
+
+// This wrapper doesn't do anything special now, but we need to figure out how to handle
+// proposal commits, drops, revision cleaning
+// TODO(amin): figure this out
+
+type KeyValueItem = Result<(merkle::Key, merkle::Value), api::Error>;
+
+/// An opaque wrapper around an Iterator.
+pub struct IteratorHandle<'db> {
+    pub iterator: Box<dyn Iterator<Item = KeyValueItem> + 'db>,
```
rkuris: We can get better performance here if we use a concrete type instead of a boxed dyn (probably MerkleKeyValueIter), which is evaluated at runtime, but that might be hard to do if you care about supporting proposals and historical revisions.

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl Debug for IteratorHandle<'_> {
```
rkuris: Seems like we should be able to derive this using the derive-where crate, or skip derive debug altogether. I don't think it adds much to implement Debug when all it prints is the name of the structure.
```
#[derive_where(Debug)]
#[derive_where(skip_inner(Debug))]
```

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("IteratorHandle").finish()
+    }
+}
+
+impl IteratorHandle<'_> {
+    pub fn iter_next(&mut self) -> Option<KeyValueItem> {
```
rkuris: This seems to stutter a little. Why not just `next`? Why not just implement Iterator here:

```
impl Iterator for IteratorHandle<'_> {
    type Item = KeyValueItem;
    fn next(&mut self) -> Option<KeyValueItem> {
        self.iterator.next()
    }
}
```

ffi/src/iterator.rs
rkuris: Why is `iterator` pub? I don't think anyone should be accessing it outside of this file.

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
...
+/// - [`IteratorResult::NullHandlePointer`] if the provided database handle is null.
+/// - [`IteratorResult::Ok`] if the iterator was created, with the iterator handle.
+/// - [`IteratorResult::Err`] if an error occurred while creating the iterator.
+///
+/// # Safety
+///
+/// The caller must:
+/// * ensure that `db` is a valid pointer to a [`DatabaseHandle`]
+/// * ensure that `root` is a valid for [`BorrowedBytes`]
+/// * ensure that `key` is a valid for [`BorrowedBytes`]
```
rkuris: "is a valid for" -> "is a valid pointer to"

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
...
+///
+/// * `db` - The database handle returned by [`fwd_open_db`]
+/// * `root` - The root hash to look up as a [`BorrowedBytes`]
+/// * `key` - The key to look up as a [`BorrowedBytes`]
+///
+/// # Returns
+///
+/// - [`IteratorResult::NullHandlePointer`] if the provided database handle is null.
+/// - [`IteratorResult::Ok`] if the iterator was created, with the iterator handle.
+/// - [`IteratorResult::Err`] if an error occurred while creating the iterator.
```
rkuris: These add very little value, consider removing them or improving them.

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
...
+/// * ensure that `key` is a valid for [`BorrowedBytes`]
+/// * call [`fwd_free_iterator`] to free the memory associated with the iterator.
+///
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn fwd_iter_on_root<'db>(
+    db: Option<&'db DatabaseHandle>,
+    root: BorrowedBytes,
+    key: BorrowedBytes,
+) -> IteratorResult<'db> {
+    invoke_with_handle(db, move |db| {
```
rkuris: [readability] insert_with_handle needs a different name, or just use map_or:
```
    db.map_or(ValueResult::NullHandlePointer, |db| db.get_latest(key).into())
```

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
...
+/// # Returns
+///
+/// - [`IteratorResult::NullHandlePointer`] if the provided database handle is null.
+/// - [`IteratorResult::Ok`] if the iterator was created, with the iterator handle.
+/// - [`IteratorResult::Err`] if an error occurred while creating the iterator.
+///
+/// # Safety
+///
+/// The caller must:
+/// * ensure that `db` is a valid pointer to a [`DatabaseHandle`]
```
rkuris: Option::None is also accepted here, and returns an error.

ffi/src/value/results.rs
```diff
@@ -195,6 +197,76 @@ pub enum ProposalResult<'db> {
...
+#[repr(C)]
+pub enum KeyValueResult {
+    /// The caller provided a null pointer to an iterator handle.
+    NullHandlePointer,
+    /// The iterator returned empty result, the iterator is exhausted
+    None,
+    /// The next item on iterator is returned.
+    Some(OwnedKeyValuePair),
+    /// An error occurred and the message is returned as an [`OwnedBytes`]. If
+    /// value is guaranteed to contain only valid UTF-8.
```
rkuris: I don't understand this. If the word "if" at the beginning is extra, then I disagree with this assertion. If this was about to say what happens in that case, then please complete the sentence.

ffi/src/value/results.rs
```diff
@@ -195,6 +197,76 @@ pub enum ProposalResult<'db> {
...
+    Err(OwnedBytes),
+}
+
+/// A result type returned from iterator FFI functions
+#[derive(Debug)]
+#[repr(C)]
+pub enum KeyValueResult {
+    /// The caller provided a null pointer to an iterator handle.
+    NullHandlePointer,
+    /// The iterator returned empty result, the iterator is exhausted
```
rkuris:
```suggestion
    /// The iterator is exhausted
```

ffi/src/value/results.rs
```diff
@@ -195,6 +197,76 @@ pub enum ProposalResult<'db> {
...
+    NullHandlePointer,
+    /// Building the proposal was successful and the proposal ID and root hash
+    /// are returned.
+    Ok {
+        /// An opaque pointer to the [`ProposalHandle`] that can be use to create
+        /// an additional proposal or later commit. The caller must ensure that this
+        /// pointer is freed with [`fwd_free_proposal`] if it is not committed.
+        ///
+        /// [`fwd_free_proposal`]: crate::fwd_free_proposal
+        // note: opaque pointers mut be boxed because the FFI does not the structure definition.
```
rkuris:
```suggestion
        /// An opaque pointer to a [`ProposalHandle`].
        /// The value should be freed either with [`fwd_free_proposal`]
        /// or [`fwd_commit_proposal`]
```

rkuris (commented):
Partial review, roughly 60% through before my time limit. Please try to keep your PR requests shorter for a better response.

Overall, this looks good but the comments need some cleanup -- they are often either tautological or not generally useful.

Think about if you're using this API what you'd care about, and make sure to document only that. For example, I usually don't care that something is an opaque pointer to something else that I don't need to know about.

ffi/firewood.h
```diff
@@ -269,6 +274,100 @@ typedef struct ValueResult {
...
+  /**
+   * The iterator returned empty result, the iterator is exhausted
+   */
+  KeyValueResult_None,
+  /**
+   * The next item on iterator is returned.
+   */
+  KeyValueResult_Some,
+  /**
+   * An error occurred and the message is returned as an [`OwnedBytes`]. If
```
bernard-avalabs:
```suggestion
   * An error occurred and the message is returned as an [`OwnedBytes`]. The
```

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
     invoke_with_handle(db, move |db| db.get_latest(key))
 }
 
+/// Return an iterator optionally starting from a key in database
```
bernard-avalabs:
```suggestion
/// Return an iterator optionally starting from a key in the provided database
```

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
...
+    db: Option<&'db DatabaseHandle>,
+    root: BorrowedBytes,
+    key: BorrowedBytes,
+) -> IteratorResult<'db> {
+    invoke_with_handle(db, move |db| {
+        db.iter_on_root(root.as_ref().try_into()?, Some(key.as_slice()))
+    })
+}
+
+/// Return an iterator on proposal optionally starting from a key
```
bernard-avalabs:
```suggestion
/// Return an iterator on the provided proposal optionally starting from a key
```

ffi/src/lib.rs
```diff
@@ -105,6 +107,120 @@ pub unsafe extern "C" fn fwd_get_latest(
...
+///
+/// - [`IteratorResult::NullHandlePointer`] if the provided proposal handle is null.
+/// - [`IteratorResult::Ok`] if the iterator was created, with the iterator handle.
+/// - [`IteratorResult::Err`] if an error occurred while creating the iterator.
+///
+/// # Safety
+///
+/// The caller must:
+/// * ensure that `handle` is a valid pointer to a [`ProposalHandle`]
+/// * ensure that `key` is a valid for [`BorrowedBytes`]
```
bernard-avalabs, line 160:
```suggestion
/// * ensure that `key` is a valid pointer to [`BorrowedBytes`]
```

ffi/src/iterator.rs
AminR443: I avoided the boxed dyn before by introducing a new enum:
```rust
enum MerkleIterator<'v> {
    View(MerkleKeyValueStream<'v, NodeStore<Committed, FileBacked>>),
    Proposal(MerkleKeyValueStream<'v, NodeStore<Arc<ImmutableProposal>, FileBacked>>),
}
```
Removed it because of Suyan's comment. If you think it's worth the performance change, I'll update. Actually, I can check the performance impact in the benchmarks and then we can discuss.

ffi/src/iterator.rs
AminR443: The iterator handle used to live in a hashmap in database handle which required Debug. I'll just remove this as it's not the case anymore, it's useless.

ffi/src/value/results.rs
AminR443: I don't understand it either. I copied these structs from @demosdemon's code and made changes, forgot to go over it after.

I'm thinking that the "If" is extra, maybe should be "The". Don't we construct these errors from `String` always, which is UTF-8 by default? 

I'll update this, but there are 4-5 other places, not relevant to this PR, that need to be updated too.

ffi/src/lib.rs
AminR443: I don't understand your comment here and the next one as well. Could you explain more please?

All the other FFI functions are following this pattern.

ffi/src/iterator.rs
AminR443: Noticed that it was needed somewhere else. added derive-where.

ffi/src/lib.rs
```diff
@@ -326,6 +436,7 @@ pub unsafe extern "C" fn fwd_commit_proposal(
 pub unsafe extern "C" fn fwd_free_proposal(
     proposal: Option<Box<ProposalHandle<'_>>>,
 ) -> VoidResult {
+    // TODO(amin): Handle iterators?
```
bernard-avalabs: Will we leak memory if there are outstanding iterators, or will they be dropped automatically? Might be worth documenting.

ffi/src/proposal.rs
```diff
@@ -97,8 +98,20 @@ impl ProposalHandle<'_> {
 
         Ok(hash_key)
     }
-}
 
+    /// Creates an iterator on the proposal starting from the given key.
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if the iterator could not be created.
```
bernard-avalabs: Perhaps expand by describing possible cases where an error may occur when calling iter_owned?

ffi/src/value/results.rs
```diff
@@ -195,6 +197,73 @@ pub enum ProposalResult<'db> {
...
+
+/// A result type returned from iterator FFI functions
+#[derive(Debug)]
+#[repr(C)]
+pub enum KeyValueResult {
+    /// The caller provided a null pointer to an iterator handle.
+    NullHandlePointer,
+    /// The iterator is exhausted
+    None,
+    /// The next item on iterator is returned.
```
bernard-avalabs:
```suggestion
    /// The next item is returned.
```

ffi/src/value/results.rs
```diff
@@ -195,6 +197,73 @@ pub enum ProposalResult<'db> {
...
+    Some(OwnedKeyValuePair),
```
bernard-avalabs, line 347: Is the owner responsible for freeing this?

firewood/src/iter.rs
```diff
@@ -62,11 +64,45 @@ enum NodeIterState {
     },
 }
 
+/// A reference to the merkle, could be borrowed, or behind an Arc
+#[derive(Debug)]
+pub enum MerkleRef<'a, T> {
+    /// reference with borrowing the merkle
```
bernard-avalabs:
```suggestion
    /// borrowing the merkle
```

firewood/src/v2/api.rs
```diff
@@ -369,6 +369,22 @@ pub trait DynDbView: Debug + Send + Sync + 'static {
     }
 }
 
+/// A view of database with possibility of getting an owned iterator from.
```
bernard-avalabs: Not sure I completely understand this line. Do you mean that this is a view of the database, and it is possible to get an owned iterator from this view?

ffi/src/value/results.rs
AminR443: This is rust-owned and go will be responsible to call the associated free function to drop it. I'll add documentation.

ffi/src/lib.rs
AminR443: Documented this in a more clear way.

firewood/src/v2/api.rs
AminR443: Yes. Fixed.

AminR443:
@rkuris @bernard-avalabs Thank you for your feedbacks. 

@rkuris I resolved almost all of your comments. A few remaining I need more explanation. In the most recent commit I made the code works for proposed roots as well, but I had to make changes to RevisionManager; please review that and let me know if you need it to be handled in a different way. The problem is that we can't simply use `ArcDynDbView` here because the iterator created by its methods will depend on it's lifetime and we can't pass that to FFI safely.

@bernard-avalabs I fixed the documentation issues and added some more documentation based on your questions.

ffi/src/handle.rs
rkuris: To be clearer here, it seems like we can't create an iterator given an arbitrary revision unless we have the hash of that revision and create a new one. I'm not sure why this is.

AminR443: I'm not sure if I'm following. Could you explain more please? which arbitrary revision are you referring to, so we could create an iterator on that?

rkuris: This function requires a HashKey. There is no function that requires a view and returns an iterator. instead, this function is doing both -- find the view from the revision and then create the iterator.

This would be the only place in the code that if you have a NodeStore you can't operate on that, unless you get the hash from that nodestore and create another nodestore. This also means it's impossible to iterate on Mutable nodestores, which we might actually need in the near future for applying range proofs.

more reading: https://en.wikipedia.org/wiki/Single-responsibility_principle

ffi/src/lib.rs
rkuris: I think you can ignore this comment. I don't like "intert_with_handle"'s semantics but you didn't introduce it.

ffi/src/lib.rs
rkuris: This comment is still unresolved.

ffi/src/iterator.rs
rkuris: What was the result of the performance check?

ffi/src/value/results.rs
rkuris: Still outstanding.

ffi/src/value/results.rs
rkuris: This is still a little confusing. The documentation is for this variant, but shouldn't the caller generally call `fwd_free_owned_bytes` regardless of which variant of KeyValueResult is returned, or must they check that it is this variant and only call that method if so?

firewood/src/iter.rs
```diff
@@ -62,11 +64,45 @@ enum NodeIterState {
     },
 }
 
+/// A reference to the merkle, could be borrowed, or behind an Arc
```
rkuris: I don't fully understand yet why you need MerkleRef at all. I would expect a reference to the merkle or an actual Merkle is sufficient here.

Is the only reason we have both because we want to understand the performance differences? If so, it seems like the faster one should prevail and we should remove the code for the slower one.

If we need both, can you explain in the comments:
 - Why we need both variants
 - When and why one should use the Borrowed variant and when one should use the Owned variant
 - If you can convert from one to the other
 - Why it's called MerkleRef (it doesn't seem to be a reference to a merkle in the owned case)
 - Why you couldn't use a Cow<Merkle<T>>

rkuris (changes requested):

ffi/src/lib.rs
AminR443: I'm following the same style as @demosdemon did with the FFI refactor, so the codebase doesn't look diverge in style. All the other methods are accepting handles as options and behaving in the same way. I could change it if you want.

firewood/src/iter.rs
AminR443: We don't need both and we can always rely on the Arc/Owned one instead. The borrowed one should be more performant, however, it's not possible to use that without unsafe in FFI. Let me explain it in a clear way; the way the MerkleKeyValueIter and MerkleNodeIter are designed is:

```rust
// here we get a MerkleKeyValueIter which borrows the nodestore.
let it = db_view.iter(...)

for item in it {
  // ...
}
```

There is no lifetime issue here and the use-case is get the iterator, retrieve some key values and then we're done. In the FFI, we want to pass the iterator to Go, which borrows `db_view`, so it's not possible to safely return that. The issue is still there if we use `ArcDynDbView` because calling `iter` is still borrowing that specific Arc reference, and is not cloning it. That's why I had to make the iterator work with Arc<NodeStore<...>> so we can safely pass the iterator to Go.

- **Why we need both variants?** 
It's not necessary to keep both. We can use Arc all the time, but this would hurt the performance for the case that we're not using the iterator for FFI purposes.
- **When and why one should use the Borrowed variant and when one should use the Owned variant**
Owned variant is only needed when we want to have a long living iterator, which is only for FFI. Borrowed should be used anywhere else.
- **If you can convert from one to the other**
If we put a constraint for T to be Clone, then we should be able to.
- **Why it's called MerkleRef (it doesn't seem to be a reference to a merkle in the owned case)**
If we decide to keep it, I'll choose a better name.
- **Why you couldn't use a Cow<Merkle>**
We're just referencing and not writing, which in this case Cow would have the same lifetime limit discussed.

I wanted the changes to be contained and not alter all the places that `&merkle` has been used and unnecessarily require usage of Arc.

demosdemon:
idk your level of git foo, but here's how I normally resolve the gnarly merge conflicts from the squash and merge:

- Find the commit from the base branch where I merged main before the squash and merge, 
- merge that specific commit into my child branch; this should be a clean merge unless there were conflicts between base branch and child branch
- merge the specific squash commit in `main` with the `ours` strategy; this should be a net-zero change (no tree updates)
- merge the rest of `main`

So, to resolve your conflicts with main now that my PR is merged, do:

```sh
# let git fetch the expired branch PR refs
git config --add remote.origin.fetch '+refs/pull/*/head:refs/remotes/origin/pr/*'
# fetch everything again
git fetch origin
git merge fcc07b8ba5e0716e3d8892a275010e77daaebbe5
git merge -s ours 8df1ccc1464f58839c732b57c6cd703e88137abc
git merge origin/main
```

The git config change will bloat your local repo. You can keep it if you want, but I do recommend removing it after, fetching again, and then running `git prune` and `git gc` to remove the bloat.

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use derive_where::derive_where;
+use firewood::merkle;
+use firewood::v2::api::{self, BoxKeyValueIter};
+
+type KeyValueItem = Result<(merkle::Key, merkle::Value), api::Error>;
+
+/// An opaque wrapper around an Iterator.
```
rkuris: nit
```suggestion
/// An opaque wrapper around a [BoxKeyValueIter].
```

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive_where(Debug)]
+#[derive_where(skip_inner)]
+pub struct IteratorHandle<'view> {
+    iterator: BoxKeyValueIter<'view>,
+}
```
rkuris: Can't this just be a newtype?
```suggestion
pub struct IteratorHandle<'view>(BoxKeyValueIter<'view>);
}
```

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+impl Iterator for IteratorHandle<'_> {
+    type Item = KeyValueItem;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iterator.next()
+    }
+}
```
rkuris: Can we just `impl Deref` and get this for free?

ffi/src/iterator.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#[derive(Debug)]
+pub struct CreateIteratorResult<'db> {
+    pub handle: IteratorHandle<'db>,
+}
```
rkuris: Also a newtype:
```suggestion
pub struct CreateIteratorResult<'db>(pub IteratorHandle<'db>);
```

ffi/src/proposal.rs
```diff
@@ -97,8 +98,19 @@ impl ProposalHandle<'_> {
 
         Ok(hash_key)
     }
-}
 
+    /// Creates an iterator on the proposal starting from the given key.
+    #[expect(clippy::missing_errors_doc)]
```
rkuris: Interesting because it seems like `iter_option` is infallible. This should be investigated, and if that's true, create an issue as a followup.

If it is actually infallible, we don't need to return a result, we can instead return a `CreateIteratorResult<'_>` directly by calling `expect` on the return from `iter_option`.

ffi/src/value/kvp.rs
```diff
@@ -61,3 +61,20 @@ impl<'a> api::KeyValuePair for &KeyValuePair<'a> {
         (*self).into_batch()
     }
 }
+
+/// Owned version of `KeyValuePair`, returned to the FFI.
```
rkuris:
```suggestion
/// Owned version of `KeyValuePair`, returned to ffi callers
```

This type needs documentation on how it should be cleaned up by the caller, since `OwnedBytes` has a caution that says "C callers must free this memory using the respective FFI function for the concrete type".

If callers aren't supposed to clean this up and instead are supposed to clean up `IteratorResult`, we should state that. In this case, presumably the only reason we're `pub` here is for generating `firewood.h`

ffi/src/value/results.rs
```diff
@@ -307,6 +307,84 @@ pub enum ProposalResult<'db> {
...
+    /// Building the iterator was successful and the iterator handle is returned
+    Ok {
+        /// An opaque pointer to the [`IteratorHandle`].
+        /// The value should be freed with [`fwd_free_iterator`]
+        ///
+        /// [`fwd_free_iterator`]: crate::fwd_free_iterator
+        handle: Box<IteratorHandle<'db>>,
+    },
+    /// An error occurred and the message is returned as an [`OwnedBytes`]. The
+    /// value is guaranteed to contain only valid UTF-8.
```
rkuris: How is this guaranteed? I think you might need to say that all constructors are expected to generate valid UTF-8, but I'm not even sure that's necessary.

ffi/src/proposal.rs
AminR443: You're right. Created #1329.

ffi/src/value/results.rs
AminR443: I removed that part. As mentioned, this mirrors @demosdemon's comments. All our FFI results use String for errors, which guarantees valid UTF-8. We just create `OwnedBytes` from it and pass it to FFI.

ffi/src/value/kvp.rs
```diff
@@ -61,3 +61,23 @@ impl<'a> api::KeyValuePair for &KeyValuePair<'a> {
         (*self).into_batch()
     }
 }
+
+/// Owned version of `KeyValuePair`, returned to ffi callers.
+///
+/// C callers must free this memory using the respective FFI function for the
+/// concrete type (but not using the `free` function from the C standard library).
```
rkuris: This is still a bit confusing. This is a comment for `OwnedKeyValuePair` which is a concrete type. What is the "respective FFI function for the concrete type" if the concrete type is `OwnedKeyValuePair`?

ffi/src/value/results.rs
rkuris: This doc looks correct to me, but the comment is not resolved. Is there more to do here? If not, please mark this resolved.

ffi/src/proposal.rs
```diff
@@ -97,8 +98,15 @@ impl ProposalHandle<'_> {
 
         Ok(hash_key)
     }
-}
 
+    /// Creates an iterator on the proposal starting from the given key.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
+    pub fn iter_from(&self, first_key: Option<&[u8]>) -> CreateIteratorResult<'_> {
+        let it = self.iter_option(first_key).expect("infallible");
```
rkuris: The expect message should be more meaningful. I'd go so far as to refer to the new issue in the string "infallible; see issue NNNN"

another option is adding "// TODO: iter_option cannot fail; see issue NNNN"

rkuris (approved):

PR #1256, opened 2025-09-03
feat(ffi-iterator): Implementation of Iterator in Go (2/4)

AminR443:
This implement a simple Iterator interface in Go (`Next`, `Key`, `Value`) using the introduced FFI functions, without batching. 

Depends On: #1255

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -728,10 +728,6 @@ typedef enum KeyValueResult_Tag {
    */
   KeyValueResult_NullHandlePointer,
   /**
-   * The provided root was not found in the database.
-   */
-  KeyValueResult_RevisionNotFound,
-  /**
    * The iterator is exhausted
    */
   KeyValueResult_None,
@@ -759,9 +755,6 @@ typedef enum KeyValueResult_Tag {
 typedef struct KeyValueResult {
   KeyValueResult_Tag tag;
   union {
-    struct {
-      struct HashKey revision_not_found;
-    };
     struct {
       struct OwnedKeyValuePair some;
     };
--- a/ffi/memory.go
+++ b/ffi/memory.go
@@ -346,10 +346,6 @@ func getKeyValueFromKeyValueResult(result C.KeyValueResult) (*ownedKeyValue, err
 	switch result.tag {
 	case C.KeyValueResult_NullHandlePointer:
 		return nil, errDBClosed
-	case C.ValueResult_RevisionNotFound:
-		// NOTE: the result value contains the provided root hash, we could use
-		// it in the error message if needed.
-		return nil, errRevisionNotFound
 	case C.KeyValueResult_None:
 		return nil, nil
 	case C.KeyValueResult_Some:
--- a/ffi/src/value/results.rs
+++ b/ffi/src/value/results.rs
@@ -336,8 +336,6 @@ pub enum IteratorResult<'db> {
 pub enum KeyValueResult {
     /// The caller provided a null pointer to an iterator handle.
     NullHandlePointer,
-    /// The provided root was not found in the database.
-    RevisionNotFound(HashKey),
     /// The iterator is exhausted
     None,
     /// The next item is returned.
@@ -362,9 +360,6 @@ impl From<Option<Result<(merkle::Key, merkle::Value), api::Error>>> for KeyValue
         match value {
             Some(value) => match value {
                 Ok(value) => KeyValueResult::Some(value.into()),
-                Err(api::Error::RevisionNotFound { provided }) => KeyValueResult::RevisionNotFound(
-                    HashKey::from(provided.unwrap_or_else(api::HashKey::empty)),
-                ),
                 Err(err) => KeyValueResult::Err(err.to_string().into_bytes().into()),
             },
             None => KeyValueResult::None,


ffi/firewood_test.go
```diff
@@ -1043,3 +1043,80 @@ func TestGetFromRootParallel(t *testing.T) {
...
+	r.NoError(err)
+
+	h2, err := db.IterOnRoot(secondRoot, nil)
+	r.NoError(err)
+
+	for i := 0; h1.Next() && h2.Next(); i += 1 {
+		r.Equal(keys[i], h1.Key())
+		r.Equal(keys[i], h2.Key())
+		r.Equal(vals[i], h1.Value())
+		r.Equal(vals[i+10], h2.Value())
```
alarso16: I'm not in love with incrementing by 10 for this section, I would prefer augmenting this into `vals1` and `vals2`, but a comment describing what you're doing at the `Update` call is sufficient if you feel strongly about it

ffi/firewood_test.go
```diff
@@ -1043,3 +1043,80 @@ func TestGetFromRootParallel(t *testing.T) {
...
+	}
+	r.NoError(h1.Err())
+	r.NoError(h2.Err())
```
alarso16: Maybe add some logs here about where the iterator failed - might help with debugging (this won't compile btw)
```suggestion
	r.NoErrorf(h2.Err(), "failed on iter %d", i)
```

ffi/firewood_test.go
```diff
@@ -1043,3 +1043,80 @@ func TestGetFromRootParallel(t *testing.T) {
...
+	p, err := db.Propose(keys, vals)
+	r.NoError(err)
+
+	it, err := p.Iter(nil)
+	r.NoError(err)
+
+	for i := 0; it.Next(); i += 1 {
+		r.Equal(keys[i], it.Key())
+		r.Equal(vals[i], it.Value())
+	}
```
alarso16: nit: Might want to check that you've iterated through all the values. Feel free to ignore

ffi/iterator.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// currentKey is the current key retrieved from the iterator
+	currentKey []byte
+	// currentVal is the current value retrieved from the iterator
+	currentVal []byte
+	// err is the error from the iterator, if any
+	err error
+}
+
+func (it *Iterator) Next() bool {
```
alarso16, line 33: Will you add some doc comments for the exported functions?

ffi/firewood.go
```diff
@@ -198,6 +198,28 @@ func (db *Database) GetFromRoot(root, key []byte) ([]byte, error) {
...
+// IterLatest creates and iterator starting from the provided key on the latest root hash.
+// pass empty slice to start from beginning
+func (db *Database) IterLatest(key []byte) (*Iterator, error) {
+	return db.IterOnRoot(nil, key)
+}
+
+// IterOnRoot creates and iterator starting from the provided key on a specific root hash.
+// If the root is not found, it returns as an error.
+// pass empty slice to start from beginning
+func (db *Database) IterOnRoot(root, key []byte) (*Iterator, error) {
```
alarso16: Does this work on proposals too?

ffi/firewood.go
```diff
@@ -198,6 +198,28 @@ func (db *Database) GetFromRoot(root, key []byte) ([]byte, error) {
 	))
 }
 
+// IterLatest creates and iterator starting from the provided key on the latest root hash.
+// pass empty slice to start from beginning
+func (db *Database) IterLatest(key []byte) (*Iterator, error) {
```
alarso16: By convention, all the other functions have excluded `Latest` from the name

alarso16 (commented):
It looks good to me, I left some non-blocking comments and didn't look at the Rust code

alarso16:
I do think there's some test cases missing. What if you're using `IterLatest` and a proposal is committed? What if you're using `Iter` on a proposal and it's committed? What happens if the database closed? These should be easy tests to right, and their behavior isn't necessarily obvious but should be well defined

ffi/iterator.go
```diff
@@ -63,3 +63,11 @@ func (it *Iterator) Value() []byte {
 func (it *Iterator) Err() error {
 	return it.err
 }
+
+// Drop drops the iterator and releases the resources
+func (it *Iterator) Drop() error {
```
alarso16, line 68: When is this called? Do you need to call it in each test?

ffi/firewood.go
AminR443: Yes it does now. I updated the rust base to support this.

ffi/iterator.go
AminR443: Right now I call it in here:
```go
func assertIteratorYields(r *require.Assertions, it kvIter, keys [][]byte, vals [][]byte) {
	i := 0
	for ; it.Next(); i += 1 {
		r.Equal(keys[i], it.Key())
		r.Equal(vals[i], it.Value())
	}
	r.NoError(it.Err())
	r.Equal(len(keys), i)
	r.NoError(it.Drop())
}
```

In a real use-case, I'm thinking of a better alternative now. In rust, I can drop the reference to the nodestore as soon as the iterator is exhausted. It is also possible to call drop automatically in go, and set the handle to nil as soon as iterator is exhausted or errored.

alarso16: In general, I think it would be more clear to duplicate the code for each and call it like this (to ensure it gets cleaned if we exit early):
```
it, err := db.Iter(...)
require.NoError(err)
t.Cleanup(func() { 
     require.NoError(it.Drop())
})
```

AminR443: Done

ffi/firewood.go
```diff
@@ -214,6 +214,17 @@ func (db *Database) Root() ([]byte, error) {
...
+func (db *Database) LatestRevision() (*Revision, error) {
+	root, err := db.Root()
+	if err != nil {
+		return nil, err
+	}
+	if bytes.Equal(root, EmptyRoot) {
+		return nil, errRevisionNotFound
+	}
+	return db.Revision(root)
+}
```
rkuris, line 226: This is racy. If a commit occurs after line 218 and before line 225, you won't have the latest revision.

ffi/firewood_test.go
```diff
@@ -1059,3 +1059,144 @@ func TestGetFromRootParallel(t *testing.T) {
 		r.NoError(err, "Parallel operation failed")
 	}
 }
+
+func assertIteratorYields(r *require.Assertions, it *Iterator, keys [][]byte, vals [][]byte) {
+	i := 0
+	for ; it.Next(); i += 1 {
+		r.Equal(keys[i], it.Key())
+		r.Equal(vals[i], it.Value())
+	}
```
rkuris, line 1145: Why did we decide not to use go iterators and range here?

ffi/iterator.go
```diff
@@ -0,0 +1,92 @@
...
+// Key returns the key of the current pair
+func (it *Iterator) Key() []byte {
+	if (it.currentKey == nil && it.currentVal == nil) || it.err != nil {
+		return nil
+	}
+	return it.currentKey
+}
+
+// Value returns the value of the current pair
+func (it *Iterator) Value() []byte {
```
rkuris, line 55: Is there ever a case where the caller will only be interested in the key or the value (not both)? If not, then it seems like having a method called `KeyValue` would be a lot more usable, which returns (key []byte, value []byte)

ffi/src/iterator.rs
```diff
@@ -8,27 +8,35 @@ use firewood::v2::api::{self, BoxKeyValueIter};
...
 
 impl Iterator for IteratorHandle<'_> {
     type Item = KeyValueItem;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.iterator.next()
+        let out = self.iterator.as_mut()?.next();
+        if out.is_none() {
+            // iterator exhausted; drop it so the NodeStore can be released
+            self.iterator.take();
```
rkuris: why take?

This should be:
```suggestion
            self.iterator = None;
```

rkuris (approved):

ffi/firewood.go
alarso16: It seems like the user's responsibility to ensure that there isn't a commit occurring.

PR #1257, opened 2025-09-03
feat(ffi-iterator): Batching support for Iterator (3/4)

AminR443:
This PR introduces batched retrieval from iterator, to reduce overheads introduced from too many calls to FFI.

Depends On: #1256

### DIFF
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -1162,54 +1162,58 @@ func TestIter(t *testing.T) {
 	thirdRoot, err := db.Update(keys[160:], vals[160:])
 	r.NoError(err)
 
-	for _, dataMode := range dataModes {
-		for _, batchMode := range batchModes {
-			t.Run(fmt.Sprintf("Latest/%s/%s", dataMode.name, batchMode.name), func(t *testing.T) {
-				r := require.New(t)
-				it, err := db.IterLatest(nil)
-				r.NoError(err)
-
-				batchMode.configFn(it)
-				assertIteratorYields(r, dataMode.configFn(it), keys, vals)
-			})
-
-			t.Run(fmt.Sprintf("OnRoot/%s/%s", dataMode.name, batchMode.name), func(t *testing.T) {
-				r := require.New(t)
-				h1, err := db.IterOnRoot(firstRoot, nil)
-				r.NoError(err)
-				h2, err := db.IterOnRoot(secondRoot, nil)
-				r.NoError(err)
-				h3, err := db.IterOnRoot(thirdRoot, nil)
-				r.NoError(err)
-				batchMode.configFn(h1)
-				batchMode.configFn(h2)
-				batchMode.configFn(h3)
-				assertIteratorYields(r, dataMode.configFn(h1), keys[:80], vals[:80])
-				assertIteratorYields(r, dataMode.configFn(h2), keys[:160], vals[:160])
-				assertIteratorYields(r, dataMode.configFn(h3), keys, vals)
-			})
-
-			t.Run(fmt.Sprintf("OnProposal/%s/%s", dataMode.name, batchMode.name), func(t *testing.T) {
-				r := require.New(t)
-				updatedValues := make([][]byte, len(vals))
-				copy(updatedValues, vals)
-
-				changedKeys := make([][]byte, 0)
-				changedVals := make([][]byte, 0)
-				for i := 0; i < len(vals); i += 4 {
-					changedKeys = append(changedKeys, keys[i])
-					newVal := []byte{byte(i)}
-					changedVals = append(changedVals, newVal)
-					updatedValues[i] = newVal
-				}
-				p, err := db.Propose(changedKeys, changedVals)
-				r.NoError(err)
-				it, err := p.Iter(nil)
-				r.NoError(err)
-
-				batchMode.configFn(it)
-				assertIteratorYields(r, dataMode.configFn(it), keys, updatedValues)
-			})
+	runForAllModes := func(parentT *testing.T, name string, fn func(*testing.T, func(it *Iterator) kvIter)) {
+		for _, dataMode := range dataModes {
+			for _, batchMode := range batchModes {
+				parentT.Run(fmt.Sprintf("%s/%s/%s", name, dataMode.name, batchMode.name), func(t *testing.T) {
+					fn(t, func(it *Iterator) kvIter {
+						batchMode.configFn(it)
+						return dataMode.configFn(it)
+					})
+				})
+			}
 		}
 	}
+	runForAllModes(t, "Latest", func(t *testing.T, fn func(it *Iterator) kvIter) {
+		r := require.New(t)
+		it, err := db.IterLatest(nil)
+		r.NoError(err)
+
+		assertIteratorYields(r, fn(it), keys, vals)
+	})
+
+	runForAllModes(t, "OnRoot", func(t *testing.T, fn func(it *Iterator) kvIter) {
+		r := require.New(t)
+		h1, err := db.IterOnRoot(firstRoot, nil)
+		r.NoError(err)
+		h2, err := db.IterOnRoot(secondRoot, nil)
+		r.NoError(err)
+		h3, err := db.IterOnRoot(thirdRoot, nil)
+		r.NoError(err)
+
+		assertIteratorYields(r, fn(h1), keys[:80], vals[:80])
+		assertIteratorYields(r, fn(h2), keys[:160], vals[:160])
+		assertIteratorYields(r, fn(h3), keys, vals)
+	})
+
+	runForAllModes(t, "OnProposal", func(t *testing.T, fn func(it *Iterator) kvIter) {
+		r := require.New(t)
+		updatedValues := make([][]byte, len(vals))
+		copy(updatedValues, vals)
+
+		changedKeys := make([][]byte, 0)
+		changedVals := make([][]byte, 0)
+		for i := 0; i < len(vals); i += 4 {
+			changedKeys = append(changedKeys, keys[i])
+			newVal := []byte{byte(i)}
+			changedVals = append(changedVals, newVal)
+			updatedValues[i] = newVal
+		}
+		p, err := db.Propose(changedKeys, changedVals)
+		r.NoError(err)
+		it, err := p.Iter(nil)
+		r.NoError(err)
+
+		assertIteratorYields(r, fn(it), keys, updatedValues)
+	})
 }
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -1199,3 +1199,30 @@ func TestIterBatched(t *testing.T) {
 	r.NoError(it2.Err())
 	r.Equal(i, 1000)
 }
+
+// Tests that batched iterator functionality works
+func TestIterBatchedBorrowed(t *testing.T) {
+	r := require.New(t)
+	db := newTestDatabase(t)
+
+	keys, vals := kvForBench(1000)
+	_, err := db.Update(keys, vals)
+	r.NoError(err)
+
+	it, err := db.IterLatest(nil)
+	r.NoError(err)
+	it.SetBatchSize(100)
+	it2, err := db.IterLatest(nil)
+	r.NoError(err)
+
+	i := 0
+	for ; it.NextBorrowed() && it2.NextBorrowed(); i += 1 {
+		r.Equal(it.Key(), it2.Key())
+		r.Equal(it.Value(), it2.Value())
+		r.Equal(keys[i], it.Key())
+		r.Equal(vals[i], it.Value())
+	}
+	r.NoError(it.Err())
+	r.NoError(it2.Err())
+	r.Equal(i, 1000)
+}
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -5,13 +5,14 @@ package ffi
 
 import (
 	"bytes"
+	"crypto/rand"
 	"encoding/hex"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -268,8 +269,8 @@ func sortKV(keys, vals [][]byte) error {
 	for i := range ord {
 		ord[i] = i
 	}
-	sort.Slice(ord, func(i, j int) bool {
-		return bytes.Compare(keys[ord[i]], keys[ord[j]]) < 0
+	slices.SortFunc(ord, func(i, j int) int {
+		return bytes.Compare(keys[i], keys[j])
 	})
 	perm := make([]int, n)
 	for dest, orig := range ord {
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -7,10 +7,12 @@ import (
 	"bytes"
 	"crypto/rand"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -254,6 +256,37 @@ func randomBytes(n int) []byte {
 	return b
 }
 
+// sortKV sorts keys lexicographically and keeps vals paired.
+func sortKV(keys, vals [][]byte) error {
+	if len(keys) != len(vals) {
+		return errors.New("keys/vals length mismatch")
+	}
+	n := len(keys)
+	if n <= 1 {
+		return nil
+	}
+	ord := make([]int, n)
+	for i := range ord {
+		ord[i] = i
+	}
+	sort.Slice(ord, func(i, j int) bool {
+		return bytes.Compare(keys[ord[i]], keys[ord[j]]) < 0
+	})
+	perm := make([]int, n)
+	for dest, orig := range ord {
+		perm[orig] = dest
+	}
+	for i := 0; i < n; i++ {
+		for perm[i] != i {
+			j := perm[i]
+			keys[i], keys[j] = keys[j], keys[i]
+			vals[i], vals[j] = vals[j], vals[i]
+			perm[i], perm[j] = perm[j], j
+		}
+	}
+	return nil
+}
+
 func kvForBench(num int) ([][]byte, [][]byte) {
 	keys := make([][]byte, num)
 	vals := make([][]byte, num)
@@ -262,6 +295,7 @@ func kvForBench(num int) ([][]byte, [][]byte) {
 		keys[i] = randomBytes(32)
 		vals[i] = randomBytes(128)
 	}
+	_ = sortKV(keys, vals)
 	return keys, vals
 }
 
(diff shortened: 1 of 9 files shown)


ffi/iterator.go
```diff
@@ -19,41 +19,120 @@ type Iterator struct {
...
+// the difference is that returned bytes in Key and Value are not copied
+// and will be freed on next call to Next or NextBorrowed
+func (it *Iterator) NextBorrowed() bool {
+	it.err = it.nextInternal()
+	if it.currentPair == nil || it.err != nil {
+		return false
+	}
+	it.currentKey = it.currentPair.key.BorrowedBytes()
+	it.currentValue = it.currentPair.value.BorrowedBytes()
+	it.err = nil
```
alarso16, line 122: This line is unnecessary I believe
```suggestion
```

ffi/firewood_test.go
```diff
@@ -1054,69 +1106,114 @@ func TestIterEmptyDb(t *testing.T) {
...
+	r.NoError(err)
+	thirdRoot, err := db.Update(keys[160:], vals[160:])
 	r.NoError(err)
 
-	for i := 0; h1.Next() && h2.Next(); i += 1 {
-		r.Equal(keys[i], h1.Key())
-		r.Equal(keys[i], h2.Key())
-		r.Equal(vals[i], h1.Value())
-		r.Equal(vals[i+10], h2.Value())
+	runForAllModes := func(parentT *testing.T, name string, fn func(*testing.T, func(it *Iterator) kvIter)) {
```
alarso16: I don't think this is a good way of refactoring these tests. I would recommend having each functionality (e.g. "Latest" or "OnRoot") as a separate test, with a constructor. The overhead of creating a new test database each time should be ignored - I'd prefer readable tests. What if it looked something like this?

```
func TestIterLatest(t *testing.T) {
      runAllModes(t, func(t *testing.T) {
             // custom test code here
       })
}
func runAllModes(t *testing, op func( t *testing.T)) {
     for _, dataMode := range dataModes {
			for _, batchMode := range batchModes {
                t.Run(fmt.Sprintf(""....), func (t *testing.T) {
				     runSpecificMode(t, dataMode, batchMode, op)
                 })
			}
		}
}
func runSpecificMode(t *testing.T, dataMode string, batchMode configFn, op func(t *testing.T)) {
       // setup database
       // do test
}
```

Re-creating the database is a little annoying, but I don't think having one big "mega test" is the right answer. Feel free to refactor a different way, but I think something along this line will be a lot easier to read and debug

ffi/iterator.go
```diff
@@ -19,41 +19,120 @@ type Iterator struct {
...
+	batchSize int
+
+	// loadedPairs is the latest loaded key value pairs retrieved
+	// from the iterator, not yet consumed by user
+	loadedPairs []*ownedKeyValue
+
+	// currentPair is the current pair retrieved from the iterator
+	currentPair *ownedKeyValue
+
+	// currentKey is the current pair retrieved from the iterator
```
alarso16: This comment is kind of wrong. But also it's relatively obvious because you named it well... maybe group `current***` into a single block and comment all of it

ffi/iterator.go
```diff
@@ -19,41 +19,120 @@ type Iterator struct {
...
-	return e == nil
+	it.currentValue = v
+	it.err = nil
+	return true
+}
+
+// NextBorrowed retrieves the next item on the iterator similar to Next
+// the difference is that returned bytes in Key and Value are not copied
+// and will be freed on next call to Next or NextBorrowed
+func (it *Iterator) NextBorrowed() bool {
```
alarso16, line 115: The doc is a little vague about the implications (i.e. holding onto the bytes returned will be UB, which is terrifying) and I think the description should make you scared...

ffi/iterator.go
```diff
@@ -19,41 +19,120 @@ type Iterator struct {
...
+		} else {
+			batch, e := getKeyValueBatchFromKeyValueBatchResult(C.fwd_iter_next_n(it.handle, C.size_t(it.batchSize)))
+			if e != nil {
+				return e
+			}
+			it.loadedPairs = batch.Copied()
+			it.currentResource = batch
+		}
+	}
+	if len(it.loadedPairs) > 0 {
```
alarso16: It would be nice if you return early at the beginning if `len(it.loadedPairs) > 0` just to clean up the code

ffi/memory.go
```diff
@@ -396,14 +450,38 @@ func getKeyValueFromKeyValueResult(result C.KeyValueResult) (*ownedKeyValue, err
...
 		return nil, fmt.Errorf("unknown C.KeyValueResult tag: %d", result.tag)
 	}
 }
 
+// getKeyValueBatchFromKeyValueBatchResult converts a C.KeyValueResult to a key value pair or error.
+//
+// It returns nil, nil if the result is None.
+// It returns a *ownedKeyValueBatch, nil if the result is Some.
+// It returns an error if the result is an error.
+func getKeyValueBatchFromKeyValueBatchResult(result C.KeyValueBatchResult) (*ownedKeyValueBatch, error) {
```
alarso16: This function name is a little long and redundant, any way we could shorten it? Here's my idea:
```suggestion
func getKeyValueBatchFromResult(result C.KeyValueBatchResult) (*ownedKeyValueBatch, error) {
```

ffi/memory.go
```diff
@@ -350,21 +350,75 @@ func getValueFromValueResult(result C.ValueResult) ([]byte, error) {
...
 	value := kv.value.CopiedBytes()
-	if err := kv.value.Free(); err != nil {
-		return nil, nil, fmt.Errorf("%w: %w", errFreeingValue, err)
+	return key, value
+}
+
+func (kv *ownedKeyValue) Free() error {
+	err := kv.key.Free()
+	if err != nil {
+		return fmt.Errorf("%w: %w", errFreeingValue, err)
```
alarso16: If there's an error freeing the first value, should we still try to free the second one? I would say yes. You can use `errors.Join` to report back both errors too

ffi/memory.go
```diff
@@ -350,21 +350,75 @@ func getValueFromValueResult(result C.ValueResult) ([]byte, error) {
 	}
 }
 
+type ownedKeyValueBatch struct {
+	owned C.OwnedKeyValueBatch
+}
+
+func (b *ownedKeyValueBatch) Copied() []*ownedKeyValue {
```
alarso16: It doesn't totally make sense to me to have exported functions on unexported structs, since it's not implementing an interface.

ffi/firewood_test.go
AminR443: Agreed. Aggregating them in a one big test because of database creation was not quite logical. I implemented a better version now.

ffi/iterator.go
AminR443: The updated one looks more terrifying and is more detailed. Let me know what you think.

ffi/memory.go
AminR443: Makes sense, I unexported the methods corresponding to unexported structs.

ffi/iterator.go
AminR443: batched these together, I think it makes more sense now.

ffi/firewood_test.go
```diff
@@ -1044,7 +1096,34 @@ func TestGetFromRootParallel(t *testing.T) {
...
+	SetBatchSize(int)
+	Next() bool
+	Key() []byte
+	Value() []byte
+	Err() error
+	Drop() error
+}
+type borrowIter struct{ it *Iterator }
+
+func (b borrowIter) SetBatchSize(int) { b.it.SetBatchSize(1) }
```
alarso16: Is this intentionally overridden? I assume so, but will you add a comment why? Even if it's just `//TODO: allow multiple batch sizes`

ffi/firewood_test.go
```diff
@@ -1055,71 +1134,113 @@ func assertIteratorYields(r *require.Assertions, it *Iterator, keys [][]byte, va
...
+	"Batched": func(it kvIter) kvIter {
+		it.SetBatchSize(100)
+		return it
+	},
+}
 
-	it, err := db.Iter(nil)
-	r.NoError(err)
-	r.False(it.Next())
+func runIteratorTestForModes(parentT *testing.T, fn func(*testing.T, iteratorConfigFn), modes ...string) {
```
alarso16: This should just be named `t`

ffi/firewood_test.go
```diff
@@ -1055,71 +1134,113 @@ func assertIteratorYields(r *require.Assertions, it *Iterator, keys [][]byte, va
...
+	r := require.New(parentT)
```
alarso16: This should moved inside the `t.Run` statement so you can tell which test failed. Otherwise, the names might get funky

ffi/firewood_test.go
```diff
@@ -1055,71 +1134,113 @@ func assertIteratorYields(r *require.Assertions, it *Iterator, keys [][]byte, va
...
+				config, ok := iterConfigs[m]
+				r.Truef(ok, "specified config mode %s does not exist", m)
+				it = config(it)
+			}
+			return it
+		})
+	})
+}
+
+func runIteratorTestForAllModes(parentT *testing.T, fn func(*testing.T, iteratorConfigFn)) {
```
alarso16, line 1231:
```suggestion
func runIteratorTestForAllModes(t *testing.T, fn func(*testing.T, iteratorConfigFn)) {
```

ffi/firewood_test.go
```diff
@@ -1044,7 +1096,34 @@ func TestGetFromRootParallel(t *testing.T) {
...
+	Next() bool
+	Key() []byte
+	Value() []byte
+	Err() error
+	Drop() error
+}
+type borrowIter struct{ it *Iterator }
+
+func (b borrowIter) SetBatchSize(int) { b.it.SetBatchSize(1) }
+func (b borrowIter) Next() bool       { return b.it.NextBorrowed() }
```
alarso16: It's kind of strange that this doesn't take a reference

ffi/firewood_test.go
```diff
@@ -247,6 +250,55 @@ func kvForTest(num int) ([][]byte, [][]byte) {
...
+			j := perm[i]
+			keys[i], keys[j] = keys[j], keys[i]
+			vals[i], vals[j] = vals[j], vals[i]
+			perm[i], perm[j] = perm[j], j
+		}
+	}
+	return nil
+}
+
+func kvForBench(num int) ([][]byte, [][]byte) {
```
alarso16: This name is weird, since it's not clear to me why I wouldn't use the other key-value generator. Maybe add a comment about its purpose

ffi/iterator.go
```diff
@@ -66,8 +151,31 @@ func (it *Iterator) Err() error {
 
 // Drop drops the iterator and releases the resources
 func (it *Iterator) Drop() error {
+	e1 := it.freeCurrentAllocation()
```
alarso16: If it's the only error being returned, it could probably be renamed
```suggestion
	err := it.freeCurrentAllocation()
```

ffi/iterator.go
```diff
@@ -66,8 +151,31 @@ func (it *Iterator) Err() error {
...
 	if it.handle != nil {
-		return getErrorFromVoidResult(C.fwd_free_iterator(it.handle))
+		return errors.Join(
```
alarso16, line 153: You should add a comment somewhere about how all errors are being returned intentionally, since someone might look at this and think it's weird.

ffi/iterator.go
alarso16: Maybe say explicitly it can cause undefined behavior. And also maybe get rid of the emoji, open to opinions but I don't think we use it anywhere else in the codebase...

ffi/firewood_test.go
AminR443: Not intentional. Thank you for catching it, fixed.

ffi/firewood_test.go
AminR443: I'm not following, could you explain a little bit more? the referred lines are from TestIterEmptyDb.

ffi/firewood_test.go
AminR443: Removed this for now. I used this for my benchmarks and just stayed for the tests as well, because I sorted the keys to make sure order matches with iterator. moved the sorting to kvForTest and removed this function. Will add it later in the benchmark PR.

ffi/firewood_test.go
alarso16: In `runtIteratorTestForModes`, you use `parentT` for the require, when you should be using the child `T`. So you should move `r := require.New(parentT)` to the inside of `t.Run`

AminR443: Fixed.

ffi/src/iterator.rs
```diff
@@ -36,6 +36,22 @@ impl Iterator for IteratorHandle<'_> {
...
+        let mut items = Vec::<KeyValueItem>::new();
+        for _ in 0..n {
+            let item = self.next();
+            if let Some(item) = item {
+                items.push(item?);
+            } else {
+                break;
+            }
+        }
+        Ok(items)
```
rkuris: This code can be greatly simplified.

```suggestion
       self.take(n).map(|r| r.map_err(Into::into)).collect()
```

ffi/src/lib.rs
```diff
@@ -195,6 +195,36 @@ pub unsafe extern "C" fn fwd_iter_next(handle: Option<&mut IteratorHandle<'_>>)
...
+///
+/// * `handle` - The iterator handle returned by [`fwd_iter_on_revision`] or
+///   [`fwd_iter_on_proposal`].
+///
+/// # Returns
+///
+/// - [`KeyValueResult::NullHandlePointer`] if the provided iterator handle is null.
+/// - [`KeyValueResult::None`] if the iterator doesn't have any remaining values/exhausted.
+/// - [`KeyValueResult::Some`] if the next item on iterator was retrieved, with the associated
+///   key value pair.
```
rkuris: This looks incorrect. Doesn't it return a batch of items?

Arguably should mention that you might not get N values back if you're at the end.

Is it guaranteed that if you get less than N items back that you're sure to get None on the next call?

ffi/src/lib.rs
```diff
@@ -195,6 +195,36 @@ pub unsafe extern "C" fn fwd_iter_next(handle: Option<&mut IteratorHandle<'_>>)
...
+/// - [`KeyValueResult::Err`] if an error occurred while retrieving the next item on iterator.
```
rkuris: What happens to the iterator at this point? Can you still fetch from it or should it be closed?

What are the errors that could happen here? Presumably the only errors are IO errors or merkle structure errors, but I'm not sure.

ffi/src/lib.rs
```diff
@@ -195,6 +195,36 @@ pub unsafe extern "C" fn fwd_iter_next(handle: Option<&mut IteratorHandle<'_>>)
...
+///
+/// # Arguments
+///
+/// * `handle` - The iterator handle returned by [`fwd_iter_on_revision`] or
+///   [`fwd_iter_on_proposal`].
+///
+/// # Returns
+///
+/// - [`KeyValueResult::NullHandlePointer`] if the provided iterator handle is null.
+/// - [`KeyValueResult::None`] if the iterator doesn't have any remaining values/exhausted.
```
rkuris: After this is returned, what happens if you call the iterator again? This should be documented and tested. See the documentation for FusedIterator for more information.

ffi/src/value/results.rs
```diff
@@ -373,6 +374,49 @@ impl From<Option<Result<(merkle::Key, merkle::Value), api::Error>>> for KeyValue
...
+
+impl From<Result<Vec<(merkle::Key, merkle::Value)>, api::Error>> for KeyValueBatchResult {
+    fn from(value: Result<Vec<(merkle::Key, merkle::Value)>, api::Error>) -> Self {
+        match value {
+            Ok(pairs) => {
+                let values: Vec<_> = pairs
+                    .into_iter()
+                    .map(|(k, v)| OwnedKeyValuePair {
+                        key: k.into(),
+                        value: v.into(),
```
rkuris: This seems overly complicated. Why doesn't `.map(Into::into)` work?

ffi/src/value/results.rs
```diff
@@ -373,6 +374,49 @@ impl From<Option<Result<(merkle::Key, merkle::Value), api::Error>>> for KeyValue
...
+                    })
+                    .collect();
+                KeyValueBatchResult::Some(values.into())
+            }
+            Err(api::Error::RevisionNotFound { provided }) => {
+                KeyValueBatchResult::RevisionNotFound(HashKey::from(
+                    provided.unwrap_or_else(api::HashKey::empty),
+                ))
```
rkuris: Why is this not a conversion?

AminR443: unnecessary error branch, existed because of previous version. removed.

ffi/src/lib.rs
AminR443: It's safe to call next on both versions (batched and single), will return None in single mode and empty batch in batch mode. Documented and added tests.

ffi/src/lib.rs
AminR443: I updated the documentation for batch mode's behavior. Your comment seems to reference a wrong function. It is guaranteed to get None if got less than N items back.

ffi/src/lib.rs
AminR443: Only possible error in Iterator is FileIoError and I updated the documentation. I'm not familiar with NodeReader/RootReader and the underlying storage access, I'll take a look tomorrow. @demosdemon: would appreciate it if you know the answer to this; what should we do if iterator returns an I/O error?

demosdemon: > What happens to the iterator at this point? 

The same as any other iterator. _It depends on the iterator._ 

Some iterators are not reentrant, meaning you cannot call next on it again. Some iterators are reentrant where you can. However, all iterators must still have their memory dropped because you are not taking ownership of the iterator when you call `next`.

And, whether or not the iterator is reentrant may also be dependent on specific error. If the error is structural, then it's probably not reentrant. But, if the error is io related, it depends on the error. An `Interrupted` or `WouldBlock` might be retryable but `NotFound` or `PermissionDenied` probably isn't.

tbh, most errors aren't reentrant and most of the time the assumption is that they aren't. In which case, the iterator should be freed immediately after it returns an error.

AminR443: @demosdemon thanks for the comment. I think it's best that I document it as not reentrant for now.

ffi/src/lib.rs
```diff
@@ -178,23 +178,61 @@ pub unsafe extern "C" fn fwd_iter_on_proposal<'p>(
...
+///
+/// Once an empty batch or items fewer than `n` is returned (iterator exhausted), subsequent calls
+/// will continue returning empty batches. You may still call this safely, but freeing the
+/// iterator with [`fwd_free_iterator`] is recommended.
+///
+/// # Safety
+///
+/// The caller must:
+/// * ensure that `handle` is a valid pointer to a [`IteratorHandle`].
+/// * call [`fwd_free_owned_key_value_batch`] on the returned batch to free any allocated memory.
```
rkuris, line 226: It seems like we could have done better here, but this is probably ok. I don't like the fact that we have a separate function to free the memory, and if someone calls the wrong one, bad things probably happen.

AminR443: I think having explicit free function for FFI types is clearer and less prone to mistakes. This also comes with a performance benefit if user borrows from the batch and all batch is freed at once when proceeding to the next batch.

PR #1258, opened 2025-09-04
feat: Parallel updates to Merkle trie (v2)

bernard-avalabs:
Second implementation of adding parallel updates to a Merkle trie. In this approach, each child of the root node is a separate trie that can be modified independently by a separate worker thread. Four steps are needed to create a parallel proposal:

1. Prepare: The trie is modified such that the root is always a branch with an empty partial path. This prevents any update from replacing the root node with one of its children, which is necessary to allow the sub-tries to be operated on independently.
2. Split: The updates from a batch are split and sent to different workers based on the first nibble of their corresponding keys.
3. Merge: After completing the batch, each worker returns the root of its sub-trie to the main thread. The main thread attaches the sub-trie roots to the children array of the root node.
4. Post-process: The Merkle trie is returned to its canonical form. This eliminates cases where a trie has a root node with no value and only one single child, or a trie has a root branch node with no value and no children.

The main changes are in parallel.rs. A test case (test_propose_parallel) has been added to test parallel insertion/deletion.

### DIFF
--- a/firewood/src/merkle/parallel.rs
+++ b/firewood/src/merkle/parallel.rs
@@ -355,7 +355,7 @@ impl ParallelMerkle {
     /// Returns a `CreateProposalError::FileIoError` if it encounters an error fetching nodes
     /// from storage, a `CreateProposalError::SendError` if it is unable to send messages to
     /// the workers, and a `CreateProposalError::InvalidConversionToPathComponent` if it is
-    /// unable to convert an u8 index into a path component.
+    /// unable to convert a u8 index into a path component.
     pub fn create_proposal<T: Parentable>(
         &mut self,
         parent: &NodeStore<T, FileBacked>,
--- a/firewood/src/v2/api.rs
+++ b/firewood/src/v2/api.rs
@@ -162,8 +162,8 @@ pub enum Error {
     #[error("send error to worker")]
     SendErrorToWorker,
 
-    // Error converting an u8 index into a path component
-    #[error("error converting an u8 index into a path component")]
+    // Error converting a u8 index into a path component
+    #[error("error converting a u8 index into a path component")]
     InvalidConversionToPathComponent,
 }
 
--- a/storage/src/node/mod.rs
+++ b/storage/src/node/mod.rs
@@ -415,7 +415,7 @@ impl Node {
     ///
     /// # Errors
     ///
-    /// Returns an `Error` if it cannot create a `PathComponent` from an u8 index.
+    /// Returns an `Error` if it cannot create a `PathComponent` from a u8 index.
     pub fn force_branch_for_insert(mut self) -> Result<Box<BranchNode>, Error> {
         // If the `partial_path` is non-empty, then create a branch that will be the new
         // root with the previous root as the child at the index returned from split_first.


firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+    #[must_use]
+    pub fn new() -> Self {
+        ParallelMerkle {
+            workers: [(); BranchNode::MAX_CHILDREN].map(|()| None),
+        }
+    }
+
+    /// Performs the prepare step to allow clean separation of the trie into an array
+    /// of subtries that can be operated on independently by the worker threads.
+    fn prepare_trie(&self, proposal: &mut NodeStore<MutableProposal, FileBacked>) {
```
rkuris: Generalize this code. I can then use the exact same code during proof validation.

To do this, rename this method to "normalize_root" or similar, and call `Node::normalize` except when the root is None) which consumes the node and generates a new one (possibly the same one).
```
impl Node {
    // untested, just for illustration purposes
    fn normalize(self) -> Self {
        if let Some(partial_path) = node.partial_path() {
            pp.split_first().map(|(child_index, child_path)| {
                let mut branch = BranchNode {
                    partial_path: Path::new(),
                    value: None,
                    children: BranchNode::empty_children(),
                };
                node.update_partial_path(child_path);
                branch.update_child(child_index, Some(Child::Node(node)));
                branch
                }
            }
        } else if let Leaf(leaf) = self {
           BranchNode{ ... }
        } else {
            self
        }
    }
}

File: firewood/src/db.rs [hunk 2]
```diff
@@ -111,6 +114,7 @@ pub struct DbConfig {
 pub struct Db {
     metrics: Arc<DbMetrics>,
     manager: RevisionManager,
+    threadpool: OnceLock<ThreadPool>,
```
rkuris: While we currently don't support it, it would be possible to have more than one RevisionManager. Placing it here means we would have one thread pool for all revisions, which I think isn't supported by your code -- they would have to support handling multiple tries.

We might end up doing something like this in the future, so it probably makes sense for now to put this in `RevisionManager` instead of here.

Metrics are (kind of) global regardless of how many RevisionManager instances we have.

firewood/src/merkle/mod.rs
```diff
@@ -582,13 +585,9 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
         self.try_into().expect("failed to convert")
     }
 
-    /// Map `key` to `value` in the trie.
-    /// Each element of key is 2 nibbles.
-    pub fn insert(&mut self, key: &[u8], value: Value) -> Result<(), FileIoError> {
-        let key = Path::from_nibbles_iterator(NibblesIterator::new(key));
-
+    /// Map `key` to `value` in the trie when `key` is a `Path`
+    pub fn insert_path(&mut self, key: Path, value: Value) -> Result<(), FileIoError> {
```
rkuris: Pull this out into a separate PR.

Instead of passing "key" we should be passing in the nibbles iterator. Path is allocated and we can avoid the allocation by just passing in the iterator.

I'm also not sure how many callers to `insert` there are and whether they can and/or should be creating nibbles iterators, could be addressed in a separate PR.

firewood/src/merkle/mod.rs
```diff
@@ -605,6 +604,13 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
         Ok(())
     }
 
+    /// Map `key` to `value` in the trie.
+    /// Each element of key is 2 nibbles.
+    pub fn insert(&mut self, key: &[u8], value: Value) -> Result<(), FileIoError> {
+        let key = Path::from_nibbles_iterator(NibblesIterator::new(key));
+        self.insert_path(key, value)
+    }
+
```
rkuris: Move this before `insert_path`.

firewood/src/merkle/mod.rs
```diff
@@ -744,13 +750,11 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
...
 
-    /// Removes the value associated with the given `key`.
+    /// Removes the value associated with the given `key` where `key` is a `Path`
     /// Returns the value that was removed, if any.
     /// Otherwise returns `None`.
     /// Each element of `key` is 2 nibbles.
-    pub fn remove(&mut self, key: &[u8]) -> Result<Option<Value>, FileIoError> {
-        let key = Path::from_nibbles_iterator(NibblesIterator::new(key));
-
+    pub fn remove_path(&mut self, key: Path) -> Result<Option<Value>, FileIoError> {
```
rkuris: Same comment here as with insert_path

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        //    during the post-processing step.
+        // 3. If the existing root does not have a partial path, then there is nothing we need
+        //    to do if it is a branch. If it is a leaf, then convert it into a branch.
+        //
+        // The result after the prepare step is that there is a branch node at the root with
+        // an empty partial path.
+        let root_node = proposal.root_mut().take();
+        if let Some(mut node) = root_node {
+            // Non-empty root. Check if it has a partial path
+            let index_path_opt: Option<(u8, Path)> = node
```
rkuris: See my snippet above for how to approach this in a rusty way.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        //     value was inserted for the empty key).
+        //     Otherwise, delete the empty root node
+        // If more than one child remains, then do nothing
+        // If only one child remains and the root has a value, then do nothing
+        // If only one child remains and the root doesn’t have a value, then deleted the root,
+        // use the child node as the new root, and update the partial path of the new root to
+        // include its previous child index
+
+        // The root should always be a branch given the prepare step.
+        let branch = new_root.as_branch_mut().expect("not a branch");
```
rkuris: This should get a branch node passed in, and we might be able to reuse this code for when deletes happen. It might be worth thinking about whether this is a method on BranchNode.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// After performing parallel modifications, it may be necessary to perform post processing
+    /// to return the Merkle trie to the correct canonical form.
+    fn postprocess_trie(
+        &self,
+        nodestore: &mut NodeStore<MutableProposal, FileBacked>,
+        mut new_root: Node,
+    ) -> Result<Option<Node>, FileIoError> {
+        // Check if the Merkle trie has an extra root node that was added to facilitate efficient
+        // parallel modification of the trie. If it does, apply transform to return trie to a
+        // valid state by following the steps below:
```
rkuris: This comment is misleading. It's quite possible that we didn't add the extra root node but now we need to collapse it.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        //
+        // If all of the children from the root has been removed, then
+        //     Convert the root into a leaf if it has a value (this should indicate that a
+        //     value was inserted for the empty key).
+        //     Otherwise, delete the empty root node
+        // If more than one child remains, then do nothing
+        // If only one child remains and the root has a value, then do nothing
+        // If only one child remains and the root doesn’t have a value, then deleted the root,
+        // use the child node as the new root, and update the partial path of the new root to
+        // include its previous child index
```
rkuris: I had to read this a few times, and while it's correct, I think it could be written clearer, and perhaps the code should model the comments closer. Here's my quick stab:

```
// If the root node has:
// 0 children and no value, the trie is empty
// 0 children and a value, the root should be a leaf
// 1 child and no value, the child should be the root
// In all other cases, the root is already correct.

File: firewood/src/merkle/parallel.rs [hunk 10]
```diff
@@ -0,0 +1,500 @@
...
+                // The child's partial path is the concatenation of its (now removed) parent, which
+                // should always be empty because of our prepare step, its (former) child index, and
+                // its partial path. Because the parent's partial path should always be empty, we
+                // replace it here with a Path::new().
+                let partial_path = Path::from_nibbles_iterator(
+                    Path::new()
+                        .iter()
+                        .copied()
+                        .chain(once(child_index as u8))
+                        .chain(child.partial_path().iter().copied()),
```
rkuris: Avoid creating the empty Path to start with (untested):
```suggestion
                    once(child_index as u8)
                        .chain(child.partial_path().iter().copied()),
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+                );
+
+                match child {
+                    Node::Branch(ref mut child_branch) => {
+                        child_branch.partial_path = partial_path;
+                    }
+                    Node::Leaf(ref mut leaf) => {
+                        leaf.partial_path = partial_path;
+                    }
+                }
```
rkuris:
```suggestion
                child.update_partial_path(partial_path);
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        pool: &ThreadPool,
+        proposal: &NodeStore<MutableProposal, FileBacked>,
+        root_branch: &mut Box<BranchNode>,
+        first_nibble: u8,
+        worker_sender: Sender<Response<FileBacked>>,
+    ) -> Result<WorkerState, FileIoError> {
+        // Create a channel for the coordinator (main thread) to send messages to this worker.
+        let child_channel = mpsc::channel();
+
+        // Get the child node from the proposal
```
rkuris: This comment could be more valuable:
```suggestion
        // The root's child becomes the root node of the worker
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        let child = root_branch
+            .children
+            .get_mut(first_nibble as usize)
+            .expect("index error")
+            .take();
```
rkuris: Consider moving this to `BranchNode`. I'm not sure if it's very reusable; we should look and see if we do this elsewhere. It's similar to `BranchNode::child`, and could be called `BranchNode::take_child` or something.

storage/src/nodestore/mod.rs
```diff
@@ -251,6 +251,14 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
         self.kind.deleted.push(node);
     }
 
+    /// Take deleted nodes from child nodestore
+    pub fn add_deleted_nodes_from_child(
+        &mut self,
+        child_nodestore: &mut NodeStore<MutableProposal, S>,
+    ) {
+        self.kind.deleted.append(&mut child_nodestore.kind.deleted);
+    }
```
rkuris: This seems overly complicated -- lots of `mut` and it requires one nodestore to understand the structure of the other.

I think all we're doing here is extending `self.kind.deleted` with `child_nodestore.kind.deleted`, which can be done with `Vec::extend` or `Vec::extend_from_slice`

One approach:

```suggestion
pub fn delete_nodes(&mut self, nodes: &[MaybePersistedNode]) {
    self.kind.deleted.extend_from_slice(nodes)
}
```

The one caller then just does:
```
proposal.delete_nodes(&child_nodestore.kind.deleted);
```

storage/src/nodestore/mod.rs
```diff
@@ -270,6 +278,45 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
     }
 }
 
+impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
+    /// Creates a new [`NodeStore`] from a child node.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`FileIoError`] if the child node cannot be read.
+    pub fn from_child(
```
rkuris: Nodestores understanding Child seems non-intuitive. The calculation of the root and the creation of the Nodestore should be separated as well, and this function really shouldn't be doing any I/O.

Maybe this is `from_root(parent: ..., root: Node)` and the caller does the I/O.

storage/src/nodestore/mod.rs
```diff
@@ -44,10 +44,10 @@ pub(crate) mod header;
 pub(crate) mod persist;
 pub(crate) mod primitives;
 
-use crate::firewood_gauge;
 use crate::linear::OffsetReader;
 use crate::logger::trace;
 use crate::node::branch::ReadSerializable as _;
+use crate::{Child, firewood_gauge};
```
rkuris: A nodestore understanding a child seems unnecessarily circular. See below.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+    /// by the value of the `first_nibble`.
+    fn create_worker(
+        pool: &ThreadPool,
+        proposal: &NodeStore<MutableProposal, FileBacked>,
+        root_branch: &mut Box<BranchNode>,
+        first_nibble: u8,
+        worker_sender: Sender<Response<FileBacked>>,
+    ) -> Result<WorkerState, FileIoError> {
+        // Create a channel for the coordinator (main thread) to send messages to this worker.
+        let child_channel = mpsc::channel();
```
rkuris: name these rather than leaving them as tuples:
```suggestion
        let (child_sender, child_receiver) = mpsc::channel();
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        // Build a nodestore from the child node
+        let worker_nodestore = NodeStore::from_child(proposal, child)?;
+
+        // Spawn a worker from the threadpool for this nibble. The worker will send messages to the coordinator
+        // using `worker_sender`.
+        pool.spawn(move || {
+            let mut merkle = Merkle::from(worker_nodestore);
+            loop {
+                // Wait for a message on the receiver child channel. Break out of loop if there is an error.
+                let request = match child_channel.1.recv() {
```
rkuris: This made me scratch my head as to why this worked. The compiler was smart enough to realize that the only thing being moved here is `child_channel.1` leaving `child_channel` is a "partially moved" state. See the comment above on how to fix this.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+};
+use rayon::{ThreadPool, ThreadPoolBuilder};
+use std::iter::once;
+use std::sync::mpsc::{Receiver, Sender};
+use std::sync::{Arc, OnceLock, mpsc};
+
+#[derive(Debug)]
+struct WorkerState {
+    sender: mpsc::Sender<Request>,
+}
```
rkuris: This is just a newtype pattern I think:
```suggestion
struct WorkerSender(mpsc::Sender<Request>);

impl std::ops::Deref for WorkerSender {
    type Target = mpsc::Sender<Request>;
    
    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```
Then, if you have a WorkerSender you can just call send on it (no reason to name sender).
    

```

File: firewood/src/merkle/parallel.rs (hunk 1 above)
Comment (bernard-avalabs) [reply firewood/src/merkle/parallel.rs]: I added a normalize_for_insert in Node, and renamed prepare_trie to normalize_root.

File: firewood/src/db.rs (hunk 2 above)
Comment (bernard-avalabs) [reply firewood/src/db.rs]: I've moved threadpool to RevisionManager.

File: firewood/src/merkle/mod.rs (hunk 3 above)
Comment (bernard-avalabs) [reply firewood/src/merkle/mod.rs]: I've changed insert_path to insert_from_iterator that takes a NibbleIterator. I've also added a TODO to describe possible future optimization using rayon scoped threads to allow a NibbleIterator to be directly passed from the main thread to a worker thread.

File: firewood/src/merkle/parallel.rs (hunk 8 above)
Comment (bernard-avalabs) [reply firewood/src/merkle/parallel.rs]: I've updated the comment.

File: firewood/src/merkle/parallel.rs (hunk 9 above)
Comment (bernard-avalabs) [reply firewood/src/merkle/parallel.rs]: I've updated the comment.

File: firewood/src/merkle/parallel.rs (hunk 13 above)
Comment (bernard-avalabs) [reply firewood/src/merkle/parallel.rs]: I've created a take_child method in BranchNode.

File: storage/src/nodestore/mod.rs (hunk 14 above)
Comment (bernard-avalabs) [reply storage/src/nodestore/mod.rs]: I've added deleted_nodes and deleted_as_slice to NodeStore.

File: storage/src/nodestore/mod.rs (hunk 15 above)
Comment (bernard-avalabs) [reply storage/src/nodestore/mod.rs]: I've added from_root and moved the I/O to the caller.

File: storage/src/nodestore/mod.rs (hunk 16 above)
Comment (bernard-avalabs) [reply storage/src/nodestore/mod.rs]: I've eliminated this dependency.

**Comment by rkuris** [comment]
This needs the lockfile updated. Please see https://github.com/ava-labs/firewood/pull/1321 for some instructions as this changed fairly recently.

File: firewood/src/merkle/parallel.rs [hunk 20]
```diff
@@ -0,0 +1,458 @@
...
+                    })))
+                } else {
+                    Ok(None) // No value. Just delete the root
+                }
+            }
+            Some((child_index, child)) => {
+                // Check if the root has a value or if there is more than one child. If yes, then
+                // just return the root unmodified
+                if branch.value.is_some() || children_iter.next().is_some() {
+                    return Ok(Some((*branch).into()));
```
rkuris: This unboxes and boxes the branch, which means allocate, copy, free. Use a move instead, which might mean you need a new method on Node (maybe `impl From<Box<BranchNode>>>`) or just `Node::Branch(branch)` might work.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+        // Build a nodestore from the child node
+        let worker_nodestore = NodeStore::from_root(proposal, child_root);
+
+        // Spawn a worker from the threadpool for this nibble. The worker will send messages to the coordinator
+        // using `worker_sender`.
+        pool.spawn(move || {
+            let mut merkle = Merkle::from(worker_nodestore);
+
+            // Wait for a message on the receiver child channel. Break out of loop if there is an error.
+            while let Ok(request) = child_receiver.recv() {
```
rkuris: Closing the sender without sending Done should be caught here, at least a log entry.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+                    return Err(err); // Early termination.
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Get a worker from the worker pool based on the `first_nibble` value. Create a worker if
+    /// it doesn't exist already.
+    fn get_worker(
```
rkuris: nit: rust avoids naming things "get". Just `worker` would be fine here.

Why this isn't caught in clippy is a good question.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    /// Removes all of the entries in the trie. For the root entry, the value is removed but the
+    /// root itself will remain. An empty root will only be removed during post processing.
+    fn remove_all_entries(&self, root_branch: &mut Box<BranchNode>) {
+        for worker in self.workers.iter().flatten() {
+            worker
+                .send(Request::DeleteRange {
+                    prefix: Box::default(), // Empty prefix
+                })
+                .expect("TODO: handle error");
```
rkuris: Let's handle this error better than panic.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+        parent: &NodeStore<T, FileBacked>,
+        batch: impl IntoIterator<IntoIter: KeyValuePairIter>,
+        threadpool: &OnceLock<ThreadPool>,
+    ) -> Result<Arc<NodeStore<Arc<ImmutableProposal>, FileBacked>>, FileIoError> {
+        // Get (or create) a threadpool
+        let pool = threadpool.get_or_init(|| {
+            ThreadPoolBuilder::new()
+                .num_threads(BranchNode::MAX_CHILDREN)
+                .build()
+                .expect("Error in creating threadpool")
```
rkuris: Error handling can be better here, maybe:
```suggestion
                .build()?
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+        // Prepare step: normalize the root in preparation for performing parallel modifications
+        // to the trie.
+        self.normalize_root(&mut proposal);
+
+        let mut root_branch = proposal
+            .root_mut()
+            .take()
+            .expect("Should have a root node after prepare step")
+            .into_branch()
+            .expect("Root should be a branch after prepare step");
```
rkuris: This raises some concerns about the type system but I don't have a good easy solution.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+            //       Box<[u8]> to the worker if we use rayon scoped threads. This change would
+            //       eliminate a memory copy but may require some code refactoring.
+            match &op {
+                BatchOp::Put { key: _, value } => {
+                    worker
+                        .send(Request::Insert {
+                            key: op.key().as_ref().into(),
+                            value: value.as_ref().into(),
+                        })
+                        .expect("send to worker error");
```
rkuris: These expects should also get cleaned up. We want to avoid panics and instead return errors back. Otherwise, the root cause of the problem will be hard to identify.

Same for the other cases below.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,458 @@
...
+
+        // Post-process step: return the trie to its canonical form.
+        *proposal.root_mut() = self.postprocess_trie(&mut proposal, root_branch)?;
+
+        // Done with these worker senders. Setting the workers to None will allow the next create
+        // proposal from the same ParallelMerkle to reuse the thread pool.
+        self.workers = [(); BranchNode::MAX_CHILDREN].map(|()| None);
+
+        let immutable: Arc<NodeStore<Arc<ImmutableProposal>, FileBacked>> =
+            Arc::new(proposal.try_into().expect("error creating immutable"));
```
rkuris: This can fail, and we should return that error. An example would be a giant node due to too large of a value.

storage/src/node/branch.rs
```diff
@@ -437,6 +437,15 @@ impl BranchNode {
...
     }
 
+    /// Takes the child at the given index.
+    /// Panics if `child_index` >= [`BranchNode::MAX_CHILDREN`].
+    pub fn take_child(&mut self, child_index: u8) -> Option<Child> {
+        self.children
+            .get_mut(child_index as usize)
+            .expect("index error")
+            .take()
+    }
```
rkuris: Ideally, we don't call expect here. Brandon recently made some changes here that might make it easier.

firewood/src/merkle/parallel.rs
bernard-avalabs: I don't believe we can propagate the error this way since the closure in `get_or_init` doesn't return a Result. This will work if we replace `get_or_init` with `get_or_try_init`, but that currently is only available in nightly releases. I've left a comment that documents this.

firewood/src/merkle/parallel.rs
bernard-avalabs: Replaced with `Node::Branch(branch)`

firewood/src/merkle/parallel.rs
bernard-avalabs: From our discussion, I have removed Done messages and instead moved up when the caller closes the channels. A worker will send back its response once the sender closes the channel.

firewood/src/merkle/parallel.rs
bernard-avalabs: Changed to worker.

firewood/src/merkle/parallel.rs
bernard-avalabs: Fixed

firewood/src/merkle/parallel.rs
bernard-avalabs: I've changed normalize_root to return a branch, which is used directly instead of refetching the branch from the proposal.

firewood/src/merkle/parallel.rs
bernard-avalabs: I've added more error handling code here. The most likely reason for receiving a send error is if the worker returned to the thread pool after receiving a `FileIoError`. Therefore, upon receiving a send error, the main thread will go through the response channel to see if there is a `FileIoError` that was sent back from a worker. If there is, then it will propagate the `FileIoError`. Otherwise, it will propagate the `SendError`.

firewood/src/merkle/parallel.rs
bernard-avalabs: Fixed

storage/src/node/branch.rs
bernard-avalabs: I've added a comment here to use u4 once it becomes available.

firewood/src/merkle/parallel.rs
rkuris: Looks like there might be a bug here. Calling ThreadPoolBuilder twice without terminating the threadpool will result in a panic, and the oncelock you're using will be recreated if the database is closed and then reopened, or if there is a second database opened at the same time.

I think this means if you do: create, insert, close, open, insert, it will fail. Can you please add a test that does that? Currently the reopen test we have doesn't do another insert after reopening the database.

Also please test open-db1, insert, open-db2, insert with parallel turned on.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        Self::new()
+    }
+}
+
+impl ParallelMerkle {
+    /// Default constructor
+    #[must_use]
+    pub fn new() -> Self {
+        ParallelMerkle {
+            workers: [(); BranchNode::MAX_CHILDREN].map(|()| None),
```
rkuris: There are less verbose ways of doing this. I would have used https://doc.rust-lang.org/stable/std/array/fn.from_fn.html

After `use std::array::from_fn;`:
```suggestion
            workers: from_fn(|_| None),
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+impl Default for ParallelMerkle {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ParallelMerkle {
+    /// Default constructor
+    #[must_use]
+    pub fn new() -> Self {
```
rkuris: Why are there two public ways of initializing ParallelMerkle? Let's just use Default::default().

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        ParallelMerkle {
+            workers: [(); BranchNode::MAX_CHILDREN].map(|()| None),
+        }
+    }
+
+    /// Normalize the root to allow clean separation of the trie into an array of subtries that
+    /// can be operated on independently by the worker threads.
+    fn normalize_root(
```
rkuris: nit: naming: maybe `force_root` or something more creative?

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        if let Some(node) = root_node {
+            return node.normalize_for_insert();
+        }
+        // Empty trie. Create a branch node with an empty partial path and a None for a value.
+        BranchNode {
+            partial_path: Path::new(),
+            value: None,
+            children: BranchNode::empty_children(),
+        }
+        .into()
```
rkuris: This is an if-let-else or a `map_or_else`

See https://doc.rust-lang.org/book/ch06-03-if-let.html and https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else

```suggestion
        if let Some(node) = root_node {
            node.normalize_for_insert()
        } else {
            // Empty trie. Create a branch node with an empty partial path and a None for a value.
            BranchNode {
                partial_path: Path::new(),
                value: None,
                children: BranchNode::empty_children(),
            }
            .into()
        }
```

or

```
    node.map_or_else(BranchNode{...}.into(), normalize_for_insert)
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+        mut branch: Box<BranchNode>,
+    ) -> Result<Option<Node>, FileIoError> {
+        // Check if the Merkle trie has an extra root node. If it does, apply transform to
+        // return trie to a valid state by following the steps below:
+        //
+        // If the root node has:
+        // 0 children and no value, the trie is empty. Just delete the root.
+        // 0 children and a value (from an empty key), the root should be a leaf
+        // 1 child and no value, the child should be the root (need to update partial path)
+        // In all other cases, the root is already correct.
```
rkuris: Move this to the doc block around line 120 instead of just a regular comment. If we do generate docs for this, it would be nice to have the explanation in the doc comment.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,500 @@
...
+
+        let mut children_iter = branch
+            .children
+            .iter_mut()
+            .enumerate()
+            .filter_map(|(index, child)| child.as_mut().map(|child| (index, child)));
+
+        let first_child = children_iter.next();
```
rkuris, line 125: I hit my 30 minute review limit timer here. Please leave this comment and request another review, as I'd like to get this merged, and splitting it now would probably generate too much churn.

firewood/src/merkle/parallel.rs
bernard-avalabs: I'm not sure I see why a panic would occur in this scenario. The threadpool is created on the first call of create_proposal for a given database (or more precisely the revision manager for the db). If the database is closed, then it would drop the OnceLock in the revision manager which in turn would drop the threadpool in the OnceLock. Opening the database again will create a new threadpool if I understand correctly.

I've added the two tests that you have suggested: 
1. Create, insert, close, open, insert
2. Open db1, insert, open db2, insert

Both tests are currently passing.

firewood/src/merkle/parallel.rs
bernard-avalabs: I've removed this constructor, but have made the same change in another part of the code that reinitailizes the workers array.

firewood/src/merkle/parallel.rs
bernard-avalabs: I've removed this constructor.

firewood/src/merkle/parallel.rs
bernard-avalabs: Changed to force_root

firewood/src/merkle/parallel.rs
bernard-avalabs: Changed to use map_or_else

firewood/src/merkle/parallel.rs
bernard-avalabs: Moved this to the doc block.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+                Ok(Some(child))
+            }
+        }
+    }
+
+    /// Call by a worker to processes requests from `child_receiver` and send back a response on
+    /// `response_sender` once the main thread closes the child sender.
+    fn worker_event_loop(
+        mut merkle: Merkle<NodeStore<MutableProposal, FileBacked>>,
+        first_nibble: u8,
```
rkuris: nit: for branch factor 256, this isn't a nibble. Perhaps we should call this `first_path_component`?

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+            None => {
+                if let Some(value) = branch.value.take() {
+                    // There is a value for the empty key. Create a leaf with the value and return.
+                    Ok(Some(Node::Leaf(LeafNode {
+                        value,
+                        partial_path: Path::new(), // Partial path should be empty
+                    })))
+                } else {
+                    Ok(None) // No value. Just delete the root
+                }
```
rkuris: All this could be written more concice with `map`. The optimizer should do the move automatically.
```suggestion
                Ok(branch.value.map(|value| Node::Leaf(LeafNode { value, ..Default::default() })
```

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+    type Target = mpsc::Sender<Request>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+/// A request to the worker.
+#[derive(Debug)]
+enum Request {
```
rkuris: This should be a `BatchOp<Key, Value>`. Then you get the convenience method `key()` which you can use later.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+    fn worker_event_loop(
+        mut merkle: Merkle<NodeStore<MutableProposal, FileBacked>>,
+        first_nibble: u8,
+        child_receiver: Receiver<Request>,
+        response_sender: Sender<Result<Response, FileIoError>>,
+    ) -> Result<(), Box<SendError<Result<Response, FileIoError>>>> {
+        // Wait for a message on the receiver child channel. Break out of loop when the sender has
+        // closed the child sender.
+        while let Ok(request) = child_receiver.recv() {
+            match request {
```
rkuris: If you used `BatchOp` here, you could factor out all the common code below easily:
```suggestion
            let mut nibbles_iter = NibblesIterator::new(&request.key());
            nibbles_iter.next();
            match request {
```

I'm also wondering if we can reuse the code to execute this from somewhere else.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+                    let mut nibbles_iter = NibblesIterator::new(&prefix);
+                    nibbles_iter.next(); // Skip the first nibble
+                    if let Err(err) = merkle.remove_prefix_from_iter(nibbles_iter) {
+                        response_sender.send(Err(err))?;
+                        break; // Stop handling additional requests
+                    }
+                }
+            }
+        }
+        // The main thread has closed the channel. Send back the worker's response.
```
rkuris, line 199: 🏆 this is a lot cleaner IMO

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+    fn worker(
+        &mut self,
+        pool: &ThreadPool,
+        proposal: &NodeStore<MutableProposal, FileBacked>,
+        root_branch: &mut BranchNode,
+        first_nibble: u8,
+        response_sender: Sender<Result<Response, FileIoError>>,
+    ) -> Result<&mut WorkerSender, FileIoError> {
+        // Find the worker's state corresponding to the first nibble which are stored in an array.
+        // Note: Should update to use u4 index to avoid expect when it is available.
```
rkuris: nit: use TODO as this highlights this in the editor later.

firewood/src/merkle/parallel.rs
rkuris: Interesting. I'm guessing this works because the OnceLock is dropped when you close the database. For the case where there are multiple databases open, each will have their own threadpool, which I think is correct.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,485 @@
...
+                // response channel and return that instead.
+                ParallelMerkle::find_fileio_error(&response_receiver)?;
+                return Err(err.into());
+            }
+        }
+
+        // Drop the sender response channel from the parent thread.
+        drop(response_sender);
+
+        // Setting the workers to None will close the senders to the workers. This will casue the
```
rkuris: nit: typo "casue" -> "cause"

firewood/src/merkle/parallel.rs
bernard-avalabs: I've updated the code to use map.

firewood/src/merkle/parallel.rs
bernard-avalabs: I've renamed `first_nibble` to `first_path_component` for the entire file.

firewood/src/merkle/parallel.rs
bernard-avalabs: Moving the common code up to above `match request` causes borrow/ownership problems as `request` is borrowed on line 169 which prevents us from moving `request.value` inside the `BatchOp::Put` arm of the match block. I can fix this by matching on `&request` and cloning `value`, but that requires an extra clone compare to the current solution.

I've left it as is for now. Alternatively, I can add an inner function that creates the nibble iterator and calls next on it, and call that function in each arm of `match request` to reduce code duplication. However, I don't think that is worth doing in this case.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,490 @@
...
+        // Get (or create) a threadpool. Note that OnceLock currently doesn't support
+        // get_or_try_init (it is available in a nightly release). The get_or_init should
+        // be replaced with get_or_try_init once it is available to allow the error to be
+        // passed back to the caller.
+        let pool = threadpool.get_or_init(|| {
+            ThreadPoolBuilder::new()
+                .num_threads(BranchNode::MAX_CHILDREN)
+                .build()
+                .expect("Error in creating threadpool")
+        });
```
rkuris: nit: feels like this should be part of the threadpool() method that was called before calling create_proposal, which would make this function a little more focused on creating the proposal instead of creating the threadpool also.

I'd consider calling this method with `pool` directly. Notably it's also not described in the doc block at the beginning of the function, and it moves the panic up a little.

https://www.goodreads.com/quotes/9395684-functions-should-do-one-thing-they-should-do-it-well

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // Create a proposal from the parent
+        let mut proposal = NodeStore::new(parent)?;
```
rkuris: nit: naming: A "proposal" is well defined in the top level README.md, so perhaps this should be called `mutable_nodestore` or maybe just `nodestore`.
This comment could also be clearer, we're actually creating a mutable nodestore, not a proposal here.

firewood/src/merkle/parallel.rs
```diff
@@ -0,0 +1,490 @@
...
+    /// The parent thread may receive a `SendError` if the worker that it is sending to has
+    /// returned to the threadpool after encountering a `FileIoError`. This function should
+    /// be called after receiving a `SendError` to find and propagate the `FileIoError`.
+    fn find_fileio_error(
+        response_receiver: &Receiver<Result<Response, FileIoError>>,
+    ) -> Result<(), FileIoError> {
+        // Go through the messages in the response channel without blocking to see if we can
+        // find the FileIoError that caused the worker to close the channel, resulting in a
+        // send error. If we can find it, then we propagate the FileIoError.
+        while let Ok(resp) = response_receiver.try_recv() {
```
rkuris: This can be a lot more readable, making it clear we're ignoring the successful responses:
```
for result in response_receiver.try_iter() {
    let _ = result?; // explicitly ignore the successful Response
}
```

Which brings up the next question: why would there ever be something other than an error in the channel at this point?

firewood/src/db.rs
```diff
@@ -555,6 +573,164 @@ mod test {
...
+        }
+        drop(db);
+
+        // Open-db1, insert, open-db2, insert
+        let db1 = TestDb::new();
+        insert_commit(&db1, 1);
+        let db2 = TestDb::new();
+        insert_commit(&db2, 2);
+        let parent1 = db1.manager.current_revision();
+        let parent2 = db2.manager.current_revision();
```
rkuris: nit: you could do this without peeking into the manager, but that's probably more work.

firewood/src/merkle/parallel.rs
bernard-avalabs: It is possible that the other workers have inserted successful Responses ahead of the worker that encountered a FileIoError.

firewood/src/db.rs
bernard-avalabs: I've changed it to `let parent1 = db1.revision(db1.root_hash().unwrap().unwrap()).unwrap();`

firewood/src/v2/api.rs
```diff
@@ -156,6 +157,14 @@ pub enum Error {
...
     #[error(transparent)]
     InvalidRootHash(#[from] firewood_storage::InvalidTrieHashLength),
+
+    // Error sending to worker
+    #[error("send error to worker")]
+    SendErrorToWorker,
+
+    // Error converting an index into a path component
+    #[error("error converting an index into a path component")]
+    InvalidIndexToPathComponent,
```
rkuris: nit: I think this is impossible, but I'm not 100% sure.

storage/src/node/mod.rs
```diff
@@ -402,6 +402,53 @@ impl Node {
...
+    ///
+    /// 1.  If the node has a partial path, then create a new node with an empty partial path
+    ///     and a None for a value. Push down the previous node as a child and return the
+    ///     branch.
+    /// 2.  If the existing node does not have a partial path, then there is nothing we need
+    ///     to do if it is a branch. If it is a leaf, then convert it into a branch.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `Error` if it cannot create a `PathComponent` from a child index.
```
rkuris: Not clear how this can happen.

storage/src/node/mod.rs
```diff
@@ -402,6 +402,53 @@ impl Node {
...
+                .ok_or_else(|| Error::other("invalid child index"))?;
+            *branch.children.get_mut(child_path_component) = Some(Child::Node(self));
+            Ok(branch.into())
+        } else {
+            Ok(match self {
+                Node::Leaf(mut leaf) => {
+                    // Root is a leaf with an empty partial path. Replace it with a branch.
+                    BranchNode {
+                        partial_path: Path::new(),
+                        value: Some(std::mem::take(&mut leaf.value)),
```
rkuris: take isn't needed here. Doesn't leaf go out of scope immediately after? So this is just:
```
value: Some(leaf.value)
```

firewood/src/v2/api.rs
bernard-avalabs: I can change that to "error converting a u8 index to a path component".

storage/src/node/mod.rs
bernard-avalabs: The error is from converting a u8 index to a path component. I'll update this.

PR #1259, opened 2025-09-04
fix: Path iterator returned wrong node

rkuris:
Enabled all proof tests, which discovered an introduced bug when MaybePersisted was introduced. The node being returned was the wrong one, which caused some proof errors when constructing proofs, notably the wrong node was being returned if the node wasn't yet persisted.

Since the tests run with unpersisted but hashed nodes, this caused some failures when the tests were re-enabled.

Also restructured the tests. Proof tests are separated now from range proof tests.

Range proof tests were moved but not enabled in this PR (will be done next).


firewood/src/proof.rs
```diff
@@ -89,6 +89,33 @@ pub struct ProofNode {
...
 }
 
+impl std::fmt::Debug for ProofNode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let key: Vec<u8> = self.key().collect();
+        let children: Vec<(usize, TrieHash)> = self
+            .children()
+            .iter()
+            .enumerate()
+            .filter_map(|(i, c)| c.as_ref().map(|h| (i, (*h).clone().into_triehash())))
```
demosdemon: I think this should probably be `IntoHashType::into_hash_type()` instead of `into_triehash`. We've sort of make TrieHash specific to storage and HashType the generic for the node id. 

But, it already is a `HashType` so probably don't need it at all. 

`Clone` in a debug impl is also a red flag. everything should work by reference.

demosdemon (approved):

firewood/src/proof.rs
```diff
@@ -89,6 +89,33 @@ pub struct ProofNode {
...
+        ds.field("key_nibbles", &key)
+            .field("value_digest", &value_digest)
+            .field("children", &children);
+
+        #[cfg(feature = "ethhash")]
+        ds.field("partial_len", &self.partial_len);
+
+        // Compute the hash and render it as well
+        let hash = firewood_storage::Preimage::to_hash(self);
+        ds.field("hash", &hash);
```
demosdemon, line 113: this is rather expensive for `Debug`. Can we gate this behind `f.alternate()`?

rkuris: It's actually not that bad, since it never recurses; the hash should be very quick, and it's only when you debug print (which should really be never in production). It's just a quick hash calculation, and saved my bacon a few times. I also used debug alternate format to mean "prettyprint" which also helped, so I'd hate to overload that.

firewood/src/merkle/tests/proof.rs
```diff
@@ -0,0 +1,271 @@
...
+    debug!("{}", merkle.dump_to_string().unwrap());
+    debug!("root_hash: {root_hash:?}");
+
+    // Non-existent key under the same parent branch
+    let missing = [0u8, 2u8];
+    let proof = merkle.prove(&missing).unwrap();
+
+    debug!("{proof:#?}");
+    assert_eq!(
+        proof.as_ref().first().unwrap().to_hash(),
```
demosdemon, line 226: test fails here with ethhash cause this is too small to be hashed into the triehash:
```suggestion
        TrieHash::from(proof.as_ref().first().unwrap().to_hash()),
```

(probably not the best way to fix this)

firewood/src/merkle/tests/proof.rs
```diff
@@ -0,0 +1,271 @@
...
+        root_hash.clone().into_hash_type()
+    );
+
+    // Ensure at least one node in the proof carries a value (proof value present)
+    assert!(proof.as_ref().iter().any(|n| n.value_digest.is_some()));
+
+    // Exclusion should verify with expected None even if proof includes node values
+    proof
+        .verify(missing, Option::<&[u8]>::None, &root_hash)
+        .unwrap();
```
demosdemon, line 236: however, fixing that breaks here because I realized just yesterday that `verify` is using the wrong hash fn

https://github.com/ava-labs/firewood/blob/0098d123e103844e2d7dfc032e911be52da80489/storage/src/hashednode.rs#L114-L126

The three other test failures might be because of the same `verify` reason.

demosdemon: although, `UnexpectedHash` is only ever produced here so that might not be the reason.

https://github.com/ava-labs/firewood/blob/0098d123e103844e2d7dfc032e911be52da80489/firewood/src/proof.rs#L175-L177

demosdemon: Oh, I see. the value is too small to be hashed so we're running into the same issue as above.

I'm not entirely sure how this should be handled. My gut reaction is to make it so that the partialeq/eq impls on HashOrRlp hashes the rlp value if the other value is already hash.

demosdemon:
#1260 addresses my comments here and fixes your test failures

PR #1260, opened 2025-09-05
feat: explicit impl of PartialEq/Eq on HashOrRlp

demosdemon:
Addresses my comments on #1259.


rkuris (approved):
This is great, it will make some of my test writing a lot easier!

PR #1262, opened 2025-09-05
chore: these tests actually work now

rkuris:


demosdemon:
heh, yeah, I fixed these with #1260.

demosdemon (approved):

PR #1268, opened 2025-09-08
fix: use `count` instead of `size_hint`

demosdemon:
`size_hint` is not guaranteed to be accurate and in some cases where we use it, it may not be. This can potentially lead to silent bugs due to incorrect assumptions about the number of elements in an iterator.

The `count` method on iterators _may_ be suboptimal and try enumerating each item in the iterator; however, most iterator implementations are smarter than that and will shortcut evaluating `count` and return a more optimal response. E.g., [`core::slice::Iter`](https://doc.rust-lang.org/stable/src/core/slice/iter.rs.html#1376-1379) or [`alloc::vec::IntoIter`](https://doc.rust-lang.org/stable/src/alloc/vec/into_iter.rs.html#256-259) and most iterator adaptors take advantage of this as well such as [`core::iter::Chain`](https://doc.rust-lang.org/stable/src/core/iter/adapters/chain.rs.html#89-99) and [`core::iter::Copied`](https://doc.rust-lang.org/stable/src/core/iter/adapters/copied.rs.html#86-88).

I also refactored nibbles_to_eth_compact to be clearer and take advantage of newer rust features to remove some redundant code.

### DIFF
--- a/storage/src/hashers/ethhash.rs
+++ b/storage/src/hashers/ethhash.rs
@@ -134,9 +134,6 @@ impl<T: Hashable> Preimage for T {
                             rlp.append(&bytes);
                         }
                     }
-                    Some(ValueDigest::Hash(hash)) => {
-                        rlp.append(&hash);
-                    }
                     None => {
                         rlp.append_empty_data();
                     }
@@ -144,7 +141,6 @@ impl<T: Hashable> Preimage for T {
             } else {
                 match self.value_digest() {
                     Some(ValueDigest::Value(bytes)) => rlp.append(&bytes),
-                    Some(ValueDigest::Hash(hash)) => rlp.append(&hash),
                     None => rlp.append_empty_data(),
                 };
             }
--- a/firewood/src/proof.rs
+++ b/firewood/src/proof.rs
@@ -129,6 +129,7 @@ impl Hashable for ProofNode {
     fn value_digest(&self) -> Option<ValueDigest<&[u8]>> {
         self.value_digest.as_ref().map(|vd| match vd {
             ValueDigest::Value(v) => ValueDigest::Value(v.as_ref()),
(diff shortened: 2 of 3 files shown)


storage/src/hashers/ethhash.rs
```diff
@@ -114,7 +102,7 @@ impl<T: Hashable> Preimage for T {
     }
 
     fn write(&self, buf: &mut impl HasUpdate) {
-        let is_account = self.key().size_hint().0 == 64;
+        let is_account = self.key().count() == 64;
```
rkuris, line 105: Doesn't this consume the iterator? I'm not sure why no tests are failing if that's so.

storage/src/hashers/ethhash.rs
```diff
@@ -94,17 +88,11 @@ impl<T: Hashable> Preimage for T {
...
-                trace!(
-                    "SIZE WAS {1} {0}",
-                    hex::encode(&collector),
-                    self.key().size_hint().0
-                );
-            }
-        }
+        trace!(
+            "SIZE WAS {} {}",
+            self.key().count(),
```
rkuris, line 93: I think this consumes the iterator.

storage/src/hashers/ethhash.rs
demosdemon: You're correct but each call to `key()` returns a unique iterator.

rkuris: Looks like we do this elsewhere anyway.

rkuris:
It does seem kinda silly to call key(), make a nibbles iterator from a set of bytes, then count them, when it would be a lot easier to just look at the raw key. `key` is probably named poorly, should be `key_nibbles_iter` or seomthing.

PR #1271, opened 2025-09-10
feat(proofs)!: add v0 serialization for RangeProofs

demosdemon:


ffi/memory.go
```diff
@@ -348,36 +348,6 @@ func getDatabaseFromHandleResult(result C.HandleResult) (*Database, error) {
 	}
 }
 
-func getRangeProofFromRangeProofResult(result C.RangeProofResult) (*RangeProof, error) {
```
alarso16, line 351: Why move this? Because it's special purpose for proofs?

ffi/proofs_test.go
```diff
@@ -0,0 +1,134 @@
...
+	r.NoError(proof.Free())
+
+	proof, err = db.RangeProof(maybe{true, root}, maybe{true, []byte("key2")}, maybe{true, []byte("key3")}, 10)
+	r.NoError(err)
+	r.NotNil(proof)
+	proof5Bytes, err := proof.MarshalBinary()
+	r.NoError(err)
+	r.NoError(proof.Free())
+
+	r.Equal(proof4Bytes, proof5Bytes)
```
alarso16: I had a giant comment about parametrizing this test typed out before I saw that you are comparing the byte outputs. This does leave me a with a few questions.
* Why are we choosing to compare these proofs? aka why would `proof1Bytes` and `proof4Bytes` possibly be the same?
* Is the serialization format deterministic? I don't think we require that, so I'm not sure we need to test it. 
* What are you testing? The name isn't very specific
If the answer to these is "We can't really test anything correctly yet, wait for a future PR", we might want to restructure the tests to be ready for those future PRs

ffi/proofs_test.go
```diff
@@ -0,0 +1,134 @@
...
+	r := require.New(t)
+	db := newTestDatabase(t)
+
+	// check empty database
+	proof, err := db.RangeProof(nil, nil, nil, 0)
+	r.ErrorIs(err, errEmptyTrie)
+	r.Nil(proof)
+
+	// check for non-existent root
+	proof, err = db.RangeProof(maybe{
```
alarso16: This seems out of scope for a test about serialization

ffi/proofs_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+package ffi
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type maybe struct {
```
alarso16, line 14: Also maybe out of scope for this PR: should we provide some sort of `maybe` implementation for consumers not also depending on avalanchego?

ffi/proofs_test.go
demosdemon: This is testing that requesting a range proof with a root key and without a root key (i.e., using the default root) returns the same proof. Deterministic proofs are a requirement of the proof itself. A range proof is just a view over a subset over the the database therefore must always provide the same result. 

I have definitely overloaded this test with some additional stuff and can split out.

ffi/memory.go
demosdemon: Yeah. This is specific to proofs and the `memory.go` file has grown really large. I started to realize that shoving all of the memory code in there was not a great idea.

Tbh, I meant to move this with the PR that added it. But, I accidentally made that change in a different commit and didn't notice until after I merged.

ffi/proofs_test.go
demosdemon: Perhaps. There is also an [existing repo](https://github.com/pmorelli92/maybe/blob/93d1dbae533897ff87c93cb135f1778ba8b10c02/maybe.go#L19-L25) that implements the same interface. But, if we do provide something, I would rather it exist outside any of the 4 main repos (avalachego, coreth, subnet-evm, and firewood) to avoid any opportunity for cyclic dependencies.

firewood/src/merkle/tests/ethhash.rs
```diff
@@ -83,7 +83,7 @@ fn test_eth_compatible_accounts(
     key_suffixes_and_values: &[(&str, &str)],
     expected_root: &str,
 ) {
-    use sha2::Digest as _;
+    use sha3::Digest as _;
```
rkuris, line 86: Curious why we would use sha3 here. Ethereum avoids it completely, preferring Keccak.

ffi/src/value.rs
```diff
@@ -22,12 +22,65 @@ pub use self::results::{
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[repr(C)]
 pub enum Maybe<T> {
     /// No value present.
     None,
     /// A value is present.
     Some(T),
 }
+
+impl<T> Maybe<T> {
```
rkuris, line 35: Shame we can't just delegate to Option.

firewood/src/proofs/header.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+const _: () = {
+    assert!(size_of::<Header>() == 32);
+};
+
+impl Header {
+    /// Construct a new header for the given proof type.
+    #[must_use]
+    pub(crate) const fn new(proof_type: ProofType) -> Self {
```
rkuris: nit: seems like this is a conversion from `ProofType` to `Header`, so maybe a `From` implementation would have worked fine here.

firewood/src/merkle/tests/ethhash.rs
demosdemon: this is just for the `Digest` trait. I removed the sha2 dependency because it was unused other than here. We could technically replace this with the digest traits library.

ffi/src/value.rs
demosdemon: Yeah, I'm annoyed that option doesn't work over the FFI library because I hated adding this.

firewood/src/proofs/header.rs
demosdemon: Yeah, I intentionally chose not to implement `From` because I didn't want the conversion to be exported outside of the crate. `ProofType` and `Header` are both visible, but I didn't want an API that enabled someone to create a `Header`.

However, in retrospect, that decision was naive because of the bytemuck traits.

PR #1280, opened 2025-09-15
fix: Check return from submit_and_wait

rkuris:
We now decrement the count of remaining items to wait for instead of assuming they all completed.


demosdemon (approved):

demosdemon (changes requested):
requesting changes since the tests are borked

rkuris:
This approach is incorrect. We probably need something like this, but this isn't it.

Issue #1250, opened 2025-08-27
test: `db::test::reopen_test` can race with the flock releasing and fail


demosdemon:
Unable to reproduce on macos with a stress test running for 2 hours. However, linux triggered the race after only 34 seconds.

```console
$ RUST_BACKTRACE=1 cargo nextest run --package=firewood --lib --features ethhash,logger --cargo-profile=maxperf --stress-duration=24h -- --exact db::test::reopen_test
...
────────────
 Stress test iteration 4552 (00:02:05/24:00:00 elapsed so far, 23:57:55 remaining)
        FAIL [   0.034s] [4552] firewood db::test::reopen_test
  stdout ───

    running 1 test
    714f239a2a43648075a0b8e11f358c4583c0633867bc50754afe244475837e88
    test db::test::reopen_test ... FAILED

    failures:

    failures:
        db::test::reopen_test

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 109 filtered out; finished in 0.01s

  stderr ───

    thread 'db::test::reopen_test' panicked at firewood/src/db.rs:807:46:
    called `Result::unwrap()` on an `Err` value: FileIO(FileIoError { inner: Custom { kind: WouldBlock, error: "WouldBlock" }, filename: Some("/tmp/.tmp2YQdrQ/testdb"), offset: 0, context: Some("unable to obtain advisory lock: database may be opened by another instance") })
    stack backtrace:
       0: __rustc::rust_begin_unwind
       1: core::panicking::panic_fmt
       2: core::result::unwrap_failed
       3: core::ops::function::FnOnce::call_once
    note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

  Cancelling due to test failure
  Cancelling [ 00:02:05] ███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████   1/1  : 0 running, 0 passed, 1 failed, 109 skipped      Stress test [   0.036s] iteration 4552: 1 test run: 0 passed, 1 failed, 109 skipped
────────────
     Summary [ 125.399s] 1 test run: 0 passed, 1 failed, 109 skipped
        FAIL [   0.034s] [4552] firewood db::test::reopen_test
error: test run failed
```

[...]

PR #1290, opened 2025-09-18
fix(#1250): add small sleep to reopen test for kthread to shutdown

demosdemon:
Closes: #1250


rkuris (changes requested):
Seems like we need to figure out why we need this especially since it still fails after this change, even at 50ms on some machines.

PR #1292, opened 2025-09-19
chore!: decorate enums and structs with `#[non_exhaustive]`

demosdemon:
This allows us to add new variants to the enums and fields to the relevant structs without it being a breaking change in the future.


firewood/src/v2/batch_op.rs
```diff
@@ -3,14 +3,17 @@
 
 use crate::v2::api::{KeyType, ValueType};
 
-/// A key/value pair operation. Only put (upsert) and delete are
-/// supported
+/// A key/value pair operation.
+///
+/// Put (upsert), Delete (single key), or Prefix Delete (range) are supported.
+#[non_exhaustive]
```
alarso16: nit: everything else had `#[derive]` first.

alarso16 (approved):

PR #1300, opened 2025-09-23
fix: incorrect gauge metrics

rkuris:
The metrics package is very strange in that gauges are stored as f64 but in a u64 container. Clueless as to why, but you must use from_bits to get the actual float value.

The metrics are reporting 4629700416936869888 as the max_revision count, which is actually 32:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=80c05d7b7f47eb7f459382e2f9950c10


AminR443:
Something weird is now going on for `firewood.nodes.unwritten`. Before your change:
image
After your change:
image
Why is it negative?

rkuris:
> Why is it negative?

Different bug.

PR #1303, opened 2025-09-25
feat: Parallel hashing of Merkle trie

bernard-avalabs:
This PR builds on https://github.com/ava-labs/firewood/pull/1258 to support parallel hashing. The workers performing parallel inserts will now perform hashing on their subtrie when they receive the Done message from the main thread. This is safe even if the trie is modified after post-processing because there is only one case in which any of the subtries is modified during post-processing (the root only has one child and no value). For this case, the updated root will be rehashed.

The public function hash_subtrie_with_index has been added in nodestore/hash.rs to allow the workers to correctly hash their subtrie by accounting for their child index.

The test fuzz_checker has been modified to randomly use either propose or propose_parallel to check the correctness of both hashing approaches (serial and parallel).

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -778,7 +778,13 @@ mod test {
                 }
                 batch
             });
-            let proposal = db.propose(batch).unwrap();
+
+            // randomly uses either propose or propose_parallel
+            let proposal = if rng.random() {
+                db.propose(batch).unwrap()
+            } else {
+                db.propose_parallel(batch).unwrap()
+            };
             proposal.commit().unwrap();
 
             // check the database for consistency, sometimes checking the hashes
--- a/firewood/src/merkle/parallel.rs
+++ b/firewood/src/merkle/parallel.rs
@@ -37,7 +37,7 @@ enum Request {
 #[derive(Debug)]
 enum Response<S> {
     // return the new root of the subtrie at the given nibble
-    Root(u8, Box<NodeStore<MutableProposal, S>>),
+    Root(u8, Box<NodeStore<MutableProposal, S>>, Option<Child>),
     Error(FileIoError),
 }
 
@@ -227,8 +227,33 @@ impl ParallelMerkle {
                     }
                     // Sent from the coordinator to the workers to signal that the batch is done.
                     Request::Done => {
+                        let root = merkle.nodestore.root_mut();
+                        let Some(root_node) = std::mem::take(root) else {
+                            worker_sender
+                                .send(Response::Root(
+                                    first_nibble,
+                                    merkle.into_inner().into(),
+                                    None,
+                                ))
+                                .expect("send from worker error");
+                            break; // Allow the worker to return to the thread pool.
+                        };
+
+                        let (root_node, root_hash, _unwritten_count) =
+                            NodeStore::<MutableProposal, FileBacked>::hash_helper(root_node)
+                                .expect("TODO");
+
+                        // Maybe add root_node, root_hash to the Response.  (Return a child. The child is added to the children
+                        // array to the root node by the main thread)
+
+                        let hashed_root = Some(Child::MaybePersisted(root_node, root_hash));
+
                         worker_sender
-                            .send(Response::Root(first_nibble, merkle.into_inner().into()))
+                            .send(Response::Root(
+                                first_nibble,
+                                merkle.into_inner().into(),
+                                hashed_root,
+                            ))
                             .expect("send from worker error");
                         break; // Allow the worker to return to the thread pool.
                     }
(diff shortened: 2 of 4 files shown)


firewood/src/db.rs
```diff
@@ -778,7 +778,13 @@ mod test {
...
                 batch
             });
-            let proposal = db.propose(batch).unwrap();
+
+            // randomly uses either propose or propose_parallel
+            let proposal = if rng.random() {
+                db.propose(batch).unwrap()
+            } else {
+                db.propose_parallel(batch).unwrap()
+            };
```
rkuris: Did we decide to always use parallel mode? If not, we probably have a lot more work to do to support both.

nit: This seems unrelated to the hashing, probably should be a separate PR.

firewood/src/merkle/parallel.rs
```diff
@@ -227,8 +227,37 @@ impl ParallelMerkle {
...
+                        let hashed_result = merkle
+                            .nodestore
+                            .root_mut()
+                            .take()
+                            .map(|root| {
+                                #[cfg(not(feature = "ethhash"))]
+                                let (root_node, root_hash) =
+                                    NodeStore::<MutableProposal, FileBacked>::hash_helper(
+                                        root,
+                                        Path::from_nibbles_iterator(once(first_nibble)),
```
rkuris: I think you can use the `impl<T: AsRef<[u8]>> From<T> for Path` instead of making an interator just to construct the path (untested):
```suggestion
                                        Path::from(&[first_nibble]),
```

(also in the ethhash case)

firewood/src/merkle/parallel.rs
```diff
@@ -227,8 +227,37 @@ impl ParallelMerkle {
                     }
                     // Sent from the coordinator to the workers to signal that the batch is done.
                     Request::Done => {
+                        // Hash this subtrie and return the root as a Child::MaybePersisted.
+                        let hashed_result = merkle
+                            .nodestore
```
rkuris: nit: I think we're done with `merkle` here so we can probably just call `Merkle::into_inner()` to consume it and just work with the resulting nodestore:
```suggestion
                        let nodestore = merkle.into_inner();
                        let hashed_result = nodestore
```

This saves you from dereferencing the merkle again later.

firewood/src/merkle/parallel.rs
```diff
@@ -35,9 +35,9 @@ enum Request {
 }
 
 #[derive(Debug)]
-enum Response<S> {
-    // return the new root of the subtrie at the given nibble
-    Root(u8, Box<NodeStore<MutableProposal, S>>),
+enum Response {
+    // return both the new root of the subtrie at the given nibble and the deleted nodes.
+    Root(u8, Option<Child>, Vec<MaybePersistedNode>),
```
rkuris: I think we can make this more rusty, as seems to be getting a bit messy. I don't love tuples with so many items as they are harder to read and easier to mix up.

`Response` also seems to just be a `Result` with the `Ok` variant containing the results.

So, I'd consider rewriting response to:
```
struct Response {
    nibble: u8,
    root: Option<Child>,
    deleted_nodes: Vec<MaybePersistedNode>,
}
```

Then, construct the `Sender` with `Sender<Result<Response, FileIoError>>`

Apologies since this should have been caught in the other parallel PR.

storage/src/nodestore/mod.rs
```diff
@@ -247,10 +247,9 @@ impl<S: ReadableStorage> NodeStore<MutableProposal, S> {
         self.kind.deleted.push(node);
     }
 
-    /// Return nodes that have been maked as deleted in this proposal as a slice.
-    #[must_use]
-    pub const fn deleted_as_slice(&self) -> &[MaybePersistedNode] {
-        self.kind.deleted.as_slice()
+    /// Take the nodes that have been marked as deleted in this proposal.
+    pub fn take_deleted_nodes(&mut self) -> Vec<MaybePersistedNode> {
+        std::mem::take(&mut self.kind.deleted)
```
rkuris: nit: `use std::mem::take` then just:
```suggestion
        take(&mut self.kind.deleted)
```

rkuris (changes requested):

firewood/src/merkle/parallel.rs
bernard-avalabs: Thanks for the feedback. I've made this change.

firewood/src/merkle/parallel.rs
bernard-avalabs: I've made this change.

firewood/src/merkle/parallel.rs
bernard-avalabs: I've made this change.

firewood/src/db.rs
bernard-avalabs: From our conversation, I've removed this change from this PR. Once we decide whether to replace propose with propose_parallel, we can determine if we need to add additional tests.

PR #1308, opened 2025-09-25
chore: Add support for enormous disk

rkuris:
I need to test firewood boostrapping when we don't delete anything so I added support for an instance with ~13T of disk space.

### DIFF
--- a/benchmark/setup-scripts/build-environment.sh
+++ b/benchmark/setup-scripts/build-environment.sh
@@ -10,7 +10,7 @@ fi
 apt upgrade -y
 
 # install the build dependency packages
-pkgs=(git protobuf-compiler build-essential apt-transport-https net-tools zfsutils-linux)
+pkgs=(git protobuf-compiler build-essential apt-transport-https net-tools zfsutils-linux mdadm)
 install_pkgs=()
 for pkg in "${pkgs[@]}"; do
   if ! dpkg -s "$pkg" > /dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


demosdemon (approved):

benchmark/setup-scripts/build-environment.sh
```diff
@@ -21,15 +21,52 @@ if [ "${#install_pkgs[@]}" -gt 0 ]; then
...
+  else
+    echo "Unsupported number of NVMe devices: ${#NVME_DEVS[@]}. Using first device only."
+    DEVICE_TO_USE="${NVME_DEVS[0]}"
+  fi
+
+  # Wait for RAID array to be ready (if created)
+  if [[ "$DEVICE_TO_USE" == "/dev/md0" ]]; then
+    echo "Waiting for RAID array to be ready..."
+    while [ ! -e "$DEVICE_TO_USE" ]; do
+      sleep 1
```
aaronbuchwald, line 57: Is this meant to wait for the RAID array to be fully assembled? I think that it will be ready almost immediately, but still in a repair state for a potentially long time.

When I was setting this up for the first time, I found that it was available to use quickly, but still being re-configured and the re-configuration process rate limits itself to ensure that it can be used more quickly, but may cause a difference in performance.

When running with `i8ge.12xlarge`, I found this command helped decrease that time from ~24hrs to ~1hr:

```bash
echo "dev.raid.speed_limit_min = 500000" | sudo tee -a /etc/sysctl.conf
echo "dev.raid.speed_limit_max = 5000000" | sudo tee -a /etc/sysctl.conf
```

rkuris: This isn't a great idea. No point increasing the rate since then it might actually slow down the load. Firewood wasn't saturating the disk anyway but leaving those disks with some headroom means the copy won't slow down the load.

The "repair" is just copying one of the slices to the other one.

aaronbuchwald: Got it, so it's preferred to leave the repair rate limited and run the benchmark simultaneously (minimal or no interference at the default rate) vs. try to accelerate it.

PR #1319, opened 2025-09-26
feat: Add nix flake for ffi

maru-ava:
## Why this should be merged 

A nix flake for the firewood ffi library enables deterministic builds both for [firewood-go-ethhash](http://github.com/ava-labs/firewood-go-ethhash) and a future nix build of avalanchego. 

## How this works

- Adds flake for ffi crate
- Updates attach-static-libs workflow to build static libs with the flake

## How was this tested

 - CI (new ffi-nix job)
 - Locally
   - Install nix: 
     - `curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install`
   - `cd ffi`
   - Build: 
     - `nix build .#firewood-ffi`
   - Test (assuming build step perfomed) outside of a nix dev shell: 
     - `GOLANG="nix run $PWD#go"`
       - Runs the version of golang referenced by the ffi flake
       - Need to capture this before changing directories to `result/ffi` because `result` is a nix store symlink so `../../` won't resolve to the `ffi` path containing the flake
       - Running golang this way instead of with `nix develop` validates usage of the build output won't require `nix develop`'s shell magic  
     - `cd result/ffi`
     - `GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=ethhash ${GOLANG} test ./... -v`

### DIFF
--- a/ffi/test-build-equivalency.sh
+++ b/ffi/test-build-equivalency.sh
@@ -40,18 +40,19 @@ fi
 echo ""
 echo "=== Relocation Count Comparison ==="
 
-# Detect OS and use appropriate tool
+# Determine os-specific reloc config
 if [[ "$OSTYPE" == "darwin"* ]]; then
     # macOS
-    otool -rv "$NIX_LIB" > "$TMPDIR/nix-relocs.txt"
-    otool -rv "$CARGO_LIB" > "$TMPDIR/cargo-relocs.txt"
+    RELOC_CMD="otool -rv"
     RELOC_PATTERN='[A-Z_]+_RELOC_[A-Z0-9_]+'
 else
     # Linux
-    readelf -r "$NIX_LIB" > "$TMPDIR/nix-relocs.txt"
-    readelf -r "$CARGO_LIB" > "$TMPDIR/cargo-relocs.txt"
+    RELOC_CMD="readelf -r"
     RELOC_PATTERN='R_[A-Z0-9_]+'
 fi
+
+$RELOC_CMD "$NIX_LIB" > "$TMPDIR/nix-relocs.txt"
+$RELOC_CMD "$CARGO_LIB" > "$TMPDIR/cargo-relocs.txt"
 
 NIX_RELOCS=$(wc -l < "$TMPDIR/nix-relocs.txt")
 CARGO_RELOCS=$(wc -l < "$TMPDIR/cargo-relocs.txt")
@@ -67,13 +68,8 @@ echo ""
 echo "=== Relocation Type Comparison ==="
 
 # Use grep with -E for better portability (avoid -P which isn't available on macOS)
-if [[ "$OSTYPE" == "darwin"* ]]; then
-    grep -Eo "$RELOC_PATTERN" "$TMPDIR/nix-relocs.txt" | sort | uniq -c > "$TMPDIR/nix-reloc-types.txt"
-    grep -Eo "$RELOC_PATTERN" "$TMPDIR/cargo-relocs.txt" | sort | uniq -c > "$TMPDIR/cargo-reloc-types.txt"
-else
-    grep -oP "$RELOC_PATTERN" "$TMPDIR/nix-relocs.txt" | sort | uniq -c > "$TMPDIR/nix-reloc-types.txt"
-    grep -oP "$RELOC_PATTERN" "$TMPDIR/cargo-relocs.txt" | sort | uniq -c > "$TMPDIR/cargo-reloc-types.txt"
-fi
+grep -Eo "$RELOC_PATTERN" "$TMPDIR/nix-relocs.txt" | sort | uniq -c > "$TMPDIR/nix-reloc-types.txt"
+grep -Eo "$RELOC_PATTERN" "$TMPDIR/cargo-relocs.txt" | sort | uniq -c > "$TMPDIR/cargo-reloc-types.txt"
 
 if diff "$TMPDIR/nix-reloc-types.txt" "$TMPDIR/cargo-reloc-types.txt" > /dev/null; then
     echo "✅ Relocation types match"


maru-ava:
> 1. Is there any change to the release process? The RELEASE.md file wasn't changed if so.

afaik the only change is to the CI job that publishes to firewood-go-ethhash, and I don't see mention of that in RELEASE.md

> 2. When dependency updates happen, are any additional steps required?

The only requirement is to keep Cargo.lock up-to-date, and Brandon already added a CI check to ensure PRs won't merge if it's out-of-sync.

> 3. I noticed all the files in the /nix result directory are dated 1969. Is this expected?

That is expected - it's the unix epoch (i.e. time=0). Nix sets all built outputs to this time to enable reproducibility, content-addressibility and build caching. 

> 4. The object files are about half as big, nearly identical text area bug the "others" section is much smaller. Any idea why?

Assuming this is an apples-to-apples comparison (i.e. maxperf and same build args), I would guess there is a difference in the build tooling that nix uses?

rkuris:
> Assuming this is an apples-to-apples comparison (i.e. maxperf and same build args), I would guess there is a difference in the build tooling that nix uses?

Yes, it was apples to apples. One concern is that it may affect performance. I'll bring this up at standup.

maru-ava:
> > Assuming this is an apples-to-apples comparison (i.e. maxperf and same build args), I would guess there is a difference in the build tooling that nix uses?
> 
> Yes, it was apples to apples. One concern is that it may affect performance. I'll bring this up at standup.

Would it make sense to remove the update to the CI job pending qualification of the build artifact with an appropriate performance test?

rkuris:
Yes, that's a good idea.

maru-ava:
Removed changes to attach-static-libs workflow and added new nix-specific ffi-nix job to ensure coverage.

ffi/flake.nix
```diff
@@ -0,0 +1,119 @@
...
+        src = craneLib.path ./..;
+        filter = path: type:
+          (lib.hasSuffix "\.md" path) ||
+          (craneLib.filterCargoSources path type);
+      };
+
+      commonArgs = {
+        inherit src;
+        strictDeps = true;
+        dontStrip = true;
```
maru-ava, line 51: Nix strips debugging details by default, which was resulting in the .a file being ~6MB instead of ~50MB.

ffi/flake.nix
```diff
@@ -0,0 +1,119 @@
...
+
+        # Build only the firewood-ffi crate
+        pname = ffiCargoToml.package.name;
+        version = workspaceCargoToml.workspace.package.version;
+
+        CARGO_PROFILE = "maxperf";
```
demosdemon: this doesn't appear to be working. the size reduction @rkuris noticed is from fewer inlined methods which tells me there's less LTO. 

~I haven't determined the correct way to set this with `rust-overlay`, but cargo will use the [`PROFILE`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#:~:text=PROFILE%20%E2%80%94%20release,settings%20being%20used.) environment variable if `--profile` is not provided on the command line.~

EDIT: that appears to be incorrect. looks like there is _no_ environment variable that cargo will use to infer the profile and it must be provided in the cli invocation. <https://doc.rust-lang.org/cargo/reference/profiles.html#profile-selection>

demosdemon (changes requested):

ffi/flake.nix
demosdemon: I see `CARGO_PROFILE` is actually a parameter to crane's `mkCargoDerivation` which is wrapped by `buildDepsOnly` and `buildPackage`. So, I'm not entirely sure why this profile doesn't appear to work.

maru-ava: As per the [most recent code comment](https://github.com/ava-labs/firewood/pull/1319#discussion_r2393444387), the reason for the reduction in size was due to nix automatically stripping debugging symbols (which is the default for nix). I've corrected that, and the .a files are now [edit: almost] identical in size.

I also believe that `CARGO_PROFILE` is being respected. This var is for configuring `crane` as per its [api docs](https://crane.dev/API.html) (search for `CARGO_PROFILE`) rather than intended for direct consumption by cargo. This can be verified by the following invocation:

 - `cd ffi && nix build .#firewood-ffi --print-build-logs --rebuild 2>&1 | grep -i "cargo build"`

demosdemon: Ah, I didn't realize the change to not strip came after Ron's comment.

ffi/flake.nix
```diff
@@ -0,0 +1,119 @@
...
+      });
+
+      firewood-ffi = craneLib.buildPackage (commonArgs // {
+        inherit cargoArtifacts;
+        cargoExtraArgs = "--package ${ffiCargoToml.package.name} --features ethhash,logger --frozen";
+
+        # Install the static library and header
+        postInstall = ''
+          mkdir -p $out/lib $out/include
+          cp target/*/libfirewood_ffi.a $out/lib/ || cp target/release/libfirewood_ffi.a $out/lib/
```
demosdemon: the `*` should probably be replaced with `maxperf` as the target directory is deterministic on the profile used. If the resulting artifact isn't in the `maxperf` directory, then we're using the wrong profile.

maru-ava: Done

maru-ava:
fwiw I added a script to validate build equivalency. Not sure whether it makes sense to merge it, I just wanted to have a common reference point of comparison.

maru-ava:
I tried to run the build equivalency script on macos but there are some issues with the macos support that I think should be corrected before merge.

maru-ava:
False alarm on the macos support being an issue, pebkac, but I did move the macos deployment target to ensure consistency between buildDepsOnly and buildPackage and to ensure better compatibility with `--frozen`.

Should be possible for anyone to check for themselves equivalency between the normal cargo build and its nix equivalent by running `./ffi/test-build-equivalency.sh`. Please let me know if you'd like to see this file removed before merge.

rkuris:
> False alarm on the macos support being an issue, pebkac, but I did move the macos deployment target to ensure consistency between buildDepsOnly and buildPackage and to ensure better compatibility with `--frozen`.
> 
> Should be possible for anyone to check for themselves equivalency between the normal cargo build and its nix equivalent by running `./ffi/test-build-equivalency.sh`. Please let me know if you'd like to see this file removed before merge.

Either way is fine with me, but if we keep it, it needs to be tested during CI or when we merge to main or it will go stale.

maru-ava:
> Either way is fine with me, but if we keep it, it needs to be tested during CI or when we merge to main or it will go stale.

I've updated the ffi-nix job to use the `test-build-equivalency.sh` script instead of just building the flake directly, so it both builds the flake and performs some equivalency checks against the non-nix cargo build.

.github/workflows/ci.yaml
```diff
@@ -244,6 +244,20 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20
+      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 #v13
+      - name: Test build equivalency between Nix and Cargo
+        run: bash -x ./ffi/test-build-equivalency.sh
+      - name: Test Go FFI bindings
+        working-directory: ffi
+        # - cgocheck2 is expensive but provides complete pointer checks
+        # - the nix flake builds with `--features ethhash`
```
rkuris: It builds with features ethhash and logger.

ffi/test-build-equivalency.sh
```diff
@@ -0,0 +1,95 @@
...
+echo ""
+echo "=== Relocation Type Distribution ==="
+cat "$TMPDIR/nix-reloc-types.txt"
+
+echo ""
+echo "=== Summary ==="
+if [ "$NIX_SYMBOLS" -eq "$CARGO_SYMBOLS" ] && [ "$NIX_RELOCS" -eq "$CARGO_RELOCS" ] && diff "$TMPDIR/nix-reloc-types.txt" "$TMPDIR/cargo-reloc-types.txt" > /dev/null; then
+    echo "✅ Builds are equivalent - both using maxperf profile"
+else
+    echo "❌ Builds differ"
```
rkuris, line 90: It looks like we just log that they differ and return success anyway, which means someone might not notice there is a difference.

Shouldn't we exit with a different exit status if the builds differ?

rkuris (changes requested):

.github/workflows/ci.yaml
maru-ava: I can add that if you want, the thing I was trying to highlight here was why `TEST_FIREWOOD_HASH_MODE=ethhash` was used instead of `TEST_FIREWOOD_HASH_MODE=firewood`

ffi/test-build-equivalency.sh
maru-ava: Good catch, done.

.github/workflows/ci.yaml
maru-ava: Updated comment, PTAL

maru-ava:
jemalloc appears to be introducing non-deterministic build behavior in CI but not locally. Updating to run `make` invocations serially appears to resolve this.

ffi/test-build-equivalency.sh
```diff
@@ -0,0 +1,96 @@
...
+echo "=== Relocation Type Comparison ==="
+
+# Use grep with -E for better portability (avoid -P which isn't available on macOS)
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    grep -Eo "$RELOC_PATTERN" "$TMPDIR/nix-relocs.txt" | sort | uniq -c > "$TMPDIR/nix-reloc-types.txt"
+    grep -Eo "$RELOC_PATTERN" "$TMPDIR/cargo-relocs.txt" | sort | uniq -c > "$TMPDIR/cargo-reloc-types.txt"
+else
+    grep -oP "$RELOC_PATTERN" "$TMPDIR/nix-relocs.txt" | sort | uniq -c > "$TMPDIR/nix-reloc-types.txt"
+    grep -oP "$RELOC_PATTERN" "$TMPDIR/cargo-relocs.txt" | sort | uniq -c > "$TMPDIR/cargo-reloc-types.txt"
+fi
```
rkuris: nit: Aren't there some grep options that work on both platforms?

otherwise, I would have preferred `GREPFLAGS=-Eo` and `GREPFLAGS=-Op` based on `$OSTYPE` and then combining all those into single statements.

.github/workflows/ci.yaml
```diff
@@ -244,6 +244,25 @@ jobs:
         # cgocheck2 is expensive but provides complete pointer checks
         run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=firewood go test -race ./...
 
+  ffi-nix:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20
+      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 #v13
```
rkuris, line 252: Questions: How and when are these hardcoded shas updated? I'm sure dependabot won't tell us when they update. Should we add something to the release checklist to see if they should be updated?

What happens if someone finds a security vulnerability in these packages?

rkuris (approved):
Good to go, please rebase and merge. Maintenance question should just get answered here. Other stuff are nits that don't need fixing.

.github/workflows/ci.yaml
maru-ava: dependabot can, in fact, tell you to update. I just checked, though, and it's not yet configured for this repo. This is what needs to be in `.github/dependabot.yml`:

```
  - package-ecosystem: github-actions
    directory: "/"
    schedule:
      interval: weekly
    open-pull-requests-limit: 0 # Disable non-security version updates
```

As per https://github.com/ava-labs/avalanchego/pull/3822, specifying SHAs for 3rd party actions (i.e. not provided by github) is best practice to avoid supply chain attack. A high-profile incident earlier this year galvanized awareness of the problem.

maru-ava: Updated dependabot config: https://github.com/ava-labs/firewood/pull/1319/commits/dc0f60bec09754dbd645b5c598db948d6d4def61

ffi/test-build-equivalency.sh
maru-ava: Updated to minimize duplication: https://github.com/ava-labs/firewood/pull/1319/commits/15742d281859a0c13486a9259f6da47f01e75f9f

ffi/test-build-equivalency.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# Use grep with -E for better portability (avoid -P which isn't available on macOS)
+grep -Eo "$RELOC_PATTERN" "$TMPDIR/nix-relocs.txt" | sort | uniq -c > "$TMPDIR/nix-reloc-types.txt"
+grep -Eo "$RELOC_PATTERN" "$TMPDIR/cargo-relocs.txt" | sort | uniq -c > "$TMPDIR/cargo-reloc-types.txt"
+
+if diff "$TMPDIR/nix-reloc-types.txt" "$TMPDIR/cargo-reloc-types.txt" > /dev/null; then
+    echo "✅ Relocation types match"
+else
+    echo "❌ Relocation types differ"
+    diff "$TMPDIR/nix-reloc-types.txt" "$TMPDIR/cargo-reloc-types.txt"
```
rkuris, line 78: nit: would have been nice for an explicit exit after diff. It's magic to know that diff exits with a non-zero exit status due to there being differences.

rkuris (approved):

PR #1321, opened 2025-09-26
ci: include dependency lockfile

demosdemon:
For reproducible builds and in order to write a nix flake, we need to include the lockfile. This is primarily for the ffi library. But, because we ship a binary artifact, we need to include it for all dependencies of the ffi library (including firewood itself).

# `--frozen` vs `--locked`

`--frozen` implies both `--locked` and `--offline`, which is what we want in most cases. However, `--locked` is what we want in all cases. `--offline` (and `--frozen`) do not work for most of the CI actions because the cache utility we use does not include `~/.cargo/registry/src` because cargo can theoretically reconstruct it from all of the `~/.cargo/registry/cache/*.crate` archives. And, it can. But, cargo refuses to do that if you specify `--frozen` or `--offline`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -317,4 +317,4 @@ jobs:
       - name: Install cargo-machete
         run: cargo install cargo-machete
       - name: Check for stale dependencies
-        run: cargo machete --locked --with-metadata
+        run: cargo machete --with-metadata


rkuris (approved):
Question: when does the `Cargo.lock` file get updated?

demosdemon:
> Question: when does the `Cargo.lock` file get updated?

The lock file is updated any time someone:

- Adds a new crate to the workspace
- Adds a new dependency to a crate in the workspace
- Runs `cargo update` which will resolve any semver compatible upgrades (making those changes to `Cargo.toml` that only bump patch versions irrelevant)
  - This is done automatically by rust-analyzer sometimes and will cause some annoyances.

PR #1326, opened 2025-10-02
feat(ffi): implement revision handles and expose to FFI layer

### LINKED ISSUE #1008: Require the FFI to explicitly track revisions

AminR443:
This PR introduces revision handles that hold a revision across the FFI while you use it, so it won’t be removed mid-operation. This should improve concurrency and avoids errors as mentioned in #1008 . It also makes iterator's design simpler, we can create them from a held RevisionHandle with no tricky lifetimes. 

This updates the existing Go bindings so `Database.Revision` returns a Revision with `Get` and `Drop` methods. Necessary changes to tests are applied.
  
Closes #1008.

### DIFF
--- a/firewood/src/merkle/tests/mod.rs
+++ b/firewood/src/merkle/tests/mod.rs
@@ -52,7 +52,7 @@ where
     V: AsRef<[u8]>,
 {
     let memstore = Arc::new(MemStore::new(Vec::with_capacity(64 * 1024)));
-    let base = Merkle::from(NodeStore::new_empty_committed(memstore.clone()).unwrap());
+    let base = Merkle::from(NodeStore::new_empty_committed(memstore.clone()));
     let mut merkle = base.fork().unwrap();
 
     for (k, v) in iter.clone() {
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -995,7 +995,7 @@ mod test {
     // We use primitive calls here to do a low-level check.
     fn checker_traverse_correct_trie() {
         let memstore = MemStore::new(vec![]);
-        let mut nodestore = NodeStore::new_empty_committed(memstore.into()).unwrap();
+        let mut nodestore = NodeStore::new_empty_committed(memstore.into());
 
         let test_trie = gen_test_trie(&mut nodestore);
         // let (_, high_watermark, (root_addr, root_hash)) = gen_test_trie(&mut nodestore);
@@ -1019,7 +1019,7 @@ mod test {
     // This test permutes the simple trie with a wrong hash and checks that the checker detects it.
     fn checker_traverse_trie_with_wrong_hash() {
         let memstore = MemStore::new(vec![]);
-        let mut nodestore = NodeStore::new_empty_committed(memstore.into()).unwrap();
+        let mut nodestore = NodeStore::new_empty_committed(memstore.into());
 
         let mut test_trie = gen_test_trie(&mut nodestore);
         let root_addr = test_trie.root_address;
@@ -1088,7 +1088,7 @@ mod test {
         let rng = crate::SeededRng::from_env_or_random();
 
         let memstore = MemStore::new(vec![]);
-        let mut nodestore = NodeStore::new_empty_committed(memstore.into()).unwrap();
+        let mut nodestore = NodeStore::new_empty_committed(memstore.into());
 
         // write free areas
         let mut high_watermark = NodeStoreHeader::SIZE;
@@ -1134,7 +1134,7 @@ mod test {
     #[test]
     fn traverse_freelist_should_skip_offspring_of_incorrect_areas() {
         let memstore = MemStore::new(vec![]);
-        let mut nodestore = NodeStore::new_empty_committed(memstore.into()).unwrap();
+        let mut nodestore = NodeStore::new_empty_committed(memstore.into());
         let TestFreelist {
             high_watermark,
             free_ranges,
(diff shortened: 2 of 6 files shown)


ffi/src/value/results.rs
```diff
@@ -306,6 +307,52 @@ pub enum ProposalResult<'db> {
...
+    NullHandlePointer,
+    /// The provided root was not found in the database.
+    RevisionNotFound(HashKey),
+    /// Getting the revision was successful and the revision handle is returned
+    Ok {
+        /// An opaque pointer to the [`RevisionHandle`].
+        /// The value should be freed with [`fwd_free_revision`]
+        ///
+        /// [`fwd_free_revision`]: crate::fwd_free_revision
+        handle: Box<RevisionHandle>,
```
demosdemon, line 325: one small thought that I don't have a strong opinion on:

It might be useful to include the hash here like I did for proposals. This allows you to make a combination `get_root` that returns a revision with the root hash at the same time. Also, calling into the FFI layer isn't free. There is some overhead on the Go side. Copying the 32-byte hash on the stack in the return would help avoid that by letting Go hold onto it.

This does pair with my other comment too.

ffi/revision.go
```diff
@@ -13,19 +13,77 @@ import "C"
...
+func getRevisionFromResult(result C.RevisionResult) (*Revision, error) {
+	switch result.tag {
+	case C.RevisionResult_NullHandlePointer:
+		return nil, errDBClosed
+	case C.RevisionResult_RevisionNotFound:
+		return nil, errRevisionNotFound
+	case C.RevisionResult_Ok:
+		body := (*C.RevisionResult_Ok_Body)(unsafe.Pointer(&result.anon0))
+		proposal := &Revision{
+			handle: body.handle,
```
demosdemon, line 89: The `Revision` struct has a `root` field that's supposed to contain the root hash. It appears unused now. So, I would remove it from the struct or set it like proposals do.

ffi/revision.go
```diff
@@ -13,19 +13,77 @@ import "C"
...
 	errRevisionNotFound  = errors.New("revision not found")
 	errInvalidRootLength = fmt.Errorf("root hash must be %d bytes", RootLength)
+	errDroppedRevision   = errors.New("revision already dropped")
 )
 
+// Revision is an immutable view over the database at a specific root hash.
+// Instances are created via Database.Revision, provide read-only access to the revision,
+// and must be released with Drop when no longer needed.
 type Revision struct {
-	database *Database
```
demosdemon, line 24: It's not strictly necessary to hold onto the db handle for revisions because they can technically survive the database. However, I think I would rather keep the handle so that when we add Cleanups they ensure the cleanup for the Revision occurs before the Database. I'll defer to @alarso16 on that though.

alarso16: Do we need to enforce this in go?

alarso16 (changes requested):
See above

ffi/revision.go
demosdemon: It would be helpful to enforce this in go.

Revision manager will be unable to reclaim the revision if there is still an outstanding handle. It also will be unable to reclaim any revision that is after the revision when there's an outstanding handle.

<https://github.com/ava-labs/firewood/blob/8d69d198052b01a67283c6dd4f93728d7c4619f8/firewood/src/manager.rs#L207-L214>

Right now we don't attempt to reclaim free areas on flush but if we did, having this enforced in go would allow us to do the reclaim.

ffi/revision.go
AminR443: I set that field after getting the revision since we already have the root hash and didn't need to return it from Rust.

https://github.com/ava-labs/firewood/blob/4e1398b6cb5004c7b6cbee51eb0f5ed6741fff84/ffi/firewood.go#L230-L233

However, for the latest revision, this requires an extra FFI call to get the root hash. I haven't changed this because the Go code expects failure when passed `nil` for root. We could pass `nil` for latest revision, but I'm unsure of side effects. If there are none, I'll make these changes. @alarso16 what do you think?

https://github.com/ava-labs/firewood/blob/4e1398b6cb5004c7b6cbee51eb0f5ed6741fff84/ffi/firewood.go#L218-L221

alarso16: There are no side effects that I know if (also allows you to get rid of the `LatestRevision` api you had. I don't think the go-ethereum code would use it though

ffi/firewood_test.go
```diff
@@ -821,6 +821,39 @@ func TestRevision(t *testing.T) {
 	}
 }
 
+// Tests that even if a proposal is committed, the corresponding revision will not go away
+// as we're holding on to it
+func TestRevisionOutlivesProposal(t *testing.T) {
```
alarso16, line 830: This is weird behavior, this means that a revision can be a proposal that can't be committed?

ffi/firewood_test.go
```diff
@@ -126,22 +126,25 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func newTestDatabase(t *testing.T) *Database {
+func newTestDatabase(t *testing.T, configureFns ...func(*Config)) *Database {
```
alarso16, line 129: Good solution! I like the optional configure function.

alarso16 (commented):
I still think that we need to document the behavior of this`Revision` - specifically, `Drop` has different implementations, even though they seem to reference the same object. Also, I think we should add to godocs the behavior for this, and when each should be used in the README.md

ffi/firewood_test.go
AminR443: Created #1348 per our discussion.

Issue #1299, opened 2025-09-22
Epic: Archival Node Support


RodrigoVillar:
Currently, nodes utilizing Firewood have access to the latest `N` revisions. Any older revisions are no longer tracked, which is a blocker for archival support as archival nodes need to be query the world state at any given accepted block height.

Implementing this epic requires the following:

- [x] Adding a mock `RootStore` to Firewood (PR #1346  which uses PRs #1351  and #1355 )
- [x] Enabling `go` calls with different rootstores (#1384)
- [x] Adding a production `RootStore` implementation to Firewood (#1432)
- [x] Enable archival support in Coreth (https://github.com/ava-labs/coreth/issues/1387) 
- [x] Add Firewood archival option to reexecution test (https://github.com/ava-labs/avalanchego/pull/4442) 
- [x] Schedule reexecution test with Firewood's archival support (#1386)
- [x] Create archival state at different block heights (#1500)
- [x] Add integration test for Firewood's archival support (#1499)

RodrigoVillar:
Originally, we had planned to `build => test => merge => benchmark` the chosen `RootStore` implementation. However, I think it makes more sense to `build => test => benchmark => merge` for the following reasons:
- We should verify that Firewood with `RootStore` is faster than using `LevelDB`
- We should verify that Firewood with `RootStore` does not fail when running the reexecution tests
- We should see how much disk space `RootStore` takes up when running the reexecution tests

Merging prior to benchmarking would provide reviewers of the `RootStore` PR with context as to what could be improved. One open question for example is the current choice of FjallStore - is FjallStore's ability to persist solely to the OS buffer worth the tradeoff of having to allocate a separate directory?

AFAIK, running the benchmark against a candidate implementation of `RootStore` would require us to push a candidate version of [firewood-go-ethhash](https://github.com/ava-labs/firewood-go-ethhash) so that AvalancheGo could then run a reexecution test in CI.

rkuris:
> AFAIK, running the benchmark against a candidate implementation of `RootStore` would require us to push a candidate version of [firewood-go-ethhash](https://github.com/ava-labs/firewood-go-ethhash) so that AvalancheGo could then run a reexecution test in CI.

This isn't true. We can run re-execution tests on arbitrary versions for testing, it's quite easy, and the instructions are in the benchmark directory.

RodrigoVillar:
With #1557 now merged in, we can march towards completing the remaining issues in this epic.

[...]

PR #1333, opened 2025-10-06
research(rootstore): implementations POC

RodrigoVillar:
A prerequisite of #1299 is choosing the underlying datastore for `RootStore`.  This PR sets to finalize the underlying choice of datastore by providing the following:

- Three implementations of `RootStore` using SQLite, LMDB, and RocksDB. 
- Read and write benchmarks for each implementation
- An initial integration of `RootStore` into both the `RevisionManager` and `DB`

I don't expect this PR to be merged in; rather, this PR complements the archival support design document by giving reviewers an idea of what the final `RootStore` implementation will look like. Once the underlying data store is selected, I will cherry-pick the relevant implementation details from this PR and utilize them in a future PR.

## Tips for Reviewing

If you want to just run the implementation benchmarks, execute the following:

```
cargo bench --bench root_store
```

This will benchmark the read and write performance of each implementation with 1m kv-pairs already inserted. 

For those wishing to understand the file structure of this PR:
- `firewood/benches/root_store.rs`: contains implementation benchmarks
- `firewood/src/root_store.rs`: contains the three implementations of `RootStore`
- `firewood/src/db.rs`: contains a simple unit test for what a correct integration of `RootStore` would look like
- `firewood/src/manager.rs`: a draft of what `RootStore` integration into the `RevisionManager` would look like

### DIFF
--- a/firewood/benches/root_store.rs
+++ b/firewood/benches/root_store.rs
@@ -1,0 +1,139 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use std::collections::HashMap;
+
+use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use firewood::root_store::{LMDBStore, RocksDBStore, RootStore, RootStoreBuilder, SQLiteStore};
+use firewood_storage::{LinearAddress, SeededRng, TrieHash};
+use rand::Rng;
+
+fn bench_lmdb_writes_prepopulated<const NKEYS: usize>(criterion: &mut Criterion) {
+    bench_writes_with_prepopulated_root_store::<LMDBStore>(criterion, NKEYS, "LMDB");
+}
+
+fn bench_rocksdb_writes_prepopulated<const NKEYS: usize>(criterion: &mut Criterion) {
+    bench_writes_with_prepopulated_root_store::<RocksDBStore>(criterion, NKEYS, "RocksDB");
+}
+
+fn bench_sqlite_writes_prepopulated<const NKEYS: usize>(criterion: &mut Criterion) {
+    bench_writes_with_prepopulated_root_store::<SQLiteStore>(criterion, NKEYS, "SQLite");
+}
+
+fn bench_lmdb_reads_prepopulated<const NKEYS: usize>(criterion: &mut Criterion) {
+    bench_reads_with_prepopulated_root_state::<LMDBStore>(criterion, NKEYS, "LMDB");
+}
+
+fn bench_rocksdb_reads_prepopulated<const NKEYS: usize>(criterion: &mut Criterion) {
+    bench_reads_with_prepopulated_root_state::<RocksDBStore>(criterion, NKEYS, "RocksDB");
+}
+
+fn bench_sqlite_reads_prepopulated<const NKEYS: usize>(criterion: &mut Criterion) {
+    bench_reads_with_prepopulated_root_state::<SQLiteStore>(criterion, NKEYS, "SQLite");
+}
+
+// bench_writes_with_prepopulated_root_store benchmarks an instance of T by
+// first prepopulating it with 1 million key-value pairs before benchmarking
+// its write performance
+#[expect(clippy::unwrap_used)]
+fn bench_writes_with_prepopulated_root_store<T: RootStore + RootStoreBuilder<T>>(
+    criterion: &mut Criterion,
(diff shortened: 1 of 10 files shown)


rkuris:
Looks like this might have some FFI complications. If we're going with rocksdb, we probably need to make sure that the C parts of rocksdb are included in the ffi build, and we should probably talk about this in more detail.

firewood/benches/root_store.rs
```diff
@@ -0,0 +1,139 @@
...
+                },
+                BatchSize::SmallInput,
+            );
+        });
+}
+
+// bench_reads_with_prepopulated_root_store benchmarks an instance of T by
+// first prepopulating it with 1 million key-value pairs before benchmarking
+// its read performance
+#[expect(clippy::unwrap_used)]
```
rkuris, line 80: I think you can move this to the file level rather than including it on each fn

firewood/src/manager.rs
```diff
@@ -260,7 +232,77 @@ impl RevisionManager {
     }
 }
 
-impl RevisionManager {
+impl RevisionManager<RocksDBStore> {
```
rkuris: nit: If you move this up to line 98, it would produce a much easier-to-review diff.

If you want to keep it here, add a note in the PR that explains that it's identical even though it's moved.

firewood/src/manager.rs
```diff
@@ -278,12 +320,28 @@ impl RevisionManager {
...
-            .ok_or(RevisionManagerError::RevisionNotFound {
-                provided: root_hash,
-            })?;
+            .cloned();
+
+        if let Some(v) = proposal {
+            return Ok(v);
+        }
+
+        // Get linear address
```
rkuris: Comments here could be improved. `view` now has three steps, so perhaps it warrants a block comment before the method itself.

Check out how this is done in more complex calls like `commit`. As soon as we hit 3 steps, it's helpful to readers later to have each step outlined somewhere.

firewood/src/manager.rs
```diff
@@ -278,12 +320,28 @@ impl RevisionManager {
...
+                provided: root_hash.clone(),
+            }
+        })?;
+
+        let cr = self.current_revision().get_underlying_storage();
+
+        let mut header = NodeStoreHeader::new();
+        header.set_root_address(Some(addr));
+
+        #[allow(clippy::useless_conversion)]
```
rkuris: There is a helper method that avoids this problem elsewhere in the code; use `root_hash.into_triehash()` instead of `root_hash.into()` which is a no-op for non-eth mode.

firewood/src/manager.rs
```diff
@@ -278,12 +320,28 @@ impl RevisionManager {
...
+        let node_store = NodeStore::new_committed(header, root_hash.into(), addr, cr);
 
-        Ok(proposal)
+        Ok(Arc::new(node_store))
```
rkuris: This means that if you request the same root repeatedly, we won't be sharing nodestores.

This is a hint that we may want to consider caching them.

firewood/src/root_store.rs
```diff
@@ -0,0 +1,435 @@
...
+    Database, Env, EnvOpenOptions, byteorder,
+    types::{self, Bytes},
+};
+use rocksdb::{DB, Options};
+use rusqlite::Connection;
+
+#[derive(Debug, thiserror::Error)]
+pub enum RootStoreError {
+    #[error("Failed to add root: {0}")]
+    AddRoot(String),
```
rkuris: This isn't very rusty, and means that we allocate memory every time we generate one of these errors. Let's just use the original error if you want these errors classified rather than turning the original error into a string.
```suggestion
    AddRoot(String),
```

firewood/src/root_store.rs
```diff
@@ -0,0 +1,435 @@
...
+    #[error("Failed to get root address: {0}")]
+    Get(String),
+    #[error("Failed to open RootStore: {0}")]
+    Open(String),
+}
+
+const ERR_EMPTY_ROOT: &str = "empty root address";
```
rkuris: This review exceeded my 30 minute time limit for reviews here. Is there any way to break this up into smaller parts?

RodrigoVillar:
Putting this into draft for the following reasons:

- We're still deciding on which implementation to use
- We should focus on adding code which RootStore requires before adding RootStore itself

RodrigoVillar:
Closing this as #1395 is the leading candidate for a production version of `RootStore`.

Issue #1230, opened 2025-08-21
New Type: PathComponent


demosdemon:
# Problem

We currently have the [`Path`](https://github.com/ava-labs/firewood/blob/1b72d7d5aa615d222238bf0f2500d07ba63aff8d/storage/src/node/path.rs#L29) structure. However, it has a couple problems:

1. It does not validate that the input bytes are valid nibbles.
2. It does not guarantee to users that all bytes are valid nibbles.
3. It allows external users to construct a path object that may not be valid.
4. It exposes direct access to its internal value.

# Proposed Solution

Create a `PathComponent` newtype that is a wrapper around `u8`. If `branch_factor_256` is enabled, `PathComponent` may be a full 8-bit value. However, when `branch_factor_256` is not enabled (the default), the `PathComponent` may only be a 4-bit value.

```rust
pub struct PathComponent(u8);

impl PathComponent {
    pub const MIN: Self = Self(0);

    pub const MAX: Self = Self(if cfg!(feature = "branch_factor_256") { 0xFF } else { 0x0F });

    pub const fn new(val: u8) -> Option<Self> {
        // If max == 255, the none branch will be dead and elide when compiled with optimizations
        if val <= Self::MAX.0 {
            Some(Self(val))
        } else {
            None
        }
    }
}
```

With the `PathComponent`, `Path` should make its inner value private and expose interfaces that only operate on `PathComponent`s and only as full bytes with explicit interfaces.

In the (not near) future, `branch_factor_256` will be a runtime option and not compile-time. Therefore, `PathComponent`'s interface should be written with the intent of converting it into a `trait` that will be shared by multiple branch factors.

PR #1336, opened 2025-10-06
feat(1/7): add U4 type to be used as a path component

demosdemon:
This change introduces a `U4` type that uses an underlying enum with only 16 states in order to enable niche optimizations. This change does not yet use the type within code as that will cause a large amount of changes that will make reviewing this difficult. The focus of this change is the `U4` type itself and nothing outside of that. This type also is not a `PathComponent`, which will appear in the next Pull Request.

## Memory Layout

The niche optimizations apply to memory layout allowing for `Option<U4>` to have the same size and alignment of `U4`, meaning there is no additional storage to keep the Option discriminant.

## Array Bounds Checking

The niche optimizations also apply to bounds checking when indexing into a fixed size array, like we do for the `Children<_>` type alias. See <https://rust.godbolt.org/z/cqdz1n9d5>

This change contributes to the resolution of #1230.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1088,6 +1088,7 @@ dependencies = [
  "cbindgen",
  "chrono",
  "coarsetime",
+ "derive-where",
  "env_logger",
  "firewood",
  "firewood-storage",
@@ -1963,9 +1964,9 @@ checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
 name = "libc"
-version = "0.2.176"
+version = "0.2.177"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "58f929b4d672ea937a23a1ab494143d968337a5f47e56d0815df1e0890ddf174"
+checksum = "2874a2af47a2325c2001a6e6fad9b16a53b802102b528163885171cf92b15976"
 
 [[package]]
 name = "libloading"
--- a/benchmark/Grafana-dashboard.json
+++ b/benchmark/Grafana-dashboard.json
@@ -1,35 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
-      "pluginName": "Prometheus"
-    }
-  ],
-  "__elements": {},
-  "__requires": [
-    {
-      "type": "grafana",
-      "id": "grafana",
-      "name": "Grafana",
-      "version": "11.5.1"
-    },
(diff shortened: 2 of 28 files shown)


storage/src/u4.rs
```diff
@@ -0,0 +1,341 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+/// An error similar to [`std::num::TryFromIntError`] but able to be created
+/// within our crate.
+///
+/// The std error does not have a public constructor, and neither does ours.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, thiserror::Error)]
+#[non_exhaustive]
+#[error("out of range integral type conversion attempted")]
```
demosdemon, line 10: The error message here is identical to that of `std::num::TryFromIntError`. I went back and forth about providing the original value, the invalid value, the index where the invalid value occurs, and some other possible error message states. But, I decided that what stdlib does is sufficient because if the caller wanted to include the invalid value, they have the value that was passed into the constructor because `u8` and `&[u8]` are both `Copy` and cheap to carry around after the call.

rkuris:
Why didn't we just depend on the existing U4 crate out there? Or at least be name compatible with those functions?

demosdemon:
> Why didn't we just depend on the existing U4 crate out there? Or at least be name compatible with those functions?

TBH, I didn't look because it was such a simple type and I didn't to add a dependency for that. After your comment, I did do some research on existing creates and I found [one](https://crates.io/crates/u4) that is implemented very similar to the way I do here but it is unmaintained and the source repo listed on crates.io no longer exists. So, I do not want to use that.

storage/src/u4.rs
```diff
@@ -0,0 +1,343 @@
...
+    #[test_case(0xC0, U4(Repr::OxC); "0xC0 -> 12")]
+    #[test_case(0xD0, U4(Repr::OxD); "0xD0 -> 13")]
+    #[test_case(0xE0, U4(Repr::OxE); "0xE0 -> 14")]
+    #[test_case(0xF0, U4(Repr::OxF); "0xF0 -> 15")]
+    #[test_case(0xFF, U4(Repr::OxF); "0xFF -> 15")]
+    fn test_new_shifted(input: u8, expected: U4) {
+        assert_eq!(U4::new_shifted(input), expected);
+    }
+
+    #[test_case(0x00, (U4(Repr::Ox0), U4(Repr::Ox0)); "0x00 -> (0, 0)")]
```
rkuris, line 334: nit: I wonder if test_matrix could have made this exhaustive

demosdemon: I attempted this; the matrix api would allow me to express the expected `(a, b)` pair but not the expected byte value.

PR #1337, opened 2025-10-06
feat(2/7): add newtype for PathComponent

demosdemon:
The new `PathComponent` type is introduced to encapsulate the possible values of a single component in the Trie path. If `branch_factor_256` is not enabled, the default, `PathComponent` wraps `U4`, which was added in the previous pull request. This type is only value for values 0-15, inclusive. If `branch_factor_256` is enabled, `PathComponent` wraps a `u8`, which can represent values 0-255, inclusive.

This change contributes to the resolution of #1230, but does not fully resolve it until all usages of `Path` are replaced with future changes.


storage/src/path/component.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub const ALL: [Self; 256] = {
+        let mut all = [Self(0); 256];
+        let mut i = 0;
+        #[expect(clippy::indexing_slicing)]
+        while i < 256 {
+            all[i] = Self(i as u8);
+            i += 1;
+        }
+        all
+    };
```
rkuris, line 71: fortunately, `core::array::from_fn` is const as of rust 1.63, so this can become:
```suggestion
pub const ALL: [Self; 256] = core::array::from_fn(|i| Self(i as u8));
```

rkuris (approved):
Approved, with suggestion

storage/src/path/component.rs
demosdemon: [`from_fn`](https://doc.rust-lang.org/stable/src/core/array/mod.rs.html#106-113) isn't const. It was added in 1.63 but has never been const.

PR #1338, opened 2025-10-06
feat: add TriePath trait and path abstraction

demosdemon:
The `TriePath` trait provides a unified interface for handling paths in the trie structure. Somtimes paths are backed by a sequence of bytes, and sometimes by a sequence of components. For branch_factor_256, they are the same. This abstraction allows us to work with both without running into type issues and inconsistencies with the two representations.

### DIFF
--- a/storage/src/path/joined.rs
+++ b/storage/src/path/joined.rs
@@ -1,0 +1,39 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use std::iter::Chain;
+
+use super::TriePath;
+
+/// Joins two path segments into a single path.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+pub struct JoinedPath<P, S> {
+    /// The prefix segment of the path.
+    pub prefix: P,
+    /// The suffix segment of the path.
+    pub suffix: S,
+}
+
+impl<P: TriePath, S: TriePath> JoinedPath<P, S> {
+    /// Creates a new joined path from the given prefix and suffix.
+    ///
+    /// This does not allocate and takes ownership of the input segments.
+    pub const fn new(prefix: P, suffix: S) -> Self {
+        Self { prefix, suffix }
+    }
+}
+
+impl<P: TriePath, S: TriePath> TriePath for JoinedPath<P, S> {
+    type Components<'a>
+        = Chain<P::Components<'a>, S::Components<'a>>
+    where
+        Self: 'a;
+
+    fn len(&self) -> usize {
+        self.prefix.len().wrapping_add(self.suffix.len())
+    }
+
+    fn components(&self) -> Self::Components<'_> {
+        self.prefix.components().chain(self.suffix.components())
+    }
+}
--- a/storage/src/path/mod.rs
+++ b/storage/src/path/mod.rs
@@ -2,5 +2,192 @@
 // See the file LICENSE.md for licensing terms.
 
 mod component;
+mod joined;
 
 pub use self::component::PathComponent;
+pub use self::joined::JoinedPath;
+
+/// A trie path of components with different underlying representations.
+///
+/// The underlying representation does not need to be a contiguous array of
+/// [`PathComponent`], but it must be possible to iterate over them in order
+/// as well as have a known length.
+pub trait TriePath {
+    /// The iterator returned by [`TriePath::components`].
+    type Components<'a>: Iterator<Item = PathComponent> + Clone + 'a
+    where
+        Self: 'a;
+
+    /// The length, in path components, of this path.
+    fn len(&self) -> usize;
+
+    /// Returns true if this path is empty (i.e. has length 0).
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns an iterator over the components of this path.
+    fn components(&self) -> Self::Components<'_>;
+
+    /// Joins this path with the given suffix, returning a new joined path.
+    ///
+    /// The returned path is a view over the two input paths and does not allocate.
+    /// The input paths are consumed and ownership is taken.
+    fn join_path<S>(self, suffix: S) -> JoinedPath<Self, S>
+    where
+        Self: Sized,
+        S: TriePath,
+    {
+        JoinedPath::new(self, suffix)
+    }
+
+    /// The inverse of [`TriePath::join_path`]; joins the given prefix with this path.
+    fn join_prefix<P>(self, prefix: P) -> JoinedPath<P, Self>
+    where
+        Self: Sized,
+        P: TriePath,
+    {
+        prefix.join_path(self)
+    }
+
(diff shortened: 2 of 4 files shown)


storage/src/path/joined.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use std::iter::Chain;
+
+use super::TriePath;
+
+/// Joins two path segments into a single path.
```
rkuris: nit: should mention why we're doing this. I think it's because we don't want to allocate more memory so we keep these separated.

If that's not the reason, then I'm wondering if we should `impl Add` instead.

storage/src/path/joined.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+impl<P: TriePath, S: TriePath> TriePath for JoinedPath<P, S> {
+    type Components<'a>
+        = Chain<P::Components<'a>, S::Components<'a>>
+    where
+        Self: 'a;
+
+    fn len(&self) -> usize {
+        self.prefix.len().wrapping_add(self.suffix.len())
```
rkuris: I think I would rather panic than wrap here, as it should never happen.
```suggestion
        self.prefix.len().checked_add(self.suffix.len()).expect("path size overflow")
```

storage/src/path/mod.rs
```diff
@@ -2,5 +2,192 @@
...
+    }
+
+    /// Returns an iterator over the components of this path.
+    fn components(&self) -> Self::Components<'_>;
+
+    /// Joins this path with the given suffix, returning a new joined path.
+    ///
+    /// The returned path is a view over the two input paths and does not allocate.
+    /// The input paths are consumed and ownership is taken.
+    fn join_path<S>(self, suffix: S) -> JoinedPath<Self, S>
```
rkuris: nit: maybe `join_append` is clearer, since the next one is a prepend, or maybe just `append`? I'm not sure `join` is doing it justice since it's just separate components, logically joined perhaps, but still kept separate.

Is it clearer to `impl Add` here (or just require a TriePath to use Add as a supertrait?)

storage/src/path/mod.rs
```diff
@@ -2,5 +2,192 @@
...
+    where
+        Self: Sized,
+        S: TriePath,
+    {
+        JoinedPath::new(self, suffix)
+    }
+
+    /// The inverse of [`TriePath::join_path`]; joins the given prefix with this path.
+    fn join_prefix<P>(self, prefix: P) -> JoinedPath<P, Self>
```
rkuris: nit: `join_prepend` might be clearer`, or even `prepend`.

storage/src/path/mod.rs
```diff
@@ -2,5 +2,192 @@
...
+    where
+        Self: Sized,
+        P: TriePath,
+    {
+        prefix.join_path(self)
+    }
+
+    /// Compares this path against another path for equality using path component
+    /// equality.
+    fn path_eq<T: TriePath>(&self, other: &T) -> bool {
```
rkuris: Same general comment here. Why can't we just `impl PartialEq` and `Eq` here?

storage/src/path/mod.rs
```diff
@@ -2,5 +2,192 @@
...
 
 pub use self::component::PathComponent;
+pub use self::joined::JoinedPath;
+
+/// A trie path of components with different underlying representations.
+///
+/// The underlying representation does not need to be a contiguous array of
+/// [`PathComponent`], but it must be possible to iterate over them in order
+/// as well as have a known length.
+pub trait TriePath {
```
rkuris, line 15: I would have expected this to have some super traits:
```suggestion
pub trait TriePath : PartialEq + Eq + PartialOrd + Ord {
```

or maybe even:
```
pub trait TriePath : PartialEq + Eq + PartialOrd + Ord + Add<Output = JoinedPath<Self, Self>>
```
(although that would force you to say TriePath is Sized, but I think that's okay)

storage/src/path/mod.rs
```diff
@@ -2,5 +2,192 @@
...
+    }
+}
+
+impl<P: TriePath + ?Sized> std::fmt::Display for DisplayPath<'_, P> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        display_path(f, self.path.components())
+    }
+}
+
+impl<T: TriePath + ?Sized> TriePath for &T {
```
rkuris, line 137: Isn't it possible to do something a bit more generic rather than re-implement this for each type that dereferences into a T? Something like:
```
impl<T, D> TriePath for D
where
    D: Deref<Target = T>,
    T: TriePath,
{
...
```

Then we don't need so many implementations for all things that dereference to T.

rkuris (changes requested):

storage/src/path/mod.rs
demosdemon: This is similar to [`Iterator::eq`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.eq) and [`Iterator::cmp`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.cmp) (and even uses them in implementation).

It is preferable to use these methods over `PartialCmp` because it allows us to ensure we can compare over any path. This is similar to adding the binding `PartialCmp<impl TriePath>` but that binding doesn't actually work.

storage/src/path/mod.rs
demosdemon: I do not want `TriePath` to be sized, because then `[PathComponent]` is no longer allowed to be a `TriePath` (and only `&[PathComponent]`) which unnecessarily requires a layer of indirection. But additionally, these trait bindings do not achieve the desired result. I want any triepath to be comparable to any other triepath, not just itself. In other words, I want to be able to compare `[PathComponent]` with `PackedPathRef` from #1341.

storage/src/path/mod.rs
demosdemon: No, this is not possible due to rust's orphaned impl rules. This is the same reason why I cannot do something like `impl<A: TriePath, B: TriePath> PartialEq<B> for A {...}` or `impl<A: TriePath, B: TriePath> Add<B> for A { type Output = JoinedPath<A, B>; ... }`

storage/src/path/joined.rs
demosdemon: this would be impossible to wrap in a well formed case, but I will update.

A slice and vector (of non-ZSTs) in rust are limited to a length `isize::MAX`. `isize::MAX * 2` is equal to `usize::MAX - 1`, therefore, adding the length of two slices technically cannot overflow a usize.

storage/src/path/mod.rs
demosdemon: My `Deref` comment isn't exactly correct. I can provide that implementation but it does prevent me from from providing a `TriePath` on any other type that implements Deref (regardless of if it yields something that impls TriePath) and also prevents me from implementing `TriePath` on anything outside of the crate because of the potential to implement `Deref`.

PR #1339, opened 2025-10-06
feat: add SplitPath trait

demosdemon:
One of the most common operations on paths is finding the longest common prefix between two paths. This is useful for a variety of reasons, such as determining if one path is a subpath of another, or finding the relative path between two paths.

A normal `TriePath` implementation does not provide a way to do this because that trait should remain `?Sized`. Whereas, splitting the path requires a sized type in order to produce a new instance of the path when splitting.

This change adds a `SplitPath` trait that provides the ability to split a path. Also added is `AsSplittablePath` which is a conversion trait to yield a splittable path. This also acts as the borrower for owned paths to yield shared references to the path.

### DIFF
--- a/storage/src/path/joined.rs
+++ b/storage/src/path/joined.rs
@@ -3,7 +3,7 @@
 
 use std::iter::Chain;
 
-use super::TriePath;
+use super::{AsSplittablePath, SplitPath, TriePath};
 
 /// Joins two path segments into a single path.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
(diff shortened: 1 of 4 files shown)


storage/src/path/joined.rs
```diff
@@ -37,3 +37,53 @@ impl<P: TriePath, S: TriePath> TriePath for JoinedPath<P, S> {
         self.prefix.components().chain(self.suffix.components())
     }
 }
+
+impl<P: SplitPath, S: SplitPath> SplitPath for JoinedPath<P, S> {
+    fn split_at(self, mid: usize) -> (Self, Self) {
+        if let Some(mid) = mid.checked_sub(self.prefix.len()) {
```
rkuris, line 52: nifty 🏆

storage/src/path/joined.rs
```diff
@@ -37,3 +37,53 @@ impl<P: TriePath, S: TriePath> TriePath for JoinedPath<P, S> {
...
+            };
+            let suffix: Self = Self {
+                prefix: b_prefix,
+                suffix: self.suffix,
+            };
+            (prefix, suffix)
+        }
+    }
+
+    fn split_first(self) -> Option<(super::PathComponent, Self)> {
```
rkuris, line 77: Is this impl substantially better than `split_at(1)`?

rkuris (approved):
nothing but nits. Seems like we should be using this in the merkle ops at some point. If so, perhaps we should make a ticket and make that a "good first issue"?

storage/src/path/joined.rs
demosdemon: Yeah, this is preferred over split_at(1) because `split_at` returns `(Self, Self)` and not `Option<(PathComponent, Self)>`. The differences:

- if len == 0, `split_at(1)` is expected to panic, similar to [`slice::split_at`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split_at) whereas split_first won't, also similar to [`slice`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split_first).
- if len > 1, the first path returned will be a single component which will need to be unwrapped into the path component, using something like `path.components().next().unwrap()`. This method relieves that buy unwrapping it ahead of time.

PR #1340, opened 2025-10-06
feat: Added DB configuration option to turn on/off parallel insert and hashing

bernard-avalabs:
Added an option to DB config and Configuration manager to turn on/off parallel propose when the database is opened. If it is turned off, then the threadpool option is set to None and normal propose is used. The current default is off, but it is always turned on for the test_proposal_parallel test case.

Also added a fastrace::span for parallel_merkle for parallel propose.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -108,6 +108,9 @@ pub struct DbConfig {
     /// Revision manager configuration.
     #[builder(default = RevisionManagerConfig::builder().build())]
     pub manager: RevisionManagerConfig,
+    // Whether to use parallel insert and hashing for propose.
+    #[builder(default = false)]
+    pub parallel: bool,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


ffi/firewood_test.go
```diff
@@ -142,7 +148,10 @@ func newTestDatabase(t *testing.T) *Database {
 func newDatabase(dbFile string) (*Database, func() error, error) {
 	conf := DefaultConfig()
 	conf.Truncate = true // in tests, we use filepath.Join, which creates an empty file
+	return newDatabaseConfig(dbFile, conf)
+}
 
+func newDatabaseConfig(dbFile string, conf *Config) (*Database, func() error, error) {
```
alarso16, line 154: nit:
```suggestion
func newDatabaseWithConfig(dbFile string, conf *Config) (*Database, func() error, error) {
```

ffi/firewood.go
```diff
@@ -68,6 +69,7 @@ func DefaultConfig() *Config {
 		FreeListCacheEntries: 40_000,
 		Revisions:            100,
 		ReadCacheStrategy:    OnlyCacheWrites,
+		Parallel:             false,
```
alarso16, line 72: If there is no functionality change and is a performance improvement, then why is this available to be set? Just to ensure that it works as expected prior to actually shipping it? It seems like this shouldn't even be available

ffi/firewood_test.go
```diff
@@ -127,11 +127,17 @@ func TestMain(m *testing.M) {
 }
 
 func newTestDatabase(t *testing.T) *Database {
+	conf := DefaultConfig()
+	conf.Truncate = true // in tests, we use filepath.Join, which creates an empty file
```
alarso16, line 131: Why is this necessary? Seems like it must have been working before...

ffi/firewood_test.go
```diff
@@ -127,11 +127,17 @@ func TestMain(m *testing.M) {
...
+	return newTestDatabaseConfig(conf, t)
+}
+
+func newTestDatabaseConfig(conf *Config, t *testing.T) *Database {
```
alarso16, line 135: nit:
```suggestion
func newTestDatabaseConfig(t *testing.T, conf *Config) *Database {
```

ffi/firewood_test.go
```diff
@@ -377,6 +386,43 @@ func TestInvariants(t *testing.T) {
 	r.Empty(got)
 }
 
+func TestProposalUsingParallel(t *testing.T) {
+	r := require.New(t)
+	conf := DefaultConfig()
+	conf.Truncate = true
```
alarso16, line 392: I don't think this is necessary

ffi/firewood.go
bernard-avalabs: There are some cases (e.g., very small proposals) in which parallel insertion/hashing is slower than serial insertion/hashing. We may want to allow the application using firewood to choose which version they want to use. 

We are still waiting on some performance numbers to determine if providing this option is necessary.

ffi/firewood_test.go
bernard-avalabs: newTestDatabase previously called newDatabase, which sets the Truncate flag to be true. 

In this PR, newTestDatabase calls newTestDatabaseConfig which calls newDatabaseConfig instead of newDatabase. None of the *Config functions sets the truncate flag since they take the configuration as a parameter. Because of this, I now set the truncate flag in newTestDatabase.

I could instead set truncate to be true in newDatabaseConfig, but I thought that might be more confusing.

ffi/firewood_test.go
bernard-avalabs: Since I'm calling newTestDatabaseConfig instead of newTestDatabase which uses the passed in configuration directly, I believe I need to manually set the truncate flag here (since the default is false). Perhaps I'm misunderstanding your suggestion.

ffi/firewood.go
rkuris: I think we should hold off on merging this until we know what the tradeoffs are. It's an easy change we can make, but nobody is going to want to manually configure this variable I think.

rkuris (changes requested):

Issue #1342, opened 2025-10-07
chore: duplicate code between Db.propose and Proposal.create_proposal


bernard-avalabs:
The two functions are nearly identical. Should be refactored to remove duplicate code. This would also allow Proposal.create_proposal to take advantage of parallel insert/hashing if it is enabled.

bernard-avalabs:
Implemented here: https://github.com/ava-labs/firewood/pull/1343

PR #1343, opened 2025-10-07
chore: Added helper to reduce code duplication between Db.propose and Proposal.create_proposal

bernard-avalabs:
Addresses https://github.com/ava-labs/firewood/issues/1342.

Created a `propose_with_parent` to allow code reuse between Db.propose and Proposal.create_proposal. `propose_with_parent` will also choose between using parallel or serial proposal creation based on the batch size. Currently the batch size for using parallel creation is 8, but there is a TODO to update the constant once we have some experimental results. We use `size_hint()` on the batch iterator to estimate the batch size.

### DIFF
--- a/.github/.golangci.yaml.patch
+++ b/.github/.golangci.yaml.patch
@@ -1,6 +1,6 @@
---- .github/.golangci.yaml	2025-10-29 08:30:25
-+++ ffi/.golangci.yaml	2025-10-29 08:30:25
-@@ -70,8 +70,6 @@
+--- .github/.golangci.yaml	2025-10-31 18:45:07.201781593 +0000
++++ ffi/.golangci.yaml	2025-10-31 18:45:07.200643422 +0000
+@@ -75,8 +75,6 @@
        rules:
          packages:
            deny:
@@ -9,7 +9,7 @@
              - pkg: github.com/golang/mock/gomock
                desc: go.uber.org/mock/gomock should be used instead.
              - pkg: github.com/stretchr/testify/assert
-@@ -86,29 +84,10 @@
+@@ -91,29 +89,10 @@
      forbidigo:
        # Forbid the following identifiers (list of regexp).
        forbid:
@@ -39,7 +39,7 @@
      revive:
        rules:
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bool-literal-in-expr
-@@ -172,17 +151,6 @@
+@@ -177,17 +156,6 @@
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#useless-break
          - name: useless-break
            disabled: false
@@ -57,7 +57,7 @@
      tagalign:
        align: true
        sort: true
-@@ -245,7 +213,7 @@
+@@ -250,7 +218,7 @@
          - standard
          - default
          - blank
--- a/.github/workflows/attach-static-libs.yaml
+++ b/.github/workflows/attach-static-libs.yaml
@@ -43,16 +43,21 @@ jobs:
             target: aarch64-unknown-linux-gnu
           - os: macos-latest
             target: aarch64-apple-darwin
-            pre-build-cmd: echo "MACOSX_DEPLOYMENT_TARGET=13.0" >> "$GITHUB_ENV"
-          - os: macos-13
+            pre-build-cmd: echo "MACOSX_DEPLOYMENT_TARGET=15.0" >> "$GITHUB_ENV"
+          - os: macos-latest
             target: x86_64-apple-darwin
-            pre-build-cmd: echo "MACOSX_DEPLOYMENT_TARGET=13.0" >> "$GITHUB_ENV"
+            pre-build-cmd: echo "MACOSX_DEPLOYMENT_TARGET=15.0" >> "$GITHUB_ENV"
     outputs:
       has_secrets: ${{ steps.check_secrets.outputs.has_secrets }}
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: ${{ matrix.target }}
+
+      - name: Add target for cross-compilation
+        run: rustup target add ${{ matrix.target }}
 
       - name: Run pre-build command
         if: matrix.pre-build-cmd != ''
@@ -183,7 +188,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, ubuntu-22.04-arm, macos-latest, macos-13]
+        os: [ubuntu-latest, ubuntu-22.04-arm, macos-latest]
     needs: push-firewood-ffi-libs
     continue-on-error: true
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -245,6 +245,7 @@ jobs:
         run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=firewood go test -race ./...
 
   ffi-nix:
+    if: false   # TODO(marun) Re-enable once non-determinism is rooted out
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1154,6 +1154,7 @@ dependencies = [
  "arc-swap",
  "bitfield",
  "bitflags 2.9.4",
+ "bumpalo",
  "bytemuck",
  "bytemuck_derive",
  "bytes",
--- a/benchmark/bootstrap/README.md
+++ b/benchmark/bootstrap/README.md
@@ -1,4 +1,81 @@
-# Bootstrap testing script
+# Bootstrap Testing Script
 
-This script creates instances and sets up users for bootstrap testing. You'll need ec2 installed on your machine
-and you'll need to `aws sso login` before running this script.
+This directory contains tools for automated Firewood blockchain database benchmarking on AWS. The `aws-launch.sh` script creates EC2 instances, sets up the complete testing environment, and executes C-chain (Avalanche) block bootstrapping tests.
+
+## Prerequisites
+
+Before running the script, you'll need:
+
+- AWS CLI installed and configured on your machine
+- Authenticated AWS session: `aws sso login`
+  - Your session should be configured to use the `Experimental` account.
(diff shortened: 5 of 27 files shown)


bernard-avalabs:
> 
> ```
> +            proposals: counter!("firewood.proposals"),
> +        describe_counter!("firewood.proposals", "Number of proposals created");
> -            proposals: counter!("firewood.proposals"),
> -        describe_counter!("firewood.proposals", "Number of proposals created");
> ```
> 
> However, the dashboard was not modified.
> 
> You may need to update `benchmark/Grafana-dashboard.json` accordingly.
> 
> _This check is automated to help maintain the dashboard._

Metrics did not change. Diff removed and re-added counters due to code movement.

firewood/src/db.rs
```diff
@@ -141,13 +141,62 @@ impl api::Db for Db {
         Ok(self.manager.all_hashes())
     }
 
-    #[fastrace::trace(short_name = true)]
     fn propose(
         &self,
         batch: impl IntoIterator<IntoIter: KeyValuePairIter>,
     ) -> Result<Self::Proposal<'_>, api::Error> {
         let parent = self.manager.current_revision();
+        self.propose_helper(batch, parent)
```
demosdemon:
```suggestion
        self.propose_helper(batch, &self.manager.current_revision())
```

firewood/src/db.rs
```diff
@@ -141,13 +141,62 @@ impl api::Db for Db {
...
+    pub fn check(&self, opt: CheckOpt) -> CheckerReport {
+        let latest_rev_nodestore = self.manager.current_revision();
+        latest_rev_nodestore.check(opt)
+    }
+
+    #[fastrace::trace(short_name = true)]
+    fn propose_helper<F: Parentable>(
+        &self,
+        batch: impl IntoIterator<IntoIter: KeyValuePairIter>,
+        parent: Arc<NodeStore<F, FileBacked>>,
```
demosdemon:
```suggestion
        parent: &NodeStore<F, FileBacked>,
```

---

nit: This can be passed in by reference. this is a minor change but allows you to avoid the atomics within `Proposal::create_proposal`.

firewood/src/db.rs
```diff
@@ -298,30 +305,7 @@ impl Proposal<'_> {
...
-        let nodestore = merkle.into_inner();
-        let immutable: Arc<NodeStore<Arc<ImmutableProposal>, FileBacked>> =
-            Arc::new(nodestore.try_into()?);
-        self.db.manager.add_proposal(immutable.clone());
-
-        Ok(Self {
-            nodestore: immutable,
-            db: self.db,
-        })
+        self.db.propose_helper(batch, parent)
```
demosdemon:
```suggestion
        self.db.propose_helper(batch, &self.nodestore)
```

firewood/src/db.rs
demosdemon:
```suggestion
            let immutable = parallel_merkle.create_proposal(parent, batch, threadpool)?;
```

demosdemon:
```suggestion
            let proposal = NodeStore::new(parent)?;
```

firewood/src/db.rs
```diff
@@ -141,13 +141,62 @@ impl api::Db for Db {
...
+        self.metrics.clone()
+    }
+
+    /// Check the database for consistency
+    pub fn check(&self, opt: CheckOpt) -> CheckerReport {
+        let latest_rev_nodestore = self.manager.current_revision();
+        latest_rev_nodestore.check(opt)
+    }
+
+    #[fastrace::trace(short_name = true)]
```
demosdemon:
```suggestion
    #[fastrace::trace(name = "propose")]
```

Let's change the name here so it's the same as before.

firewood/src/db.rs
```diff
@@ -141,13 +141,62 @@ impl api::Db for Db {
...
+    fn propose_helper<F: Parentable>(
```
rkuris: nit: naming. How about `propose_with_parent`. `helper` doesn't give much additional info as to what it's helping, whereas `propose_with_parent` says you're applying a batch against a specific parent.

Note that we may have the requirement of parenting proposals against revisions that are not the latest and are not other proposals, but we should probably document what kinds of parents we currently support.

rkuris (approved):
Looks good, but please apply the comments before merging.

firewood/src/db.rs
bernard-avalabs: I've made the change.

firewood/src/db.rs
bernard-avalabs: I've made the change.

firewood/src/db.rs
bernard-avalabs: I've made the change.

firewood/src/db.rs
bernard-avalabs: I've made the change.

firewood/src/db.rs
bernard-avalabs: I've changed the name and added a comment about what parents we currently support.

firewood/src/db.rs
```diff
@@ -235,20 +212,68 @@ impl<RS> Db<RS> {
         latest_rev_nodestore.check(opt)
     }
 
-    /// Propose a new batch that is processed in parallel.
+    /// Create a proposal with a specified parent. Currently, the parent can be another proposal or the
+    /// current revision. Proposals with a batch size larger than or equal to `MIN_BATCH_SIZE_FOR_PARALLEL`
+    /// are processed in parallel.
```
rkuris: nit: At this level I don't think anything stops us from creating a proposal against a non-current revision, so I'm not sure why this is mentioned in the comment.

firewood/src/db.rs
```diff
@@ -206,6 +174,13 @@ impl<RS: RootStore> Db<RS> {
...
         let manager =
             RevisionManager::new(db_path.as_ref().to_path_buf(), config_manager, root_store)?;
+        #[cfg(test)]
+        let db = Self {
+            metrics,
+            manager,
+            force_parallel: false,
+        };
+        #[cfg(not(test))]
         let db = Self { metrics, manager };
```
rkuris:
```suggestion
        let db = Self {
            metrics,
            manager,
            #[cfg(test)]
            force_parallel: false,
        };
```

firewood/src/db.rs
```diff
@@ -217,6 +192,8 @@ impl<RS: RootStore> Db<RS> {
 }
 
 impl<RS> Db<RS> {
+    const MIN_BATCH_SIZE_FOR_PARALLEL: usize = 8;
```
rkuris: Add a TODO that indicates we don't know this is the optimal size (or a comment that says how "8" was shown to be the best value)

firewood/src/db.rs
```diff
@@ -235,20 +212,68 @@ impl<RS> Db<RS> {
...
-            parallel_merkle.create_proposal(&parent, batch, self.manager.threadpool())?;
+        // Perform parallel proposal creation if the batch size is >= MIN_BATCH_SIZE_FOR_PARALLEL.
+        // Otherwise create the proposal serially as it is faster for small batches.
+        // TODO: Experimentally determine the right value for MIN_BATCH_SIZE_FOR_PARALLEL.
+        let batch = batch.into_iter();
+        #[cfg(test)]
+        let use_parallel =
+            self.force_parallel || batch.size_hint().0 >= Db::<RS>::MIN_BATCH_SIZE_FOR_PARALLEL;
+        #[cfg(not(test))]
+        let use_parallel = batch.size_hint().0 >= Db::<RS>::MIN_BATCH_SIZE_FOR_PARALLEL;
```
rkuris: Duplicating this code is going to possibly create testing errors. If I change the first line and forget to change the second one, I am going to test with the wrong conditions, and the tests may pass or fail but production will behave differently.

IMO the right way to fix this is to use a value set at database open time as the value of force_parallel. This is because I think we'll eventually need some way to mark parallel as always disabled. Maybe `force_parallel: Option<bool>` where None means it's automatically computed, true means always on, and false means always off.

firewood/src/db.rs
bernard-avalabs: I've moved the constant to be the default value of BatchSize and added a TODO there.

firewood/src/db.rs
bernard-avalabs: I've removed these different cases in favor of using a db option set at open time.

firewood/src/db.rs
bernard-avalabs: I've added a UseParallel enum that gets set at database open time. This code now checks the value of UseParallel and does not have separate code for test/non-test builds.

bernard-avalabs:
> IMO the right way to fix this is to use a value set at database open time as the value of force_parallel. This is because I think we'll eventually need some way to mark parallel as always disabled.
> 
> IMO the best way to fix this is to create an enum with values NEVER, BATCH_SIZE(usize), ALWAYS set at database open time. That has the higest readability and would probably make this code change much cleaner.

I don't think we need both a force_parallel Option and an enum. I've instead added a enum field in Db, and set BatchSize (8 currently) as the default configuration value. Instead of checking for an option, it performs a match on an Enum during proposal creation. Please let me know if I've misunderstood your comment.

firewood/src/db.rs
```diff
@@ -231,20 +215,65 @@ impl Db {
...
+            UseParallel::Always => true,
+            UseParallel::BatchSize(required_size) => batch.size_hint().0 >= required_size,
+        };
+        let immutable = if use_parallel {
+            let mut parallel_merkle = ParallelMerkle::default();
+            let span = fastrace::Span::enter_with_local_parent("parallel_merkle");
+            let immutable =
+                parallel_merkle.create_proposal(parent, batch, self.manager.threadpool())?;
+            drop(span);
+            immutable
```
rkuris: This seems more complicated than necessary. Why isn't this:
```suggestion
            parallel_merkle.create_proposal(parent, batch, self.manager.threadpool())?
```
since `span` goes out of scope and will be dropped automatically.

firewood/src/db.rs
```diff
@@ -231,20 +215,65 @@ impl Db {
...
+                        merkle.remove_prefix(prefix.as_ref())?;
+                    }
+                }
+            }
+
+            drop(span);
+            let span = fastrace::Span::enter_with_local_parent("freeze");
+
+            let nodestore = merkle.into_inner();
+            let immutable: Arc<NodeStore<Arc<ImmutableProposal>, FileBacked>> =
```
rkuris: same issue here

rkuris (approved):
Please make requested changes and merge

PR #1344, opened 2025-10-08
feat(6/7): Children newtype

demosdemon:
This change upgrades the `Children<T>` type from an alias to a formal new-type. This makes it possible to implement iterators and other functions that operate with the `PathComponent` type, making indexing and manipulation of the array safer and more ergonomic.


firewood/src/iter.rs
```diff
@@ -240,6 +241,8 @@ fn get_iterator_intial_state<T: TrieReader>(
 
                         return Ok(NodeIterState::Iterating { iter_stack });
                     };
+                    let next_unmatched_key_nibble =
+                        PathComponent::try_new(next_unmatched_key_nibble).expect("valid nibble");
```
demosdemon, line 245: this try_new/expect combo that is added here and in several other places will go away in the next PR

storage/src/node/branch.rs
```diff
@@ -414,135 +416,62 @@ impl Debug for BranchNode {
...
 
-    /// Returns the address of the child at the given index.
-    /// Panics if `child_index` >= [`BranchNode::MAX_CHILDREN`].
-    #[must_use]
-    pub fn child(&self, child_index: u8) -> &Option<Child> {
-        self.children
-            .get(child_index as usize)
-            .expect("child_index is in bounds")
-    }
+    // /// Convenience function to create a new array of empty children.
```
rkuris: Why is this code commented out instead of being removed?

storage/src/node/children.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// Returns the number of [`Some`] elements in this collection.
+    #[must_use]
+    pub fn count(&self) -> usize {
+        self.0.iter().filter(|c| c.is_some()).count()
+    }
+
+    /// Sets the element at `index` to `None`, returning the previous value.
+    #[must_use]
+    pub const fn take(&mut self, index: PathComponent) -> Option<T> {
```
rkuris, line 149: nit: Not crazy about the take/replace naming here, but no good suggestion on how to fix it.

storage/src/node/path.rs
```diff
@@ -192,6 +192,7 @@ impl Iterator for NibblesIterator<'_> {
 
     #[cfg(feature = "branch_factor_256")]
     fn next(&mut self) -> Option<Self::Item> {
+        #![expect(clippy::indexing_slicing)]
```
rkuris, line 195: This is odd, what are we indexing_slicing?

storage/src/node/children.rs
demosdemon: take/replace match the api names in `std::mem` [1] [2] and `Option` [3] [4]. They're used here for parity with the general ecosystem.

[1]: https://doc.rust-lang.org/stable/std/mem/fn.take.html
[2]: https://doc.rust-lang.org/stable/std/mem/fn.replace.html
[3]: https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.take
[4]: https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.replace

storage/src/node/path.rs
demosdemon: there is an index operation further down in this method. The lint was exposed in my testing, but I don't think this would ever appear in CI because we don't run it on this feature.

storage/src/node/branch.rs
```diff
@@ -106,6 +102,12 @@ pub enum Child {
     MaybePersisted(MaybePersistedNode, HashType),
 }
 
+impl Default for Child {
+    fn default() -> Self {
+        Child::Node(Node::default())
+    }
+}
+
```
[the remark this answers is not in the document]
demosdemon: bot is technically correct. The only place where `Child::default` is currently used is inside `<Merkle<NodeStore<MutableProposal, S>>>::flatten_branch` when converting a branch node with only one child into the child. There, we use `mem::take` to remove the child out of it's old memory. I could instead try to `Option::take` it from it's old slot instead of replacing the old slot with the Default value.

PR #1345, opened 2025-10-09
feat: Add just task runner to ensure reproducibility of CI

maru-ava:
As a follow-on to #1319, this PR introduces the [just](https://github.com/casey/just) command runner to ensure the new CI ffi-nix CI job can be trivially reproduced locally.


ffi/flake.nix
```diff
@@ -116,13 +116,25 @@
         program = "${go}/bin/go";
       };
 
+      apps.jq = {
```
maru-ava, line 119: Used by the `check-golang-version` recipe.

rkuris:
Rebasing as I think the build failure is due to changes fixed in main

rkuris (approved):

PR #1346, opened 2025-10-09
chore(db/manager): add `MockStore`

RodrigoVillar:
As part of #1299, this PR introduces the `RootStore` trait and makes both the revision manager and database generic over `RootStore`. The revision manager, in particular, is modified to make calls to its `RootStore` as described in the `RootStore` design doc.  It's expected that the archival state datastore will implement this `RootStore` trait in production.

This PR also introduces two implementations of `RootStore`:
- `NoOpStore`: a no-op implementation of `RootStore` that is used in all contexts outside of testing
- `MockStore`: a mock implementation of `RootStore` that is used for testing the `RootStore` API.

This PR also introduces two database tests to test the `RootStore` API.

### DIFF
--- a/firewood/src/root_store.rs
+++ b/firewood/src/root_store.rs
@@ -6,12 +6,18 @@ use std::{cell::RefCell, collections::HashMap, rc::Rc};
 
 use firewood_storage::{LinearAddress, TrieHash};
 
-#[derive(Debug, thiserror::Error)]
-pub enum RootStoreError {
-    #[error("Failed to add root")]
+#[derive(Debug)]
+pub enum RootStoreMethod {
     Add,
-    #[error("Failed to get root")]
     Get,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("A RootStore error occurred.")]
+pub struct RootStoreError {
+    pub method: RootStoreMethod,
+    #[source]
+    pub source: Box<dyn std::error::Error + Send + Sync>,
 }
 
 pub trait RootStore {
@@ -76,7 +82,10 @@ impl MockStore {
 impl RootStore for MockStore {
     fn add_root(&self, hash: &TrieHash, address: &LinearAddress) -> Result<(), RootStoreError> {
         if self.should_fail {
-            return Err(RootStoreError::Add);
+            return Err(RootStoreError {
+                method: RootStoreMethod::Add,
+                source: "Adding roots should fail".into(),
+            });
         }
 
         self.roots.borrow_mut().insert(hash.clone(), *address);
@@ -85,7 +94,10 @@ impl RootStore for MockStore {
 
     fn get(&self, hash: &TrieHash) -> Result<Option<LinearAddress>, RootStoreError> {
         if self.should_fail {
-            return Err(RootStoreError::Get);
+            return Err(RootStoreError {
+                method: RootStoreMethod::Get,
+                source: "Getting roots should fail".into(),
+            });
         }
 
         Ok(self.roots.borrow().get(hash).copied())


firewood/src/db.rs
```diff
@@ -110,16 +111,16 @@ pub struct DbConfig {
 
 #[derive(Debug)]
 /// A database instance.
-pub struct Db {
+pub struct Db<T: RootStore = NoOpStore> {
```
rkuris: Generally, structures shouldn't have bounds on them. Here's an article from stackoverflow that takes it to the max: https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl

Literally: "you should not put a bound on a struct unless the struct literally will not compile without it"
```suggestion
pub struct Db<T> {
```

firewood/src/db.rs
```diff
@@ -110,16 +111,16 @@ pub struct DbConfig {
...
     metrics: Arc<DbMetrics>,
-    manager: RevisionManager,
+    manager: RevisionManager<T>,
 }
 
-impl api::Db for Db {
+impl<T: RootStore> api::Db for Db<T> {
```
rkuris: You should attempt to remove these bounds too. I don't think you can, but make sure (has to be done after the struct has it removed).

firewood/src/db.rs
```diff
@@ -221,12 +234,12 @@ impl Db {
 
 #[derive(Debug)]
 /// A user-visible database proposal
-pub struct Proposal<'db> {
+pub struct Proposal<'db, T: RootStore> {
```
rkuris:
```suggestion
pub struct Proposal<'db, T> {
```

firewood/src/db.rs
```diff
@@ -867,4 +926,36 @@ mod test {
...
+
+    impl<T: RootStore> TestDb<T> {
+        fn path(&self) -> PathBuf {
+            [self.tmpdir.path().to_path_buf(), PathBuf::from("testdb")]
+                .iter()
+                .collect()
+        }
+    }
+
+    impl TestDb<MockStore> {
```
rkuris: Feels like this should be combined with the previous block, especially since reopen should get used with any RootStore implementation.

firewood/src/manager.rs
```diff
@@ -20,11 +20,15 @@ use metrics::gauge;
 use typed_builder::TypedBuilder;
 
 use crate::merkle::Merkle;
+#[cfg(test)]
+use crate::root_store::MockStore;
```
rkuris: nit: Move this down so it's not in the main block of use statements to improve readability

firewood/src/manager.rs
```diff
@@ -65,7 +69,7 @@ type CommittedRevision = Arc<NodeStore<Committed, FileBacked>>;
 type ProposedRevision = Arc<NodeStore<Arc<ImmutableProposal>, FileBacked>>;
 
 #[derive(Debug)]
-pub(crate) struct RevisionManager {
+pub(crate) struct RevisionManager<T: RootStore> {
```
rkuris:
```suggestion
pub(crate) struct RevisionManager<T> {
```
(Check all remaining structs, chances are they don't need bounds on T; I'm not mentioning any remaining ones)

firewood/src/manager.rs
```diff
@@ -162,7 +187,8 @@ impl RevisionManager {
     /// 4. Persist to disk. This includes flushing everything to disk.
     /// 5. Set last committed revision.
     ///    Set last committed revision in memory.
-    /// 6. Proposal Cleanup.
+    /// 6. Persist the revision to `RootStore`.
```
rkuris: Curious why this isn't before step 5. In some crazy world someone could be fast enough to start working with the revision before it was ever written to the RootStore.

firewood/src/manager.rs
```diff
@@ -260,30 +291,50 @@ impl RevisionManager {
     }
 }
 
-impl RevisionManager {
+impl<T: RootStore> RevisionManager<T> {
     pub fn add_proposal(&self, proposal: ProposedRevision) {
         self.proposals.lock().expect("poisoned lock").push(proposal);
     }
 
+    /// View the database at a specific time.
```
rkuris:
```suggestion
    /// View the database at a specific hash.
```

firewood/src/manager.rs
```diff
@@ -260,30 +291,50 @@ impl RevisionManager {
...
-            .cloned()
-            .ok_or(RevisionManagerError::RevisionNotFound {
-                provided: root_hash,
-            })?;
+            .cloned();
 
-        Ok(proposal)
+        if let Some(v) = proposal {
+            return Ok(v);
+        }
```
rkuris: I would have combined this with the previous block, something like:
```
        if let Some(proposal) = self.proposals.lock()...cloned() {
            return Ok(proposal);
        }
```

Avoids `v`. If you want to keep your old code, at rename `v` and shadow as `proposal`:
```suggestion
        if let Some(proposal) = proposal {
            return Ok(proposal);
        }
```

firewood/src/manager.rs
```diff
@@ -260,30 +291,50 @@ impl RevisionManager {
...
+                    provided: root_hash.clone(),
+                })?;
+
+        let node_store = NodeStore::with_root(
+            root_hash.into_hash_type(),
+            revision_addr,
+            self.current_revision(),
+        );
+
+        Ok(Arc::new(node_store))
```
rkuris, line 316: I would have done `Ok(node_store.into())` but I like this better actually

firewood/src/manager.rs
```diff
@@ -311,10 +362,18 @@ impl RevisionManager {
     }
 }
 
+#[cfg(test)]
```
rkuris: Was it not possible to put this in the `mod tests` block?

firewood/src/manager.rs
```diff
@@ -329,14 +388,14 @@ mod tests {
             .build();
 
         // First database instance should open successfully
-        let first_manager = RevisionManager::new(db_path.clone(), config.clone());
+        let first_manager = RevisionManager::new(db_path.clone(), config.clone(), NoOpStore {});
```
rkuris, line 409: The RootStore belongs in the ConfigManager. Arguably the path does too, but that would be a different PR.

firewood/src/db.rs
```diff
@@ -181,6 +182,16 @@ impl api::Db for Db {
 impl Db {
     /// Create a new database instance.
     pub fn new<P: AsRef<Path>>(db_path: P, cfg: DbConfig) -> Result<Self, api::Error> {
+        Self::new_with_root_store(db_path, cfg, NoOpStore {})
```
rkuris: You can avoid this by putting the root_store in DbConfig.

firewood/src/root_store.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[cfg(test)]
+#[derive(Debug, Clone)]
+pub struct MockStore {
+    roots: Rc<RefCell<HashMap<TrieHash, LinearAddress>>>,
+    should_add_root_fail: bool,
+    should_get_fail: bool,
+}
+
+#[allow(clippy::new_without_default)]
```
rkuris: Prior comment eliminates this warning (and the new method, just use `MockStore::default()`)

firewood/src/root_store.rs
```diff
@@ -0,0 +1,103 @@
...
+#[cfg(test)]
+impl MockStore {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            roots: Rc::new(RefCell::new(HashMap::new())),
+            should_add_root_fail: false,
+            should_get_fail: false,
```
rkuris: Are these two variables always the same value for all tests? If so, let's combine them:
```suggestion
            simulate_failures: false,
```

firewood/src/root_store.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use firewood_storage::{LinearAddress, TrieHash};
+
+#[derive(Debug, thiserror::Error)]
+pub enum RootStoreError {
+    #[error("Failed to add root")]
+    Add,
+    #[error("Failed to get root")]
+    Get,
+}
```
rkuris: We need the reason the Add or Get failed. I don't want to wonder when we get a call from the field.
```suggestion
pub enum RootStoreError<E: std::error::Error> {
    #[error("RootStore add")]
    Add(E),
    #[error("RootStore get")]
    Get(E),
}
```

A better approach might be to use `thiserror`'s `#[source]` method and an actual struct.

firewood/src/root_store.rs
```diff
@@ -0,0 +1,103 @@
...
+        Ok(())
+    }
+
+    fn get(&self, _hash: &TrieHash) -> Result<Option<LinearAddress>, RootStoreError> {
+        Ok(None)
+    }
+}
+
+#[cfg(test)]
+#[derive(Debug, Clone)]
```
rkuris: Magic!
```suggestion
#[derive(Debug, Default, Clone)]
```

firewood/src/root_store.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// Returns an instance of `MockStore` that fails for all `add_root` and `get` calls.
+    #[must_use]
+    pub fn with_failures() -> Self {
+        Self {
+            roots: Rc::new(RefCell::new(HashMap::new())),
+            should_add_root_fail: true,
+            should_get_fail: true,
+        }
```
rkuris, line 77:
```suggestion
        Self {
            simulate_failures: true,
            ..Default::default()
        }
```

firewood/src/manager.rs
```diff
@@ -129,6 +143,17 @@ impl RevisionManager {
             nodestore.flush_header_with_padding()?;
         }
 
+        // On startup, we always write the latest revision to RootStore
+        if let Some(root_hash) = manager.current_revision().root_hash() {
+            let root_address = manager.current_revision().root_address().ok_or(
+                RevisionManagerError::RevisionWithoutAddress {
```
rkuris, line 150: This means you can't open an empty database. I guess we don't have a test for that, please add one. Create a database, don't insert anything, and use the `reopen` method.

rkuris (changes requested):

firewood/src/db.rs
rkuris: Also, please consider renaming T (the generic). Maybe RS or R make more sense. For some reason, rust likes keeping these really short but RS carries some meaning whereas T doesn't really. Don't use S since we're using that all over the codebase for "storage".

firewood/src/manager.rs
RodrigoVillar: Done: a8fdb69

firewood/src/root_store.rs
RodrigoVillar: Done: e71f515

firewood/src/root_store.rs
RodrigoVillar: Done: e71f515 and 973d65293e84a677d9a1f78ea9d6aa1c6d6e63cb

firewood/src/root_store.rs
RodrigoVillar: Good call; done: 973d65293e84a677d9a1f78ea9d6aa1c6d6e63cb

firewood/src/root_store.rs
RodrigoVillar: Done: e71f515961181d1110d28a24d2cdb2f454540e26

firewood/src/manager.rs
RodrigoVillar: Decided to keep the original code since it's more legible IMO; renamed `v` here: fa40dd5deb66686728a786075d882e2ce902ec3a

firewood/src/manager.rs
RodrigoVillar: TIL you can put `impl` blocks inside `mod tests` blocks; done: 3118ccb86991700326e57236575a95ed8ca56b5e

firewood/src/manager.rs
RodrigoVillar: Done: 92af351

firewood/src/manager.rs
RodrigoVillar: Added a test here: 2a6b6b42876b047a7206845ca49e09dd1693e664. The test passes though so it seems that you can indeed open an empty database? Feel free to resolve if this is the case.

firewood/src/db.rs
RodrigoVillar: Done: 462baca31a7ce845d6c738035fbc551db62d0539

firewood/src/db.rs
RodrigoVillar: Unfortunately cannot remove these bounds due the following later in `db.rs`:

```rust
impl<'db, T: RootStore> api::Proposal for Proposal<'db, T> { 
    // ...
}
```

firewood/src/manager.rs
RodrigoVillar: Good point; I believe I confused `Set last committed revision` with actually persisting the proposal (which needs to occur before writing to `RootStore`). Modified here: 6b640bc

firewood/src/db.rs
RodrigoVillar: This would be awesome; my original idea was to add the following to the revision manager:

```rust
impl<T: RootStore> RevisionManager<T> {
        pub fn root_store(&self) -> T {
            self.root_store
        }
    }
```

However, to do this, I would need to add the copy/clone trait to `RootStore`. I'm fine with this but wanted to pass this by you as this would mean mixing production code with test code.

firewood/src/manager.rs
RodrigoVillar: Done: 462baca31a7ce845d6c738035fbc551db62d0539 (also moved code that didn't need the `RootStore` bound into their own `impl` blocks)

firewood/src/db.rs
RodrigoVillar: Thanks for linking the article (and educating me on the consequences of using trait bounds). I've gone ahead and removed the trait bounds wherever possible in 462baca31a7ce845d6c738035fbc551db62d0539 and 5cf563b92e764fc6e695f8613a42c52b4cec4162.

One thing I've left behind in these commits is the default generic type `NoOpStore`. My rationale for this is that consumers of `Db` outside of the `firewood` crate will only be using a `NoOpStore`. By attaching the default generic here, consumers of `Db` in other crates such as `benchmark` and `ffi` need not worry about attaching a generic type parameter to `Db` which will always be set to `NoOpStore`. Open to feedback/responses on this though as this isn't discussed in the article linked.

firewood/src/db.rs
RodrigoVillar: Update: was able to unify the `MockStore` and `NoOpStore` code here: b714961

firewood/src/db.rs
rkuris: I don't understand why proposals need to know about the type of the rootstore. Why can't you remove that bound too?

firewood/src/root_store.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn get(&self, _hash: &TrieHash) -> Result<Option<LinearAddress>, RootStoreError> {
+        Ok(None)
+    }
+}
+
+#[cfg(test)]
+#[cfg_attr(test, derive(Clone))]
+#[derive(Debug, Default)]
+pub struct MockStore {
+    roots: Rc<RefCell<HashMap<TrieHash, LinearAddress>>>,
```
rkuris, line 65: Interesting to note that you were able to use a non-thread-safe variant here. A multi-threaded test won't be able to use MockStore. I probably would have made this thread-safe just in case someone wants to use it in a multi-threaded test later, but I can see that's not required yet.

firewood/src/manager.rs
```diff
@@ -262,32 +274,68 @@ impl RevisionManager {
...
+        Ok(Arc::new(node_store))
+    }
+}
+
+impl<T> RevisionManager<T> {
+    pub fn add_proposal(&self, proposal: ProposedRevision) {
+        self.proposals.lock().expect("poisoned lock").push(proposal);
+    }
+
+    pub fn all_hashes(&self) -> Vec<TrieHash> {
```
rkuris, line 326: This needs some more thought. Do we need to support fetching all hashes from the RootStore? This might be handy for debugging. If so, we probably don't want to return a `Vec` anymore, this should be an iterator.

Maybe this shouldn't be called `all_hashes`...

I think a TODO here would be sufficient, as this would be an API change.

firewood/src/db.rs
RodrigoVillar: Proposals need to know about the type of RootStore because of the commit method: 

```rs
fn commit(self) -> Result<(), api::Error> {
        Ok(self.db.manager.commit(self.nodestore)?)
    }
```

This method calls the `commit()` method on the manager, which needs the RootStore bound because this is where we persist the revision.

firewood/src/root_store.rs
RodrigoVillar: Ack.

firewood/src/db.rs
RodrigoVillar: Also renamed `T` in these commits: ee8a716 and 0ad97a1

firewood/src/manager.rs
```diff
@@ -262,32 +274,68 @@ impl RevisionManager {
...
+            .iter()
+            .filter_map(|r| r.root_hash().or_default_root_hash())
+            .chain(
+                self.proposals
+                    .lock()
+                    .expect("poisoned lock")
+                    .iter()
+                    .filter_map(|p| p.root_hash().or_default_root_hash()),
+            )
+            .collect()
```
RodrigoVillar, line 339: Note: this is not new code. I moved these methods to their own `impl` block as these methods don't require a trait bound on `RS`.

firewood/src/manager.rs
RodrigoVillar: Great point; added a `TODO` comment here: 00fd700

firewood/src/manager.rs
RodrigoVillar: Discussed offline - we shouldn't add `RootStore` to the `ConfigManager` as this would be adding a data structure to a set of parameters.

firewood/src/db.rs
RodrigoVillar: Discussed offline - we shouldn't add RootStore to the `DbConfig` as this would be adding a data structure to a set of parameters.

firewood/src/root_store.rs
RodrigoVillar: Done: 2b038ca

PR #1349, opened 2025-10-10
feat(ffi): `Database.Close()` guarantees proposals committed or freed

ARR4N:
Explicit freeing of proposals propagated through `libevm` (i.e. `geth`) plumbing has proven difficult when not being committed as they are simply dropped for the GC to collect. Furthermore, strict ordering of calls to `Proposal.Drop()` (or `Commit()`) before `Database.Close()` is required to avoid segfaults. This PR implements a fix for both issues:

1. All new `Proposal`s have a GC finalizer attached, which calls `Drop()`. This is safe because it is a no-op if called twice or after a call to `Commit()`.
2. The `Database` has a `sync.WaitGroup` introduced, which tracks all outstanding proposals. Calls to `Commit()` / `Drop()` decrement the group counter (only once per `Proposal`).
3. `Database.Close()` waits on the `WaitGroup` before freeing its own handle, avoiding segfaults.

Assuming that all calls to `Database.Propose()` and `Proposal.Propose()` occur _before_ the call to `Database.Close()` then this is a correct usage of `sync.WaitGroup`'s documented requirement for ordering of calls to `Add()` and `Wait()`.

An integration test demonstrates blocking and eventual return of `Database.Close()`, specifically due to the unreachability of un-dropped, un-committed `Proposal`s, resulting in their finalizers decrementing the `WaitGroup`.

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -251,6 +251,11 @@ func (db *Database) Revision(root []byte) (*Revision, error) {
 	return rev, nil
 }
 
+// defaultCloseTimeout is the duration by which the [context.Context] passed to
+// [Database.Close] is limited. A minute is arbitrary but well above what is
+// reasonably required, and is chosen simply to avoid permanently blocking.
+var defaultCloseTimeout = time.Minute
+
 // Close releases the memory associated with the Database.
 //
 // This blocks until all outstanding Proposals are either unreachable or one of
@@ -275,7 +280,7 @@ func (db *Database) Close(ctx context.Context) error {
 		close(done)
 	}()
 
-	ctx, cancel := context.WithTimeout(ctx, time.Minute /*arbitrary but definitely long enough*/)
+	ctx, cancel := context.WithTimeout(ctx, defaultCloseTimeout)
 	defer cancel()
 	select {
 	case <-done:


ffi/proposal.go
```diff
@@ -104,25 +100,32 @@ func (p *Proposal) Drop() error {
...
 		proposal := &Proposal{
-			db:     db,
-			handle: body.handle,
-			root:   hashKey[:],
+			handle:        body.handle,
+			root:          hashKey[:],
+			openProposals: openProposals,
 		}
+		openProposals.Add(1)
+		runtime.SetFinalizer(proposal, (*Proposal).Drop)
```
alarso16, line 157: Any reason not to use `runtime.AddCleanup`?

ffi/proposal.go
```diff
@@ -104,25 +100,32 @@ func (p *Proposal) Drop() error {
...
 	}
-
-	p.handle = nil // Prevent double free
-
+	p.afterDisowned()
 	return nil
 }
 
+func (p *Proposal) afterDisowned() {
+	p.freeOnce.Do(func() {
```
alarso16: It's good that we've prevented some racy behavior here anyway

ffi/proposal.go
```diff
@@ -58,8 +56,8 @@ func (p *Proposal) Get(key []byte) ([]byte, error) {
 	return getValueFromValueResult(C.fwd_get_from_proposal(p.handle, newBorrowedBytes(key, &pinner)))
 }
 
-// Propose creates a new proposal with the given keys and values.
-// The proposal is not committed until Commit is called.
+// Propose is equivalent to [Database.Close] except that the new proposal is
```
alarso16:
```suggestion
// Propose is equivalent to [Database.Propose] except that the new proposal is
```

ffi/proposal.go
```diff
@@ -14,25 +14,23 @@ import (
...
 	// handle is an opaque pointer to the proposal within Firewood. It should be
 	// passed to the C FFI functions that operate on proposals
 	//
 	// It is not safe to call these methods with a nil handle.
 	//
 	// Calls to `C.fwd_commit_proposal` and `C.fwd_free_proposal` will invalidate
 	// this handle, so it should not be used after those calls.
-	handle *C.ProposalHandle
+	handle        *C.ProposalHandle
+	openProposals *sync.WaitGroup
```
alarso16, line 36: Maybe add a comment for how `openProposals` is used (even if it just points you to `[Database.Close]`

ffi/firewood_test.go
```diff
@@ -1050,3 +1050,60 @@ func TestGetFromRootParallel(t *testing.T) {
...
+	go func() {
+		require.NoErrorf(t, db.Close(), "%T.Close()", db)
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		t.Errorf("%T.Close() returned with undropped %T", db, p0) //nolint:forbidigo // Use of require is impossible without a hack like require.False(true)
+	case <-time.After(300 * time.Millisecond):
+		// TODO(arr4n) use `synctest` package when at Go 1.25
```
alarso16, line 1211: This is neat, I've never heard of this. This does seem to solve a pretty common pattern in testing.

ffi/proposal.go
ARR4N: It was only introduced in Go 1.24 so would force a `libevm` update. If we have to update anyway then we should use it instead.

ffi/proposal.go
ARR4N: What do you mean? Can you give examples? Technically this `Once` isn't necessary, but I put it in defensively in case the rest of the code is refactored and current invariants no longer hold.

ffi/proposal.go
ARR4N: 🤣

ffi/firewood_test.go
ARR4N: Yeah, I can't wait to start using it!

ffi/proposal.go
```diff
@@ -28,7 +28,11 @@ type Proposal struct {
 	//
 	// Calls to `C.fwd_commit_proposal` and `C.fwd_free_proposal` will invalidate
 	// this handle, so it should not be used after those calls.
-	handle        *C.ProposalHandle
+	handle *C.ProposalHandle
+
+	// [Database.Close] blocks on this WaitGroup, which is incremented by
+	// [getProposalFromProposalResult], and decremented by either
+	// [Proposal.Commit] or [Proposal.Done].
```
alarso16:
```suggestion
	// [Proposal.Commit] or [Proposal.Drop].
```

ffi/proposal.go
alarso16: We do rely that the consumer isn't trying to simultaneously Commit and Drop, which seems reasonable. However, in that case, you would get UB. It's probably best to make UB completely impossible, even if the actions to make it happen are unreasonable

ffi/proposal.go
alarso16: Firewood did already update to 1.24.7 (probably should update to 1.24.8, but they don't use any of the vulnerable packages)

ffi/proposal.go
ARR4N: They can definitely still race and both call the Rust code, but only under invalid usage as you say. This just guarantees that the `WaitGroup` never panics by going negative.

demosdemon: This won't prevent racy behavior. If we wanted to do that, the `freeOnce` call also needs to wrap the `C.fwd_free_proposal` and `C.fwd_commit_proposal` calls as there's nothing preventing a concurrent call to `Drop()`/`Commit()` while this is running.

ffi/firewood.go
```diff
@@ -233,17 +237,23 @@ func (db *Database) Revision(root []byte) (*Revision, error) {
...
+// nothing. The pointer will be set to nil after freeing to prevent double free.
+// However, it is not safe to call this method concurrently from multiple
+// goroutines.
 func (db *Database) Close() error {
 	if db.handle == nil {
 		return nil
 	}
 
+	runtime.GC()
+	db.proposals.Wait()
```
rkuris: Does this mean that a leaked proposal now becomes a hang at exit time?

If so, we should consider returning an error instead of hanging. This would make it much harder to debug if it happens on a system we have no control over, but if there's a log that says "hey there was a leaked proposal" somewhere that would make debugging a lot easier.

One way to do this is via a timeout, maybe with some large amount of time (60 seconds).

ffi/firewood_test.go
rkuris: In theory we could use it now if we add `GOEXPERIMENT=synctest`

ffi/firewood.go
```diff
@@ -233,17 +237,23 @@ func (db *Database) Revision(root []byte) (*Revision, error) {
...
+// This is safe to call if the handle pointer is nil, in which case it does
+// nothing. The pointer will be set to nil after freeing to prevent double free.
+// However, it is not safe to call this method concurrently from multiple
+// goroutines.
 func (db *Database) Close() error {
 	if db.handle == nil {
 		return nil
 	}
 
+	runtime.GC()
```
demosdemon: Not a fan of the GC call. I assume it's to try to eagerly run any outstanding finalizers. But, GC will also include everything else and may penalize us more than necessary.

ffi/proposal.go
alarso16: Ahh I see. Is that worth preventing (in a separate PR)?

ffi/firewood.go
ARR4N: > I assume it's to try to eagerly run any outstanding finalizers

Yup

> GC will also include everything else and may penalize us more than necessary

Good point. I've put it in a separate go routine to avoid this, but I think it's important to still include due to the above.

ffi/proposal.go
ARR4N: Putting them in the `Once` would be problematic as, if either returned an error, then they couldn't be called again.

There's no need to place it in a separate PR IMO as it's simply a mutex. It's also worth doing to allay @demosdemon's concerns [here](https://github.com/ava-labs/firewood/pull/1349#pullrequestreview-3324518115):

> That could also potentially allow the finalizer to run on the proposal concurrently while commit is running; but, I believe that won't actually happen in practice because of the `Once`.

It's true that it won't actually happen because the `Proposal` remains alive long enough, but an explicit lock is much easier to reason about than GC lifetime.

ffi/firewood_test.go
ARR4N: > In theory we could use it now if we add `GOEXPERIMENT=synctest`

Unfortunately it [requires compiling Go itself with this](https://pkg.go.dev/testing/synctest@go1.24.7#:~:text=only%20exists%20when%20using%20Go%20compiled%20with%20GOEXPERIMENT%3Dsynctest), not just running the test.

ARR4N:
> Really like the use of `WaitGroup`. My main concern is the explicit GC call.

[Addressed](https://github.com/ava-labs/firewood/pull/1349#discussion_r2425867230) in your code-specific comment

> But, something else I noticed is that there's nothing clearing the finalizer that's been set. So, the finalizer will _always_ call `Drop` even if `Drop` or `Commit` was called outside of the finalizer. Ideally we would clear the finalizer to prevent that from happening if there was an explicit call. That could also potentially allow the finalizer to run on the proposal concurrently while commit is running; but, I believe that won't actually happen in practice because of the `Once`.

A call to `Drop` after a call to either of the others is a no-op, and the two are now thread-safe with respect to each other although your final point is correct about it not actually happening. Further details [here](https://github.com/ava-labs/firewood/pull/1349#discussion_r2425929258).

ffi/firewood.go
ARR4N: The idiomatic approach is to accept a `Context` and then add an extra timeout. @alarso16 do we absolutely have to conform to the `kvBackend` interface (which precludes adding the `Context`)?

ffi/proposal.go
ARR4N: `AddCleanup` requires the `arg` to be different to the value having the cleanup added, which would require a bunch of weird refactoring (moving everything except the `root` field) into a different struct. The benefit would be that it couldn't be cleared externally, but I'm not sure that outweighs the cost. Using the `(*Proposal).Drop` pattern negates the primary foot-gun of finalizers.

ffi/firewood.go
alarso16: No, this is consumed by `triedb`. This struct doesn't implement any particular interface required by libevm, but will be called on `DBOverride.Close()`. Accepting a context would be my first thought anywhere else, so maybe we just require the consumer to understand that the operation may hang by sending a context. libevm can create an ephemeral context I guess.

ffi/proposal.go
```diff
@@ -87,22 +88,36 @@ func (p *Proposal) Propose(keys, vals [][]byte) (*Proposal, error) {
 	if err != nil {
 		return nil, err
 	}
+	return getProposalFromProposalResult(C.fwd_propose_on_proposal(p.handle, kvp), p.openProposals)
+}
+
+// disownHandle is the common path of [Proposal.Commit] and [Proposal.Drop], the
+// `fn` argument defining the method-specific behaviour.
+func (p *Proposal) disownHandle(fn func(*C.ProposalHandle) error, disownEvenOnErr bool) error {
```
alarso16, line 96: I like that the behavior is clarified for what happens to the lifetime in the error case

ffi/proposal.go
```diff
@@ -112,32 +127,34 @@ func (p *Proposal) Commit() error {
 //
 // The pointer will be set to nil after freeing to prevent double free.
 func (p *Proposal) Drop() error {
-	if p.handle == nil {
-		return nil
+	if err := p.disownHandle(dropProposal, false); err != nil && err != errDroppedProposal {
```
alarso16, line 130: Is using `errors.Is` better practice? Or since we know that the err isn't wrapped, this check is easier?

ARR4N: > Is using `errors.Is` better practice?

Nope, it's redundant here.

> Or since we know that the err isn't wrapped, this check is easier?

Exactly. It's useful when there's a desire to wrap an error, but in this case there isn't any.

ffi/proposal.go
ARR4N: Me too. It actually caught me off guard when I first did this refactoring.

ARR4N:
> Should `Drop` even be exposed to the user if we guarantee to free memory on GC? I think yes, it should still be available, but wanted to check with others

I think it absolutely should be, otherwise the only way to close the database is to guarantee that every proposal becomes unreachable.

ffi/firewood.go
ARR4N: `Context` added.

ffi/firewood.go
```diff
@@ -247,17 +253,36 @@ func (db *Database) Revision(root []byte) (*Revision, error) {
...
+
+	ctx, cancel := context.WithTimeout(ctx, time.Minute /*arbitrary but definitely long enough*/)
+	defer cancel()
+	select {
+	case <-done:
+	case <-ctx.Done():
+		return fmt.Errorf("at least one reachable %T neither dropped nor committed", &Proposal{})
+	}
+
 	if err := getErrorFromVoidResult(C.fwd_close_db(db.handle)); err != nil {
```
alarso16, line 291: So if the user makes a mistake and forgets to drop a proposal, the database won't close. I think this is the behavior we should enforce, but it does seem weird

ARR4N: > I think this is the behavior we should enforce

Me too because it suggests a leak somewhere. Not necessarily a raw resource leak, but something smelly on the consuming side.

> but it does seem weird

How so? I'm not disagreeing as much as wanting to understand more precisely in case there are alternative approaches.

alarso16: If I make a mistake as a user, even if I should have dropped any outstanding proposals, should we really rely on the OS to close the file later? We know that we can handle it more gracefully

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -78,7 +79,7 @@ func newMerkleTriePair(t *testing.T) *merkleTriePair {
 	tr, err := tdb.OpenTrie(ethRoot)
 	r.NoError(err)
 	t.Cleanup(func() {
-		r.NoError(db.Close())
+		r.NoError(db.Close(context.Background())) //nolint:usetesting // t.Context() will already be cancelled
```
alarso16, line 82: Interesting! I didn't know when it canceled.

ffi/firewood.go
alarso16: I don't have a strong opinion on this - just worth thinking about

ffi/tests/eth/eth_compatibility_test.go
ARR4N: Me neither, but I used it and then the test failed.

ffi/firewood.go
```diff
@@ -247,17 +253,36 @@ func (db *Database) Revision(root []byte) (*Revision, error) {
...
 
+	go runtime.GC()
+
+	done := make(chan struct{})
+	go func() {
+		db.proposals.Wait()
+		close(done)
+	}()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Minute /*arbitrary but definitely long enough*/)
```
demosdemon: nit: can we move the one minute to a var that is set by config or envvar on `init`? This would allow you to also set the timeout to something other than 1 minute for testing.

ffi/firewood.go
demosdemon: While we cannot rely on this forever, as of today the storage on disk for firewood will not be in an indeterminate state as long as there are no goroutines invoking FFI at the time the process terminates. This is because firewood does not yet have deferred persist implemented; therefore, firewood will complete all writes to disk (except for fsync) before the call to `Commit` returns. This also means it is safe for the process to terminate while the database is unclosed. 

Once deferred persist is implemented (and enabled), that won't be true as closing the database will be required to flush and terminate the workers and any outstanding proposals at that point would prevent us from doing so.

ffi/firewood.go
ARR4N: Abstracted into a variable that can be modified. In testing I assume it would be shorter, so the passed-in context should instead just have its own timeout added.

ffi/firewood.go
```diff
@@ -245,19 +251,43 @@ func (db *Database) Revision(root []byte) (*Revision, error) {
...
+// [Database.Close] is limited. A minute is arbitrary but well above what is
+// reasonably required, and is chosen simply to avoid permanently blocking.
+var defaultCloseTimeout = time.Minute
+
 // Close releases the memory associated with the Database.
 //
-// This is not safe to call while there are any outstanding Proposals. All proposals
-// must be freed or committed before calling this.
+// This blocks until all outstanding Proposals are either unreachable or one of
+// [Proposal.Commit] or [Proposal.Drop] has been called on them. Unreachable
```
rkuris, line 262: This is no longer correct. It also fails with an error if more than 1 minute passes.

ffi/firewood.go
rkuris: Yeah but generally speaking that's not a big deal. It just means that some additional blocks will have to be played forward on restart. The same thing will happen if the application crashes or it's killed.

rkuris (approved):
One documentation nit to fix, then merge it

PR #1350, opened 2025-10-10
chore(storage): add `into_hashtype()`

RodrigoVillar:
This PR breaks up #1346 by adding the `into_hashtype()` method.

Currently, `RootStore` plans to be a mapping from root hashes (`TrieHash`) to root addresses (`LinearAddress`). However, to construct a `NodeStore` (and therefore, a `Child`), we need to use a `HashType` value as seen:

https://github.com/ava-labs/firewood/blob/18a95bdfcebf86f7843bfd9d88bcd42ae8cc4985/storage/src/node/branch.rs#L97-L107

This PR fixes this by allowing root hashes of type `TrieHash` to convert to a `HashType` value, allowing for their usage in constructing a `Child`.

### DIFF
--- a/storage/src/trie_hash.rs
+++ b/storage/src/trie_hash.rs
@@ -1,8 +1,8 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-use crate::node::ExtendableBytes;
 use crate::node::branch::Serializable;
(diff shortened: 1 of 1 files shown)


RodrigoVillar:
@rkuris I thought this didn't exist as I couldn't find the `into_triehash()` method mentioned here: https://github.com/ava-labs/firewood/pull/1333#discussion_r2406949515 ; apologies for this duplicate code PR.

Issue #1347, opened 2025-10-09
bug: Bad freelist after an IO error


AminR443:
One of my block re-execution tests failed with this error:

```
[10-09|13:51:02.221] INFO vm-executor c/vm_reexecute_test.go:406 executing block {"height": 57584000, "eta": "9h55m18s"}
[10-09|13:51:25.594] INFO vm-executor c/vm_reexecute_test.go:406 executing block {"height": 57585000, "eta": "9h55m6s"}
[10-09|13:51:31.905] CRIT <2q9e4r6Mu3U68nU1fYjgbR6JvwrRx36CohpAX5UQxse55x1Q5 Chain> core/blockchain.go:612 unable to flatten snapshot from acceptor blockHash=1187a6..01948f err="firewood: error committing proposal 0x624419e82ea9de200bd938b9713a32c8519abc7bdd0d3135075b9e4449bae984: File IO error: Interrupted system call (os error 4) at offset 0 of file '/mnt/nvme/ubuntu/exec-data/current-state/chain-data-dir/firewood.db' io-uring submit_and_wait"
exit status 1
FAIL    github.com/ava-labs/avalanchego/tests/reexecute/c       112153.371s
FAIL
task: Failed to run task "reexecute-cchain-range": exit status 1
```

Attempting to resume from last block leads to a new error indicating a bad freelist:

```
[10-09|17:57:35.318] INFO <2q9e4r6Mu3U68nU1fYjgbR6JvwrRx36CohpAX5UQxse55x1Q5 Chain> core/blockchain.go:744 Loaded most recent local header          number=57,585,326 hash=1187a6..01948f age=7mo3w14h
[10-09|17:57:35.318] INFO <2q9e4r6Mu3U68nU1fYjgbR6JvwrRx36CohpAX5UQxse55x1Q5 Chain> core/blockchain.go:745 Loaded most recent local full block      number=57,585,326 hash=1187a6..01948f age=7mo3w14h
[10-09|17:57:35.318] INFO <2q9e4r6Mu3U68nU1fYjgbR6JvwrRx36CohpAX5UQxse55x1Q5 Chain> core/blockchain.go:1822 Loaded Acceptor tip                      hash=1187a6..01948f
[...]

PR #1354, opened 2025-10-10
fix: EINTR during iouring calls

rkuris:
golang does not guarantee signals are using SA_RESTART, so any system call on a "slow device" might cause the EINTR error. We saw this in the error in issue #1347.

This doesn't resolve that issue completely because IO errors can occur during these writes. That will be fixed under a separate issue.

My research indicated that re-calling the system call is the right thing to do here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1153,7 +1153,6 @@ dependencies = [
  "indicatif",
  "integer-encoding",
  "io-uring",
- "libc",
  "log",
  "lru",
  "metrics",


storage/src/nodestore/persist.rs
```diff
@@ -316,6 +319,33 @@ impl NodeStore<Committed, FileBacked> {
...
+            wait_nr: u32,
+            storage: &FileBacked,
+            operation_name: &str,
+        ) -> Result<(), FileIoError> {
+            loop {
+                match ring.submit_and_wait(wait_nr as usize) {
+                    Ok(_) => return Ok(()),
+                    Err(e) => {
+                        // Check if the error is EINTR (interrupted system call)
+                        if e.raw_os_error() == Some(libc::EINTR) {
```
demosdemon: Curious why you need libc or the raw error. Is the ErrorKind for these errors not [`ErrorKind::Interrupted`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.Interrupted)?

rkuris: Totally should use that. Done.

Cargo.lock
```diff
@@ -1964,9 +1964,9 @@ checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 
 [[package]]
 name = "libc"
-version = "0.2.176"
+version = "0.2.177"
```
rkuris, line 1967: Interesting that this got updated. I suppose we should be running `cargo update` periodically now.

demosdemon: This was updated because you added libc as a direct dependency. You've since removed that change; however, it's still updated in the lockfile and you can't easily downgrade it without completely reverting the lockfile.

I don't want to run cargo update periodically. I would rather dependabot do that but we've disabled that for whatever reason.

PR #1355, opened 2025-10-10
chore(nodestore): access persisted nodestores

RodrigoVillar:
This PR breaks up #1346 by factoring out the `NodeStore` changes. 

This PR allows for persisted nodestores which are no longer in memory to be retrieved again. This is useful for `RootStore` as this is what it'll use to construct prior revisions in the db `view()` method.

### DIFF
--- a/storage/src/nodestore/mod.rs
+++ b/storage/src/nodestore/mod.rs
@@ -158,8 +158,12 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
     /// address and hash, typically when reconstructing a [`NodeStore`] from
     /// a committed state. The `latest_nodestore` provides access to the underlying
     /// storage backend containing the persisted trie data.
+    ///
+    /// ## Panics
+    ///
+    /// Panics in debug builds if the hash of the node at `root_address` does
+    /// not equal `root_hash`.
     #[must_use]
-    #[allow(clippy::missing_panics_doc)]
     pub fn with_root(
         root_hash: HashType,
         root_address: LinearAddress,


storage/src/nodestore/mod.rs
```diff
@@ -151,6 +151,38 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
+    /// This constructor is used when you have an existing root node at a known
+    /// address and hash, typically when reconstructing a [`NodeStore`] from
+    /// a committed state.
+    pub fn new_committed(
+        root_hash: HashType,
+        root_address: LinearAddress,
+        storage: Arc<S>,
+    ) -> Self {
+        let mut header = NodeStoreHeader::new();
+        header.set_root_address(Some(root_address));
```
rkuris: Nit: create `NodeStoreHeader::with_root` to avoid `mut`.

Although it's not always possible to remove mut, the only point of it here is for construction of a new one, which should be a different function.

storage/src/nodestore/mod.rs
```diff
@@ -151,6 +151,38 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
             },
         }
     }
+
+    /// Create a new Committed [`NodeStore`] with a specified root node.
+    ///
+    /// This constructor is used when you have an existing root node at a known
+    /// address and hash, typically when reconstructing a [`NodeStore`] from
+    /// a committed state.
+    pub fn new_committed(
```
rkuris: Probably should name this `with_root`

storage/src/nodestore/mod.rs
```diff
@@ -151,6 +151,38 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
+        root_hash: HashType,
+        root_address: LinearAddress,
+        storage: Arc<S>,
+    ) -> Self {
+        let mut header = NodeStoreHeader::new();
+        header.set_root_address(Some(root_address));
+
```
rkuris, line 174: Seems like a good place for a `debug_assert`. If we hash the given root, we should get the hash provided. Something like this but less messy:

```
debug_assert_eq!(root_hash, hash_node(read_node, nodestore.read_node(root_address).expect("Root node read should succeed"));
```

You'd also have to create the nodestore first I think. debug_assert only because read_node isn't cheap, nor is calculating the hash, but this might help you debug some problems later.

firewood/src/db.rs
```diff
@@ -766,6 +766,54 @@ mod test {
...
+        let latest_root_hash = db.root_hash().unwrap().unwrap();
+        let latest_revision = db.revision(latest_root_hash).unwrap();
+
+        let latest_value = latest_revision.val(key).unwrap().unwrap();
+        assert_eq!(new_value, latest_value.as_ref());
+
+        let node_store = NodeStore::new_committed(
+            root_hash.into_hash_type(),
+            root_address,
+            latest_revision.underlying_storage(),
```
rkuris: Let's keep the surface area of the API to storage smaller here. I don't think firewood should be able to see or manipulate the storage implementations, but it knows about nodestores...

More specifically, move the call to `underlying_storage` to `new_committed` to reduce the visibility of it.

firewood/src/db.rs
```diff
@@ -766,6 +766,54 @@ mod test {
...
+        let new_value = b"new_value";
+        let batch = vec![BatchOp::Put {
+            key,
+            value: new_value,
+        }];
+
+        let proposal = db.propose(batch).unwrap();
+        proposal.commit().unwrap();
+
+        // Finally, reopen the database and make sure that we can retrieve the first revision
```
rkuris, line 798: 🏆 great test

firewood/src/db.rs
```diff
@@ -766,6 +766,54 @@ mod test {
         });
     }
 
+    #[test]
+    fn test_committed_nodestore_retrieval() {
```
rkuris: nit: name this `resurrect_unpersisted_root` or something. You don't really need to say `test_` unless you want to.

storage/src/nodestore/mod.rs
```diff
@@ -151,6 +151,38 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
+    /// Create a new Committed [`NodeStore`] with a specified root node.
+    ///
+    /// This constructor is used when you have an existing root node at a known
+    /// address and hash, typically when reconstructing a [`NodeStore`] from
+    /// a committed state.
+    pub fn new_committed(
+        root_hash: HashType,
+        root_address: LinearAddress,
+        storage: Arc<S>,
```
rkuris: maybe `nodestore: NodeStore<T, S>`

storage/src/nodestore/mod.rs
RodrigoVillar: Done: 6181b73

storage/src/nodestore/mod.rs
RodrigoVillar: Done: 5c921ac

storage/src/nodestore/mod.rs
RodrigoVillar: Agree that this will help with debugging later; added here: a86f0c3

firewood/src/db.rs
RodrigoVillar: Great point; moved the storage operation into `with_root()` here: d6bb55cd1b7a86806e4993038ddc3be99be92658

firewood/src/db.rs
RodrigoVillar: Renamed to `test_resurrect_unpersisted_root()` here: 59f3333. While I agree that the `test_` prefix is not necessary, I would prefer to be consistent with the naming of the other tests in `db.rs` (which mostly have the `test_` prefix).

storage/src/nodestore/mod.rs
RodrigoVillar: Done: d6bb55cd1b7a86806e4993038ddc3be99be92658 (named to `latest_nodestore` since the common pattern will be to pass in the most recent revision's nodestore here).

storage/src/nodestore/mod.rs
```diff
@@ -151,6 +151,46 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
     }
+
+    /// Create a new Committed [`NodeStore`] with a specified root node.
+    ///
+    /// This constructor is used when you have an existing root node at a known
+    /// address and hash, typically when reconstructing a [`NodeStore`] from
+    /// a committed state. The `latest_nodestore` provides access to the underlying
+    /// storage backend containing the persisted trie data.
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
```
rkuris: nit: This is a bit tricky to remove because panics only can happen for debug builds. I would prefer to. have documented the panics (specifically mentioning they only can occur in debug builds and why).

RodrigoVillar: Great point; updated the documentation to talk about panics here: 48062f5

PR #1356, opened 2025-10-13
feat(7/7): replace Path with various newtypes

demosdemon:
`Path` has been replaced with the components added in previous PRs inclusing using `[PathComponent]` that make it easier to reason about paths in code.


demosdemon:
ultra stale

PR #1359, opened 2025-10-13
feat: Use PathComponent in proofs

demosdemon:
This change primarily replaces the soon to be removed `Path` struct with `PathBuf` (which is a type alias for the smallvec of `PathComponent`) within the proof structure and related hashing traits. Changing only the proof and related components was less invasive changing `LeafNode`, `BranchNode`, or the iterator states. Those changes will come in future updates.


storage/src/path/buf.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use crate::{IntoSplitPath, TriePath};
+
+use super::PathComponent;
+
+/// An owned buffer of path components.
+pub type PathBuf = smallvec::SmallVec<[PathComponent; 32]>;
+
+/// A trie path represented as a slice of path components, either borrowed or owned.
+pub enum PartialPath<'a> {
```
rkuris, line 10: nit: Not quite sure why we couldn't use a `Cow` here

demosdemon: the standard library `Cow` doesn't allow us to swap in our own `Owned` variants. `Cow<'_, [PathComponent]>` would force the [`Owned`](https://doc.rust-lang.org/stable/src/alloc/borrow.rs.html#189) variant to be a `Vec` and not `SmallVec` (or any other container).

This is because the `ToOwned` impl for `[T]` has [`Owned` set to `Vec<T>`](https://doc.rust-lang.org/stable/src/alloc/slice.rs.html#835)

PR #1360, opened 2025-10-13
chore: upgrade dependencies

demosdemon:
The transitive upgrade on `generic-array` caused a large number of deprecation warnings and is beginning to interfere with CI on pull requests. To resolve this, I replaced the generic-array with a plain array in TrieHash and annotated the remaining instances with the lint expectation.


demosdemon:
> Sad to see GenericArray go because the idea here is that we might need to make hashes larger at some point.

Ideally by that point the ecosystem under surrounding the `digest` crate has upgraded to generic-array 1.x (which is why 0.14 is deprecated).

PR #1363, opened 2025-10-14
feat: add TrieNode trait and related functionality

demosdemon:
`TrieNode` is a generic trait that abstracts over the behavior of nodes in a fixed-arity radix trie. This trait allows for different structures to represent tries while enabling similar operations on them.

`TrieNode` is merkle-aware, meaning it can represent nodes that are a mix of locally described children or children referenced by only their hash. This is accomplished through `TrieEdgeState` and is used later in proof verification.

The `HashedTrieNode` trait is for post-hashed nodes and is distinctly different from the hash described on `TrieNode`. This trait is intended to represent the computed hash of the node itself; whereas, the hash accessed via `TrieNode` is the hash of a child node as seen from its parent. While they are usually the same, they can differ during intermediate states of a trie or during proof verification when some information is incomplete.

`KeyValueTrieRoot` is an implementation of `TrieNode` that only represents a key-value store and is not merkleized. This is really a separate change and is a component of proof verification but is also needed here as it is used to test and verify the iterator.

### DIFF
--- a/storage/src/path/buf.rs
+++ b/storage/src/path/buf.rs
@@ -130,6 +130,10 @@ impl<'a> IntoSplitPath for &'a PartialPath<'_> {
 }
 
 /// A RAII guard that resets a path buffer to its original length when dropped.
+///
+/// The guard exposes append-only methods to add components to the path buffer
+/// but not remove them without dropping the guard. This ensures that the guard
+/// will always restore the path buffer to its original state when dropped.
 #[must_use]
 pub struct PathGuard<'a> {
     buf: &'a mut PathBuf,
@@ -178,6 +182,13 @@ impl<'a> PathGuard<'a> {
         fork.extend(path.components());
         fork
     }
+
+    /// Appends the given component to the path buffer.
+    ///
+    /// This component will be removed when the guard is dropped.
+    pub fn push(&mut self, component: PathComponent) {
+        self.buf.push(component);
+    }
 }
 
 impl std::ops::Deref for PathGuard<'_> {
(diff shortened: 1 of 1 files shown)


storage/src/path/buf.rs
```diff
@@ -128,3 +128,68 @@ impl<'a> IntoSplitPath for &'a PartialPath<'_> {
...
+        }
+    }
+
+    /// Creates a new guard that will reset this guard's buffer to its current
+    /// length when dropped.
+    ///
+    /// This allows for nested guards that can be used in recursive algorithms.
+    pub fn fork(&mut self) -> PathGuard<'_> {
+        PathGuard::new(self.buf)
+    }
```
rkuris, line 171: Why isn't this the `clone` implementation?

storage/src/path/buf.rs
```diff
@@ -128,3 +128,68 @@ impl<'a> IntoSplitPath for &'a PartialPath<'_> {
         self
     }
 }
+
+/// A RAII guard that resets a path buffer to its original length when dropped.
```
rkuris, line 132: Actually, this isn't completely true. If you create a PathGuard and then shorten PathBuf, it will remain shortened after dropping it, as the truncate will have no effect.

rkuris (approved):

storage/src/path/buf.rs
demosdemon: `Clone` returns the same lifetime and the original type. This returns a new lifetime bound to self. 

You also cannot `Clone` mutable references.

PR #1365, opened 2025-10-15
feat: add hashed key-value trie implementation

demosdemon:
This expands on the previously added key-value trie by adding a hashed variant and mechanism to convert from unhashed to hashed.

### DIFF
--- a/storage/src/tries/kvp.rs
+++ b/storage/src/tries/kvp.rs
@@ -427,11 +427,21 @@ mod tests {
 
     use super::*;
 
-    /// in constant context, convert an ASCII hex string to a byte array
+    /// In constant context, convert an ASCII hex string to a byte array.
+    ///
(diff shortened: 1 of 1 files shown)


storage/src/node/branch.rs
```diff
@@ -193,14 +193,23 @@ mod ethhash {
...
-    #[derive(Clone, Debug)]
+    #[derive(Clone)]
     pub enum HashOrRlp {
         Hash(TrieHash),
         // TODO: this slice is never larger than 32 bytes so smallvec is probably not our best container
         // the length is stored in a `usize` but it could be in a `u8` and it will never overflow
         Rlp(SmallVec<[u8; 32]>),
     }
 
+    impl std::fmt::Debug for HashOrRlp {
```
rkuris, line 207: nit: would be nice to explain why it was hand rolled (I had to look)

PR #1367, opened 2025-10-15
feat: add TriePathIter for iterating over the nodes along a path

demosdemon:
This is used to walk along a path to yield all of the edges. In proof verification, this is used to verify the paths of the proof are valid.


demosdemon:
> Let's chat about this and how it differs from the existing MerkleNodeIter. The existing code is tested heavily as it forms the basis of MerkleKeyValueIter.

`MerkleNodeIter` requires a `TrieReader` and yields nodes as `SharedNode`s. They have different enough functionality that it isn't easy to replace one with the other. But, it would be possible to make `TriePathIter` replace `MerkleNodeIter` with some adjustments. However, the inverse wouldn't be true due to the dependency on `TrieReader` and `SharedNode`.

demosdemon:
stale

PR #1369, opened 2025-10-15
feat: build trie from Proofs

demosdemon:
Key proofs are a sequence of trie nodes that follow a linear path through the trie. This means we can reconstruct a narrow view of the trie over that linear path. In this narrow view, each node can refer to another full node or just its hash.

This resulting trie can (and will in an upcoming change) be hashed to generate the same root hash. While merging with the key-value tries from #1363 and #1365, we can iteratively verify the hash of each layer and detect early if any node is incomplete.

The `Remote` edges can also point outside of the key range. We can use these remote edges to identify holes in our overall trie and continue synchronizing down those paths.

### DIFF
--- a/firewood/src/proofs/tests.rs
+++ b/firewood/src/proofs/tests.rs
@@ -3,6 +3,10 @@
 
 #![expect(clippy::unwrap_used, clippy::indexing_slicing)]
 
+use firewood_storage::{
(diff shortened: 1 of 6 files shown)


rkuris (approved):
LGTM

demosdemon:
stale

PR #1378, opened 2025-10-20
feat: add more info for IO error message

qusuyan:


rkuris:
@qusuyan this commit is not signed, and can't be merged until it's signed.

qusuyan:
> @qusuyan this commit is not signed, and can't be merged until it's signed.

Done! @rkuris

PR #1380, opened 2025-10-20
chore: update Go to 1.24.9

RodrigoVillar:
AvalancheGo currently uses go1.24.9 while Coreth is also has a PR open to use go.1.24.9 [[ref](https://github.com/ava-labs/coreth/pull/1352)]. This PR updates our go version to 1.24.9.

### DIFF
--- a/ffi/go.mod
+++ b/ffi/go.mod
@@ -2,7 +2,7 @@ module github.com/ava-labs/firewood/ffi
 
 go 1.24
 
-toolchain go1.24.8
+toolchain go1.24.9
 
 require (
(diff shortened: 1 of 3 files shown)


RodrigoVillar:
@alarso16 I don't think so; referencing #1361 as an example.

rkuris:
Anyone have any suggestions to help keep these in sync later?

RodrigoVillar:
@rkuris We could use something similar to https://github.com/ava-labs/firewood/blob/main/.github/scripts/verify_golangci_yaml_changes.sh

PR #1382, opened 2025-10-22
Ensure golang input is up-to-date for nix flake

maru-ava:
The PR introducing the nix flake missed ensuring that the golang input from avalanchego reflected the update for 1.24.9. This change remedies that oversight and ensures that CI will catch it in the future.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -253,9 +253,9 @@ jobs:
       - name: Check golang flake input is up to date
         working-directory: ffi
         run: |
-          nix flake lock --update-input golang
+          nix flake update golang
           if ! git diff --exit-code flake.lock; then
-            echo "Error: ffi/flake.lock is out of date. Run 'nix flake lock --update-input golang' in the ffi directory and commit the changes."
+            echo "Error: ffi/flake.lock is out of date. Run 'nix flake update golang' in the ffi directory and commit the changes."
             exit 1
           fi
       - name: Test build equivalency between Nix and Cargo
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -250,6 +250,14 @@ jobs:
       - uses: actions/checkout@v4
       - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20
       - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 #v13
+      - name: Check golang flake input is up to date
```
maru-ava, line 253: The [just](https://github.com/ava-labs/firewood/pull/1345) PR simplifies local usage.

.github/workflows/ci.yaml
```diff
@@ -250,6 +250,14 @@ jobs:
...
+        working-directory: ffi
+        run: |
+          nix flake update golang
```
rkuris, line 256: Does this command update golang to the latest version? If so, that isn't what we want -- we want to be in sync with coreth.

rkuris: I'd like to figure this out myself but I couldn't figure out where to find the source or steps that this actually performs. Can you point me to those as well?

maru-ava: This ensures that the lockfile is consistent with the version of the `golang` flake input. 

Maybe it makes sense to merge the just PR first instead of second? Using just makes it both more obvious what is going on and makes it trivial to execute locally.

maru-ava: Fixed the issues with https://github.com/ava-labs/firewood/pull/1345 (command naming and broken just invocation)

maru-ava:
Closing in favor of #1345

PR #1388, opened 2025-10-24
chore(ffi/firewood): remove `RootStore` generics

RodrigoVillar:
While going through a first pass of adding `ProductionStore`, I found that adding a `RootStore` generic to `Db` resulted in the generic bubbling up to the `ffi` layer, which caused a myriad of issues from the Golang side (and a massive refactoring).

This PR removes the `RootStore` generic and instead sets the `root_store` field of the `RevisionManager` to be a `Box` pointer.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -397,6 +397,7 @@ mod test {
 
     impl<T: Iterator> IterExt for T {}
 
+    #[cfg(test)]
     impl Db {
         /// Extract the root store by consuming the database instance.
         /// This is primarily used for reopening or replacing the database with the same root store.
@@ -1112,9 +1113,7 @@ mod test {
             let path = self.path();
             let TestDb { db, tmpdir } = self;
 
-            // Explicit scope ensures the database is fully dropped before we
-            // attempt to open a new database instance
-            let root_store = { db.into_root_store() };
+            let root_store = db.into_root_store();
 
             let dbconfig = DbConfig::builder().truncate(false).build();
             let db = Db::with_root_store(path, dbconfig, root_store).unwrap();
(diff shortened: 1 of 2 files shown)


firewood/src/root_store.rs
```diff
@@ -20,7 +24,7 @@ pub struct RootStoreError {
     pub source: Box<dyn std::error::Error + Send + Sync>,
 }
 
-pub trait RootStore {
+pub trait RootStore: Debug {
```
RodrigoVillar, line 27: Note to reviewer: added the `Debug` trait here since the `RevisionManager` has a `#[derive(Debug)]` macro on it.

firewood/src/db.rs
```diff
@@ -1093,48 +1097,42 @@ mod test {
...
         fn reopen(self) -> Self {
             let path = self.path();
-            let root_store = self.manager.root_store();
-            drop(self.db);
-            let dbconfig = DbConfig::builder().truncate(false).build();
+            let TestDb { db, tmpdir } = self;
+
+            // Explicit scope ensures the database is fully dropped before we
+            // attempt to open a new database instance
+            let root_store = { db.into_root_store() };
```
RodrigoVillar: Note to reviewer: since `into_root_store()` implicitly drops `db`, I added the scope here to make this drop more clear (since `drop()` can't be used afterwards). Open to suggestions on how to make this better.

rkuris: This method is never used in production so into_root_store (or the whole impl block) should be marked `#[cfg(test)]` to make it clear this is only used during testing and avoid including this code in debug builds (it will get optimized away in release builds, but no point building it in debug builds).

You also don't need these brackets. When I see `into` I realize it's going to destroy the db without a comment.

rkuris (approved):
Please make the requested small changes and merge.
No additional review is needed unless you change something else :)

PR #1389, opened 2025-10-27
fix(storage): flush freelist early to prevent corruption

rkuris:
Fix a critical bug where nodes were written to storage before the freelist was updated, leaving the database in an inconsistent state if an I/O error occurred during node persistence.

Problem:
When persisting nodes, if the freelist flush happened after node writes, a crash or I/O error between node writes and freelist flush would leave allocated storage space untracked in the freelist. On recovery, these areas could be reallocated, leading to data corruption or loss.

Solution:
- Flush freelist immediately after allocating node addresses but before writing any node data
- Add NodeAllocator::flush_freelist() to enable explicit freelist updates
- Add flush_freelist_from() to persist a specific header's freelist state

Additional improvements:
- Refactor to use bumpalo bump allocator with bounded memory for node serialization (prevents unbounded memory growth)
- Extract common serialization and batching logic into shared helper functions (serialize_node_to_bump, process_unpersisted_nodes)
- Eliminate ~90 lines of code duplication between io-uring and non-io-uring paths while ensuring both use identical logic

### DIFF
--- a/.github/.golangci.yaml.patch
+++ b/.github/.golangci.yaml.patch
@@ -1,6 +1,6 @@
---- .github/.golangci.yaml	2025-09-10 10:51:41
-+++ ffi/.golangci.yaml	2025-08-22 11:42:25
-@@ -69,8 +69,6 @@
+--- .github/.golangci.yaml	2025-10-29 08:30:25
++++ ffi/.golangci.yaml	2025-10-29 08:30:25
+@@ -70,8 +70,6 @@
        rules:
          packages:
            deny:
@@ -9,7 +9,7 @@
              - pkg: github.com/golang/mock/gomock
                desc: go.uber.org/mock/gomock should be used instead.
              - pkg: github.com/stretchr/testify/assert
-@@ -85,29 +83,10 @@
+@@ -86,29 +84,10 @@
      forbidigo:
        # Forbid the following identifiers (list of regexp).
        forbid:
@@ -39,7 +39,7 @@
      revive:
        rules:
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bool-literal-in-expr
-@@ -171,17 +150,6 @@
+@@ -172,17 +151,6 @@
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#useless-break
          - name: useless-break
            disabled: false
@@ -57,7 +57,7 @@
      tagalign:
        align: true
        sort: true
-@@ -227,7 +195,7 @@
+@@ -245,7 +213,7 @@
          - standard
          - default
          - blank
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,7 @@ metrics-util = "0.20.0"
 nonzero_ext = "0.3.0"
 rand_distr = "0.5.1"
 sha2 = "0.10.9"
-smallvec = "1.15.1"
+smallvec = { version = "1.15.1", features = ["write", "union", "const_new"] }
 test-case = "3.3.1"
 thiserror = "2.0.17"
 
--- a/ffi/.golangci.yaml
+++ b/ffi/.golangci.yaml
@@ -60,6 +60,7 @@ linters:
     - tagalign
     - testifylint
     - unconvert
+    - usetesting
     - unparam
     - unused
     - usestdlibvars
@@ -167,6 +168,14 @@ linters:
       disable:
         - go-require
         - float-compare
+    usetesting:
+      os-create-temp:     true # Disallow `os.CreateTemp("", ...)`
+      os-mkdir-temp:      true # Disallow `os.MkdirTemp()`
+      os-setenv:          true # Disallow `os.Setenv()`
+      os-temp-dir:        true # Disallow `os.TempDir()`
+      os-chdir:           true # Disallow `os.Chdir()`
+      context-background: true # Disallow `context.Background()`
+      context-todo:       true # Disallow `context.TODO()`
     unused:
       # Mark all struct fields that have been written to as used.
       # Default: true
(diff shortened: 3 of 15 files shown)


storage/src/node/mod.rs
```diff
@@ -149,6 +149,21 @@ impl ExtendableBytes for Vec<u8> {
     }
 }
 
+impl ExtendableBytes for bumpalo::collections::Vec<'_, u8> {
+    fn extend<T: IntoIterator<Item = u8>>(&mut self, other: T) {
+        std::iter::Extend::extend(self, other);
+    }
+    fn reserve(&mut self, reserve: usize) {
+        self.reserve(reserve);
```
demosdemon: extremely pedantic nit:

I'm confident that this is not triggering recursion because there is a built-in lint in [rustc](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#unconditional-recursion) that warns about unconditional recursion (why it's a warning and not an error, idk, probably something about observable effects on the stack); but, I do think that this should use the fully qualified path like the other methods for consistency.

storage/src/nodestore/persist.rs
```diff
@@ -84,10 +84,22 @@ impl<T, S: WritableStorage> NodeStore<T, S> {
...
+    ///
+    /// This function is used to ensure that the freelist is advanced after allocating
+    /// nodes for writing. This allows the database to be recovered from an I/O error while
+    /// persisting a revision to disk.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`FileIoError`] if the free list cannot be written to storage.
+    #[fastrace::trace(name = "firewood.flush_freelist")]
+    pub fn flush_freelist_from(&self, header: &NodeStoreHeader) -> Result<(), FileIoError> {
```
demosdemon: This api feels weird and I'm not sure how much I like it being exposed. What guarantees do we have that the header passed in is a valid header and was born from this nodestore? If any addresses are wrong (or from a different instance), this will corrupt the database. I do not see it used anywhere else (yet?), so maybe remove the `pub` for now. If it does need to be exported, the exported method should verify the lineage of the header.

storage/src/nodestore/alloc.rs
```diff
@@ -358,6 +358,13 @@ impl<S: WritableStorage> NodeAllocator<'_, S> {
 
         Ok(())
     }
+
+    pub fn flush_freelist(&mut self) -> Result<(), FileIoError> {
```
demosdemon, line 362: did you intend to use `flush_freelist_from` here?

storage/src/nodestore/persist.rs
```diff
@@ -290,6 +394,67 @@ impl<S: WritableStorage + 'static> NodeStore<Committed, S> {
...
+    }
+}
+
+#[cfg(feature = "io-uring")]
+use std::pin::Pin;
+
+#[cfg(feature = "io-uring")]
+#[derive(Clone, Debug)]
+struct PinnedBufferEntry<'a> {
+    pinned_buffer: Pin<&'a [u8]>,
```
demosdemon: the `Pin` here is not necessary. an immutable reference is always pinned. The arena allocator effectively pins the memory in place for the entirety of the immutable reference, and is required to pin the memory because of immutable rules. Pinning is only relevant for mutable pointers because you need to guarantee that writers will not move the pointer. Immutable pointers already provide that guarantee by the nature of being immutable.

storage/src/nodestore/persist.rs
```diff
@@ -463,22 +559,69 @@ impl NodeStore<Committed, FileBacked> {
...
+    storage: &FileBacked,
+    completion_queue: io_uring::cqueue::CompletionQueue<'_>,
+    saved_pinned_buffers: &mut [Option<PinnedBufferEntry<'_>>],
+    cached_nodes: &mut Vec<MaybePersistedNode>,
+) -> Result<(), FileIoError> {
+    for entry in completion_queue {
+        // user data contains the index of the entry in the saved_pinned_buffers array
+        let item = entry.user_data() as usize;
+        let pbe = saved_pinned_buffers
+            .get_mut(item)
```
demosdemon: you could probably call `take` here for simplicity. If you return an error, everything within `saved_pinned_buffers` will fall off the stack regardless.

storage/src/nodestore/persist.rs
```diff
@@ -290,6 +394,67 @@ impl<S: WritableStorage + 'static> NodeStore<Committed, S> {
     }
 }
 
+#[cfg(feature = "io-uring")]
```
demosdemon: there is a lot of io-uring specific stuff here and having `use` statements in the middle of the file is weird. The io-uring related code should be moved into a dedicated module which is conditionally included instead of every item. That should result in only 2-3 conditionally included statements instead of 10+.

storage/src/nodestore/alloc.rs
rkuris: No. I only use that when you have a special `NodeAllocator` and are in the middle of flushing a bunch of nodes. The final flush should use the regular `flush_freelist` which uses the freelist in the `NodeStore`.

storage/src/nodestore/persist.rs
```diff
@@ -199,30 +279,30 @@ impl<S: WritableStorage + 'static> NodeStore<Committed, S> {
...
+                    file_backed.flush_nodes_io_uring()?
+                } else {
+                    self.flush_nodes_generic()?
+                }
+            }
+            #[cfg(not(feature = "io-uring"))]
+            {
+                self.flush_nodes_generic()?
+            }
+        };
```
demosdemon, line 300: I would probably leave the `downcast_to_file_backed` method as this removes the documentation as to why it exists and looks more unwieldily. But I think I'm forming an idea of a better way of doing this that doesn't require this `dyn Any` downcast so maybe not worth changing.

Issue #1393, opened 2025-10-29
Chunking large databases


alarso16:
For many use cases, having a single file of, say, 500GB is somewhat prohibitive, especially for snapshotting state. Either as a config option, or by default, it would be helpful to set how the database should be chunked. The ava-labs infrastructure team encouraged file sizes of approximately 20GB. The scheme in which this is decided (maximum file size, number of files, etc.) is unknown at this time. More planning should be put into this.

Likely, this would change the firewood interface to just accept a folder, and contain all files (like the root store) within this folder.

RodrigoVillar:
If we do go about changing the Firewood interface to accept a folder, we'll need to check for (and perhaps optimize) RootStore as well [[ref](https://github.com/ava-labs/firewood/pull/1395#discussion_r2504769920)].

PR #1395, opened 2025-10-29
feat(firewood/ffi): add `FjallStore`

RodrigoVillar:
This PR adds `FjallStore`, an implementation of `RootStore` that allows for persistent storage of revisions (in contrast to `MockStore`, which keeps everything in-memory). The main changes of this PR are as follows:

- Adds `FjallStore` implementation
- Extends the `DbConfig` to take in a path to where `FjallStore` will be stored
- Extends the FFI database config to enable archival support
- Adds `FjallStore` tests in both Rust and Golang

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -6,7 +6,7 @@ use std::path::PathBuf;
 use firewood::{
     db::{Db, DbConfig},
     manager::RevisionManagerConfig,
-    v2::api::{self, ArcDynDbView, Db as _, DbView, HashKey, HashKeyExt, KeyType},
+    v2::api::{self, ArcDynDbView, Db as _, DbView, HashKey, HashKeyExt, IntoBatchIter, KeyType},
 };
 
 use crate::{BorrowedBytes, CView, CreateProposalResult, KeyValuePair, arc_cache::ArcCache};
@@ -234,6 +234,18 @@ impl DatabaseHandle {
     pub(crate) fn clear_cached_view(&self) {
         self.cached_view.clear();
     }
+
+    pub(crate) fn merge_key_value_range(
+        &self,
+        first_key: Option<impl KeyType>,
+        last_key: Option<impl KeyType>,
+        key_values: impl IntoIterator<Item: api::KeyValuePair>,
+    ) -> Result<CreateProposalResult<'_>, api::Error> {
+        CreateProposalResult::new(self, || {
+            self.db
+                .merge_key_value_range(first_key, last_key, key_values)
+        })
+    }
 }
 
 impl From<Db> for DatabaseHandle {
@@ -250,11 +262,11 @@ impl<'db> CView<'db> for &'db crate::DatabaseHandle {
         self
     }
 
-    fn create_proposal<'kvp>(
+    fn create_proposal(
         self,
-        values: impl AsRef<[KeyValuePair<'kvp>]> + 'kvp,
+        values: impl IntoBatchIter,
     ) -> Result<firewood::db::Proposal<'db>, api::Error> {
-        self.db.propose(values.as_ref().iter())
+        self.db.propose(values)
     }
 }
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -244,17 +244,6 @@ jobs:
         # cgocheck2 is expensive but provides complete pointer checks
         run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=firewood go test -race ./...
 
-  ffi-nix:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20
-      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 #v13
-      - name: Check that FFI flake is up-to-date
-        run: ./scripts/run-just.sh check-ffi-flake
-      - name: Test nix build of Golang FFI bindings
-        run: ./scripts/run-just.sh test-ffi-nix
-
   firewood-ethhash-differential-fuzz:
     needs: build
     runs-on: ubuntu-latest
--- a/.github/workflows/ffi-nix.yaml
+++ b/.github/workflows/ffi-nix.yaml
@@ -1,0 +1,18 @@
+name: ffi-nix
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  ffi-nix:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20
+      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 #v13
+      - name: Check that FFI flake is up-to-date
+        run: ./scripts/run-just.sh check-ffi-flake
+      - name: Test nix build of Golang FFI bindings
+        run: ./scripts/run-just.sh test-ffi-nix
+
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -1290,8 +1290,8 @@ struct HashResult fwd_db_verify_and_commit_change_proof(const struct DatabaseHan
  * concurrently. The caller must ensure exclusive access to the proof context
  * for the duration of the call.
  */
-struct HashResult fwd_db_verify_and_commit_range_proof(const struct DatabaseHandle *_db,
-                                                       struct VerifyRangeProofArgs _args);
+struct HashResult fwd_db_verify_and_commit_range_proof(const struct DatabaseHandle *db,
+                                                       struct VerifyRangeProofArgs args);
 
 /**
  * Verify a change proof and prepare a proposal to later commit or drop.
@@ -1342,8 +1342,8 @@ struct VoidResult fwd_db_verify_change_proof(const struct DatabaseHandle *_db,
  * concurrently. The caller must ensure exclusive access to the proof context
  * for the duration of the call.
  */
-struct VoidResult fwd_db_verify_range_proof(const struct DatabaseHandle *_db,
-                                            struct VerifyRangeProofArgs _args);
+struct VoidResult fwd_db_verify_range_proof(const struct DatabaseHandle *db,
+                                            struct VerifyRangeProofArgs args);
 
 /**
  * Frees the memory associated with a `ChangeProofContext`.
@@ -1874,7 +1874,7 @@ struct ProposalResult fwd_propose_on_proposal(const struct ProposalHandle *handl
  * concurrently. The caller must ensure exclusive access to the proof context
  * for the duration of the call.
  */
-struct NextKeyRangeResult fwd_range_proof_find_next_key(struct RangeProofContext *_proof);
+struct NextKeyRangeResult fwd_range_proof_find_next_key(struct RangeProofContext *proof);
 
 /**
  * Deserialize a `RangeProof` from bytes.
@@ -1932,7 +1932,7 @@ struct ValueResult fwd_range_proof_to_bytes(const struct RangeProofContext *proo
  * concurrently. The caller must ensure exclusive access to the proof context
  * for the duration of the call.
  */
-struct VoidResult fwd_range_proof_verify(struct VerifyRangeProofArgs _args);
+struct VoidResult fwd_range_proof_verify(struct VerifyRangeProofArgs args);
 
 /**
  * Get the root hash of the latest version of the database
--- a/ffi/proofs.go
+++ b/ffi/proofs.go
@@ -8,6 +8,7 @@ package ffi
 import "C"
 
 import (
+	"encoding"
 	"errors"
 	"fmt"
 	"runtime"
@@ -19,10 +20,44 @@ var (
 	errEmptyTrie   = errors.New("a range proof was requested on an empty trie")
 )
 
+var (
+	_ encoding.BinaryMarshaler   = (*RangeProof)(nil)
+	_ encoding.BinaryUnmarshaler = (*RangeProof)(nil)
+)
+
 // RangeProof represents a proof that a range of keys and their values are
 // included in a trie with a given root hash.
+//
+// RangeProofs can be created via [Database.RangeProof] and are marshallable via
+// [encoding.BinaryMarshaler] and [encoding.BinaryUnmarshaler]. They can be
+// verified independent of a database via [RangeProof.Verify] or with a database
+// via [Database.VerifyRangeProof]. Verified range proofs can be committed to a
+// database via [Database.VerifyAndCommitRangeProof] (where verification will
+// be skipped if it was already verified). Verifying a range proof with a
+// database will optimistically prepare a proposal that can be committed later.
 type RangeProof struct {
+	// handle is an opaque pointer to the range proof within Firewood. It should be
+	// passed to the C FFI functions that operate on range proofs
+	//
+	// It is not safe to call these methods with a nil handle.
+	//
+	// Calls to `C.fwd_free_range_proof` will invalidate this handle, so it
+	// should not be used after those calls.
+	//
+	// Calls to `C.fwd_db_verify_range_proof` will cause the range proof to
+	// build and retain ownership of an embedded proposal, which also retains
+	// a reference to the database. Therefore, while the range proof owns an
+	// embedded proposal, the database must be kept alive. The proposal and
+	// reference to the database are released after calling
+	// `C.fwd_db_verify_and_commit_range_proof` or `C.fwd_free_range_proof`.
 	handle *C.RangeProofContext
+
+	// keepAliveHandle keeps the database alive while this range proof
+	// owns an embedded proposal. It is initialized when the range proof is
+	// verified with a database handle ([Database.VerifyRangeProof]) and not by
+	// unmarshalling or when [RangeProof.Verify] is used. It is disowned after
+	// [Database.VerifyAndCommitRangeProof] or [RangeProof.Free].
+	keepAliveHandle databaseKeepAliveHandle
 }
 
 // ChangeProof represents a proof of changes between two roots for a range of keys.
@@ -93,6 +128,9 @@ func (p *RangeProof) Verify(
 // the proof is valid, a proposal containing the changes is prepared. The
 // call to [*Database.VerifyAndCommitRangeProof] will skip verification and commit the
 // prepared proposal.
+//
+// Because this method prepares a proposal, the database must be kept alive
+// until the proof is committed or freed.
 func (db *Database) VerifyRangeProof(
 	proof *RangeProof,
 	startKey, endKey Maybe[[]byte],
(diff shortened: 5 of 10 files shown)


firewood/src/root_store.rs
```diff
@@ -108,3 +109,83 @@ impl RootStore for MockStore {
...
+}
+
+impl Debug for FjallStore {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FjallStore")
+            .field("keyspace", &"<Keyspace>")
+            .field("items", &"<PartitionHandle>")
+            .finish()
+    }
+}
```
RodrigoVillar: Since `Keyspace` and `Partition` both don't implement the `Debug` trait, I manually implemented `Debug` for `FjallStore`. Open to any suggestions on how to make this better though.

RodrigoVillar:
Note to reviewers: I've marked the PR as `DO NOT MERGE` since I'm working on getting benchmark results for Firewood + FjallStore when compared to LevelDB (and we should only merge once we have benchmark results). However, I figured we can get the code review process going since benchmarking can take place async.

ffi/firewood_test.go
```diff
@@ -1264,6 +1264,55 @@ func TestHandlesFreeImplicitly(t *testing.T) {
...
+		r.NoError(err)
+	}
+
+	// Close and reopen the database
+	r.NoError(db.Close(t.Context()))
+
+	db, err = New(dbFile, config)
+	r.NoError(err)
+
+	// Verify that we can access all revisions
```
rkuris, line 1298: In theory, we should be able to fix the original issue here and then this test will pass even if fjall is not doing anything. Can we configure the revision manager when creating the database to only save 5 revisions?

Additionally, we should test that all 10 are readable before reopening the database as well...

ffi/src/handle.rs
```diff
@@ -100,9 +113,19 @@ impl DatabaseHandle {
...
+            .map_err(|e| invalid_data(format!("root store path contains invalid utf-8: {e}")))?;
+
+        let root_store_dir = Some(root_store_path)
+            .filter(|s| !s.is_empty())
+            .map(PathBuf::from);
+
         let cfg = DbConfig::builder()
             .truncate(args.truncate)
             .manager(args.as_rev_manager_config()?)
+            .root_store_dir(root_store_dir)
```
rkuris, line 128: this works, but I think we need to switch to a directory per our discussion (probably a different PR).

firewood/src/db.rs
```diff
@@ -122,6 +122,9 @@ pub struct DbConfig {
     // TODO: Experimentally determine the right value for BatchSize.
     #[builder(default = UseParallel::BatchSize(8))]
     pub use_parallel: UseParallel,
+    /// `RootStore` directory path
+    #[builder(default = None)]
```
rkuris, line 129: Question: Isn't `None` the default if you don't specify?

firewood/src/root_store.rs
rkuris: You can use [derive_where](https://docs.rs/derive-where/latest/derive_where/) with [skip_inner](https://docs.rs/derive-where/latest/derive_where/#skipping-fields).

firewood/src/root_store.rs
```diff
@@ -108,3 +109,83 @@ impl RootStore for MockStore {
...
+    /// # Errors
+    ///
+    /// Will return an error if unable to create or open an instance of `FjallStore`.
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<FjallStore, RootStoreError> {
+        let keyspace = Config::new(path).open().map_err(|e| RootStoreError {
+            method: RootStoreMethod::New,
+            source: Box::new(e),
+        })?;
+        let items = keyspace
+            .open_partition("firewood", PartitionCreateOptions::default())
```
rkuris: Let's pull "firewood" out as a const for better readability as to what this is used for:
```
const FJALL_PARTITION_NAME: &str = "firewood";
```

firewood/src/root_store.rs
```diff
@@ -108,3 +109,83 @@ impl RootStore for MockStore {
...
+            .map_err(|e| RootStoreError {
+                method: RootStoreMethod::New,
```
rkuris: nit: in looking at this, I'm not convinced having the method here adds a ton of value and it increases the amount of code significantly. Your call though, I'm okay either way.

`let items = keyspace.open_partition(...)?;` is soo much easier to read.

storage/src/trie_hash.rs
```diff
@@ -105,6 +105,12 @@ impl TrieHash {
         self
     }
 
+    /// Converts `TrieHash` to an array of bytes.
+    #[must_use]
+    pub const fn to_bytes(&self) -> [u8; TRIE_HASH_LEN] {
+        self.0
```
rkuris: This feels like `impl Deref`, although perhaps you need it in const context in which case you can ignore this comment.

firewood/src/db.rs
RodrigoVillar: No; by removing the attribute here, we'll need to always call `root_store_dir(None)` on the builder whenever we have an instance of the `Db` builder. We also have similar attributes for `create_if_missing` and `truncate`, which are booleans (and should therefore be `false` by default):

https://github.com/ava-labs/firewood/blob/e775eeb178b36fdfb808540b59a4d91f6cc2dac6/firewood/src/db.rs#L110-L116

ffi/src/handle.rs
RodrigoVillar: Noted here: https://github.com/ava-labs/firewood/issues/1393#issuecomment-3504342934

firewood/src/root_store.rs
RodrigoVillar: Do you mean the `method` field of `RootStoreError`?

firewood/src/root_store.rs
RodrigoVillar: Ah thank you! Switched to using `derive_where` here: [9e3b9e2](https://github.com/ava-labs/firewood/pull/1395/commits/9e3b9e2d61858cf9fe7f59b15f028a704c6d5623)

storage/src/trie_hash.rs
RodrigoVillar: Yup I could use the existing `impl Deref` - done: 2691ac6

firewood/src/root_store.rs
RodrigoVillar: Done: 9d40bef

firewood/src/root_store.rs
rkuris: yes

rkuris: If you really really want to keep `method` then you could create a specialized function that works like `map_err` but removes the boilerplate. Then the code would be something like:

```
let items = keyspace
            .open_partition("firewood", PartitionCreateOptions::default())
            .map_err(wrap_with(RootStoreMethod::New))?;
```

This is still better as you avoid all the closure cognitive load.

RodrigoVillar: Assuming that we can figure out which method returned the error, I think it's better to remove the `RootStoreError` type altogether - PR which does that can be found here: https://github.com/ava-labs/firewood/pull/1446

RodrigoVillar: Resolving this as #1446 was merged in.

ffi/firewood_test.go
RodrigoVillar: Done here: [aa64746](https://github.com/ava-labs/firewood/pull/1395/commits/aa64746278d87c8a19d9e167a676389c740ea6f4).

Modifying the database to save only five revisions revealed an edge case that we didn't test - what happens to the oldest revision when we hit our `max_revisions` limit? It turns out that the oldest revision's nodes were reaped, resulting in `TestFjallStore()` failing. 

To fix this, I modified the `commit()` method so that if the oldest revision is referenced by `RootStore`, it's still removed from memory but its nodes are no longer reaped. I also added an additional test `test_fjall_store_with_capped_max_revisions()` on the Rust side to verify that this is indeed true now.

RodrigoVillar: Hmm on second thought, I wonder if it's reasonable to just add a `root_store_enabled` bool to the `RevisionManager` and add a revision's nodes to the free list only if `root_store_enabled == true`. In my mind, a revision being referenced by `RootStore` is (basically) logically equivalent to root store being enabled - this would save us from having to make a read to `RootStore` every time we commit.

firewood/src/manager.rs
```diff
@@ -167,7 +167,7 @@ impl RevisionManager {
     ///    The proposal's parent must be the last committed revision, otherwise the commit fails.
     ///    It only contains the address of the nodes that are deleted, which should be very small.
     /// 2. Revision reaping. If more than the maximum number of revisions are kept in memory, the
-    ///    oldest revision is reaped.
+    ///    oldest revision is reaped if not referenced by `RootStore`.
```
rkuris: This comment is confusing to me. It's not clear by what "referenced by `RootStore`" means.

What I'm reading this as is "if RootStore knows about this revision, then don't reap it" which isn't what you want. Reaping here refers to removing the revision from memory, which we want.

I'm also worried about when we don't have a RootStore. Aren't all revisions "not referenced by `RootStore`" in that case? Which implies we're never reaping any revisions, which clearly is wrong.

firewood/src/manager.rs
```diff
@@ -76,7 +76,7 @@ pub(crate) struct RevisionManager {
     /// stored in the filebacked storage.
     historical: RwLock<VecDeque<CommittedRevision>>,
     proposals: Mutex<Vec<ProposedRevision>>,
-    // committing_proposals: VecDeque<Arc<ProposedImmutable>>,
+    /// The list of revisions that are on disk by hash.
```
rkuris: This comment will become incorrect when deferred persistence is implemented. I think it's self-documenting already so the doc comment doesn't add much value.

The `committing_proposals` thing is dead code that should probably go away.

firewood/src/manager.rs
```diff
@@ -199,27 +200,40 @@ impl RevisionManager {
...
-            if let Some(oldest_hash) = oldest.root_hash().or_default_root_hash() {
-                self.by_hash
-                    .write()
-                    .expect("poisoned lock")
-                    .remove(&oldest_hash);
-            }
+            let oldest_hash = oldest
+                .root_hash()
+                .or_default_root_hash()
+                .expect("root hash should not be none");
```
rkuris: I'm not sure this is true. If you delete all keys and are not using eth compatible mode, this will be None, which will crash.

Interesting that this is not tested.

firewood/src/manager.rs
```diff
@@ -339,14 +353,29 @@ impl RevisionManager {
     }
 
     pub fn revision(&self, root_hash: HashKey) -> Result<CommittedRevision, RevisionManagerError> {
-        self.by_hash
+        if let Some(revision) = self
```
rkuris, line 340: This multi-step flow could use some block level comments.

```suggestion
        // First, check the in-memory revision manager
        if let Some(revision) = self
```

firewood/src/manager.rs
```diff
@@ -339,14 +353,29 @@ impl RevisionManager {
...
+            .root_store
+            .get(&root_hash)
+            .map_err(RevisionManagerError::RootStoreError)?
             .ok_or(RevisionManagerError::RevisionNotFound {
-                provided: root_hash,
-            })
+                provided: root_hash.clone(),
+            })?;
+
+        Ok(Arc::new(NodeStore::with_root(
```
rkuris: If you have 2 requests for the same revision, they will not share the NodeStore, and instead each one will get a separate nodestore with a reference count of 1.

One way to address this is to keep a cache of existing created NodeStores by hash. This might be a use case for weak arcs so those NodeStores will be dropped when unreferenced.

Maybe something like this:

In `RevisionManager`:
```rust
    by_rootstore: Mutex<HashMap<TrieHash, Weak<NodeStore<Committed, FileBacked>>>>,
```

Then, when looking for a revision, you lock, check, and populate that map (sloppy locking and unwraps to show the idea here):

```rust
    let nodestore_entry = self.by_rootstore.lock().unwrap().entry(&root_hash);
    let result = match nodestore_entry {
        Occupied(entry) => Some(entry.upgrade().unwrap_or_else(rootstore_lookup(&root_hash).unwrap())),
        Vacant(entry) => {
            if let Some(ns) = rootstore_lookup(&root_hash) {
                entry.insert(ns);
                Some(ns)
            } else {
                None
            }
        }
    }
```

This leaves you with the problem of removing stale hashmap entries that point to weak references that are deallocated, which can be solved by cleaning them up. Since you'll be traversing quite often, and the list will be small, maybe you avoid a hashmap and just walk a list looking for the hash you're interested in and dropping anything you find that's stale along the way.

rkuris (changes requested):
I'm not sure I love changing the reaping logic to fix this issue. The commit should have the ability to reap any revision at any time.

I think there are a few possible approaches here:

 1 - Part of reaping is adding it to the `RootStore`
 2 - Every commit adds to `RootStore` so entries are both in the RootStore and in the revision manager (preferred).

ffi/firewood_test.go
rkuris: Rather than using a bool, let's just use the rootstore that doesn't store anything. This way we don't have to sprinkle "if" statements in various places -- we always call the appropriate functions for the rootstore and the rootstore can choose to do what it wants, including nothing.

rkuris:
> Note to reviewers: I've marked the PR as `DO NOT MERGE` since I'm working on getting benchmark results for Firewood + FjallStore when compared to LevelDB (and we should only merge once we have benchmark results). However, I figured we can get the code review process going since benchmarking can take place async.

I think we determined that this PR does not affect performance, so I think this is mergeable at this point.

firewood/src/manager.rs
RodrigoVillar: Will undo this diff + defer this to a separate issue: https://github.com/ava-labs/firewood/issues/1459

firewood/src/manager.rs
RodrigoVillar: I learned this the hard way 😄(ref: aa64746). Fortunately, our tests failed immediately when I had this logic.

Fixed this here: 5bb0f67

firewood/src/manager.rs
RodrigoVillar: Reverted diff here: d848744

firewood/src/manager.rs
RodrigoVillar: Good call; done: f27fc4a

firewood/src/manager.rs
RodrigoVillar: To me, reaping means adding the nodes of the revision to the free list - our current code seems to support this definition: 

https://github.com/ava-labs/firewood/blob/8ad5e19b56487388ea66e86d1726e3c8999a3543/firewood/src/manager.rs#L214-L216

This is different than rolling out the revision from memory, which is what I think you're referring to. I've updated the documentation to make this clear: ccca265 - feel free to resolve if the documentation for `commit()` is clear now.

ffi/firewood_test.go
```diff
@@ -1244,6 +1244,58 @@ func TestHandlesFreeImplicitly(t *testing.T) {
...
+	for i := range numRevisions {
+		revision, err := db.Revision(revisionRoots[i])
+		r.NoError(err)
+
+		v, err := revision.Get(key)
+		r.NoError(err)
+
+		r.Equal(vals[i], v)
+		r.NoError(revision.Drop())
```
alarso16, line 1311: You should do this in a defer above, imo. This way, you won't get an error closing the database

firewood/src/manager.rs
RodrigoVillar: Thanks for the suggestion. As discussed offline, I went with using a weak table to cache the revisions. Once the strong count of a weak pointer becomes 0, the key-value pair will be (lazily) removed: 18e6f13

ffi/firewood_test.go
RodrigoVillar: @rkuris Now that we're on the same page when it comes to the concept of reaping, could you revisit your comment above and update it if needed?

ffi/firewood_test.go
RodrigoVillar: Done: 5665a09

ffi/firewood_test.go
rkuris: Yes, I think you'll need to know whether the RootStore implementation requires keeping the old roots around. This isn't something that should not be owned by the RevisionManager, since they can get out of sync or not be initialized properly, it should be owned by the RootStore.

You can do that easily with a trait associated constant, maybe named ALLOW_SPACE_REUSE or something:

```rust
trait RootStore {
    const ALLOWSPACE_SPACE_REUSE: bool;
    ...
}
```

Then, each implementation of `RootStore` will require defining the value of that const.

firewood/src/manager.rs
```diff
@@ -338,15 +340,56 @@ impl RevisionManager {
...
+
+        let nodestore = Arc::new(NodeStore::with_root(
+            root_hash.clone().into_hash_type(),
+            root_address,
+            self.current_revision(),
+        ));
+
+        // Cache the nodestore (stored as a weak reference).
+        self.by_rootstore
+            .lock()
```
rkuris: There's a race here (not a bug, but it wastes time if two callers hit here at the same time).

Reuse the lock from above; earlier, do:

```
let cache_guard = self.by_rootstore.lock().expect(...);
```

Then, leave `cache_guard` around and reuse it both earlier and here.

rkuris (changes requested):
I'm not sure I love the separation of concerns here, but I could be convinced it's okay.

Perhaps the Rootstore interface should be that it returns nodestores themselves, and then the RootStore can deal with its own caching (only in the real implementation).

I think that cleans up the manager itself to focus on doing what it's supposed to be doing (tracking nodestores) instead of constructing them.

firewood/src/manager.rs
RodrigoVillar: Done: a6cfff6

RodrigoVillar:
@rkuis I'm definitely on-board with having `RootStore` own constructing and maintaining its own nodestores. With our current code, refactoring would fail due to the following circular dependency:

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore

```

One way to fix this would be to pass in the latest revision to `get()`:

```rust
fn get(
        &self,
        hash: &TrieHash,
        latest_revision: CommittedRevision,
    ) -> Result<Option<LinearAddress>, Box<dyn std::error::Error + Send + Sync>>;
```

However, passing in the latest revision seems odd (we expect `RootStore` to handle it's own nodestores). I think the better alternative would be to delegate the `historical` and `by_hash` fields of the `RevisionManager` into their own DS and have that be shared by both the `RevisionManager` and `RootStore`:

```rust
pub struct InMemoryRevisions {
    historical: VecDeque<CommittedRevision>,
    by_hash: HashMap<TrieHash, CommittedRevision>
}
```

Refactoring our code to use this DS would lead to the following dependency graph:

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore
    s3: InMemoryRevisions

    s1 --> s3: Maintains latest revisions
    s1 --> s2: Wants historical revision
    s2 --> s3: Wants latest revision
```

ffi/firewood_test.go
RodrigoVillar: In the world of `dyn`'s, trait associated constants cannot be used unfortunately [[ref](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility)]. 

I went ahead and added `allow_space_reuse()` as a method here: d20cd81. This should still be much more performant than having to call `get()` every time we commit.

rkuris:
> ```rust
> pub struct InMemoryRevisions {
>     historical: VecDeque<CommittedRevision>,
>     by_hash: HashMap<TrieHash, CommittedRevision>
> }
> ```
Let's enter a new issue and do this. We should also have methods on this struct to manage and fetch from the list.

RodrigoVillar:
Created #1470 which will track this.

PR #1396, opened 2025-10-29
feat: join two key proofs into one trie

demosdemon:
This change adds support for merging key proof tries to create key range proof tries, along with error handling for merge failures and corresponding test coverage.

* Added `KeyRangeProofTrieRoot` to represent the merged result of zero, one, or two key proof tries, allowing construction of a trie that bounds a key range. This includes logic for merging tries and handling unbounded ranges.
* Implemented the `merge` and `merge_opt` methods for `KeyProofTrieRoot` and `KeyProofTrieNode`, enabling iterative merging of key proof tries and their nodes.
* Added `MergeKeyProofError` for detailed errors of merge failures, including disjoint edges, disjoint children, differing hashes, and duplicate keys with different values.

### DIFF
--- a/firewood/src/proofs/tests.rs
+++ b/firewood/src/proofs/tests.rs
@@ -4,8 +4,8 @@
 #![expect(clippy::unwrap_used, clippy::indexing_slicing)]
 
 use firewood_storage::{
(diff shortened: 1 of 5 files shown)


demosdemon:
stale

PR #1427, opened 2025-11-04
feat: merge key-value range into trie

demosdemon:
This change adds a new mechanism for merging key-value pairs with a merkle trie, refactors iterator and batch operation traits, and updates error handling to allow for batch iterators that may produce errors.

### Merkle Trie Merge Functionality

* Added the `merge_key_value_range` method to the `Merkle` struct, which merges a sorted sequence of key-value pairs with the base trie and yields a sequence of `BatchOp`s describing the changes. This is implemented via the new `MergeKeyValueIter` iterator, supporting efficient diffing and batch operation generation. 

### Iterator and Batch Trait Refactoring

* Refactored batch operation traits: replaced `KeyValuePairIter` and related traits with `TryIntoBatch`, `BatchIter`, and `IntoBatchIter`, providing a more flexible and type-safe interface for batch operations. 

### Error Handling and Iterator Return Types

* Changed iterator return types throughout the codebase to use `FileIoError` instead of the generic API `Error`, ensuring consistent error propagation for IO-related failures. Iterating over a trie can fail, but only because of IO-reasons (including data inconsistencies). This was also the driving change for changing the batch api to accept errors. Updated trait bounds and type aliases accordingly.
* Implemented `From<std::convert::Infallible>` for the API `Error` type to allow for seamless error conversion in infallible cases.


firewood/src/db.rs
```diff
@@ -315,11 +363,7 @@ impl api::DbView for Proposal<'_> {
 impl<'db> api::Proposal for Proposal<'db> {
     type Proposal = Proposal<'db>;
 
-    #[fastrace::trace(short_name = true)]
-    fn propose(
-        &self,
-        batch: impl IntoIterator<IntoIter: KeyValuePairIter>,
-    ) -> Result<Self::Proposal, api::Error> {
+    fn propose(&self, batch: impl IntoBatchIter) -> Result<Self::Proposal, api::Error> {
```
rkuris, line 366: What happened to the fastrace trace annotation? Does it not work with this new signature or something?

rkuris (approved):

firewood/src/db.rs
demosdemon: It was redundant. `propose_with_parent` (which `create_proposal` calls) had the same annotation so it was duplicated:

https://github.com/ava-labs/firewood/blob/7c313a360d55fe5a1f8f5511703107896b69d89f/firewood/src/db.rs#L159-L161

https://github.com/ava-labs/firewood/blob/7c313a360d55fe5a1f8f5511703107896b69d89f/firewood/src/db.rs#L224-L229

Issue #1422, opened 2025-11-04
ffi-nix test is flaky


rkuris:
See https://github.com/ava-labs/firewood/actions/runs/19077399006/job/54496653989 for an example. Earliest known failure is Oct 24.

rkuris:
@maru-ava reports it is not reproducible outside of CI. It may be x86 only.

rkuris:
The biggest clue is that the number of symbols is different:
```
+ echo 'Nix build:   20534 symbols'
Nix build:   20534 symbols
+ echo 'Cargo build: 20535 symbols'
Cargo build: 20535 symbols
+ '[' 20534 -eq 20535 ']'
```

@copilot can you change the failing script so that it outputs the symbol differences?

PR #1428, opened 2025-11-04
chore(ffi): Set -O2 for jemalloc nix build to ensure reproducibility

maru-ava:
See flake comment for the rational for this change.

Assuming this PR actually resolves the build indeterminism reported in #1422, the change in performance from `-O3` to `-O2` should probably be measured and weighed against the value of build determinism.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -246,6 +246,11 @@ jobs:
 
   ffi-nix:
     runs-on: ubuntu-latest
+    # Run 10 jobs in parallel to try to provoke the non-deterministic result
+    strategy:
+      fail-fast: false
+      matrix:
+        run_number: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30]
     steps:
       - uses: actions/checkout@v4
       - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20


rkuris:
Definitely need to add some performance numbers to this ticket. @RodrigoVillar recently updated the instructions on how to run these, they aren't too hard to run. The thing I care about is the amount of time it takes to load 10M blocks I think. If this causes more than a 1% perf drop, I think we should keep looking for other options, but I suspect it will not cause this much of a problem.

rkuris:
Looping in @Elvis339 perhaps he can run these or give you the other way of running them.

rkuris:
One thing to note for other reviewers: this only affects CFLAGS which is only used by jemalloc or any other C code we happen to link in.

.github/workflows/ci.yaml
```diff
@@ -245,14 +245,26 @@ jobs:
         run: GOEXPERIMENT=cgocheck2 TEST_FIREWOOD_HASH_MODE=firewood go test -race ./...
 
   ffi-nix:
-    if: false   # TODO(marun) Re-enable once non-determinism is rooted out
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: DeterminateSystems/nix-installer-action@786fff0690178f1234e4e1fe9b536e94f5433196 #v20
-      - uses: DeterminateSystems/magic-nix-cache-action@565684385bcd71bad329742eefe8d12f2e765b39 #v13
```
maru-ava, line 253: This change should be reverted before merge.

maru-ava:
Replaced by https://github.com/ava-labs/firewood/pull/1450

Issue #738, opened 2024-10-11
Finish range proof verification


rkuris:
Already started in https://github.com/ava-labs/firewood/pull/693 but needs completion

rkuris:
Rebased, warning cleanups next

demosdemon:
note to self: create separate issues for sub-prs. this auto-close is annoying and I wasn't able to figure out how to stop it for some PRs

rkuris:
There are still a bunch of tests marked disabled because of this ticket. Reopened so we can track removing them.

demosdemon:
> There are still a bunch of tests marked disabled because of this ticket. Reopened so we can track removing them.

yep, I mistook this for a different task. I realized it late last night. Thanks for re-opening!

PR #1429, opened 2025-11-04
feat(ffi): fill in ffi methods for range proofs

demosdemon:
This change fills in the Range Proof ffi methods for verify and committing a range proof. Verification is stubbed for now, but applying the range proof is supported.

`verify_and_propose` is part of `db.VerifyRangeProof` whereas a plain `verify` is used as part of `Verify` on the `RangeProof`.

Closes: #1387

### DIFF
--- a/ffi/proofs_test.go
+++ b/ffi/proofs_test.go
@@ -4,10 +4,8 @@
 package ffi
 
 import (
-	"context"
 	"runtime"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/require"
 )
@@ -52,7 +50,8 @@ func assertProofNotNil(t *testing.T, proof *RangeProof) {
 }
 
 // newVerifiedRangeProof generates a range proof for the given parameters and
-// verifies using [RangeProof.Verify] which does not prepare a proposal.
+// verifies using [RangeProof.Verify] which does not prepare a proposal. A
+// cleanup is registered to free the proof when the test ends.
 func newVerifiedRangeProof(
 	t *testing.T,
 	db *Database,
@@ -89,26 +88,6 @@ func newSerializedRangeProof(
 	r.NoError(err)
 
 	return proofBytes
-}
-
-func tryCloseDatabase(ctx context.Context, db *Database) error {
-	ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
-	defer cancel()
-	return db.Close(ctx)
-}
-
-// assertDatabaseCloseable verifies that the database can be closed within a short timeout.
-func assertDatabaseCloseable(t *testing.T, db *Database) {
-	t.Helper()
-	r := require.New(t)
-	r.NoError(tryCloseDatabase(t.Context(), db), "Database should be closeable with no active keep-alive handles")
-}
-
-// assertDatabaseNotCloseable verifies that the database cannot be closed due to active keep-alive handles.
-func assertDatabaseNotCloseable(t *testing.T, db *Database) {
-	t.Helper()
-	r := require.New(t)
-	r.ErrorIs(tryCloseDatabase(t.Context(), db), ErrActiveKeepAliveHandles, "Database should not be closeable with active keep-alive handles")
 }
 
 func TestRangeProofEmptyDB(t *testing.T) {
@@ -215,7 +194,7 @@ func TestRangeProofVerify(t *testing.T) {
 	proof := newVerifiedRangeProof(t, db, root, nothing(), nothing(), rangeProofLenTruncated)
 
 	// Database should be immediately closeable (no keep-alive)
-	assertDatabaseCloseable(t, db)
+	r.NoError(db.Close(oneSecCtx(t)))
 
 	// Verify with wrong root should fail
 	root[0] ^= 0xFF
@@ -304,13 +283,13 @@ func TestRangeProofFreeReleasesKeepAlive(t *testing.T) {
 	r.NoError(db.VerifyRangeProof(proof, nothing(), nothing(), root, rangeProofLenTruncated))
 
 	// Database should not be closeable while proof has keep-alive
-	assertDatabaseNotCloseable(t, db)
+	r.ErrorIs(db.Close(oneSecCtx(t)), ErrActiveKeepAliveHandles)
 
 	// Free the proof (releases keep-alive)
 	r.NoError(proof.Free())
 
 	// Database should now be closeable
-	assertDatabaseCloseable(t, db)
+	r.NoError(db.Close(oneSecCtx(t)))
 }
 
 func TestRangeProofCommitReleasesKeepAlive(t *testing.T) {
@@ -328,14 +307,14 @@ func TestRangeProofCommitReleasesKeepAlive(t *testing.T) {
 	r.NoError(db.VerifyRangeProof(proof, nothing(), nothing(), root, rangeProofLenTruncated))
 
 	// Database should not be closeable while proof has keep-alive
-	assertDatabaseNotCloseable(t, db)
+	r.ErrorIs(db.Close(oneSecCtx(t)), ErrActiveKeepAliveHandles)
 
 	// Commit the proof (releases keep-alive)
 	_, err = db.VerifyAndCommitRangeProof(proof, nothing(), nothing(), root, rangeProofLenTruncated)
 	r.NoError(err)
 
 	// Database should now be closeable
-	assertDatabaseCloseable(t, db)
+	r.NoError(db.Close(oneSecCtx(t)))
 
 	marshalledAfterCommit, err := proof.MarshalBinary()
 	r.NoError(err)
@@ -364,7 +343,7 @@ func TestRangeProofFinalizerCleanup(t *testing.T) {
 	r.NoError(db.VerifyRangeProof(proof, nothing(), nothing(), root, rangeProofLenTruncated))
 
 	// Database should not be closeable while proof has keep-alive
-	assertDatabaseNotCloseable(t, db)
+	r.ErrorIs(db.Close(oneSecCtx(t)), ErrActiveKeepAliveHandles)
 
 	runtime.KeepAlive(proof)
 	proof = nil //nolint:ineffassign // necessary to drop the reference for GC


ffi/src/proofs/range.rs
```diff
@@ -255,9 +418,30 @@ pub extern "C" fn fwd_db_verify_and_commit_range_proof(
 /// for the duration of the call.
 #[unsafe(no_mangle)]
 pub extern "C" fn fwd_range_proof_find_next_key(
-    _proof: Option<&mut RangeProofContext>,
+    proof: Option<&mut RangeProofContext>,
 ) -> NextKeyRangeResult {
-    CResult::from_err("not yet implemented")
+    // TODO(#352): proper implementation, this naively retuns the last key in
+    // in the range, which is correct, but not sufficient.
```
rkuris, line 416: nit: it's sufficient, just not ideal :)

ffi/proofs_test.go
```diff
@@ -37,6 +37,11 @@ func nothing() maybe {
 	}
 }
 
+func proofNotNil(r *require.Assertions, proof *RangeProof) {
```
alarso16: I think the more common way would be do to:
```suggestion
func proofNotNil(t *testing.T, proof *RangeProof) {
     t.Helper()
```

ffi/proofs_test.go
```diff
@@ -165,3 +170,201 @@ func rangeProofWithAndWithoutRoot(
...
+	proof, err := db.RangeProof(something(root), nothing(), nothing(), maxProofLen)
+	r.NoError(err)
+	proofNotNil(r, proof)
+	defer func() { r.NoError(proof.Free()) }()
+
+	// Test: Verify with correct root should succeed
+	err = proof.Verify(root, nothing(), nothing(), maxProofLen)
+	r.NoError(err)
+
+	// TODO(#738): re-enabled after verification is implemented
```
alarso16: This is probably the wrong issue

ffi/proofs_test.go
```diff
@@ -165,3 +170,201 @@ func rangeProofWithAndWithoutRoot(
...
+	if nextRange != nil {
+		r.NoError(nextRange.Free())
+	}
+
+	// Calling VerifyAndCommit should skip verification and commit
+	committedRoot, err := db.VerifyAndCommitRangeProof(
+		proof, nothing(), nothing(), root, maxProofLen,
+	)
+	r.NoError(err)
+	r.NotNil(committedRoot)
```
alarso16:
```suggestion
	r.NotEmpty(committedRoot)
```

ffi/proofs_test.go
alarso16: But also, I'm not sure why you need to poke into this implementation in the first place

alarso16 (commented):
I won't claim credit for this idea, but it might be worth finding some way of encoding bad proofs into bytes and trying to verify them. I'm not sure if it's worth the effort, but wanted to suggest it in case you know a good way of doing it

benchmark/bootstrap/aws-launch.sh
```diff
@@ -58,7 +58,7 @@ show_usage() {
     echo "  --firewood-branch BRANCH    Firewood git branch to checkout"
     echo "  --avalanchego-branch BRANCH AvalancheGo git branch to checkout"
     echo "  --coreth-branch BRANCH      Coreth git branch to checkout"
-    echo "  --libevm-branch BRANCH      LibEVM git branch to checkout"
+    echo "  --libevm-commit COMMIT      LibEVM git commit to checkout"
```
rkuris, line 62: Technically these are all commits not branches. Why are we renaming this one without renaming the others?

benchmark/bootstrap/aws-launch.sh
```diff
@@ -323,7 +322,19 @@ runcmd:
     git clone --depth 100 __CORETH_BRANCH_ARG__ https://github.com/ava-labs/coreth.git
   - >
     sudo -u ubuntu -D /mnt/nvme/ubuntu
-    git clone --depth 100 __LIBEVM_BRANCH_ARG__ https://github.com/ava-labs/libevm.git
+    git clone https://github.com/ava-labs/libevm.git
+  - >
+    sudo -u ubuntu -D /mnt/nvme/ubuntu/libevm
+    git checkout __LIBEVM_COMMIT__
```
rkuris, line 335: Why are we checking out main and then switching to this branch? Seems like we can do this in one just like the others.

demosdemon:
> Something is amiss here as there are a lot of changes to the bootstrapping scripts in this diff.

you were looking at the changes that had diverged from main before I merged them in.

demosdemon:
> I won't claim credit for this idea, but it might be worth finding some way of encoding bad proofs into bytes and trying to verify them.

I can add that test in when verification is implemented (#738). Because verification isn't implemented yet, that test would fail.

ffi/proofs_test.go
demosdemon: ~this was updated in #1437~ I added it to this PR instead.

demosdemon:
note: swapped the order of PRs so that this depends on #1437 instead of the other way around

ffi/proofs_test.go
```diff
@@ -54,6 +65,7 @@ func TestRangeProofNonExistentRoot(t *testing.T) {
 	keys, vals := kvForTest(100)
 	root, err := db.Update(keys, vals)
 	r.NoError(err)
+	r.NotEmpty(root)
```
alarso16:
```suggestion
	r.NotZero(root)
```

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	// Generate proof
+	proof, err := dbSource.RangeProof(sourceRoot, nothing(), nothing(), 0)
+	r.NoError(err)
+	proofNotNil(t, proof)
+	defer func() { r.NoError(proof.Free()) }()
+
+	// Verify and commit to target
+	committedRoot, err := dbTarget.VerifyAndCommitRangeProof(proof, nothing(), nothing(), sourceRoot, 0)
+	r.NoError(err)
+	r.NotEmpty(committedRoot)
```
alarso16:
```suggestion
	r.NotZero(committedRoot)
```

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	endKey := keys[30] // end before start
+
+	proof, err := db.RangeProof(
+		root,
+		something(startKey),
+		something(endKey),
+		maxProofLen,
+	)
+	// This might succeed with empty proof or fail - test actual behavior
+	if err == nil {
```
alarso16: Why wouldn't this fail?

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	endKey := keys[75]
+	proof, err := db.RangeProof(
+		root,
+		something(startKey),
+		something(endKey),
+		maxProofLen,
+	)
+	r.NoError(err)
+	proofNotNil(t, proof)
+	defer func() { r.NoError(proof.Free()) }()
```
alarso16: It has the finalizer, right?

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	r.ErrorIs(err, errNotPrepared)
+
+	// Transition to Proposed state
+	err = db.VerifyRangeProof(proof, nothing(), nothing(), root, maxProofLen)
+	r.NoError(err)
+
+	// State: Proposed - FindNextKey should work
+	nextRange, err := proof.FindNextKey()
+	r.NoError(err)
+	if nextRange != nil {
```
alarso16: Why would it be no-nil?

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+		r.NoError(nextRange.Free())
+	}
+}
+
+// createAndVerifyProof is a helper that creates a proof and verifies it with VerifyRangeProof,
+// which acquires a keep-alive handle on the database.
+func createAndVerifyProof(t *testing.T, db *Database) (*RangeProof, Hash) {
+	t.Helper()
```
alarso16: I'm not sure you want this
```suggestion
```
If the RangeProof creation fails, it won't point to a line number, so you would only know what part of the test fails based on the error message

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	proofNotNil(t, proof)
+
+	err = db.VerifyRangeProof(proof, nothing(), nothing(), root, 0)
+	r.NoError(err)
+
+	return proof, root
+}
+
+// assertDatabaseCloseable verifies that the database can be closed within a short timeout.
+func assertDatabaseCloseable(t *testing.T, db *Database) {
```
alarso16: I think this would be better done as:
```
func tryDatabaseClose(db) error {
```
And then you can check the error elsewhere, reducing the duplicate code below

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	ctx, cancel := context.WithTimeout(t.Context(), 100*time.Millisecond)
+	defer cancel()
+
+	err := db.Close(ctx)
+	r.ErrorIs(err, ErrActiveKeepAliveHandles, "Database should not be closeable with active keep-alive handles")
+}
+
+// TestRangeProofKeepsDbAlive verifies that VerifyRangeProof acquires a keep-alive handle
+// that prevents the database from being closed until the proof is freed.
+func TestRangeProofKeepsDbAlive(t *testing.T) {
```
alarso16: It seems like the next few tests could be made into a table

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+
+	// Unmarshal into new proof
+	proof2 := new(RangeProof)
+	err = proof2.UnmarshalBinary(proofBytes)
+	r.NoError(err)
+	defer func() { r.NoError(proof2.Free()) }()
+
+	// Database should be closeable (unmarshalled proof has no keep-alive)
+	// Note: We need to create a separate database without t.Cleanup to test this
+	dbFile := filepath.Join(t.TempDir(), "test2.db")
```
alarso16: What is the point of this? This is a separate database, this seems unrelated

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+
+	proofBytes, err := proof1.MarshalBinary()
+	r.NoError(err)
+	r.NoError(proof1.Free())
+
+	// Unmarshal into new proof
+	proof2 := new(RangeProof)
+	err = proof2.UnmarshalBinary(proofBytes)
+	r.NoError(err)
+	defer func() { r.NoError(proof2.Free()) }()
```
alarso16: This is freed before the database closes, so it doesn't seem like you're testing anything

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+// TestRangeProofFinalizerCleanup verifies that the finalizer properly releases
+// the keep-alive handle when the proof goes out of scope.
+func TestRangeProofFinalizerCleanup(t *testing.T) {
+	r := require.New(t)
+	dbFile := filepath.Join(t.TempDir(), "test.db")
+	db, err := newDatabase(dbFile)
+	r.NoError(err)
+
+	// Create and verify proof in a nested scope
+	func() {
```
alarso16: Is this the common way to do it? Why not just set the pointer to nil?

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+		proof, _ := createAndVerifyProof(t, db)
+		// Proof goes out of scope without explicit Free
+		_ = proof
+	}()
+
+	// Force GC to run finalizers
+	runtime.GC()
+	runtime.GC() // Run twice to be sure
```
alarso16: The second one definitely doesn't do anything, right?

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+	time.Sleep(100 * time.Millisecond)
+
+	// Database should eventually become closeable after finalizer runs
+	// Give it a bit more time since finalizers are not immediate
+	maxAttempts := 10
+	for i := range maxAttempts {
```
alarso16: Maybe just have a longer time limit (say, 1 second). You also made a helper for this

ffi/proofs_test.go
```diff
@@ -144,10 +156,556 @@ func rangeProof(
...
+
+	// Database should still be open due to active proof
+	// Free the proof to unblock the close
+	r.NoError(proof.Free())
+
+	// Close should complete now
+	err = <-closeDone
+	// Either succeeded or timed out is acceptable for this test
+	// The important part is that it doesn't panic or deadlock
+	if err != nil {
```
alarso16: If this error occurs, then the test shouldn't pass. That could mean that the proof isn't ever properly removed from the waitgroup

alarso16 (commented):
I left a lot of comments about testing

ffi/proofs_test.go
demosdemon: yeah, but the finalizer doesn't check if the result did not error

demosdemon: Actually no, the finalizer is only added after `Database.VerifyRangeProof`; we don't add finalizers for the other freeable types either (though, we probably should).

demosdemon:
> I left a lot of comments about testing

yeah, I let claude write most of the tests and didn't review them thoroughly. I've gone and written my own and cleaned up a lot of the weirdness.

ffi/proofs_test.go
```diff
@@ -37,11 +43,79 @@ func nothing() maybe {
...
+// assertProofNotNil verifies that the given proof and its inner handle are not nil.
+func assertProofNotNil(t *testing.T, proof *RangeProof) {
+	t.Helper()
+	r := require.New(t)
+	r.NotNil(proof)
+	r.NotNil(proof.handle)
+}
+
+// newVerifiedRangeProof generates a range proof for the given parameters and
+// verifies using [RangeProof.Verify] which does not prepare a proposal.
```
alarso16: Will you add that it will be freed after the test ends?

ffi/proofs_test.go
```diff
@@ -37,11 +43,79 @@ func nothing() maybe {
...
+
+	proof := newVerifiedRangeProof(t, db, root, startKey, endKey, proofLen)
+
+	proofBytes, err := proof.MarshalBinary()
+	r.NoError(err)
+
+	return proofBytes
+}
+
+func tryCloseDatabase(ctx context.Context, db *Database) error {
```
alarso16: I think we have a new function for these three. Instead of `assertDatabaseCloseable`, You could just do:

```
r.NoError(db.Close(oneSecCtx(t.Context()))
```
You can do something similar for the other one

PR #1430, opened 2025-11-04
chore: go workspace

alarso16:
I was frustrated all day trying to code from `firewood/` in VS Code because I couldn't see anything besides a bunch of red scriblles on my screen. I tracked it down to an issue with gopls, but it turns out this is a known problem: https://github.com/golang/go/issues/32394

`gopls`, the Go language server, can only recognize the modules you specifically point it to. This can be addressed by adding all three projects explicitly in VS code as workspaces, but this is the Go supported option for a multi-module workspace solution. Note that all modules need a unique path.

I'm not totally positive this is the perfect solution, since it didn't use to cause problems for me (circa 3 months ago). If anyone has a better solution, I'm happy to try it, but I assume I'm not the only one that's been bothered by this. The only other solution that I know if is "Add Folder To Workspace" for each go module, and then save this file somewhere and use this to open VS Code in the future


maru-ava:
Commentary on what appears to be a relevant [golang issue](https://github.com/golang/go/issues/53502) suggests that checking in go.work files may not always be desirable due to the lack of support for gracefully supporting committed and local go.work files.

[This comment](https://github.com/golang/go/issues/53502#issuecomment-1949199667) suggests a potential alternative - generating the go.work file.

alarso16:
Interesting references @maru-ava! It does seem like it might be a more flexible solution to encourage each developer to resolve this however they need. I'll close this and create a separate PR (#1434) adding advice on how to generate the go.work files and adding them to the .gitignore. If there's little interest beside mine in adding the workspace dependency, this is probably the easiest thing to do.

This is, of course, a reversible decision. If others feel that committing the go workspace is a better solution for everybody, this can be re-opened and adjusted.

PR #1435, opened 2025-11-05
feat(ffi)!: Add finalization logic for Revisions

alarso16:
When integrating code from the previous auto-drop method for Proposals in #1349, it's only reasonable to support this for Revisions as well. The logic here is exactly the same as for proposals, and they even share the same WaitGroup.

### DIFF
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -1178,6 +1178,10 @@ func TestHandlesFreeImplicitly(t *testing.T) {
 	db, err := newDatabase(filepath.Join(t.TempDir(), "test_GC_drops_implicitly.db"))
 	require.NoError(t, err)
 
+	// make the db non-empty
+	_, err = db.Update([][]byte{keyForTest(1)}, [][]byte{valForTest(1)})
+	require.NoError(t, err)
+
 	var (
 		explicitlyDropped []any
 		implicitlyDropped []any


demosdemon:
I agree with calling this "reasonable" and we should probably go ahead with it. But, this change isn't strictly necessary in the sense that it was for proposals. Holding onto a revision after closing the database will prevent firewood from reclaiming the space for that revision, if it were to be pruned, and will also prevent firewood from actually closing the file descriptor on the file (and its advisory lock). However, it will not introduce any undefined behavior that could lead to invalid memory accesses, like holding onto a proposal after closing the database will.

alarso16:
> But, this change isn't strictly necessary in the sense that it was for proposals. 

Yeah, no UB issues is great, but it does cause the file closure to hang. That was unexpected to me.

ffi/firewood_test.go
alarso16, line 1: The hardest part about this change is effective testing. @ARR4N's test was very clean for just proposals, but there's a lot of combined cases for revisions, where they may be treated different based on whether they are backed by proposals or not. Since most of the code is the same, I elected to combine into a single test. The other option is to create two more tests: one testing just revisions, and one testing that both revisions and proposals must be freed prior to close.

ffi/firewood_test.go
```diff
@@ -1206,25 +1236,26 @@ func TestProposalHandlesFreed(t *testing.T) {
...
 
-	runtime.KeepAlive(p0)
-	runtime.KeepAlive(p1)
-	p0 = nil
-	p1 = nil //nolint:ineffassign // Makes the value unreachable, allowing the finalizer to call Drop()
+	// This is the last time we must guarantee that these handles are reachable.
+	// After this, they must be unreachable so that the GC can collect them.
+	for _, h := range implicitlyDropped {
+		runtime.KeepAlive(h)
+	}
```
alarso16, line 1252: This doesn't explicitly remove all references. Go is supposedly smart enough to garbage collect here, but the defensive thing would be to set each reference to nil.

PR #1436, opened 2025-11-05
fix(benchmark/bootstrap): consistent go deps

RodrigoVillar:
Currently, there are two bugs when it comes to utilizing the bootstrap script:

- Currently, we force AvalancheGo to utilize the local versions of Coreth, Libevm, and Firewood. However, we don't apply this same process to Coreth. This is problematic because if we specify different versions of Libevm and Firewood in the bootstrap script than what's in the Coreth `go.mod`, then AvalancheGo and Coreth will diverge on what versions of Libevm and Firewood they use. 
- Utilizing the same version of libevm that's in the AvalancheGo/Coreth `go.mod` (`v1.13.15`) results in the error below. To my understanding, this is because checking out a tag breaks Go module resolutions even if you run `go mod tidy` inside of Libevm.
```
../libevm/common/bytes.go:24:2: missing go.sum entry for module providing package github.com/ethereum/go-ethereum/common/hexutil (imported by github.com/ava-labs/libevm/core); to add:
        go get github.com/ava-labs/libevm/core@v1.13.15-0.20251016142715-1bccf4f2ddb2
FAIL    github.com/ava-labs/avalanchego/tests/reexecute/c [setup failed]
FAIL
task: Failed to run task "reexecute-cchain-range": exit status 1
```

This PR fixes these issues as follows:
- Adds an additional step to force Coreth to use the local versions of Libevm and Firewood
- Replaces passing in a branch with passing in a commit for Libevm


benchmark/bootstrap/aws-launch.sh
```diff
@@ -456,6 +469,7 @@ else
 
     INSTANCE_ID=$(eval "$AWS_CMD")
     echo "instance id $INSTANCE_ID started"
+    echo -e "\033[33mREMINDER: Please manually shut down your EC2 instance after running the benchmark to avoid unexpected costs!\033[0m"
```
RodrigoVillar, line 472: I can remove this diff, but thought it was useful to log that the client needs to manually shut down their instance (in yellow text).

benchmark/bootstrap/aws-launch.sh
```diff
@@ -58,7 +58,7 @@ show_usage() {
     echo "  --firewood-branch BRANCH    Firewood git branch to checkout"
     echo "  --avalanchego-branch BRANCH AvalancheGo git branch to checkout"
     echo "  --coreth-branch BRANCH      Coreth git branch to checkout"
-    echo "  --libevm-branch BRANCH      LibEVM git branch to checkout"
+    echo "  --libevm-commit COMMIT      LibEVM git commit to checkout"
```
RodrigoVillar, line 61: Open to switching branches for commits for all the other dependencies to be consistent, but changing just Libevm got the reexecution tests to build for me.

PR #1437, opened 2025-11-05
feat(ffi): add keepalive struct to own waitgroup logic

demosdemon:
Expanding on Arran and Austin's work, I added the `databaseKeepAliveHandle` to manage the WaitGroup logic for keeping the database alive while various objects needed.

In the future, the `RangeProof` object will also use this keep-alive handle, but only during certain stages of its lifecycle. Because of this, the disown method on the keep-alive handle is capable of being called on the nil receiver without causing a panic.


ffi/firewood.go
alarso16, line 1: nit: not sure this is the file for this. Clutters an otherwise relatively exported file with unexported details shared between all structs (except the one defined here)

ffi/firewood.go
```diff
@@ -306,3 +316,30 @@ func (db *Database) Close(ctx context.Context) error {
 
 	return nil
 }
+
+func (h *databaseKeepAliveHandle) keepAlive(wg *sync.WaitGroup) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	if h.outstandingHandles != nil {
```
alarso16: The purpose of this if block is unclear to me. Shouldn't we only call this once? If we call it multiple times, then isn't that a bug?

ffi/proofs.go
```diff
@@ -109,7 +117,13 @@ func (db *Database) VerifyRangeProof(
 		max_length: C.uint32_t(maxLength),
 	}
 
-	return getErrorFromVoidResult(C.fwd_db_verify_range_proof(db.handle, args))
+	err := getErrorFromVoidResult(C.fwd_db_verify_range_proof(db.handle, args))
+	if err == nil {
+		proof.keepAlive(&db.outstandingHandles)
```
alarso16: I think the slightly more go-ey thing to do would be to early return on err:
```
	if err := getErrorFromVoidResult(C.fwd_db_verify_range_proof(db.handle, args)); err != nil {
		return err
	}

	proof.keepAlive(&db.outstandingHandles)
	runtime.SetFinalizer(proof, (*RangeProof).Free)
	return nil
```
Doesn't matter much, but thought I'd mention it

ffi/firewood.go
alarso16: This seems like a recipe for strange behavior, since we only ever create one waitgroup. I think it's invalid after the first call to `Done()`

ffi/firewood.go
```diff
@@ -306,3 +316,30 @@ func (db *Database) Close(ctx context.Context) error {
...
+		h.outstandingHandles.Done()
+		h.outstandingHandles = nil
+	}
+
+	h.outstandingHandles = wg
+	h.outstandingHandles.Add(1)
+}
+
+func (h *databaseKeepAliveHandle) disown(disownEvenOnErr bool, withLock func() error) error {
```
alarso16: `withLock` is a very strangely named field. I don't really have any better names. Maybe `attemptDisown`?

ffi/proofs.go
alarso16, line 1: These changes aren't tested, is there enough work on range proofs to accurately test this?

ffi/firewood.go
alarso16: This error message isn't quite correct now, but I probably should have resolved it myself. Will you fix it?

ffi/firewood.go
demosdemon: This is because calling `Add(1)` without also calling `Add(-1)` is weird. The alternative is to panic if `outstandingHandles` is already not nil and it would be incorrect to assume `outstandingHandles` will always be nil when this is called.

With proofs, I'm forced to use the nilness of `outstandingHandles` to check if I need to call `Drop` because its the only indicator to the Proof because there is no exposed handle to the internal proposal.

ffi/firewood.go
demosdemon: yeah, the name is weird and the `disownEvenOnErr` makes the `attempt` part of `attemptDisown` weird. The intention behind `withLock` when I first wrote it was to indicate that this is part of the disown behavior that occurs with the mutex acquired as opposed to other disown work that doesn't rely on the block.

ffi/firewood.go
```diff
@@ -295,7 +296,7 @@ func (db *Database) Close(ctx context.Context) error {
 	select {
 	case <-done:
 	case <-ctx.Done():
-		return fmt.Errorf("at least one reachable %T neither dropped nor committed", &Proposal{})
+		return fmt.Errorf("at least one reachable %T not disowned", &databaseKeepAliveHandle{})
```
alarso16: What if we create a global, exported error, and we're able to test that this actually does trigger? Also, for the user, referencing an internal (and rather confusing) type in the error message is probably not helpful

ffi/keepalive.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// init initializes the keep-alive handle to track a new outstanding handle.
+func (h *databaseKeepAliveHandle) init(wg *sync.WaitGroup) {
+	// lock not necessary today, but will be necessary in the future for types
+	// that initialize the handle at some point after construction (#1429).
+	h.mu.Lock()
+
+	if h.outstandingHandles != nil {
+		h.mu.Unlock()
```
alarso16: Looks like you could move this to a defer statement: https://go.dev/play/p/ImVrsuhlwfb

demosdemon: yeah, decided to skip the `defer` because it does add a non-trivial amount of overhead and I felt that was unnecessary for the _very_ common case where this will be nil. Handling the non-null case is only necessary for RangeProofs and I didn't want to penalize the more common proposals since they won't ever panic. (anymore than I already am by adding the lock before setting the handle)

demosdemon: but, looking at the internals of `Add`, it can panic if called with a positive delta while there are ongoing `Wait` calls, so it's better to use defer

alarso16 (approved):
Your tests are super thorough, love to see it

PR #1442, opened 2025-11-07
chore(release): prepare for v0.0.14

demosdemon:
- Updated `.golangci.yaml`
- Updated cargo dependencies
- Resolved deprecated warning in `fwdctl`
- Resolved new 1.91 lint warnings
- Updated release documentation


CHANGELOG.md
```diff
@@ -2,6 +2,79 @@
...
+- Use PathComponent in proofs ([#1359](https://github.com/ava-labs/firewood/pull/1359))
+- Replace NodeAndPrefix with HashableShunt ([#1362](https://github.com/ava-labs/firewood/pull/1362))
+- Parallel updates to Merkle trie (v2) ([#1258](https://github.com/ava-labs/firewood/pull/1258))
+- Parallel hashing of Merkle trie ([#1303](https://github.com/ava-labs/firewood/pull/1303))
+- Add nix flake for ffi ([#1319](https://github.com/ava-labs/firewood/pull/1319))
+- Add more info for IO error message ([#1378](https://github.com/ava-labs/firewood/pull/1378))
+- Add TrieNode trait and related functionality ([#1363](https://github.com/ava-labs/firewood/pull/1363))
+- Add hashed key-value trie implementation ([#1365](https://github.com/ava-labs/firewood/pull/1365))
+- Define explicit associated types on Hashable ([#1366](https://github.com/ava-labs/firewood/pull/1366))
+- *(ffi)* `Database.Close()` guarantees proposals committed or freed ([#1349](https://github.com/ava-labs/firewood/pull/1349))
```
alarso16, line 29: This is breaking

CHANGELOG.md
```diff
@@ -2,6 +2,79 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.0.14] - 2025-11-07
+
+### 🚀 Features
+
+- *(monitoring)* Grafana automatic dashboard and auth provisioning ([#1307](https://github.com/ava-labs/firewood/pull/1307))
+- *(ffi)* Implement revision handles and expose to FFI layer ([#1326](https://github.com/ava-labs/firewood/pull/1326))
```
alarso16, line 10: I think this is breaking, since you now have to explicitly drop revisions

RELEASE.md
```diff
@@ -73,15 +122,30 @@ firewood-storage.workspace = true
...
+git cliff --tag v0.0.15 -o CHANGELOG.md
 ```
 
+## Commit
+
+Commit the version bump and change log updates. Using the summary prefix:
+
+> chore(release): prepare for
+
+will cause `git-cliif` to omit the commit from the changelog. This is why
```
alarso16: nit:
```suggestion
will cause `git-cliff` to omit the commit from the changelog. This is why
```

CHANGELOG.md
demosdemon: indicating that would require the original commit to include the `!` and I don't have a way to manually override that without having to do it every time the CHANGELOG is generated (#1443); however, all version bumps are technically a breaking change until we go to 0.1.0 because applying the bump downstream requires manual updates.

Issue #1383, opened 2025-10-23
perf: parallel insert/hash performance and commit time regression


AminR443:
### Current State
#1258 introduced parallel insert and hash operations that improved propose times. However, commit times have regressed compared to the original implementation.

### Goals
1. Reliable benchmarks to quantify the overall performance impact of parallel operations
2. Identify bottlenecks causing longer commit times
3. Determine optimal configuration and improvements to maximize performance gains

### Tasks
- [x] Create comprehensive benchmark suite comparing:
  - Original serial implementation
  - Current parallel implementation
  - Various parallelism configurations
- [x] Profile commit phase to identify performance bottlenecks
- [x] Document performance characteristics under different workloads
- [x] Investigate potential optimizations for commit phase
- [ ] Provide configuration guidelines for best performance (#1368, #1358, #1357)

AminR443:
I finally discovered the reason why parallel propose+hashing changes by @bernard-avalabs were slower in commit times although with no changes to the commit. We were suspicious of disk and I/O, but I did several profilings and block traces that showed no signficant difference on I/O.
After more careful profiling, I found out the reason is behind this data structure:
```rust
pub struct MaybePersistedNode(Arc<ArcSwap<MaybePersisted>>);
```
After writing the nodes to the disk we call `persist_at` for each node, which subsequently calls `ArcSwap::store`, to swap the existing data with the new `MaybePersisted::Persisted`. In parallel propose version, although that we returned the threads back to threadpool, this operation took significantly longer than the main version.

To mitigate, I replaced it with `std::RwLock` and it provided a good performance boost. I explored ArcSwap's implementation to find the reason of performance regression. As ArcSwap is lock-free, it tracks the references with per-thread debt slot, and when trying to `store` a new data, it makes sure to pay all the debts (bump the reference counts on the old reference to safely drop it). Because all of this is tracked with per-thread debt slots (a global linked list), as we increased the number of threads (which helped propose time), more threads read from this node and acquired slots, and finally when doing the `store` the overhead grow and hurt commit time. Returning the threads back to threadpool didn't matter because this work is done regardless of readers state.

Issue #966, opened 2025-06-16
feat: add "fwdctl import" which reads all the unambiguous "fwdctl dump" formats


qusuyan:
As mentioned in #936, a command for bulk loading data into firewood can help recreate a database image efficiently. This can be helpful in testing.

fatihmcicek:
I'd like to pick this up. Quick scope check before I start:

`fwdctl dump` currently emits four formats (`csv`, `json`, `stdout`, `dot`), and `fwdctl import` only reads `csv`. I read `csv` and `json` as the unambiguous ones — `stdout` is for human consumption and `dot` is a graph visualization rather than a data dump — so the remaining work would be adding a `Json` variant to `InputFormat`, mirroring the `JsonOutputHandler` on the dump side.

Plan: JSON reader in `fwdctl/src/import.rs` + CLI tests following the existing `test_slow_fwdctl_import_csv` pattern, including a dump→import→root-hash roundtrip to verify fidelity, plus a malformed-input case.

Does that match the intent here? Happy to adjust scope if you had other formats in mind.

PR #1447, opened 2025-11-12
feat(storage): replace `ArcSwap` with `Mutex` for better performance

AminR443:
Per benchmarks done for discovering source of commit time regressions in parallel propose (#1383), we identified that the reason was `ArcSwap` usage in `MaybePersistedNode`. The lock-free mechanism of `ArcSwap` introduces write overhead when multiple threads read the data, growing linearly by the number of threads. More details are provided in #1383.

This PR replaces the `ArcSwap<MaybePersisted>` usage with `Mutex<Arc<MaybePersisted>>`. Re-execution tests for blocks 50-60m show 20% improvement (MGas/s) with parallel propose using `Mutex`, while the `ArcSwap` version provided no benefit over normal propose and slightly degraded performance.

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -21,7 +21,7 @@ use rayon::{ThreadPool, ThreadPoolBuilder};
 use typed_builder::TypedBuilder;
 
 use crate::merkle::Merkle;
-use crate::root_store::{RootStore, RootStoreError};
+use crate::root_store::RootStore;
 use crate::v2::api::{ArcDynDbView, HashKey, OptionalHashKeyExt};
 
 pub use firewood_storage::CacheReadStrategy;
@@ -97,8 +97,8 @@ pub(crate) enum RevisionManagerError {
     },
     #[error("An IO error occurred during the commit: {0}")]
     FileIoError(#[from] FileIoError),
-    #[error("A RootStore error occurred")]
-    RootStoreError(#[from] RootStoreError),
+    #[error("A RootStore error occurred: {0}")]
+    RootStoreError(#[source] Box<dyn std::error::Error + Send + Sync>),
 }
 
 impl RevisionManager {
@@ -152,7 +152,10 @@ impl RevisionManager {
                 },
             )?;
 
-            manager.root_store.add_root(&root_hash, &root_address)?;
+            manager
+                .root_store
+                .add_root(&root_hash, &root_address)
+                .map_err(RevisionManagerError::RootStoreError)?;
         }
 
         Ok(manager)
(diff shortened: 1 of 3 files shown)


AminR443:
Closes #1383

storage/src/node/persist.rs
```diff
@@ -132,35 +128,38 @@ impl MaybePersistedNode {
...
+                MaybePersisted::Unpersisted(node) | MaybePersisted::Allocated(_, node) => {
+                    node.clone()
+                }
+                MaybePersisted::Persisted(_) => {
+                    unreachable!("Cannot allocate a node that is already persisted on disk");
+                }
             }
-        }
+        };
+        *self.0.lock().expect("poisoned lock") = MaybePersisted::Allocated(addr, node);
```
demosdemon: don't need to lock, unlock, lock. that allows for TOCTOU bugs. should be able acquire the lock only once

```suggestion
        let guard = self.0.lock().expect("poisoned lock");
        let node = {
            match &*guard {
                MaybePersisted::Unpersisted(node) | MaybePersisted::Allocated(_, node) => {
                    node.clone()
                }
                MaybePersisted::Persisted(_) => {
                    unreachable!("Cannot allocate a node that is already persisted on disk");
                }
}
        };
        *guard = MaybePersisted::Allocated(addr, node);
```

storage/src/node/persist.rs
```diff
@@ -189,9 +188,11 @@ impl MaybePersistedNode {
 /// If instead you want the node itself, use [`MaybePersistedNode::as_shared_node`] first.
 impl Display for MaybePersistedNode {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self.0.load().as_ref() {
+        match &*self.0.lock().expect("poisoned lock") {
```
demosdemon: could potentially use [`try_lock`](https://doc.rust-lang.org/stable/std/sync/struct.Mutex.html#method.try_lock) for a non-blocking display impl (this is what `Debug` does by default). it's not important though.

storage/src/node/persist.rs
AminR443: Thanks for catching this. I didn't notice this as I replaced most of the stuff in-place.

storage/src/node/persist.rs
AminR443: I switched to `try_lock`, not sure if we'll need to change in future. If we do #966, we'll use this dumped data to import to database, and we may need to have the data anyway, even if it's blocking.

PR #1448, opened 2025-11-12
feat(ffi)!: Remove unused kvBackend interface

alarso16:
This is straight techdebt. This interface isn't even exported, and the two functions are no-ops


demosdemon (approved):
I thought this interface was so we maintained some api compatible with coreth/libevm. But if we don't need it, sure.

alarso16:
> I thought this interface was so we maintained some api compatible with coreth/libevm. But if we don't need it, sure.

We had it for Aaron's benchmarking, but I think that is very stale now

alarso16:
I checked with @aaronbuchwald offline, so I will queue this to merge

PR #1449, opened 2025-11-12
perf(ffi)!: Use fixed size Hash

alarso16:
There's no sense casting between byte slices and fixed sizes all the time, and it only encourages bad usage. I expect a teeny tiny performance win out of this, but more this is a point of style.

The root is still passed to Rust as a slice, but everything else is now stored in Go as a fixed length array. One potentially unnecessary side effect is a change to the interface for Range Proofs, since I don't know why you would possibly want a "most recent range proof" anyway.

Potential future work may include optimizing the FFI layer to actually use these fixed length arrays, rather than typecasting to slices.

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -165,18 +165,14 @@ impl RevisionManager {
     /// To commit a proposal involves a few steps:
     /// 1. Commit check.
     ///    The proposal's parent must be the last committed revision, otherwise the commit fails.
-    /// 2. Persist delete list.
-    ///    The list of all nodes that were to be deleted for this proposal must be fully flushed to disk.
-    ///    The address of the root node and the root hash is also persisted.
-    ///    Note that this is *not* a write ahead log.
     ///    It only contains the address of the nodes that are deleted, which should be very small.
-    /// 3. Revision reaping. If more than the maximum number of revisions are kept in memory, the
+    /// 2. Revision reaping. If more than the maximum number of revisions are kept in memory, the
     ///    oldest revision is reaped.
-    /// 4. Persist to disk. This includes flushing everything to disk.
-    /// 5. Persist the revision to `RootStore`.
-    /// 6. Set last committed revision.
+    /// 3. Persist to disk. This includes flushing everything to disk.
+    /// 4. Persist the revision to `RootStore`.
+    /// 5. Set last committed revision.
     ///    Set last committed revision in memory.
-    /// 7. Proposal Cleanup.
+    /// 6. Proposal Cleanup.
     ///    Any other proposals that have this proposal as a parent should be reparented to the committed version.
     #[fastrace::trace(short_name = true)]
     #[crate::metrics("firewood.proposal.commit", "proposal commit to storage")]
@@ -192,9 +188,8 @@ impl RevisionManager {
 
         let mut committed = proposal.as_committed(&current_revision);
 
-        // 2. Persist delete list for this committed revision to disk for recovery
-
-        // 3 Take the deleted entries from the oldest revision and mark them as free for this revision
+        // 2. Revision reaping
+        // Take the deleted entries from the oldest revision and mark them as free for this revision
         // If you crash after freeing some of these, then the free list will point to nodes that are not actually free.
         // TODO: Handle the case where we get something off the free list that is not free
         while self.historical.read().expect("poisoned lock").len() >= self.max_revisions {
@@ -232,19 +227,19 @@ impl RevisionManager {
             gauge!("firewood.max_revisions").set(self.max_revisions as f64);
         }
 
-        // 4. Persist to disk.
+        // 3. Persist to disk.
         // TODO: We can probably do this in another thread, but it requires that
         // we move the header out of NodeStore, which is in a future PR.
         committed.persist()?;
 
-        // 5. Persist revision to root store
+        // 4. Persist revision to root store
         if let (Some(hash), Some(address)) = (committed.root_hash(), committed.root_address()) {
             self.root_store
                 .add_root(&hash, &address)
                 .map_err(RevisionManagerError::RootStoreError)?;
         }
 
-        // 6. Set last committed revision
+        // 5. Set last committed revision
         let committed: CommittedRevision = committed.into();
         self.historical
             .write()
@@ -257,7 +252,7 @@ impl RevisionManager {
                 .insert(hash, committed.clone());
         }
 
-        // 7. Proposal Cleanup
+        // 6. Proposal Cleanup
         // Free proposal that is being committed as well as any proposals no longer
         // referenced by anyone else.
         self.proposals


ffi/src/proofs/range.rs
```diff
@@ -110,12 +110,10 @@ pub extern "C" fn fwd_db_range_proof(
     args: CreateRangeProofArgs,
 ) -> RangeProofResult {
     crate::invoke_with_handle(db, |db| {
-        let root_hash = match args.root {
-            Maybe::Some(root) => root.as_ref().try_into()?,
-            Maybe::None => db
-                .current_root_hash()?
-                .ok_or(api::Error::RangeProofOnEmptyTrie)?,
-        };
+        let root_hash: firewood_storage::TrieHash = args.root.as_ref().try_into()?;
```
alarso16: I apologize for the un-Rusty code. Any ideas for improvement?

ffi/src/proofs/range.rs
```diff
@@ -110,12 +110,10 @@ pub extern "C" fn fwd_db_range_proof(
...
+        if firewood_storage::TrieHash::empty() == root_hash {
+            return Err(api::Error::RangeProofOnEmptyTrie);
+        }
```
demosdemon: This isn't necessary and is incorrect; comparing `root_hash` with `empty()` is only correct for merkledb hashing, not ethhash. `TrieHash::default_root_hash()` is the correct empty trie hash that considers the build flags.

I originally had it here because it was correct to return early if I was unable to obtain the root hash. If the user is providing the hash, I would always pass it through to the function. Removing this will cause `RevisionNotFound` to be returned instead if the revision was truly not found, which is a more correct response when the user provides a root hash and if the trie is empty, the empty error is provided later down <https://github.com/ava-labs/firewood/blob/607b735c45853ce53d96a15aa84843a1bde92b48/firewood/src/merkle/mod.rs#L414-L418>

> since I don't know why you would possibly want a "most recent range proof" anyway

The reason for having the ability to get a range proof without providing the root was to skip the two-step action of querying the root and then querying for the range proof with that root. It was provided as an optimization since `eth_getProof` allows you to specify `"latest"` for `blockNumber`, if felt like range proofs may want the same ability. This was also before I knew anything about the peer protocol.

ffi/src/proofs/range.rs
demosdemon: this is fine for now because I'm probably going to remove it; unless you want to go through and replace all of the instances where we cast a slice into the hashkey with instead using the [`HashKey`](https://github.com/ava-labs/firewood/blob/607b735c45853ce53d96a15aa84843a1bde92b48/ffi/src/value/hash_key.rs#L6-L21) ffi type and using [`.into()`](https://github.com/ava-labs/firewood/blob/607b735c45853ce53d96a15aa84843a1bde92b48/ffi/src/value/hash_key.rs#L34-L39) to convert it into the [`HashKey`](https://github.com/ava-labs/firewood/blob/607b735c45853ce53d96a15aa84843a1bde92b48/firewood/src/v2/api.rs#L33-L40) firewood type (they are slightly different for FFI reasons).

combined with my other note, that would turn this into:

```
        let view = db.get_root(args.root.into())?;
```

To create the ffi HashKey in Go:

```go
func newCHashKey(hash Hash) C.HashKey {
	return *(*C.HashKey)(unsafe.Pointer(&hash))
}
```

ffi/src/proofs/range.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
 pub struct CreateRangeProofArgs<'a> {
     /// The root hash of the revision to prove. If `None`, the latest revision
     /// is used.
-    pub root: Maybe<BorrowedBytes<'a>>,
+    pub root: BorrowedBytes<'a>,
```
demosdemon: re your try_into; it can be removed if this is a `HashKey` (`use crate::HashKey;`).

ffi/proofs.go
demosdemon: missed one

ffi/src/proofs/range.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
 pub struct CreateRangeProofArgs<'a> {
     /// The root hash of the revision to prove. If `None`, the latest revision
```
RodrigoVillar: I think we can remove the `None` reference.

ffi/firewood.go
```diff
@@ -194,13 +191,13 @@ func (db *Database) Get(key []byte) ([]byte, error) {
...
 // If key is not found, it returns (nil, nil).
-func (db *Database) GetFromRoot(root, key []byte) ([]byte, error) {
+func (db *Database) GetFromRoot(root Hash, key []byte) ([]byte, error) {
 	if db.handle == nil {
 		return nil, errDBClosed
 	}
 
 	// If the root is empty, the database is empty.
-	if len(root) == 0 || bytes.Equal(root, EmptyRoot) {
+	if root == (EmptyRoot) {
```
RodrigoVillar: I think we could remove the parentheses here.

ffi/firewood.go
```diff
@@ -209,50 +206,40 @@ func (db *Database) GetFromRoot(root, key []byte) ([]byte, error) {
...
+	return getHashKeyFromHashResult(C.fwd_root_hash(db.handle))
 }
 
 func (db *Database) LatestRevision() (*Revision, error) {
 	root, err := db.Root()
 	if err != nil {
 		return nil, err
 	}
-	if bytes.Equal(root, EmptyRoot) {
+	if root == (EmptyRoot) {
```
RodrigoVillar: Same comment regarding parentheses.

ffi/proofs_test.go
```diff
@@ -57,11 +57,10 @@ func TestRangeProofNonExistentRoot(t *testing.T) {
 	r.NotNil(root)
 
 	// create a bogus root
-	bogusRoot := make([]byte, len(root))
-	copy(bogusRoot, root)
+	bogusRoot := root
 	bogusRoot[0] ^= 0xFF
```
RodrigoVillar:
```suggestion
    // corrupt the root
    root[0] ^= 0xFF
```

nit: could we remove creating `bogusRoot`? See above for an example

ffi/src/proofs/range.rs
alarso16: We won't support `eth_getProof`, but also even in this case where they request latest, the root is abstracted

ffi/firewood_test.go
```diff
@@ -1088,13 +1073,8 @@ func TestGetFromRoot(t *testing.T) {
...
 	}
 
-	// Test with invalid root hash
-	invalidRoot := []byte("this is not a valid 32-byte hash")
-	_, err = db.GetFromRoot(invalidRoot, []byte("key"))
-	r.Error(err, "GetFromRoot with invalid root should return error")
-
 	// Test with valid-length but non-existent root
-	nonExistentRoot := make([]byte, RootLength)
+	nonExistentRoot := EmptyRoot
```
RodrigoVillar:
```suggestion
	var nonExistentRoot Hash
```

demosdemon (approved):

PR #1464, opened 2025-11-17
feat: remove the last bits of arc-swap

demosdemon:
I noticed there was one remaining component using arc-swap. This change replaces it with a standard Mutex. And, removes the dependency on arc-swap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,12 +140,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arc-swap"
-version = "1.7.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


AminR443:
Sounds great. I planned on doing this sometime soon, was not sure about performance implications though; Are we sure that mutex is the best option here too?

demosdemon:
> Are we sure that mutex is the best option here too?

The parent here has nearly the same usages as `MaybePersisted`; it is only accessed during commit -- to verify the parent is committed and has the expected hash (`parent_hash_is`) or to indicate the parent is committed and set its hash (`commit_reparent`). There should be no contention on the lock right now since there's no parallelism on commits.

Issue #1054, opened 2025-07-11
As a result of unpersisted nodes in proposals, giant nodes are allowed


rkuris:
The giant_node test in nodestore no longer panics because we don't persist nodes when the proposal becomes immutable.

We should, however, prevent someone from creating a node so large that it won't fit into our maximum allocatable node size (currently ~16M). This is not trivial because we actually don't know how large the node will be when we serialize it, and we want to avoid serializing it until it is getting persisted.

The right fix is probably to prevent inserting (key.len() + value.len()) > AREA_SIZES.last() - MAX_OVERHEAD but I don't know how to calculate MAX_OVERHEAD exactly. An estimate might work, but that seems quite fragile.

demosdemon:
Issue Grooming: Ron's recent arena allocator changes may have fixed this issue; and if not, may have made it much easier to fix. I will follow up.

PR #1217, opened 2025-08-18
chore: only allocate the area needed

qusuyan:
As shown from statistics, Firewood currently has a lot of internal fragmentations because it will allocate any large enough free area from the free list, often times in which the allocated area is several times bigger than the area needed. We have found 3 possible mitigations:
1. When allocating, only allocate the area needed to store the node but not larger areas (this PR)
2. Since allocation happens in batches, we can allocate areas in the order of area sizes - so that large areas have priority and it may avoid allocating large areas from file end in a lot of cases. 
3. If only large areas are available, split the area and only allocate one. Because Merkle Trie contains dominantly small trie nodes, this design should lead to low external fragmentation. 

We need to run more experiments to see if these approaches actually work well.

PR #1465, opened 2025-11-17
test: fix `giant_node` test

demosdemon:
The previous work to defer allocating nodes until commit time had the side effect of allowing nodes to exceed the maximum allowed size at hash time. This defers the error until commit time, which is not ideal.

This change updates the test to confirm that the allocator still errors when trying to allocate a node that is too large. However, we would prefer to do this earlier but without serializing the node multiple times.

As part of this change, I created a helper type that the test can use to verify the exact error was returned instead of comparing against the string representation of the error. As a part of that, I discovered that there was some vestiges of old code where the allocator would potentially return an area larger than necessary when allocating from the free list. However, that functionality was removed in #1217; thus, I have simplified the remaining code to reflect that which also simplified my `giant_node` test changes. Similarly, the removed wasted metric was no longer relevant because we no longer select free nodes larger than needed.

I added `DisplayTruncatedHex` to the `Debug` representations of the leaf and branch nodes to truncate the node values when printing debug info. This was necessary before fixing the `giant_node` test because the debug output included the full contents of the node, which would flood the terminal when the test failed and cause it to hang.

Closes: #1054

### DIFF
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -239,6 +239,13 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
                 *free_stored_area_addr = free_head.next_free_block;
             }
 
+            firewood_counter!(
+                "firewood.space.reused",
+                "Bytes reused from free list by index",
+                "index" => index_name(index)
+            )
+            .increment(index.size());
+
             // Return the address of the newly allocated block.
             trace!("Allocating from free list: addr: {address:?}, size: {index}");
             return Ok(Some(address));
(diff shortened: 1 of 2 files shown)


storage/src/nodestore/alloc.rs
```diff
@@ -247,22 +239,9 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
                 *free_stored_area_addr = free_head.next_free_block;
             }
 
-            firewood_counter!(
```
rkuris, line 250: Why were these removed? They were useful before and might still be useful later

rkuris (approved):

storage/src/nodestore/alloc.rs
demosdemon: I removed them because after #1217 we no longer selected free-nodes that are larger than the minimum needed. Therefore, `wasted` is no longer a relevant metric. But I'll add `reused` back. I was focused on the `wasted` part at the time and after looking back, `reused` is still useful.

Issue #1413, opened 2025-10-30
docs(ffi): add godoc comments to `Config`


RodrigoVillar:
The `Config` type in the `ffi` package is as follows:

https://github.com/ava-labs/firewood/blob/9ed71b121dae9de79465acda4254e935d68d193a/ffi/firewood.go#L55-L62

However, it feels a bit weird that we don't have documentation for any of the fields of `Config`- anyone wondering what these fields mean would need to look through the Rust code.

Issue #1113, opened 2025-07-22
docs: Go doc


alarso16:
Although #1078 addresses Rust docs and the `README.md`, one would still expect the [Go docs](https://pkg.go.dev/github.com/ava-labs/firewood-go-ethhash/ffi#section-documentation) to be populated appropriately. This likely requires a little investigation into licensing and how they are rendered.

PR #1466, opened 2025-11-17
docs(ffi): Update godoc

### LINKED ISSUE #1348: docs(ffi): Clarify terminology and behavior for Revisions in docs
## Problem

The documentation uses inconsistent terminology across the FFI layer. In Rust, a revision can be either a `CommittedRevision` (also called a historical revision) or a `ProposedRevision` (a revision created by a proposal but not yet committed). However, in Go, the term "revision" almost always refers to historical revisions only, leading to confusion across language bindings.

## Context

In #1326, we introduced revisions to the FFI layer as a way to get a handle to a specific revision and prevent it from being removed. The current approach returns a "view" that can be either proposed or committed. According to @alarso16, coreth/libevm uses `GetFromRoot(root, key)` to access proposals during commit and never reads directly from a proposal.

## Action Items

- [ ] Decide if FFI should only expose historical revisions
- [ ] Update docs to use consistent terminology
- [ ] Consider renaming "Revision" to "View" in FFI for clarity

alarso16:
This doesn't totally complete #1113 since I didn't touch the Proof or Iterator files. However, this does clarify a lot of the use cases, and what is valid for concurrent operations.

The best way to review this is open the godocs and reading through it, commenting on what doesn't seem clear. To reproduce the godocs, run:
```
cd ./ffi
godoc -http=:<PORT_NUM>
```

Please leave nit-picky comments! It would be great if any reader of the docs would know how to use the code.

Closes #1348, #1413

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,43 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.0.15] - 2025-11-18
+
+### 🚀 Features
+
+- Merge key-value range into trie ([#1427](https://github.com/ava-labs/firewood/pull/1427))
+- *(storage)* Replace `ArcSwap` with `Mutex` for better performance ([#1447](https://github.com/ava-labs/firewood/pull/1447))
+- *(ffi)* [**breaking**] Remove unused kvBackend interface ([#1448](https://github.com/ava-labs/firewood/pull/1448))
+- *(ffi)* Add keepalive struct to own waitgroup logic ([#1437](https://github.com/ava-labs/firewood/pull/1437))
+- Remove the last bits of arc-swap ([#1464](https://github.com/ava-labs/firewood/pull/1464))
+- *(ffi)* Fill in ffi methods for range proofs ([#1429](https://github.com/ava-labs/firewood/pull/1429))
+- *(firewood/ffi)* Add `FjallStore` ([#1395](https://github.com/ava-labs/firewood/pull/1395))
+
+### 🚜 Refactor
+
+- *(rootstore)* Replace RootStoreError with boxed error ([#1446](https://github.com/ava-labs/firewood/pull/1446))
+- Remove default 1-minute timeout on `Database.Close()` and set 1-sec in tests ([#1458](https://github.com/ava-labs/firewood/pull/1458))
+
+### 📚 Documentation
+
+- Add agent instructions ([#1445](https://github.com/ava-labs/firewood/pull/1445))
+- *(firewood)* Clean up commit steps ([#1453](https://github.com/ava-labs/firewood/pull/1453))
+- Merge code review process into CONTRIBUTING.md ([#1397](https://github.com/ava-labs/firewood/pull/1397))
+- Add comprehensive metrics documentation in METRICS.md ([#1402](https://github.com/ava-labs/firewood/pull/1402))
+
+### ⚡ Performance
+
+- *(ffi)* [**breaking**] Use fixed size Hash ([#1449](https://github.com/ava-labs/firewood/pull/1449))
+
+### 🧪 Testing
+
+- Fix `giant_node` test ([#1465](https://github.com/ava-labs/firewood/pull/1465))
+
+### ⚙️ Miscellaneous Tasks
+
+- *(ci)* Re-enable ffi-nix job ([#1450](https://github.com/ava-labs/firewood/pull/1450))
+- Relegate build equivalent check to post-merge job ([#1469](https://github.com/ava-labs/firewood/pull/1469))
+
 ## [0.0.14] - 2025-11-07
 
 ### 🚀 Features
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,22 +101,22 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e231f6134f61b71076a3eab506c379d4f36122f2af15a9ff04415ea4c3339e2"
+checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.10"
+version = "3.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e0633414522a32ffaac8ac6cc8f748e090c5717661fddeea04219e2344f5f2a"
+checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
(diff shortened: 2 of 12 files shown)


demosdemon (approved):

ffi/firewood.go
```diff
@@ -33,18 +33,35 @@ import (
...
-// A Database is a handle to a Firewood database.
-// It is not safe to call these methods with a nil handle.
+// Database is an FFI wrapper for the Rust Firewood database.
+// All functions rely on CGO to call into the underlying Rust implementation.
+// Instances are created via [New] and must be closed with [Database.Close]
+// when no longer needed.
+//
+// A Database can have outstanding [Revision] and [Proposal], which
+// access the database's memory. These must be released before closing the
+// database. See [Database.Close] for more details.
```
demosdemon, line 58: Since I merged #1429, this should include how `RangeProof` also holds onto the database memory after `Database.VerifyRangeProof` (or, that can hold off until #1468).

ffi/firewood.go
```diff
@@ -33,18 +33,35 @@ import (
 	"sync"
 )
 
+// RootLength is the hash length for all Firewood hashes.
```
rkuris, line 36: Nit: it's rare we see Firewood capitalized.

ffi/firewood.go
```diff
@@ -53,16 +70,28 @@ type Database struct {
 	outstandingHandles sync.WaitGroup
 }
 
-// Config configures the opening of a [Database].
+// Config sets the configuration parameters used when opening a [Database].
```
rkuris: Nit: prefer "holds" or "defines" to "sets" which is an action.

ffi/firewood.go
```diff
@@ -53,16 +70,28 @@ type Database struct {
...
+	// Must be non-zero.
 	FreeListCacheEntries uint
-	Revisions            uint
-	ReadCacheStrategy    CacheStrategy
+	// Revisions is the maximum number of historical revisions to keep on disk.
+	// Must be at least 2.
+	Revisions uint
+	// ReadCacheStrategy is the caching strategy used for the node cache.
+	ReadCacheStrategy CacheStrategy
 }
```
demosdemon: A thought I had unrelated to your PR, I'm curious if we should refactor this into the option functions model instead of passing a config struct. It would allow us to copy the builder pattern to the ffi later. It's not necessary now for the current options; but, if they get more convoluted, it might be worth it.

ffi/firewood.go
```diff
@@ -146,13 +188,16 @@ func (db *Database) Update(keys, vals [][]byte) (Hash, error) {
...
-// freeing proposals.
+// is not committed until [Proposal.Commit] is called. See [Database.Close] regarding
+// freeing proposals. All proposals must be freed before closing the database.
 //
 // Value Semantics:
 //   - nil value (vals[i] == nil): Performs a DeleteRange operation using the key as a prefix
 //   - empty slice (vals[i] != nil && len(vals[i]) == 0): Inserts/updates the key with an empty value
 //   - non-empty value: Inserts/updates the key with the provided value
+//
+// This function is not thread-safe with respect to other calls that reference the latest
```
rkuris, line 204: Nit: I wonder if a lock would prevent stupid errors here (probably a separate issue).

An uncontended mutex is pretty cheap.

If you agree, create an issue from this.

ffi/firewood.go
```diff
@@ -146,13 +188,16 @@ func (db *Database) Update(keys, vals [][]byte) (Hash, error) {
 }
 
 // Propose creates a new proposal with the given keys and values. The proposal
-// is not committed until [Proposal.Commit] is called. See [Database.Close] re
-// freeing proposals.
+// is not committed until [Proposal.Commit] is called. See [Database.Close] regarding
+// freeing proposals. All proposals must be freed before closing the database.
```
rkuris: Should. There is a defined behavior when you break this rule.

ffi/firewood.go
```diff
@@ -53,16 +70,28 @@ type Database struct {
...
+	// Must be non-zero.
+	NodeCacheEntries uint
+	// FreeListCacheEntries is the number of entries in the freelist cache.
+	// Must be non-zero.
 	FreeListCacheEntries uint
-	Revisions            uint
-	ReadCacheStrategy    CacheStrategy
+	// Revisions is the maximum number of historical revisions to keep on disk.
+	// Must be at least 2.
+	Revisions uint
```
RodrigoVillar: Technically, this isn't true if `RootStore` is enabled (all revisions will be kept on disk) now that #1395 is merged in.

ffi/firewood.go
alarso16: I don't think its rare? A lot of the docs that you wrote also capitalize it. It's a proper noun!

ffi/firewood.go
alarso16: I'll let #1468 do that, if that's ok. I'm not sure of what other requirements are necessary, so I think you should document it

ffi/firewood.go
alarso16: An issue is opened, worth a discussion for sure

ffi/firewood.go
alarso16: Issue created

ffi/firewood.go
```diff
@@ -87,15 +122,21 @@ const (
...
+// the same file path until the database is closed.
 func New(filePath string, conf *Config) (*Database, error) {
 	if conf == nil {
 		conf = DefaultConfig()
 	}
 	if conf.ReadCacheStrategy >= invalidCacheStrategy {
 		return nil, fmt.Errorf("invalid %T (%[1]d)", conf.ReadCacheStrategy)
 	}
-	if conf.Revisions < 2 {
+	if conf.Revisions < 2 && len(conf.RootStoreDir) == 0 {
```
alarso16: Is this correct?

RodrigoVillar: Yup

rkuris: Doesn't seem to be. Revisions must be at least 2 in all cases.

RodrigoVillar: Ah apologies, @rkuris is correct here.

ffi/firewood.go
```diff
@@ -53,17 +70,32 @@ type Database struct {
...
+	NodeCacheEntries uint
+	// FreeListCacheEntries is the number of entries in the freelist cache.
+	// Must be non-zero.
 	FreeListCacheEntries uint
-	Revisions            uint
-	ReadCacheStrategy    CacheStrategy
-	RootStoreDir         string
+	// Revisions is the maximum number of historical revisions to keep on disk.
+	// This value is ignored if RootStoreDir is set.
+	// Otherwise, must be >= 2.
```
RodrigoVillar: Hmm this isn't true; even if `RootStore` is enabled, `Revisions` will be still be used (it'll represent the number of revisions to keep around in memory).

I would clarify this as follows (feel free to modify):
```go
// Revisions is the maximum number of historical revisions to keep in memory.
// If RootStoreDir is set, then any revisions removed from memory will still be kept on disk.
// Otherwise, any revisions removed from memory will no longer be kept on disk.
// Must be >= 2
```

This ignores the edge case of Firewood (with RootStore disabled) shutting down with revisions in memory (spoiler alert: those revisions will still be kept on disk), but we can probably punt that for a later issue 😅

RodrigoVillar (approved):

PR #1474, opened 2025-11-18
feat(ffi): protect unsafe concurrent ops

### LINKED ISSUE #1473: feat(ffi): Latest revision lock in go
Nit: I wonder if a lock would prevent stupid errors here (probably a separate issue).

An uncontended mutex is pretty cheap.

If you agree, create an issue from this.

_Originally posted by @rkuris in https://github.com/ava-labs/firewood/pull/1466#discussion_r2539098012_

Especially for the latest update stuff, it seems very reasonable to provide this. I think a wider discussion about whether all necessary locking can/should be implemented in the library or given to the user, since #1466 clearly defined the concurrency requirements.

Additionally, we need to do some performance testing on whatever we decide to ensure that it doesn't degrade performance

alarso16:
The goal of this is to prevent potentially unintended side effects of using the most recent revision in a concurrent manner. This isn't intended to remove the burden of correct concurrency from the user, but prevent easy mistakes. We expect these invariants to hold, so why not guarantee it?

Closes #1473


ffi/firewood.go
```diff
@@ -174,13 +179,16 @@ func New(filePath string, conf *Config) (*Database, error) {
...
+// This function conflicts with all other calls that access the latest state of the database,
+// and will lock for the duration of this function.
 func (db *Database) Update(keys, vals [][]byte) (Hash, error) {
 	if db.handle == nil {
 		return EmptyRoot, errDBClosed
 	}
 
+	db.stateLock.Lock()
+	defer db.stateLock.Unlock()
+
```
demosdemon: if we check that the handle is null before locking, we need to check again after locking to make sure the handle wasn't set to null while we were waiting for the lock

ffi/firewood.go
```diff
@@ -347,11 +379,21 @@ func (db *Database) Close(ctx context.Context) error {
...
 
+	db.stateLock.Lock()
+	defer db.stateLock.Unlock()
 	if err := getErrorFromVoidResult(C.fwd_close_db(db.handle)); err != nil {
 		return fmt.Errorf("unexpected error when closing database: %w", err)
 	}
 
+	db.handleLock.Lock()
 	db.handle = nil // Prevent double free
+	db.handleLock.Unlock()
```
demosdemon: the lock needs to be around the entire access of the handle (i.e., when we're making the ffi call) and not just here otherwise we have no way of preventing concurrent threads from using the handle while we're closing it.

ffi/firewood.go
```diff
@@ -347,11 +379,21 @@ func (db *Database) Close(ctx context.Context) error {
...
 
 	return nil
 }
+
+func (db *Database) isClosed() bool {
+	db.handleLock.Lock()
+	defer db.handleLock.Unlock()
+	return db.handle == nil
```
demosdemon: checking if the db is nil doesn't need to be locked if we're going to bail on if it is null; checking if the db is nil needs to be locked only if we are about to immediately use it. locking and then unlocking when the result is to return if nil is unnecessary. 

In other words, TOCTOU doesn't exist if there is no use. Otherwise, TOCTOU exists if you are using the handle after unlocking.

demosdemon:
I don't know if Go has any read-write locks in stdlib, but this is a case where we need it. Most of the operations with the database handle are safe to execute concurrently (e.g., read operations). And, reading is also safe to do concurrently while committing. However, dropping is not safe to do while there are any concurrent reads or ongoing commit. So, before we drop, we need to acquire a semaphore that waits for ongoing reads to complete and prevents future reads, a lock to prevent commits, and a lock to prevent concurrent drops (which is handled by the keepAliveHandle and also why I wanted to name the function `withLock`).

alarso16:
> I don't know if Go has any read-write locks in stdlib...

It does, my bad for not noticing this! I think I fixed this issue.

ffi/proofs.go
```diff
@@ -169,6 +177,8 @@ func (db *Database) VerifyAndCommitRangeProof(
 	rootHash Hash,
 	maxLength uint32,
 ) (Hash, error) {
```
demosdemon, line 179: I think we also need to acquire the commit lock here. And should we rename it to commit lock? "State lock" is correct but wasn't my immediate thought when looking for it.

rkuris: I think this is still unresolved. Can you fix this before merging? Otherwise it looks okay to me.

rkuris (approved):

ffi/proofs.go
demosdemon: the commit lock is now acquired within the `disown` function. That seems good as it's protecting the smallest scope around actually executing the commit function.

PR #1476, opened 2025-11-19
feat: use `parking_lot` to eliminate lock poisoning

### LINKED ISSUE #1030: chore: move to parking_lot to avoid poisoned lock panics
If we ever want to get rid of these panics for poisoned locks, we can switch to [parking_lot](https://docs.rs/parking_lot) which doesn't do poisoning. Poisoning only happens when a thread panics while a lock is acquired. If we are strict about not panicking, then we don't need the poisoning feature and its extra overhead.

_Originally posted by @demosdemon in https://github.com/ava-labs/firewood/pull/1027#discussion_r2183245436_

AminR443:
This PR closes #1030 by migrating from the standard library's `Mutex` and `RwLock` implementations to `parking_lot` equivalents throughout the codebase.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -16,4 +16,9 @@ disallowed-methods = [
 disallowed-types = [
   { path = "rand::SeedableRng", replacement = "firewood_storage::SeededRng", reason = "use a prng with a user-defined seed instead", allow-invalid = true },
   { path = "rand::rngs::StdRng", replacement = "firewood_storage::SeededRng", reason = "use a prng with a user-defined seed instead", allow-invalid = true },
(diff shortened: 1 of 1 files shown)


demosdemon:
oh, we also have the ability to lint when someone uses the stdlib mutex by adding it here:
https://github.com/ava-labs/firewood/blob/78f459fe1e6d1fd64d2a3a032a3a93755e6ac5f3/clippy.toml#L16-L19

AminR443:
Oh, cool! I'll add that.

rkuris:
Did this actually improve performance? If not, let's remove the perf flags. I suspect it didn't make any measurable difference.

Issue #1470, opened 2025-11-18
refactor: in-memory revisions data structure


RodrigoVillar:
@rkuis I'm definitely on-board with having `RootStore` own constructing and maintaining its own nodestores. With our current code, refactoring would fail due to the following circular dependency:

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore

    s1 --> s2: Wants historical revision X
    s2 --> s1: Wants latest revision
```

One way to fix this would be to pass in the latest revision to `get()`:

```rust
fn get(
        &self,
        hash: &TrieHash,
        latest_revision: CommittedRevision,
    ) -> Result<Option<LinearAddress>, Box<dyn std::error::Error + Send + Sync>>;
```

However, passing in the latest revision seems odd (we expect `RootStore` to handle it's own nodestores). I think the better alternative would be to delegate the `historical` and `by_hash` fields of the `RevisionManager` into their own DS and have that be shared by both the `RevisionManager` and `RootStore`:

```rust
pub struct InMemoryRevisions {
    historical: VecDeque<CommittedRevision>,
    by_hash: HashMap<TrieHash, CommittedRevision>
}
```

Refactoring our code to use this DS would lead to the following dependency graph:

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore
    s3: InMemoryRevisions

    s1 --> s3: Maintains latest revisions
    s1 --> s2: Wants historical revision
    s2 --> s3: Wants latest revision
```

_Originally posted by @RodrigoVillar in https://github.com/ava-labs/firewood/issues/1395#issuecomment-3548415477_

RodrigoVillar:
We should also have the following: https://github.com/ava-labs/firewood/pull/1395#issuecomment-3548725875

PR #1480, opened 2025-11-20
refactor(rootstore): add `CommittedRevisionCache`

RodrigoVillar:
As mentioned in #1470, there exists the following circular dependency: 

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore

```

This PR eliminates the circular dependency above by introducing `CommittedRevisionCache`, a new FIFO data structure which holds recently committed revisions.  With this PR, we have the following dependency graph:

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore
    s3: CommittedRevisionCache

    s1 --> s3: Maintains latest revisions
    s1 --> s2: Wants historical revision
    s2 --> s3: Wants latest revision
```

This PR does the following:
- Adds the `CommittedRevisionCache` data structure
- Updates `RevisionManager` and `RootStore` to use `CommittedRevisionCache`
    - Revision construction is now handled by `RootStore` instead of the `RevisionManager`
    - Moves revision caching from the `RevisionManager` to `RootStore`
- Adds `TestRootStore` test helper to test `RootStore` caching

### DIFF
--- a/firewood/src/root_store.rs
+++ b/firewood/src/root_store.rs
@@ -23,7 +23,8 @@ pub struct RootStore {
     items: PartitionHandle,
     committed_revision_cache: Arc<RwLock<CommittedRevisionCache>>,
     /// Cache of reconstructed revisions by hash.
-    cache: Mutex<WeakValueHashMap<TrieHash, Weak<NodeStore<Committed, FileBacked>>>>,
+    reconstructed_revision_cache:
+        Mutex<WeakValueHashMap<TrieHash, Weak<NodeStore<Committed, FileBacked>>>>,
 }
 
 impl RootStore {
@@ -43,13 +44,13 @@ impl RootStore {
         let keyspace = Config::new(path).open()?;
         let items =
             keyspace.open_partition(FJALL_PARTITION_NAME, PartitionCreateOptions::default())?;
-        let cache = Mutex::new(WeakValueHashMap::new());
+        let reconstructed_revision_cache = Mutex::new(WeakValueHashMap::new());
 
         Ok(Self {
             keyspace,
             items,
             committed_revision_cache,
-            cache,
+            reconstructed_revision_cache,
         })
     }
 
@@ -99,7 +100,7 @@ impl RootStore {
         hash: &TrieHash,
     ) -> Result<Option<CommittedRevision>, Box<dyn std::error::Error + Send + Sync>> {
         // 1. Check if the committed revision is cached.
-        if let Some(v) = self.cache.lock().get(hash) {
+        if let Some(v) = self.reconstructed_revision_cache.lock().get(hash) {
             return Ok(Some(v));
         }
 
@@ -128,7 +129,9 @@ impl RootStore {
         ));
 
         // Cache for future lookups.
-        self.cache.lock().insert(hash.clone(), nodestore.clone());
+        self.reconstructed_revision_cache
+            .lock()
+            .insert(hash.clone(), nodestore.clone());
 
         Ok(Some(nodestore))
     }
@@ -188,7 +191,7 @@ mod tests {
 
         let hash = TrieHash::from_bytes([1; 32]);
         root_store
-            .cache
+            .reconstructed_revision_cache
             .lock()
             .insert(hash.clone(), revision.clone());
 


firewood/src/manager.rs
```diff
@@ -69,25 +68,39 @@ pub struct ConfigManager {
...
-    by_rootstore: Mutex<WeakValueHashMap<TrieHash, Weak<NodeStore<Committed, FileBacked>>>>,
     threadpool: OnceLock<ThreadPool>,
     root_store: Box<dyn RootStore + Send + Sync>,
 }
 
+#[derive(Debug, Default)]
+pub struct InMemoryRevisions {
+    latest: VecDeque<CommittedRevision>,
+    by_hash: HashMap<TrieHash, CommittedRevision>,
+    allow_space_reuse: bool,
```
demosdemon: Want to point out that this is replacing two locks (`RwLock<VecDeque<CommittedRevision>>` and `RwLock<HashMap<TrieHash, CommittedRevision>>`) with one (essentially `RwLock<(VecDeque<CommittedRevision>, HashMap<TrieHash, CommittedRevision>)>`. This is a non-trivial change with an unknown performance hit and it may be better to move the two locks into `InMemoryRevisions` instead around it.

We also need to acquire a lock in order to determine if `allow_space_reuse` is true or false, when the value will not change after it is set. If the locks were places around `latest` and `by_hash`, the same behavior as before where we lock each independently and also gives us the ability to read `allow_space_reuse` lock-free.

firewood/src/manager.rs
```diff
@@ -122,28 +135,37 @@ impl RevisionManager {
...
 
-        if let Some(hash) = nodestore.root_hash().or_default_root_hash() {
-            manager.by_hash.write().insert(hash, nodestore.clone());
+        // Add the initial nodestore to the in-memory revisions
+        {
+            let mut revisions = in_memory_revisions.write();
+            revisions.latest.push_back(nodestore.clone());
+            if let Some(hash) = nodestore.root_hash().or_default_root_hash() {
+                revisions.by_hash.insert(hash, nodestore.clone());
+            }
```
demosdemon: I wonder if there a way we can construct the `InMemoryRevisions` with this initial revision before we wrap it in the mutex.

firewood/src/manager.rs
```diff
@@ -69,25 +68,39 @@ pub struct ConfigManager {
...
 }
 
-type CommittedRevision = Arc<NodeStore<Committed, FileBacked>>;
+pub type CommittedRevision = Arc<NodeStore<Committed, FileBacked>>;
 type ProposedRevision = Arc<NodeStore<Arc<ImmutableProposal>, FileBacked>>;
 
 #[derive(Debug)]
 pub(crate) struct RevisionManager {
     /// Maximum number of revisions to keep on disk
     max_revisions: usize,
```
demosdemon: nit: excess whitespace around the field. I'd remove it or add a doc-comment.

firewood/src/manager.rs
```diff
@@ -203,19 +225,20 @@ impl RevisionManager {
         // If `RootStore` allows space reuse, add the oldest revision's nodes to the free list.
         // If you crash after freeing some of these, then the free list will point to nodes that are not actually free.
         // TODO: Handle the case where we get something off the free list that is not free
-        while self.historical.read().len() >= self.max_revisions {
-            let oldest = self
-                .historical
-                .write()
+        let mut in_memory_revisions = self.in_memory_revisions.write();
```
demosdemon: This `write` lock over `InMemoryRevisions` will prevent any thread from acquiring the read lock. The change to wrap the lock around `InMemoryRevisions` instead of each field means a parallel thread will be unable to read from the `by_hash` map while you're pruning in-memory revisions. Your change does keep us from repeatedly locking and re-locking the `by_hash` map while draining it. The lock is also held the entire time `persist` is executing; which is not desired at all.

firewood/src/manager.rs
RodrigoVillar: 100% agree on being able to read `allow_space_reuse` lock-free.

W.R.T the two locks, I'm unsure about the performance hit but isn't a single lock what we actually want? In the case of writes, any inserts/removals to `latest` should be reflected in `by_hash` and vice versa. This way, `latest` and `by_hash` are always in sync except for the following case:

```rust
// Not in sync if None
if let Some(hash) = committed.root_hash().or_default_root_hash() {
    self.by_hash.insert(hash, committed.clone());
}
```

firewood/src/manager.rs
RodrigoVillar: Removed the whitespace here: 29a6a07

firewood/src/manager.rs
```diff
@@ -122,28 +163,26 @@ impl RevisionManager {
...
         let nodestore = Arc::new(NodeStore::open(storage.clone())?);
+
+        let committed_revision_cache = Arc::new(RwLock::new(CommittedRevisionCache::default()));
+        committed_revision_cache.write().push(nodestore.clone());
+
+        let root_store = config
+            .root_store_dir
+            .map(|path| RootStore::new(path, committed_revision_cache.clone()))
+            .transpose()
+            .map_err(RevisionManagerError::RootStoreError)?;
```
RodrigoVillar, line 176: This code resolves #1485

firewood/src/manager.rs
RodrigoVillar: As of the time of writing this comment, I've added several methods to `CommittedRevisionCache` including the `push()` method, which makes adding the initial revision trivial.

firewood/src/manager.rs
RodrigoVillar: Good point; with the introducing of the methods on `CommittedRevisionCache`, we no longer hold a write lock throughout most of `commit()`.

firewood/src/root_store.rs
```diff
@@ -57,24 +76,138 @@ impl RootStore {
...
+            .get_latest_revision()
+            .expect("there is always one revision")
+            .clone();
+
+        // 3. Construct the committed revision.
+        let nodestore = Arc::new(NodeStore::with_root(
+            hash.clone().into_hash_type(),
+            addr,
+            latest_nodestore,
+        ));
```
demosdemon, line 129: I think the `with_root` method needs a change to better reflect what you're doing.

Passing in `latest_nodestore` doesn't make sense until I realize you're only doing it to acquire a handle to the storage backend. I'd rather not do that because it's very unclear the intent here. The revision cache should probably own its own handle to the file backend. It would remove the `expect` from looking up the latest revision and also make it clearer that you only need the storage and not the entire node store.

There's also a debug-only panic if node at the address has the wrong hash, but I think that probably should be a runtime check. Hashing the root node should be relatively quick because it's not recursive.

RodrigoVillar: Thanks for the feedback. After some thought, I agree that passing in `latest_nodestore` to `with_root()` seems odd and we can just pass in the file backend instead to both `with_root()` and `RootStore` itself (this is also much cleaner). 

Going to close this PR in favor of #1495 since the `CommittedRevisionCache` logic (which is no longer necessary) takes up a majority of this PR.

PR #1481, opened 2025-11-20
refactor(firewood/db): unify reopen methods

RodrigoVillar:
IMO, the `reopen()` method of `TestDb` is a misnomer - while this method does not truncate the existing Firewood file, it uses the default configuration for the new `TestDb` instance, ignoring the prior instance's configuration. As a consumer, I would expect `reopen()` to maintain the existing configuration.

This PR does the following:
- Unifies `reopen()` and `reopen_with_config()` into a single method
- Stores the `dbconfig` in `testDb` to allow the database configuration to persist across reopens
- Adds documentation to both `reopen()` and `replace()`
- Removes `into_root_store()` test helpers since `RootStore` reopens are now handled via the database configuration 

This PR is a precursor to #1481.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -406,7 +406,6 @@ mod test {
 
     use crate::db::{Db, Proposal, UseParallel};
     use crate::manager::RevisionManagerConfig;
-    use crate::root_store::RootStore;
     use crate::v2::api::{Db as _, DbView, Proposal as _};
 
     use super::{BatchOp, DbConfig};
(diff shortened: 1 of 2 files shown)


firewood/src/db.rs
```diff
@@ -667,12 +657,7 @@ mod test {
                 .build(),
         );
         insert_commit(&db, 1);
-        let db = db.reopen_with_config(
-            DbConfig::builder()
-                .truncate(false)
-                .use_parallel(UseParallel::Always)
-                .build(),
-        );
+        let db = db.reopen();
```
RodrigoVillar, line 660: Note to reviewers: `truncate` is `false` by default and the `use_parallel()` setting is preserved in `testDb`'s configuration, so using `db.reopen()` is equivalent here.

PR #1482, opened 2025-11-21
refactor(db/revisionmanager): move `RootStore` creation to `RevisionManager`

RodrigoVillar:
In preparation of moving in-memory revisions to their own data structure (#1470), this PR moves creation of `RootStore` into the `RevisionManager` - this move will allow us to inject the in-memory revisions data structure into `RootStore` during creation.

This PR is a precursor to #1480.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -1101,7 +1101,7 @@ mod test {
         let dbconfig = DbConfig::builder()
             .manager(RevisionManagerConfig::builder().max_revisions(5).build())
             .build();
-        let db = TestDb::new_with_fjall_store(dbconfig.clone());
+        let db = TestDb::new_with_fjall_store(dbconfig);
 
         // Create and commit 10 proposals
         let key = b"root_store";


firewood/src/manager.rs
```diff
@@ -61,6 +61,9 @@ pub struct ConfigManager {
     /// existing contents will be lost.
     #[builder(default = false)]
     pub truncate: bool,
+    /// `RootStore` directory path
+    #[builder(default = None)]
+    pub root_store_dir: Option<PathBuf>,
```
RodrigoVillar, line 66: Q for reviewers: this new config field does not remove the existing `RootStoreDir` field in` DbConfig`:

https://github.com/ava-labs/firewood/blob/b5e22b7c17aa7ce95d14c2bf72eaba874f8763b6/firewood/src/db.rs#L128-L130

Similar patterns exist for the `truncate` and `create` fields and so I though this was fine, but wanted to double check.

firewood/src/manager.rs
```diff
@@ -123,6 +122,13 @@ impl RevisionManager {
         // from opening the same database simultaneously
         fb.lock()?;
 
+        let root_store: Box<dyn RootStore + Send + Sync> = match config.root_store_dir {
+            Some(path) => {
+                Box::new(FjallStore::new(path).map_err(RevisionManagerError::RootStoreError)?)
+            }
+            None => Box::new(NoOpStore {}),
```
demosdemon, line 129: Its starting to feel like we don't need `NoOpStore` and could instead implement `RootStore` on `Option<T> where T: RootStore` that does the same thing as `NoOpStore` if the option is `None` (which may also require a pass-through implementation of `RootStore` on `Box<T>` OR `Box<dyn RootStore + Send + Sync>` (but not both).

demosdemon (approved):

firewood/src/manager.rs
RodrigoVillar: Hmm good point - creating an issue for it here: #1483.

Issue #1483, opened 2025-11-21
Investigate removing `NoOpStore`


RodrigoVillar:
Its starting to feel like we don't need `NoOpStore` and could instead implement `RootStore` on `Option<T> where T: RootStore` that does the same thing as `NoOpStore` if the option is `None` (which may also require a pass-through implementation of `RootStore` on `Box<T>` OR `Box<dyn RootStore + Send + Sync>` (but not both).

_Originally posted by @demosdemon in https://github.com/ava-labs/firewood/pull/1482#discussion_r2550410186_

PR #1484, opened 2025-11-21
refactor(rootstore): remove box pointer

RodrigoVillar:
While taking a stab at #1483, I noticed after removing `NoOpStore` in favor of `Option<T>`, there's actually no need for the `Box` pointer since, if we have `Some`, it'll always be an instance of `FjallStore`. 

This PR does the following:
- Removes the usage of `Box` pointers for `RootStore`
- Removes the `RootStore` trait and renames `FjallStore => RootStore`
- Updates code to use new `RootStore` naming and logic

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -239,12 +239,12 @@ impl RevisionManager {
         committed.persist()?;
 
         // 4. Persist revision to root store
-        if let Some(store) = &self.root_store {
-            if let (Some(hash), Some(address)) = (committed.root_hash(), committed.root_address()) {
-                store
-                    .add_root(&hash, &address)
-                    .map_err(RevisionManagerError::RootStoreError)?;
-            }
+        if let Some(store) = &self.root_store
+            && let (Some(hash), Some(address)) = (committed.root_hash(), committed.root_address())
+        {
+            store
+                .add_root(&hash, &address)
+                .map_err(RevisionManagerError::RootStoreError)?;
         }
 
         // 5. Set last committed revision


firewood/src/manager.rs
```diff
@@ -214,8 +213,8 @@ impl RevisionManager {
                 self.by_hash.write().remove(hash);
             }
 
-            // We reap the revision's nodes only if `RootStore` allows space reuse.
-            if self.root_store.allow_space_reuse() {
+            // We reap the revision's nodes only if `RootStore` does not exist.
+            if self.root_store.is_none() {
```
RodrigoVillar, line 217: This is because `allow_space_reuse() == true` if and only if `self.root_store.is_none() == true`.

demosdemon (approved):

firewood/src/manager.rs
```diff
@@ -122,11 +122,9 @@ impl RevisionManager {
         // from opening the same database simultaneously
         fb.lock()?;
 
-        let root_store: Box<dyn RootStore + Send + Sync> = match config.root_store_dir {
-            Some(path) => {
-                Box::new(FjallStore::new(path).map_err(RevisionManagerError::RootStoreError)?)
-            }
-            None => Box::new(NoOpStore {}),
+        let root_store: Option<RootStore> = match config.root_store_dir {
```
rkuris, line 125: This match degenerated into a map and could be written much simpler.

rkuris (approved):

firewood/src/manager.rs
RodrigoVillar: Thanks for the feedback! Tracking this in #1485 and should be closed in #1480.

PR #1486, opened 2025-11-24
fix(ffi): hold onto verification context in next key range

demosdemon:
We need to know the expected end key when determining the next key in of the range after applying the range proof.


ffi/proofs.go
```diff
@@ -199,6 +199,12 @@ func (db *Database) VerifyAndCommitRangeProof(
 //
 // FindNextKey can only be called after a successful call to [*Database.VerifyRangeProof] or
 // [*Database.VerifyAndCommitRangeProof].
+//
+// The next key range indicates the next `(startKey, endKey]` range of keys that
```
alarso16, line 203: That is also the way this was pre-this PR, correct?

ffi/src/proofs/range.rs
```diff
@@ -168,27 +164,77 @@ impl<'db> RangeProofContext<'db> {
...
+
+        let hash = result?;
+        self.proposal_state = Some(ProposalState::Committed(hash.clone()));
 
-        self.state = RangeProofState::Committed(hash.clone());
         Ok(hash)
     }
+
+    fn find_next_key(&mut self) -> Result<Option<KeyRange>, api::Error> {
+        // TODO(#352): proper implementation, this naively retuns the last key in
```
alarso16:
```suggestion
        // TODO(#352): proper implementation, this naively returns the last key in
```

ffi/src/proofs/range.rs
```diff
@@ -168,27 +164,77 @@ impl<'db> RangeProofContext<'db> {
...
         self.verify(root, start_key, end_key, max_length)?;
 
-        let proposal_handle = if let Some(proposal) = self.state.take_proposal() {
-            proposal
-        } else {
-            db.merge_key_value_range(start_key, end_key, self.proof.key_values())?
-                .handle
+        let mut allow_rebase = true;
+        let proposal_handle = match self.proposal_state.take() {
+            Some(ProposalState::Committed(hash)) => {
```
alarso16, line 214: Does this mean it's safe to call `VerifyAndCommit` multiple times?

ffi/proofs.go
demosdemon: Correct. And, after looking at <https://github.com/ava-labs/avalanchego/blob/70148edc6ecabb2836b89d890718011b892b7063/x/sync/manager.go#L716-L729>, I think this is the expected return value.

Also looking at `completeWorkItem`, a better implementation of `FindNextKey` would be an iterator of something that can be easily converted into many `unprocessedWorkItems` and `processedWorkItems`.

ffi/src/proofs/range.rs
demosdemon: this does three things, and I'll come back to add comments before merging since its not clear:

1. if already committed, calling `verify_and_commit` is a no-op and returns the committed hash as is
2. if a proposal has been built before, try to commit but if the commit fails because the latest revision has changed, reparent and try again. This is because we don't provide any other way of doing that.
3. If there isn't a proposal, then create one and don't try to reparent if cause that shouldn't happen (TOCTOU is still possible but less likely).

ffi/proofs.go
alarso16: `completeWorkItem` does split the remaining range in 2 by looking at the midpoint. Do we want to "hijack" that implementation? What I mean is, we can achieve sufficient parallelism just by getting a next key and relying on the `Manager` to partition ranges?

PR #1489, opened 2025-11-25
fix: mark node as allocated before serializing parent

demosdemon:
#1488 found a panic during serialization of a node, indicating that a child node was either not hashed or not allocated and therefore could not serialize the parent node. After a walk through of the code, I noticed that after #1389, we began serializing nodes in batches in order to prevent corrupting the free-list after a failed write and introduced a case where this could occur. Previously, we would serialize a node, write it to storage, and then mark it as allocated in one step. Now we serialize and allocate the node and add the node to a batch before moving onto the next node. Like before, we only mark the node as allocated after writing it to disk. This means that if the batch contained child and parent, the parent node fails to serialize because it cannot reference the address.

This change refactors the allocation logic to mark nodes as allocated during batching. This ensures that if its parent is also in the batch, it can successfully serialize. This linear address itself is not read from by viewers of the `Allocated` state because the node is not considered committed until the entire write has completed (i.e., after the root and header are written). This means this change does not introduce any early reads of uncommitted data.

In the course of writing tests, I discovered that the arena allocator returns the number of bytes by the arena, not the number of bytes occupied by items within the arena. This meant that the batching logic should have always generated a batch of 1 element. To fix this, I track the allocated length of each area. This successfully triggered the bug I discovered; however, this also means the bug might not be the source of the original panic.

### DIFF
--- a/storage/src/nodestore/persist.rs
+++ b/storage/src/nodestore/persist.rs
@@ -237,6 +237,7 @@ where
     F: FnMut(Vec<(&[u8], crate::LinearAddress, MaybePersistedNode)>) -> Result<(), FileIoError>,
 {
     let mut allocated_objects = Vec::new();
+    let mut allocated_len = 0_usize;
 
     // Process each unpersisted node directly from the iterator
     for node in UnPersistedNodeIterator::new(node_store) {
@@ -248,18 +249,23 @@ where
         let (slice, persisted_address, idx_size) =
             serialize_node_to_bump(bump, &shared_node, node_allocator)?;
 
+        // NOTE(#1488): we need to set the address so that the parent node can
+        // reference it when they are serialized within the same batch.
+        node.allocate_at(persisted_address);
+        allocated_len = allocated_len.saturating_add(idx_size);
         allocated_objects.push((slice, persisted_address, node));
 
         // we pause if we can't allocate another node of the same size as the last one
         // This isn't a guarantee that we won't exceed bump_size_limit
         // but it's a good enough approximation
-        let might_overflow = bump.allocated_bytes() > bump_size_limit.saturating_sub(idx_size);
+        let might_overflow = allocated_len > bump_size_limit.saturating_sub(idx_size);
         if might_overflow {
             // must persist freelist before writing anything
             node_allocator.flush_freelist()?;
             write_fn(allocated_objects)?;
             allocated_objects = Vec::new();
             bump.reset();
+            allocated_len = 0;
         }
     }
     if !allocated_objects.is_empty() {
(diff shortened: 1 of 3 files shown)


storage/src/nodestore/persist.rs
```diff
@@ -248,18 +249,23 @@ where
         let (slice, persisted_address, idx_size) =
             serialize_node_to_bump(bump, &shared_node, node_allocator)?;
 
+        // NOTE(#1488): we need to set the address so that the parent node can
+        // reference it when they are serialized within the same batch.
+        node.allocate_at(persisted_address);
```
rkuris, line 254: You cannot do this. If you do, someone else can (and will) call read on another thread before it is persisted, returning garbage.

rkuris: > You cannot do this. If you do, someone else can (and will) call read on another thread before it is persisted, returning garbage.

actually wait, you can. The reader will use the in-memory copy.

storage/src/nodestore/persist.rs
```diff
@@ -248,18 +249,23 @@ where
...
+        allocated_len = allocated_len.saturating_add(idx_size);
         allocated_objects.push((slice, persisted_address, node));
 
         // we pause if we can't allocate another node of the same size as the last one
         // This isn't a guarantee that we won't exceed bump_size_limit
         // but it's a good enough approximation
-        let might_overflow = bump.allocated_bytes() > bump_size_limit.saturating_sub(idx_size);
+        let might_overflow = allocated_len > bump_size_limit.saturating_sub(idx_size);
```
rkuris, line 261: This is fine, but I think we should use a method on the allocator itself if possible. There is [this chunk_capacity method](https://docs.rs/bumpalo/latest/bumpalo/struct.Bump.html#method.chunk_capacity) we might be able to use.

storage/src/nodestore/persist.rs
demosdemon: Can you explain how that is possible? The address is not used for reading unless the MaybePersistedNode is `MaybePersisted::Persisted`. Do you have another suggestion for fixing the problem of letting parents discover their children's addresses during serialization?

rkuris: Totally my mistake, I misread this as "persist_at" which would have been bad.

storage/src/nodestore/persist.rs
demosdemon: I started implementing using `chunk_capacity`; but, I found that it only reports the capacity of the "current" slab used by the arena. If the arena were to allocate an additional slab, the reported `chunk_capacity` will jump up to reflect what's unused in the new slab. And, the `Vec` container we write to for serialization will perform [exponential growth on reallocs](https://github.com/fitzgen/bumpalo/blob/e1eff20bee477aa34b4b040eae6731e95dc386f7/src/collections/raw_vec.rs#L349-L365) which could trigger a new slab to be allocated even if it is mostly unused. We could use `Bump::set_allocation_limit` but I would want to do that only after refactoring the serialization logic check for allocation failures instead of panicking.

PR #1491, opened 2025-11-26
fix(parallel): track deleted nodes from root branch in parallel propose

AminR443:
Parallel propose was missing some deleted nodes that sequential propose tracked. When workers take children from the root branch, those persisted nodes weren't being marked as deleted. 

This fix adds tracking for root-level children taken during worker creation and deduplicates the deleted list by address to handle repeated updates across parallel workers. Also adds a test comparing deleted nodes between parallel and normal propose.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -833,9 +833,7 @@ mod test {
         initial_parallel_proposal.commit().unwrap();
         initial_single_proposal.commit().unwrap();
 
-        // Second batch: update only the first key. This exercises the deletion
-        // bookkeeping in a configuration where parallel workers were used in
-        // the initial batch and only a subset of keys are updated.
+        // Second batch: update only the first key.
         let update_keys: Vec<[u8; 1]> = vec![[0x00]];
         let update_values: Vec<Box<[u8]>> = vec![Box::new([3u8])];
 


storage/src/nodestore/mod.rs
```diff
@@ -590,10 +590,23 @@ impl<S: ReadableStorage> TryFrom<NodeStore<MutableProposal, S>>
     fn try_from(val: NodeStore<MutableProposal, S>) -> Result<Self, Self::Error> {
         let NodeStore {
             header,
-            kind,
+            mut kind,
             storage,
         } = val;
 
+        // Deduplicate deleted nodes by persisted address within a batch across parallel workers.
```
bernard-avalabs: I'm curious to know where the duplicates are coming from. The changes in this PR should only be adding deleted nodes from the children of the root (one per worker) which should be distinct from the deleted nodes added by the workers as they are only adding nodes from their own sub-trie.

Maybe the duplicates are being added in `postprocess_trie` as that function is returning the trie back to its canonical form which may require removing the root node and adding a deleted node. If it is only happening there, perhaps we can remove this dedup and add a dedup check in `postprocess_trie`.

AminR443: I removed the deduplication code, it was no longer necessary and survived the refactor.

PR #1493, opened 2025-11-27
ci(perf): Track Firewood Performance via AvalancheGo Benchmarks

### LINKED ISSUE #1494: Track Firewood Performance via AvalancheGo Reexecution Benchmarks
## Problem

We have no safeguard against performance regressions. When code slows down Firewood by 10%, we don't catch it until production. This costs engineering time debugging issues that could have been prevented.

We need a system that shows when a commit degrades performance.

## Stage 1: Setup Integration Points

Build the cross-repo workflow infrastructure to enable performance tracking.

**Integration points to establish:**

1. Firewood workflow triggers AvalancheGo benchmark workflow
2. AvalancheGo builds Firewood from specified commit/branch
3. AvalancheGo runs C-Chain reexecution benchmark
4. AvalancheGo uploads results as artifact
5. Firewood downloads artifact
6. Firewood stores results with **github-action-benchmark**
7. Results visualized on GitHub Pages

**Why this architecture:**
- Reexecution benchmarks require AvalancheGo infrastructure (Coreth, block data, AWS credentials)
- On-demand triggers avoid dependency matrix failures (avalanchego <-> coreth -> firewood)
- Each repo owns its domain: AvalancheGo runs tests, Firewood tracks results

```mermaid
flowchart TB
    subgraph Firewood["**Firewood Repo** (Performance Tracking Owner)"]
        FW_Trigger["1. Trigger Workflow<br/>via GitHub API"]
        FW_Poll["4. Poll for Completion"]
        FW_Download["5. Download Artifact"]
        FW_Benchmark["6. github-action-benchmark"]
        FW_Pages["7. GitHub Pages<br/>Visualization"]
        
        FW_Trigger --> FW_Poll
        FW_Poll --> FW_Download
        FW_Download --> FW_Benchmark
        FW_Benchmark --> FW_Pages
    end
    
    subgraph AvalancheGo["**AvalancheGo Repo** (Benchmark Execution Owner)"]
        AG_Workflow["Workflow triggered with<br/>firewood-commit param"]
        AG_Build["2. Build Firewood<br/>from source"]
        AG_Run["3. Run C-Chain<br/>Reexecution Benchmark"]
        AG_Upload["Upload Results Artifact"]
        
        AG_Workflow --> AG_Build
        AG_Build --> AG_Run
        AG_Run --> AG_Upload
    end
    
    FW_Trigger -->|"GitHub API<br/>repository_dispatch"| AG_Workflow
    AG_Upload -->|"Artifact"| FW_Poll
```

**Stage 1 tasks:**

Firewood (this repo):
- Create workflow that triggers AvalancheGo via GitHub API
- Poll for completion and download artifact
- Feed results to github-action-benchmark
- Configure GitHub Pages
- Documentation

PR Progress:
- https://github.com/ava-labs/firewood/pull/1493

AvalancheGo:
- Workflow accepts firewood-commit parameter (done, PR: [link])
- Builds Firewood from source and runs benchmark (done)
- Uploads results artifact (done)

PR Progress:
- https://github.com/ava-labs/avalanchego/pull/4650

**Outcome:** Performance data flows from AvalancheGo to Firewood, visualized by commit.
**Trade-off:** Undisclosed number of false positives for integration (pipeline) and historical baseline tracking

## Stage 2: Statistical Analysis

Once the pipes are working, add PerfProbe for statistical rigor, confidence intervals, and automated regression detection.

Use Polyrepo tool for managing cross-repo workflow infrastructure

Elvis339:
## Why

Track C-Chain reexecution benchmark performance over time. Catch regressions before production.

Closes #1494

## How

Firewood → triggers AvalancheGo benchmark → downloads results → publishes to GitHub Pages

Screenshot 2026-01-27 at 16 12 02

## Changes

- `scripts/bench-cchain-reexecution.sh`
  - Trigger AvalancheGo's C-Chain reexecution benchmark
  - Poll for workflow registration
  - Wait for completion and download artifacts

- `.github/workflows/track-performance.yml`
  - Orchestrate benchmark trigger via the script
  - Publish results to GitHub Pages (main → `bench/`, branches → `dev/bench/{branch}/`)

## Usage
```
# Auth
nix run ./ffi#gh -- auth login
export GH_TOKEN=$(nix run ./ffi#gh -- auth token)

# Predefined test
./scripts/bench-cchain-reexecution.sh trigger firewood-101-250k

# With specific Firewood version
FIREWOOD_REF=v0.0.18 ./scripts/bench-cchain-reexecution.sh trigger firewood-101-250k

# Custom block range
RUNNER=avalanche-avalanchego-runner-2ti \
FIREWOOD_REF=v0.0.18 \
CONFIG=firewood \
START_BLOCK=1 \
END_BLOCK=100 \
BLOCK_DIR_SRC=cchain-mainnet-blocks-200-ldb \
./scripts/bench-cchain-reexecution.sh trigger

# Other commands
./scripts/bench-cchain-reexecution.sh tests    # list available tests
./scripts/bench-cchain-reexecution.sh list     # list recent runs
./scripts/bench-cchain-reexecution.sh status <run_id>
```
- Set `FIREWOOD_REF=v0.0.18` explicitly. Without it, the workflow builds from `HEAD`, which currently fails due to changes in FFI layer

### Related
AvalancheGo: https://github.com/ava-labs/avalanchego/pull/4650
Local iteration PR: https://github.com/ava-labs/firewood/pull/1642
Followup: https://github.com/ava-labs/firewood/issues/1639

### DIFF
--- a/scripts/bench-cchain-reexecution.sh
+++ b/scripts/bench-cchain-reexecution.sh
@@ -75,7 +75,6 @@ declare -A TESTS=(
 
 log() { echo "==> $1" >&2; }
 
-# Use GitHub Actions error annotation format in CI for better visibility
 err() {
     if [[ "${GITHUB_ACTIONS:-}" == "true" ]]; then
         echo "::error::$1"
@@ -84,21 +83,19 @@ err() {
     fi
 }
 
-die() { err "$1"; exit 1; }
-
 require_gh() {
     if ! command -v gh &>/dev/null; then
-        die "gh CLI not found. Run from nix shell: nix develop ./ffi"
+        err "gh CLI not found. Run from nix shell: nix develop ./ffi"; exit 1
     fi
     if [[ -z "${GH_TOKEN:-}" ]]; then
-        die "GH_TOKEN is required"
+        err "GH_TOKEN is required"; exit 1
     fi
 }
 
 # Prevent command injection via malicious ref names passed to gh CLI
 validate_ref() {
     local ref="$1" name="$2"
-    [[ "$ref" =~ ^[a-zA-Z0-9/_.-]+$ ]] || die "$name contains invalid characters: $ref"
+    [[ "$ref" =~ ^[a-zA-Z0-9/_.-]+$ ]] || { err "$name contains invalid characters: $ref"; exit 1; }
 }
 
 # Verify a run's inputs match what we triggered with
@@ -152,7 +149,7 @@ poll_workflow_registration() {
     
     # Verify we can list runs (fail fast on permission issues)
     if ! gh run list --repo "$AVALANCHEGO_REPO" --workflow "$WORKFLOW_NAME" --limit 1 &>/dev/null; then
-        die "Cannot list runs in $AVALANCHEGO_REPO. Check token permissions."
+        err "Cannot list runs in $AVALANCHEGO_REPO. Check token permissions."; exit 1
     fi
     
     for ((i=0; i<POLL_TIMEOUT; i++)); do
@@ -187,7 +184,7 @@ poll_workflow_registration() {
         ((i % 10 == 0)) && ((i > 0)) && log "Polling (${i}s)"
     done
     
-    die "Workflow not found after ${POLL_TIMEOUT}s. The workflow may have been triggered but couldn't be detected. Check: https://github.com/$AVALANCHEGO_REPO/actions/workflows"
+    err "Workflow not found after ${POLL_TIMEOUT}s. The workflow may have been triggered but couldn't be detected. Check: https://github.com/$AVALANCHEGO_REPO/actions/workflows"; exit 1
 }
 
 trigger_workflow() {
@@ -217,7 +214,7 @@ trigger_workflow() {
     else
         # Custom mode: block params required, CURRENT_STATE_DIR_SRC optional (empty = genesis)
         [[ -z "${START_BLOCK:-}${END_BLOCK:-}${BLOCK_DIR_SRC:-}" ]] && \
-            die "Provide a test name or set START_BLOCK, END_BLOCK, BLOCK_DIR_SRC"
+            err "Provide a test name or set START_BLOCK, END_BLOCK, BLOCK_DIR_SRC"; exit 1
         : "${START_BLOCK:?START_BLOCK required}"
         : "${END_BLOCK:?END_BLOCK required}"
         : "${BLOCK_DIR_SRC:?BLOCK_DIR_SRC required}"
@@ -271,7 +268,7 @@ download_artifact() {
     find "$DOWNLOAD_DIR" -name "benchmark-output.json" -type f -exec mv {} "$output_file" \;
     find "$DOWNLOAD_DIR" -mindepth 1 -type d -delete 2>/dev/null || true
     
-    [[ -f "$output_file" ]] || die "No benchmark results found"
+    [[ -f "$output_file" ]] || { err "No benchmark results found"; exit 1; }
     log "Results: $output_file"
 }
 
@@ -309,7 +306,7 @@ cmd_trigger() {
 }
 
 cmd_status() {
-    [[ -z "${1:-}" ]] && die "usage: $0 status <run_id>"
+    [[ -z "${1:-}" ]] && { err "usage: $0 status <run_id>"; exit 1; }
     require_gh
     check_status "$1"
 }


Elvis339:
~Do not merge blocked on: https://github.com/ava-labs/firewood/actions/runs/19737463191/job/56552736887#step:7:46~

Unblocked, test runs:
- v0.0.15: https://github.com/ava-labs/firewood/actions/runs/19871663903
- v0.0.14: https://github.com/ava-labs/firewood/actions/runs/19871880354

README.md
RodrigoVillar, line 1: Could we move this to `METRICS.md`? I think it would fit better there + given that this is software that only we (Ava Labs) can run, I don't think it should be in the `README.md` (external facing file)

.github/workflows/track-performance.yml
RodrigoVillar, line 1: As discussed offline, I think it would be easier if we could encapsulate invocation of this workflow in a `just` command.

Elvis339: Good feedback, added: https://github.com/ava-labs/firewood/pull/1493/commits/73fc78193b29ff8bb67d352678058e92cf02f314

README.md
Elvis339: Done: https://github.com/ava-labs/firewood/pull/1493/commits/668cef309346d3e5e44276c26e43fd0b05bf74cd

METRICS.md
```diff
@@ -284,3 +284,46 @@ rate(firewood_space_from_end[5m])
...
+### Running Benchmarks Locally
+
+Benchmarks can be triggered locally using just commands (requires nix):
+
+```bash
+# Run full benchmark: trigger, wait, download results
+just benchmark
+
+# With specific versions
+just benchmark v0.0.15 master c-chain-reexecution-firewood-101-250k avalanche-avalanchego-runner-2ti
```
RodrigoVillar: Could we improve this? Right now, it's not clear what this command does (IMO, this is analogous to having something like `just benchmark true true true` where it's unclear what each of the arguments does)

RodrigoVillar (changes requested):

METRICS.md
Elvis339: Just supports named params: https://github.com/ava-labs/firewood/pull/1493/commits/1e2bdd3d2699859fca3dbf4007b03236af56b631

Thanks for the feedback.

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+      avalanchego:
+        description: 'AvalancheGo commit/branch/tag to test against'
+        default: 'master'
+      task:
+        description: 'Predefined task (leave empty to use custom parameters below)'
+        type: choice
+        options:
+          - c-chain-reexecution-firewood-101-250k
+          - c-chain-reexecution-firewood-33m-33m500k
+          - c-chain-reexecution-firewood-33m-40m
```
RodrigoVillar: Q: why do we have these predefined tasks?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+      config:
+        description: 'VM config (e.g., firewood, hashdb)'
+        default: ''
+      start-block:
+        default: ''
+      end-block:
+        default: ''
+      block-dir-src:
+        description: 'Block directory source e.g., cchain-mainnet-blocks-1m-ldb (without S3 path)'
```
RodrigoVillar:
```suggestion
        description: 'Block directory source (e.g., cchain-mainnet-blocks-1m-ldb [without S3 path])'
```

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+        default: ''
+      current-state-dir-src:
+        description: 'Current state directory source e.g., cchain-mainnet-blocks-30m-40m-ldb (without S3 path)'
```
RodrigoVillar:
```suggestion
        description: 'Current state directory source (e.g., cchain-mainnet-blocks-30m-40m-ldb [without S3 path])'
```

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+              -f firewood="$FIREWOOD" \
+              -f config="${{ inputs.config }}" \
+              -f start-block="${{ inputs.start-block }}" \
+              -f end-block="${{ inputs.end-block }}" \
+              -f block-dir-src="${{ inputs.block-dir-src }}" \
+              -f current-state-dir-src="${{ inputs.current-state-dir-src }}" \
+              -f runner="${{ inputs.runner }}" \
+              $LIBEVM_FLAG
+            
+            sleep 10
```
RodrigoVillar: Q: what's the purpose of this `sleep` here?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
+name: Track Performance
+
+on:
+  workflow_dispatch:
+    inputs:
+      firewood:
+        description: 'Firewood commit/branch/tag to test (leave empty for current HEAD)'
```
rkuris: Q: Current HEAD of what? main? Or whatever is actually checked out?

.github/workflows/track-performance.yml
rkuris: looks like there's a race between when you run 'gh workflow run' and 'gh run list' as it might not yet be running. 10s might work, sometimes, but if no runners are available it might take longer. It does seem like we should wait a second and loop up to 60 seconds until it shows up.

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          gh-pages-branch: benchmark-data
+          benchmark-data-dir-path: ${{ steps.download.outputs.data-dir }}
+          # Don't fail the workflow if there's an issue with benchmark storage
+          fail-on-alert: false
+          comment-on-alert: false
+
+      - name: Summary
+        if: always()
```
rkuris: nit: I don't think you need this

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+        run: just benchmark-wait "${{ steps.trigger.outputs.run_id }}"
+        env:
+          GH_TOKEN: ${{ secrets.FIREWOOD_AVALANCHEGO_GITHUB_TOKEN }}
+        timeout-minutes: 60
+
+      - name: Download benchmark results
+        id: download
+        shell: nix develop ./ffi --command bash {0}
+        run: |
+          just benchmark-download "${{ steps.trigger.outputs.run_id }}"
```
rkuris: Seems like it would be more readable if this command were the same as the name on line 132 -- "just download-benchmark-results"

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,196 @@
...
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: C-Chain Reexecution Performance
+          tool: 'go'
+          output-file-path: ./results/benchmark-output.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          gh-pages-branch: benchmark-data
+          benchmark-data-dir-path: ${{ steps.download.outputs.data-dir }}
+          # Don't fail the workflow if there's an issue with benchmark storage
```
rkuris: seems like we should know if this happens, otherwise we'll silently stop seeing results, right?

METRICS.md
```diff
@@ -284,3 +284,69 @@ rate(firewood_space_from_end[5m])
 rate(firewood_proposal_commit{success="false"}[5m]) /
 rate(firewood_proposal_commit[5m])
 ```
+
+## Performance Tracking
+
+Firewood tracks its performance over time by running C-Chain reexecution benchmarks in AvalancheGo. This allows us to:
```
rkuris: For the uninitiated, someone might not know what the "C-Chain reexecution benchmarks" are. Can you add a link to the yaml in avalanchego or give a brief explanation of what they are?

Also, this is a quick summary but it still should say generally how long it's expected to run and how many blocks will be processed by default.

METRICS.md
```diff
@@ -284,3 +284,69 @@ rate(firewood_space_from_end[5m])
...
+just benchmark-trigger \
+  firewood=v0.0.15 \
+  avalanchego=master \
+  task=c-chain-reexecution-firewood-101-250k \
+  runner=avalanche-avalanchego-runner-2ti
+
+# Wait for a specific run to complete
+just benchmark-wait run_id=19938272417
+
+# Download results from a specific run
```
rkuris: where do these get downloaded to?

METRICS.md
```diff
@@ -284,3 +284,69 @@ rate(firewood_space_from_end[5m])
...
+  libevm=v1.0.0
+```
+
+**Parameters:**
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `firewood` | Firewood commit/branch/tag | `HEAD` |
+| `avalanchego` | AvalancheGo commit/branch/tag | `master` |
+| `task` | Benchmark task name | `c-chain-reexecution-firewood-101-250k` |
```
rkuris: Where can I find a list of possible task names?

METRICS.md
```diff
@@ -284,3 +284,69 @@ rate(firewood_space_from_end[5m])
...
+
+### Running Benchmarks Locally
+
+Benchmarks can be triggered locally using just commands (requires nix):
+
+```bash
+# Run with defaults (current HEAD, master, default task/runner)
+just benchmark
+
+# With named parameters
```
rkuris: Readability here was strange. Would prefer the parameter descriptions first and then an examples section afterwards.

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+    
+    $GH workflow run "Firewood Reexecution Benchmark" \
+        --repo ava-labs/avalanchego \
+        --ref "{{ avalanchego }}" \
+        -f firewood="{{ firewood }}" \
+        -f task="{{ task }}" \
+        -f runner="{{ runner }}" \
+        $LIBEVM_FLAG
+    
+    sleep 10
```
rkuris: This sleep is non-deterministic as well, might fail later if the runners are very slow to start for some reason. See my earlier comment.

I'm also worried about too much cut-paste. Can we consolidate these two "start this thing" and wait for the job to start into a single location somewhere?

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+        exit 1
+    fi
+    
+    echo "$RUN_ID"
+
+# Wait for reexecution benchmark run to complete
+# Usage: just benchmark-wait <run_id>
+benchmark-wait run_id: check-nix
+    #!/usr/bin/env bash
+    set -euo pipefail
```
rkuris, line 165: These two lines can be replaced with:
```suggestion
    #!/usr/bin/env -S bash -euo pipefail
```

This works since bash 4.4 (released roughly 10 years ago).

Not commenting on other instances of this, probably should change them all.

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+        FIREWOOD=$(git rev-parse HEAD)
+    fi
+    
+    echo "Firewood: $FIREWOOD"
+    echo "AvalancheGo: {{ avalanchego }}"
+    echo "Task: {{ task }}"
+    echo "Runner: {{ runner }}"
+    if [ -n "{{ libevm }}" ]; then
+        echo "LibEVM: {{ libevm }}"
+    fi
```
rkuris:
```suggestion
    else
        echo "LibEVM: default"
    fi
```

Probably should say what the default is here too

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+    
+    echo "Triggering reexecution benchmark in AvalancheGo..."
```
rkuris: Maybe move this before the others and indent the parameters?

```suggestion
    echo "Triggering reexecution benchmark in AvalancheGo with parameters:"
    echo "  - Firewood: $FIREWOOD"
    # ...
```

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+    RUN_ID=$(just benchmark-trigger "$FIREWOOD" "{{ avalanchego }}" "{{ task }}" "{{ runner }}" "{{ libevm }}")
+    echo "  Run ID: $RUN_ID"
+    echo "  URL: https://github.com/ava-labs/avalanchego/actions/runs/$RUN_ID"
```
rkuris: Let's be more verbose here so someone knows what the URL does. Maybe something like:
```suggestion
    echo "  Monitor progress here: https://github.com/ava-labs/avalanchego/actions/runs/$RUN_ID"
```

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+    fi
+    
+    echo "$RUN_ID"
+
+# Wait for reexecution benchmark run to complete
+# Usage: just benchmark-wait <run_id>
+benchmark-wait run_id: check-nix
+    #!/usr/bin/env bash
+    set -euo pipefail
+    nix run ./ffi#gh -- run watch "{{ run_id }}" --repo ava-labs/avalanchego --exit-status
```
rkuris: This can hang if it never finishes, which is possible if someone introduces some infinite loop somewhere.

I think github will let it run for a day, which is way too long. 2h seems like plenty but perhaps a shorter time makes sense.

```suggestion
    nix run ./ffi#gh -- timeout -k 1m 2h run watch "{{ run_id }}" --repo ava-labs/avalanchego --exit-status
```

justfile
```diff
@@ -138,3 +138,99 @@ update-ffi-flake: check-nix
...
+        echo "LibEVM: {{ libevm }}"
+    fi
+    
+    echo "Triggering reexecution benchmark in AvalancheGo..."
+    RUN_ID=$(just benchmark-trigger "$FIREWOOD" "{{ avalanchego }}" "{{ task }}" "{{ runner }}" "{{ libevm }}")
+    echo "  Run ID: $RUN_ID"
+    echo "  URL: https://github.com/ava-labs/avalanchego/actions/runs/$RUN_ID"
+    echo ""
+    
+    echo "Waiting for benchmark completion..."
```
rkuris: 👍  This looks polished. I probably would have just thrown a "set -x" here and called it a day but I like this better.

.github/workflows/track-performance.yml
Elvis339: Currently checked out commit, I updated docs to reduce ambiguity. 

https://github.com/ava-labs/firewood/blob/19d921ee9a0d3b030f44087fed7aa5566f730d2c/.github/workflows/track-performance.yml#L7

.github/workflows/track-performance.yml
Elvis339: Without predefined tasks, you'd need to specify all the granular S3 parameters manually (block-dir-src, current-state-dir-src, etc.). These predefined tasks exist in AvalancheGo's Taskfile for convenience and reusability.

.github/workflows/track-performance.yml
Elvis339: Not needed you're right, removed.

.github/workflows/track-performance.yml
Elvis339: Yes exactly sometimes the runners are busy so I've increased this to 3 min.

https://github.com/ava-labs/firewood/blob/19d921ee9a0d3b030f44087fed7aa5566f730d2c/justfile#L163

RodrigoVillar: Could we add a comment here explaining why we need to `sleep` here?

.github/workflows/track-performance.yml
Elvis339: Yes, good catch added warning: https://github.com/ava-labs/firewood/blob/19d921ee9a0d3b030f44087fed7aa5566f730d2c/.github/workflows/track-performance.yml#L165

.github/workflows/track-performance.yml
Elvis339: Good insight, I updated related commands to be more readable and intuitive:
- benchmark-download → download-benchmark-results
- benchmark-trigger → trigger-benchmark
- benchmark-wait → wait-benchmark
- benchmark-list → list-benchmarks

METRICS.md
Elvis339: You're right, updated: https://github.com/ava-labs/firewood/blob/5e32e9da7e34cc56485d72d1bef0c45e614aaa01/METRICS.md#performance-tracking

METRICS.md
Elvis339: They get downloaded in `./results` https://github.com/ava-labs/firewood/blob/5e32e9da7e34cc56485d72d1bef0c45e614aaa01/justfile#L191

METRICS.md
Elvis339: Extended docs with instructions on finding task names: METRICS.md#running-benchmarks-locally

Once merged, you can trigger this workflow from the GitHub UI with preset task options in a dropdown no need to look them up.

justfile
Elvis339: Addressed. I replaced the fixed sleep 10 with a retry loop (up to 3min, 10s intervals) that polls for the run ID until it appears. The trigger + wait-for-registration logic is now consolidated in trigger-benchmark.

justfile
Elvis339: https://github.com/ava-labs/firewood/blob/80082648a0c7d86b2dda9ee3049553a178721d0e/justfile#L208

justfile
Elvis339: https://github.com/ava-labs/firewood/blob/80082648a0c7d86b2dda9ee3049553a178721d0e/justfile#L224

justfile
Elvis339: https://github.com/ava-labs/firewood/blob/80082648a0c7d86b2dda9ee3049553a178721d0e/justfile#L185

.github/workflows/track-performance.yml
Elvis339: Replaced `sleep` with bit more robust implementation with comment:

https://github.com/ava-labs/firewood/pull/1493/commits/f2cf5024631365abc2423bec5a486479d9589863

Elvis339:
We need to get this merged first: https://github.com/ava-labs/avalanchego/pull/4650 before merging this PR.

DO NOT MERGE.

.github/workflows/gh-pages.yaml
```diff
@@ -30,6 +30,14 @@ jobs:
         run: |
           cp -rv target/doc/* ./_site
           cp -rv docs/assets ./_site
+      - name: Include benchmark data
+        run: |
+          git fetch origin benchmark-data || true
```
rkuris: why are we ignoring errors here?

.github/workflows/gh-pages.yaml
```diff
@@ -30,6 +30,14 @@ jobs:
...
+          if git rev-parse origin/benchmark-data >/dev/null 2>&1; then
+            git checkout origin/benchmark-data -- dev bench 2>/dev/null || true
+            cp -rv dev _site/ 2>/dev/null || true
+            cp -rv bench _site/ 2>/dev/null || true
```
rkuris: same here. Seems like if these fail, the build should fail

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
+name: Track Performance
+
+on:
+  workflow_dispatch:
```
rkuris, line 6: Why is this manually dispatched? Shouldn't this happen automatically on pushes to main, or is this just an intermediate step for testing purposes?

If the latter, please change the PR description to show the followup or create another PR/task and link it.

.github/workflows/track-performance.yml
rkuris: Why did you pick these particular ones?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+        options:
+          - avalanche-avalanchego-runner-2ti
+          - avago-runner-i4i-4xlarge-local-ssd
+          - avago-runner-m6i-4xlarge-ebs-fast
+
+jobs:
+  c-chain-benchmark:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write  # Required for github-action-benchmark to push to gh-pages
```
rkuris, line 49: This comment was helpful as a good explanation of what this task does. IMO this would be better documented as a block comment or even better by changing the name from "c-chain-benchmark" to "record-bench-to-gh-pages".

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+    steps:
+      - name: Checkout Firewood
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Needed for github-action-benchmark
```
rkuris: why is this? Seems like this will be pretty slow.

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@02a151ada4993995686f9ed4f1be7cfbb229e56f # v31
```
rkuris: why do we need this specific version? How will this be maintained going forward?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+        run: |
+          FIREWOOD="${{ inputs.firewood || github.sha }}"
+          echo "firewood=$FIREWOOD" >> "$GITHUB_OUTPUT"
+          
+          if [ -n "${{ inputs.task }}" ]; then
+            # Task-based mode: use just command
+            RUN_ID=$(just trigger-benchmark "$FIREWOOD" "${{ inputs.avalanchego }}" "${{ inputs.task }}" "${{ inputs.runner }}" "${{ inputs.libevm }}")
+          else
+            # Granular mode: use direct gh with custom params
+            LIBEVM_FLAG=""
```
rkuris: This seems a little fragile too. What happens if someone changes the just file? How can they know this needs to be kept in sync?

Can't we just pass the parameters to just?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+            
+            if [ -z "$RUN_ID" ] || [ "$RUN_ID" = "null" ]; then
+              echo "Could not find triggered workflow run after 3min"
+              exit 1
+            fi
+          fi
+          
+          echo "run_id=$RUN_ID" >> "$GITHUB_OUTPUT"
+          echo "run_url=https://github.com/ava-labs/avalanchego/actions/runs/$RUN_ID" >> "$GITHUB_OUTPUT"
+        env:
```
rkuris, line 65: nit: move to top rather than hiding it here

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+              -f firewood="$FIREWOOD" \
+              -f config="${{ inputs.config }}" \
+              -f start-block="${{ inputs.start-block }}" \
+              -f end-block="${{ inputs.end-block }}" \
+              -f block-dir-src="${{ inputs.block-dir-src }}" \
+              -f current-state-dir-src="${{ inputs.current-state-dir-src }}" \
+              -f runner="${{ inputs.runner }}" \
+              $LIBEVM_FLAG
+            
+            # Wait for GitHub API to register the workflow run (retry up to 3min)
```
rkuris: Is this the same logic we use in just, or are we already out of sync with what just does?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+        run: just wait-benchmark "${{ steps.trigger.outputs.run_id }}"
+        env:
+          GH_TOKEN: ${{ secrets.FIREWOOD_AVALANCHEGO_GITHUB_TOKEN }}
+        timeout-minutes: 60
+
+      - name: Download benchmark results
+        id: download
+        shell: nix develop ./ffi --command bash {0}
+        run: |
+          just download-benchmark-results "${{ steps.trigger.outputs.run_id }}"
```
rkuris: curious as to what happens if the benchmark fails to start at all

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+          auto-push: true
+          gh-pages-branch: benchmark-data
+          benchmark-data-dir-path: ${{ steps.download.outputs.data-dir }}
+          fail-on-alert: false
+          comment-on-alert: false
+
+      - name: Summary
+        run: |
+          if [ "${{ steps.store.outcome }}" == "failure" ]; then
+            echo "::warning::Benchmark storage failed - results were not saved to GitHub Pages"
```
rkuris, line 112: looks like we keep going anyway. Is this expected?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,205 @@
...
+          fi
+          echo "## Firewood Performance Benchmark Results" >> $GITHUB_STEP_SUMMARY
+          echo "" >> $GITHUB_STEP_SUMMARY
+          echo "**Configuration:**" >> $GITHUB_STEP_SUMMARY
+
```
rkuris: Group all these together (probably with the ones below too):
```suggestion
          { 
              echo "## Firewood Performance Benchmark Results"
              echo
              echo "**Configuration:**"
              # ...
        } > $GITHUB_STEP_SUMMARY
```

It's a little particular about spacing though. There must be a space after the opening { and a newline or a semicolon ; before the closing }.

justfile
```diff
@@ -138,3 +138,100 @@ update-ffi-flake: check-nix
...
+        --repo ava-labs/avalanchego \
+        --ref "{{ avalanchego }}" \
+        -f firewood="{{ firewood }}" \
+        -f task="{{ task }}" \
+        -f runner="{{ runner }}" \
+        $LIBEVM_FLAG
+    
+    # Wait for GitHub to register the run (retry up to 3min)
+    for i in {1..18}; do
+        sleep 10
```
rkuris: Can we consolidate this loop and reduce the possibility of these getting out of sync?

A great example might be lets assume we want to change this to check every 5 seconds. How many places have to be updated?

.github/workflows/track-performance.yml
Elvis339: It's representative of the preset defined here:
- https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/c-chain-reexecution-benchmark-container.json
- https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/c-chain-reexecution-benchmark-gh-native.json

.github/workflows/track-performance.yml
Elvis339: Yes, this is intentional. The `continue-on-error: true` on the store step lets us still show the summary even if publishing to GitHub Pages fails.

The idea is: if the benchmark ran and we got results, we want to surface what was tested and provide links to the AvalancheGo run - even if we couldn't save to gh-pages. The warning ::warning::Benchmark storage failed... makes the failure visible without hiding the useful context.
Alternative would be failing the workflow, but then we lose the summary which makes debugging harder.

justfile
```diff
@@ -168,3 +168,51 @@ release-step-refresh-changelog tag:
 
     echo "Generating changelog..."
     git cliff -o CHANGELOG.md --tag "{{tag}}"
+
+# Run a C-Chain reexecution benchmark
+# just bench <test>                                    - predefined test
+# FIREWOOD_REF=v0.1.0 just bench firewood-101-250k     - with specific ref
+# Custom: START_BLOCK=101 END_BLOCK=250000 BLOCK_DIR_SRC=<s3> CURRENT_STATE_DIR_SRC=<s3> just bench
+bench test="":
```
Elvis339: `just` provides the golden path for developers to run benchmarks locally. Same commands work in CI.

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,134 @@
...
+    permissions:
+      contents: write  # Required for github-action-benchmark to push to gh-pages
+    steps:
+      - name: Checkout Firewood
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Needed for github-action-benchmark
+
+      - name: Trigger C-Chain Reexecution Benchmark
+        run: ./scripts/bench-cchain-reexecution.sh trigger "${{ inputs.test }}"
```
Elvis339: Not using `just` here it's a thin wrapper around this script. Avoids installing Nix/just in CI for no benefit.

.github/workflows/track-performance.yml
Elvis339: It's intermediate step for testing purposes. I created new issue for follow-up: https://github.com/ava-labs/firewood/issues/1639 and update PR description.

.github/workflows/track-performance.yml
Elvis339: You were right it's not needed. Updated.

.github/workflows/gh-pages.yaml
```diff
@@ -30,6 +30,32 @@ jobs:
...
           cp -rv docs/assets ./_site
+      # GitHub Pages deploys from a single source (overwrites, doesn't merge).
+      # Benchmark history lives on benchmark-data branch (for append-only storage).
+      # We merge both into _site/ so a single deployment serves docs + benchmarks.
+      # See track-performance.yml for how benchmark data is collected and stored.
+      #
+      # Structure on benchmark-data branch (see track-performance.yml for how this is populated):
+      #   bench/              - Official benchmark history (main branch only)
+      #   dev/bench/{branch}/ - Feature branch benchmarks (experimental)
+      - name: Include benchmark data
```
Elvis339, line 41: Build working: https://github.com/ava-labs/firewood/actions/runs/21334822723/job/61405063779
Deploy only works from main due to environment protection rules.

.github/check-license-headers.yaml
```diff
@@ -58,7 +58,7 @@
       "clippy.toml",
       "**/tests/compile_*/**",
       "justfile",
-      "scripts/run-just.sh",
+      "scripts/**",
```
Elvis339, line 61: Added new benchmark script (bench-cchain-reexecution.sh) which 
caused CI to fail with "Config does not cover the file". Shell 
scripts aren't checked for license content (only .rs/.go/.h are), 
but must be explicitly listed in the config. Exclude entire 
scripts/ directory to avoid listing each script individually.

https://github.com/ava-labs/firewood/blob/main/.github/check-license-headers.yaml

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,137 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      firewood:
+        description: 'Firewood commit/branch/tag to test (leave empty to use the commit that triggered the workflow)'
+        default: ''
+      libevm:
+        description: 'libevm commit/branch/tag to test (leave empty to skip)'
+        default: ''
```
rkuris, line 13: Isn't libevm part of the monorepo now, or is this still pending?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,137 @@
...
+
+jobs:
+  record-benchmark-to-gh-pages:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write  # Required for github-action-benchmark to push to gh-pages
+    steps:
+      - name: Checkout Firewood
+        uses: actions/checkout@v4
```
rkuris, line 56: Doesn't this always check out the default branch? Shouldn't it be parameterized with something like:
```
ref: {{ inputs.firewood || github.sha }}
```

Otherwise I think this checks out main, and then the script later switches it to a different branch, which is less efficient and perhaps confusing.

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,137 @@
...
+          # Execution
+          RUNNER: ${{ inputs.runner }}
+          TIMEOUT_MINUTES: ${{ inputs.timeout-minutes }}
+
+      # Store main branch results in bench/, feature branches in dev/bench/{branch}/
+      # to keep official benchmark history clean
+      - name: Determine results location
+        id: location
+        run: |
+          if [[ "${{ github.ref }}" == "refs/heads/main" ]]; then
```
rkuris, line 88: Is this correct? I'm not following if `github.ref` is the tag we really want here, especially if `inputs.firewood` points to something completely different?

Shouldn't this be doing a `git rev-parse --abbrev-ref HEAD` or something to get the correct firewood tag?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,137 @@
...
+          if [[ -n "${{ inputs.test }}" ]]; then
+            ./scripts/bench-cchain-reexecution.sh trigger "${{ inputs.test }}"
+          else
+            ./scripts/bench-cchain-reexecution.sh trigger
+          fi
+        env:
+          GH_TOKEN: ${{ secrets.FIREWOOD_AVALANCHEGO_GITHUB_TOKEN }}
+          # Custom mode (ignored when test is specified)
+          CONFIG: ${{ inputs.config }}
+          START_BLOCK: ${{ inputs.start-block }}
```
rkuris, line 69: It seems like if you set START_BLOCK you better also be setting CURRENT_STATE_DIR_SRC to let it know where to get the bootstrap database, is that correct?

If so, we should verify that either neither is provided or both are.

.github/workflows/track-performance.yml
RodrigoVillar: `libevm` will never be a part of the monorepo as it's a direct fork of `go-ethereum` (makes syncs easier) and for licensing reasons.

Elvis339:
> Early review, could use another pass
> 
> There's a lot of code here and I barely was able to complete the review in my maximum review time. Please consider breaking this up for more timely reviews, especially if the review is anything larger than this.
> 
> I'm also a little confused about how we track which firewood and avalanchego versions we ran the test against.
> 
> Let's say the performance loss was due to a change in avalanchego, how can we know that?

I've split into two PRs per your feedback:
- This PR (CI): Workflows, benchmark script, license config
- [#1642 (Local tooling):](https://github.com/ava-labs/firewood/pull/1642) justfile recipe, METRICS.md docs, flake.nix

On version tracking `github-action-benchmark` stores time-series data (mgas/s over time) without version metadata tracing a regression requires digging through run history.

This is a conscious tradeoff to get data flowing now with minimal setup, then iterate to something more robust (S3 storage, richer metadata, export for analysis). Planning to revisit after ~2 weeks of data collection.

For now, each run logs Firewood / Avalanchego refs in the GitHub Actions summary, so the info exists just not queryable. Added your concern to the tracking doc as something to think through for the next iteration. How does that sound? 

Summary example: https://github.com/ava-labs/firewood/actions/runs/21334552166

.github/workflows/track-performance.yml
Elvis339: The checkout here is only to get the `bench-cchain-reexecution.sh` script we're not building or testing Firewood in this workflow. The script just triggers AvalancheGo's workflow via API and passes `FIREWOOD_REF` to it. AvalancheGo is responsible for checking out and building Firewood at that ref.

So the checkout ref doesn't affect which Firewood version gets benchmarked that's controlled by `FIREWOOD_REF` (inputs.firewood) passed to AvalancheGo. The script itself is stable and backward compatible.

scripts/bench-cchain-reexecution.sh
RodrigoVillar, line 1: Could we rename this to `benchmark_cchain_range.sh`? This is a variation of the reexecution script found in AvalancheGo and ideally, I shouldn't have to remember two different script names.

.github/workflows/track-performance.yml
Elvis339: `github.ref` here is intentional it determines who can write to official benchmark history, not what is being benchmarked.

The logic is:
- Only workflows triggered from main write to bench/ (official history)
- Feature branches write to dev/bench/{branch}/ (experimental, won't pollute official trends)

This prevents a feature branch from accidentally (or intentionally) writing to official history by setting inputs.firewood=main.

`inputs.firewood` controls what gets benchmarked; `github.ref` controls where results are stored.

.github/workflows/track-performance.yml
Elvis339: `CURRENT_STATE_DIR` is optional, not required because you might want to start from Genesis. 
https://github.com/ava-labs/firewood/blob/cb0f6bed0b97cc1ccd9f8ac95fad43ce7ac3482e/scripts/bench-cchain-reexecution.sh#L31

Validation: https://github.com/ava-labs/firewood/blob/cb0f6bed0b97cc1ccd9f8ac95fad43ce7ac3482e/scripts/bench-cchain-reexecution.sh#L219

scripts/bench-cchain-reexecution.sh
Elvis339: The script name matters less here since the primary interface is the `just` wrapper for local iterations:

You'll actually use `just bench-cchain firewood-101-250k`

Not: `./scripts/bench-cchain-reexecution.sh trigger firewood-101-250k`

See [justfile L171-L261](https://github.com/ava-labs/firewood/blob/12e170be6390e94fc9b90774a5de9912e9a826d8/justfile#L171) in #1642.

The script is an implementation detail developers interact through `just`. 

Also, the interfaces are different:

| | AvalancheGo | Firewood |
|---|---|---|
| **Script** | `benchmark_cchain_range.sh` | `bench-cchain-reexecution.sh` |
| **Runs** | Locally on machine | Triggers remote workflow via API |
| **Interface** | `./script.sh <test>` | `./script.sh trigger <test>` |
| **Extra commands** | None | `status`, `list`, `tests` |

Given the different behavior and that developers won't use the script directly because some dependencies such as gh are hidden in Nix, I'd keep the distinct name to avoid confusion if someone does look at the script.

This context got lost with the PR split
https://github.com/ava-labs/firewood/pull/1642

scripts/bench-cchain-reexecution.sh
```diff
@@ -0,0 +1,358 @@
...
+    log "firewood:    ${firewood:-<AvalancheGo go.mod default>}"
+    log "runner:      $RUNNER"
+    
+    # Record time BEFORE triggering to avoid race condition: we only look for
+    # runs created after this timestamp, so concurrent triggers don't collide
+    local trigger_time
+    trigger_time=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+    
+    # TODO: use $AVALANCHEGO_REF once ava-labs/avalanchego#4560 is merged
+    gh workflow run "$WORKFLOW_NAME" \
```
Elvis339, line 240: TODO

Elvis339: https://github.com/ava-labs/firewood/pull/1493/commits/5ef931320d1d4cf959ad24b68832b0ef93dcf47c

Elvis339:
Executing:

```
RUNNER=avalanche-avalanchego-runner-2ti \
FIREWOOD_REF=v0.0.18 \
CONFIG=firewood \
START_BLOCK=1 \
END_BLOCK=100 \
BLOCK_DIR_SRC=cchain-mainnet-blocks-200-ldb \
./scripts/bench-cchain-reexecution.sh trigger
```

Triggered reexecution benchmark in AvalancheGo https://github.com/ava-labs/avalanchego/actions/runs/21408080471

scripts/bench-cchain-reexecution.sh
```diff
@@ -0,0 +1,357 @@
...
+    cat <<EOF
+USAGE
+  ./bench-cchain-reexecution.sh <command> [args]
+
+COMMANDS
+  trigger [test]   Trigger benchmark, wait, download results
+  status <run_id>  Check run status
+  list             List recent runs
+  tests            Show available tests
+  help             Show this help
```
RodrigoVillar, line 333:
```suggestion
  help             Show this help message
```

scripts/bench-cchain-reexecution.sh
```diff
@@ -0,0 +1,357 @@
...
+#
+# USAGE
+#   ./bench-cchain-reexecution.sh <command> [args]
+#
+# COMMANDS
+#   trigger [test]       Trigger benchmark, wait, download results
+#   status <run_id>      Check run status
+#   list                 List recent runs
+#   tests                Show available tests
+#   help                 Show this help
```
RodrigoVillar: This is a repetition of the help message, could we just tell the client to call `help` if they want to find out more information?

scripts/bench-cchain-reexecution.sh
```diff
@@ -0,0 +1,357 @@
...
+# Use GitHub Actions error annotation format in CI for better visibility
+err() {
+    if [[ "${GITHUB_ACTIONS:-}" == "true" ]]; then
+        echo "::error::$1"
+    else
+        echo "error: $1" >&2
+    fi
+}
+
+die() { err "$1"; exit 1; }
```
RodrigoVillar: Could we combine these two into one (i.e. just append the exit code to `err` and remove `die`)?

scripts/bench-cchain-reexecution.sh
```diff
@@ -0,0 +1,357 @@
...
+#
+# COMMANDS
+#   trigger [test]       Trigger benchmark, wait, download results
+#   status <run_id>      Check run status
+#   list                 List recent runs
+#   tests                Show available tests
+#   help                 Show this help
+#
+# ENVIRONMENT
+#   GH_TOKEN              (string, required)                          GitHub token for API access
```
RodrigoVillar: All values in bash are strings and so adding the type annotation here adds cognitive load when the name of the parameter itself is sufficient (e.g. `GH_TOKEN` tells me I need to input the token which is inherently a string, `TIMEOUT_MINUTES` tells me I need to input a numerical value). 

Could the environment variables instead be formatted as something along the lines of:
```
TIMEOUT_MINUTES       Workflow timeout in minutes (optional)  
```

scripts/bench-cchain-reexecution.sh
```diff
@@ -0,0 +1,357 @@
...
+#   TEST                  (string, optional)                          Predefined test name (alternative to arg)
+#   FIREWOOD_REF          (string, optional)                          Firewood commit/tag/branch (empty = AvalancheGo's go.mod default)
+#   AVALANCHEGO_REF       (string, master)                            AvalancheGo ref to test against
+#   RUNNER                (string, avalanche-avalanchego-runner-2ti)  GitHub Actions runner label
+#   LIBEVM_REF            (string, optional)                          Optional libevm ref
+#   TIMEOUT_MINUTES       (int, optional)                             Workflow timeout
+#   DOWNLOAD_DIR          (string, ./results)                         Directory for downloaded artifacts
```
RodrigoVillar: It's not clear that `./results` is a default value, could we instead say:
```
TIMEOUT_MINUTES       Workflow timeout in minutes (optional, default: ./results)  
```

.github/workflows/track-performance.yml
rkuris: Let's add some comments explaining this to avoid confusion later. In particular we should mention that the only reason to check this out is to get the script.

One interesting side effect of this decision is that if you want to update the script, it won't use the new one until it lands in main I think.

.github/workflows/track-performance.yml
rkuris: If you want to start from genesis, then you either must set START_BLOCK to 0 (1?) or not set it. So, if START_BLOCK is not 0, then CURRENT_STATE_DIR_SRC must be set. Isn't that correct?

.github/workflows/track-performance.yml
rkuris: A comment here would be helpful as it's not obvious.

.github/workflows/track-performance.yml
Elvis339: Added clarifying comment to void confusion later. 

> One interesting side effect of this decision is that if you want to update the script, it won't use the new one until it lands in main I think.

Since the checkout doesn't specify a ref, it uses whichever branch you trigger from for manual runs. So you can test script changes by triggering the workflow from your feature branch. For scheduled runs (coming later), it will use main - which is the expected behavior for automated tracking.

.github/workflows/track-performance.yml
Elvis339: There was a comment: https://github.com/ava-labs/firewood/blob/79a2def5d9fb61ac6cf0759756247d337ee17bdb/.github/workflows/track-performance.yml#L77, but I updated it to be more clarifying.

.github/workflows/track-performance.yml
Elvis339: Close, `START_BLOCK` should be `1` (not `0`) for `Genesis`, and when `CURRENT_STATE_DIR_SRC` is empty it means starting from genesis (no pre-existing state to bootstrap from). So the valid combinations are:

- Genesis run: `START_BLOCK=1`, `CURRENT_STATE_DIR_SRC` empty
- Resume run: `START_BLOCK=N`, `CURRENT_STATE_DIR_SRC` points to state at block N-1

scripts/bench-cchain-reexecution.sh
Elvis339: It's intentional the header is for reading the source directly, help is for runtime. I often just open the file and read the `ENVIRONMENT` without calling `help`.

scripts/bench-cchain-reexecution.sh
Elvis339: Keeping them separate as `err` logs without exiting (useful if we need to log multiple errors before bailing), `die` is the "fatal" version. Clearer intent at call sites.

scripts/bench-cchain-reexecution.sh
Elvis339: Good call, simplified the format dropped type annotations and moved required/optional to the end of the description.

https://github.com/ava-labs/firewood/pull/1493/commits/809c8ea1824396e0193987712cbd82d2e1153c0f

scripts/bench-cchain-reexecution.sh
Elvis339: Addressed: https://github.com/ava-labs/firewood/pull/1493/commits/809c8ea1824396e0193987712cbd82d2e1153c0f

scripts/bench-cchain-reexecution.sh
RodrigoVillar: The only place we call `err` in `bench_cchain_reexecution.sh` is when we pass in an unknown command and even then, we exit afterwards with code `1`.

> Clearer intent at call sites.

I think replacing uses of `die` with `err` and `exit 1` would be clearer in this case.

Elvis339: 👍 Done

https://github.com/ava-labs/firewood/pull/1493/commits/d5027de19810843719ec49cf6d405edf8acfefc2

Elvis339:
> Approving on the condition that [#1493 (comment)](https://github.com/ava-labs/firewood/pull/1493#discussion_r2733730171) be addressed

Thank you, it was addressed:

- https://github.com/ava-labs/firewood/pull/1493/changes/809c8ea1824396e0193987712cbd82d2e1153c0f
- https://github.com/ava-labs/firewood/blob/d5027de19810843719ec49cf6d405edf8acfefc2/scripts/bench-cchain-reexecution.sh#L14

RodrigoVillar:
@Elvis339 I meant this:

https://github.com/ava-labs/firewood/blob/06c4d0a373adcf3957761acac33b6028be5cdcab/scripts/bench-cchain-reexecution.sh#L333

Issue #1485, opened 2025-11-24
Clean up `RootStore` creation


RodrigoVillar:
This match degenerated into a map and could be written much simpler.

_Originally posted by @rkuris in https://github.com/ava-labs/firewood/pull/1484#discussion_r2553370239_

PR #1495, opened 2025-12-01
refactor(rootstore): own nodestore logic

RodrigoVillar:
As mentioned in #1470, there exists the following circular dependency: 

```mermaid
stateDiagram-v2
    s1: Revision Manager
    s2: RootStore

```

This PR eliminates the circular dependency above by giving `RootStore` a reference to the file backend, which allows it to construct revisions without having to rely on the storage of another `NodeStore` (previously, we needed the latest `NodeStore` whose storage we would clone).

This PR does the following:
- `NodeStore` construction with a specified root hash + address now takes in a storage reference rather than another `NodeStore.
- Revision construction is now handled by `RootStore` instead of the RevisionManager
- Moves revision caching from the RevisionManager to `RootStore`
- Adds tests for `RootStore` caching logic

Note: there exists an earlier version of this PR (#1480). However, https://github.com/ava-labs/firewood/pull/1480#discussion_r2566078111 pointed out that having to pass in a `NodeStore` to `NodeStore::with_root()` seemed odd. After some thought, I agree with this point and found that passing in a storage reference instead removed the need to have to share revisions between the `RevisionManager` and `RootStore`. Hence, I went with creating a new PR rather than clobbering the existing one.

### DIFF
--- a/firewood/src/root_store.rs
+++ b/firewood/src/root_store.rs
@@ -98,8 +98,11 @@ impl RootStore {
         &self,
         hash: &TrieHash,
     ) -> Result<Option<CommittedRevision>, Box<dyn std::error::Error + Send + Sync>> {
+        // Obtain the lock to prevent multiple threads from caching the same result.
+        let mut revision_cache = self.revision_cache.lock();
+
         // 1. Check if the committed revision is cached.
-        if let Some(v) = self.revision_cache.lock().get(hash) {
+        if let Some(v) = revision_cache.get(hash) {
             return Ok(Some(v));
         }
 
@@ -121,9 +124,7 @@ impl RootStore {
         ));
 
         // Cache for future lookups.
-        self.revision_cache
-            .lock()
-            .insert(hash.clone(), nodestore.clone());
+        revision_cache.insert(hash.clone(), nodestore.clone());
 
         Ok(Some(nodestore))
     }


firewood/src/manager.rs
```diff
@@ -122,20 +120,20 @@ impl RevisionManager {
...
-        };
-
         let storage = Arc::new(fb);
         let nodestore = Arc::new(NodeStore::open(storage.clone())?);
+        let root_store = config
+            .root_store_dir
+            .map(|path| RootStore::new(path, storage.clone()))
+            .transpose()
+            .map_err(RevisionManagerError::RootStoreError)?;
+
```
RodrigoVillar, line 130: This code resolves #1485.

firewood/src/root_store.rs
```diff
@@ -57,24 +75,126 @@ impl RootStore {
...
+        }
+
+        // 2. If the committed revision is not cached, query the underlying
+        //    datastore for the revision's root address.
         let Some(v) = self.items.get(**hash)? else {
             return Ok(None);
         };
 
         let array: [u8; 8] = v.as_ref().try_into()?;
+        let addr = LinearAddress::new(u64::from_be_bytes(array))
```
rkuris, line 116: This code seems racy. There's nothing to prevent two threads from calling get() on the same hash and ending up here:

Thread 1 runs to this point and releases the revision_cache lock, not seeing the hash it's looking for
Thread 2 also runs to this point and releases the revision_cache lock, again not seeing the hash it's looking for

I think you'll need to hold the revision_cache lock while doing all this other work. I think the mutex will only be locked while querying the underlying store, which should be pretty fast, so I think we should keep the lock around.

RodrigoVillar: Good catch; updated so that the lock is held for the lifetime of `get()`: e9fd23f

rkuris (approved):

PR #1496, opened 2025-12-01
feat: Difference iterator for change proofs

bernard-avalabs:
Building on https://github.com/ava-labs/firewood/pull/839, this is an alternate implementation of a difference iterator that creates a sequence of `BatchOp` operations that represent the difference between two tries.

The iterator works by performing lock-step synchronized pre-order traversal of the two tries. It does this by transitioning between the differents states specified by `DiffIterationNodeState`. The pre-order traversal implementation has been modified to allow it to skip the children of a node from traversal. This allows sub-tries that have matching path, value, and hash to be skipped during the traversal as they must be identical.

The iteration can start at a specified `start_key`, which will skip the traversal of all keys smaller than `start_key` from both tries.

Tests have primarily been taken from https://github.com/ava-labs/firewood/pull/839 and https://github.com/ava-labs/firewood/tree/amin/change-proofs-docs. The tests show that sub-tries are skipped if there is a hash match, but additional tests that verify that all possible matches have been detected should be added in a future PR.

### DIFF
--- a/benchmark/bootstrap/aws-launch.sh
+++ b/benchmark/bootstrap/aws-launch.sh
@@ -5,7 +5,7 @@ set +e
 INSTANCE_TYPE=i4g.large
 FIREWOOD_BRANCH=""
 AVALANCHEGO_BRANCH=""
-LIBEVM_COMMIT=""
+LIBEVM_BRANCH=""
 NBLOCKS="1m"
 CONFIG="firewood"
 REGION="us-west-2"
@@ -60,7 +60,7 @@ show_usage() {
     echo "  --instance-type TYPE        EC2 instance type (default: i4g.large)"
     echo "  --firewood-branch BRANCH    Firewood git branch to checkout"
     echo "  --avalanchego-branch BRANCH AvalancheGo git branch to checkout"
-    echo "  --libevm-commit COMMIT      LibEVM git commit to checkout"
+    echo "  --libevm-branch BRANCH      LibEVM git branch to checkout"
     echo "  --nblocks BLOCKS            Number of blocks to download (default: 1m)"
     echo "  --config CONFIG             The VM reexecution config to use (default: firewood)"
     echo "  --region REGION             AWS region (default: us-west-2)"
@@ -110,8 +110,8 @@ while [[ $# -gt 0 ]]; do
             AVALANCHEGO_BRANCH="$2"
             shift 2
             ;;
-        --libevm-commit)
-            LIBEVM_COMMIT="$2"
+        --libevm-branch)
+            LIBEVM_BRANCH="$2"
             shift 2
             ;;
         --nblocks)
@@ -176,7 +176,7 @@ done
 if [ ${#TERMINATE_INSTANCES[@]} -gt 0 ]; then
     # Check if any incompatible options are present
     if [ -n "$FIREWOOD_BRANCH" ] || [ -n "$AVALANCHEGO_BRANCH" ] || \
-       [ -n "$LIBEVM_COMMIT" ] || [ "$INSTANCE_TYPE" != "i4g.large" ] || [ "$NBLOCKS" != "1m" ] || \
+       [ -n "$LIBEVM_BRANCH" ] || [ "$INSTANCE_TYPE" != "i4g.large" ] || [ "$NBLOCKS" != "1m" ] || \
        [ "$CONFIG" != "firewood" ] || [ "$DRY_RUN" = true ] || [ "$SPOT_INSTANCE" = true ] || \
        [ "$SHOW_INSTANCES" = true ] || [ "$TERMINATE_MINE" = true ]; then
         echo "Error: --terminate cannot be used with other options except --region"
@@ -194,7 +194,7 @@ fi
 if [ "$TERMINATE_MINE" = true ]; then
     # Check if any incompatible options are present
     if [ -n "$FIREWOOD_BRANCH" ] || [ -n "$AVALANCHEGO_BRANCH" ] || \
-       [ -n "$LIBEVM_COMMIT" ] || [ "$INSTANCE_TYPE" != "i4g.large" ] || [ "$NBLOCKS" != "1m" ] || \
+       [ -n "$LIBEVM_BRANCH" ] || [ "$INSTANCE_TYPE" != "i4g.large" ] || [ "$NBLOCKS" != "1m" ] || \
        [ "$CONFIG" != "firewood" ] || [ "$DRY_RUN" = true ] || [ "$SPOT_INSTANCE" = true ] || \
        [ "$SHOW_INSTANCES" = true ]; then
         echo "Error: --terminate-mine cannot be used with other options except --region"
@@ -224,7 +224,7 @@ fi
 if [ "$SHOW_INSTANCES" = true ]; then
     # Check if any incompatible options are present
     if [ -n "$FIREWOOD_BRANCH" ] || [ -n "$AVALANCHEGO_BRANCH" ] || \
-       [ -n "$LIBEVM_COMMIT" ] || [ "$INSTANCE_TYPE" != "i4g.large" ] || [ "$NBLOCKS" != "1m" ] || \
+       [ -n "$LIBEVM_BRANCH" ] || [ "$INSTANCE_TYPE" != "i4g.large" ] || [ "$NBLOCKS" != "1m" ] || \
        [ "$CONFIG" != "firewood" ] || [ "$DRY_RUN" = true ] || [ "$SPOT_INSTANCE" = true ]; then
         echo "Error: --show cannot be used with other options except --region"
         exit 1
@@ -246,7 +246,7 @@ echo "Configuration:"
 echo "  Instance Type: $INSTANCE_TYPE ($TYPE)"
 echo "  Firewood Branch: ${FIREWOOD_BRANCH:-default}"
 echo "  AvalancheGo Branch: ${AVALANCHEGO_BRANCH:-default}"
-echo "  LibEVM Commit: ${LIBEVM_COMMIT:-default}"
+echo "  LibEVM Branch: ${LIBEVM_BRANCH:-default}"
 echo "  Number of Blocks: $NBLOCKS"
 echo "  Config: $CONFIG"
 echo "  Region: $REGION"
@@ -288,8 +288,8 @@ if [ -n "$AVALANCHEGO_BRANCH" ]; then
     AVALANCHEGO_BRANCH_ARG="--branch $AVALANCHEGO_BRANCH"
 fi
 
-# For libevm, default to main branch if no commit specified
-LIBEVM_COMMIT_CHECKOUT="${LIBEVM_COMMIT:-main}"
+# For libevm, default to main branch if no branch specified
+LIBEVM_BRANCH_CHECKOUT="${LIBEVM_BRANCH:-main}"
 
 # set up this script to run at startup, installing a few packages, creating user accounts,
 # and downloading the blocks for the C-chain
@@ -411,7 +411,7 @@ runcmd:
     git clone https://github.com/ava-labs/libevm.git
   - >
     sudo -u ubuntu -D /mnt/nvme/ubuntu/libevm
-    git checkout __LIBEVM_COMMIT__
+    git checkout __LIBEVM_BRANCH__
   # force coreth to use the checked-out version of firewood and libevm
   - >
     sudo -u ubuntu -D /mnt/nvme/ubuntu/avalanchego/graft/coreth
@@ -471,7 +471,7 @@ esac
 USERDATA=$(echo "$USERDATA_TEMPLATE" | \
   sed "s|__FIREWOOD_BRANCH_ARG__|$FIREWOOD_BRANCH_ARG|g" | \
   sed "s|__AVALANCHEGO_BRANCH_ARG__|$AVALANCHEGO_BRANCH_ARG|g" | \
-  sed "s|__LIBEVM_COMMIT__|$LIBEVM_COMMIT_CHECKOUT|g" | \
+  sed "s|__LIBEVM_BRANCH__|$LIBEVM_BRANCH_CHECKOUT|g" | \
   sed "s|__NBLOCKS__|$NBLOCKS|g" | \
   sed "s|__END_BLOCK__|$END_BLOCK|g" | \
   sed "s|__CONFIG__|$CONFIG|g" | \
@@ -481,7 +481,8 @@ export USERDATA
 fi  # End of DRY_RUN=false conditional
 
 
-SUFFIX=$(hexdump -vn4 -e'4/4 "%08X" 1 "\n"' /dev/urandom)
+# Generate a compact 8-hex-digit suffix without spaces/newlines
+SUFFIX=$(hexdump -vn4 -e '1/4 "%08X"' /dev/urandom)
 
 # Build instance name with branch info
 INSTANCE_NAME="$USER-fw-$SUFFIX"
@@ -491,8 +492,8 @@ fi
 if [ -n "$AVALANCHEGO_BRANCH" ]; then
     INSTANCE_NAME="$INSTANCE_NAME-ag-$AVALANCHEGO_BRANCH"
 fi
-if [ -n "$LIBEVM_COMMIT" ]; then
-    INSTANCE_NAME="$INSTANCE_NAME-le-$LIBEVM_COMMIT"
+if [ -n "$LIBEVM_BRANCH" ]; then
+    INSTANCE_NAME="$INSTANCE_NAME-le-$LIBEVM_BRANCH"
 fi
 
 # Build spot instance market options if requested


firewood/Cargo.toml
```diff
@@ -36,6 +36,7 @@ thiserror.workspace = true
 # Regular dependencies
 typed-builder = "0.23.1"
 rayon = "1.11.0"
+triomphe = "0.1.15"
```
rkuris: Let's try to avoid using triomphe in firewood. The abstract type `SharedNode` exported by storage, is already a triomphe::Arc<Node>, and that should be the only use case of it here.

firewood/Cargo.toml
```diff
@@ -68,6 +69,7 @@ hash-db = "0.16.0"
 plain_hasher = "0.2.3"
 rlp = "0.6.1"
 sha3 = "0.10.8"
+metrics-util = "0.20.0"
```
rkuris: If possible, let's avoid bleeding metrics-util here. It looks like you're only using it to install a metrics listener. That code is duplicated from somewhere else and should be consolidated.

It's particularly challenging here because metrics can only be initialized once, so you may run into problems if tests share the same executable.

AI's plan looks like this:

Plan: Consolidate Metrics Testing Infrastructure

The PR added metrics-util to test diff iterator optimizations. Currently there's only one Rust test using metrics recording, but it will be skipped if any other code sets the global recorder first. The FFI crate already has production metrics setup, creating a good consolidation opportunity.

Steps
 - Create shared test recorder module in test_utils.rs since it already houses test utilities and is conditionally compiled with #[cfg(any(test, feature = "test_utils"))]. Move the TestRecorder implementation from diff.rs there, making it reusable across all test modules.

Add thread-local or test-scoped metrics capture by implementing a with_test_metrics helper that uses a DebuggingRecorder (from metrics-util) or similar approach to capture metrics without requiring global recorder installation. This avoids conflicts when tests run in the same process.

Update the diff test in diff.rs to use the new shared infrastructure instead of trying to set a global recorder, removing the silent test-skip behavior.

Document metrics testing constraints in [CONTRIBUTING.md] or [test_utils.rs], explaining that global recorders can only be set once and how to use the shared testing utilities properly.

firewood/src/iter.rs
```diff
@@ -544,7 +544,7 @@ fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(nibbles: Iter) -> Key {
 }
 
 #[cfg(not(feature = "branch_factor_256"))]
-fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Key {
+pub(crate) fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Key {
```
rkuris, line 551: Can you verify this isn't needed during branch_factor_256? I think this breaks those builds, even though we currently don't do much testing for it. Easy fix here would be to change the code a few lines earlier too.

firewood/src/merkle/mod.rs
```diff
@@ -1091,14 +1091,18 @@ pub fn nibbles_to_bytes_iter(nibbles: &[u8]) -> impl Iterator<Item = u8> {
 /// The `unique_*` properties, [`unique_a`][`PrefixOverlap::unique_a`] and [`unique_b`][`PrefixOverlap::unique_b`]
 /// are set based on the argument order passed into the [`from`][`PrefixOverlap::from`] constructor.
 #[derive(Debug)]
-struct PrefixOverlap<'a, T> {
-    shared: &'a [T],
-    unique_a: &'a [T],
-    unique_b: &'a [T],
+pub struct PrefixOverlap<'a, T> {
```
rkuris: I think this bleeds outside the crate, which is not good. Consider moving PrefixOverlap to someplace that both consumers can see it, or better yet, adding more helper methods to get the parts you need. I'll comment about this later in the PR if I get that far.

firewood/src/merkle/mod.rs
```diff
@@ -1091,14 +1091,18 @@ pub fn nibbles_to_bytes_iter(nibbles: &[u8]) -> impl Iterator<Item = u8> {
...
+    pub shared: &'a [T],
+    pub unique_a: &'a [T],
+    pub unique_b: &'a [T],
 }
 
 impl<'a, T: PartialEq> PrefixOverlap<'a, T> {
-    fn from(a: &'a [T], b: &'a [T]) -> Self {
+    /// # Panics
```
rkuris: Again, this isn't great, because now the visibility of these internal routines is getting bigger.

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+// TODO: Currently only tests use the code in this file. Remove allow(dead_code) once change
+//       proof has been implemented.
+#![allow(dead_code)]
```
rkuris: This is likely to remain after we start using it, since it's not a `expect(dead_code)`. Also, it's better to just put this on the top level function it's complaining about, rather than the whole file, in case there's something stray you're not using.

Moving this to `#[expect(dead_code)]` to `DiffMerkleNodeStream` probably solves both of these.

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
...
+
+use crate::{
+    db::BatchOp,
+    iter::key_from_nibble_iter,
+    merkle::{Key, PrefixOverlap, Value},
+};
+use firewood_storage::{Child, FileIoError, HashedNodeReader, Node, Path, TrieHash, TrieReader};
+use metrics::counter;
+use std::{cmp::Ordering, iter::once};
+use triomphe::Arc;
```
rkuris: See earlier comment, this should be `SharedNode`

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
...
+        );
+        Ok(NodeState {
+            path: full_path,
+            node,
+            hash: node_hash,
+        })
+    }
+
+    fn next(&mut self) -> Result<Option<NodeState>, FileIoError> {
+        counter!("firewood.diff_merkle", "traversal" => "next").increment(1);
```
rkuris: This doesn't properly register the metric. Use firewood_counter! here instead.

Also, let's change this to "firewood.change_proof.diff_traversal" or, even better, "firewood.change_proof.next"

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+struct PreOrderIterator<'a> {
+    trie: &'a dyn TrieReader,
+    stack: Vec<TraversalStackFrame>,
+    prev_num_children: u8,
+}
+
+/// Ignoring possible arithmetic side effects since we are only incrementing
+/// `prev_num_children` once per child and there can only be at most 16 children per
+/// node and `prev_num_children` is an u8.
+#[allow(clippy::arithmetic_side_effects)]
```
rkuris: Reduce the scope here and use `expect` to avoid this from going stale. I want to verify each addition doesn't cause any overflow problems at each point the warning occurs.

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
...
+            Child::Node(node) => node.clone().into(),
+            Child::AddressWithHash(addr, hash) => {
+                node_hash = Some(hash.into());
+                self.trie.read_node(addr)?
+            }
+            Child::MaybePersisted(maybe_persisted, hash) => {
+                node_hash = Some(hash.into());
+                maybe_persisted.as_shared_node(&self.trie)?
+            }
+        };
```
rkuris: This code duplication can be avoided using the `into_triehash` method, which works correctly in both ethhash and non-ethhash modes.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+            }
+            Child::MaybePersisted(maybe_persisted, hash) => {
+                node_hash = Some(hash.into());
+                maybe_persisted.as_shared_node(&self.trie)?
+            }
+        };
+
+        // Compute the full path for the node. This is used to compare with the full path
+        // of the current node from the other trie.
+        let full_path = Path::from_nibbles_iterator(
```
rkuris: I'm not sure if we can do better here. Is there a way to avoid all this path copying? I know there are some helpers that allow paths to be temporarily enlarged and then shrunk back down.

This comment is probably not very clear but we can chat about it in-person if you'd like.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+            path: full_path,
+            node,
+            hash: node_hash,
+        })
+    }
+
+    fn next(&mut self) -> Result<Option<NodeState>, FileIoError> {
+        counter!("firewood.diff_merkle", "traversal" => "next").increment(1);
+        // Pop the next node state from the stack and reset `prev_num_children`
+        self.prev_num_children = 0;
```
rkuris: The name `prev_num_children` suggests it refers to children from a previous iteration, but it actually tracks the children of the last processed node (which is the "current" node from the caller's perspective). Maybe `pushed_children_count` or `children_to_skip` or something like that might be clearer, or I'm not following this code well enough.

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
...
+/// is big enough since there can at most be 16 children per node.
+///
+/// Note that this skipping the children functionality can only be called once after a
+/// call to `next`. We cannot call `skip_children` twice in order to skip all of the
+/// children from the current node's parent. This functionality is not necessary for
+/// our use case as a hash match on the parent would have been determined previously
+/// since we are performing in-order traversal.
+struct PreOrderIterator<'a> {
+    trie: &'a dyn TrieReader,
+    stack: Vec<TraversalStackFrame>,
```
rkuris:
```suggestion
    traversal_stack: Vec<TraversalStackFrame>,
```
This might be less typing, but it's easier to grok the code if it's spelled out.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+            // Stack is empty, iteration is complete. Return None.
+            Ok(None)
+        }
+    }
+
+    /// Used to not iterate through the children of a branch node if there is a hash match.
+    fn skip_children(&mut self) {
+        for _ in 0..self.prev_num_children {
+            self.stack.pop();
+        }
```
rkuris: Use `Vec::truncate` here which will be faster:

```suggestion
        let truncated_length = self.stack.len().saturating_sub(self.prev_num_children);
        self.stack.truncate(truncated_length);
```

This also calls out that you should check that there are N elements in the stack before doing this, perhaps as a `debug_assert`.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+///
+/// Note that this skipping the children functionality can only be called once after a
+/// call to `next`. We cannot call `skip_children` twice in order to skip all of the
+/// children from the current node's parent. This functionality is not necessary for
+/// our use case as a hash match on the parent would have been determined previously
+/// since we are performing in-order traversal.
+struct PreOrderIterator<'a> {
+    trie: &'a dyn TrieReader,
+    stack: Vec<TraversalStackFrame>,
+    prev_num_children: u8,
```
rkuris: nit: This doesn't strictly need to be a `u8`, and might be easier as a `usize` but I'll leave that decision to you.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+                        );
+                        self.stack.push(TraversalStackFrame {
+                            pre_path: child_pre_path,
+                            node: child,
+                            hash: None,
+                        });
+                        // Increment the number of children that has been added to the stack.
+                        self.prev_num_children += 1;
+                    }
+                }
```
rkuris: Consider using `filter_map` to eliminate the `None` values for you, something like this (AI generated, so needs further review):

```suggestion
for (path_comp, child) in branch.children.clone().into_iter().rev().filter_map(|(pc, c)| c.map(|child| (pc, child))) {
    // Generate the pre-path for this child, and push it onto the traversal stack.
    let child_pre_path = Path::from_nibbles_iterator(
        state.path.iter().copied().chain(once(path_comp.as_u8())),
    );
    self.stack.push(TraversalStackFrame {
        pre_path: child_pre_path,
        node: child,
        hash: None,
    });
    // Increment the number of children that has been added to the stack.
    self.prev_num_children += 1;
}
```

rkuris (changes requested):

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+/// Adding support for the Iterator trait
+impl Iterator for DiffMerkleNodeStream<'_> {
+    type Item = Result<BatchOp<Key, Value>, firewood_storage::FileIoError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.next_internal().transpose()? {
+            Ok(batch) => Some(Ok(batch)),
+            Err(e) => Some(Err(e)),
+        }
+    }
```
AminR443, line 391: This can be simplified: 
```suggestion
    fn next(&mut self) -> Option<Self::Item> {
        self.next_internal().transpose()
    }
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+        left_hash: Option<TrieHash>,
+        left_tree: PreOrderIterator<'a>,
+        right_hash: Option<TrieHash>,
+        right_tree: PreOrderIterator<'a>,
+    ) -> DiffIterationNodeState<'a> {
+        if match (left_hash, right_hash) {
+            (Some(left_hash), Some(right_hash)) => left_hash == right_hash,
+            _ => false,
+        } {
+            DiffIterationNodeState::SkipChildren {
```
AminR443: We can simplify the expression here by using `matches!` macro:

```suggestion
    ) -> DiffIterationNodeState<'a> {
        let hashes_equal = matches!((left_hash, right_hash),
            (Some(l), Some(r)) if l == r
        );
        if hashes_equal {
            DiffIterationNodeState::SkipChildren {
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+                // this by returning it in the second entry of the tuple in the return value.
+                (
+                    DiffIterationNodeState::TraverseLeft {
+                        left_tree,
+                        right_tree,
+                        right_state,
+                    },
+                    left_state.node.value().map(|_val| BatchOp::Delete {
+                        key: key_from_nibble_iter(left_state.path.iter().copied()),
+                    }),
```
AminR443: can be replaced with `deleted_values` helper:

```suggestion
                    Self::deleted_values(left_state.node, left_state.path),
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+                (
+                    DiffIterationNodeState::TraverseRight {
+                        left_tree,
+                        right_tree,
+                        left_state,
+                    },
+                    right_state.node.value().map(|val| BatchOp::Put {
+                        key: key_from_nibble_iter(right_state.path.iter().copied()),
+                        value: val.into(),
+                    }),
```
AminR443: Similar to above:

```suggestion
                    Self::additional_values(right_state.node, right_state.path),
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,2028 @@
...
+                    None,
+                ),
+                (Some(_val), None) => (
+                    DiffIterationNodeState::TraverseBoth {
+                        left_tree,
+                        right_tree,
+                    },
+                    Some(BatchOp::Delete {
+                        key: key_from_nibble_iter(left_state.path.iter().copied()),
+                    }),
```
AminR443: It'd be nice if we could generalize the helpers for all of the cases, kind of repetitive patterns here.

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
...
+                        )
+                    }
+                }
+            },
+        }
+    }
+
+    /// Helper function that returns a key/value to be added to the delete list if the current
+    /// node from the left trie has a value.
+    fn deleted_values(left_node: Arc<Node>, left_path: Path) -> Option<BatchOp<Key, Value>> {
```
AminR443: why is this function taking parameters by value? Should be ok to take references instead:

```suggestion
    fn deleted_values(left_node: &Node, left_path: &Path) -> Option<BatchOp<Key, Value>> {
```

firewood/src/proofs/diff.rs
```diff
@@ -0,0 +1,2028 @@
...
+        left_node.value().map(|_val| BatchOp::Delete {
+            key: key_from_nibble_iter(left_path.iter().copied()),
+        })
+    }
+
+    /// Helper function that returns a key/value to be added to the addition list if the current
+    /// node from the right trie has a value.
+    fn additional_values(right_node: Arc<Node>, right_path: Path) -> Option<BatchOp<Key, Value>> {
```
AminR443: Similar to above:

```suggestion
    fn additional_values(right_node: &Node, right_path: &Path) -> Option<BatchOp<Key, Value>> {
```

firewood/src/proofs/diff.rs
AminR443: Also If we change the first parameter to take the node's value directly (passing .value() at the call site), this function becomes more reusable. See [here](https://github.com/ava-labs/firewood/pull/1496/files#r2597099891).

firewood/src/merkle/changes.rs
bernard-avalabs: In my opinion, the original is slightly easier to read than the version using `matches!`. I would prefer to keep the original unless there is some other reason that I'm missing.

Issue #1293, opened 2025-09-19
bug(io-uring): debug_assert triggered in


demosdemon:
CI action: https://github.com/ava-labs/firewood/actions/runs/17745669480/job/50430325118

```
--- FAIL: FuzzTree (41.12s)
    --- FAIL: FuzzTree (0.00s)
        merkle_compatibility_test.go:353: dbBatch
        merkle_compatibility_test.go:201: 
            	Error Trace:	/home/runner/work/firewood/firewood/ffi/tests/firewood/merkle_compatibility_test.go:201
            	            				/home/runner/work/firewood/firewood/ffi/tests/firewood/merkle_compatibility_test.go:360
            	            				/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.2.linux-amd64/src/reflect/value.go:584
            	            				/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.2.linux-amd64/src/reflect/value.go:368
            	            				/home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.2.linux-amd64/src/testing/fuzz.go:340
            	Error:      	Received unexpected error:
            	            	Found entry with node still set: Some(PinnedBufferEntry { pinned_buffer: [0, 9, 3, 3, 1, 4, 4, 155, 44, 109, 40], node: Some((LinearAddress(2048), MaybePersistedNode(ArcSwapAny(Allocated(LinearAddress(2048), Leaf([Leaf 3 3 1 4  9b2c6d28])))))) })
            	Test:       	FuzzTree
    
    Failing input written to testdata/fuzz/FuzzTree/8bf45eff27af0b01
    To re-run:
    go test -run=FuzzTree/8bf45eff27af0b01
FAIL

```

panic is here: https://github.com/ava-labs/firewood/blob/15e9a99fb5fc0648aaca523c4cdbf6c8de314294/storage/src/nodestore/persist.rs#L492-L496

(this is also evidence that the panic handler code works correctly when `panic=abort` is not set)

rkuris:
During backlog grooming we noted this hasn't happened and there have been several rewrites. Next backlog grooming we should consider closing it if we haven't seen it again

demosdemon:
https://github.com/ava-labs/firewood/actions/runs/19337032295/job/55315050786

AminR443:
https://github.com/ava-labs/firewood/actions/runs/20016225158/job/57394550147

PR #1498, opened 2025-12-02
fix: refactor io-uring to drive the queue to completion

demosdemon:
Previously, we were encountering issues with the io-uring writes where we discovered we were not completing writes.

After extensively studying `liburing` and the `io-uring` crate, I realized our submission/processing loop was unnecessarily complicated and did not continuously drive the queue until completion. The crux of the fix is the `backlog` dequeue that keeps track of the backlogged entries and the `outstanding` hashmap that keeps track of the entries we've submitted to the kernel. CQEs that don't indicate a complete write are re-queued for submission.

Closes: #1293 
Closes: #1488

## References

- <https://github.com/tokio-rs/io-uring/blob/f297f5bdb4802915b48a532edf9b817d6c91de8f/examples/tcp_echo.rs>
- <https://github.com/axboe/liburing/blob/e1ef1e680ee38ed9116989155fca47921698c25f/examples/io_uring-cp.c>

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -385,7 +385,8 @@ mod test {
     use std::path::PathBuf;
 
     use firewood_storage::{
-        CheckOpt, CheckerError, HashedNodeReader, IntoHashType, NodeStore, TrieHash,
+        CheckOpt, CheckerError, HashedNodeReader, IntoHashType, LinearAddress, MaybePersistedNode,
+        NodeStore, TrieHash,
     };
 
     use crate::db::{Db, Proposal, UseParallel};
@@ -796,6 +797,64 @@ mod test {
             }
             proposal.commit().unwrap();
         }
+    }
+
+    #[test]
+    fn test_propose_parallel_vs_normal_propose() {
+        fn persisted_deleted(nodes: &[MaybePersistedNode]) -> Vec<LinearAddress> {
+            let mut addresses: Vec<_> = nodes
+                .iter()
+                .filter_map(MaybePersistedNode::as_linear_address)
+                .collect();
+            addresses.sort();
+            addresses
+        }
+
+        let parallel_db = TestDb::new_with_config(
+            DbConfig::builder()
+                .use_parallel(UseParallel::Always)
+                .build(),
+        );
+        let single_threaded_db =
+            TestDb::new_with_config(DbConfig::builder().use_parallel(UseParallel::Never).build());
+
+        // First batch: insert two keys with different first nibbles so they are
+        // handled by different workers in the parallel merkle implementation.
+        let initial_keys: Vec<[u8; 1]> = vec![[0x00], [0x10]];
+        let initial_values: Vec<Box<[u8]>> = vec![Box::new([1u8]), Box::new([2u8])];
+
+        let initial_parallel_proposal = parallel_db
+            .propose(initial_keys.iter().zip(initial_values.iter()))
+            .unwrap();
+        let initial_single_proposal = single_threaded_db
+            .propose(initial_keys.iter().zip(initial_values.iter()))
+            .unwrap();
+
+        initial_parallel_proposal.commit().unwrap();
+        initial_single_proposal.commit().unwrap();
+
+        // Second batch: update only the first key.
+        let update_keys: Vec<[u8; 1]> = vec![[0x00]];
+        let update_values: Vec<Box<[u8]>> = vec![Box::new([3u8])];
+
+        let update_parallel_proposal = parallel_db
+            .propose(update_keys.iter().zip(update_values.iter()))
+            .unwrap();
+        let update_single_proposal = single_threaded_db
+            .propose(update_keys.iter().zip(update_values.iter()))
+            .unwrap();
+
+        let parallel_deleted = update_parallel_proposal.nodestore.deleted().to_vec();
+        let single_deleted = update_single_proposal.nodestore.deleted().to_vec();
+
+        assert_eq!(
+            persisted_deleted(&parallel_deleted),
+            persisted_deleted(&single_deleted),
+            "persisted deleted nodes should match between parallel and single proposals",
+        );
+
+        update_parallel_proposal.commit().unwrap();
+        update_single_proposal.commit().unwrap();
     }
 
     /// Test that proposing on a proposal works as expected
--- a/firewood/src/merkle/parallel.rs
+++ b/firewood/src/merkle/parallel.rs
@@ -7,12 +7,11 @@ use crate::v2::api::IntoBatchIter;
 use firewood_storage::logger::error;
 use firewood_storage::{
     BranchNode, Child, Children, FileBacked, FileIoError, ImmutableProposal, LeafNode,
-    MaybePersistedNode, MutableProposal, NibblesIterator, Node, NodeReader, NodeStore, Parentable,
-    Path, PathComponent,
+    MaybePersistedNode, MutableProposal, NibblesIterator, Node, NodeStore, Parentable, Path,
+    PathComponent,
 };
 use rayon::ThreadPool;
 use std::iter::once;
-use std::ops::Deref;
 use std::sync::mpsc::{Receiver, SendError, Sender};
 use std::sync::{Arc, mpsc};
 
@@ -227,7 +226,7 @@ impl ParallelMerkle {
     /// by the value of the `first_path_component`.
     fn create_worker(
         pool: &ThreadPool,
-        proposal: &NodeStore<MutableProposal, FileBacked>,
+        proposal: &mut NodeStore<MutableProposal, FileBacked>,
         root_branch: &mut BranchNode,
         first_path_component: PathComponent,
         response_sender: Sender<Result<Response, FileIoError>>,
@@ -235,7 +234,6 @@ impl ParallelMerkle {
         // Create a channel for the coordinator (main thread) to send messages to this worker.
         let (child_sender, child_receiver) = mpsc::channel();
 
-        // The root's child becomes the root node of the worker
         let child_root = root_branch
             .children
             .get_mut(first_path_component)
@@ -244,10 +242,11 @@ impl ParallelMerkle {
                 match child {
                     Child::Node(node) => Ok(node),
                     Child::AddressWithHash(address, _) => {
-                        Ok(proposal.read_node(address)?.deref().clone())
+                        // Track deletion of the removed child from the root (if it was persisted).
+                        Ok(proposal.read_for_update(address.into())?)
                     }
                     Child::MaybePersisted(maybe_persisted, _) => {
-                        Ok(maybe_persisted.as_shared_node(proposal)?.deref().clone())
+                        Ok(proposal.read_for_update(maybe_persisted)?)
                     }
                 }
             })
@@ -301,7 +300,7 @@ impl ParallelMerkle {
     fn worker(
         &mut self,
         pool: &ThreadPool,
-        proposal: &NodeStore<MutableProposal, FileBacked>,
+        proposal: &mut NodeStore<MutableProposal, FileBacked>,
         root_branch: &mut BranchNode,
         first_path_component: PathComponent,
         response_sender: Sender<Result<Response, FileIoError>>,
@@ -439,7 +438,7 @@ impl ParallelMerkle {
             // doesn't already exist.
             let worker = self.worker(
                 pool,
-                &mutable_nodestore,
+                &mut mutable_nodestore,
                 &mut root_branch,
                 first_path_component,
                 response_sender.clone(),
--- a/fwdctl/src/check.rs
+++ b/fwdctl/src/check.rs
@@ -69,32 +69,35 @@ pub(super) fn run(opts: &Options) -> Result<(), api::Error> {
     };
 
     let nodestore = NodeStore::open(storage)?;
-    let db_stats = if opts.fix {
-        let (nodestore, report) = nodestore.check_and_fix(check_ops);
-        if let Err(e) = nodestore {
-            println!("Error fixing database: {e}");
-        }
-        println!("Fixed Errors ({}):", report.fixed.len());
-        for error in report.fixed {
-            println!("\t{error}");
-        }
-        println!();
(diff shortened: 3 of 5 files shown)


rkuris:

This is actually correct. We can just create a followup if you'd rather defer this to later.

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+            // context switching on submit; but, does add a significant
+            // amount of overhead for small writes and after the thread sleeps.
+            .setup_sqpoll(SUBMISSION_QUEUE_IDLE_TIME_MS)
+            // do not stop processing the submission queue on error, submit
+            // all events and report errors in the completion queue
+            .setup_submit_all()
+            // NOTE: .setup_defer_taskrun() is not used because it conflicts
+            // with the `setup_sqpoll` thread and causes the kernel to return
+            // EINVAL when initializing the ring. With `setup_single_issuer` and
+            // `setup_sqpoll`, we are allowed to sumit from multiple threads and
```
rkuris: typo
```suggestion
            // `setup_sqpoll`, we are allowed to submit from multiple threads and
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+    /// the operation failed, returns a `BatchError` describing the failure.
+    ///
+    /// If `length` is non-zero after a successful write, the entry must be
+    /// re-submitted to write the remaining data.
+    fn handle_result(&mut self, res: i32) -> Result<usize, BatchError> {
+        #![expect(clippy::arithmetic_side_effects, clippy::cast_sign_loss)]
+
+        if res < 0 {
+            let err = std::io::Error::from_raw_os_error(-res);
+            // rust stdlib maps EAGIN to WouldBlock; re-submit in that case
```
rkuris:
```suggestion
            // rust stdlib maps EAGAIN to WouldBlock; re-submit in that case
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+            if matches!(err.kind(), std::io::ErrorKind::WouldBlock) {
+                trace!(
+                    "io-uring write at offset {} returned EAGIN, re-submitting",
```
rkuris:
```suggestion
                    "io-uring write at offset {} returned EAGAIN, re-submitting",
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+    /// If `length` is non-zero after a successful write, the entry must be
+    /// re-submitted to write the remaining data.
+    fn handle_result(&mut self, res: i32) -> Result<usize, BatchError> {
+        #![expect(clippy::arithmetic_side_effects, clippy::cast_sign_loss)]
+
+        if res < 0 {
+            let err = std::io::Error::from_raw_os_error(-res);
+            // rust stdlib maps EAGIN to WouldBlock; re-submit in that case
+            if matches!(err.kind(), std::io::ErrorKind::WouldBlock) {
+                trace!(
```
rkuris: IMO this should be `debug!` instead of trace, but I'm okay with it like this too.
```suggestion
                debug!(
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+                    "io-uring write at offset {} returned EAGIN, re-submitting",
+                    self.original_offset
+                );
+                firewood_counter!(
+                    "ring.eagin_write_retry",
+                    "amount of io-uring write entries that have been re-submitted due to EAGIN io error"
```
rkuris:
```suggestion
                    "amount of io-uring write entries that have been re-submitted due to EAGAIN io error"
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+
+        let written = res as u32;
+        if written > self.length {
+            return Err(BatchError {
+                batch_offset: self.original_offset,
+                error_offset: self.offset,
+                err: std::io::Error::other("kernel wrote more data than requested"),
+                incurable: false,
+            });
+        }
```
rkuris: Why is incurable not true? These seem like catastrophic errors. As mentioned, something is very wrong if we got zero bytes or more than we requested was written.

Maybe the overflow is clearer with checked_sub?

```
self.length = self.length.checked_sub(written).ok_or_else(BatchError {
...
```

Would also be good to know what `written` actually is in the error message as well

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+        // The kernel will spurrously interrupt this call if this thread needs
+        // to be woken up for an unrelated reason; we ignore those interrupts.
+        // The kernel will also return EBUSY if its internal submission queue is
+        // full, in which case we break out to process completions. If the CQ
+        // is empty, we may spin here until the syscall blocks.
+        match self.submitter.submit_and_wait(1) {
+            Ok(_) => {}
+            Err(ref err) if ignore_poll_error(err) => {}
+            Err(err) => return Err(err),
+        }
```
rkuris: I think this is clearer than the match, but YMMV:
```suggestion
        if let Err(err) = self.submitter.submit_and_wait(1) {
            if !ignore_poll_error(&err) {
                return Err(err);
            }
        }
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+    }
+
+    fn enqueue_single_op(&mut self) -> EnqueueResult {
+        let Some(entry) = self.pop_next_entry() else {
+            return EnqueueResult::Empty;
+        };
+
+        // debug_assert_eq so it `Debug` prints the entry on failure
+        debug_assert_eq!(
+            self.hashtable_remove(entry.original_offset),
```
rkuris: This is a sneaky way to verify that the entry is not already there, and only happens in debug mode. I think this belongs in the insert path -- I added another comment there.

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+
+    fn is_finished(&self) -> bool {
+        // we never take from the iterator without submitting it to the kernel
+        // or adding it to the backlog, so we don't need to check the iterator
+        self.backlog.is_empty() && self.outstanding.is_empty()
+    }
+
+    fn hashtable_insert(&mut self, entry: QueueEntry<'batch>) {
+        self.outstanding
+            .insert_unique(entry.hashtable_hash(), entry, QueueEntry::hashtable_hash);
```
rkuris: The docs for this method say the behavior is undefined if the value already exists. We only check for the existence of the value in the debug path, so we'd get UB.

I'd rather have this method be fallable and use the `entry` api.

storage/src/linear/filebacked.rs
```diff
@@ -346,6 +284,35 @@ impl OffsetReader for PredictiveReader<'_> {
...
+struct UnlockOnDrop(File);
+
+impl Drop for UnlockOnDrop {
+    fn drop(&mut self) {
+        // ignore the error, we might not have ever called `lock`
+        _ = self.0.unlock();
+    }
+}
+
+impl std::fmt::Debug for UnlockOnDrop {
```
rkuris: Why couldn't we `#[derive(Debug)]` on UnlockOnDrop?

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+            HashTableEntry::Vacant(_) => None,
+        }
+    }
+
+    fn pop_next_entry(&mut self) -> Option<QueueEntry<'batch>> {
+        if let Some(entry) = self.backlog.pop_front() {
+            Some(entry)
+        } else {
+            self.entries.next()
+        }
```
rkuris:
```suggestion
        self.backlog.pop_front().or_else(|| self.entries.next())
```

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+    }
+
+    fn add_entry_to_backlog(&mut self, entry: QueueEntry<'batch>) {
+        if entry.offset == entry.original_offset {
+            // new entry, queue at the back
+            self.backlog.push_back(entry);
+        } else {
+            // partially written entry, re-queue at the front for priority
```
rkuris, line 444: why is this needed?

storage/src/linear/io_uring.rs
```diff
@@ -0,0 +1,811 @@
...
+            }
+        }
+    }
+}
+
+impl std::fmt::Debug for IoUringProxy {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("IoUringProxy").finish_non_exhaustive()
+    }
+}
```
rkuris: nit: we can use the derive_where crate for this boilerplate code

rkuris (changes requested):

storage/src/linear/io_uring.rs
demosdemon: `EAGIN` is the error code, not `EAGAIN`

storage/src/linear/io_uring.rs
demosdemon: `EAGIN` is the error code, not `EAGAIN`.

storage/src/linear/io_uring.rs
demosdemon: why is what needed?

storage/src/linear/filebacked.rs
demosdemon: Because that would include `UnlockOnDrop` in the debug output when it's not relevant. This forwards to the File `Debug`.

storage/src/linear/io_uring.rs
demosdemon: `EAGIN` is the error code, not `EAGAIN`.

storage/src/linear/io_uring.rs
demosdemon: This is an unlikely error case and we can check for it, but it really should be checked before we make it into the write batch code.

storage/src/linear/io_uring.rs
demosdemon: that looks worse IMO

storage/src/linear/io_uring.rs
demosdemon: This is an unlikely error case and we can check for it, but it really should be checked before we make it into the write batch code.

storage/src/linear/filebacked.rs
rkuris: Why isn't it relevant?

storage/src/linear/io_uring.rs
demosdemon: As described on the field for `incurable`, `incurable` is reserved for errors that prevent us from communicating with the kernel. This doesn't.

This is also a very unlikely error. We should probably panic instead.

storage/src/linear/io_uring.rs
rkuris: pushing to the front sometimes

demosdemon: As mentioned, it's to prioritize completing new blocks before we enqueue new blocks. This should also prevent the dequeue from allocating more space in most cases because we will try to drain it from priority inserts before inserting items we grabbed from the iterator.

storage/src/linear/io_uring.rs
demosdemon: The assumption here is that the node allocator didn't cause a duplicate or overlapping region to be written. But, if we want to be defensive against that then checking for duplicate inserts is also not sufficient. We also need to check for overlapping inserts.

Are we assuming the node allocator is broken here?

storage/src/linear/io_uring.rs
demosdemon: 🍳 egg on my face. it is `EAGAIN`. I swear I this check earlier cause I questioned it myself.

rkuris: It's okay, egg on your face and AGIN on mine. It seems to be a common enough error that searching for it puts you down a deep rabbit hole.

storage/src/linear/io_uring.rs
rkuris: Can we rename 'incurable' to `kernel_comm` or `kcomm_error` to avoid this for casual readers?

storage/src/linear/io_uring.rs
rkuris: are we checking this elsewhere if we don't have debug assertions enabled?

storage/src/linear/io_uring.rs
rkuris: I'm just wanting a check before we get into an undefined state if somehow this does happen, even in release mode, or alternatively use the entry() api.

storage/src/linear/filebacked.rs
```diff
@@ -142,35 +83,21 @@ impl FileBacked {
...
         Ok(Self {
-            fd,
             cache: Mutex::new(LruCache::new(node_cache_size)),
             free_list_cache: Mutex::new(LruCache::new(free_list_cache_size)),
             cache_read_strategy,
             filename: path,
             #[cfg(feature = "io-uring")]
-            ring: Mutex::new(ManuallyDrop::new(ring)),
+            ring,
+            fd: UnlockOnDrop(fd),
```
rkuris, line 100: This seems unnecessarily complicated. All file locks are automatically dropped by the OS when the file is closed, and we seem to ignore any errors during unlock anyway. Is there a reason we can't just close the file when we drop it normally?

demosdemon: this is not new behavior
https://github.com/ava-labs/firewood/pull/1352

rkuris:
Sure but I think this was the cause of all that too.

demosdemon:
> Sure but I think this was the cause of all that too.

The cause of what? What do you mean? Do you have feedback or not? You are commenting on something that is unrelated to this PR.

rkuris:
What I meant was I bet the reason the locks were not getting released on
close was because of the pending io-uring thread held by the kernel.

Let's merge this anyway and see if that is true in a different issue.

PR #1504, opened 2025-12-03
test: optimize init_merkle by removing redundant verification passes

rkuris:
The init_merkle test helper was performing verification after each insert and after each phase (insert, hash, commit), resulting in 3 full passes over the dataset with get_value() calls.

This change removes the redundant verification passes and keeps only a single final verification pass at the end to ensure correctness.

Performance improvements:
- Debug mode: 33.4s → 0.35s (95x faster)
- Release mode: 4.2s → 0.015s (280x faster)

33s is way too long, especially since this method will be used to initialize quite a few merkles in some upcoming tests.

The merkle tree is still fully verified at the end by checking that all key-value pairs can be retrieved correctly from the committed state.

### DIFF
--- a/firewood/src/merkle/tests/mod.rs
+++ b/firewood/src/merkle/tests/mod.rs
@@ -60,40 +60,12 @@ where
         let value = v.as_ref();
 
         merkle.insert(key, value.into()).unwrap();
-
-        assert_eq!(
-            merkle.get_value(key).unwrap().as_deref(),
-            Some(value),
-            "Failed to insert key: {key:?}",
-        );
-    }
-
-    for (k, v) in iter.clone() {
-        let key = k.as_ref();
-        let value = v.as_ref();
-
-        assert_eq!(
-            merkle.get_value(key).unwrap().as_deref(),
-            Some(value),
-            "Failed to get key after insert: {key:?}",
-        );
     }
 
     let merkle = merkle.hash();
-
-    for (k, v) in iter.clone() {
-        let key = k.as_ref();
-        let value = v.as_ref();
-
-        assert_eq!(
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/tests/mod.rs
```diff
@@ -60,40 +60,12 @@ where
...
-
-    for (k, v) in iter.clone() {
-        let key = k.as_ref();
-        let value = v.as_ref();
-
-        assert_eq!(
-            merkle.get_value(key).unwrap().as_deref(),
-            Some(value),
-            "Failed to get key after insert: {key:?}",
-        );
```
demosdemon, line 79: fun fact, removing this is the bulk of your time savings. I discovered this a while back but I never dug in further to understand why it was so expensive with an unhashed proposal.

firewood/src/merkle/tests/mod.rs
```diff
@@ -60,40 +60,12 @@ where
         let value = v.as_ref();
 
         merkle.insert(key, value.into()).unwrap();
-
-        assert_eq!(
-            merkle.get_value(key).unwrap().as_deref(),
-            Some(value),
-            "Failed to insert key: {key:?}",
-        );
```
RodrigoVillar, line 68: This `assert_eq!()` was grossly redundant considering we do the same check right afterwards.

firewood/src/merkle/tests/mod.rs
rkuris: I think it's because it verifies all the keys starting at 1..N on each pass. That is, first time it just checks K1, then it checks K1+K2, then K1+K2+K3. This results in roughly 8 million `get_value` calls.

Issue #1502, opened 2025-12-03
perf: `HashOrRlp` perf loss from `Hash` implementation


demosdemon:
Performance analysis shows the commit from #1363 caused a noticeable hit to performance. I suspect that is because of the change to the [`Hash` implementation](https://github.com/ava-labs/firewood/blob/216426207c735aa77ff74032ecae55ff6152b557/storage/src/node/branch.rs#L268-L275) that runs `keccak256` over the data if it is an RLP. We do need to replace the SmallVec with an inlined array and that will help perf, but only negligibly compared to whatever mitigation we can devise to prevent that extra hashing.

demosdemon:
Hmm. Removing the `Hash` implementation on `HashOrRlp` yields a successful build meaning this isn't the problem. We're not even hashing them like we are the `TrieHash`.

PR #1505, opened 2025-12-04
perf: avoid extra copying when comparing `HashOrRlp`

demosdemon:
The change I previously made to `HashOrRlp` to reduce some code redundancy added extra overhead in the form of unnecessary copying when comparing two `HashOrRlp` instances for equality or hashing them. That added a noticable increase to the average propose+commit benchmark times.

This change reverts to the prior method of comparing without copying.

I have also moved the `HashType` into its own module with `TrieHash` and `HashOrRlp` as submodules.

Closes: #1502

### DIFF
--- a/storage/src/hashtype.rs
+++ b/storage/src/hashtype.rs
@@ -28,18 +28,17 @@ pub trait IntoHashType {
     fn into_hash_type(self) -> HashType;
 }
 
-#[cfg(feature = "ethhash")]
 impl IntoHashType for crate::TrieHash {
     #[inline]
     fn into_hash_type(self) -> HashType {
-        self.into()
-    }
-}
+        #[cfg(feature = "ethhash")]
+        {
+            self.into()
+        }
 
-#[cfg(not(feature = "ethhash"))]
-impl IntoHashType for crate::TrieHash {
-    #[inline]
-    fn into_hash_type(self) -> HashType {
-        self
+        #[cfg(not(feature = "ethhash"))]
+        {
+            self
+        }
     }
 }


storage/src/hashtype.rs
```diff
@@ -0,0 +1,45 @@
...
+/// This is used to allow different hash types to be conditionally used, e.g., when the
+/// `ethhash` feature is enabled. When not enabled, this suppresses the clippy warnings
+/// about useless `.into()` calls.
+pub trait IntoHashType {
+    /// Converts the value into a `HashType`.
+    #[must_use]
+    fn into_hash_type(self) -> HashType;
+}
+
+#[cfg(feature = "ethhash")]
```
rkuris: nit: we could avoid some code duplication by putting this on the `self.into()` below

storage/src/hashtype/ethhash.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            // if both are hash or rlp, we can skip hashing
+            (HashOrRlp::Hash(h1), HashOrRlp::Hash(h2)) => h1 == h2,
+            (HashOrRlp::Rlp(r1), HashOrRlp::Rlp(r2)) => r1 == r2,
+            // otherwise, one is a hash and the other isn't, so hash the rlp and compare
+            (HashOrRlp::Hash(h), HashOrRlp::Rlp(r)) | (HashOrRlp::Rlp(r), HashOrRlp::Hash(h)) => {
+                // avoid copying by comparing the hash directly; copying into a
+                // TrieHash adds a noticeable overhead
+                Keccak256::digest(r).as_slice() == h.as_ref()
```
rkuris, line 88: 🆒

PR #1506, opened 2025-12-05
refactor(checker): separate check and fix

RodrigoVillar:
While going through the checker code, I noticed two things:

- We always check but sometimes fix. Therefore, it didn't make sense to have a `check_and_fix()` method since this can be simplified into just a `fix()` method (as `check()` already exists).
- `check_and_fix()` returns `(Result<NodeStore<Arc<ImmutableProposal>, S>, FileIoError>, FixReport)` but we never use the returned `NodeStore` and so this can be simplified into just `Result<FixReport, FileIoError>`

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -385,7 +385,8 @@ mod test {
     use std::path::PathBuf;
 
     use firewood_storage::{
-        CheckOpt, CheckerError, HashedNodeReader, IntoHashType, NodeStore, TrieHash,
+        CheckOpt, CheckerError, HashedNodeReader, IntoHashType, LinearAddress, MaybePersistedNode,
+        NodeStore, TrieHash,
     };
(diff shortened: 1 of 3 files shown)


storage/src/checker/mod.rs
```diff
@@ -578,31 +577,12 @@ pub struct FixReport {
 }
 
 impl<S: WritableStorage> NodeStore<Committed, S> {
-    /// Check the node store and fix any errors found.
+    /// Given a check report, fixes any errors found.
     /// Returns a report of the fix operation.
-    // TODO: return a committed revision instead of an immutable proposal
```
RodrigoVillar, line 583: Note to reviewer: we never used the returned `NodeStore` but if we plan on doing so, I think this should be a GH issue rather than an in-code TODO.

PR #1519, opened 2025-12-09
fix: Race condition between view and commit

rkuris:
During commit(), a revision transitions from being in proposals (step 6) to by_hash (step 5). The view() method was checking by_hash first, then proposals, creating a window where a concurrent view() could fail to find a revision that exists but is mid-transition.

Fix by reversing the check order in view() to match the write order in commit(). Now view() checks proposals first (where revisions start), then by_hash (where they end up), ensuring the revision is always findable during the transition.

Also added a concurrency stress test that reproduces the race condition with 10 viewer threads and 5 committer thread over 1000 iterations. This fails without the fix and works with it.

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -442,7 +442,6 @@ mod tests {
 
         let config = ConfigManager::builder()
             .create(true)
-            .truncate(true)
             .manager(
                 RevisionManagerConfig::builder()
                     .max_revisions(100_000) // Set very high to prevent reaping during test


firewood/src/manager.rs
```diff
@@ -421,4 +421,188 @@ mod tests {
...
+        const NUM_VIEWER_THREADS: usize = 10;
+        const NUM_COMMITTER_THREADS: usize = 5;
+
+        // Create a temporary database
+        let temp_file = NamedTempFile::new().unwrap();
+        let db_path = temp_file.path().to_path_buf();
+
+        let config = ConfigManager::builder()
+            .create(true)
+            .truncate(true)
```
RodrigoVillar:
```suggestion
```

nit: I don't think truncation is necessary here since `temp_file` will be empty.

firewood/src/manager.rs
```diff
@@ -421,4 +421,188 @@ mod tests {
...
+                    .max_revisions(100_000) // Set very high to prevent reaping during test
+                    .build(),
+            )
+            .build();
+
+        let manager = Arc::new(RevisionManager::new(db_path, config).unwrap());
+
+        // Create an initial proposal and commit it to have a non-empty base
+        let base_revision = manager.current_revision();
+        let mut proposal = NodeStore::new(&*base_revision).unwrap();
```
RodrigoVillar, line 456:
```suggestion
        let mut proposal = NodeStore::new(base_revision.as_ref()).unwrap();
```

nit: this is easier to read IMO (however, will defer to you as to whether this usage is idiomatic or not).

demosdemon: I prefer `&*` over adding `as_ref` everywhere but I also won't go out of my way to change one. They're also technically different, `&*` invokes the `Deref` trait whereas `as_ref` invokes the `AsRef<_>` trait. Most of the time that distinction is irrelevant but important when an object has multiple `AsRef<_>` implementations and the compiler can't infer which one to use.

rkuris: Totally a style thing. Let's leave it.

RodrigoVillar: > Totally a style thing. Let's leave it.

100%, thanks for the context @demosdemon !

PR #1520, opened 2025-12-09
feat: prevent accidental UB on iterators

### LINKED ISSUE #1467: docs(ffi): Clarify Iterator docs
After #1466, much of the docs were updated, but not these operations. A few minutes should be spent making `godoc`-renderable documentation describing the methods.

alarso16:
I didn't add the keepAliveHandle to iterators, mostly because I didn't understand how it should be used. I think this is a sufficient test for this. Additionally, I added a big godoc about it, since it was convenient to do now.

Closes #1467

### DIFF
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -1567,6 +1567,57 @@ func TestIterDone(t *testing.T) {
 	r.NoError(it.Err())
 }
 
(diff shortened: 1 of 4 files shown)


alarso16:
This isn't how iterators work, which is clear form the stack corruption seen. @AminR443 will be looking for a different way to guarantee that iterators work as expected

PR #1521, opened 2025-12-09
test: Iterator outlives revision

alarso16:
I just realized that we test nothing about the guarantees of iterators to be used after their underlying revision/proposal is out of scope/dropped. Since I was working on #1520 and noticed some weird errors, I decided to test this directly. It appears that something strange is happening for sure...


alarso16:
@AminR443 talked to me, and this isn't expected to work. He will be working on a permanent solution to guarantee valid behavior

PR #1523, opened 2025-12-10
fix(io-uring): use sqwait to correctly synchronize kernel updates

demosdemon:
This is necessary to ensure the kernel does not drop completion events.

Closes: #1522

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,6 +1172,7 @@ dependencies = [
  "bytemuck",
  "bytemuck_derive",
  "bytes",
+ "cfg-if",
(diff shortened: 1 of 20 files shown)


rkuris:
Not sure this matters so much since we're mostly disabling io-uring for now.

demosdemon:
> Not sure this matters so much since we're mostly disabling io-uring for now.

True. I'm going to continue testing io-uring to figure out why it's slower. May need to continue tuning the options like disabling SQPOLL, since we're doing this dedicate pub/sub loop.

PR #1526, opened 2025-12-10
feat(ffi): Stub for code hashes in proof

alarso16:
This is just a stub for the codehashes method that will be added eventually to enable development in the next release.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,23 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Possible Solutions**
+If known, give some ideas of what may be done to fix the problem.
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: enhancement
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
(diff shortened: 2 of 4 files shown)


ffi/proofs.go
```diff
@@ -221,6 +221,19 @@ func (p *RangeProof) FindNextKey() (*NextKeyRange, error) {
...
+// If the proof does not contain any account nodes, an empty list is returned.
+//
+// Note: this method is only relevant for Ethereum tries.
+// This method can only be called after a succesful verification of the range proof,
+// otherwise an error is returned.
+//
+// TODO(#1157): implement this method to extract code hashes from account nodes.
+func (*RangeProof) CodeHashes() ([]Hash, error) {
+	return []Hash{}, nil
+}
```
demosdemon: Do we want to return a slice, which requires collecting them all at once, or do we want to return an iterator, which allows lazily collecting them? 

In Go, functions that return [`iter.Seq`](https://pkg.go.dev/iter#Seq) or [`iter.Seq2`](https://pkg.go.dev/iter#Seq2) (for key-value pairs) functions can be used in range loops making them as ergonomic as rust iterators.

Ignore the below the HR, a `Seq` would work fine as expected as we would construct the FFI iterator and drop it all within the same executing body within the `Seq` executor. Our signature is simply

```go
func (*RangeProof) CodeHashes() iter.Seq2[Hash, error] {
    return func(yield (Hash, error) bool) {
        iter, err := buildIter()
        if err != nil {
            yield(Hash{}, err)
            return
        }

        defer iter.Free()

        // poll iter loop, return if empty or if yield returns false
        for {
            hash, err = iter.Next()
            if err == errExhausted {
                return
            }
            if !yield(hash, err) || err != nil {
                return
            }
        }
    }
}
```

callers using this would look something like:

```go
for hash, err := range proof.CodeHashes() {
    if err != nil {
        // handle err and break loop
    }

    // handle hash
}
```

--- 

Our signature would need to be something like:


```go
// CodeHashes returns (dropper, iterator, nil) or (nil, nil, error)
func (*RangeProof) CodeHashes() (func() error, iter.Seq[Hash], error) {
    panic("todo")
}
```

a [`Pull`er](https://pkg.go.dev/iter#Pull) may be more ergonomic by ensuring the stop function is part of the contract.

alarso16: As you suggested, an `iter.Seq2[Hash, error]` might be a great way to do this!

alarso16: For now, I think we should return nothing, rather than the first response returning an error. I don't think it matters too much though.

Issue #1441, opened 2025-11-07
Enhancement: populate read cache with nodes from key-value iter on drop


demosdemon:
TBD

rkuris:
LGTM, but add that if the iterator is exhausted we should not cache the nodes at drop time. This can happen if the iterator asks for a specific range and we fetched all of them.

PR #1531, opened 2025-12-11
refactor: partially refactor iter types into submodules

demosdemon:
This partially refactors some of the iterator types into submodules. 

The `peek` method was added to `ReturnableIterator` because it will be necessary in a future PR  for #1441.

### DIFF
--- a/firewood/src/iter.rs
+++ b/firewood/src/iter.rs
@@ -1,9 +1,12 @@
 // Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
-use crate::merkle::{Key, Value};
-use crate::v2::api::{KeyType, KeyValuePair};
+mod filtered_key_range;
+pub(crate) mod returnable;
 
+use crate::merkle::{Key, Value};
+
+pub use self::filtered_key_range::{FilteredKeyRangeExt, FilteredKeyRangeIter};
 use firewood_storage::{
     BranchNode, Child, FileIoError, NibblesIterator, Node, PathBuf, PathComponent, PathIterItem,
     SharedNode, TriePathFromUnpackedBytes, TrieReader,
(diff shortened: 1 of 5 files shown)


firewood/src/iter.rs
```diff
@@ -568,61 +565,6 @@ fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter) -> Key {
     data.into_boxed_slice()
 }
 
-/// An iterator over key-value pairs that stops after a specified final key.
-#[derive(Debug)]
-#[must_use = "iterators are lazy and do nothing unless consumed"]
-pub enum FilteredKeyRangeIter<I, K> {
```
rkuris, line 574: nit: naming

`FilteredKeyRangeIter::Unfiltered` seems odd. Perhaps `MaybeFilteredKeyRangeIter` is better?

firewood/src/iter/filtered_key_range.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// encountered.
+    pub fn new(iter: I, last_key: Option<K>) -> Self {
+        match last_key {
+            Some(k) => FilteredKeyRangeIter::Filtered { iter, last_key: k },
+            None => FilteredKeyRangeIter::Unfiltered { iter },
+        }
+    }
+}
+
+impl<I: Iterator<Item = T>, T: KeyValuePair, K: KeyType> Iterator for FilteredKeyRangeIter<I, K> {
```
rkuris, line 37: Should also `impl FusedIterator` since this code will also fuse the underlying iterator.

Probably should document that it will fuse it as well.

firewood/src/iter/returnable.rs
```diff
@@ -0,0 +1,68 @@
...
+pub(crate) trait ReturnableIteratorExt: Iterator + Sized {
+    /// Wraps this iterator in a [`ReturnableIterator`].
+    fn returnable(self) -> ReturnableIterator<Self> {
+        ReturnableIterator::new(self)
+    }
+}
+
+impl<I: Iterator> ReturnableIteratorExt for I {}
+
+/// Similar to a peekable iterator. In addition to being able to peek at the
```
rkuris, line 13: I'm not sure I'm following what this does that Peekable's blanket implementation does not do., except it's a little less flexible. Peekable keeps a `peeked: Option<Option<I::Item>>` inside it, but also supports stuff like `peek_mut` which can be used to implement `return_item`.

demosdemon: > also supports stuff like `peek_mut` which can be used to implement `return_item`.

[`peek_mut`](https://doc.rust-lang.org/stable/std/iter/struct.Peekable.html#method.peek_mut) does not do what I need. If it returned `&mut Option<I::Item>`, that would be sufficient. However, it returns `Option<&mut I::Item>` which is not the same thing.

`Peekable` provides no way for us to push the item back to the front of the iterator.

This is similar to using `VecDeque` to `push_front` but with a slot for only one element.

demosdemon:
stale

PR #1532, opened 2025-12-11
feat!: open database with directory instead of filepath

### LINKED ISSUE #1463: feat!: Open database should be a folder
We have several use cases for keeping all database files in a folder at this point:
 - RootStore wants one or more separate files
 - Deletion log should be in an append-only file #832 
 - Database chunking or freezing a section of the database to make copying easier

This issue simply changes the parameter to open to require it to be a directory instead of a file, and fixes all the tests and API callers both internally and externally.

RodrigoVillar:
## Why this should be merged

This PR closes #1463 by changing database initialization to take in a directory rather than a filepath. With this change, the Firewood database structure is as follows:

```
├── <INSERT_DIR_NAME_HERE>
│   ├── firewood.db
│   ├── root_store (optional)
```

## How this works

- Updates database + revision manager to use directories
- Updates documentation to use directories rather than filepaths in examples
- Updates tests to use directory when creating Firewood instance

## How this was tested

CI + added an additional unit test

Yes

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -32,6 +32,8 @@ use firewood_storage::{
     TrieHash,
 };
 
+const DB_FILE_NAME: &str = "firewood.db";
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, TypedBuilder)]
 /// Revision manager configuratoin
 pub struct RevisionManagerConfig {
@@ -114,8 +116,7 @@ impl RevisionManager {
             std::fs::create_dir_all(&db_dir).map_err(RevisionManagerError::IOError)?;
         }
 
-        let file = db_dir.join("firewood.db");
-
+        let file = db_dir.join(DB_FILE_NAME);
         let fb = FileBacked::new(
             file,
             config.manager.node_cache_size,


firewood/src/manager.rs
```diff
@@ -99,16 +100,24 @@ pub(crate) enum RevisionManagerError {
         provided: Option<HashKey>,
         expected: Option<HashKey>,
     },
-    #[error("An IO error occurred during the commit: {0}")]
+    #[error("An FileIO error occurred during the commit: {0}")]
```
rkuris: Nit: "A FileIO"

fwdctl/src/check.rs
```diff
@@ -41,7 +41,7 @@ pub struct Options {
 }
 
 pub(super) fn run(opts: &Options) -> Result<(), api::Error> {
-    let db_path = PathBuf::from(&opts.database.dbpath);
+    let db_path = PathBuf::from(&opts.database.dbpath).join("firewood.db");
```
rkuris, line 44: This important string should be a const somewhere both for emphasis and just in case we rebrand it later.

firewood/src/manager.rs
RodrigoVillar: Good catch; done: c4b4421

fwdctl/src/check.rs
RodrigoVillar: Done: 97e9e8f

PR #1535, opened 2025-12-11
feat!: FFI logging options in Go open config; remove StartLogs

rkuris:
Adds WithLogPath and WithLogFilter functional options to initialize env_logger via FFI at database open time.
- Full `RUST_LOG` filter support via env_logger::Builder::parse_filters()
- Log path required; use /dev/stdout for stdout; no defaults
- Logging initialized before DB open; remains initialized even if open fails
- Global per-process logger; fails when already initialized
- Remove deprecated StartLogs and LogConfig from Go; keep fwd_start_logs called internally
- Add trace log on Db::new() and comprehensive TestLogging covering error/success cases

BREAKING CHANGE: StartLogs API removed; use WithLogPath/WithLogFilter options on New()

## Why this should be merged

* We need this immediately to help with debugging the rust side
* Logging must be configured at open to enable future per-database log strategies.
* Aligns Go-side configuration with Rust [env_logger](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) semantics using RUST_LOG.
* Unblocks downstream integration (e.g., coreth) to set logging level and output programmatically.

Now we support any logging option as configured from the go side, which means we probably need some configuration options in coreth to set up logging the way we want.

## How this works

- Go: [New(path, WithLogPath("/dev/stdout"), WithLogFilter("info"))](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) initializes logging before DB open.
- Rust FFI: LogArgs validates UTF-8; [env_logger::Builder](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) parses RUST_LOG filters and targets the provided path.
- Logging is global per-process; subsequent init attempts return an error.
- If DB open fails, logger remains initialized (explicitly documented).

## How this was tested

- [TestLogging](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (Go):
- Empty path: logging disabled, DB opens
- Invalid path: initialization fails with clear error
- Valid path + trace: log file created and non-empty; contains “Opening Firewood database …”
- Second init: fails with “attempted to set a logger after the logging system was already initialized”

### Workspace checks

 - [cargo fmt](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
- cargo test --workspace --features ethhash,logger --all-targets
- cargo clippy --workspace --features ethhash,logger --all-targets -- -D warnings
- cargo doc --no-deps

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,9 @@
+[profile.default]
+# Skip tests prefixed with test_slow in default profile
+default-filter = "not test(test_slow)"
+
 [profile.ci]
 # Do not cancel the test run on the first failure.
 fail-fast = false
+# Override default filter to run all tests including slow ones
+default-filter = "all()"
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -20,11 +20,12 @@ jobs:
         run: |
           cargo login ${{ secrets.CARGO_TOKEN }}
           cargo publish -p firewood-macros
-      - name: publish firewood-triehash crate
-        continue-on-error: false
-        run: |
-          cargo login ${{ secrets.CARGO_TOKEN }}
-          cargo publish -p firewood-triehash
+      # TODO(demosdemon): detect when version is bumped and only publish then
+      # - name: publish firewood-triehash crate
+      #   continue-on-error: false
+      #   run: |
+      #     cargo login ${{ secrets.CARGO_TOKEN }}
+      #     cargo publish -p firewood-triehash
       - name: publish firewood-storage crate
         continue-on-error: false
         run: |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -119,6 +119,16 @@ cargo doc --no-deps                                                     # Ensure
 
 All tests must pass, and there should be no clippy warnings.
 
+### Slow Tests
+
+If your PR modifies code that is tested by any test prefixed with `test_slow_`, you should also run the full test suite with the `ci` profile to ensure those tests pass:
+
+```bash
+cargo nextest run --workspace --features ethhash,logger --all-targets --profile ci
+```
+
+The `ci` profile includes slow tests that are skipped in the default profile for faster local development.
+
 ### Markdown Linter
 
 If your PR touches any Markdown file, run the following:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,53 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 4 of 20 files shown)


ffi/src/logging.rs
```diff
@@ -9,57 +9,58 @@ use crate::BorrowedBytes;
 pub struct LogArgs<'a> {
     /// The file path where logs for this process are stored.
     ///
-    /// If empty, this is set to `${TMPDIR}/firewood-log.txt`.
+    /// This is required and must not be empty. Use "/dev/stdout" for stdout logging.
```
demosdemon, line 12: We shouldn't recommend use of `/dev/stdout` as it will cause interleaving of the stdout from the parent process -- even if the parent process is also implemented Rust sharing the same stdlib statics.

- Any synchronization in Go around printing to stdout will be bypassed.
- Because `File::open("/dev/stdout")` returns a new fd, any synchronization in Rust will also be bypassed.

ffi/src/logging.rs
```diff
@@ -78,29 +79,27 @@ impl LogArgs<'_> {
...
                     e.kind(),
                     format!("failed to open log file `{}`: {e}", log_path.display()),
                 )
             })?;
 
-        env_logger::Builder::new()
-            .filter_level(level)
-            .target(Pipe(Box::new(file)))
+        let mut builder = env_logger::Builder::new();
+        builder.target(Pipe(Box::new(file)));
```
demosdemon, line 96: We should wrap `file` with `std::io::BufWriter` to make using `stdout` slightly more copacetic (doesn't prevent interleaving but makes it harder if we only write whole lines). `env_logger` calls `flush` after writing out every event.

rkuris:
> I don't think we can do this until logs are supported per-database. Otherwise, we can't even make multiple databases in tests (since coreth Firewood doesn't know it's a test at instantiation). If this is added for the next release, I think I would just have to remove logging all together

We could still do this. If the database open fails because the logger is already initialized, just create the database with no logging. That would be a workaround until we implement per-database logging correctly.

PR #1541, opened 2025-12-13
ci: switch to cargo-nextest

RodrigoVillar:
## Why this should be merged

[cargo nextest](https://nexte.st/) is a test runner similar to `cargo test` which offers faster testing times along with customization of our test suite. This tool was recommended to me by @demosdemon and after playing with it for some time, I think the team would benefit from usage of this tool.

## How this works

- Replaces usage of `cargo test` with `cargo nextest` in CI
- Replaces recommended usage of `cargo test` with `cargo nextest run` in documentation

## How this was tested

CI + ran `cargo nextest run` locally


.github/workflows/ci.yaml
```diff
@@ -151,17 +151,18 @@ jobs:
...
       - uses: dtolnay/rust-toolchain@stable
+      - uses: taiki-e/install-action@nextest
       - uses: Swatinem/rust-cache@v2
         with:
           save-if: "false"
           shared-key: ${{ matrix.profile-key }}
-      - name: Run tests with features enabled
-        run: cargo test --locked --verbose ${{ matrix.profile-args }}
+      - name: Run tests with cargo-nextest
+        run: cargo nextest run --locked --profile ci ${{ matrix.profile-args }}
```
rkuris: This is a bit strange. The `profile` option is set in `profile args` so sometimes you'll be getting `--profile ci --profile maxperf`. I think this means you get maxperf and `profile ci` is ignored, but I'm not sure.

RodrigoVillar: The CLI configuration for `cargo nextest run` can be found here: https://nexte.st/docs/running/#options-and-arguments.

In particular, `--profile` refers to the `cargo nextest` profile to use (which lives in `.config/nextest.toml` and `--cargo-profile` refers to the cargo build profile to use. Looking here,

https://github.com/ava-labs/firewood/blob/b6d2b5fcda850156f38aa2c8665496936239ed28/.github/workflows/ci.yaml#L153-L154

we see that `max-perf` is using the correct argument flag (i.e. `--cargo-profile`).

PR #1542, opened 2025-12-14
fix(ffi/iterator): keep view alive for iterator lifetime

AminR443:
## Why this should be merged
- Fixes a lifetime bug in the FFI where an iterator could outlive its backing Revision/Proposal view, leading to invalid access/crashes.
- Makes iterator behavior safe and predictable by keeping the underlying `DbView` alive for the iterator’s lifetime.
- Adds proper tests.

## How this works
IteratorHandle now holds both the `BoxKeyValueIter` and an `ArcDynDbView` clone.

## How this was tested
Added Go tests in `ffi/firewood_test.go`: `TestIterOutlivesRevision` and `TestIterOutlivesProposal`.

Yes.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,5 +3,3 @@
 ## How this works
 
 ## How this was tested
-
-## Need to be documented in RELEASES.md?
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1131,7 +1131,6 @@ dependencies = [
 name = "firewood-fwdctl"
 version = "0.0.16"
 dependencies = [
- "anyhow",
  "askama",
  "assert_cmd",
  "clap",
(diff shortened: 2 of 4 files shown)


ffi/src/iterator.rs
```diff
@@ -3,31 +3,32 @@
...
-pub struct IteratorHandle<'view>(Option<BoxKeyValueIter<'view>>);
-
-impl<'view> From<BoxKeyValueIter<'view>> for IteratorHandle<'view> {
-    fn from(value: BoxKeyValueIter<'view>) -> Self {
-        IteratorHandle(Some(value))
-    }
+pub struct IteratorHandle<'view> {
+    // order is important to ensure the view is dropped after the iterator
+    iter: Option<BoxKeyValueIter<'view>>,
+    view: Option<ArcDynDbView>,
```
rkuris: This should be `Option<(BoxKeyValueIter<'view>, ArcDynDbView)>` otherwise you are representing some invalid states where one is `Some` and the other is `None`.

AminR443: Good point. Done.

PR #1543, opened 2025-12-14
refactor(ffi/iterator): move tests into a separate file

AminR443:
## Why this should be merged
As discussed with @alarso16 before, this moves the iterator-related tests into a separate file to keep things organized.

## How this works
Moves iterator-related tests out of `firewood_test.go` into their own file.

## How this was tested
Existing go tests.

N/A


alarso16 (approved):
Thanks! So much simpler

PR #1546, opened 2025-12-15
chore: Remove reference to RELEASES.md

rkuris:
## Why this should be merged

We don't use this process and so asking for it on a PR is useless.

## How this works

Removes the line

## How this was tested

Touché


rkuris:
duplicate

PR #1547, opened 2025-12-15
test(firewood): reduce iteration count

RodrigoVillar:
## Why this should be merged

When running `cargo nextest --features ethhash,logger --all-targets`, this test takes over a minute to run as it's running 5000 iterations. We can just run this test with 500 iterations and still achieve the same test coverage.

## How this works

Reduces number of iterations run from 5000 to 500 (10x reduction)

## How this was tested

CI


rkuris (approved):
It's still only 8 threads, but we don't need to loop 5000 times to find a problem in this code.

PR #1553, opened 2025-12-16
feat: Difference iterator for change proofs part 1

bernard-avalabs:
## Why this should be merged

This is part 1 of the change proof implementation. It includes a pre-order traversal implementation of the Merkle trie that supports skipping traversal of a node's children. This skipping feature will be especially useful when performing a diff traversal between two tries, as sub-tries that have a matching hash can be skipped during the traversal. This implementation also supports starting the traversal at a start key.

## How this works

This pre-order traversal implementation works by pushing a node's children onto a traversal stack, and traversing the nodes in stack order. A pre-order traversal of a trie will iterate through the nodes and keys in lexicographical order. This is mostly a textbook design with the only difference being that a node's children is not pushed onto the stack on the same next call that returned the node. Instead, they are pushed during the subsequent next call. This is an optimization to omit the need to push the children onto the stack in the event of a hash match that allow the children of the current node to be skipped.

## How this was tested

There is a test case that tests the correctness of the iteration with and without a start key using random keys/values. It also does a simple test of the skip children functionality. A more thorough test for skip children will be included in part 2 when the feature is used as part of the diff iterator.


firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+            Child::Node(node) => node.clone().into(),
+            Child::AddressWithHash(addr, child_hash) => {
+                hash = Some(child_hash.into_triehash());
+                trie.read_node(addr)?
+            }
+            Child::MaybePersisted(maybe_persisted, child_hash) => {
+                hash = Some(child_hash.into_triehash());
+                maybe_persisted.as_shared_node(&trie)?
+            }
+        };
```
rkuris: This would be easier to read if we add Child::as_shared_node and then it becomes this:
`let node = frame.node.as_shared_node();`

The new function should mention that it will allocate a new Arc in the Child::Node case.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+            traversal_stack: vec![],
+            trie,
+        };
+        // If the root node is not None, then push a `TraversalStackFrame` for the
+        // root onto the traversal stack. It will be used on the first call to
+        // next or next_internal.
+        if let Some(root) = trie.root_node() {
+            preorder.traversal_stack.push(TraversalStackFrame {
+                pre_path: Path::default(),
+                node: Child::Node((*root).clone()),
```
rkuris: Why do we need to create a new child here? Can't we just clone or reference the original root_node?

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+        if let Some(prev_node_state) = self.node_state.take()
+            && let Node::Branch(branch) = &*prev_node_state.node
+        {
+            // Since a stack is LIFO and we want to perform pre-order traversal, we pushed the
+            // children in reverse order.
+            for (path_comp, child) in branch
+                .children
+                .iter()
+                .rev()
+                .filter_map(|(pc, c)| c.as_ref().map(|child| (pc, child)))
```
rkuris: This looks a lot like what `iter_present` is doing. Perhaps `iter_present` should be renamed `into_iter_present` since it consumes, and I think you need to write the non-consuming one.

I think it would make the code a lot more readable since it took me some time to figure out what the filter_map was doing.

recommend: `iter_present().rev()`

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+            // the start key onto the traversal stack.
+            if let Some(prev_node_state) = self.node_state.take()
+                && let Node::Branch(branch) = &*prev_node_state.node
+            {
+                let mut passed_check = false;
+                for (path_comp, child) in branch
+                    .children
+                    .iter()
+                    .rev()
+                    .filter_map(|(pc, c)| c.as_ref().map(|child| (pc, child)))
```
rkuris: Same as earlier comment

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+                    let child_pre_path = Path::from_nibbles_iterator(
+                        prev_node_state
+                            .path
+                            .iter()
+                            .copied()
+                            .chain(once(path_comp.as_u8())),
+                    );
+                    // A previous child has already been pushed onto the traversal stack. The
+                    // remaining children can be pushed without additional checks.
+                    if passed_check {
```
rkuris: nit: I probably would have written this as:
 - get the iterator
 - read from the iterator until we get the state we want
 - exhaust the iterator in the new state
 
 That avoids the mutable switch and might be a bit more readable. Something like:
 
 ```
let reversed_children_with_full_path = branch.into_iter().map(|(path_comp, child)| (path_comp, child, Path::from_nibbles_iterator(...));

while let Some((path_comp, child, full_path)) = reversed_children_with_full_path.next() {
    let child_key = ... // along with all the other code in the else clause below
    if unique_node.is_empty() || ... { ...; break }
}
self.traversal_stack.extend(reversed_children_with_full_path.map(...))
```

File: firewood/src/merkle/changes.rs [hunk 6]
```diff
@@ -0,0 +1,416 @@
...
+
+                // Since pre-order traversal of a trie iterates through the nodes in lexicographical
+                // order, we can stop the traversal once we see a node key that is larger than or
+                // equal to the key. We stop the traversal by pushing  the current node with its
+                // pre-path and hash back to the stack. Calling `next` will process this node.
+                let node_key = key_from_nibble_iter(state.path.iter().copied());
+                if node_key >= *key {
+                    self.traversal_stack.push(TraversalStackFrame {
+                        pre_path,
+                        node: Child::Node((*state.node).clone()),
```
rkuris: Same as earlier comment.

rkuris:
I'm confused as to why you're handling the Child::Node case. This case only happens when the child has not been hashed. Does this algorithm work when the children are not hashed yet? I think the answer is no.

In that case, I think you can make an assumption you can always get a SharedNode from the child, and call unreachable for the case where the child has no hash yet.

bernard-avalabs:
> I'm confused as to why you're handling the Child::Node case. This case only happens when the child has not been hashed. Does this algorithm work when the children are not hashed yet? I think the answer is no.
> 
> In that case, I think you can make an assumption you can always get a SharedNode from the child, and call unreachable for the case where the child has no hash yet.

The traversal algorithm **does not** require child nodes to have hashes. The diff algorithm in part 2 of the PR also works without hashes, although it wouldn't be able to skip matching sub-tries.

I assume that we want to support child nodes without hashes so we can perform diffs on MutableProposals.

firewood/src/merkle/changes.rs
bernard-avalabs: In the `next` function, we push each child of the current node (along with its pre-path and hash if available as part of a TraversalFrame) onto the traversal stack. We don't actually load the Child into a Node until we pop the child from the traversal stack for processing. The only cases where we have a Node instead of a Child when writing to the traversal stack are when we are pushing the root, or when we are pushing a node back onto the traversal stack once we have satisfied the end condition for `iterate_to_key`. For uniformity I create a Child from a Node in these two cases.

I could change the algorithm to eagerly load the Child into a Node and store NodeStates instead of TraversalStackFrames on the traversal stack. This would be slower if we need to terminate a traversal early as it may load Childs into Nodes that are never used.

firewood/src/merkle/changes.rs
bernard-avalabs: I wasn't aware of `iter_present`. The `filter_map` was suggested in a comment in the previous PR. I've replaced `iter().rev().filter_map(...)` with `iter_present().rev()`.

firewood/src/merkle/changes.rs
bernard-avalabs: See my earlier response.

firewood/src/merkle/changes.rs
rkuris: I was more thinking that `node` should be a `SharedNode`. Otherwise, what you're doing is copying the contents of the node each time, whereas if this were a `SharedNode` you'd just be cloning an Arc of the node, which may even be in the cache.

rkuris:
> I assume that we want to support child nodes without hashes so we can perform diffs on MutableProposals.

We do not want or need to support this. The only operations on a MutableProposal should be mutating it until it reaches the final state, and then converting it to an ImmutableProposal.

Having this around will raise the complexity when we later make the children of a MutableProposal different from ImmutableProposal.

firewood/src/merkle/changes.rs
bernard-avalabs: But then I would need to load the Child into a Node when I push it onto the traversal stack. That's not necessary here since I already have the Node, but it is in `next` where I am pushing each `Child` of the node onto the traversal stack.

If you think it is preferable, I can change the algorithm to eagerly load the Child to Node in `next`, which would allow me to avoid this copy here. But as mentioned previously, it would cause extra storage reads if we decide to terminate a traversal early.

bernard-avalabs:
> > I assume that we want to support child nodes without hashes so we can perform diffs on MutableProposals.
> 
> We do not want or need to support this. The only operations on a MutableProposal should be mutating it until it reaches the final state, and then converting it to an ImmutableProposal.
> 
> Having this around will raise the complexity when we later make the children of a MutableProposal different from ImmutableProposal.

Okay, I assumed that we wanted to support MutableProposals since your previous PR (https://github.com/ava-labs/firewood/pull/839/) had tests for MutableProposals (`test_diff_empty_mutable_trees`).

I'll remove support for Child nodes without hashes.

storage/src/node/branch.rs
```diff
@@ -141,6 +142,27 @@ impl Child {
...
+    /// - `Ok(SharedNode)` contains the node if successfully read
+    /// - `Err(FileIoError)` if there was an error reading from storage
+    pub fn as_shared_node(&self, storage: &dyn NodeReader) -> Result<SharedNode, FileIoError> {
+        Ok(match self {
+            Child::Node(node) => node.clone().into(),
+            Child::AddressWithHash(addr, _child_hash) => storage.read_node(*addr)?,
+            Child::MaybePersisted(maybe_persisted, _child_hash) => {
+                maybe_persisted.as_shared_node(&storage)?
+            }
+        })
```
demosdemon:
```suggestion
        match self {
            Child::Node(node) => Ok(node.clone().into()),
            Child::AddressWithHash(addr, _) => storage.read_node(*addr),
            Child::MaybePersisted(maybe_persisted, _) => maybe_persisted.as_shared_node(&storage),
        }
```

This is a very, very pedantic suggestion. Feel free to discard, but I'll explain my reasoning: overuse of the `?` operator can cause LLVM to generate worse code. This is because the `?` actually desugars to:

```rust
match storage.read_node(*addr) {
    Ok(t) => t,
    Err(err) => return <Result<Shared, FileIoError>>::from_residual(ControlFlow::Break(FileIoError:from(err))),
}
```

LLVM may fail to inline or deduplicate any of those calls resulting in us losing what little tail-call optimization is in Rust.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,412 @@
...
+    trie: &'a dyn TrieReader,
+    traversal_stack: Vec<TraversalStackFrame>,
+}
+
+impl<'a> PreOrderIterator<'a> {
+    /// Create a pre-order iterator for the trie that starts at `start_key`. The iterator takes an
+    /// optional root hash which is available when the trie is from an `ImmutableProposal`.
+    #[cfg_attr(not(test), expect(dead_code))]
+    fn new(
+        trie: &'a dyn TrieReader,
```
demosdemon: `dyn` references aren't free. They eliminate monomorphization and force all method calls to be done via dynamic dispatch. I don't particularly see a reason why we'd need to use a dyn reference here.

firewood/src/merkle/changes.rs
demosdemon: We do have a rather large amount of code that fully clones a node when it only needs a reference. We need an enum like this to prevent that.

```rust
enum MaybeOwnedNodeRef<'a> {
    Owned(SharedNode),
    Borrowed(&'a Node),
}
```

storage/src/node/branch.rs
```diff
@@ -141,6 +142,27 @@ impl Child {
...
+    /// # Arguments
+    ///
+    /// * `storage` - A reference to a `NodeReader` implementation that can read nodes from storage.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result<SharedNode, FileIoError>` where:
+    /// - `Ok(SharedNode)` contains the node if successfully read
+    /// - `Err(FileIoError)` if there was an error reading from storage
+    pub fn as_shared_node(&self, storage: &dyn NodeReader) -> Result<SharedNode, FileIoError> {
```
demosdemon: There's no need for this to be a dyn reference.

```suggestion
    pub fn as_shared_node<S: NodeReader>(&self, storage: &S) -> Result<SharedNode, FileIoError> {
```

firewood/src/merkle/changes.rs
bernard-avalabs: I used `dyn` to make it easier to support both hashed and unhashed nodes (although it might have been possible to not need it even with that requirement). I've removed it now that we are only supporting hashed nodes.

firewood/src/merkle/changes.rs
bernard-avalabs: I've removed this clone as part of a larger update to push NodeStates directly to the traversal stack.

firewood/src/merkle/changes.rs
bernard-avalabs: I've updated the code to push NodeState directly onto the TraversalStack which avoids creating a Child here.

rkuris:
> Okay, I assumed that we wanted to support MutableProposals since your previous PR (#839) had tests for MutableProposals (`test_diff_empty_mutable_trees`).

Yeah, that was an outstanding task. My original thought was it would be easy to support, but it isn't actually necessary, and arguably we shouldn't allow anything to happen to MutableProposals except mutate them further or convert them to immutable.

> I'll remove support for Child nodes without hashes.
👍

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+use crate::{
+    iter::key_from_nibble_iter,
+    merkle::{Key, PrefixOverlap},
+};
+
+/// Contains all of a node's state that is needed for node comparison in `DiffMerkleNodeStream`.
+/// It includes the nodes full path and its hash if available.
+#[cfg_attr(not(test), expect(dead_code))]
+#[derive(Clone)]
+struct NodeState {
```
rkuris: nit: Maybe this should be called "NodeDetails" or "FullNode" or something. When I was rereading this, I was thinking "oh this defines the state of a Node?" for a moment, but it doesn't.

Another option is "ComparableNodeInfo".

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+    hash: Option<TrieHash>,
+}
+
+impl NodeState {
+    // Creates a NodeState from a `Child`, its pre-path, and trie for reading the node from storage.
+    fn new<T: NodeReader>(pre_path: Path, child: &Child, trie: &T) -> Result<Self, FileIoError> {
+        let node = child.as_shared_node(trie)?;
+        // Create the full path for the node. This is used to compare with the full path
+        // of the current node from the other trie.
+        let mut path = pre_path;
```
rkuris: nit: path -> fullpath

Then you can shorten the comment by removing the first sentence.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+    ///
+    /// This implementation flips the order such that we don't push a node's children onto the
+    /// traversal stack until the subseqent call to `next`. This is done by saving the node state of
+    /// the current node in `node_state` and keeping that available for the next call. Skipping
+    /// traversal of the children then just involves setting `node_state` to None.
+    #[cfg_attr(not(test), expect(dead_code))]
+    fn next(&mut self) -> Result<Option<&NodeState>, FileIoError> {
+        firewood_counter!(
+            "firewood.change_proof.next",
+            "number of next calls on PreOrderIterator",
```
rkuris: nobody will know what PreOrderIterator is in this context
```suggestion
            "number of nodes iterated to calculate a change proof",
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+            "traversal" => "next")
```
rkuris: This tag should be unnecessary, as nobody else uses `firewood.change_proof.next` with a different value for traversal.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// of a node in a different trie, as we would then need to pop its children off the traversal
+    /// stack.
+    ///
+    /// This implementation flips the order such that we don't push a node's children onto the
+    /// traversal stack until the subseqent call to `next`. This is done by saving the node state of
+    /// the current node in `node_state` and keeping that available for the next call. Skipping
+    /// traversal of the children then just involves setting `node_state` to None.
+    #[cfg_attr(not(test), expect(dead_code))]
+    fn next(&mut self) -> Result<Option<&NodeState>, FileIoError> {
+        firewood_counter!(
```
rkuris, line 111: We need to tweak the dashboard to show this data as well. I have a PR that adds a ton of metrics almost ready for review, and we could add this to that batch, but perhaps we need another ticket.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+            hash: child.hash().map(|hash| hash.clone().into_triehash()),
+        })
+    }
+}
+
+/// Contains the state required for performing pre-order iteration of a Merkle trie. This includes
+/// the state of the current node, a reference to a trie that implements the `HashedNodeReader`
+/// trait, and a stack that contains the `NodeState` of nodes to be traversed with calls to `next`.
+struct PreOrderIterator<'a, T: HashedNodeReader> {
+    node_state: Option<NodeState>,
```
rkuris: Let's document here what `None` means, as it seems fairly nuanced.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+                // Traversal stack is empty. This means the key is lexicographically larger than
+                // all of the keys in the trie. Calling `next` will return None.
+                return Ok(());
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+#[allow(
```
rkuris: nit: expect, so these don't go stale. Not so important in tests, but still a good habit.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+            })
+    }
+
+    fn gen_random_keys(
+        rng: &SeededRng,
+        committed_keys: &HashSet<Vec<u8>>,
+        num_keys: usize,
+        start_val: usize,
+    ) -> (Vec<BatchOp<Vec<u8>, Box<[u8]>>>, usize) {
+        const CHANCE_DELETE: usize = 2;
```
rkuris: This needs units
```suggestion
        const CHANCE_DELETE_PERCENT: usize = 2;
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+        let mut seen_keys = std::collections::HashSet::new();
+        let mut i = 0;
+        let mut batch = Vec::new();
+
+        while batch.len() < num_keys {
+            if !committed_keys.is_empty() && rng.random_range(1..=100) <= CHANCE_DELETE {
```
rkuris:
```suggestion
            if !committed_keys.is_empty() && rng.random_range(0..100) < CHANCE_DELETE {
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+                let del_key = gen_delete_key(rng, committed_keys, &mut seen_keys);
+                if let Some(key) = del_key {
+                    batch.push(BatchOp::Delete { key });
+                    continue;
+                }
+                // If we couldn't generate a delete key, then just fall through and create
+                // a BatchOp::Put.
+            }
+            let key_len = rng.random_range(1..=32);
```
rkuris: This will tend to generate a lot of duplicates, since there is a 1/32 chance that the key length will be 1, and then a 1/256 chance that the same key will be chosen. That might be okay here but calling it out in case it's a problem (it was in some other tests we wrote).

A relatively easy fix is to force keys to be a minimum length of 2.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+            let key: Vec<u8> = (0..key_len).map(|_| rng.random()).collect();
+            // Only add if key is unique
+            if seen_keys.insert(key.clone()) {
+                batch.push(BatchOp::Put {
+                    key,
+                    value: Box::from(format!("value{}", i + start_val).as_bytes()),
+                });
+            }
+            i += 1;
```
rkuris: Why can't we just always use `batch.len()` instead of keeping track of `i`?

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+        // Just skip the key if it has been seen. Otherwise return it and add the key
+        // to seen keys.
+        key_opt
+            .filter(|del_key| !seen_keys.contains(del_key))
+            .inspect(|del_key| {
+                seen_keys.insert(del_key.clone());
+            })
+    }
+
+    fn gen_random_keys(
```
rkuris: name here is misleading. This doesn't just generate random keys. It generates random(ish) BatchOps.
```suggestion
    fn gen_random_test_batchops(
```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    #[test]
+    fn test_preorder_iterator() {
+        let rng = firewood_storage::SeededRng::from_env_or_random();
+        let (batch, _) = gen_random_keys(&rng, &HashSet::new(), 1000, 0);
+
+        // Keep a sorted copy of the batch.
+        let mut batch_sorted = batch.clone();
+        batch_sorted.sort_by_key(|op| op.key().clone());
```
rkuris: nit: It's a test, but I'm curious why a clone was needed here. This will clone 2*n*log(n) times due to the sort.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+        for item in &batch {
+            // All of the ops should be Puts
+            merkle
+                .insert(
+                    item.key(),
+                    item.value().unwrap().to_vec().into_boxed_slice(),
+                )
+                .unwrap();
+        }
+        let merkle = make_immutable(merkle);
```
rkuris:
```suggestion
        // freeze and compute the hash
        let merkle: Merkle<NodeStore<Arc<ImmutableProposal>, MemStore>> = merkle.into().unwrap();
```
this function is only called from here, so a comment and doing this inline is probably easier. I had to check what make_immutable does, and it's just a simple into() method call with an unwrap.

Historical note: this used to be called `freeze` before it was a fallable type conversion.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// would be inefficient if we want to skip traversing a node's children if its hash matches that
+    /// of a node in a different trie, as we would then need to pop its children off the traversal
+    /// stack.
+    ///
+    /// This implementation flips the order such that we don't push a node's children onto the
+    /// traversal stack until the subseqent call to `next`. This is done by saving the node state of
+    /// the current node in `node_state` and keeping that available for the next call. Skipping
+    /// traversal of the children then just involves setting `node_state` to None.
+    #[cfg_attr(not(test), expect(dead_code))]
+    fn next(&mut self) -> Result<Option<&NodeState>, FileIoError> {
```
rkuris: This should be an implementation of Iterator, as that has a ton more flexibility for any future callers and allows for chaining and other optimizations. Oterators return Option<Result<T>> instead of Result<Option<T>>.

```suggestion
}
impl Iterator for PreOrderIterator {
    type Item = Result<&NodeState, FileIoError>;
    fn next(&mut self) -> Option<Self::Item> {
```

You'll have a little bit of work to do to handle the returned lifetime, but it should be doable without too much trouble.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+                        break;
+                    }
+                }
+
+                // Add the rest of the children (if any) to the traversal stack.
+                for node_state in reversed_children_with_pre_path
+                    .map(|(child, child_pre_path)| NodeState::new(child_pre_path, child, self.trie))
+                {
+                    self.traversal_stack.push(node_state?);
+                }
```
rkuris: Why can't we use extend here?
```suggestion
                    self.traversal_stack.extend(
                        reversed_children_with_pre_path
                            .map(|(child, child_pre_path)| NodeState::new(child_pre_path, child, self.trie)));

```

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+
+impl<'a, T: HashedNodeReader> PreOrderIterator<'a, T> {
+    /// Create a pre-order iterator for the trie that starts at `start_key`.
+    #[cfg_attr(not(test), expect(dead_code))]
+    fn new(trie: &'a T, start_key: &Key) -> Result<PreOrderIterator<'a, T>, FileIoError> {
+        let mut preorder = Self {
+            node_state: None,
+            traversal_stack: vec![],
+            trie,
+        };
```
rkuris: This feels strange. Rather than allocating a vec and then pushing stuff to it, let's figure out whether we need a vec of 1 or zero and fill the vec with that. Something like:

```rust
let traversal_stack = trie
    .root_node()
    .into_iter()
    .map(|root| NodeState { ... })
    .collect();

let mut preorder = Self {
    node_state: None,
    traversal_stack,
    trie,
};
```

Bonus here is that iterating on `Option<T>` is rare but here's a perfect case for it.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,389 @@
...
+        // path of the root node is just its partial path since it has no pre-path.
+        if let Some(root) = trie.root_node() {
+            preorder.traversal_stack.push(NodeState {
+                path: root.partial_path().clone(),
+                node: root,
+                hash: trie.root_hash().map(|hash| hash.clone().into_triehash()),
+            });
+        }
+        preorder.iterate_to_key(start_key)?;
+        Ok(preorder)
```
rkuris:
```suggestion
        preorder.iterate_to_key(start_key)
```
And change iterate_to_key to consume self and return it:
```
fn iterate_to_key(self, key: &Key) -> Result<Self, FileIoError> {
...
```

firewood/src/merkle/changes.rs
bernard-avalabs: I believe your suggested code would try to append `Results` to the traversal stack. I don't believe adding a `?` after NodeState::new(...) is allowed inside a map closure.

firewood/src/merkle/changes.rs
bernard-avalabs: It would be great to add this metric to your batch if this PR is merged first.

firewood/src/merkle/changes.rs
bernard-avalabs: I've changed it to ComparableNodeInfo

firewood/src/merkle/changes.rs
bernard-avalabs: After digging into this for a bit, I believe it would be challenging to return a reference to a NodeState/ComparableNodeInfo in a regular Iterator (see https://rust-lang.github.io/generic-associated-types-initiative/explainer/motivation.html). I've instead implemented PreOrderIterator as a Lending Iterator (https://docs.rs/lender/latest/lender/) which has most of the functionality of Iterator.

firewood/src/merkle/changes.rs
bernard-avalabs: It gives a lifetime error without the clone. I've changed that line to:

`batch_sorted.sort_by(|op1, op2| op1.key().cmp(op2.key()))` which compiles without needing a clone.

firewood/src/merkle/changes.rs
bernard-avalabs: I use `make_immutable` a few times in part 2 of this PR chain. However, given that this is a one-liner, I can replace those calls with `try_into`s as well.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+#[expect(
+    clippy::unwrap_used,
+    clippy::arithmetic_side_effects,
+    clippy::type_complexity
```
demosdemon: nit: `type_complexity` is one I tend to agree with even for tests. I suspect it's complaining about the return type to `gen_random_test_batchops`; if so, then `BatchOp<Vec<u8>, Box<[u8]>>` should probably be hoisted to a type alias.

firewood/src/merkle/changes.rs
```diff
@@ -0,0 +1,416 @@
...
+                // back to the stack. Calling `next` or `next_node_info` will process this node.
+                let node_key = key_from_nibble_iter(node_info.path.iter().copied());
+                if node_key >= *key {
+                    self.traversal_stack.push(node_info);
+                    return Ok(self);
+                }
+
+                // If this node is a leaf, then we don't need to save its info to `node_info` as
+                // it has no children to traverse.
+                if let Node::Branch(_branch) = &*node_info.node {
```
demosdemon: small nit: since you don't need the `_branch` value

```suggestion
                if matches!(*node_info.node, Node::Branch(_)) {
```

demosdemon (approved):

PR #1554, opened 2025-12-16
feat: Difference iterator for change proofs part 2

bernard-avalabs:
## Why this should be merged

This is part 2 of the change proof implementation. It includes the diff iterator to determine all of the changes between two Merkle tries.

## How this works

Lock-step pre-order traversals are performed on the two tries. Depending on the paths of the nodes of the two tries, the diff iterator transitions between different states that determine which keys should be in the addition list and which keys should be in the deletion list. If two sub-tries are determined to be identical due to a matching node hash, then traversal of their children can be skipped.

## How this was tested

The PR includes a number of test cases. One test case that verifies that hash matches between two sub-tries reduces the number of iterations will be in part 3 where additional metrics handling will be introduced.

### DIFF
--- a/firewood/src/merkle/changes.rs
+++ b/firewood/src/merkle/changes.rs
@@ -69,20 +69,20 @@ impl ComparableNodeInfo {
 }
 
 /// Iterator that outputs the difference between two tries and skips matching sub-tries.
-struct DiffMerkleNodeStream<'a, LEFT: HashedNodeReader, RIGHT: HashedNodeReader> {
+struct DiffMerkleNodeStream<'a, Left: HashedNodeReader, Right: HashedNodeReader> {
     // Contains the state of the diff traversal. It is only None after calling `next` or
     // `next_internal` if we have reached the end of the traversal.
     state: Option<DiffIterationNodeState>,
-    left_tree: PreOrderIterator<'a, LEFT>,
-    right_tree: PreOrderIterator<'a, RIGHT>,
+    left_tree: PreOrderIterator<'a, Left>,
+    right_tree: PreOrderIterator<'a, Right>,
 }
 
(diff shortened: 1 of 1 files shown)


firewood/src/merkle/changes.rs
```diff
@@ -49,6 +77,322 @@ impl NodeState {
...
+    right_tree: PreOrderIterator<'a>,
+}
+
+impl<'a> DiffMerkleNodeStream<'a> {
+    /// Constructor where the left and right tries implement the trait `TrieReader`. This
+    /// allows a `DiffMerkleNodeStream` to be constructed with `MutableProposal`s. The
+    /// drawback to using `new_without_hash` is that we don't have the root hashes for
+    /// these tries. The `new` constructor should be used for `ImmutableProposal`s.
+    #[cfg_attr(not(test), expect(dead_code))]
+    pub fn new_without_hash<T: TrieReader, U: TrieReader>(
```
rkuris: Nit: use L and R (or even better, LEFT and RIGHT) as the type names.

rkuris (approved):
This looks really good.

Test coverage for start key is very light. Maybe modify the test that does the interleave to start at each of those keys (and keys inbetween them) and verify you get the correct results.

firewood/src/merkle/changes.rs
```diff
@@ -42,6 +68,294 @@ impl ComparableNodeInfo {
     }
 }
 
+/// Iterator that outputs the difference between two tries and skips matching sub-tries.
+struct DiffMerkleNodeStream<'a, LEFT: HashedNodeReader, RIGHT: HashedNodeReader> {
```
demosdemon: pedantic nit: I saw Ron's other comment; however, convention for type generics is `PascalCase` even if they are multi-character whereas `const` generics are typically `SCREAMING_SNAKE_CASE`. However, I see clippy isn't complaining about this so it's not part of the pedantic style therefore I won't expect us to actually conform to that everywhere.

PR #1555, opened 2025-12-16
feat: Difference iterator for change proofs part 3

bernard-avalabs:
## Why this should be merged

Includes an additional test that uses metrics to verify that the hash optimization in the diff traversal implementation is working.

## How this works

A counter is incremented on each call to next in the pre-order iterator. The test verifies that the hash optimization reduces the number of next calls. A local test recorder has been added in test_utils.rs to work with the metrics framework.

## How this was tested

As mentioned previously, this PR introduces a test case that uses metrics to verify the correctness of the hash optimization.

### DIFF
--- a/storage/Cargo.toml
+++ b/storage/Cargo.toml
@@ -26,7 +26,7 @@ fastrace.workspace = true
 hex.workspace = true
 integer-encoding.workspace = true
 metrics.workspace = true
-metrics-util.workspace = true
+metrics-util = { workspace = true, optional = true }
 nonzero_ext.workspace = true
 rand = { workspace = true, optional = true }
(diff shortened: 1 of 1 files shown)


storage/Cargo.toml
```diff
@@ -26,6 +26,7 @@ fastrace.workspace = true
 hex.workspace = true
 integer-encoding.workspace = true
 metrics.workspace = true
+metrics-util.workspace = true
```
demosdemon: Like `rand`, make this optional here, add it as a non-optional dependency under `dev-dependencies`, and set it to be enabled under `features` when `test_utils` is used.

Issue #1302, opened 2025-09-24
bug: hash mismatch happens sometimes during bootstrapping


rkuris:
I think this is a separate bug from the memory growth issue, so I'm creating a new one.

rkuris:
Known failing blocks and their source:

Antithesis:
- 266
- 308
- 326
- 21109
- 29830
- 34497
- 34502
- 34505


Aaron's runs:
- 308 (dup from Antethesis)
- 34497 (dup from Antethesis)
- 27881 (unique)

Rodrigo:
- 308 (dup from Antithesis + Aaron)
- 21109 (dup from Antithesis)
- 34502 (dup from Antithesis)
- 2519497
- 2532403 :/

aaronbuchwald:
ref https://github.com/ava-labs/avalanchego/issues/4598

aaronbuchwald:
Merged support for re-execution w/ firewood on August 26th here: https://github.com/ava-labs/avalanchego/pull/4180
Reported the issue as the flake started showing up non-deterministically in CI immediately after on August 27th: https://github.com/ava-labs/avalanchego/issues/4598
Removed from CI on August 28th: https://github.com/ava-labs/avalanchego/pull/4227

aaronbuchwald:
Within a single failure case here: https://github.com/ava-labs/avalanchego/actions/runs/19315010303/job/55244272668?pr=4505#step:4:4019

The key-value set from the first attempt to create a proposal on the parent is here:

https://github.com/ava-labs/avalanchego/actions/runs/19315010303/job/55244272668?pr=4505#step:4:4019

The key-value set when we attempt to verify a proposal on top of the same root is different: 

https://github.com/ava-labs/avalanchego/actions/runs/19315010303/job/55244272668?pr=4505#step:4:4027

aaronbuchwald:
> Within a single failure case here: https://github.com/ava-labs/avalanchego/actions/runs/19315010303/job/55244272668?pr=4505#step:4:4019
> 
> The key-value set from the first attempt to create a proposal on the parent is here:
> 
> https://github.com/ava-labs/avalanchego/actions/runs/19315010303/job/55244272668?pr=4505#step:4:4019
> 
> The key-value set when we attempt to verify a proposal on top of the same root is different:
> 
> https://github.com/ava-labs/avalanchego/actions/runs/19315010303/job/55244272668?pr=4505#step:4:4027

The same appears to be true for the success case.

The first attempt to propose on the parent root here: 

[...]

PR #1557, opened 2025-12-17
fix: hash mismatch when proposing on uncommitted proposal

AminR443:
## Why this should be merged

This PR fixes our hash mismatch bug (#1302) which was observed in some of our bootstrapping/reexecution tests.

## How this works

When computing the root hash in `ethhash` mode, we handle a special case where if we have an only child of an account branch, we hash it as if it is a root node. This is done by taking the nibble from its parent and prefixing it to the partial path. Now when this account branch is updated and gets another child, in the new hash computation process we need to update the previous child's hash, and hash it normally this time (pass the nibble back to parent).

To determine whether we need to update the child's hash or not, we call the `ethhash_classify_children` helper function, which could incorrectly classifies a hashed node as unhashed in the case that we create a proposal based on a not-yet-committed proposal:

https://github.com/ava-labs/firewood/blob/3db72fe361230a7d73ed0333d11328c53f6c269f/storage/src/nodestore/hash.rs#L97-L108

Even though the node isn't persisted, it's already hashed and there are cases that we may need to rehash it.

This bug didn't happen all the time because `ProposeOnProposal` only happens when the previous proposal hasn't finished committing yet. When hashing the new proposal, existing `MaybePersisted::Unpersisted` nodes from the previous proposal were skipped and their hash was reused; this classification isn't correct in the described case that a child's hash needs to be updated.

This PR fixes the bug by correctly classifying `MaybePersistedNode` as hashed.

## How this was tested

This PR introduces `TestProposeOnProposalRehash`, which captures an edge case from replay logs. The proposed key values from block 307 and 308 have the characteristic that results in hash mismatch if we commit them through `p307.Propose` (proposal built on block 307's proposal) instead of `db.Propose`. The test fails with `ethhash` enabled without the proposed changes in this PR.

### Block Info
Block 307 proposed batch:

```
d20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7eb10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6:944abef613822fb2031d897e792f89c896ddafc466
d20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7e:f8450180a00000000000000000000000000000000000000000000000000000000000000000a0a33ec8100b06cfbacc612bc2baa4d36f01d5d52df97bcef04f54d5b8ceccbd9280
f22bb46edf31af855938befaa870ed3d86a4ad93a9ecf7c63ceaa80daea9ac4d:f84e8089063443e4a44dc98c00a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0611445c10cb8404ad2103d510e139c3ace61b5acec80505bd1f5870528f587d780
71645710251b71938795615f63f7bcf27fd5aa4ffbaeb7ace16cbd1c76a41e42:f84d1188662cdf1b7d1bc000a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a47080
```

Block 308 proposed batch:

```
d20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7e290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563:944abef613822fb2031d897e792f89c896ddafc466
d20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7e:f8450180a00000000000000000000000000000000000000000000000000000000000000000a0a33ec8100b06cfbacc612bc2baa4d36f01d5d52df97bcef04f54d5b8ceccbd9280
f22bb46edf31af855938befaa870ed3d86a4ad93a9ecf7c63ceaa80daea9ac4d:f84e808906348c71086eb3ac00a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0611445c10cb8404ad2103d510e139c3ace61b5acec80505bd1f5870528f587d780
71645710251b71938795615f63f7bcf27fd5aa4ffbaeb7ace16cbd1c76a41e42:f84d128865e452b75c31a000a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a47080
```

The first key in both of the blocks are associated with the same account (prefix), but the suffix is different.

### Diagram

```mermaid
flowchart LR
    block-307 --> acc
    acc -->|"b"| child_b
    child_b["2304<br/>H=Hash(7389cf2f7a39755d7b9eab4daa654300d1b920ffaddeb65bfae313304af4e2c2)<br/>pp=10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6<br/>val=944abe..."]
    acc["2656<br/>H=Hash(2b64d343059ea3753f1f75ff0d6b5502879da1a36826913e8660ca5e7305d343)<br/>pp=20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7e<br/>val=f84501..."]

    block-308-correct --> acc2
    acc2 -->|"2"| c2_2
    acc2 -->|"b"| c2_b
    c2_2["3296<br/>H=Hash(dd3b2e5a365c80ed79fa853153c6c2469bbaf1d3c17b76d306137470b84091a0)<br/>pp=90decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563<br/>val=944abe..."]
    c2_b["2304<br/>H=Hash(cd70fffcb0193c0841e94e0383aabc3cb995b2e8e253435dd4c17f9eb1a89197)<br/>pp=10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6<br/>val=944abe..."]
    acc2["3648<br/>H=Hash(9651a6cd5fe95088a7204d30fd7bf927833b4187d6da3d8b1bbc826083c0f6f3)<br/>pp=20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7e<br/>val=f84501..."]

    block-308-incorrect --> acc3
    acc3 -->|"2"| c3_2
    acc3 -->|"b"| c3_b
    c3_b["2304<br/>H=Hash(7389cf2f7a39755d7b9eab4daa654300d1b920ffaddeb65bfae313304af4e2c2)<br/>pp=10e2d527612073b26eecdfd717e6a320cf44b4afac2b0732d9fcbe2b7fa0cf6<br/>val=944abe..."]
    c3_2["3296<br/>H=Hash(dd3b2e5a365c80ed79fa853153c6c2469bbaf1d3c17b76d306137470b84091a0)<br/>pp=90decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563<br/>val=944abe..."]
    acc3["3648<br/>H=Hash(01cfe6a279a0b9e175ca3a72a78b30449cdaa058f4901fbe309f2ebf6e2ae78e)<br/>pp=20628954718b36081ad06b1a04fb8896f3090eb5c7ab3cd9086e50e61d88b7e<br/>val=f84501..."]

    style c3_b stroke:#ff0000,stroke-width:3px
    style c2_b stroke:#00ff00,stroke-width:3px
```

### DIFF
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -1598,16 +1598,16 @@ func parseKVFromDumpFormat(dumpstr string) ([][]byte, [][]byte, error) {
 		}
 		parts := strings.Split(line, ":")
 		if len(parts) != 2 {
-			return nil, nil, fmt.Errorf("invalid line format")
+			return nil, nil, errors.New("invalid line format")
 		}
 		key, value := parts[0], parts[1]
 		keyBytes, err := hex.DecodeString(key)
 		if err != nil {
-			return nil, nil, fmt.Errorf("decode key: %w", err)
+			return nil, nil, err
 		}
 		valueBytes, err := hex.DecodeString(value)
 		if err != nil {
-			return nil, nil, fmt.Errorf("decode value: %w", err)
+			return nil, nil, err
 		}
 		keys = append(keys, keyBytes)
 		values = append(values, valueBytes)


AminR443:
I wasn't sure if this was the real source, so I ran a little experiment. I replayed and extracted the key values that are being committed in each commit, and checked to see if that commit makes a change that could fall under this bug. For this I just check if we had that account with single child before and if this commit adds a new child to it.

I ran the test up to block 34550, there are 896 blocks that have this characteristic, and all the blocks mentioned in #1302 (except the ones larger than 34550) had this characteristic as well:

```
Block 266 was in list
Block 308 was in list
Block 326 was in list
Block 21109 was in list
Block 27881 was in list
Block 29830 was in list
Block 34497 was in list
Block 34502 was in list
Block 34505 was in list
```

demosdemon (approved):

PR #1566, opened 2025-12-17
Simple range proof verification

rkuris:
## Why this should be merged

Partial verification of range proofs, without generating tries and checking hashes.

## How this works

The current implementation provides a foundation for range proof verification by:

* Checking structural validity (key ordering, non-empty proof)
* Validating that the requested range matches the provided proof
* Cryptographically verifying the start and end boundary proofs

What's NOT included (tracked in #738):

* Full trie reconstruction from the proof
* Verification of intermediate nodes between boundaries
* Complete validation that no keys are missing within the proven range

## How this was tested

The test suite in range.rs, which includes tests for:

* Valid range proofs with various configurations
* Invalid proofs (modified keys, values, gaps, etc.)
* Edge cases (single element, empty ranges, full trie)

Note: Most tests are currently #ignore because they are not checked yet and will be fixed in followup PRs.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -126,12 +126,14 @@ RUN . "${CARGO_HOME}/env" \
     just \
     ripgrep \
     rustfilt \
-    starship
+    starship \
+  && rm -rf "${CARGO_HOME}/registry" "${CARGO_HOME}/git" "${SCCACHE_DIR}"
 
 # Install Go tools
 RUN curl -sSfL https://golangci-lint.run/install.sh | sh -s -- -b $(go env GOPATH)/bin \
   && go install github.com/reteps/dockerfmt@latest \
-  && go install mvdan.cc/sh/v3/cmd/shfmt@latest
+  && go install mvdan.cc/sh/v3/cmd/shfmt@latest \
+  && go clean -cache -testcache -modcache -fuzzcache
 
 # Install Claude Code CLI
 RUN curl -fsSL https://claude.ai/install.sh | bash
--- a/.github/.golangci.yaml.patch
+++ b/.github/.golangci.yaml.patch
@@ -1,5 +1,5 @@
---- .github/.golangci.yaml	2026-02-20 16:14:53.739528860 +0000
-+++ ffi/.golangci.yaml	2026-02-20 16:00:50.182324273 +0000
+--- .github/.golangci.yaml	2026-03-19 15:26:49
++++ ffi/.golangci.yaml	2026-03-19 15:18:18
 @@ -40,7 +40,7 @@
          - standard
          - default
@@ -41,7 +41,13 @@
      - nolintlint
      - perfsprint
      - prealloc
-@@ -97,12 +97,10 @@
+@@ -93,12 +93,12 @@
+     - usestdlibvars
+     - whitespace
+   settings:
++    errcheck:
++      disable-default-exclusions: true
+     depguard:
        rules:
          packages:
            deny:
@@ -50,11 +56,7 @@
              - pkg: github.com/golang/mock/gomock
                desc: go.uber.org/mock/gomock should be used instead.
              # Note: testify/assert is checked separately by test_warn_testify_assert
-             # in lint.sh to produce warnings instead of errors.
-             - pkg: io/ioutil
-               desc: io/ioutil is deprecated. Use package io or os instead.
-     errorlint:
-@@ -113,33 +111,10 @@
+@@ -113,33 +113,10 @@
      forbidigo:
        # Forbid the following identifiers (list of regexp).
        forbid:
@@ -88,7 +90,7 @@
      revive:
        rules:
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bool-literal-in-expr
-@@ -203,17 +178,6 @@
+@@ -203,17 +180,6 @@
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#useless-break
          - name: useless-break
            disabled: false
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -3,8 +3,14 @@ name: gh-pages
 on:
   push:
     branches:
-      - "main"
-      - "rkuris/gh-pages"
+      - main
+  pull_request:
+    paths:
+      # trigger if a pr touches the workflow definition to validate it, but not
+      # for other changes to avoid unnecessary builds.
+      - .github/workflows/gh-pages.yaml
+  # allow manual trigger because we cannot "retry" a deploy step
+  workflow_dispatch:
   # Rebuild pages after benchmark workflow completes.
   # Currently only scheduled runs (which run on main) trigger this.
   # If we later add feature branch benchmarks, remove `branches` filter -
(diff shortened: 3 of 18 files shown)


firewood/src/lib.rs
```diff
@@ -175,7 +175,7 @@ pub use firewood_storage::logger;
 /// In the event of unexpected behavior in testing, disable this first before
 /// looking elsewhere.
 fn init_logger() {
-    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("trace"))
+    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info"))
```
rkuris, line 184: trace logging for all tests is very expensive, noticed a big improvement here.

Sorry to sneak this in to this PR, really it belongs as a separate PR. Complain if you agree and I'll move it.

firewood/src/merkle/tests/range.rs
```diff
@@ -104,10 +102,12 @@ fn test_bad_range_proof() {
             0 => {
                 // Modified key
                 keys[index] = rng.random::<[u8; 32]>(); // In theory it can't be same
+                continue 'skip_test;
```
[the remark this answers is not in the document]
rkuris: The whole point of this PR is to enable the parts of the test that do work, and out of order is an integral part of this test. The others will get skipped.

firewood/src/merkle/mod.rs
```diff
@@ -259,12 +259,64 @@ impl<T: TrieReader> Merkle<T> {
...
+        {
+            return Err(api::Error::InvalidRange {
+                start_key: requested_first.as_ref().to_vec().into(),
+                end_key: left.0.as_ref().to_vec().into(),
+            });
+        }
+
+        proof
+            .start_proof()
+            .verify(&left.0, Some(&left.1), root_hash)?;
```
demosdemon: This is incorrect. When we generate proofs, we always generate a proof for the lower bound. The proof for the lower bound may or may not prove the first key provided in the range. If the lower bound does not exist, the provided proof may not have all the information required to prove the inclusion of the first key.

firewood/src/merkle/mod.rs
```diff
@@ -259,12 +259,64 @@ impl<T: TrieReader> Merkle<T> {
...
+        {
+            return Err(api::Error::InvalidRange {
+                start_key: right.0.as_ref().to_vec().into(),
+                end_key: requested_last.as_ref().to_vec().into(),
+            });
+        }
+
+        proof
+            .end_proof()
+            .verify(&right.0, Some(&right.1), root_hash)?;
```
demosdemon: Similarly here, when we generate the upper bound proof, we generate a proof for the final key in the yielded range only if we truncated the range. Otherwise, we generate a proof for the requested upper bound. This proof may or may not be sufficient to verify the final yielded key.

firewood/src/merkle/mod.rs
```diff
@@ -259,12 +259,64 @@ impl<T: TrieReader> Merkle<T> {
...
+        {
+            return Err(api::Error::ProofError(
+                ProofError::NonMonotonicIncreaseRange,
+            ));
+        }
+
+        // check that the start and end proofs are valid
+        let left = key_values
+            .first()
+            .ok_or(api::Error::ProofError(ProofError::Empty))?;
```
demosdemon: This is incorrect. An empty set of key-values is only invalid if both key bounds are empty. An empty set of key-values with a bounded range correctly represents an empty set of key-values between the two keys.

firewood/src/merkle/mod.rs
rkuris: Seems like we should perhaps construct the key provided and make sure its <= to the first key.

firewood/src/merkle/tests/range.rs
```diff
@@ -69,7 +68,6 @@ fn test_range_proof() {
 #[test]
 // Tests a few cases which the proof is wrong.
 // The prover is expected to detect the error.
-#[ignore = "https://github.com/ava-labs/firewood/issues/738"]
 fn test_bad_range_proof() {
```
RodrigoVillar: (q / feel-free-to-ignore): I wonder if it's cleaner to split this test into two - one test that tests with verification and one test that doesn't. Having the `'skip_test` loop label is hard to follow considering that we also have `continue` statements which should be kept around.

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,64 @@ impl<T: TrieReader> Merkle<T> {
...
-        _proof: &RangeProof<impl KeyType, impl ValueType, impl ProofCollection>,
+        first_key: Option<impl KeyType>,
+        last_key: Option<impl KeyType>,
+        root_hash: &TrieHash,
+        proof: &RangeProof<impl KeyType, impl ValueType, impl ProofCollection>,
     ) -> Result<(), api::Error> {
-        todo!()
+        // check that the keys are in ascending order
+        let key_values = proof.key_values();
+        if !key_values
```
bernard-avalabs, line 256: nit: Using is_sorted_by is slightly simpler:

```
        if !key_values
            .iter()
            .is_sorted_by(|a, b| a.0.as_ref() < b.0.as_ref())
```

firewood/src/merkle/mod.rs
rkuris: This should be resolved with the changes I made.

firewood/src/merkle/mod.rs
rkuris: Fixed.

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,131 @@ impl<T: TrieReader> Merkle<T> {
...
+                root_hash,
+            )?;
+
+            // If the requested lower bound differs from the first yielded key,
+            // the boundary proof should still be able to prove that yielded key.
+            // This rejects range proofs that skip keys between the bound and payload.
+            if let Some((left_key, _left_value)) = left
+                && requested_first.as_ref() != left_key.as_ref()
+            {
+                let mut gap_iter = self
```
bernard-avalabs: Clarification question:

Is this trying to verify that there is no gap between the end of the last range proof partition that was applied and the start of this range proof partition? 

What if the original trie that the range proofs were being applied to was no empty? Do range proofs include deletes for its range to ensure that existing keys are deleted?

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,131 @@ impl<T: TrieReader> Merkle<T> {
...
+                    .key_value_iter_from_key(requested_first.as_ref())
+                    .stop_after_key(Some(left_key.as_ref()));
+
```
demosdemon: This doesn't really have anything to do with verification and is conflating local state with the proof state. This only works if the `self` trie already contains all of the keys and values in the range proof. If it doesn't, which is typical when verifying a range proof before applying, this will fail.

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,131 @@ impl<T: TrieReader> Merkle<T> {
...
+                .end_proof()
+                .verify(requested_last.as_ref(), expected_end_value, root_hash)?;
+
+            // Symmetric boundary-to-edge consistency check for the upper bound.
+            if let Some((right_key, _right_value)) = right
+                && requested_last.as_ref() != right_key.as_ref()
+            {
+                let mut gap_iter = self
+                    .key_value_iter_from_key(right_key.as_ref())
+                    .stop_after_key(Some(requested_last.as_ref()));
```
demosdemon: same here, checking the `self` trie will only yield the expected result if the local trie already has the same state as the proof

firewood/src/merkle/mod.rs
rkuris: No. Range proofs do NOT contain deletes. It is assumed that all values between the two endpoints specified in the range proof contain no other values. If there are any values in that range other than the ones specified in the proof, they are not in the target revision, and should be removed implicitly if this range proof is being applied.

Unlike a change proof, since there is no "base" revision for a range proof, we don't know what keys and/or values might be there. We only know what's in the version being asked about.

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,131 @@ impl<T: TrieReader> Merkle<T> {
...
+
+        if key_values.is_empty() && first_key.is_none() && last_key.is_none() {
+            return Err(api::Error::ProofError(ProofError::Empty));
+        }
+
+        let left = key_values.first();
+        let right = key_values.last();
+
+        // Verify that first_key (if provided) is <= the first key in the proof
+        if let (Some(ref requested_first), Some((left_key, _))) = (first_key.as_ref(), left)
```
alarso16: I _think_ this `ref` is unnecessary (also below)

firewood/src/merkle/mod.rs
rkuris: Removed key_value_iter_from_key gap checks in 5379b1b077faecd6966b025cfd8556a29a624b25

firewood/src/merkle/mod.rs
rkuris: fixed in 5379b1b077faecd6966b025cfd8556a29a624b25

firewood/src/merkle/mod.rs
rkuris: fixed all 4 in 21c9a65b54aad5aa849b08197946dd84e4b62a71

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,131 @@ impl<T: TrieReader> Merkle<T> {
...
+                        ProofError::RangeProofStartBeyondFirstKey,
+                    ));
+                };
+
+                if actual_first.as_ref() != left_key.as_ref() {
+                    return Err(api::Error::ProofError(
+                        ProofError::RangeProofStartBeyondFirstKey,
+                    ));
+                }
+            }
```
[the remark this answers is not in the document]
rkuris: fixed in 5379b1b077faecd6966b025cfd8556a29a624b25

firewood/src/merkle/mod.rs
```diff
@@ -262,12 +262,131 @@ impl<T: TrieReader> Merkle<T> {
...
+                    return Err(api::Error::ProofError(
+                        ProofError::RangeProofStartBeyondFirstKey,
+                    ));
+                };
+
+                if actual_first.as_ref() != left_key.as_ref() {
+                    return Err(api::Error::ProofError(
+                        ProofError::RangeProofStartBeyondFirstKey,
+                    ));
+                }
```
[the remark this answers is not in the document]
rkuris: fixed in 5379b1b077faecd6966b025cfd8556a29a624b25

PR #1570, opened 2025-12-17
fix: correct firewood_gauge! macro to use describe_gauge!

rkuris:
## Why this should be merged

Previously, the firewood_gauge! macro was incorrectly calling
metrics::describe_counter! instead of metrics::describe_gauge!,
which would register gauge metrics with incorrect descriptions.

## How this works

This commit:
- Fixes the macro to properly call describe_gauge! for both
  labeled and unlabeled gauge variants
- Updates existing gauge usage in manager.rs to use firewood_gauge!
  macro with proper descriptions
- Adds tests for both firewood_counter! and
  firewood_gauge! macros
- Updates documentation to include .set() method in gauge usage

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,5 @@
 # CODEOWNERS
 # Note: more specific rules overrule wildcard
-*        @rkuris @demosdemon
-/ffi     @rkuris @demosdemon @alarso16
-/.github @rkuris @demosdemon @aaronbuchwald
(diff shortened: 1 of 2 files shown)


AminR443:
It seems you already made this change in #1569.

rkuris:
> It seems you already made this change in #1569.

Yep, it was supposed to be separate. Closed.

PR #1571, opened 2025-12-17
feat(metrics)!: proposal tracking

rkuris:
BREAKING CHANGE: Renamed metrics for clarity
- firewood.proposals.based_on_proposal to firewood.proposals.created (with base=proposal tag)
- firewood.proposals.outstanding to firewood.proposals.uncommitted

## Why this should be merged

Add metrics for tracking proposal lifecycle:
- firewood.proposals.created with base tag (db/proposal).
This allows us to see how many propose-on-propose versus propose-on-db
happen during normal operations.

- firewood.proposals.uncommitted gauge for active proposals
This allows us to detect leaks

- firewood.proposals.discarded counter for dropped proposals
This helps us identify how many proposals are just discarded instead of
getting committed

- firewood.proposals.reparented counter for commit reparenting
This shows how many times we reparent a proposal to a committed revision

## How this was tested

Marked DO NOT MERGE until we can properly test this in bootstrapping

### DIFF
--- a/benchmark/bootstrap/README.md
+++ b/benchmark/bootstrap/README.md
@@ -9,6 +9,7 @@ Before running the script, you'll need:
 - AWS CLI installed and configured on your machine
 - Authenticated AWS session: `aws sso login`
   - Your session should be configured to use the `Experimental` account.
+  - This must either be the AWS default or named and passed to this script as `--profile PROFILE`.
 
 ## What It Does
 
--- a/benchmark/bootstrap/aws-launch.sh
+++ b/benchmark/bootstrap/aws-launch.sh
@@ -8,12 +8,15 @@ AVALANCHEGO_BRANCH=""
 LIBEVM_BRANCH=""
 NBLOCKS="1m"
 CONFIG="firewood"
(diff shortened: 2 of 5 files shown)


firewood/src/manager.rs
```diff
@@ -296,10 +296,33 @@ impl RevisionManager {
...
-            .lock()
-            .retain(|p| !Arc::ptr_eq(&proposal, p) && Arc::strong_count(p) > 1);
+        // referenced by anyone else. Track how many were discarded (dropped without commit).
+        {
+            let mut lock = self.proposals.lock();
+            let mut discarded = 0u64;
+            lock.retain(|p| {
+                let should_retain = !Arc::ptr_eq(&proposal, p) && Arc::strong_count(p) > 1;
+                if !should_retain {
+                    discarded = discarded.saturating_add(1);
```
demosdemon: wrapping add is cheaper and guaranteed to never actually wrap because allocator rules prevent you from ever having more than `isize::MAX` items in a slice/vector.

demosdemon (approved):

Issue #1524, opened 2025-12-10
ffi/metrics: provide option to enable expensive metric calculations


demosdemon:
AvalancheGo has this flag. Firewood should provide this flag as well. This can be done either when opening the database or configuration metrics. If done when configuring metrics, this can be done by setting global static atomic bool that is read with relaxed ordering when determining whether or not to calculate the metrics. If done when opening the database, this can be added as an open option but will need to be threaded through the codebase everywhere the option is needed.

The former option is easier but the later option is preferred long-term but is more invasive. 

I am also open to a convoluted third option where we use Thread-Local Storage (TLS) to set a context variable when entering database methods and then check that TLS for the context variable when needed which prevents tainting global state.

alarso16:
I think this is great. In the mean time, prior to #1090, there's no reason not to do the first option. It should be very not-invasive, but still most of the decision making about what might be expensive.

PR #1579, opened 2025-12-22
fix(bootstrap): Enable bootstrap metrics listener

rkuris:
Due to a change in avalanchego, we're no longer collecting metrics locally when running the bootstrapping tests. This was super convenient before, and adding new metrics and changing dashboards is much harder without this test grafana environment working.

This appears to be the result of several separate problems:

 1. Avalanchego doesn't enable a prometheus listener on port 6060, most likely due to the changes in https://github.com/ava-labs/avalanchego/pull/4418.
 2. Firewood is no longer opened with a prometheus listener on port 3000.
 3. Firewood does not enable the metrics server when opening the database, unless expensive metrics are enabled.

I think we don't want firewood listening directly, so I'm not fixing #2. However, prometheus should still attempt to listen on port 3000 for the other benchmark tests, so I'm leaving the prometheus listener running on that port for launched instances.

Issue #3 is a problem in avalanchego. We probably should not consider all metrics expensive though, so we should initialize the metrics server regardless of this setting. This is a separate avalanchego bug that should be fixed as part of https://github.com/ava-labs/firewood/issues/1524

### DIFF
--- a/benchmark/bootstrap/aws-launch.sh
+++ b/benchmark/bootstrap/aws-launch.sh
@@ -454,7 +454,7 @@ runcmd:
   # execute bootstrapping
   - >
     sudo -u ubuntu -D /mnt/nvme/ubuntu/avalanchego --login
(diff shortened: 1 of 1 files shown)


rkuris:
Bootstrap script is gone

Issue #1578, opened 2025-12-20
Metrics not getting enabled by default


rkuris:
**Describe the bug**
No firewood metrics are being reported to prometheus in the latest version of avalanchego/firewood combo.

**To Reproduce**
Spin up a bootstrap test, the metrics aren't making it to grafana

**Expected behavior**
I expected the firewood metrics would be enabled.

**Possible Solutions**
This is likely a problem on the coreth side as the metrics initialization has moved.

rkuris:
It seems that StartMetrics is only called when expensive metrics are enabled. I think this may have changed as previously the metrics were initialized on all re-execution tests.

rkuris:
It seems that enabling expensive metrics is still not enough to get metrics reported through our local grafana instance.

rkuris:
Likely related to side effects from [#4418](https://github.com/ava-labs/avalanchego/pull/4418)

rkuris:
There's a rumor that collected metrics from firewood are prefixed as well. If it's true, we need to figure this out and remove any prefixes added by avalanchego.

alarso16:
Closed by ava-labs/avalanchego#4793

PR #1580, opened 2025-12-22
fix(benchmark): align metrics config with new avalanchego task format

Elvis339:
## Why this should be merged

Patch is related to https://github.com/ava-labs/firewood/issues/1578

The `aws-launch.sh` script is currently not collecting metrics because it uses the old avalanchego task interface. The new version of avalanchego's [Taskfile.yml](https://github.com/ava-labs/avalanchego/blob/10f374649645f3093023633d60825875c5645589/Taskfile.yml#L204) changed how metrics configuration is passed:

- Old: `METRICS_ENABLED` passed as a task variable
- New: `METRICS_SERVER_ENABLED` and / or `METRICS_COLLECTOR_ENABLED` passed as environment variables

This mismatch causes metrics to silently not be configured when running benchmarks on EC2 instances.

Related PRs:
- [ava-labs/avalanchego#4761](https://github.com/ava-labs/avalanchego/pull/4761) - [Move from task vars to env](https://github.com/ava-labs/avalanchego/blob/0eebefb80b34849c44c7657a431b57b50e194300/scripts/benchmark_cchain_range.sh#L15)
- [ava-labs/avalanchego#4443](https://github.com/ava-labs/avalanchego/pull/4443)

## How this works

Changed the bootstrap command from:
```bash
task reexecute-cchain-range ... METRICS_ENABLED=false
```

To:
```bash
METRICS_SERVER_ENABLED=true task reexecute-cchain-range ...
```

This aligns with how avalanchego's `benchmark_cchain_range.sh` now reads these settings from the environment rather than from task variables.

### 1. AvalancheGo environment variable behavior

Navigate to or clone AvalancheGo:
```bash
git clone git@github.com:ava-labs/avalanchego.git
cd avalanchego
```

**Default (metrics disabled):**
```bash
./scripts/run_task.sh reexecute-cchain-range \
  BLOCK_DIR=/tmp/test \
  CURRENT_STATE_DIR=/tmp/test \
  START_BLOCK=101 \
  END_BLOCK=102 2>&1 | grep -o '"metrics-server-enabled": "[^"]*"'
```
Expected: `"metrics-server-enabled": "false"`

**With metrics enabled:**
```bash
METRICS_SERVER_ENABLED=true ./scripts/run_task.sh reexecute-cchain-range \
  BLOCK_DIR=/tmp/test \
  CURRENT_STATE_DIR=/tmp/test \
  START_BLOCK=101 \
  END_BLOCK=102 2>&1 | grep -o '"metrics-server-enabled": "[^"]*"'
```
Expected: `"metrics-server-enabled": "true"`

### 2. Firewood `aws-launch.sh` flag behavior

**Default (metrics enabled):**
```bash
./benchmark/bootstrap/aws-launch.sh --dry-run 2>&1 | grep "Metrics Server:"
```
Expected: `Metrics Server: true`

**Disable metrics:**
```bash
./benchmark/bootstrap/aws-launch.sh --dry-run --metrics-server false 2>&1 | grep "Metrics Server:"
```
Expected: `Metrics Server: false`

**Case insensitivity:**
```bash
./benchmark/bootstrap/aws-launch.sh --dry-run --metrics-server TRUE 2>&1 | grep "Metrics Server:"
```
Expected: `Metrics Server: true` (normalized to lowercase)

**Invalid value (fail fast):**
```bash
./benchmark/bootstrap/aws-launch.sh --metrics-server tru 2>&1
```
Expected:
```
Error: Invalid --metrics-server value 'tru'
Valid values: true, false
```

### DIFF
--- a/.github/.golangci.yaml.patch
+++ b/.github/.golangci.yaml.patch
@@ -1,5 +1,5 @@
---- .github/.golangci.yaml	2025-11-26 17:52:36.814736814 +0000
-+++ ffi/.golangci.yaml	2025-11-26 17:52:31.624079933 +0000
+--- .github/.golangci.yaml	2025-12-29 01:55:58
++++ ffi/.golangci.yaml	2025-12-29 00:09:32
 @@ -40,7 +40,7 @@
          - standard
          - default
@@ -9,7 +9,7 @@
          - alias
          - dot
        custom-order: true
-@@ -93,8 +93,6 @@
+@@ -93,12 +93,10 @@
        rules:
          packages:
            deny:
@@ -17,16 +17,27 @@
 -              desc: github.com/ava-labs/avalanchego/utils/linked should be used instead.
              - pkg: github.com/golang/mock/gomock
                desc: go.uber.org/mock/gomock should be used instead.
-             - pkg: github.com/stretchr/testify/assert
-@@ -109,29 +107,10 @@
+-            # Note: testify/assert is checked separately by test_warn_testify_assert
+-            # in lint.sh to produce warnings instead of errors.
++            - pkg: github.com/stretchr/testify/assert
++              desc: github.com/stretchr/testify/require should be used instead.
+             - pkg: io/ioutil
+               desc: io/ioutil is deprecated. Use package io or os instead.
+     errorlint:
+@@ -109,33 +107,10 @@
      forbidigo:
        # Forbid the following identifiers (list of regexp).
        forbid:
+-        - pattern: assert\.Error$(# ErrorIs should be used instead)?
+-        - pattern: assert\.ErrorContains$(# ErrorIs should be used instead)?
+-        - pattern: assert\.EqualValues$(# Equal should be used instead)?
+-        - pattern: assert\.NotEqualValues$(# NotEqual should be used instead)?
 -        - pattern: require\.Error$(# ErrorIs should be used instead)?
 -        - pattern: require\.ErrorContains$(# ErrorIs should be used instead)?
 -        - pattern: require\.EqualValues$(# Equal should be used instead)?
 -        - pattern: require\.NotEqualValues$(# NotEqual should be used instead)?
-         - pattern: ^(t|b|tb|f)\.(Fatal|Fatalf|Error|Errorf)$(# the require library should be used instead)?
+-        - pattern: ^(t|b|tb|f)\.(Fatal|Fatalf|Error|Errorf)$(# the assert and require libraries should be used instead)?
++        - pattern: ^(t|b|tb|f)\.(Fatal|Fatalf|Error|Errorf)$(# the require library should be used instead)?
          - pattern: ^sort\.(Slice|Strings)$(# the slices package should be used instead)?
        # Exclude godoc examples from forbidigo checks.
        exclude-godoc-examples: false
@@ -48,7 +59,7 @@
      revive:
        rules:
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bool-literal-in-expr
-@@ -195,17 +174,6 @@
+@@ -199,17 +174,6 @@
          # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#useless-break
          - name: useless-break
            disabled: false


benchmark/bootstrap/aws-launch.sh
```diff
@@ -454,7 +454,8 @@ runcmd:
   # execute bootstrapping
   - >
     sudo -u ubuntu -D /mnt/nvme/ubuntu/avalanchego --login
-    time task reexecute-cchain-range CURRENT_STATE_DIR=/mnt/nvme/ubuntu/exec-data/current-state BLOCK_DIR=/mnt/nvme/ubuntu/exec-data/blocks START_BLOCK=1 END_BLOCK=__END_BLOCK__ CONFIG=__CONFIG__ METRICS_ENABLED=false
+    METRICS_SERVER_ENABLED=false METRICS_COLLECTOR_ENABLED=false
```
Elvis339: Question for reviewers: Should metrics be configurable via a CLI flag (e.g., `--enable-metrics`)?

Currently I kept metrics disabled (`METRICS_SERVER_ENABLED=false`) to match the previous behavior. To test with metrics enabled, you would need to:

1. Manually edit the script to change `false` to `true`:
    ```bash
    METRICS_SERVER_ENABLED=true METRICS_COLLECTOR_ENABLED=true
    ```

2. Or we could add a `--enable-metrics` flag to the script that sets these environment variables accordingly.

Let me know if you'd like me to add this as a configurable option.

demosdemon: We shouldn't need two different ways to pass parameters to the task. This is why I was saying the parameter needed to be added to the taskfile.

Elvis339: You're right we don't need both. I initially included `METRICS_COLLECTOR_ENABLED` to show it's possible to pass multiple env vars.

The two options serve different purposes:
- `METRICS_SERVER_ENABLED` - Starts a metrics HTTP server that exposes a `/metrics` endpoint
- `METRICS_COLLECTOR_ENABLED` - Controls whether a Prometheus agent is started to collect and forward metrics to a remote Prometheus instance (used by tmpnet and CI workflows)

Since this scripts runs Grafana locally and scrape the metrics endpoint, you only need `METRICS_SERVER_ENABLED`.

As for why these are environment variables instead of task variables, that's just how we designed it https://github.com/ava-labs/avalanchego/pull/4443. Task variables are for the benchmark itself (which blocks, which config), environment variables are for the runtime context (metrics, Prometheus creds, runner type). Either approach would work, this is just convention.

Elvis339:
> I would prefer the default on, not off. Why would someone not want metrics?

Default was false, updated it to be true.

rkuris:
This just needs a little more TLC.

As Brandon mentioned, configuration variables for the script should be handled the same way. This means setting METRICS_SERVER_ENABLED should be just another parameter to the task file.

We could use some way to shut it off via an option when the script is run.

How are invalid values handled? Is the error sane? If I set it to "tru" will it give me a good error message?

This must be tested. Did it work? Please update the PR to show how you tested it. Ideally you tested it with firewood bootstrap.

rkuris:
> This just needs a little more TLC.
> 
> As Brandon mentioned, configuration variables for the script should be handled the same way. This means setting METRICS_SERVER_ENABLED should be just another parameter to the task file.
> 
> We could use some way to shut it off via an option when the script is run.
> 
> How are invalid values handled? Is the error sane? If I set it to "tru" will it give me a good error message?
> 
> This must be tested. Did it work? Please update the PR to show how you tested it. Ideally you tested it with firewood bootstrap.

Summary of offline discussion:

 - There is a move toward environment variables rather than add additional task variables. Code is being refactored anyway in https://github.com/ava-labs/avalanchego/blob/0eebefb80b34849c44c7657a431b57b50e194300/scripts/benchmark_cchain_range.sh#L15
 - Still could use an option to turn it off.
 - Invalid options are ignored, this should be fixed. It's a line or two to validate either true|TRUE|false|FALSE and provide a sane error if none of these are specified, potentially saving some debugging down the road. Fail fast is always better, especially for a tool like this that is run manually during debugging.
 - Some testing was done, I'll do some final acceptance testing once the PR is fully ready.

Elvis339:
@rkuris this commit https://github.com/ava-labs/firewood/pull/1580/commits/189d359ce14953124ed6af6be8e44a825a97a193 

- Added --metrics-server flag to control whether the metrics server is enabled during benchmark runs.
Default is true (metrics enabled)
- Use --metrics-server false to disable
- Validates input and normalizes case (e.g., TRUE → true)
- Rejects invalid values like tru with a clear error

Validation is done here because the underlying AvalancheGo task only recognizes exact true/false - anything else silently disables metrics. This makes the behavior explicit on the consumer (Firewood) rather than letting typos fail silently.

benchmark/bootstrap/aws-launch.sh
```diff
@@ -128,6 +130,17 @@ while [[ $# -gt 0 ]]; do
             CONFIG="$2"
             shift 2
             ;;
+        --metrics-server)
+            # Normalize to lowercase
+            METRICS_SERVER="${2,,}"
+            # Validate boolean value
+            if [[ "$METRICS_SERVER" != "true" && "$METRICS_SERVER" != "false" ]]; then
```
Elvis339, line 137: If the team agrees we can also map with common boolean aliases: `true|t|1|yes|y|on` and `false|f|0|no|n|off`

demosdemon: This is fine. Fancier handling can be implemented in `fwdctl` (#1318).

PR #1584, opened 2025-12-29
fix(ci): update golang-ci patch

AminR443:
## Why this should be merged
The `expected-golangci-yaml-diff` CI workflow is failing because AvalancheGo updated their `.golangci.yml` file, adding new `assert.*` forbidigo patterns and a comment about testify/assert. The existing patch no longer applies cleanly (hunk #3 fails at line 107).

## How this works
Regenerates `.github/.golangci.yaml.patch` against the current upstream AvalancheGo `.golangci.yml` so all hunks apply cleanly.

## How this was tested
Ran `.github/scripts/verify_golangci_yaml_changes.sh` check locally - patch applies successfully and `ffi/.golangci.yaml` matches the expected patched output.

### DIFF
--- a/.github/.golangci.yaml.patch
+++ b/.github/.golangci.yaml.patch
@@ -1,5 +1,5 @@
---- .github/.golangci.yaml	2025-11-26 17:52:36.814736814 +0000
-+++ ffi/.golangci.yaml	2025-11-26 17:52:31.624079933 +0000
+--- .github/.golangci.yaml	2025-12-29 01:55:58
++++ ffi/.golangci.yaml	2025-12-29 00:09:32
 @@ -40,7 +40,7 @@
          - standard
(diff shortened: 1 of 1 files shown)


AminR443:
This doesn't adopt any new changes from AvalancheGo's `.golangci.yaml`. If we want any of those changes, let me know and I'll include it.

demosdemon:
> I'm not sure whether we should deviate from avalanchego here, or add the warning for assert, or just allow assert to be used. Any thoughts on this?

I already hate this synchronizing process and I don't want to track the differences of a second file. If we're going to merge into the monorepo, then I don't want to invest any more time in keeping them in sync since that'll go away after the merge. If we abandon the monorepo idea, then I'll reconsider.

PR #1587, opened 2025-12-29
feat!: Use AWS profile in launch script

alarso16:
## Why this should be merged

Allows easier use of launch script with AWS credentials

## How this works

Specifies profile explicitly

## How this was tested

Locally, using a profile name

### DIFF
--- a/benchmark/bootstrap/aws-launch.sh
+++ b/benchmark/bootstrap/aws-launch.sh
@@ -9,6 +9,7 @@ LIBEVM_BRANCH=""
 NBLOCKS="1m"
 CONFIG="firewood"
 PROFILE=""
+METRICS_SERVER="true"
 REGION="us-west-2"
 DRY_RUN=false
 SPOT_INSTANCE=false
@@ -66,6 +67,7 @@ show_usage() {
     echo "  --nblocks BLOCKS            Number of blocks to download (default: 1m)"
     echo "  --config CONFIG             The VM reexecution config to use (default: firewood)"
     echo "  --profile PROFILE           AWS CLI profile to use (default: AWS default)"
+    echo "  --metrics-server BOOL       Enable metrics server (true/false, default: true)"
     echo "  --region REGION             AWS region (default: us-west-2)"
     echo "  --spot                      Use spot instance pricing (default depends on instance type)"
     echo "  --dry-run                   Show the aws command that would be run without executing it"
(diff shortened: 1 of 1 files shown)


benchmark/bootstrap/README.md
```diff
@@ -9,6 +9,8 @@ Before running the script, you'll need:
 - AWS CLI installed and configured on your machine
 - Authenticated AWS session: `aws sso login`
   - Your session should be configured to use the `Experimental` account.
+  - This must either be the default profile or named and passed to this script as `--profile PROFILE`.
+  - `$AWS_PROFILE` is ignored.
```
demosdemon: I'd rather not make this a true statement. `--profile` should be optional and not override `AWS_PROFILE` unless actually provided on the cli.

alarso16: We can do that, it's just messier. Every aws command will need to be constructed with an optional append.

demosdemon: that's not too messy. we can have a bash array to store common aws command args

```bash
COMMON_AWS_ARGS=()

if test "... condition for --region"; then
    COMMON_AWS_ARGS+=(--region "$REGION")
fi

if test "... condition for --profile"; then
    COMMON_AWS_ARGS+=(--profile "$PROFILE")
fi

# ...

aws "${COMMON_AWS_ARGS[@]}" ... other args
```

alarso16: Ahh done, this is overall better anyway

benchmark/bootstrap/aws-launch.sh
```diff
@@ -506,8 +519,7 @@ fi
 if [ "$DRY_RUN" = true ]; then
     echo "DRY RUN - Would execute the following command:"
     echo ""
-    echo "aws ec2 run-instances \\"
-    echo "  --region \"$REGION\" \\"
+    echo "aws ec2 run-instances ${COMMON_ARGS[@]} \\"
```
demosdemon: nit: this one should be a `*` instead of `@` due to how shell expansions work. Bash will already do the work of the star expansion anyways because it's not a single quoted expansion. But, it's better to be explicit when dealing with array expansions. 

Ref: <https://www.gnu.org/software/bash/manual/html_node/Arrays.html#:~:text=These%20subscripts%20differ%20only%20when%20the%20wordappears%20within%20double%20quotes.If%20the%20word%20is%20double%2Dquoted%2C$%7Bname%5B*%5D%7D%20expands%20to%20a%20single%20word%20withthe%20value%20of%20each%20array%20member%20separated%20by%20the%20first%20character%20of%20theIFS%20variable%2C%20and%20$%7Bname%5B@%5D%7D%20expands%20each%20element%20ofname%20to%20a%20separate%20word.>

PR #1590, opened 2026-01-07
feat(replay): add `firewood-replay` crate for operation recording and replay (1/5)

AminR443:
## Why this should be merged

This PR introduces a new `firewood-replay` crate that enables recording and replaying database operations against Firewood. This is essential for:

- Performance benchmarking with real workloads
- Reproducing specific operation sequences for debugging
- Validating consistent database behavior across code changes

## How this works

The crate defines serializable types for all database operations (reads, batches, proposals, commits) using a `DbOperation` enum. Operations are serialized as length-prefixed MessagePack segments, allowing streaming without loading entire logs into memory. The replay engine (`replay_from_reader`) applies operations against a `Db` instance, tracking proposal IDs across operations and collecting performance metrics.

## How this was tested

Unit tests verify batch inserts, propose/commit workflows, chained proposals, and empty log handling.

## Design Notes
1. **Separate crate decision:** The choice to make this a separate crate versus including it in the existing `ffi` crate was to enable use outside FFI use cases, like new commands in `fwdctl` for benchmarking.
2. **Serialization format:** Initial serialization used `rkyv`, which was zero-copy and didn't introduce much overhead. However, Go can't read from it easily, so I switched to MessagePack. I also considered Cap'n Proto, but it seemed like overkill for our use case.

### DIFF
--- a/replay/Cargo.toml
+++ b/replay/Cargo.toml
@@ -26,4 +26,4 @@ tempfile.workspace = true
 workspace = true
 
 [package.metadata.cargo-machete]
-ignored = ["metrics"]
+ignored = ["metrics", "serde_bytes"]


demosdemon:
Adding a new package will require some updates to CI workflows as well. Please insert the package here in a topologically correct position: <https://github.com/ava-labs/firewood/blob/b8fc6677dca8002302a6573d8aac19d391038375/.github/workflows/publish.yaml#L9-L53>

Cargo.toml
```diff
@@ -63,6 +64,7 @@ firewood-ffi = { path = "ffi", version = "0.0.18" }
 
 # no longer bumping with workspace
 firewood-triehash = { path = "triehash", version = "0.0.16" }
+firewood-replay = { path = "replay", version = "0.0.18" }
```
demosdemon: move this above the comment

replay/src/lib.rs
```diff
@@ -0,0 +1,556 @@
...
+use thiserror::Error;
+
+/// Serde helper for `Option<Box<[u8]>>` using efficient byte representation.
+mod option_bytes {
+    use serde::{Deserialize, Deserializer, Serializer};
+
+    #[expect(clippy::ref_option, reason = "serde requires &T for serialize")]
+    pub fn serialize<S: Serializer>(value: &Option<Box<[u8]>>, ser: S) -> Result<S::Ok, S::Error> {
+        match value {
+            Some(bytes) => ser.serialize_some(&serde_bytes::Bytes::new(bytes)),
```
demosdemon: `serde_with` may be more versatile here:  
 
- <https://docs.rs/serde_with/latest/serde_with/struct.Bytes.html>

Which you can combine with `Option<_>` using

```rust
#[serde_as]
#[derive(/* ..., */ Serialize, Deserialize)]
pub struct KeyValueOp {
// ...
    #[serde_as(as = "Option<serde_with::Bytes>")]
    pub value: Option<Box<[u8]>>,
// ...
}
```

demosdemon (approved):
LGTM after other comments.

replay/src/lib.rs
```diff
@@ -0,0 +1,556 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
```
demosdemon: Oh, and copyright year. New files should try to get the year in which they were committed.

AminR443: Fixed. Sorry the code was from last year.

replay/src/lib.rs
AminR443: Good idea. Thanks.

PR #1591, opened 2026-01-07
feat(ffi/replay): add `block-replay` feature for recording FFI operations (2/5)

AminR443:
## Why this should be merged

Enables recording of FFI database operations for performance analysis and replay benchmarking. The feature is opt-in, only active when the `block-replay` feature is enabled and `FIREWOOD_BLOCK_REPLAY_PATH` env var is set. It's zero-cost when disabled, as feature-gated code paths are compiled out.

## How this works

This PR adds a `block-replay` Cargo feature that optionally depends on `firewood-replay`. A new `ffi/src/replay.rs` module provides a global `Recorder` behind `OnceLock<Option<Mutex<Recorder>>>`.

FFI entry points (`fwd_get_latest`, `fwd_propose_on_db`, `fwd_commit_proposal`, etc.) are instrumented to record operations when the feature is active. Operations are buffered and periodically flushed as length-prefixed MessagePack segments to the path specified in `FIREWOOD_BLOCK_REPLAY_PATH`. Manual flushing is available via `fwd_block_replay_flush()`, with auto-flush on `fwd_close_db()`.

Recorded operation types: `GetLatest`, `GetFromRoot`, `GetFromProposal`, `Batch`, `ProposeOnDB`, `ProposeOnProposal`, `Commit`.

## How this was tested

Compiled with the feature enabled and verified replay log generation via the environment variable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1256,6 +1256,7 @@ dependencies = [
  "log",
  "lru",
  "metrics",
+ "metrics-util",
  "nonzero_ext",
  "parking_lot",
  "pprof",
--- a/ffi/flake.lock
+++ b/ffi/flake.lock
@@ -39,33 +39,35 @@
       },
       "locked": {
         "dir": "nix/go",
-        "lastModified": 1760973838,
-        "narHash": "sha256-UnngvRB45lUeWwot7cvB0MaedaQEQmcw+q8Y6WbeGtE=",
+        "lastModified": 1767633935,
+        "narHash": "sha256-SzWj4hzLTPHrPct8jejMIf5pB4DIOxe+PMpXDPKaDg8=",
         "owner": "ava-labs",
         "repo": "avalanchego",
-        "rev": "f10757d594eedf0f016bc1400739788c542f005f",
+        "rev": "1a59a6f646ef18167dde10d24d4efbf05e3de177",
         "type": "github"
       },
       "original": {
         "dir": "nix/go",
         "owner": "ava-labs",
-        "ref": "f10757d594eedf0f016bc1400739788c542f005f",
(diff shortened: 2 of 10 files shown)


ffi/src/lib.rs
```diff
@@ -462,7 +482,17 @@ pub unsafe extern "C" fn fwd_propose_on_db<'db>(
     db: Option<&'db DatabaseHandle>,
     values: BorrowedKeyValuePairs<'_>,
 ) -> ProposalResult<'db> {
-    invoke_with_handle(db, move |db| db.create_proposal_handle(values))
+    #[cfg(feature = "block-replay")]
+    let values_for_log = values;
```
demosdemon: I see why you need this, `BorrowedSlice` improperly implements Copy. This is an annoying caveat `derive` with generics. The auto-derived code yields:

```rust
impl<T: Clone> Clone for BorrowedSlice<'_, T> {
    fn clone(&self) -> Self {
        Self {
            ptr: self.ptr.clone(),
            len: self.len.clone(),
            marker: self.marker.clone(),
        }
    }
}

impl<T: Copy> Copy for BorrowedSlice<'_, T> {}
```

Which is incorrect once you see what it is doing. We need the following manual implementation:

```rust
impl<T> Clone for BorrowedSlice<'_, T> {
    fn clone(&self) -> Self {
        *self
    }
}

impl<T> Copy for BorrowedSlice<'_, T> {}
```

I honestly would prefer we fix `BorrowedSlice` and manually implement `Clone` and `Copy` for all `T` so that you can delete this pull out.

AminR443: Fixed. Thanks for pointing it out.

PR #1592, opened 2026-01-07
feat(ffi/replay): add Go replay log decoder and execution tests (3/5)

AminR443:
## Why this should be merged

Adds Go-side code to decode and replay Firewood operation logs recorded by the Rust `block-replay` feature. This enables performance benchmarking of real workloads captured from production and deterministic replay testing.

## How this works

Replay log types are Go structs mirroring the Rust `ReplayLog` and `DbOperation` enums, using msgpack for deserialization. Read ops include `GetLatest`, `GetFromProposal`, and `GetFromRoot`. Write ops include `Batch`, `ProposeOnDB`, `ProposeOnProposal`, and `Commit`.

`applyReplayLogs()` replays decoded operations against a `Database`, tracking proposals by ID and optionally verifying commit hashes.

## How this was tested

- **TestBlockReplayRoundTrip** — Records operations via env var, flushes, replays to a fresh database, asserts root hashes match
- **TestReplayLogExecution** — Replays an external log file (set `REPLAY_LOG` env var); skips if not set
- **BenchmarkReplayLog** — Benchmarks replay execution.


demosdemon:
Pinging @alarso16 for second look at the Go code

ffi/replay_test.go
```diff
@@ -0,0 +1,412 @@
...
+)
+
+// ReplayLog mirrors the Rust ReplayLog type.
+type ReplayLog struct {
+	Operations []DbOperation `msgpack:"operations"`
+}
+
+// DbOperation is the Go representation of the Rust DbOperation enum.
+// msgpack will populate exactly one of the pointer fields below.
+type DbOperation struct {
```
alarso16:
```suggestion
type DBOperation struct {
```
Kind of surprised that the linter didn't yell at you. Alternatively, are there other kinds of operations? Maybe `ReplayOperation` unless there's going to be more

ffi/replay_test.go
```diff
@@ -0,0 +1,412 @@
...
+	GetFromProposal   *GetFromProposal   `msgpack:"GetFromProposal,omitempty"`
+	GetFromRoot       *GetFromRoot       `msgpack:"GetFromRoot,omitempty"`
+	Batch             *Batch             `msgpack:"Batch,omitempty"`
+	ProposeOnDB       *ProposeOnDB       `msgpack:"ProposeOnDB,omitempty"`
+	ProposeOnProposal *ProposeOnProposal `msgpack:"ProposeOnProposal,omitempty"`
+	Commit            *Commit            `msgpack:"Commit,omitempty"`
+}
+
+// GetLatest represents a read from the latest revision.
+type GetLatest struct {
```
alarso16: None of these structs are actually exported, so maybe it would be better to not capitalize everything. I'm not sure if this is exported (maybe Go does something to prevent it because it's a test file?)

ffi/replay_test.go
```diff
@@ -0,0 +1,412 @@
...
+	r.NoError(err, "replay logs against database")
+	elapsed := time.Since(start)
+
+	root, err := db.Root()
+	r.NoError(err, "get root after replay")
+	r.NotEqual(EmptyRoot, root, "root should not be EmptyRoot after replay")
+
+	t.Logf("Replay completed in %v (%d commits), final root: %x", elapsed, commits, root)
+
+	r.NoError(db.Close(oneSecCtx(t)))
```
alarso16: Doesn't this happen automatically with the database open?

ffi/firewood.go
```diff
@@ -484,3 +484,12 @@ func (db *Database) Dump() (string, error) {
 
 	return string(bytes), nil
 }
+
+// FlushBlockReplay flushes buffered block replay operations to disk.
+//
+// This function is only meaningful when the FFI library was compiled with
+// the `block-replay` feature and the `FIREWOOD_BLOCK_REPLAY_PATH` environment
+// variable is set. Otherwise, it is a no-op.
+func FlushBlockReplay() error {
```
alarso16, line 507: This is global? So if multiple databases are instantiated in the same thread, then everything will be jumbled up?

ffi/replay_test.go
```diff
@@ -0,0 +1,412 @@
...
+// Returns the number of commits applied and any error encountered.
+func applyReplayLogs(db *Database, logs []ReplayLog, cfg ReplayConfig) (int, error) {
+	proposals := make(map[uint64]*Proposal)
+	totalCommits := 0
+
+	for _, segment := range logs {
+		for _, op := range segment.Operations {
+			switch {
+			case op.GetLatest != nil:
+				// Read operations - errors are non-fatal during replay
```
alarso16: Why are they "non-fatal"?

ffi/replay_test.go
```diff
@@ -0,0 +1,412 @@
...
+type ReplayConfig struct {
+	// MaxCommits limits the number of commits to replay. 0 means unlimited.
+	MaxCommits int
+	// VerifyHashes enables verification of returned hashes after commits.
+	VerifyHashes bool
+}
+
+// applyReplayLogs applies replay logs to a database.
+// Returns the number of commits applied and any error encountered.
+func applyReplayLogs(db *Database, logs []ReplayLog, cfg ReplayConfig) (int, error) {
```
alarso16: Might be easier if you pass in `t` and fail

ffi/replay_test.go
RodrigoVillar: AFAIK, these types won't be exported. However, I agree that they should still be lowercased to make it clear to readers that these types are used only for testing.

ffi/firewood.go
AminR443: Thanks for noting this out. My initial code tracked this and now I realized that it got lost when simplifying it for experiments. I'll create an issue and a follow up PR to resolve this in both Rust and go, so we have database id like proposal id.

ffi/replay_test.go
AminR443: I prefer to keep the helper functions independent of testing. makes them easier to reuse/copy in future cases.

ffi/replay_test.go
AminR443: will fix.

ffi/firewood.go
alarso16: If it's debug-only, might not be worth all the effort. Just documenting (at least until Firewood officially supports multiple databases)

ffi/replay_test.go
AminR443: Done.

ffi/replay_test.go
AminR443: Done.

ffi/firewood.go
AminR443: Documented.

ffi/replay_test.go
AminR443: Resolved. remained from a previous experiment.

ffi/replay_test.go
AminR443: You're right, done.

PR #1594, opened 2026-01-07
feat(replay): performance action (5/5) 

AminR443:

### DIFF
--- a/.github/scripts/update_benchmark_reports.py
+++ b/.github/scripts/update_benchmark_reports.py
@@ -102,12 +102,42 @@ def apply_theme(report_dir, css):
         )
 
 
+def fix_base_commit_ids(path, base_sha, base_subject, head_sha):
+    """Fix commit IDs in comparison data where benchmark-action recorded the
+    base benchmark under the HEAD SHA (it ignores the ref input in PR context)."""
+    data = load_data(path)
+    if not data or not base_sha or not head_sha or base_sha == head_sha:
+        return
+    changed = False
+    for suites in (data.get("entries") or {}).values():
+        if not isinstance(suites, list) or len(suites) < 2:
+            continue
+        # Sort by date to find the earliest entry — that's the base benchmark
+        dated = [(s.get("date", 0), s) for s in suites if isinstance(s, dict)]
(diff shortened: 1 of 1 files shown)


JuanLeon2:
I love this.  It is too bad that per Rodrigo the benchmark action can only do one dimension (e.g, time), but super duper anyway.

Some comments:
1. Does it require write access to contents?  (Is this needed for the push to the gh-pages-based graph?)
2. Can we add the permissions on the jobs that require them, rather then the whole workflow?
3. It uses your personal repo.  Can we use context.repo.repo instead?
4. Large binary files are not kind to github; shoudl we put the reply log on S3, perhaps even the same volume that Aaron uses for the replay logs?
5. We have been limiting what is on GH actions to things that don't make sense to do/cannot be done from the command line.  In Go, we also use "task" (equivalent is "just" in Rust) to serve as both a dictionary of things that are useful and as specification of "golden paths" (e.g, sets of parameters that make sense).   Should the build and run steps here follow that pattern?
6. Should we add an "extra" input whose value we can change when we make changes to the becnhmark that result in discontinuities (e.g, if we replace one metric with another, the historical perf graph might show a jump -- the text in the "extra" input appears in the tooltip, and would be helpful to someone looking at the trend).
7. Should we split this action into two: one contains the trigger and the  response to the PR/comment; the other one is a composabel action that builds and executes the benchmark, and uses github-benchmark-action.   The latter would have the name of the benchmark as an input, hence it could be called from longer-running benchmarks that we might execute overnight, or on merge e.g

AminR443:
Thanks for your comment. 

- For 1 and 2, we can change those and I haven't yet cleaned it up as it's a WIP. 
- About 3, it's for deploying the gh-pages, last I remembered we can't use that in our repo because rust docs depend on it? I'll double check but I think we discussed it in one of standups, mentioned by Elvis I suppose.
- (4) Definitely we should go with S3. This was just a demonstration and I took a shortcut to integrate S3 after it looks good. 
- (5) I agree. I was planning to add some commands in justfile we have, and use them. We can take a look to our existing workflows and refactor them too.
- (6, 7) Both great ideas, we can discuss further in a call and I'll make sure to add them.

PR #1595, opened 2026-01-07
feat: add `FlushAndSync()`

### LINKED ISSUE #1575: Provide an API to force the current revision to disk
In an attempt to freeze the API, we will need some method to force the current version of firewood to disk without closing the database. This can be used by coreth to force syncs at a specific blockno%INTERVAL. Since firewood has no access to the block number, we'd need a way for coreth to tell this to firewood.

This is now a no-op, as all revisions are forced to disk, so this just adds a stub that can be integrated into coreth. We should name the go method PersistNow()

We could add a breaking change to the API later, or force the database to be closed and reopened, but that's way more tricky.

RodrigoVillar:
## Why this should be merged

Closes #1575.

## How this works

Adds a `PersistNow()` method to both the Go and FFI library. This method is currently a no-op (this is documented in the code).

## How this was tested

Added a unit test to verify that `db.PersistNow()` doesn't fail.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -334,7 +334,7 @@ impl Db {
 
     /// Forces the database to flush and sync the revision associated with root
     /// to disk.
-    #[allow(clippy::missing_const_for_fn)]
+    #[expect(clippy::missing_const_for_fn)]
     pub fn flush_and_sync_root(&self, _root: HashKey) -> Result<(), api::Error> {
         Ok(())
     }


firewood/src/db.rs
```diff
@@ -323,6 +323,12 @@ impl Db {
         let merge_ops = merkle.merge_key_value_range(first_key, last_key, key_values);
         self.propose_with_parent(merge_ops, merkle.nodestore())
     }
+
+    /// Forces the database to persist the current state to disk.
+    #[allow(clippy::missing_const_for_fn)]
```
RodrigoVillar: Ignoring the `clippy` warning for now means that we don't have to remove the `const` (and therefore, modify the API) once this is no longer a no-op.

ffi/firewood.go
```diff
@@ -399,6 +399,17 @@ func (db *Database) LatestRevision() (*Revision, error) {
 	return db.Revision(root)
 }
 
+// PersistNow forces the database to persist the current state to disk.
+// This method can be used to force syncs at specific intervals.
```
alarso16: Will you add more details about how this works? Like the state will be around forever?

alarso16: Talked in the office - the details are not clear yet, so we can leave this for now

alarso16 (approved):

demosdemon (changes requested):
The api should require the caller to specify the root they want to persist; which may or may not be the tip revision.

I'm also not sure about the name. We should probably call it something like `FlushRoot` and/or `FlushAndSyncRoot` where `Flush` means we write the nodes to the filesystem and `Sync` means we wait for the kernel to sync the nodes to the block device (`fsync`).

RodrigoVillar:
> The api should require the caller to specify the root they want to persist; which may or may not be the tip revision.
> 
> I'm also not sure about the name. We should probably call it something like `FlushRoot` and/or `FlushAndSyncRoot` where `Flush` means we write the nodes to the filesystem and `Sync` means we wait for the kernel to sync the nodes to the block device (`fsync`).

On board with renaming `PersistNow()` and specifying which root to persist - was following what #1575 outlined.

demosdemon:
> On board with renaming `PersistNow()` and specifying which root to persist - was following what #1575 outlined.

Understood! I just had some real-time thoughts about how the API should look before we freeze it. And, for documentation: providing a root hash will flush that revision as well as any preceding revision to disk. If the revision is already flushed to disk, it is a no-op. Unless the `Sync` variant is used which case we will always make the `fsync` syscall even if we did not write any nodes.

ffi/firewood.go
demosdemon: I think I can clarify this a bit. This API is needed to service "delayed persist" where the caller can commit a revision but it may not actually be written to disk. Right now, delayed persist isn't implemented so that's why this method is a no-op. But, when it is implemented this will allow the caller to manually flush a revision that may or may not be persisted.

Additionally and similar to delayed persist, we want the ability to 1) flush a known revision to disk; 2) signal to stop writing to the file; 3) copy the file; and then 4) signal to start writing to the file, so that we can atomically copy the database while running.

firewood/src/db.rs
demosdemon: nit: use `expect` so that the expectation itself becomes a lint violation when it is no longer suppressing any warnings.

```suggestion
    #[expect(clippy::missing_const_for_fn)]
```

RodrigoVillar: Done: 5a64c78

PR #1597, opened 2026-01-08
feat: CodeHashes() for range proofs

### LINKED ISSUE #1157: Range Proofs: eth compat needs ability to get code hashes from applied proof

alarso16:
## Why this should be merged

To enable syncing all Ethereum state, non-empty code hashes must be provided from each proof.

Closes #1157

## How this works

Iterates through all keys in the range proof, and if the length is 32 (account node), RLP decodes to find the code hash, and returns from the iterator. This is hidden behind the `ethhash` config option. Otherwise, an error is returned.

## How this was tested

UT in Go, and a Fuzz test in the `ethhash` case

### DIFF
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -56,9 +56,15 @@ static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 /// information.
 #[inline]
 fn invoke<T: CResult, V: Into<T>>(once: impl FnOnce() -> V) -> T {
+    #[cfg(panic = "unwind")]
     match std::panic::catch_unwind(std::panic::AssertUnwindSafe(once)) {
         Ok(result) => result.into(),
         Err(panic) => T::from_panic(panic),
+    }
+
+    #[cfg(not(panic = "unwind"))]
+    {
+        once().into()
     }
 }
 
@@ -107,10 +113,7 @@ fn invoke_with_handle<H, T: NullHandleResult, V: Into<T>>(
 ///
 /// [`BorrowedBytes`]: crate::value::BorrowedBytes
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_get_latest(
-    db: Option<&DatabaseHandle>,
-    key: BorrowedBytes,
-) -> ValueResult {
+pub extern "C" fn fwd_get_latest(db: Option<&DatabaseHandle>, key: BorrowedBytes) -> ValueResult {
     #[cfg(feature = "block-replay")]
     if db.is_some() {
         replay::record_get_latest(key);
@@ -140,7 +143,7 @@ pub unsafe extern "C" fn fwd_get_latest(
 /// * call [`fwd_free_iterator`] to free the memory associated with the iterator.
 ///
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_iter_on_revision<'view>(
+pub extern "C" fn fwd_iter_on_revision<'view>(
     revision: Option<&'view RevisionHandle>,
     key: BorrowedBytes,
 ) -> IteratorResult<'view> {
@@ -169,7 +172,7 @@ pub unsafe extern "C" fn fwd_iter_on_revision<'view>(
 /// * call [`fwd_free_iterator`] to free the memory associated with the iterator.
 ///
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_iter_on_proposal<'p>(
+pub extern "C" fn fwd_iter_on_proposal<'p>(
     handle: Option<&'p ProposalHandle<'_>>,
     key: BorrowedBytes,
 ) -> IteratorResult<'p> {
@@ -203,7 +206,7 @@ pub unsafe extern "C" fn fwd_iter_on_proposal<'p>(
 ///   to free the memory associated with the returned value.
 ///
 #[unsafe(no_mangle)]
-pub unsafe extern "C" fn fwd_iter_next(handle: Option<&mut IteratorHandle<'_>>) -> KeyValueResult {
+pub extern "C" fn fwd_iter_next(handle: Option<&mut IteratorHandle<'_>>) -> KeyValueResult {
     invoke_with_handle(handle, Iterator::next)
 }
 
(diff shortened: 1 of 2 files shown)


ffi/src/proofs/change.rs
```diff
@@ -80,6 +89,91 @@ pub struct NextKeyRange {
     pub end_key: Maybe<OwnedBytes>,
 }
 
+#[derive(Debug)]
+#[cfg_attr(not(feature = "ethhash"), allow(dead_code))]
+pub struct CodeIteratorHandle<'a> {
+    key_values: &'a KeyValueBox,
+    i: usize,
```
demosdemon: Manually indexing into the slice like this is an anti-pattern.

These two fields should be replaced with `std::slice::Iter<'a, KeyValueBox>`

demosdemon:
Relatively minor feedback, but it was too much to put in a comment so I created a commit for it: https://github.com/ava-labs/firewood/pull/1602

- Manually indexing and incrementing the counter is an anti-pattern and wrapping the built-in Iter is preferred
- The `void` field is [uninhabitable](https://doc.rust-lang.org/nomicon/exotic-sizes.html#empty-types) (not even "zero-sized") which prevents the compiler from even emitting the struct during compilation (and all associated code as well).

demosdemon (approved):

ffi/proofs.go
```diff
@@ -228,10 +232,39 @@ func (p *RangeProof) FindNextKey() (*NextKeyRange, error) {
...
+	return func(yield func(Hash, error) bool) {
+		iter, err := getCodeHashIteratorFromCodeHashIteratorResult(C.fwd_range_proof_code_hash_iter(p.handle))
+		if err != nil {
+			yield(EmptyRoot, err)
+			return
+		}
+		defer func() {
+			if err := iter.Free(); err != nil {
+				panic(err)
+			}
```
AminR443, line 244: Is use of panic ideal/necessary here? we usually don't do that on other cleanup/free errors, do we?

ffi/proofs.go
```diff
@@ -505,6 +538,19 @@ func getNextKeyRangeFromNextKeyRangeResult(result C.NextKeyRangeResult) (*NextKe
 	}
 }
 
+func getCodeHashIteratorFromCodeHashIteratorResult(result C.CodeIteratorResult) (*codeIterator, error) {
+	switch result.tag {
+	case C.CodeIteratorResult_Ok:
+		ptr := *(**C.CodeIteratorHandle)(unsafe.Pointer(&result.anon0))
+		return &codeIterator{handle: ptr}, nil
```
AminR443, line 546: This misses a case for `C.CodeIteratorResult_NullHandlePointer`.

ffi/proofs_test.go
```diff
@@ -271,6 +272,37 @@ func TestRangeProofFindNextKey(t *testing.T) {
...
+	r.NoError(err)
+	codeHash := stringToHash(t, "044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d")
+
+	root, err := db.Update([][]byte{key[:]}, [][]byte{val})
+	r.NoError(err)
+
+	proof := newVerifiedRangeProof(t, db, root, nothing(), nothing(), rangeProofLenUnbounded)
+
+	i := 0
+	mode, err := inferHashingMode(t.Context())
```
AminR443, line 291: `TestMain` already does `inferHashingMode`, is it possible to re-use it? Feel free to ignore though, I think Brandon's changes will remove this eventually.

ffi/Cargo.toml
```diff
@@ -31,6 +31,7 @@ oxhttp = "0.3.1"
 tikv-jemallocator = "0.6.1"
 derive-where = "1.6.0"
 metrics-exporter-prometheus = { version = "0.17.2", default-features = false }
+rlp = "0.6.1"
```
AminR443: This can be made optional and only used with ethhash feature. Not important though, making hash a runtime option will change this already.

ffi/src/proofs/change.rs
```diff
@@ -80,6 +96,83 @@ pub struct NextKeyRange {
...
+        #[cfg(feature = "ethhash")]
+        self.inner.find_map(|(key, value)| {
+            if key.len() != 32 {
+                return None;
+            }
+
+            let Ok(code_hash_slice) = Rlp::new(value).at(3).and_then(|r| r.data()) else {
+                return Some(Err(api::Error::ProofError(ProofError::InvalidValueFormat)));
+            };
+            let code_hash: HashKey = TrieHash::try_from(code_hash_slice).ok()?.into();
```
AminR443, line 129: `ok()?` will skip entries when `code_hash_slice` isn't exactly 32 bytes. Is this expected? or should this be an error.

ffi/src/proofs/range.rs
```diff
@@ -524,6 +528,77 @@ pub extern "C" fn fwd_range_proof_find_next_key(
     crate::invoke_with_handle(proof, RangeProofContext::find_next_key)
 }
 
+/// Returns an iterator over the code hashes contained in the range proof.
+/// The iterator must be freed after use.
+///
+/// Can be called at any time after the proof has been created.
```
AminR443, line 534: The go doc says:
```
// This method can only be called after a successful verification of the proof,
// otherwise an error is returned on the first iteration.
```
Rust doc seems to be the correct one, If I didn't miss sth.

.github/workflows/ci.yaml
```diff
@@ -276,7 +276,8 @@ jobs:
           sudo prlimit --pid $$ --memlock=-1:-1 || echo "prlimit not available, locking kernel memory may fail when setting up io-uring"
           ulimit -Sa
           ulimit -Ha
-          go test -fuzz=. -fuzztime=1m
+          go test -fuzz=FuzzFirewoodTree -fuzztime=1m
+          go test -fuzz=FuzzRangeProofCodeHashes -fuzztime=1m
```
AminR443: This should be, rn it doesn't run and skips because of the wrong name:

```suggestion
          go test -fuzz=FuzzRangeProofCreation -fuzztime=1m
```

ffi/proofs.go
alarso16: The way that the `iter.Seq2` works is really interesting:
```
for hash, err := range p.CodeHashes() {
    if ok := internal(hash, err); !ok {
          break
    }
}
```

Is essentially syntactic sugar for:
```
iter := p.CodeHashes()
iter(internal)
```

However, can we trust the user to explicitly end the for loop? What if there's a panic in a test? We can still attempt to free it. Additionally, if a user calls `break`, then we don't have the opportunity to return an error to them

ffi/proofs_test.go
alarso16: The way to fix it is Brandon's changes, so I'll leave it until then. If he beats me, then I'll change it

ffi/src/proofs/change.rs
alarso16: This is ok. We expect there to also be 64 byte entries, but we are looking for all 32 byte entries

PR #1602, opened 2026-01-09
chore: pr feedback

demosdemon:
Relatively minor feedback for #1597, but it was too much to put in a comment so I created a commit for it.

- Manually indexing and incrementing the counter is an anti-pattern and wrapping the built-in Iter is preferred
- The void field is [uninhabitable](https://doc.rust-lang.org/nomicon/exotic-sizes.html#empty-types) (not even "zero-sized") which prevents the compiler from even emitting the struct during compilation (and all associated code as well).

### DIFF
--- a/ffi/src/proofs/change.rs
+++ b/ffi/src/proofs/change.rs
@@ -15,6 +15,13 @@ use crate::{
     NextKeyRangeResult, OwnedBytes, ValueResult, VoidResult,
 };
 
+#[cfg(feature = "ethhash")]
+const EMPTY_CODE_HASH: [u8; 32] = [
+    // "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
+    0xc5, 0xd2, 0x46, 0x01, 0x86, 0xf7, 0x23, 0x3c, 0x92, 0x7e, 0x7d, 0xb2, 0xdc, 0xc7, 0x03, 0xc0,
+    0xe5, 0x00, 0xb6, 0x53, 0xca, 0x82, 0x27, 0x3b, 0x7b, 0xfa, 0xd8, 0x04, 0x5d, 0x85, 0xa4, 0x70,
+];
+
 /// Arguments for creating a change proof.
 #[derive(Debug)]
 #[repr(C)]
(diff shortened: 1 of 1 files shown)


ffi/src/proofs/change.rs
```diff
@@ -90,52 +97,42 @@ pub struct NextKeyRange {
 }
 
 #[derive(Debug)]
-#[cfg_attr(not(feature = "ethhash"), allow(dead_code))]
+#[non_exhaustive]
```
alarso16, line 100: What's the point of this?

ffi/src/proofs/change.rs
```diff
@@ -90,52 +97,42 @@ pub struct NextKeyRange {
...
 
 impl Iterator for CodeIteratorHandle<'_> {
     type Item = Result<HashKey, api::Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
         #[cfg(not(feature = "ethhash"))]
-        {
-            None
-        }
+        match self.void {}
```
alarso16, line 118: What is this line doing? I just don't understand they syntax

ffi/src/proofs/change.rs
```diff
@@ -90,52 +97,42 @@ pub struct NextKeyRange {
...
 pub struct CodeIteratorHandle<'a> {
-    key_values: &'a KeyValueBox,
-    i: usize,
+    #[cfg(feature = "ethhash")]
+    inner: std::slice::Iter<'a, KeyValuePair>,
+    // uninhabitable fields make the struct impossible to construct when the feature is disabled
+    #[cfg(not(feature = "ethhash"))]
+    void: std::convert::Infallible,
+    #[cfg(not(feature = "ethhash"))]
+    marker: std::marker::PhantomData<&'a ()>,
```
alarso16, line 108: This name isn't referenced anywhere, why isn't the compiler angry about it?

alarso16 (approved):

ffi/src/proofs/change.rs
demosdemon: This is technically not needed. I added it before I added the `Infallible` field. An empty struct decorated with `#[non_exhaustive]` cannot be created outside of the crate; which wasn't a strong enough variant.

ffi/src/proofs/change.rs
demosdemon: See the documentation for [uninhabitable types](https://doc.rust-lang.org/nomicon/exotic-sizes.html#empty-types), and [irrefutable patterns](https://doc.rust-lang.org/reference/patterns.html#refutability).

This is a combination of both concepts; attempting to match on the uninhabitable enum yields no branches. This expression's return type is `!` (aka the [never](https://doc.rust-lang.org/std/primitive.never.html) type).

You can see this in action in the source for [`Infallible`](https://doc.rust-lang.org/stable/src/core/convert/mod.rs.html#936) and even here: <https://github.com/ava-labs/firewood/blob/1ef907e789f86186ac84a8f5484452d8c0967446/firewood/src/v2/api.rs#L177-L181>

ffi/src/proofs/change.rs
demosdemon: PhantomData is a special type. The compiler doesn't complain because it knows only the compiler cares about it.

PR #1604, opened 2026-01-09
feat: Export RevisionNotFound error

alarso16:
## Why this should be merged

In xsync, the "error we couldn't form a range proof" and  the "error we couldn't find the state root you wanted", since one is logged as an error, and the other is dropped. This distinction seems helpful for other contexts as well.

this isn't strictly necessary, but will be used to prevent DoS warning logs

## How this works

Exports an error

## How this was tested

No need


alarso16:
Closing for now, this part of x/sync may be re-worked

Issue #1564, opened 2025-12-17
DB Open Options: Hashing Mode


demosdemon:
Before we make hashing mode a runtime option, we need callers to specify which hashing mode they require when opening the database.

At the beginning, we will compare this option with the compile time option and return and error if they do not match. After #1088, this option will be used to decide which hashing mode the database uses.

PR #1608, opened 2026-01-12
feat!: Make `NodeHashAlgorithm` a required option on storage

demosdemon:
## Why this should be merged

In order to facilitate specifying which method of hashing nodes a runtime option, we must first expose the option to callers.

## How this works

- Adds the `NodeHashAlgorithm` enum; which can be expanded to include more options in the future.
- Adds the `node_hash_algorithm` to `ReadableStorage` which ensures that it will be available to anyone that may need to hash a node.
- Adds the parameter to MemStore and FileBacked
- Adds the parameter to DbConfig and ConfigManager
- The default for only tests was added to lessen the number of changes needed

I originally added it to only the `NodeStore`; but, the changes for that were surprisingly more invasive. Additionally, I later realized that long-term I will need this on the storage layer to make it easier to share to everyone that may need to hash a node.

## How this was tested

CI and UT

---

This is a continuation of #1568

Co-authored-by: Ron Kuris <ron.kuris@avalabs.org>

Closes: #1568
Closes: #1564

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -49,6 +49,17 @@ fn is_valid_key(key: &Path) -> bool {
     key.0.len().is_multiple_of(2)
 }
 
+#[expect(clippy::result_large_err)]
+const fn check_area_aligned(
+    address: LinearAddress,
+    parent: StoredAreaParent,
+) -> Result<(), CheckerError> {
+    if !address.is_aligned() {
+        return Err(CheckerError::AreaMisaligned { address, parent });
+    }
+    Ok(())
+}
+
 /// Options for the checker
 #[derive(Debug)]
 pub struct CheckOpt {
@@ -155,7 +166,6 @@ struct SubTrieMetadata {
 }
 
 /// [`NodeStore`] checker
-#[expect(clippy::result_large_err)]
 impl<T, S: ReadableStorage> NodeStore<T, S>
 where
     NodeStore<T, S>: HashedNodeReader,
@@ -282,7 +292,7 @@ where
         } = subtrie;
 
         // check that address is aligned
-        self.check_area_aligned(subtrie_root_address, StoredAreaParent::TrieNode(parent))?;
+        check_area_aligned(subtrie_root_address, StoredAreaParent::TrieNode(parent))?;
 
         // read the node from the disk - we avoid cache since we will never visit the same node twice
         let (area_index, area_size) =
@@ -498,7 +508,7 @@ where
             };
 
             // check that the area is aligned
-            if let Err(e) = self.check_area_aligned(addr, StoredAreaParent::FreeList(parent)) {
+            if let Err(e) = check_area_aligned(addr, StoredAreaParent::FreeList(parent)) {
                 errors.push(e);
                 free_list_iter.move_to_next_free_list();
                 continue;
@@ -551,17 +561,6 @@ where
             },
             errors,
         )
-    }
-
-    const fn check_area_aligned(
-        &self,
-        address: LinearAddress,
-        parent: StoredAreaParent,
-    ) -> Result<(), CheckerError> {
-        if !address.is_aligned() {
-            return Err(CheckerError::AreaMisaligned { address, parent });
-        }
-        Ok(())
     }
 }
 


firewood/src/manager.rs
```diff
@@ -55,6 +55,11 @@ pub struct RevisionManagerConfig {
 #[non_exhaustive]
 /// Configuration manager that contains both truncate and revision manager config
 pub struct ConfigManager {
+    /// The directory where the database files will be stored (required).
+    pub root_dir: PathBuf,
+    /// The algorithm used for hashing nodes (required).
+    pub node_hash_algorithm: NodeHashAlgorithm,
```
RodrigoVillar, line 61: I'm not sure how I feel about this. From what I've seen in our Rust code, the builder pattern is usually reserved for optional fields that have defaults. Given that both `root_dir` and `node_hash_algorithm` are required fields, I would think it's easier to extract out these fields (similar to what we already do for `Db`).

storage/src/linear/memory.rs
```diff
@@ -11,22 +11,34 @@
...
+
+#[cfg(any(test, feature = "test_utils"))]
+impl Default for MemStore {
+    fn default() -> Self {
+        Self {
+            bytes: Mutex::new(Vec::new()),
+            node_hash_algorithm: NodeHashAlgorithm::compile_option(),
+        }
+    }
 }
```
RodrigoVillar, line 33: Nice!

firewood/src/db.rs
```diff
@@ -113,6 +114,9 @@ pub enum UseParallel {
 #[derive(Clone, TypedBuilder, Debug)]
 #[non_exhaustive]
 pub struct DbConfig {
+    /// The algorithm used for hashing nodes (required).
```
RodrigoVillar, line 117: Is this required if we already provide a default for `node_hash_algorithm`?

ffi/src/handle.rs
```diff
@@ -12,6 +12,27 @@ use crate::{BorrowedBytes, CView, CreateProposalResult, KeyValuePair, arc_cache:
 use crate::revision::{GetRevisionResult, RevisionHandle};
 use firewood_storage::firewood_counter;
 
+/// The hashing mode to use for the database.
+///
+/// This determines the cryptographic hash function and trie structure used.
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NodeHashAlgorithm {
```
alarso16, line 20: Why not just `HashAlgorithm`? Also, you note the different here (SHA-256 vs. Keccak-256), but that's not the only difference, right? Ethhash also does all the funky account vs. storage stuff, and really (should) only support keys of length 32 or 64

ffi/firewood.go
```diff
@@ -39,6 +39,16 @@ const RootLength = C.sizeof_HashKey
 // Hash is the type used for all firewood hashes.
 type Hash [RootLength]byte
 
+// NodeHashAlgorithm represents the hashing mode used by the database.
+type NodeHashAlgorithm uint8
```
```suggestion
type HashAlgorithm uint8
```

ffi/firewood.go
```diff
@@ -39,6 +39,16 @@ const RootLength = C.sizeof_HashKey
...
+
+const (
+	// NodeHashAlgorithmMerkleDB uses MerkleDB-compatible SHA-256 based hashing
+	NodeHashAlgorithmMerkleDB NodeHashAlgorithm = C.NodeHashAlgorithm_MerkleDB
+	// NodeHashAlgorithmEthereum uses Ethereum-compatible Keccak-256 based hashing
+	NodeHashAlgorithmEthereum NodeHashAlgorithm = C.NodeHashAlgorithm_Ethereum
```
```suggestion
	// NodeHashAlgorithmMerkleDB uses MerkleDB-compatible SHA-256 based hashing
	MerkleDBHashing NodeHashAlgorithm = C.NodeHashAlgorithm_MerkleDB
	// NodeHashAlgorithmEthereum uses Ethereum-compatible Keccak-256 based hashing
	EthereumHashing NodeHashAlgorithm = C.NodeHashAlgorithm_Ethereum
```

alarso16: These names are all very long (seemingly unnecessarily)

ffi/firewood_test.go
```diff
@@ -180,6 +201,48 @@ func TestUpdateSingleKV(t *testing.T) {
 	r.Equal(vals[0], got)
 }
 
+func TestNodeHashAlgorithmValidation(t *testing.T) {
+	r := require.New(t)
+
+	// Try to create a database with ethhash
+	dbDirEthereum := t.TempDir()
+	dbEthereum, errEthereum := New(dbDirEthereum, NodeHashAlgorithmEthereum, WithTruncate(true))
```
alarso16:
```suggestion
	dbEth, errEth := New(dbDirEthereum, NodeHashAlgorithmEthereum, WithTruncate(true))
```
nit

ffi/firewood_test.go
```diff
@@ -160,8 +161,28 @@ func newTestDatabase(tb testing.TB, opts ...Option) *Database {
...
+var (
+	// detectedNodeHashAlgorithm is cached after first detection to avoid repeated attempts
+	detectedNodeHashAlgorithm     NodeHashAlgorithm
+	detectedNodeHashAlgorithmOnce sync.Once
+)
+
 func newDatabase(dbDir string, opts ...Option) (*Database, error) {
-	f, err := New(dbDir, opts...)
+	// Detect the correct algo once and cache it
+	detectedNodeHashAlgorithmOnce.Do(func() {
```
alarso16, line 172: What might be clearer is moving this to a `test` package and setting this (and the roots) in an init function. Feel free to disagree, I'm not sure it's better

firewood/src/manager.rs
alarso16: That's also the typical perspective in Go

alarso16 (commented):
I left several opinionated nits. The FFI code looks correct though

firewood/src/manager.rs
demosdemon: There's no need for that requirement here. The typed builders we use fail to compile if you do not provide the required arguments. Passing in three variables (db path, algo, and the rest) is excessive when "The rest" struct can carry everything.

firewood/src/db.rs
demosdemon: The default is only for when `cfg(test)` is true. It is required otherwise. The default for tests was added otherwise there would be an additional +20 lines.

ffi/src/handle.rs
demosdemon: There are additional differences to how we hash a node other than the digest algorithm

- The preimage for MerkleDB hashes includes the full node path whereas the Ethereum compatible hashes do not
- The order in which fields are hashed is different
- For ethereum compatible hashes, we manipulate account nodes and inject an subtrie.

The binary choice of Sha256 vs Keccak256 isn't detailed enough. Future NodeHashAlgorithm choices may be added that change some of those expectations without changing the digest algorithm. This controls both the preimage algorithm and the digest algorithm. And, previously Ron and I discussed how this will likely be the way to control branching factors when b256 is added back in because that changes the way we hash nodes (i.e., incompatible ethereum hashes).

ffi/firewood_test.go
demosdemon: This was copied from #1568. I have no opinion. This gives the opportunity for it be lazy checked but that's likely irrelevant since this is a test and we're always checking (... except if you run a single test by name, the init will still run).

ffi/firewood.go
demosdemon: I don't remember what the standard style was for enums. I thought it was `TypeName{Variant}` thus why these are verbose.

ffi/firewood_test.go
alarso16: If you have no opinion, and I don't necessarily think my idea is better, let's just leave it

ffi/firewood.go
alarso16: I can't find anything specific about enums, but I did find this: https://google.github.io/styleguide/go/decisions#variable-name-vs-type

https://github.com/uber-go/guide/blob/master/style.md#start-enums-at-one

ffi/src/handle.rs
alarso16: Fair enough! I'm content

RodrigoVillar (approved):

ffi/firewood_test.go
```diff
@@ -180,6 +201,48 @@ func TestUpdateSingleKV(t *testing.T) {
...
+		result = ethereumOnly
+	case errMerkledb == nil && errEth != nil:
+		result = merkledbOnly
+	}
+
+	switch result {
+	case ethereumOnly:
+		r.NoError(dbEth.Close(oneSecCtx(t)))
+		r.Error(errMerkledb)
+		r.Contains(errMerkledb.Error(), "node store hash algorithm mismatch: want to initialize with MerkleDB, but build option is for Ethereum")
```
alarso16: nit: use `r.ErrorContains`

alarso16 (approved):
Can't trust my rust review too much, not my specialty

Issue #970, opened 2025-06-17
storage: version string improvements


demosdemon:
what a mess. I wonder if we can shorten "firewood" to "fwd" going forward. Sounds like a different PR for someone else to address later though.

_Originally posted by @rkuris in https://github.com/ava-labs/firewood/pull/964#discussion_r2153182751_

demosdemon:
# NodeStoreHeader Expansion Plan for Git Version Information

## Current State Analysis

### Header Structure

The `NodeStoreHeader` struct currently occupies only 264 bytes of the [allocated 2048 bytes][allocation]:

- `version: Version` - 16 bytes (stores "firewood X.Y.Z")
- `endian_test: u64` - 8 bytes
- `size: u64` - 8 bytes
- `free_lists: FreeLists` - 184 bytes (`[Option<LinearAddress>; 23]`)
- `root_address: Option<LinearAddress>` - 8 bytes
- `area_size_hash: [u8; 32]` - 32 bytes
- `ethhash: u64` - 8 bytes
- **Total: 264 bytes out of 2048 available**

### Key Observations

1. **Ample Space Available**: The header has 1784 bytes of unused padding space.

2. **Version Validation**: The current [validation] checks:

   - Valid UTF-8
   - Starts with "firewood " prefix
   - Parseable as semver
   - Meets minimum version requirement (>= 0.0.4)

3. **Serialization**: Uses bytemuck for zero-copy serialization with Pod/Zeroable traits.

4. **Backward Compatibility**: The header location and size (first 2048 bytes) is assumed by multiple components.

## Proposed Solution

### Option 1: Add New Field (Recommended)

Add a new `git_version` field after the existing fields:

```rust
#[derive(Copy, Debug, PartialEq, Eq, Clone, Zeroable, Pod)]
#[repr(C)]
pub struct NodeStoreHeader {
    version: Version,           // 16 bytes
    endian_test: u64,           // 8 bytes
    size: u64,                  // 8 bytes
    free_lists: FreeLists,      // 184 bytes
    root_address: Option<LinearAddress>, // 8 bytes
    area_size_hash: [u8; 32],   // 32 bytes
    ethhash: u64,               // 8 bytes
    git_version: GitVersion,    // NEW: 64 bytes
    // Still 1720 bytes of padding available
}
```

Define GitVersion as:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Zeroable, Pod)]
#[repr(transparent)]
[...]

PR #1611, opened 2026-01-13
feat: add cargo_version and git_describe to storage header

demosdemon:
## Why this should be merged

This expands the database header to include the `CARGO_PKG_VERSION` in a different place as well as the output from `git describe`.

## How this works

This changes magic version to a fixed string, which is more common anyways. The magic version string will be used to track the major database file format as future changes are planned. Two fields were added to the header (plus a third unused field which will hold the root hash).

This is backwards compatible with all current versions of firewood (0.0.4+). The new fields will only be read if the header has the correct magic header. Instead of parsing the semver, a fixed set of valid version strings is checked.

## How this was tested

UT and CI

A few tests reopen a database and the log message below appears:

```text
[2026-01-13T23:56:13Z DEBUG firewood_storage::nodestore::header] Database was created with firewood version v0.0.18-34-gd3fcc5595
```

where `d3fcc5595` is the short representation of the [commit](https://github.com/ava-labs/firewood/pull/1611/changes/d3fcc5595a84d20992116a9e662f5676cabbf060).

Closes: #970


storage/src/nodestore/header.rs
```diff
@@ -32,128 +32,188 @@ use super::primitives::{LinearAddress, area_size_hash};
...
 use crate::logger::{debug, trace};
 
-/// Can be used by filesystem tooling such as "file" to identify
-/// the version of firewood used to create this `NodeStore` file.
+/// Can be used by filesystem tooling such as "file" to identify the version of
+/// firewood used to create this `NodeStore` file.
+///
+/// From firewood v0.0.4 to sometime between v0.0.18, the 16 bytes "firewood 0.0.x"
+/// followed by enought NUL bytes to pad to 16 bytes. Between v0.0.18 and v0.1.0,
+/// the string was change to the literal "firewood-v1" plus padding. The version
```
RodrigoVillar, line 40: q: why `firewood-v1` and not `firewood-v0.1`?

storage/src/nodestore/header.rs
```diff
@@ -281,6 +383,31 @@ impl NodeStoreHeader {
...
+        } else {
+            None
+        }
+    }
+
+    pub fn firewood_version_str(&self) -> Option<std::borrow::Cow<'_, str>> {
+        self.git_describe()
+            .map(GitDescribe::as_str)
+            .or_else(|| self.cargo_version().map(CargoVersion::as_str))
+    }
```
RodrigoVillar, line 452: (nit/feel-free-to-ignore): the term "firewood version" seems to be overloaded here (can refer to git version or cargo version). Would it make sense to add a function doc that makes it clear what "firewood version" means here?

storage/src/nodestore/header.rs
demosdemon: I wanted it to signify the version of data format which is independent of the firewood version. This file format may continue all the way to Firewood v1.0 or be changed in ways that require radically different interpretations.

RodrigoVillar: Ah gotcha - makes sense.

firewood/src/db.rs
```diff
@@ -1266,6 +1266,42 @@ mod test {
...
     }
 
+    #[test]
+    fn test_backwards_compatible_magic_string() {
+        use std::os::unix::fs::FileExt;
+
+        let testdb = TestDb::new();
+
+        testdb
+            .db
```
RodrigoVillar, line 1276: nit: should we just call `testdb.propose()` (i.e. no need to refer to inner database)

AminR443 (approved):

Issue #1605, opened 2026-01-12
Feat: Include hash of root revision in Header


demosdemon:
DB header currently only includes the address of the root revision. In order to verify consistency for the root outside of external sources, the hash should be included in the header as well.

PR #1612, opened 2026-01-14
feat!: add the hash of the root node to the header

demosdemon:
## Why this should be merged

Prior to this, we did not include the root hash in the header when writing it to disk. This means that when opening the database, the root node must be read and hashed in order fill in the nodestore structure. This also means that the checker cannot independently verify the root node for consistency.

## How this works

This change adds the hash to the header and uses it when opening the database.

## How this was tested

CI and UT.

Closes: #1605


storage/src/nodestore/header.rs
```diff
@@ -29,8 +29,11 @@ use std::io::{Error, ErrorKind, Read};
 
 use super::alloc::FreeLists;
 use super::primitives::{LinearAddress, area_size_hash};
-use crate::NodeHashAlgorithm;
 use crate::logger::{debug, trace};
+use crate::{NodeHashAlgorithm, TrieHash};
+
+/// A pair of root location information: (address, hash)
+pub type RootLocationPair = (LinearAddress, TrieHash);
```
RodrigoVillar: I'm not sure if the root node hash would be considered a description of location (whereas a name like `RootNodeInfo` would be more general). However, I don't feel strongly about this so please resolve if you think `RootLocationPair` is fitting.

RodrigoVillar (approved):

PR #1614, opened 2026-01-14
feat!: Remove `GetFromRoot` API

### LINKED ISSUE #1613: Remove `fwd_get_from_root`

As mentioned in #1609, `get_from_root` is a stale API that is no longer needed. We want to remove the `get_from_root` api prior to freezing the v0.1.0 api and instead require callers to get values after acquiring a reference to the revision. While the caller holds a reference to the revision, we can guarantee that the revision will not be deleted from any concurrent operations.

alarso16:
## Why this should be merged

Closes #1613

## How this works

Deletes the API

## How this was tested

CI

### DIFF
--- a/ffi/src/replay.rs
+++ b/ffi/src/replay.rs
@@ -17,8 +17,8 @@ use std::path::PathBuf;
 use std::sync::OnceLock;
 
 use firewood_replay::{
-    Batch, Commit, DbOperation, GetFromProposal, GetLatest, KeyValueOp, ProposalId,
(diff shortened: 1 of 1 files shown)


RodrigoVillar (commented):
I think we may need to remove the `get_from_root()` references in `ffi/src/replay.rs`

ffi/src/replay.rs
RodrigoVillar:
```suggestion
    Batch, Commit, DbOperation, GetFromProposal, GetLatest, KeyValueOp, ProposalId,
```

One last mention of `GetFromRoot`

PR #1618, opened 2026-01-14
refactor: decouple header from `NodeStore`

### LINKED ISSUE #1616: Extract `NodeStoreHeader` from `NodeStore`
When persisting, we currently require a mutable reference to `NodeStore` as persisting modifies the header. This creates an architectural incompatibility with deferred persistence as we'll eventually need to modify a committed `NodeStore` instance, but it is already being referenced elsewhere (e.g. a proposal being built on top of the committed revision).

To solve this, we extract the header from NodeStore so that it can be managed independently as a single source of truth

RodrigoVillar:
## Why this should be merged

Closes #1616

## How this works

- Extracts the header from `NodeStore` type
  - There is now a "global" header handled by the `RevisionManager`
- Instantiation and persists of the `NodeStoreHeader` are handled by the header itself.
- Moves instantation of header and `by_hash` to before we create the manager
  - This avoids us having to modify these values through a mutex/read-write-lock. 

## How this was tested

CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,6 +35,8 @@ jobs:
             os: macos-latest
           - profile-key: debug-ethhash-logger
             profile-args: "--features ethhash,logger"
+          - profile-key: debug-all-features
+            profile-args: "--all-features"
           - profile-key: maxperf-ethhash-logger
             profile-args: "--profile maxperf --features ethhash,logger"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
@@ -89,6 +91,8 @@ jobs:
             os: macos-latest
           - profile-key: debug-ethhash-logger
             profile-args: "--features ethhash,logger"
+          - profile-key: debug-all-features
+            profile-args: "--all-features"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
     steps:
       - uses: actions/checkout@v4
@@ -121,6 +125,8 @@ jobs:
             os: macos-latest
           - profile-key: debug-ethhash-logger
             profile-args: "--features ethhash,logger"
+          - profile-key: debug-all-features
+            profile-args: "--all-features"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
     steps:
       - uses: actions/checkout@v4
@@ -150,6 +156,8 @@ jobs:
             os: macos-latest
           - profile-key: debug-ethhash-logger
             profile-args: "--features ethhash,logger"
+          - profile-key: debug-all-features
+            profile-args: "--all-features"
           - profile-key: maxperf-ethhash-logger
             profile-args: "--cargo-profile maxperf --features ethhash,logger"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
@@ -178,6 +186,8 @@ jobs:
             os: macos-latest
           - profile-key: debug-ethhash-logger
             profile-args: "--features ethhash,logger"
+          - profile-key: debug-all-features
+            profile-args: "--all-features"
           - profile-key: maxperf-ethhash-logger
             profile-args: "--profile maxperf --features ethhash,logger"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
--- a/clippy.toml
+++ b/clippy.toml
@@ -11,6 +11,7 @@ doc-valid-idents = [
 
 disallowed-methods = [
   { path = "rand::rng", replacement = "firewood_storage::SeededRng::from_env_or_random", reason = "use a prng with a user-defined seed instead", allow-invalid = true },
+  { path = "std::os::unix::fs::FileExt::write_at", replacement = "write_all_at", reason = "use write_all_at instead of write_at to handle short writes ensuring the entire buffer is written", allow-invalid = true },
 ]
 
 disallowed-types = [
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -1285,7 +1285,7 @@ mod test {
         assert_eq!(&version, b"firewood-v1\0\0\0\0\0");
 
         // overwrite the magic string to simulate an older version
-        file.write_at(b"firewood 0.0.18\0", 0).unwrap();
+        file.write_all_at(b"firewood 0.0.18\0", 0).unwrap();
         drop(file);
 
         let testdb = testdb.reopen();
--- a/storage/src/linear/filebacked.rs
+++ b/storage/src/linear/filebacked.rs
@@ -174,7 +174,8 @@ impl ReadableStorage for FileBacked {
 impl WritableStorage for FileBacked {
     fn write(&self, offset: u64, object: &[u8]) -> Result<usize, FileIoError> {
         self.fd
-            .write_at(object, offset)
+            .write_all_at(object, offset)
+            .map(|()| object.len())
             .map_err(|e| self.file_io_error(e, offset, Some("write".to_string())))
     }
 


storage/src/nodestore/persist.rs
```diff
@@ -46,30 +47,33 @@ use super::{Committed, NodeStore};
...
+    /// Persist this header to storage including all the padding.
+    /// This is only done the first time we write the header.
     ///
     /// # Errors
     ///
     /// Returns a [`FileIoError`] if the header cannot be written.
-    pub fn flush_header_with_padding(&self) -> Result<(), FileIoError> {
-        let mut header_bytes = bytemuck::bytes_of(&self.header).to_vec();
-        header_bytes.resize(NodeStoreHeader::SIZE as usize, 0);
-        debug_assert_eq!(header_bytes.len(), NodeStoreHeader::SIZE as usize);
```
demosdemon, line 70: I know this was here before and you touching it is only because of the refactor; however, this method seems unnecessary.

I would much rather remove this method altogether. In its place where it is called, within the `if config.truncate { ... }` branch, replace it with a normal flush to of the header and then call [`set_len`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.set_len) on the file to `NodeStoreHeader::SIZE` afterwards to complete the truncation.

firewood/src/manager.rs
```diff
@@ -181,7 +196,10 @@ impl RevisionManager {
         }
 
         if config.truncate {
-            nodestore.flush_header_with_padding()?;
+            manager
+                .header
+                .lock()
+                .flush_with_padding_to(storage.as_ref())?;
```
demosdemon: This can be done before wrapping the header in the mutex. Previously, we needed to do this after constructing the node store because the method was on the node store and it also owned the header. But, with it detached from the nodestore we can do this write before constructing the manager.

firewood/src/manager.rs
demosdemon: this can now be done before wrapping the `by_hash` map with the RwLock.

firewood/src/manager.rs
```diff
@@ -89,6 +89,12 @@ pub(crate) struct RevisionManager {
     /// is preserved on disk for historical queries.
     max_revisions: usize,
 
+    /// Persisted metadata for the database.
+    ///
+    /// Loaded from disk on startup and updated during commits when nodes
+    /// are persisted or deleted nodes are added to the free lists.
+    header: Mutex<NodeStoreHeader>,
```
demosdemon, line 96: I'm curious if there are any performance implications of this change, cc @AminR443.

firewood/src/manager.rs
```diff
@@ -405,6 +424,12 @@ impl RevisionManager {
             .clone()
     }
 
+    /// Returns a clone of the current header.
+    /// This is used for read-only operations like checking the database.
+    pub fn header(&self) -> NodeStoreHeader {
+        *self.header.lock()
+    }
```
demosdemon: This hides a lock and potentially expensive copy operation. Since this is currently only used by the checker, let's make this `pub(crate)` and instead return a [`parking_lot::MutexGuard<'_, NodeStoreHeader>`](https://docs.rs/parking_lot/latest/parking_lot/type.MutexGuard.html) which implement `DerefMut`.

storage/src/nodestore/persist.rs
RodrigoVillar: Done: 8bae809 (since we only have access to `FileBacked` and not `File`, I added a `set_len()` method on `FileBacked` for this case).

firewood/src/manager.rs
RodrigoVillar: Done: 063cc54

firewood/src/manager.rs
RodrigoVillar: Done: 7eb9045

firewood/src/manager.rs
RodrigoVillar: Good call; done: ad35cdd

demosdemon (approved):

firewood/src/manager.rs
rkuris: There is only one consumer of header (well, arguably the checker is also a consumer, and in theory the checker can run while the database is running, but we don't currently support that) so I suspect there is no contention here.

firewood/src/manager.rs
```diff
@@ -424,10 +427,10 @@ impl RevisionManager {
             .clone()
     }
 
-    /// Returns a clone of the current header.
+    /// Returns a guard to the current header.
     /// This is used for read-only operations like checking the database.
-    pub fn header(&self) -> NodeStoreHeader {
-        *self.header.lock()
+    pub(crate) fn header(&self) -> MutexGuard<'_, NodeStoreHeader> {
```
rkuris: nit: not sure I love this pattern. It lowers the readability of the caller. I would assume "header" just returns self.header, but it actually gets a lock and can block. Maybe naming this "locked_header" would be clearer, or just have the callers do this themselves (there are very few).

firewood/src/manager.rs
```diff
@@ -181,27 +181,26 @@ impl RevisionManager {
             })
             .transpose()?;
 
+        if config.truncate {
+            header.flush_to(storage.as_ref())?;
+            storage.set_len(NodeStoreHeader::SIZE)?;
```
rkuris, line 186: Unfortunately, the way this is implemented is not what we want. What this does on the OS side is create a hole, particularly on some file systems. I think I'd prefer to write zeroes instead of relying on the OS hole.

At a minimum, this should have been mentioned in the PR description as it's an unrelated change, arguably should even be in a different PR.

firewood/src/manager.rs
```diff
@@ -167,23 +181,26 @@ impl RevisionManager {
...
+        }
+
+        let mut by_hash = HashMap::new();
+        if let Some(hash) = nodestore.root_hash().or_default_root_hash() {
+            by_hash.insert(hash, nodestore.clone());
```
rkuris, line 191: This also appears unrelated to this PR. Are you fixing a bug where the initial revision is not recorded in the by_hash array?

At a minimum, this should be mentioned in the PR description. Ideally it should be a separate tiny PR.

firewood/src/manager.rs
```diff
@@ -298,6 +316,10 @@ impl RevisionManager {
             self.by_hash.write().insert(hash, committed.clone());
         }
 
+        // At this point, we can release the lock on the header as the header
+        // and the last committed revision are up-to-date.
+        drop(header);
+
```
rkuris, line 322: nit: I probably would have just ignored this. Keeping this lock around during the very short next phase is not very expensive and it would lower cognitive load on this function.

storage/src/nodestore/persist.rs
```diff
@@ -46,30 +47,15 @@ use super::{Committed, NodeStore};
...
-    /// # Errors
-    ///
-    /// Returns a [`FileIoError`] if the header cannot be written.
-    pub fn flush_header_with_padding(&self) -> Result<(), FileIoError> {
-        let mut header_bytes = bytemuck::bytes_of(&self.header).to_vec();
-        header_bytes.resize(NodeStoreHeader::SIZE as usize, 0);
-        debug_assert_eq!(header_bytes.len(), NodeStoreHeader::SIZE as usize);
-
-        self.storage.write(0, &header_bytes)?;
+    pub fn flush_to<S: WritableStorage>(&self, storage: &S) -> Result<(), FileIoError> {
```
rkuris: nit: naming: `write_to`, as we aren't calling flush or anything.

storage/src/nodestore/persist.rs
```diff
@@ -46,30 +47,15 @@ use super::{Committed, NodeStore};
...
+        let header_bytes = bytemuck::bytes_of(self);
+        storage.write(0, header_bytes)?;
         Ok(())
```
rkuris, line 59: This hides the fact that `storage.write` returns a number of bytes written, which better be equal to the number of bytes in header_bytes, or we should be returning an error.

```suggestion
        if storage.write(0, header_bytes)? != header_bytes.len() {
           Err(...)
        } else {
           Ok(())
        }
```

firewood/src/manager.rs
RodrigoVillar: This was requested here: https://github.com/ava-labs/firewood/pull/1618#discussion_r2709576640 - going to document this in the PR description.

firewood/src/manager.rs
demosdemon: If we don't want to rely on the OS holes, then we should **always** write all 2048 bytes and not just when we want to truncate the file.

firewood/src/manager.rs
demosdemon: disagree. there is a non-trivial amount of work that follows this drop including acquiring a mutex twice.

storage/src/nodestore/persist.rs
demosdemon: This check is not performed anywhere else and it feels odd to enforce that here but not when we write nodes.

- https://github.com/ava-labs/firewood/blob/98c8762a27b4cc0439bc09adfaf42dd486d1b8af/storage/src/nodestore/alloc.rs#L315
- https://github.com/ava-labs/firewood/blob/98c8762a27b4cc0439bc09adfaf42dd486d1b8af/storage/src/nodestore/alloc.rs#L329
- https://github.com/ava-labs/firewood/blob/98c8762a27b4cc0439bc09adfaf42dd486d1b8af/storage/src/nodestore/alloc.rs#L521
- https://github.com/ava-labs/firewood/blob/98c8762a27b4cc0439bc09adfaf42dd486d1b8af/storage/src/nodestore/persist.rs#L57
- https://github.com/ava-labs/firewood/blob/98c8762a27b4cc0439bc09adfaf42dd486d1b8af/storage/src/nodestore/persist.rs#L288-L292

If we do want to start enforcing that, then it should be a more holistic approach instead of one-offing it here.

firewood/src/manager.rs
RodrigoVillar: @rkuris removing the mutex from the header means we'd need to make the reference on the revision manager in `commit()` mutable; with multiple threads having access to the revision manager (e.g. proposals reference the database which hold the revision manager), we'd need to propagate the mutex which is invasive and worse off IMO than having the mutex over just the header.

rkuris: I know, I was just commenting that this is generally speaking an uncontested mutex, which means the performance is very unlikely to be any different.

firewood/src/manager.rs
rkuris: I'm. not following this. Once we've written the zeroes a hole should never get created, is that incorrect?

storage/src/nodestore/persist.rs
rkuris: Good callout, I created an issue. We should fix this.

firewood/src/manager.rs
RodrigoVillar: Renamed to `locked_header`: 3f56336

firewood/src/manager.rs
AminR443: Yeah, I think that in current setup it should have negligible overhead, nanosecond effect on each commit (an additional `compare_exchange` happening). I can run my quick performance check later on this PR. How'll this work with deferred persist though? I haven't reviewed this PR yet, is there any possibility of contention in future?

RodrigoVillar: In a world with deferred persistence, the persist worker will be the consumer of the header and so there won't be an increase in contention.

firewood/src/manager.rs
rkuris: It's a nit, so I don't care much about this, but this is the only holder of this mutex (it can't contend, except in the very obscure case of a parallel check) so releasing it early just adds code, comments, and cognitive load with no real performance implications.

RodrigoVillar: Releasing the mutex early means we can move onto committing the child proposal of the current revision being committed. I'm not sure what the actual runtime cost of proposal cleanup is, but IMO the cognitive load this adds (which is low) is a fair tradeoff for adding the lock.

firewood/src/manager.rs
demosdemon: > I'm. not following this. Once we've written the zeroes a hole should never get created, is that incorrect?

What about when we want to make changes to the header such as removing a field? Removing a field from the end will leave garbage since we're not overwriting it with zeros. Additionally, we insert holes in the file every time we write a node [1] and do not handle reading from those holes any differently [2]. When reading the header, we do not read all 2048 bytes [3]. The linux kernel guarantees that reading from a hole will transparently return all zeros [4]. And finally, because 2048 is less than 4096 (the default `PAGE_SIZE` which all blocks must be GTE [5]), the hole is filled with zeros anyways when writing to the filesystem because holes are only recorded when they span whole blocks [6]. So, I don't understand what the problem is if there is a hole? 

[1]: https://github.com/ava-labs/firewood/blob/f7e086309d515b259d5ee8fabbc011b79e7a6719/storage/src/linear/filebacked.rs#L168-L184
[2]: https://github.com/ava-labs/firewood/blob/f7e086309d515b259d5ee8fabbc011b79e7a6719/storage/src/linear/filebacked.rs#L252-L269
[3]: https://github.com/ava-labs/firewood/blob/f7e086309d515b259d5ee8fabbc011b79e7a6719/storage/src/nodestore/header.rs#L314-L321
[4]: https://lwn.net/Articles/357767/#:~:text=A%20hole%20in,to%20user%20space.
[5]: https://github.com/torvalds/linux/blob/a66191c590b3b58eaff05d2277971f854772bd5b/block/bdev.c#L140-L153
[6]: https://github.com/torvalds/linux/blob/a66191c590b3b58eaff05d2277971f854772bd5b/fs/ext4/inode.c#L4440

rkuris: Valid point about it being only 2k anyway. Chances of having the block size be less than that is pretty much nothing.

Issue #1586, opened 2025-12-29
Metrics macros do not deduplicate calls to `describe_*`


demosdemon:
## Problem 

Our metrics macros use a callsite local `Once` to ensure the describe macros is called only once. However, there's nothing preventing the a metric with the same name from creating its own `Once` at a different callsite and calling the describe macros.

All of the metric callsites are now peppered with `Once` locks which adds unnecessary overhead on every metric invocation, even if metrics are not enabled.

As an example, `firewood/src/merkle/mod.rs` invokes `firewood_counter!("insert", ...)` 5 times:
- https://github.com/ava-labs/firewood/blob/939d85d318e10d28d72ff0dd8fad9c0a527e9171/firewood/src/merkle/mod.rs#L661
- https://github.com/ava-labs/firewood/blob/939d85d318e10d28d72ff0dd8fad9c0a527e9171/firewood/src/merkle/mod.rs#L682
- https://github.com/ava-labs/firewood/blob/939d85d318e10d28d72ff0dd8fad9c0a527e9171/firewood/src/merkle/mod.rs#L704
- https://github.com/ava-labs/firewood/blob/939d85d318e10d28d72ff0dd8fad9c0a527e9171/firewood/src/merkle/mod.rs#L727
- https://github.com/ava-labs/firewood/blob/939d85d318e10d28d72ff0dd8fad9c0a527e9171/firewood/src/merkle/mod.rs#L757

That is 5 different `Once` tokens each with their own state meaning we will invoke `describe_counter!("insert", ...)` five different times. That's also 5 different opportunities to have a different description for the same metric and which one is applied is non-deterministic. 

## Suggestion

Remove the custom metrics macros and move all of the `describe_*` macro calls into a dedicated method that is called only once when initializing the metrics.

Issue #1090, opened 2025-07-18
feat: have some way to fetch metrics for only one database


rkuris:
It's not clear if we want this. It seems like collecting metrics should be done globally by golang, but if it wants them separated it shouldn't be hard to do that once we switch to tracing.

Issue #1089, opened 2025-07-18
feat: use tracing to separate logs and metrics by database


rkuris:
We need some automated way of adding metadata to each log and/or metric event. The tracing library has support for this. We'd just need to open a span at the start of each ffi call with the database name as some metadata, then exit the span when returning.

Logging can then be modified to support logging to separate files.

Tracing library: https://crates.io/crates/tracing

Steps:
- [ ] Keep existing functionality, but start using tracing
- [ ] Create spans in each ffi call, including the database name when started
- [ ] Split out the log files based on open parameters

demosdemon:
We've decided to use labels instead of splitting the library for metrics.

PR #1619, opened 2026-01-15
feat(metrics): metrics crate, separate registries and expensive metrics (1/2)

AminR443:
## Why this should be merged
Centralizes metric definitions into per-crate registries and introduces infrastructure for expensive metrics. This:

  1. **Improves maintainability** - Metric names/descriptions are defined once in registry modules (`ffi/src/registry.rs`, `firewood/src/registry.rs`, `storage/src/registry.rs`, `replay/src/registry.rs`) rather than scattered across callsites
  2. **Enables expensive metric recording** - Adds thread-local `MetricsContext` so callers can opt-in to expensive metrics at API boundaries (e.g., FFI entrypoints)
  3. **Simplifies callsites** - New macros (`firewood_increment!`, `firewood_set!`, `firewood_record!`) are cleaner than the old description-registering macros

Note: This doesn't record any expensive metrics, and doesn't set the metrics context in FFI, those are handled in the next PR (#1620).

Closes: #1586.

Also this can be used in future to solve #1090, #1089, as databases can be separated using the same TLS context, by adding a database name/tag/id field to the struct.

## How this works

  - New `firewood-metrics` crate provides:
    - `MetricsContext` - thread-local context set at API boundaries
    - `set_metrics_context()` - returns a guard that restores previous context on drop
    - Recording macros with optional `expensive` flag that checks `expensive_metrics_enabled()` before recording
  - Each crate defines a `registry` module with metrics and a `register()` function.
  - At startup, `ffi/src/metrics_setup.rs` calls each registry's `register()` function
  - Callsites use the new macros with registry constants:
    ```rust
    firewood_increment!(crate::registry::COMMIT_COUNT, 1);
    firewood_increment!(crate::registry::SLOW_OP, 1, expensive);  // only if enabled
    ```

## How this was tested
Existing UT in Go, added some tests in Rust.

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -832,7 +832,7 @@ mod test {
         test_write_header(
             nodestore,
             high_watermark,
-            Some(root_addr),
+            Some((root_addr, root_hash.clone().into_triehash())),
             FreeLists::default(),
         );
 
--- a/storage/src/nodestore/alloc.rs
+++ b/storage/src/nodestore/alloc.rs
@@ -567,6 +567,7 @@ pub mod test_utils {
 
     use crate::NodeHashAlgorithm;
     use crate::node::Node;
+    use crate::nodestore::header::RootNodeInfo;
     use crate::nodestore::{Committed, NodeStore, NodeStoreHeader};
 
     // Helper function to wrap the node in a StoredArea and write it to the given offset. Returns the size of the area on success.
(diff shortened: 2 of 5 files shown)


ffi/src/metrics_setup.rs
```diff
@@ -18,6 +18,9 @@ static RECORDER: OnceLock<PrometheusHandle> = OnceLock::new();
 /// This happens on a per-process basis, meaning that the metrics system cannot
 /// be initialized if it has already been set up in the same process.
 pub fn setup_metrics() -> Result<(), Box<dyn Error>> {
+    crate::registry::register();
+    firewood::registry::register();
+    firewood_storage::registry::register();
```
demosdemon, line 23: The `replay` register should be invoked here too.

demosdemon (approved):

demosdemon:
One thing I just realized, this needs to provide the metrics context to the parallel hasher threads as they won't pick up the TLS variable on spawn: <https://github.com/ava-labs/firewood/blob/6778a05be1e2fb1f5505d8caf6a0486db53a66da/firewood/src/merkle/parallel.rs#L260>

ffi/src/metrics_setup.rs
AminR443: Thanks for mentioning.

AminR443:
Thanks for mentioning this too. I had it in original changes and it disappeared somehow in splitting.

PR #1620, opened 2026-01-15
feat(metrics): integrate FFI with expensive metrics and add Go tests (2/2)

### LINKED ISSUE #1524: ffi/metrics: provide option to enable expensive metric calculations

AminR443:
## Why this should be merged

This PR adds support for **expensive metrics** (like histograms) to the FFI layer, allowing operators to opt-in to more detailed performance tracking. Histograms provide distribution data (percentiles, buckets) for commit, propose, and batch operations - information not available from simple counters. Since histogram recording has overhead, this is disabled by default and can be enabled per-database handle.

Closes #1524.

## How this works

1. **New `MetricsContext` propagation**: A `MetricsContext` struct tracks whether expensive metrics are enabled for a given database handle. This context is propagated through:
    - `DatabaseHandle` → `RevisionHandle` → `IteratorHandle`
    - `DatabaseHandle` → `ProposalHandle` → `IteratorHandle`

2. **`HasMetricsContext` trait**: Introduced to provide a consistent way for FFI handle types to expose their metrics context.

3. **New histogram metrics**: Three new histogram metrics are recorded when expensive metrics are enabled:
    - `ffi.commit_ms_bucket` - Commit duration distribution
    - `ffi.propose_ms_bucket` - Propose duration distribution
    - `ffi.batch_ms_bucket` - Batch processing duration distribution

4. **Go API**: Added `WithExpensiveMetrics()` option to enable expensive metrics when creating a database.

5. **Metrics-free paths**: Operations that don't need metrics recording (proof verification, serialization, memory freeing) now use `invoke_without_metrics` instead of `invoke_with_handle` to avoid unnecessary context setup.

## How this was tested

- Added `TestExpensiveMetrics` Go test that creates a database with `WithExpensiveMetrics()`, performs operations, and verifies histogram metrics are recorded
- Refactored existing metrics tests to share initialization logic via `sync.Once` patterns

## Performance Note
At first, I thought that TLS will have more overhead compared to Atomic global, as we need to set the context on every FFI operation. I did benchmarks and it turns out TLS is cheaper, I suppose the reads are cheaper and thread cache helps here.

30k block execution (average of 6 runs):

Method | ExpensiveEnabled | ExpensiveDisabled
-- | -- | --
Global | 21.389851963 | 21.204608595
TLS | 21.281072396 | 21.177187864
Normal (no expensive metrics, enabled) | 20.954168492 | 20.954168492
No Metrics (metrics disabled completely) | 20.623543854 | 20.623543854

Non-active overhead is 1%. This is pure firewood; given that Firewood represents 60–70% of execution time in the reexecution test, this translates to roughly 0.60-0.70% overall overhead on bootstrapping.

### DIFF
--- a/ffi/metrics_test.go
+++ b/ffi/metrics_test.go
@@ -87,7 +87,9 @@ func TestMetrics(t *testing.T) {
 	r.NoError(err)
 
 	assertMetrics(t, metricsPort, expectedMetrics)
-	r.True((logPath == "") || assertNonEmptyFile(r, logPath))
+	if logPath != "" {
+		r.True(assertNonEmptyFile(t, logPath))
+	}
 }
 
 func TestExpensiveMetrics(t *testing.T) {
@@ -104,10 +106,11 @@ func TestExpensiveMetrics(t *testing.T) {
 	assertMetrics(t, metricsPort, merged)
 }
 
-func assertNonEmptyFile(r *require.Assertions, path string) bool {
+func assertNonEmptyFile(t *testing.T, path string) bool {
+	t.Helper()
 	f, err := os.ReadFile(path)
-	r.NoError(err)
-	r.NotEmpty(f)
+	require.NoError(t, err)
+	require.NotEmpty(t, f)
 	return true
 }
 
--- a/ffi/src/metrics.rs
+++ b/ffi/src/metrics.rs
@@ -14,31 +14,31 @@ use std::time::Duration;
 
 static RECORDER: OnceLock<PrometheusHandle> = OnceLock::new();
 
-/// Trait for types that carry a context.
+/// Trait for types that carry a [`MetricsContext`].
 ///
 /// Implemented for FFI handle types.
 /// Concrete impls live in their respective modules (handle, revision, proposal, iterator).
-pub(crate) trait HasContext {
-    fn metrics(&self) -> Option<MetricsContext>;
+pub(crate) trait MetricsContextExt {
+    fn metrics_context(&self) -> Option<MetricsContext>;
 }
 
 // some blanket implementations. can't go with Deref approach because of
 // tuple handle in range proofs.
-impl<T: HasContext + ?Sized> HasContext for Box<T> {
-    fn metrics(&self) -> Option<MetricsContext> {
-        (**self).metrics()
+impl<T: MetricsContextExt + ?Sized> MetricsContextExt for Box<T> {
+    fn metrics_context(&self) -> Option<MetricsContext> {
+        (**self).metrics_context()
     }
 }
 
-impl<T: HasContext + ?Sized> HasContext for &T {
-    fn metrics(&self) -> Option<MetricsContext> {
-        (**self).metrics()
+impl<T: MetricsContextExt + ?Sized> MetricsContextExt for &T {
+    fn metrics_context(&self) -> Option<MetricsContext> {
+        (**self).metrics_context()
     }
 }
 
-impl<T: HasContext + ?Sized> HasContext for &mut T {
-    fn metrics(&self) -> Option<MetricsContext> {
-        (**self).metrics()
+impl<T: MetricsContextExt + ?Sized> MetricsContextExt for &mut T {
+    fn metrics_context(&self) -> Option<MetricsContext> {
+        (**self).metrics_context()
     }
 }
 
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -242,11 +242,6 @@ impl DatabaseHandle {
         })
     }
 
-    #[must_use]
-    pub(crate) const fn metrics_context(&self) -> MetricsContext {
-        self.metrics_context
-    }
-
     pub(crate) fn get_root(&self, root: HashKey) -> Result<ArcDynDbView, api::Error> {
         let mut cache_miss = false;
         let view = self.cached_view.get_or_try_insert_with(root, |key| {
@@ -302,8 +297,8 @@ impl<'db> CView<'db> for &'db crate::DatabaseHandle {
     }
 }
 
-impl crate::HasContext for DatabaseHandle {
-    fn metrics(&self) -> Option<MetricsContext> {
+impl crate::MetricsContextExt for DatabaseHandle {
+    fn metrics_context(&self) -> Option<MetricsContext> {
         Some(self.metrics_context)
     }
 }
--- a/ffi/src/iterator.rs
+++ b/ffi/src/iterator.rs
@@ -17,7 +17,7 @@ type KeyValueItem = (merkle::Key, merkle::Value);
 pub struct IteratorHandle<'view> {
     iter: Option<BoxKeyValueIter<'view>>,
     view: Option<ArcDynDbView>,
-    metrics_context: MetricsContext,
+    metrics_context: Option<MetricsContext>,
 }
 
 impl fmt::Debug for IteratorHandle<'_> {
@@ -33,7 +33,7 @@ impl<'view> IteratorHandle<'view> {
     pub(crate) fn new(
         view: ArcDynDbView,
         iter: BoxKeyValueIter<'view>,
-        metrics_context: MetricsContext,
+        metrics_context: Option<MetricsContext>,
     ) -> Self {
         Self {
             iter: Some(iter),
@@ -69,8 +69,8 @@ impl IteratorHandle<'_> {
 #[derive(Debug, Default)]
 pub struct CreateIteratorResult<'db>(pub IteratorHandle<'db>);
 
-impl crate::HasContext for IteratorHandle<'_> {
-    fn metrics(&self) -> Option<MetricsContext> {
-        Some(self.metrics_context)
+impl crate::MetricsContextExt for IteratorHandle<'_> {
+    fn metrics_context(&self) -> Option<MetricsContext> {
+        self.metrics_context
     }
 }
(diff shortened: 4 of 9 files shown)


ffi/tests/eth/eth_compatibility_test.go
alarso16, line 1: Are these not being linted?

ffi/generate_cgo.go
alarso16, line 1: Ok this file is being linted for sure, so why are you reformatting unrelated things? Not accusatory, since they're correct reformats, but unsure how they happened

ffi/metrics_test.go
```diff
@@ -17,37 +19,98 @@ import (
...
+		"ffi_commit_ms_bucket":  dto.MetricType_HISTOGRAM,
+		"ffi_propose_ms_bucket": dto.MetricType_HISTOGRAM,
+		"ffi_batch_ms_bucket":   dto.MetricType_HISTOGRAM,
+	}
+	initMetrics   sync.Once
+	initLogs      sync.Once
+	activeLogPath string
+)
+
+func ensureMetricsStarted(r *require.Assertions) {
```
alarso16: nit: It's more conventional to pass around `t *testing.T` instead. Using `r := require.New(t)` is somewhat discouraged anyway

ffi/metrics_test.go
```diff
@@ -17,37 +19,98 @@ import (
...
+}
+
+func ensureLogsStarted(r *require.Assertions, logPath string) {
+	initLogs.Do(func() {
+		logConfig := &LogConfig{
+			Path:        logPath,
+			FilterLevel: "trace",
+		}
+		if err := StartLogs(logConfig); err != nil {
+			r.Contains(err.Error(), "Logging is not available")
```
alarso16:
```suggestion
			r.ErrorContains(err, "Logging is not available")
```

Why does this happen?

ffi/generate_cgo.go
AminR443: I usually use claude code to clean up the large branch I've been working on and give it instructions on how to split up. It makes sure everything compiles and passes. It sometimes touches unnecessary files, and since this PR touched many files, that slipped past my own review.

ffi/tests/eth/eth_compatibility_test.go
AminR443: explained below.

ffi/generate_cgo.go
AminR443: I can revert those changes if you want.

metrics/src/lib.rs
```diff
@@ -104,6 +128,78 @@ pub fn expensive_metrics_enabled() -> bool {
...
+    pub name: &'static str,
+    /// The bucket boundaries.
+    pub buckets: &'static [f64],
+}
+
+/// Global registry of histogram bucket configurations.
+///
+/// Crates register their histogram buckets during their `register()` call,
+/// and the metrics setup code reads from this registry to configure the exporter.
+static HISTOGRAM_BUCKETS: RwLock<Vec<HistogramBucketConfig>> = RwLock::new(Vec::new());
```
demosdemon: Looks like the RwLock is not really needed. The same could be accomplished by passing a `&mut Vec<_>` to the register functions without needing any synchronization.

metrics/src/lib.rs
```diff
@@ -104,6 +128,78 @@ pub fn expensive_metrics_enabled() -> bool {
...
+///
+/// // In your registry::register() function:
+/// register_histogram_buckets!(
+///     "ffi.commit_ms_bucket",
+///     "Commit duration via FFI in milliseconds",
+///     &[0.1, 0.5, 1.0, 2.0, 5.0, 10.0]
+/// );
+/// ```
+#[macro_export]
+macro_rules! register_histogram_buckets {
```
demosdemon: Similarly, the macro seems unnecessary if all it is doing is calling a function with the same arguments.

demosdemon:
Re your performance note:

> I suppose the reads are cheaper and thread cache helps here.

This is exactly why. A `lock read` operation on x86_64 still takes 2-7ns without contention and cannot be optimized by the compiler or the CPU. Since TLS doesn't involve atomic operations it performs a straight `read` on the pointer without any synchronization barriers. This allows the compiler to optimize when and how any memory access for the TLS slot is actually performed.

I recommend playing around with Godbolt. Claude is good at analyzing assembly and explaining differences ([A](https://rust.godbolt.org/z/3xnYTbY76) vs [B](https://rust.godbolt.org/z/8vjecbK79) with some `block_box` calls peppered in to try to disable optimizations).

ffi/metrics_test.go
```diff
@@ -17,37 +19,102 @@ import (
...
+	_, err := db.Update(keys, vals)
+	r.NoError(err)
 
-	var logsDisabled bool
-	if err := StartLogs(logConfig); err != nil {
-		r.Contains(err.Error(), "Logging is not available")
-		logsDisabled = true
-	}
+	assertMetrics(t, metricsPort, expectedMetrics)
+	r.True((logPath == "") || assertNonEmptyFile(r, logPath))
```
alarso16:
```suggestion
	if logPath != "" {
		r.True(assertNonEmptyFile(r, logPath))
	}
```
I think this is more clear (even though it's technically the same). Feel free to push back and close

ffi/metrics_test.go
```diff
@@ -17,37 +19,102 @@ import (
...
 	_, err := db.Update(keys, vals)
 	r.NoError(err)
 
+	merged := make(map[string]dto.MetricType, len(expectedMetrics)+len(expectedExpensiveMetrics))
+	maps.Copy(merged, expectedMetrics)
+	maps.Copy(merged, expectedExpensiveMetrics)
+	assertMetrics(t, metricsPort, merged)
+}
+
+func assertNonEmptyFile(r *require.Assertions, path string) bool {
```
alarso16:
```suggestion
func assertNonEmptyFile(t *testing.T, path string) bool {
	t.Helper()
```

You should pass around `t`. Also, I think it makes sense to mark the function as a helper. From the docs:
```
Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped. Helper may be called simultaneously from multiple goroutines.
```

You should just play around with it to see how it looks

ffi/metrics_test.go
```diff
@@ -65,25 +132,13 @@ func TestMetrics(t *testing.T) {
 	r.NoError(resp.Body.Close())
 
 	// Check that batch op was recorded (no prefix)
-	r.Contains(string(body), "ffi_batch 1")
+	r.NotContains(string(body), "ffi_batch 0")
```
alarso16, line 138: I don't understand, why are we checking this?

ffi/src/metrics.rs
```diff
@@ -5,35 +5,68 @@ use std::error::Error;
...
 use oxhttp::Server;
 use oxhttp::model::{Body, Response, StatusCode};
 use std::net::Ipv4Addr;
 use std::time::Duration;
 
-use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
-
 static RECORDER: OnceLock<PrometheusHandle> = OnceLock::new();
 
+/// Trait for types that carry a context.
```
alarso16: What is a "context"? Is this a specific Rust meaning like it is in Go?

demosdemon: Austin's comment is a good point. Let's rename this trait to `MetricsContextExt` to follow the common pattern in rust of naming [extension traits](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html) with an `Ext` suffix. Similarly, I would name the method `metrics_context` as well as clarify in the doc string that this is for MetricsContexts.

Essentially, this makes it easier to add more "context" extension traits without overloading this one trait. 

If we want to consider this a "provider," I'd be okay with naming it `MetricsContextProvider`.

ffi/metrics_test.go
AminR443: I usually opt-in for shorter syntax, and don't like if branches like this. Although I agree that this might be a little confusing at first sight, so I'll apply your suggestion.

ffi/metrics_test.go
AminR443: I don't use Go much these days, and my Go is rusty:) I didn't know that passing around require.Assertions isn't a good practice. I fixed it in the other PR and used test helpers but forgot to fix this one. Thanks for mentioning it.

ffi/metrics_test.go
AminR443: Previous test checked this to make sure that ffi_batch was recorded and is equal to 1 (although it would pass for any number beginning with 1, but we had only one metric test). Now we have two tests (one for expensive and one for normal) and go runs these in parallel which causes those dbs to share metrics. This makes the metrics value equal to 2. I replaced it with just checking for non-zero. We can get rid of this I think, but I didn't want to remove it because it isn't one of my changes.

ffi/src/metrics.rs
AminR443: Done

ffi/metrics_test.go
AminR443: Done

ffi/metrics_test.go
AminR443: Done

alarso16 (approved):
Looks good from my perspective.

demosdemon (approved):

PR #1624, opened 2026-01-16
feat(ffi): BatchOp in FFI

### LINKED ISSUE #1433: ffi: nil vs empty slice distinction in API
As of #1420, nil values and empty slices are treated as differently in the FFI. This is atypical in Go, and is super prone to misuse. If we do this internally, that is ok, but it's not recommended to expose this through an API. Some better long term solutions might be:

1. Exposing some `Maybe` structure.
2. Creating a `BatchOp` with `Put`, `Delete`, and `PrefixDelete`(?). This can then be translated into a list of keys and values
3. Create a `Batch`, starting empty, with the same operations as `BatchOp`. This batch can be passed to a proposal later, or can be created from a base proposal

Option 1 is the least invasive, but probably is avoiding the core problem. Options 2 and 3 offer clarity and potentially more features (namely, Delete vs PrefixDelete). Option 2 likely requires breaking API changes, but option 3 also leaves questions about proposing in the concurrent world.

AminR443:
## Why this should be merged
The current FFI API relies on nil vs empty slice semantics to distinguish between delete and put operations: `nil` value means DeleteRange while `[]byte{}` means Put with empty value. This is problematic because in Go, the distinction between `nil` and empty slice is subtle and error-prone. Many Go developers don't expect these to behave differently, making it easy to accidentally trigger a deletion when intending to store an empty value.

Closes #1433.

## How this works
This PR introduces an explicit `BatchOpTag` enum across the FFI layer with three variants: `Put`, `Delete`, and `DeleteRange`. On the Rust side, `KeyValuePair` now includes an `op` field, and `try_into_batch()` uses this tag instead of null-pointer checking.

On the Go side, a new `Batch` type provides builder methods (`Put`, `Delete`, `PrefixDelete`) for constructing operations. New `UpdateBatch()` and `ProposeBatch()` methods accept this type. The existing `Update()` and `Propose()` methods remain for backward compatibility, internally converting to `Batch` while preserving the old nil-means-delete semantics. These are marked as deprecated.

## How this was tested
UT, extends existing tests to cover new APIs.

## Performance Note
30k block execution (average of 12 runs):
- Main branch: `17.345905360s`
- This branch: `17.503642220s`

Overhead is 0.90%. This is pure firewood; given that Firewood represents 60–70% of execution time in the reexecution test, this translates to roughly 0.60% overall overhead on bootstrapping.

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -111,6 +111,8 @@ type config struct {
 	readCacheStrategy CacheStrategy
 	// rootStore defines whether to enable storing all historical revisions on disk.
 	rootStore bool
+	// expensiveMetricsEnabled controls whether expensive metrics recording is enabled.
+	expensiveMetricsEnabled bool
 }
 
 func defaultConfig() *config {
@@ -182,6 +184,15 @@ func WithRootStore() Option {
 	}
 }
 
+// WithExpensiveMetrics enables expensive metrics recording for this database.
+// Expensive metrics are disabled by default since they can introduce overhead.
+// Default: false
+func WithExpensiveMetrics() Option {
+	return func(c *config) {
+		c.expensiveMetricsEnabled = true
+	}
+}
+
 // A CacheStrategy represents the caching strategy used by a [Database].
 type CacheStrategy uint8
 
@@ -246,6 +257,7 @@ func New(dbDir string, nodeHashAlgorithm NodeHashAlgorithm, opts ...Option) (*Da
 		strategy:             C.uint8_t(conf.readCacheStrategy),
 		truncate:             C.bool(conf.truncate),
 		root_store:           C.bool(conf.rootStore),
+		expensive_metrics:    C.bool(conf.expensiveMetricsEnabled),
 		node_hash_algorithm:  C.enum_NodeHashAlgorithm(nodeHashAlgorithm),
 	}
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,7 +1172,6 @@ version = "0.0.18"
 dependencies = [
  "cbindgen",
  "coarsetime",
- "derive-where",
  "env_logger",
  "firewood",
  "firewood-metrics",
--- a/ffi/Cargo.toml
+++ b/ffi/Cargo.toml
@@ -30,7 +30,6 @@ parking_lot.workspace = true
 # Regular dependencies
 oxhttp = "0.3.1"
 tikv-jemallocator = "0.6.1"
-derive-where = "1.6.0"
 metrics-exporter-prometheus = { version = "0.17.2", default-features = false }
 # Optional dependencies
 rlp = { version = "0.6.1", optional = true }
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -1007,6 +1007,12 @@ typedef struct DatabaseHandleArgs {
    */
   bool truncate;
   /**
+   * Whether to enable expensive metrics recording for this database handle.
+   *
+   * Expensive metrics are disabled by default.
+   */
+  bool expensive_metrics;
+  /**
    * The hashing mode to use for the database.
    *
    * This must match the compile-time feature:
--- a/ffi/generate_cgo.go
+++ b/ffi/generate_cgo.go
@@ -127,7 +127,7 @@ func changeCgoDirectivesForFile(targetMode string, targetFile string) error {
 		fmt.Printf("[INFO] No changes needed to %s\n", targetFile)
 		return nil
 	}
-	return os.WriteFile(targetFile, []byte(newContents), 0644)
+	return os.WriteFile(targetFile, []byte(newContents), 0o644)
 }
 
 func isCGODirective(line string) bool {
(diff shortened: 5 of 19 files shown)


ffi/firewood.go
```diff
@@ -265,14 +280,32 @@ func (db *Database) Update(keys, vals [][]byte) (Hash, error) {
...
-// Value Semantics:
+// Deprecated: Use [Database.ProposeBatch] instead which provides explicit operation types.
+// This function maintains backward compatibility by treating:
 //   - nil value (vals[i] == nil): Performs a DeleteRange operation using the key as a prefix
 //   - empty slice (vals[i] != nil && len(vals[i]) == 0): Inserts/updates the key with an empty value
 //   - non-empty value: Inserts/updates the key with the provided value
 //
 // This function conflicts with all other calls that access the latest state of the database,
 // and will lock for the duration of this function.
 func (db *Database) Propose(keys, vals [][]byte) (*Proposal, error) {
```
alarso16: In my personal opinion, if we are going to deprecate these immediately before the first "real" release, then let's just delete them and have `Propose` and `Update` take batches

ffi/batch_op.go
```diff
@@ -0,0 +1,90 @@
...
+func (b *Batch) Put(key, value []byte) {
+	b.ops = append(b.ops, batchOp{
+		key:   key,
+		value: value,
+		op:    C.BatchOpTag_Put,
+	})
+}
+
+// Delete adds an operation to delete a specific key.
+func (b *Batch) Delete(key []byte) {
```
alarso16: I think this is untested

ffi/firewood.go
RodrigoVillar: +1, it's easier to add this back than to remove them post `v0.1`

ffi/batch_op.go
```diff
@@ -0,0 +1,90 @@
...
+	key   []byte
+	value []byte
+	op    C.enum_BatchOpTag
+}
+
+// Batch accumulates database operations to be applied atomically.
+// Create a new Batch with [NewBatch], add operations with [Batch.Put],
+// [Batch.Delete], and [Batch.PrefixDelete], then apply with
+// [Database.UpdateBatch] or [Database.ProposeBatch].
+type Batch struct {
```
demosdemon: If the `Batch` type owned a `Pinner`, it could build up a collection `C.KeyValuePair`.

ffi/src/value/kvp.rs
```diff
@@ -10,19 +10,42 @@ use firewood::v2::api;
...
 pub type OwnedKeyValueBatch = OwnedSlice<OwnedKeyValuePair>;
 
-/// A `KeyValue` represents a key-value pair, passed to the FFI.
+/// Tag indicating the type of batch operation.
+///
+/// This is used by the FFI to explicitly distinguish between different
+/// operation types instead of relying on nil vs empty pointer semantics.
+#[repr(C)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BatchOpTag {
```
demosdemon: This dedicated enum is kinda unnecessary. A `#[repr(C)]` `enum` will automatically create one too.

```rust
#[repr(C)]
enum BatchOp<'a> {
    Put { key: BorrowedBytes<'a>, value: BorrowedBytes<'a> },
    Delete { key: BorrowedBytes<'a> },
    DeleteRange { prefix: BorrowedBytes<'a> },
}
```

Would generate this in `firewood.h`:

```c
typedef enum BatchOp_Tag {
  BatchOp_Put,
  BatchOp_Delete,
  BatchOp_DeleteRange,
} BatchOp_Tag;

typedef struct BatchOp_Put_Body {
  BorrowedBytes key;
  BorrowedBytes value;
} BatchOp_Put_Body;

typedef struct BatchOp_Delete_Body {
  BorrowedBytes key;
} BatchOp_Delete_Body;

typedef struct BatchOp_DeleteRange_Body {
  BorrowedBytes prefix;
} BatchOp_DeleteRange_Body;

typedef struct BatchOp {
  BatchOp_Tag tag;
  union {
    BatchOp_Put_Body put;
    BatchOp_Delete_Body delete_;
    BatchOp_DeleteRange_Body delete_range;
  };
} BatchOp;
```

However, it does make creating the BatchOp in `go` a little uglier.

ffi/firewood.go
demosdemon: Agreed. We can remove this since we already have other breaking changes to reconcile with avalanchego.

ffi/firewood_test.go
```diff
@@ -355,66 +352,76 @@ func sortKV(keys, vals [][]byte) error {
...
 			name: "Propose",
-			insert: func(db dbView, keys, vals [][]byte) (dbView, error) {
-				proposal, err := db.Propose(keys, vals)
+			insert: func(t *testing.T, db dbView, batch []BatchOp) (dbView, error) {
+				proposal, err := db.Propose(batch)
 				if err != nil {
 					return nil, err
 				}
+				t.Cleanup(func() {
+					require.NoError(t, proposal.Drop())
```
alarso16, line 411: Good catch, kind of surprised you noticed this. Did you add it because you had an issue? I wouldn't expect this to cause any issues

demosdemon (approved):

ffi/firewood_test.go
AminR443: No, just noticed it when replacing those key values with batch ops.

PR #1627, opened 2026-01-16
chore(release): prepare for v0.1.0

demosdemon:
Bumps package versions to 0.1.0 and decouples their version from the main Cargo.toml

Upgrades all dependencies except for fjall as it introduces significant breaking changes.

PR #1630, opened 2026-01-20
fix(flake): grab version from correct location

demosdemon:
## Why this should be merged

#1627 broke the flake nix by removing a field from the cargo.toml that is parsed.

## How this works

This corrects the flake to grab the version from the correct location.

## How this was tested

Added a PR trigger for the flake nix when touching the flake file or go modules.

### DIFF
--- a/.github/workflows/ffi-nix.yaml
+++ b/.github/workflows/ffi-nix.yaml
@@ -3,6 +3,12 @@ name: ffi-nix
 on:
   push:
     branches: [main]
+  pull_request:
+    paths:
+      - ffi/flake.nix
+      - ffi/flake.lock
+      - ffi/go.mod
+      - ffi/go.sum
 
 jobs:
   ffi-nix:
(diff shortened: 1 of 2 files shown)


.github/workflows/ffi-nix.yaml
```diff
@@ -3,6 +3,12 @@ name: ffi-nix
 on:
   push:
     branches: [main]
+  pull_request:
```
maru-ava, line 6: Presumably this gets removed before merge?

demosdemon: Should it? With this config, it will run this workflow whenever the flake is edited. If this was here and also included `Cargo.toml`, then I would have caught the breakage before merging the other PR. At least this way when editing the flake, it also runs on the same PR. However, the workflow will not run if none of these 4 files are touched.

maru-ava: Fair enough, keep it.

PR #1637, opened 2026-01-23
feat: Creating change proofs using FFI

bernard-avalabs:
## Why this should be merged

This PR implements `fwd_db_change_proof` in `ffi/src/proof/changes.rs` to allow AvalancheGo to create a change proof. This is the first PR in a series of PR to allow change proofs to be created, verified, committed, and serialized/deserialized using FFI calls.

## How this works

The design largely follows range proof creation. It mainly differs by using a `DiffMerkleNodeStream` to create an array of `BatchOp`s that represents the difference between two revisions.

## How this was tested

Change proof to iterator tests have been added in `firewood/src/merkle/changes.rs`. Basic change proof creation tests from golang have been added in `ffi/proofs_test.go`. More tests from golang will be added in the followup PRs.

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -284,7 +284,21 @@ impl DatabaseHandle {
     ///
     /// # Errors
     ///
-    /// An error is returned if the revision cannot be found or if there is an I/O error.
+    /// * `api::Error::StartRevisionNotFound` - If the revision for `start_hash` cannot
+    ///   be found. Note that only an `EndRevisionNotFound` is returned when both
+    ///   `start_hash` and `end_hash` cannot be found.
+    ///
+    /// * `api::Error::EndRevisionNotFound` - If the revision for `end_hash` cannot be
+    ///   found. Note that only an `EndRevisionNotFound` is returned when both
+    ///   `start_hash` and `end_hash` cannot be found.
+    ///
+    /// * `api::Error::InvalidRange` - If `start_key` > `end_key` when both are provided.
+    ///   This ensures the range bounds are logically consistent.
+    ///
+    /// * `api::Error` - Various other errors can occur during proof generation, such as:
+    ///   - I/O errors when reading nodes from storage
+    ///   - Corrupted trie structure
+    ///   - Invalid node references
     pub(crate) fn change_proof(
         &self,
         start_hash: HashKey,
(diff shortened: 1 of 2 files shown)


ffi/proofs.go
```diff
@@ -579,6 +579,10 @@ func getChangeProofFromChangeProofResult(result C.ChangeProofResult) (*ChangePro
 	switch result.tag {
 	case C.ChangeProofResult_NullHandlePointer:
 		return nil, errDBClosed
+	case C.ChangeProofResult_RevisionNotFound:
+		// NOTE: the result value contains the provided root hash, we could use
+		// it in the error message if needed.
+		return nil, errRevisionNotFound
```
alarso16: Whether this is the correct architecture decision or not is kind of related, but right now, x/sync server asks the DB for a change proof, and if the starting revision is not found, the x/sync server will instead ask the DB for range proof. If the end revision is not found then we can't do anything and drop the request.

Anyway, I think we need two separate errors exported: `ErrStartRevisionNotFound` and `ErrEndRevisionNotFound`, or we need to change the behavior of x/sync

ffi/revision.go
```diff
@@ -16,8 +16,10 @@ import (
 )
 
 var (
-	ErrDroppedRevision  = errors.New("revision already dropped")
-	errRevisionNotFound = errors.New("revision not found")
+	ErrDroppedRevision       = errors.New("revision already dropped")
+	errRevisionNotFound      = errors.New("revision not found")
+	errEndRevisionNotFound   = errors.New("end revision not found")
+	errStartRevisionNotFound = errors.New("start revision not found")
```
alarso16:
```suggestion
	ErrEndRevisionNotFound   = errors.New("end revision not found")
	ErrStartRevisionNotFound = errors.New("start revision not found")
```
Will you export these?

ffi/proofs.go
bernard-avalabs: Thanks for letting me know about x/sync's behavior with respect to missing revisions.

I've added separate errors for missing start revision and end revision. In the case where both are missing, change proof creation will just return a missing end revision as it terminates on the first error.

alarso16: I think that's a good case to call out, since I didn't think about it really. Will you put it in the comment?

bernard-avalabs: I've added a comment in `fw_db_change_proof`

> /// - [`ChangeProofResult::StartRevisionNotFound`] if the caller provided a start root
> ///   that was not found in the database. The missing root hash is included in the result.
> ///   If both the start root and end root are missing, then only the end root is
> ///   reported.

ffi/revision.go
bernard-avalabs: I've made this change.

ffi/src/handle.rs
```diff
@@ -274,6 +280,43 @@ impl DatabaseHandle {
...
+            end_key,
+            Merkle::from(self.db.revision(start_hash).map_err(|err| {
+                if let api::Error::RevisionNotFound { provided } = err {
+                    api::Error::StartRevisionNotFound { provided }
+                } else {
+                    err
+                }
+            })?)
+            .nodestore(),
+            limit,
```
rkuris: nit: improve readability with some local variables and a helper, something like:
```
    let start_merkle = Merkle::from(self.db.revision(start_hash).map_err(...)?);
    let end_merkle = Merkle::from(revision(end_hash).map_err(...)?);
    end_merkle.change_proof(start_key, end_key, start_merkle.nodestore(), limit)
```

It's unfortunate we didn't use tuple variants for `api::Error`. Might also be cleaner to create an enum helper that converts a `RevisionNotFound` into a `StartRevisionNotFound` or a `EndRevisionNotFound` as part of the api::Error type, but I'm not sure that's worth it.

firewood/src/merkle/mod.rs
```diff
@@ -458,6 +460,74 @@ impl<T: TrieReader> Merkle<T> {
 }
 
 impl<T: HashedNodeReader> Merkle<T> {
+    /// Generate a change proof
+    ///
+    /// # Errors
+    ///
+    /// * `api::Error::InvalidRange` - If `start_key` > `end_key` when both are provided.
```
rkuris, line 467: This is much better, perhaps the other function earlier could use the same clarity.

bernard-avalabs: I've added additional comments on errors in `DatabaseHandle::change_proof`.

ffi/src/handle.rs
bernard-avalabs: I've made this change.

PR #1638, opened 2026-01-25
feat: Change proof serialization/deserialization

bernard-avalabs:
## Why this should be merged

Second PR for change proof FFI (builds on https://github.com/ava-labs/firewood/pull/1637). This PR includes `fwd_change_proof_to_bytes` and `fwd_change_proof_from_bytes` for serialization/deserization of a change proof.

## How this works
Mostly follows the serialization/deserialization implementation in range proof. Main change is to support serializing/deserializing `BatchOp`s which are used for storing the difference between two revisions.

## How this was tested

Basic serialization and deserialization tests in `proofs_test.go`, including a round trip test where a change proof is serialized, deserialized, and serialized again, and verifying the two serialized proofs match.

### DIFF
--- a/firewood/src/proofs/de.rs
+++ b/firewood/src/proofs/de.rs
@@ -21,6 +21,7 @@ use super::{
 use crate::{
     db::BatchOp,
     merkle::{Key, Value},
+    proofs::magic::{BATCH_DELETE, BATCH_DELETE_RANGE, BATCH_PUT},
     v2::api::{FrozenChangeProof, FrozenRangeProof},
 };
 
@@ -150,14 +151,14 @@ impl Version0 for BatchOp<Key, Value> {
             .read_item::<u8>()
             .map_err(|err| err.set_item("option discriminant"))?
         {
-            0 => Ok(BatchOp::Put {
+            BATCH_PUT => Ok(BatchOp::Put {
                 key: reader.read_item()?,
                 value: reader.read_item()?,
             }),
-            1 => Ok(BatchOp::Delete {
+            BATCH_DELETE => Ok(BatchOp::Delete {
                 key: reader.read_item()?,
             }),
-            2 => Ok(BatchOp::DeleteRange {
+            BATCH_DELETE_RANGE => Ok(BatchOp::DeleteRange {
                 prefix: reader.read_item()?,
             }),
             found => Err(reader.invalid_item("option discriminant", "0, 1, or 2", found)),


ffi/proofs_test.go
```diff
@@ -412,3 +431,59 @@ func TestChangeProofCreation(t *testing.T) {
 	_, err = db.ChangeProof(root1, root2, nothing(), nothing(), changeProofLenUnbounded)
 	r.NoError(err)
 }
+
+func TestChangeProofDiffersAfterUpdate(t *testing.T) {
```
alarso16, line 435: It's not totally clear what this is testing. Wouldn't being able to commit a change proof and resulting in the correct root mean that they contain the correct data? It seems that this might just be an intermediate test until that functionality is present, but  if that's the case, will you add a `//TODO: delete this once verification and commitment are implemented`. If you think this test will continue to be necessary after that, can you argue why?

firewood/src/proofs/de.rs
```diff
@@ -72,6 +76,45 @@ impl<T: Version0> Version0 for Box<[T]> {
...
+                    Err(reader.invalid_item(
+                        "trailing bytes",
+                        "no data after the proof",
+                        format!("{} bytes", reader.remainder().len()),
+                    ))
+                }
+            }
+            found => Err(ReadError::InvalidHeader(
+                InvalidHeader::UnsupportedVersion { found },
+            )),
```
rkuris: nit: would have preferred this error is handled before the main code instead of indenting. Also a shame bool::ok_or is not stable yet. Maybe:
```
if header.version != 0 {
    return Err(ReadError::InvalidHeader(
                InvalidHeader::UnsupportedVersion { found },
            ))
}
let mut reader = ...
```

ffi/proofs_test.go
bernard-avalabs: This test is based on `TestRangeProofDiffersAfterUpdate`. Similar to `TestChangeProofEmptyDB`, its mainly to check basic functionality, namely the ability to create a serialized change proof from two non-empty revisions without error, and to verify that serialized versions of distinct change proofs are different.

It can catch serialization issues that simply checking the root hash will miss. However, `TestRoundTripChangeProofSerialization` is a more comprehensive test for serialization/deserialization.

It don't think it hurts to keep it since it is a short test. But I can remove it if you don't think it is necessary.

firewood/src/proofs/de.rs
bernard-avalabs: I've made this change.

firewood/src/proofs/de.rs
```diff
@@ -86,6 +130,41 @@ impl Version0 for FrozenRangeProof {
...
+    }
+}
+
+impl Version0 for BatchOp<Key, Value> {
+    fn read_v0_item(reader: &mut V0Reader<'_>) -> Result<Self, ReadError> {
+        match reader
+            .read_item::<u8>()
+            .map_err(|err| err.set_item("option discriminant"))?
+        {
+            0 => Ok(BatchOp::Put {
```
demosdemon: nit: you can use the constants here in the match pattern as well.

demosdemon (approved):

PR #1641, opened 2026-01-26
feat: Change proof verification and proposal creation

bernard-avalabs:
## Why this should be merged

Builds on https://github.com/ava-labs/firewood/pull/1638. Verifies a change proof and creates a proposal on the destination trie.

## How this works

Mostly follows the implementation from range proofs with some additional checks that are specific to change proofs. The interface for range proofs has a separate function for verify (`fwd_range_proof_verify`) and verify and propose (`fwd_db_verify_range_proof`). The change proof interface only has verify, but the comments explains that it also prepares a proposal. I have follow the comment by implementing proof verification and proposal creation in the verify function.

## How this was tested

A basic test was added in `proofs_test.go` that creates a change proof and verifies it and check that it does not return an error. More end-to-end tests will be added in a later PR.

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -56,7 +56,8 @@ use std::{
 };
 
 use firewood_storage::{
-    Committed, FileBacked, FileIoError, HashedNodeReader, NodeStore, NodeStoreHeader,
+    Committed, FileBacked, FileIoError, HashedNodeReader, LinearAddress, NodeStore,
+    NodeStoreHeader, TrieHash,
 };
 use parking_lot::{Condvar, Mutex, MutexGuard};
 
@@ -396,7 +397,7 @@ impl PersistLoop {
     ) -> Result<(), PersistError> {
         let result = self
             .persist_to_disk(revision)
-            .and_then(|()| self.save_to_root_store(revision));
+            .and_then(|()| self.maybe_save_to_root_store(revision));
         self.release_permits(commits_since_persist);
         result
     }
@@ -426,16 +427,25 @@ impl PersistLoop {
     }
 
     /// Saves the revision's root address to `RootStore` if configured.
-    fn save_to_root_store(&self, revision: &CommittedRevision) -> Result<(), PersistError> {
+    fn maybe_save_to_root_store(&self, revision: &CommittedRevision) -> Result<(), PersistError> {
         if let Some(ref store) = self.shared.root_store
             && let (Some(hash), Some(addr)) = (revision.root_hash(), revision.root_address())
         {
-            store.add_root(&hash, &addr).map_err(|e| {
-                error!("Failed to persist revision address to RootStore: {e}");
-                PersistError::RootStore(e.into())
-            })?;
+            save_to_root_store(store, &hash, &addr)?;
         }
 
         Ok(())
     }
+}
+
+#[crate::metrics("persist.root_store", "persist revision address to root store")]
+fn save_to_root_store(
+    store: &RootStore,
+    hash: &TrieHash,
+    addr: &LinearAddress,
+) -> Result<(), PersistError> {
+    store.add_root(hash, addr).map_err(|e| {
+        error!("Failed to persist revision address to RootStore: {e}");
+        PersistError::RootStore(e.into())
+    })
 }


ffi/proofs_test.go
```diff
@@ -487,3 +486,28 @@ func TestRoundTripChangeProofSerialization(t *testing.T) {
...
+func TestVerifyChangeProof(t *testing.T) {
+	r := require.New(t)
+	db1 := newTestDatabase(t)
+	db2 := newTestDatabase(t)
+
+	// Insert some data.
+	_, _, batch := kvForTest(10)
+	root1, err := db1.Update(batch[:5])
+	r.NoError(err)
+	root2, err := db2.Update(batch[:5])
```
alarso16: Should we ensure these roots match? Seems weird that `root2` corresponds to the second database, not the second root

ffi/src/proofs/change.rs
```diff
@@ -105,9 +107,100 @@ impl From<FrozenChangeProof> for ChangeProofContext<'_> {
...
+        start_key: Option<&[u8]>,
+        end_key: Option<&[u8]>,
+        max_length: Option<NonZeroUsize>,
+    ) -> Result<(), api::Error> {
+        if let Some(ref ctx) = self.verification {
+            if ctx.start_root == start_root
+                && ctx.end_root == end_root
+                && ctx.start_key.as_deref() == start_key
+                && ctx.end_key.as_deref() == end_key
+                && ctx.max_length == max_length
```
rkuris: Curious as to why we're checking max_length equality here. The others I understand but why would this matter?

ffi/src/proofs/change.rs
```diff
@@ -105,9 +107,100 @@ impl From<FrozenChangeProof> for ChangeProofContext<'_> {
...
+                // already verified with the same context
+                return Ok(());
+            }
+
+            return Err(api::Error::ProofError(ProofError::ValueMismatch));
+        }
+
+        let batch_ops = self.proof.batch_ops();
+        if batch_ops.is_empty() {
+            return Err(api::Error::ProofError(ProofError::Empty));
```
rkuris: Is this really an error? It's possible that within a specific start_key and end_key there are no changes between two revisions, in which case batch_ops would be empty but this is still a valid change proof.

ffi/src/proofs/change.rs
```diff
@@ -105,9 +107,100 @@ impl From<FrozenChangeProof> for ChangeProofContext<'_> {
...
+        start_key: Option<&[u8]>,
+        end_key: Option<&[u8]>,
+        max_length: Option<NonZeroUsize>,
+    ) -> Result<(), api::Error> {
+        self.verify(start_root, end_root, start_key, end_key, max_length)?;
+        if self.proposal_state.is_some() {
+            return Ok(());
+        }
+        let proposal = db.apply_change_proof_to_parent(start_root.into(), &self.proof)?;
+        self.proposal_state = Some(ProposalState::Immutable(proposal.handle));
```
rkuris: nit: There are a lot of states here and I would have preferred a generic state.

ffi/src/handle.rs
```diff
@@ -330,6 +330,39 @@ impl DatabaseHandle {
...
+    ///
+    /// Returns a `ParentNotLatest` error if `start_hash` does not match the current root hash.
+    /// Returns a `LatestIsEmpty` error if the trie is empty. A range proof should be used in
+    /// this case.
+    pub fn apply_change_proof_to_parent(
+        &self,
+        start_hash: HashKey,
+        change_proof: &FrozenChangeProof,
+    ) -> Result<CreateProposalResult<'_>, api::Error> {
+        // Verify that the current hash is the start_hash
```
rkuris: I don't think this is correct to do here.

Imagine if we have R1 and R2, and two change proofs for different start and end keys, CP1 and CP2, each covering half of the keyspace.

If you apply CP1 to R1 you get some hash that is no longer R1, and therefore cannot apply CP2.

ffi/src/proofs/change.rs
bernard-avalabs: I followed what was done in range proofs. I assume it is to ensure that this is an exact match from the previous verify. I can remove the max_length check if that is your preference.

ffi/src/proofs/change.rs
bernard-avalabs: I'll remove this check.

ffi/src/proofs/change.rs
bernard-avalabs: I also followed what was done in range proofs here. What do you have in mind for a generic state?

ffi/src/proofs/change.rs
rkuris: More importantly, please add a test to verify this will work.

ffi/src/handle.rs
bernard-avalabs: Perhaps I misunderstood the interface. In my tests (mostly in the next PR), I have been passing the committed hash from the last range as the `start_hash` in the next verify call, not the `start_root` in the change proof. Here is a snippet of a test:

```
	startKey := nothing()
	for {
		proof, err := db1.ChangeProof(root1, root1_updated, startKey, nothing(), changeProofLenTruncated)
		r.NoError(err)
		t.Cleanup(func() { r.NoError(proof.Free()) })

		// Verify the proof
		r.NoError(db2.VerifyChangeProof(proof, root2, root1_updated, startKey, nothing(), changeProofLenTruncated))

		// Commit the proof
		root2, err = db2.VerifyAndCommitChangeProof(proof, root2, root1_updated, startKey, nothing(), changeProofLenTruncated)
		r.NoError(err)

		// Find the next start key
		nextRange, err := proof.FindNextKey()
		r.NoError(err)
		if nextRange == nil {
			break
		}
		startKey = maybe{
			hasValue: true,
			value:    append([]byte{}, nextRange.StartKey()...), // copy to new slice to avoid use-after-free
		}
		r.NoError(nextRange.Free())
	}
```

In the above snippet, I would retain root2 which is returned from a VerifyAndCommitChangeProof and use it as the start root for the next call to VerifyChangeProof.

I assumed that this would be a helpful check to ensure that no other operation was committed to the trie in between  committing two ranges of a change proof.

Should this just always use the current root hash?

ffi/src/proofs/change.rs
rkuris: It's okay the way it is. A comment might be helpful here.

I wonder if it makes more sense to implement `PartialEq` instead, to move the comparison logic out of this function.

ffi/proofs_test.go
bernard-avalabs: I've changed the databases to dbA and dbB to make it more clear. I've also added the root check.

ffi/src/handle.rs
rkuris: Let's chat in person.

We should also add a test that limits the size of the change proof to 1.

bernard-avalabs: Summary from discussion: 
- Change proof ranges may be applied out of order.
- The start hash parameter should be the start root from the change proof. 
- Verification of the the start hash is not needed here. Correctness is handled by the above layer.

Additional test to add:
- Each range of a change proof is returning the right size.
- Empty proof is handled correctly.

bernard-avalabs: I'm going to defer the proof size test to the next PR. The size of a change proof cannot be easily checked in golang because it is opaque. With commit from the next PR, I can manually verify by checking the key returned from `find_next_key` and also the root hash after committing a change proof range.

For the same reason, the empty proof test does not check the proof size. I have verified the proof is empty manually, and I will add an explicit test that shows that committing the empty proof does not change the root hash in the next PR.

ffi/src/proofs/change.rs
rkuris: I'm not crazy about the readability and invariants here. Can you create a table of of what the following states mean and which ones are valid and invalid?

|verification|proposal_state|explanation|
|--|--|--|
|None|None|They get created in this state?|
|None|Immutable|Invalid?|
|None|Committed|Invalid?|
|Some(VerificationContext)|None|Verified but not proposed?|
|Some(VerificationContext)|Immutable|Proposed but not committed?|
|Some(VerificationContext)|Committed|Proposed and committed?|

Some of these states are invalid (maybe?) so we either must (a) structure it so they cannot be represented, or (b) add lots of state checks and assertions to prevent bad state.

(a) is definitely more rusty, here's one way to do that (assuming I'm right above, but I could easily be wrong):
```
pub struct ChangeProofContext<'db, State> {
    state: State<'db>,
    proof: FrozenChangeProof,
}

struct InitialState<'_>;
struct VerifiedState<'_>(VerificationContext));
struct ProposedState<'db>(crate::ProposalHandle<'db>);
struct CommittedState<'db>(HashKey);
```

I also don't really like "initialState" in this design either. I think I'd prefer to only have these contain verified proofs. Wrapping it in a ChangeProofContext when the verification is incomplete seems incorrect. I think this leaves you with:
 - a `TryInto` implementation from `FrozenChangeProof` to `ChangeProofContext<'db, VerifiedState>`
 - a `propose()` method on `ChangeProofContext<'db, VerifiedState>`
 - a `commit()` method on `ChangeProofContext<'db, ProposedState>`

If you find you really still need 'verify_and_propose' then it should go through both states. This prevents anyone from being able to write code that creates a proposal from an unverified CP.

I'm also a bit worried about the `db` lifetime and what happens if you make some of these calls with a different db (propose on db1, commit on db2). This could be solved by keeping a reference to the db inside the context and not specifying it after it's created.

bernard-avalabs: I've changed the interface based on your feedback. One difference is that, instead of following the typestate pattern with different states for ChangeProofContext, I've created separate structs for each context: ChangeProofContext, VerifiedProofContext, ProposedProofContext, etc. I think this makes it slightly more readable given the matching ChangeProofResult, VerifiedProofResult, and ProposedProofResult, and more cleanly distinguishes the different states in Golang.

I've also kept ChangeProofContext to represent an unverified change proof. This allows (without more interface changes) a change proof to be created without calling verify on it.  It shouldn't be possible to use an unverified change proof directly as it must be verified first before a proposal can be created from it.

rkuris:
I think this is a lot better, but I think we can even do more here. I'll approve this PR so you can merge it, but we should consider the following for the next release.

I don't like the current split state variables (`ChangeProofContext` -> `VerifiedChangeProofContext` -> `ProposedChangeProofContext`). I think we can do this with a single stateful struct that supports idempotent verification, proposal preparation, commit, and next-range introspection.

I think this simplifies the design a bit, as there are odd edge cases in this design. An example might be that verification is not idempotent. Calling verify again on the same ChangeProofContext returns ProofIsNone because the proof was consumed.

It might make the rust code a tiny bit more complicated but is likely to simplify the ffi interface significantly. I'd have to try it to see.

Maybe something like this:

```
pub struct ChangeProofContext<'db> {
    proof: FrozenChangeProof,

    // Cached verification context to support idempotent verify calls.
    verification: Option<VerificationContext>,

    // Proposal/commit lifecycle state.
    proposal_state: ProposalState<'db>,
}

struct VerificationContext {
    start_root: HashKey,
    end_root: HashKey,
    start_key: Option<Box<[u8]>>,
    end_key: Option<Box<[u8]>>,
    max_length: Option<NonZeroUsize>,
}

enum ProposalState<'db> {
    // No prepared proposal yet.
    Unprepared,

    // Proposal prepared against a DB snapshot, not committed yet.
    Proposed(crate::ProposalHandle<'db>),

    // Already committed; value is resulting root (None => empty trie).
    Committed(Option<HashKey>),
}
```

bernard-avalabs:
> Maybe something like this:
> 
> ```
> pub struct ChangeProofContext<'db> {
>     proof: FrozenChangeProof,
> 
>     // Cached verification context to support idempotent verify calls.
>     verification: Option<VerificationContext>,
> 
>     // Proposal/commit lifecycle state.
>     proposal_state: ProposalState<'db>,
> }
> 
> struct VerificationContext {
>     start_root: HashKey,
>     end_root: HashKey,
>     start_key: Option<Box<[u8]>>,
>     end_key: Option<Box<[u8]>>,
>     max_length: Option<NonZeroUsize>,
> }
> 
> enum ProposalState<'db> {
>     // No prepared proposal yet.
>     Unprepared,
> 
>     // Proposal prepared against a DB snapshot, not committed yet.
>     Proposed(crate::ProposalHandle<'db>),
> 
>     // Already committed; value is resulting root (None => empty trie).
>     Committed(Option<HashKey>),
> }
> ```

I'll merge the current PR and will create a new issue to address your new feedback. (Update: https://github.com/ava-labs/firewood/issues/1709)

Perhaps we can discuss your new interface suggestion in person, as it seems very similar to the interface I had in my initial implementation.

PR #1642, opened 2026-01-26
chore(track-performance): local iteration

Elvis339:
Add local developer tooling for triggering C-Chain reexecution benchmarks. Includes just recipe, gh CLI in nix devshell, and usage documentation in METRICS.md.

Note: should be merged after https://github.com/ava-labs/firewood/pull/1493

## Why this should be merged

Enables developers to trigger C-Chain reexecution benchmarks from their local environment without navigating GitHub Actions UI. This complements the `track-performance.yml` workflow by providing a CLI interface.

## How this works

The `bench-cchain` recipe:
1. Validates inputs (predefined test name OR custom block params)
2. Triggers `track-performance.yml` via `gh workflow run`
3. Polls for workflow registration (handles GitHub's async dispatch)
4. Watches the run until completion

Environment variables support custom configurations:
- `FIREWOOD_REF` - specific commit/tag to benchmark
- `AVALANCHEGO_REF` - AvalancheGo version to test against
- `START_BLOCK`, `END_BLOCK`, `BLOCK_DIR_SRC` - custom block ranges

## How this was tested

```
nix run ./ffi#gh -- gh auth login

GH_TOKEN=$(gh auth token) RUNNER=avalanche-avalanchego-runner-2ti FIREWOOD_REF=v0.0.18 CONFIG=firewood START_BLOCK=1 END_BLOCK=100 BLOCK_DIR_SRC=cchain-mainnet-blocks-200-ldb just bench-cchain
```

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -271,11 +271,8 @@ impl RevisionManager {
 
             // We reap the revision's nodes only if `RootStore` does not exist.
             if self.root_store.is_none() {
-                // This `try_unwrap` is safe because nobody else will call `try_unwrap` on this Arc
-                // in a different thread, so we don't have to worry about the race condition where
-                // the Arc we get back is not usable as indicated in the docs for `try_unwrap`.
-                // This guarantee is there because we have a `&mut self` reference to the manager, so
-                // the compiler guarantees we are the only one using this manager.
+                // The warning in the docs for `Arc::try_unwrap` does not apply here
+                // because `original` is retained and not immediately dropped.
                 match Arc::try_unwrap(oldest) {
                     Ok(oldest) => oldest.reap_deleted(&mut header)?,
                     Err(original) => {


justfile
```diff
@@ -168,3 +168,93 @@ release-step-refresh-changelog tag:
...
+    fi
+    
+    test="{{ test }}"
+    
+    # Validate: need either test name OR custom block params
+    if [[ -z "$test" && -z "${START_BLOCK:-}" ]]; then
+        echo "error: Provide a test name or set START_BLOCK, END_BLOCK, BLOCK_DIR_SRC" >&2
+        echo "" >&2
+        echo "Predefined tests:" >&2
+        echo "  firewood-101-250k, firewood-33m-33m500k, firewood-33m-40m" >&2
```
rkuris, line 214: For consistency:
```suggestion
        echo "  firewood-101k-250k, firewood-33m-33m500k, firewood-33m-40m" >&2
```

I think this would also be a lot better if the "predefined tests" list was in one place, probably in the just file, making maintenance easier, otherwise they're sure to get out of sync if someone adds or renames one.

justfile
```diff
@@ -168,3 +168,93 @@ release-step-refresh-changelog tag:
...
+    
+    if [[ -z "$run_id" ]]; then
+        echo "warning: Could not find run ID. Check manually at:"
+        echo "  https://github.com/ava-labs/firewood/actions/workflows/track-performance.yml"
+        exit 0
+    fi
+    
+    echo ""
+    echo "Workflow triggered. Run ID: $run_id"
+    echo "https://github.com/ava-labs/firewood/actions/runs/$run_id"
```
rkuris:
```suggestion
    echo "Monitor this workflow with cli: $GH run watch $run_id"
    echo " or with this URL: https://github.com/ava-labs/firewood/actions/runs/$run_id"
```

justfile
```diff
@@ -168,3 +168,93 @@ release-step-refresh-changelog tag:
...
+    echo ""
+    
+    $GH run watch "$run_id"
```
rkuris, line 267: Is it better to use this or watch it on the web? Maybe we shouldn't run this and let them choose.

Elvis339: Since the script already resolves and wraps the gh CLI (falling back to nix if needed), waiting by default keeps the experience consistent for users who have the tooling set up. The URL is printed before the watch starts, so users can open it in their browser and Ctrl+C to exit if they prefer the web UI.

https://github.com/ava-labs/firewood/blob/12e170be6390e94fc9b90774a5de9912e9a826d8/justfile#L190

Does that work for you, or would you prefer adding a something like `no-wait` flag?

METRICS.md
```diff
@@ -298,3 +298,104 @@ rate(firewood_space_from_end[5m])
...
+
+```bash
+nix run ./ffi#gh -- gh auth login
+export GH_TOKEN=$(gh auth token)
+
+# Predefined test
+just bench-cchain firewood-101-250k
+
+# With specific Firewood version
+FIREWOOD_REF=v0.1.0 just bench-cchain firewood-33m-40m
```
RodrigoVillar: I believe you mentioned that `just` supports named parameters. If so, could you please used named parameters wherever possible?

METRICS.md
```diff
@@ -298,3 +298,104 @@ rate(firewood_space_from_end[5m])
...
+
+**Custom mode variables** (when no test specified):
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `CONFIG` | firewood | VM config (firewood, hashdb, etc.) |
+| `START_BLOCK` | required | First block number |
+| `END_BLOCK` | required | Last block number |
+| `BLOCK_DIR_SRC` | required | S3 block directory |
+| `CURRENT_STATE_DIR_SRC` | - | S3 state directory (empty = genesis run) |
```
RodrigoVillar: Given that this is a repetition of:

https://github.com/ava-labs/firewood/blob/06c4d0a373adcf3957761acac33b6028be5cdcab/scripts/bench-cchain-reexecution.sh#L16-L31

Would it make sense to just tell clients to look at the documentation in `scripts/bench-cchain-reexecution.sh` for all possible options?

METRICS.md
RodrigoVillar, line 1: One of the requirements of adding GAB to Firewood was that developers would be able to create a branch from main, push a commit, and run GAB against their new branch. From this though, it's not evident that it's supported and that there's no documentation for this common workflow. Could we add a section walking through what running benchmarks against a branch would look like?

Elvis339: Yes but it's intentionally on-demand for now. CI integration requires understanding wider understanding from the team these benchmarks run 5+ hours on dedicated hardware. For example, some self-hosted runners surfaced deeper issue that needs resolving first. 

Feature branch benchmarks work today: select your branch in the UI dropdown or use --ref via CLI. Results appear at dev/bench/{branch}/.

METRICS.md
Elvis339: Good recommendation, updated. Thank you

justfile
RodrigoVillar, line 1: q: why is the `justfile` not using `./scripts/bench-cchain-reexecution.sh`?

Elvis339: They serve different purposes:
- just bench-cchain → triggers track-performance.yml → results published to GitHub Pages for the current branch
- bench-cchain-reexecution.sh → triggers AvalancheGo directly → downloads results locally

The justfile ensures benchmarks appear in Firewood's GitHub Pages trends. The script is used by the workflow (and can be used standalone if you don't need publishing).

Documented here: https://github.com/ava-labs/firewood/blob/401318f891df84f0982a484d276e61fc3901edaa/justfile#L172

RodrigoVillar: Why can't the script and `justfile` logic be merged into one common script? To my understanding, the only difference here is that one command publishes the benchmark results while the other downloads them locally, implying that an if-else statement could be used to deduplicate code here.

Elvis339: Fair point. It could be merged, but there are tradeoffs:

Current design:
- `bench-cchain-reexecution.sh` → single purpose: trigger AvalancheGo, wait, download
- `track-performance.yml` → orchestrates: calls script + publishes to GitHub Pages
- justfile → triggers the workflow (user entry point)

If merged:
- Script would need to know about GitHub Pages publishing, workflow context, branch-based data directories
- Adds conditional logic that only matters in CI context
- Script becomes harder to use standalone for quick local tests

The workflow already uses the script internally. The separation keeps the script simple and reusable while the workflow handles CI-specific orchestration.

That said, I see this as a **temporary solution**. The more I work on this, the less value I see in the justfile wrapper the scheduled runs handle official tracking, and for development, truly local iteration (running benchmarks on local hardware without the `gh` CLI roundtrip) would be far more useful for the Firewood team.

I originally added the justfile recipe to make it easy to test the workflow integration during initial setup, but long-term we'd benefit more from investing in local benchmark tooling rather than polishing this remote trigger path.

Open to simplifying or removing it if it's adding more confusion than value.

RodrigoVillar: Could you create an issue to come back and possibly deduplicate code here? My issue is around parameter validation which we do in both the `justfile` and `scripts/bench-cchain-reexecution.sh` but I don't think it's a blocker to this PR.

> That said, I see this as a temporary solution. The more I work on this, the less value I see in the justfile wrapper the scheduled runs handle official tracking, and for development, truly local iteration (running benchmarks on local hardware without the gh CLI roundtrip) would be far more useful for the Firewood team.

My ideal workflow would be to effortlessly run the reexecution tests with our GH runners, so I'm not sure I agree with the local iteration but I digress.

PR #1643, opened 2026-01-27
ci: add benchmark workflow

RodrigoVillar:
## Why this should be merged

A key component of measuring our performance is seeing trends in our benchmarks. Currently, we have benchmarks in Firewood that are not part of our CI. By automatically running these benchmarks on a daily basis, we can start getting an idea of which components are getting faster/slower. 

In the future, these benchmarks can be consumed by tools such as Github-Action-Benchmark but for now, the results are uploaded as artifacts.

## How this works

Adds a benchmark workflow which runs daily and can be triggered via a manual dispatch; the benchmarks run are as follows:

- `hashops`
- `serializer`

## How this was tested

https://github.com/ava-labs/firewood/actions/runs/21412797668/job/61653777101?pr=1643

### DIFF
--- a/.github/check-license-headers.yaml
+++ b/.github/check-license-headers.yaml
@@ -58,7 +58,7 @@
       "clippy.toml",
       "**/tests/compile_*/**",
       "justfile",
-      "scripts/run-just.sh",
+      "scripts/**",
     ],
   }
 ]
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -30,6 +30,32 @@ jobs:
         run: |
           cp -rv target/doc/* ./_site
           cp -rv docs/assets ./_site
+      # GitHub Pages deploys from a single source (overwrites, doesn't merge).
+      # Benchmark history lives on benchmark-data branch (for append-only storage).
+      # We merge both into _site/ so a single deployment serves docs + benchmarks.
+      # See track-performance.yml for how benchmark data is collected and stored.
+      #
+      # Structure on benchmark-data branch (see track-performance.yml for how this is populated):
+      #   bench/              - Official benchmark history (main branch only)
+      #   dev/bench/{branch}/ - Feature branch benchmarks (experimental)
+      - name: Include benchmark data
+        run: |
+          # Fetch benchmark-data branch (may not exist on first run)
+          if ! git fetch origin benchmark-data 2>/dev/null; then
+            echo "No benchmark-data branch yet, skipping"
+            exit 0
+          fi
+          
+          # Extract benchmark directories from benchmark-data branch into current dir
+          # These may not exist if no benchmarks have been run yet for that category
+          git checkout origin/benchmark-data -- dev 2>/dev/null || echo "No dev/ directory (no feature branch benchmarks yet)"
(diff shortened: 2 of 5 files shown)


.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,35 @@
+name: firewood-benchmarks
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
```
Elvis339, line 6: Should there be some inputs that control the benchmarks?

.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
```
Elvis339, line 14: Should this run on a dedicated/self-hosted runner instead of ubuntu-latest? Shared runners have variable performance which can make benchmark results unreliable for tracking regressions across runs.

.github/workflows/benchmarks.yaml
RodrigoVillar: I think the current parameter setup (i.e. hardcoded) is fine. I don't have much context into these benchmarks but would prefer we start tracking them + we can expand their inputs in the future.

.github/workflows/benchmarks.yaml
RodrigoVillar: What dedicated/self-hosted runners are available to use? Open to switching but I'm unaware of the options.

Elvis339: For reference, these self-hosted runners are available (managed via ARC):

| Runner | Spec | Storage |
|--------|------|---------|
| avalanche-avalanchego-runner | 4 vCPU, 16 GB | EBS |
| avago-runner-m6i-4xlarge-ebs-fast | 16 vCPU, 64 GB | Fast EBS |
| avago-runner-i4i-4xlarge-local-ssd | 16 vCPU, 128 GB | Local NVMe |

Worth reaching out to security/GitHub admin for a full list. Not a blocker tho ubuntu-latest is fine to start.

Elvis339:
Having benchmarks in CI is a great step forward.

One question tho the goal is seeing trends, but artifacts alone don't provide visualization or comparison. Was there a reason for deferring `github-action-benchmark` to a follow-up? Since the integration is fairly small, including it here would complete the goal.

RodrigoVillar:
With PRs such as https://github.com/ava-labs/firewood/pull/1645 still in review, I didn't want to integrate GAB only for those changes to be modified. If you want, you can take ownership of adding these benchmarks to GAB once all prerequisite PRs are merged in.

.github/workflows/benchmarks.yaml
Elvis339: Full list of self-hosted runners: https://github.com/ava-labs/devops-argocd/blob/main/base/system/actions-runners/action-runner.yaml

rkuris (approved):

.github/workflows/benchmarks.yaml
RodrigoVillar: Going to stick with `ubuntu-latest` for now. Once GAB starts tracking this workflow as well, we can switch the runner.

Issue #1639, opened 2026-01-25
ci: Add automatic benchmark triggers on main


Elvis339:
**Is your feature request related to a problem? Please describe.**

The C-Chain Performance Tracking workflow is currently manual-only (`workflow_dispatch`). Without automated runs, performance regressions may go unnoticed until someone remembers to trigger a benchmark.

**Describe the solution you'd like**

Add scheduled (e.g., daily) benchmark runs to enable continuous performance tracking and regression detection.

**Open questions:**

### 1. Which test configuration?

The benchmark supports various starting states (from block 100 to 69M+) and block ranges. Key trade-offs:

| Factor | Early blocks (< 1M) | Mid-range (10M-40M) | Late blocks (50M+) |
|--------|---------------------|---------------------|---------------------|
| Trie state size | Small | Medium-Large | Very large |
| Runtime | Minutes | Hours (>5h) | Days |
| Cost | $ | $$$ | $$$$$$ |

Additional dimensions:
- Block range size: Narrow (500k blocks) vs wide (7M+ blocks)
- Archive vs full node: Archive retains all historical state; full prunes

### 2. Which runner?

Available runners:

| Runner | Spec | Storage |
|--------|------|---------|
| `avalanche-avalanchego-runner` | 4 vCPU, 16 GB | EBS |
| `avago-runner-m6i-4xlarge-ebs-fast` | 16 vCPU, 64 GB | Fast EBS |
| `avago-runner-i4i-4xlarge-local-ssd` | 16 vCPU, 128 GB | Local NVMe |

### 3. Frequency?

| Option | Trade-off |
|--------|-----------|
| Daily | Catches regressions quickly, higher cost |
| Weekly | Lower cost, delayed feedback |
| Rotating | e.g., daily short test + weekly long test balances coverage and cost |

**Describe alternatives you've considered**

- Manual only (current): Full control, but easy to forget
- Trigger on push to main: Reactive, but potentially noisy and expensive
- Scheduled: Predictable cost, consistent trend data

**Additional context**

State snapshots and block data are available in S3 at various checkpoints (100, 1M, 10M, 20M, 30M, 40M, 50M+), enabling flexible test configuration without needing to define every combination upfront.

Requires https://github.com/ava-labs/firewood/pull/1493 to be merged before proceeding.

rkuris:
[...]

PR #1645, opened 2026-01-28
ci: add scheduled benchmark runs (#1639)

Elvis339:
## Why this should be merged
This patch is closing [#1639](https://github.com/ava-labs/firewood/issues/1639) it adds automated benchmark triggers to catch performance regressions without manual intervention.

- Daily (40M→41M): Quick 1M block test for fast regression detection
- Weekly (50M→60M): Comprehensive 10M block stress test
- Runner: avago-runner-i4i-2xlarge-local-ssd for consistent NVMe performance

Note: should be merged after https://github.com/ava-labs/firewood/pull/1493

## How this works
1. configure-benchmark job: Shell case statement matches cron schedule → benchmark parameters (block range, runner, timeout)
2. benchmark job: Triggers AvalancheGo's C-Chain reexecution workflow with resolved parameters
3. github-action-benchmark: Publishes results to benchmark-data branch (bench/ for main, dev/bench/{branch}/ for feature branches)
4. workflow_run trigger: After benchmark completes, gh-pages.yaml auto-rebuilds to include new results

Manual dispatch still works via workflow inputs.

## How this was tested
- [x] Verified S3 paths exist
- [x] Confirmed storage (~443 GB) fits i4i.4xlarge capacity (3.75 TB)
- [x] Manual workflow dispatch validation (pending)

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,35 @@
+name: firewood-benchmarks
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: "maxperf-ethhash-logger"
+          cache-all-crates: true
+          cache-workspace-crates: true
+      - name: Cargo Fetch (run `cargo generate-lockfile` if this fails)
+        run: cargo fetch --locked --verbose
+      - name: Build
+        run: cargo build --frozen --profile maxperf --features ethhash,logger --workspace --benches
+      - name: Run firewood crate benchmarks
+        run: cargo bench --frozen --profile maxperf --features ethhash,logger -p firewood --bench hashops -- --noplot
+      - name: Run storage crate benchmarks
+        run: cargo bench --frozen --profile maxperf --features ethhash,logger -p firewood-storage --bench serializer -- --noplot
+      - name: Upload benchmark results
+        uses: actions/upload-artifact@v4
+        with:
+          name: benchmark-results
+          path: target/criterion


.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,186 @@
...
+        id: store
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: C-Chain Reexecution with Firewood
+          tool: 'customBiggerIsBetter'
+          output-file-path: ./results/benchmark-output.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          summary-always: true
+          auto-push: true
+          fail-on-alert: true
```
RodrigoVillar, line 163: q: what's the rationale for setting `summary-always` and `fail-on-alert` to true?

.github/benchmark-schedules.json
```diff
@@ -0,0 +1,29 @@
+{
+  "schedule": {
+    "include": [
+      {
+        "name": "daily-40m-41m",
+        "cron": "0 5 * * *",
+        "runner": "avago-runner-i4i-4xlarge-local-ssd",
```
rkuris: This instance type is likely wasting a ton of space and is double the cost of i4i-2xlarge and 4x the cost of i4i-xlarge. The latter has 937G of local disk. Let's lower the cost of this by using smaller instances.

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,186 @@
...
+    - cron: '0 5 * * *'
+    # Weekly Sunday at 05:00 UTC (00:00 ET) - 10M blocks: 50M → 60M
+    - cron: '0 5 * * 0'
+  workflow_dispatch:
+    inputs:
+      firewood:
+        description: 'Firewood commit/branch/tag to test (leave empty to use the commit that triggered the workflow)'
+        default: ''
+      libevm:
+        description: 'libevm commit/branch/tag to test (leave empty to skip)'
```
rkuris: What does 'skip' mean here? Does that mean use main/master?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,186 @@
...
+        default: ''
+      avalanchego:
+        description: 'AvalancheGo commit/branch/tag to test against'
+        default: 'master'
+      test:
+        description: 'Predefined test (leave empty to use custom parameters below)' # https://github.com/ava-labs/avalanchego/blob/a85295d87193b30ff17c594680dadd6618022f5e/scripts/benchmark_cchain_range.sh#L63
+        default: ''
+      config:
+        description: 'Config (e.g., firewood, hashdb)'
+        default: ''
```
rkuris: Should this be 'firewood'? What happens if none is specified in the json?

Why do we even need a default here? Same with many of the others. Is it required?

.github/workflows/track-performance.yml
```diff
@@ -0,0 +1,186 @@
...
+      - uses: actions/checkout@v4
+      - name: Resolve benchmark config
+        id: resolve
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');
+            let matrix;
+
+            if (context.eventName === 'schedule') {
```
rkuris: Is this true when fired from the scheduled cron job and false otherwise?

.github/benchmark-schedules.json
Elvis339: Infra team opened Jira ticket for this, we can track it here: https://ava-labs.atlassian.net/browse/INFRA-6420

.github/workflows/track-performance.yml
Elvis339: `default: ''` isn't strictly required omitting it has the same behavior (empty string when not provided). We use it for consistency with [AvalancheGo's pattern](https://github.com/ava-labs/avalanchego/blob/be457390191174c6c9888b21121def18b1c5fc88/.github/workflows/c-chain-reexecution-benchmark-container.yml#L9).

Yes you're right in this context default value for config should be Firewood.

.github/workflows/track-performance.yml
Elvis339: Yes, `context.eventName` returns the trigger type:
- `schedule` when fired by cron
- `workflow_dispatch` when triggered manually
- `push`, `pull_request`, etc. for other triggers

.github/workflows/track-performance.yml
Elvis339: Both are about visibility and catching regressions early. summary-always ensures results are always visible in the job summary. fail-on-alert ensures we don't silently regress - if performance drops significantly, the workflow fails rather than quietly recording bad data.

Elvis339:
> Nit: Overall, this seems like a lot of somewhat-fragile code to parse json instead of just specifying the parameters in the normal github style.

Updated was trying to stay consistent with [AvalancheGo's JSON config approach](https://github.com/ava-labs/avalanchego/blob/master/.github/workflows/c-chain-reexecution-benchmark-container.json) but I agree with you inline with native GitHub outputs feels more natural here. No strong need for consistency on this at this point in time.

.github/workflows/track-performance.yml
Elvis339: Skip means using libevm defined in AvalancheGo, I updated description.

Elvis339:
@rkuris while testing the scheduled benchmark workflow, I uncovered a couple of issues:
1. `v0.1.0` has a broken Nix build
The `flake.nix` at that tag tries to read `workspaceCargoToml.workspace.package.version`, but that field doesn't exist in our `Cargo.toml`. This was already fixed on `main` (reads from `ffiCargoToml.package.version` instead), but the tag is frozen with the bug.

When it's time to cut the next release, this will naturally resolve. Alternatively, we could cut a quick `v0.1.1` patch to get performance trends.

2. GitHub API quirk with commit SHAs
Discovered that `gh workflow run --ref` only accepts branch/tag names, not commit SHAs. Added a note to the script documenting this. Doesn't affect scheduled runs since they use master.

Let me know if you'd like me to prepare a `v0.1.1` patch or if we're fine waiting for the next release cycle.

.github/workflows/track-performance.yml
RodrigoVillar: > if performance drops significantly

Hmm what does it mean for performance to drop significantly?

Elvis339:
Test case manual dispatch with custom dependencies

Command:
```
gh workflow run track-performance.yml \
  --ref es/scheduled-perf-tracking \
  -f test=firewood-101-250k \
  -f runner=avago-runner-i4i-2xlarge-local-ssd \
  -f firewood=v0.0.18 \
  -f avalanchego=firewood-benchmark-base
```

Run: https://github.com/ava-labs/firewood/actions/runs/21489879261

- Predefined test: firewood-101-250k
- Runner: avago-runner-i4i-2xlarge-local-ssd
- Firewood: v0.0.18
- AvalancheGo: firewood-benchmark-base branch

Graph: https://ava-labs.github.io/firewood/dev/bench/es-scheduled-perf-tracking/

.github/workflows/track-performance.yml
Elvis339: Good question, I've added explicit `alert-threshold: "150%"`  to define what "significant" means - the workflow will fail if performance degrades by more than `50%` compared to the previous run.

`fail-on-alert` signals that the workflow failed specifically because of a performance regression (vs. an infrastructure issue), and `alert-threshold` defines the sensitivity. We can tune this based on observed variance once we have more data points.

.github/benchmark-schedules.json
Elvis339: New runner added and updated.

.github/workflows/track-performance.yml
Elvis339: Workflow changed this comment is no longer relevant.

.github/workflows/track-performance.yml
```diff
@@ -33,17 +38,69 @@ on:
...
+                echo "start-block=40000001" >> "$GITHUB_OUTPUT"
+                echo "end-block=41000000" >> "$GITHUB_OUTPUT"
+                echo "block-dir-src=cchain-mainnet-blocks-40m-50m-ldb" >> "$GITHUB_OUTPUT"
+                echo "current-state-dir-src=cchain-current-state-firewood-40m" >> "$GITHUB_OUTPUT"
+                echo "runner=avago-runner-i4i-2xlarge-local-ssd" >> "$GITHUB_OUTPUT"
+                echo "timeout-minutes=720" >> "$GITHUB_OUTPUT"
+                ;;
+              "0 5 * * 6")  # Weekly Saturday at 05:00 UTC: 10M blocks (50M → 60M)
+                echo "name=weekly-50m-60m" >> "$GITHUB_OUTPUT"
+                echo "config=firewood" >> "$GITHUB_OUTPUT"
```
rkuris: most of these are the same. Can we put the common ones outside the event.schedule case statement? This would make the differences more obvious to a casual reader.

.github/workflows/track-performance.yml
```diff
@@ -3,13 +3,18 @@
 name: C-Chain Reexecution Performance Tracking
 
 on:
+  schedule:
+    # Daily at 05:00 UTC (00:00 ET) - 1M blocks: 40M → 41M
```
rkuris, line 8: Please add a comment that says that these schedules are matched later to set the benchmark parameters, and consider adding a comment down below that says these are matched exactly from the strings here.

.github/workflows/track-performance.yml
```diff
@@ -102,42 +159,35 @@ jobs:
           summary-always: true
           auto-push: true
           fail-on-alert: true
+          alert-threshold: "150%"
```
rkuris, line 164: What is the default, and why did we choose 150%?

Elvis339: [Default is 200%](https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#alert-threshold-optional) we chose 150% to be more sensitive to regressions it alerts if performance degrades by more than 50% from the previous run.

.github/workflows/track-performance.yml
Elvis339: Sure, extracted these two defaults. 

```
echo "config=firewood" >> "$GITHUB_OUTPUT"
echo "runner=avago-runner-i4i-2xlarge-local-ssd" >> "$GITHUB_OUTPUT"
```

PR #1646, opened 2026-01-30
feat: Change proof commit and find next key

bernard-avalabs:
## Why this should be merged

Builds on https://github.com/ava-labs/firewood/pull/1641. This is likely the last in the current series of Change Proof PRs. It implements `fwd_db_verify_and_commit_change_proof` and `fwd_change_proof_find_next_key`. 

## How this works

Mostly follows the implementation from range proofs with some additional checks that are specific to change proofs. The find next key approach is the same basic one used in range proofs and should later be replaced.

## How this was tested

Added end-to-end tests for committing a change proof, applying multiple rounds of change proofs with different start keys determined by calling find next key, and testing change proofs with deletes instead of just puts.

### DIFF
--- a/ffi/src/proofs/change.rs
+++ b/ffi/src/proofs/change.rs
@@ -1,6 +1,7 @@
 // Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE.md for licensing terms.
 
+use std::convert::Into;
 use std::num::NonZeroUsize;
 
 use firewood_metrics::firewood_increment;
@@ -191,7 +192,7 @@ impl VerifiedChangeProofContext {
             return Err(api::Error::ProofError(ProofError::ProofIsNone));
         };
         let proposal = db.apply_change_proof_to_parent(self.params.start_root.into(), &proof)?;
-        let root_hash = proposal.handle.root_hash().map(std::convert::Into::into);
+        let root_hash = proposal.handle.root_hash().map(Into::into);
         Ok(ProposedChangeProofContext {
             proof,
             db,
@@ -225,11 +226,6 @@ impl<'db> ProposedChangeProofContext<'db> {
             return Ok(None);
         };
 
-        if self.root_hash.as_ref() == Some(&self.end_root) {
-            // already at the target root, so we are done
-            return Ok(None);
-        }
-
         if self.proof.end_proof().is_empty() {
             // unbounded, so we are done
             return Ok(None);
@@ -252,12 +248,12 @@ impl<'db> ProposedChangeProofContext<'db> {
         };
 
         let metrics_cb = |commit_time: coarsetime::Duration| {
-            firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis());
-            firewood_increment!(crate::registry::MERGE_COUNT, 1);
+            firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis(), "change" => "commit");
+            firewood_increment!(crate::registry::MERGE_COUNT, 1, "change" => "commit");
         };
 
         let result = proposal_handle.commit_proposal(metrics_cb);
-        let hash = result?.map(std::convert::Into::into);
+        let hash = result?.map(Into::into);
         Ok(hash)
     }
 }
@@ -493,8 +489,7 @@ pub extern "C" fn fwd_db_propose_change_proof<'db>(
 #[unsafe(no_mangle)]
 pub extern "C" fn fwd_db_commit_change_proof(args: CommittedChangeProofArgs<'_>) -> HashResult {
     crate::invoke_with_handle(args.proof, |ctx| {
-        ctx.commit()
-            .map(|hash_key| hash_key.map(std::convert::Into::into))
+        ctx.commit().map(|hash_key| hash_key.map(Into::into))
     })
 }
 


ffi/proofs.go
```diff
@@ -130,7 +130,7 @@ func (p *RangeProof) Verify(
 	return getErrorFromVoidResult(C.fwd_range_proof_verify(args))
 }
 
-// VerifyChangeProof verifies the provided change [proof] proves the changes
+// VerifyRangeProof verifies the provided change [proof] proves the changes
```
alarso16: Haha good catch, but will you fix the entire line?

ffi/src/proofs/change.rs
```diff
@@ -423,11 +496,33 @@ pub extern "C" fn fwd_db_verify_change_proof<'db>(
...
+        let end_key = end_key.into_option();
+        ctx.verify_and_commit(
+            db,
+            start_root,
+            end_root,
+            start_key.as_deref(),
+            end_key.as_deref(),
+            NonZeroUsize::new(max_length as usize),
+        )
+        .map(|hash_key| hash_key.map(std::convert::Into::into))
```
rkuris: nit: `use std::convert::Into` then just `Into::into`.

ffi/src/proofs/change.rs
```diff
@@ -197,6 +197,79 @@ impl<'db> ChangeProofContext<'db> {
...
+    fn verify_and_commit(
+        &mut self,
+        db: &'db crate::DatabaseHandle,
+        start_root: HashKey,
+        end_root: HashKey,
+        start_key: Option<&[u8]>,
+        end_key: Option<&[u8]>,
+        max_length: Option<NonZeroUsize>,
+    ) -> Result<Option<HashKey>, api::Error> {
+        self.verify(start_root, end_root, start_key, end_key, max_length)?;
```
rkuris: nit: naming: This assumes `verify` is idempotent. Perhaps it should be named `verify_if_needed` or `ensure_verified` or `verify_idempotent`.

ffi/src/proofs/change.rs
```diff
@@ -197,6 +197,79 @@ impl<'db> ChangeProofContext<'db> {
...
+            Some(ProposalState::Immutable(proposal)) => proposal,
+            None => {
+                // Create an immutable proposal
+                db.apply_change_proof_to_parent(start_root.into(), &self.proof)?
+                    .handle
+            }
+        };
+
+        let metrics_cb = |commit_time: coarsetime::Duration| {
+            firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis());
```
rkuris: Do we want to annotate these with a label that they are happening during the application of a change proof?

ffi/src/proofs/change.rs
```diff
@@ -197,6 +197,79 @@ impl<'db> ChangeProofContext<'db> {
...
+
+    fn find_next_key(&mut self) -> Result<Option<KeyRange>, api::Error> {
+        let verification = self
+            .verification
+            .as_ref()
+            .ok_or(api::Error::ProofError(ProofError::Unverified))?;
+
+        let Some(last_op) = self.proof.batch_ops().last() else {
+            // no BatchOps in the proof, so we are done
+            return Ok(None);
```
rkuris, line 226: Note that this is correct here, the previous PR would have failed the verification though.

ffi/src/proofs/change.rs
```diff
@@ -197,6 +197,79 @@ impl<'db> ChangeProofContext<'db> {
...
+        };
+
+        let root_hash = match self.proposal_state {
+            Some(ProposalState::Committed(ref hash)) => Ok(*hash),
+            Some(ProposalState::Immutable(ref proposal)) => proposal
+                .root_hash()
+                .map(|hash| hash.map(std::convert::Into::into)),
+            None => Err(api::Error::ProofError(ProofError::Unverified)),
+        }?;
+        if root_hash.as_ref() == Some(&verification.end_root) {
```
rkuris: I don't think this check is needed and isn't guaranteed to ever happen unless the CP covers the entire range or the remaining range.

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+	root2_updated, err := db2.VerifyAndCommitChangeProof(change_proof, root2, root1_updated, nothing(), nothing(), changeProofLenUnbounded)
+	r.NoError(err)
+	r.Equal(root1_updated, root2_updated)
+
+	// Verify all keys are now in db2
+	for i, key := range keys {
+		got, err := db2.Get(key)
+		r.NoError(err, "Get key %d", i)
+		r.Equal(vals[i], got, "Value mismatch for key %d", i)
+	}
```
rkuris, line 609: Better than this would be to check that the hashes match.

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+}
+
```
rkuris, line 611: We could use a test that verifies that applies two change proofs, CP1 and CP2 which represent different halves of a change proof, and then apply them in series.

R1 -> R2
CP1 consists of a change proof between R1 and R2 with half the keyspace
CP2 consists of a change proof between R1 and R2 with the other half of the keyspace

I think this will call out a few issues with the code in this PR and the prior one.

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+func TestVerifyAndCommitChangeProof(t *testing.T) {
+	r := require.New(t)
+	db1 := newTestDatabase(t)
+	db2 := newTestDatabase(t)
+
+	// Insert some data.
+	keys, vals, batch := kvForTest(100)
+	root1, err := db1.Update(batch[:50])
+	r.NoError(err)
+	root2, err := db2.Update(batch[:50])
```
alarso16:
```suggestion
	_, err := db2.Update(batch[:50])
```
Shouldn't these match? Just adds what seems like more mental overhead then necessary. Then you only need `root1, root2, gotRoot`

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+	r.NoError(err)
+
+	// Insert more data into db1 but not db2.
+	root1_updated, err := db1.Update(batch[50:])
+	r.NoError(err)
+
+	// Create a change proof from db1.
+	change_proof, err := db1.ChangeProof(root1, root1_updated, nothing(), nothing(), changeProofLenUnbounded)
```
alarso16:
```suggestion
	changeProof, err := db1.ChangeProof(root1, root1_updated, nothing(), nothing(), changeProofLenUnbounded)
```
camelCase is conventional in Go

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+	// Verify the proof
+	r.NoError(db2.VerifyChangeProof(proof, root2, root1_updated, nothing(), nothing(), changeProofLenTruncated))
+
+	// Now FindNextKey should work
+	nextRange, err := proof.FindNextKey()
+	r.NoError(err)
+	r.NotNil(nextRange)
+	startKey := nextRange.StartKey()
+	r.NotEmpty(startKey)
+	startKey = append([]byte{}, startKey...) // copy to new slice to avoid use-after-free
```
alarso16:
```suggestion
```
I believe `startKey` is copied when it's returned from the API, so this is unnecessary

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+
+		// Find the next start key
+		nextRange, err := proof.FindNextKey()
+		r.NoError(err)
+		if nextRange == nil {
+			break
+		}
+		startKey = maybe{
+			hasValue: true,
+			value:    append([]byte{}, nextRange.StartKey()...), // copy to new slice to avoid use-after-free
```
alarso16:
```suggestion
			value:    StartKey(),
```

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+	root2, err := db2.Update(batch[:50])
+	r.NoError(err)
+
+	// Insert the rest in the second batch
+	root1_updated, err := db1.Update(batch[50:])
+	r.NoError(err)
+
+	// Create and commit multiple change proofs to update db2 to match db1.
+	startKey := nothing()
+	for {
```
alarso16: This seems prone to busy loop forever if there's ever a bug. Do we know the exact amount of proofs this should take? or can we add some debugging with `t.Logf`? Just think this test will be very hard to debug

ffi/proofs_test.go
```diff
@@ -506,8 +507,216 @@ func TestVerifyChangeProof(t *testing.T) {
...
+
+	// Verify all keys are now in db2
+	for i, key := range keys {
+		got, err := db2.Get(key)
+		r.NoError(err, "Get key %d", i)
+		r.Equal(vals[i], got, "Value mismatch for %s", string(key))
+	}
+}
+
+func TestMultiRoundChangeProofWithDeletes(t *testing.T) {
```
alarso16: Can you combine the tests above? It's very conventional for functional in Go to have a test table like this:
```go
tests := []struct {
    name string,
    params any, // put everything necessary in other fields
}

for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
        r := require.New(t) // optional, but make sure you use the new `*testing.T`
        actualTest(tt.Params) // or write the test here
    })
}
```

ffi/proofs.go
bernard-avalabs: I'll rewrite the comment. This was something that I was planning to come back to after catching the error, but I must have forgotten.

ffi/proofs_test.go
bernard-avalabs: I will add a loop limit.

ffi/proofs_test.go
bernard-avalabs: I didn't know that this is Go convention. I'll make this change.

ffi/proofs.go
bernard-avalabs: Fixed

ffi/proofs_test.go
bernard-avalabs: Fixed

ffi/proofs_test.go
bernard-avalabs: Fixed

ffi/proofs_test.go
bernard-avalabs: Fixed

ffi/proofs_test.go
bernard-avalabs: Fixed

ffi/proofs_test.go
bernard-avalabs: Added loop limit

ffi/proofs_test.go
bernard-avalabs: Combined the two tests.

ffi/src/proofs/change.rs
bernard-avalabs: This has changed significantly since your review due to changes to part3 of the PR chain. `verify` is no longer idempotent. Given the change, I prefer to keep it as `verify`.

ffi/proofs_test.go
bernard-avalabs: There is a r.Equal on the root hashes above this `for` loop block.

ffi/proofs_test.go
bernard-avalabs: I believe the multi-round change proof tests in `TestMultiRoundChangeProof` already does unless you specifically mean two change proofs instead of more than one.

PR #1650, opened 2026-02-03
feat(db): add explicit close() method

RodrigoVillar:
## Why this should be merged

As part of deferred persistence, we need a way to signal to the background thread to stop and return. 

This PR is followed by #1656 and #1657.

## How this works

- Adds a `close` method to `Db` which closes the database and consumes the value.
- Updates testing code to use `close()` on reopens and replacements
- Implements the `Drop` trait on the `RevisionManager`
    - This allows us to gracefully terminate the background thread whenever the database is dropped, but only a warning is emitted if an error occurs (by virtue of `drop()` not returning anything)

For inspiration of the latter, I followed the `close()` / `drop()` pattern used by SQLite:
- https://github.com/rusqlite/rusqlite/blob/890be7bab5a8cef767b2c373ae14b19959c4f3de/src/lib.rs#L792-L808
- https://github.com/rusqlite/rusqlite/blob/890be7bab5a8cef767b2c373ae14b19959c4f3de/src/lib.rs#L357-L363

## How this was tested

CI

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -340,10 +340,7 @@ impl DatabaseHandle {
     }
 
     /// Closes the database gracefully.
-    ///
-    /// # Errors
-    ///
-    /// An error is returned if there was an error when closing the database.
+    #[expect(clippy::missing_errors_doc)]
     pub fn close(self) -> Result<(), api::Error> {
         self.db.close()
     }
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -447,7 +447,7 @@ impl RevisionManager {
     }
 
     /// Closes the revision manager gracefully.
-    #[allow(clippy::missing_const_for_fn)]
+    #[expect(clippy::missing_const_for_fn)]
     pub fn close(&self) -> Result<(), RevisionManagerError> {
         Ok(())
     }


firewood/src/manager.rs
```diff
@@ -445,6 +445,21 @@ impl RevisionManager {
...
+    pub fn close(&self) -> Result<(), RevisionManagerError> {
+        Ok(())
+    }
+}
+
+impl Drop for RevisionManager {
+    fn drop(&mut self) {
+        // Best-effort graceful shutdown - users are suggested to call `close()` instead.
+        if let Err(e) = self.close() {
+            warn!("Error during RevisionManager shutdown: {e}");
```
alarso16: Maybe an error log would be better

RodrigoVillar: Done: 1579b8f

ffi/src/handle.rs
```diff
@@ -338,6 +338,15 @@ impl DatabaseHandle {
     pub fn dump_to_string(&self) -> Result<String, api::Error> {
         self.db.dump_to_string().map_err(api::Error::from)
     }
+
+    /// Closes the database gracefully.
```
rkuris, line 342: nit: some extra documentation that states that this will persist any unpersisted versions might be helpful

ffi/src/handle.rs
```diff
@@ -338,6 +338,15 @@ impl DatabaseHandle {
...
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if there was an error when closing the database.
```
rkuris: Nit: This is a tautology and doesn't add any value. Either: (a) remove it, or (b) enumerate the expected errors. I think the expected errors here are just I/O errors.

firewood/src/db.rs
```diff
@@ -332,6 +332,11 @@ impl Db {
         let merge_ops = merkle.merge_key_value_range(first_key, last_key, key_values);
         self.propose_with_parent(merge_ops, merkle.nodestore())
     }
+
+    /// Closes the database gracefully.
+    pub fn close(self) -> Result<(), api::Error> {
```
rkuris, line 337: nit: I think at this point it's just a "flush unpersisted revisions".

RodrigoVillar: #1656 and #1657 add to the documentation of `close()` (i.e. `close()` in this PR is a no-op).

ffi/src/handle.rs
RodrigoVillar: https://github.com/ava-labs/firewood/pull/1650#discussion_r2765528505 applies here as well

ffi/src/handle.rs
RodrigoVillar: Went with removing the errors section for now: 792f3ee.

Will add this back in #1656 as `close()` will no longer be an no-op.

Issue #1651, opened 2026-02-03
`C.BatchOp` pointer misalignment


alarso16:
**Describe the bug**

Although the `C.BatchOp` struct appears to be properly aligned, the allocated struct by Go doesn't appear to be aligned properly.

Relevant code:
```go
func newCBatchOp(op BatchOp, pinner Pinner) C.BatchOp {
	var cOp C.BatchOp
	cOp.tag = op.tag
	switch op.tag {
	case C.BatchOp_Put:
		ptr := unsafe.Pointer(&cOp.anon0)
		putBody := C.BatchOp_Put_Body{
			key:   newBorrowedBytes(op.key, pinner),
			value: newBorrowedBytes(op.value, pinner),
		}
		
        // normal fmt.Println rearranges memory...
		println("addr:", uintptr(ptr))
		println("mod8:", uintptr(ptr)%8) // this prints 4
		println("align:", unsafe.Alignof(putBody)) // this prints 8

		*(*C.BatchOp_Put_Body)(ptr) = putBody
	case C.BatchOp_Delete:
		*(*C.BatchOp_Delete_Body)(unsafe.Pointer(&cOp.anon0)) = C.BatchOp_Delete_Body{
			key: newBorrowedBytes(op.key, pinner),
		}
	case C.BatchOp_DeleteRange:
		*(*C.BatchOp_DeleteRange_Body)(unsafe.Pointer(&cOp.anon0)) = C.BatchOp_DeleteRange_Body{
			prefix: newBorrowedBytes(op.key, pinner),
		}
	}
	return cOp
}
```

Output:
```
addr: 824634019484
mod8: 4
align: 8
fatal error: checkptr: misaligned pointer conversion

goroutine 24 gp=0xc000003c00 m=0 mp=0x1050a8120 [running]:
runtime.throw({0x10418fa0c?, 0x3800000000000000?})
	/usr/local/go/src/runtime/panic.go:1101 +0x38 fp=0xc000048e40 sp=0xc000048e10 pc=0x1027474e8
runtime.checkptrAlignment(0x1041719bc?, 0x46?, 0x103837a08?)
	/usr/local/go/src/runtime/checkptr.go:20 +0x80 fp=0xc000048e60 sp=0xc000048e40 pc=0x1026de030
[...]

PR #1652, opened 2026-02-03
fix(ffi): force correct alignment on certain c structs

demosdemon:
## Why this should be merged

When enabling race detection in Go, it alerted us that Go was producing structs that were not properly aligned. This was due to the union part being reduced down to a byte array which always has an alignment of 1. The enum part had an automatic alignment of 4 bytes, resulting in a 4-byte padding between the tag and the union as well as a 4-byte alignment when the struct was created in Go. The 4-byte alignment is incorrect on 64-bit platforms and should be pointer-width alignment for the tagged enums that contain pointers.

## How this works

This resolves the issue by explicitly setting the alignment of the tag part of the tagged union structs to be pointer-width aligned. This is flexible enough to handle both 32-bit and 64-bit architectures as well as any future 128-bit architectures. I included all enums with pointers in their bodies and not just the ones created in Go for consistency and to reduce ambiguity.

## How this was tested

Race detection was enabled in the tests, which no longer fails.

Closes: #1651

### DIFF
--- a/.github/workflows/track-performance.yml
+++ b/.github/workflows/track-performance.yml
@@ -68,6 +68,7 @@ jobs:
             # Cron strings must match exactly from schedule definitions above
             case "${{ github.event.schedule }}" in
               "0 5 * * *")  # Daily at 05:00 UTC: 1M blocks (40M → 41M)
+                echo "::notice::Schedule: daily-40m-41m (blocks 40000001-41000000)"
                 echo "name=daily-40m-41m" >> "$GITHUB_OUTPUT"
                 echo "start-block=40000001" >> "$GITHUB_OUTPUT"
                 echo "end-block=41000000" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -286,8 +286,8 @@ jobs:
           sudo prlimit --pid $$ --memlock=-1:-1 || echo "prlimit not available, locking kernel memory may fail when setting up io-uring"
           ulimit -Sa
           ulimit -Ha
-          go test -fuzz=FuzzFirewoodTree -fuzztime=1m
-          go test -fuzz=FuzzRangeProofCreation -fuzztime=1m
+          go test -race -fuzz=FuzzFirewoodTree -fuzztime=1m
+          go test -race -fuzz=FuzzRangeProofCreation -fuzztime=1m
```
alarso16, line 290: Can you add it for the merkle fuzzes too?

PR #1653, opened 2026-02-03
docs: Add docs for performance numbers

rkuris:
## Why this should be merged

Improves discoverability of metrics data.

## How this was tested

Untested, but we can fix it if the links do not work later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 [![Ecosystem license](https://img.shields.io/badge/License-Ecosystem-blue.svg)](./LICENSE.md)
 
 > :warning: Firewood is beta-level software.
-> The Firewood API may change with little to no warning.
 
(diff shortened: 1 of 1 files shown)


Elvis339:
It seems like duplicate of #1642 - [METRICS.md#viewing-results](https://github.com/ava-labs/firewood/blob/95783ec1f3e026dce1c68a78f81dcf9a5f62a03d/METRICS.md#viewing-results)

rkuris:
> It seems like duplicate of #1642 - [METRICS.md#viewing-results](https://github.com/ava-labs/firewood/blob/95783ec1f3e026dce1c68a78f81dcf9a5f62a03d/METRICS.md#viewing-results)

Ah yes it totally is. I was trying to capture the data you sent me in chat, but it's already there, just way at the bottom.

PR #1656, opened 2026-02-03
feat: defer persist every commit

RodrigoVillar:
## Why this should be merged

Implements deferred persistence on every commit.

This PR follows #1650 and precedes #1657.

## How this works
 
A new `PersistWorker` type coordinates with the `PersistLoop` running in a background thread to handle all disk I/O:                                                                                                                                       
  - Persistence: Writing committed revisions to disk and updating `RootStore`                                                                                                                                                                      
  - Reaping: Adding deleted nodes from old revisions to the free lists (avoids `Mutex` contention as `PersistLoop` is the sole user of the `NodeStoreHeader`).
                                                                                                                                                                    
```mermaid
sequenceDiagram
    participant Caller
    participant Main as Main Thread
    participant BG as Background Thread
    participant Disk
                                                                                                                                                                                                                                                
    Caller->>Main: commit(proposal)
    Main->>Main: Validate proposal
    Main->>Main: Update in-memory state
    Main->>Main: Acquire semaphore permit                                                                                                                                                                                                               

    Main->>BG: Send Persist message
    Main-->>Caller: Return
    BG->>Disk: Write revision
    BG->>Disk: Update RootStore                                                                                                                                                                                                    
    BG->>BG: Release semaphore permit
```

**Backpressure Mechanism**: The `PersistSemaphore` limits how far commits can get ahead of persistence. When the limit is reached (currently 1 unpersisted commit), subsequent commits block until persistence catches up. This ensures the deferred persistence invariant is maintained. If the background thread fails, the error is propagated to the next `commit()` or `propose()` call.

**Graceful Shutdown**: Extends the `close()` method to close the background thread.            

## How this was tested

Added UTs + CI

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -340,7 +340,7 @@ impl Db {
     /// This method shuts down the background persistence worker. If not called
     /// explicitly, `Drop` will attempt a best-effort shutdown but cannot report
     /// errors.
-    pub fn close(self) -> Result<(), api::Error> {
+    pub fn close(mut self) -> Result<(), api::Error> {
         self.manager.close().map_err(Into::into)
     }
 }
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -465,7 +465,7 @@ impl RevisionManager {
     /// This method shuts down the background persistence worker. If not called
     /// explicitly, `Drop` will attempt a best-effort shutdown but cannot report
     /// errors.
-    pub fn close(&self) -> Result<(), RevisionManagerError> {
+    pub fn close(&mut self) -> Result<(), RevisionManagerError> {
         self.persist_worker
             .close()
             .map_err(RevisionManagerError::PersistError)
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -71,8 +71,6 @@ enum PersistMessage {
     Persist(CommittedRevision),
     /// A persisted revision to be reaped.
     Reap(NodeStore<Committed, FileBacked>),
-    /// The background thread should shutdown.
-    Shutdown,
 }
 
 /// Handle for managing the background persistence thread.
@@ -82,7 +80,10 @@ pub(crate) struct PersistWorker {
     handle: Mutex<Option<JoinHandle<Result<(), PersistError>>>>,
 
     /// Channel for sending messages to the background thread.
-    sender: Sender<PersistMessage>,
+    ///
+    /// Wrapped in `Option` so that `close()` can drop the sender to signal
+    /// shutdown to the background thread via channel close.
+    sender: Option<Sender<PersistMessage>>,
 
     /// Shared state with background thread.
     shared: Arc<SharedState>,
@@ -115,7 +116,7 @@ impl PersistWorker {
 
         Self {
             handle: Mutex::new(Some(handle)),
-            sender,
+            sender: Some(sender),
             shared,
         }
     }
@@ -126,6 +127,8 @@ impl PersistWorker {
         self.shared.commit_throttle.acquire();
 
         self.sender
+            .as_ref()
+            .ok_or(PersistError::ChannelDisconnected)?
             .send(PersistMessage::Persist(committed))
             .map_err(|_| self.resolve_worker_error())
     }
@@ -141,6 +144,8 @@ impl PersistWorker {
             // revision with no root can still carry deleted nodes from the
             // previous revision that need their disk space freed.
             self.sender
+                .as_ref()
+                .ok_or(PersistError::ChannelDisconnected)?
                 .send(PersistMessage::Reap(nodestore))
                 .map_err(|_| self.resolve_worker_error())?;
         }
@@ -165,10 +170,10 @@ impl PersistWorker {
     ///
     /// This method is idempotent for compatibility with `drop` (as `drop` is
     /// called after this function).
-    pub(crate) fn close(&self) -> Result<(), PersistError> {
-        // Signal to the background thread to exit.
-        // We ignore any errors here as the background thread may already have exited.
-        let _ = self.sender.send(PersistMessage::Shutdown);
+    pub(crate) fn close(&mut self) -> Result<(), PersistError> {
+        // Drop the sender to close the channel, signaling the background
+        // thread to exit.
+        drop(self.sender.take());
 
         self.join_handle();
         self.check_error()
@@ -322,10 +327,9 @@ impl PersistLoop {
         result
     }
 
-    /// Processes messages until shutdown or error.
+    /// Processes messages until the channel is closed or an error occurs.
     ///
-    /// Upon receiving a message, this can do one of three things:
-    /// - On `Shutdown` or channel close: exits gracefully.
+    /// Upon receiving a message, this can do one of two things:
     /// - On `Reap`: drops the revision.
     ///   If persisted, the revision's nodes are added to the free lists only if
     ///   not running in archival mode.
@@ -334,12 +338,8 @@ impl PersistLoop {
     fn event_loop(&mut self) -> Result<(), PersistError> {
         let mut num_commits = nonzero!(1u64);
 
-        loop {
-            // An error indicates that the channel is closed.
-            let message = self.receiver.recv().unwrap_or(PersistMessage::Shutdown);
-
+        while let Ok(message) = self.receiver.recv() {
             match message {
-                PersistMessage::Shutdown => return Ok(()),
                 PersistMessage::Reap(nodestore) => self.reap(nodestore)?,
                 PersistMessage::Persist(revision) => {
                     if num_commits
@@ -353,6 +353,8 @@ impl PersistLoop {
                 }
             }
         }
+
+        Ok(())
     }
 
     /// Persists the given revision and releases semaphore permits.


ffi/src/handle.rs
```diff
@@ -340,7 +340,11 @@ impl DatabaseHandle {
     }
 
     /// Closes the database gracefully.
-    #[expect(clippy::missing_errors_doc)]
+    ///
+    /// # Errors
+    ///
+    /// An error is returned if the background thread failed to return or if
```
rkuris: nit: probably should call this "the persistence background thread" since from the caller's perspective they didn't do anything to create this background thread. It's also not clear what could cause "failed to return" and whether they could do anything about it.

ffi/src/lib.rs
```diff
@@ -687,7 +689,8 @@ pub extern "C" fn fwd_start_logs(args: LogArgs) -> VoidResult {
 ///
 /// - [`VoidResult::NullHandlePointer`] if the provided database handle is null.
 /// - [`VoidResult::Ok`] if the database handle was successfully closed and freed.
-/// - [`VoidResult::Err`] if the process panics while closing the database handle.
+/// - [`VoidResult::Err`] if the process panics while closing the database
```
rkuris:
```suggestion
/// - [`VoidResult::Err`] if the background persistence worker thread panics while closing the database
```

firewood/src/db.rs
```diff
@@ -468,6 +474,13 @@ mod test {
 
     impl<T: Iterator> IterExt for T {}
 
+    impl Db {
+        /// Wait until all pending commits have been persisted.
+        fn wait_persisted(&self) {
+            self.manager.wait_persisted();
+        }
```
rkuris, line 481: This almost makes me want to implement `Deref` for Db.

I'm worried about how long this could wait for (forever?) Is there a timeout?

Debugging a hang here will be nearly impossible, at an absolute minimum after a few minutes it should generate a warning. Another possible option here would be to add a heartbeat, which is just an atomic integer that increases whenever an IO operation completes, and we check this periodically.

Threads are difficult. Debugging them is even harder.,

firewood/src/manager.rs
```diff
@@ -117,7 +112,14 @@ pub(crate) struct RevisionManager {
...
     /// queries of arbitrarily old revisions without keeping them in memory.
-    root_store: Option<RootStore>,
+    root_store: Option<Arc<RootStore>>,
+
+    /// Worker responsible for persisting revisions to disk.
+    persist_worker: PersistWorker,
+
+    /// Lock for upholding the invariant that only one revision may be comitting
+    /// at any given time.
+    commit_lock: Mutex<()>,
```
rkuris: This is scary, since nothing enforces that the commit lock is held when someone is committing. This is how you do it in golang. I'll keep reading on to figure out how to rustify this.

firewood/src/manager.rs
```diff
@@ -221,25 +229,35 @@ impl RevisionManager {
...
     #[fastrace::trace(short_name = true)]
     #[crate::metrics("proposal.commit", "proposal commit to storage")]
     pub fn commit(&self, proposal: ProposedRevision) -> Result<(), RevisionManagerError> {
-        // 1. Commit check
+        // 1. Check if the persist worker has failed.
+        self.persist_worker
+            .check_error()
+            .map_err(RevisionManagerError::PersistError)?;
+
+        let commit_lock = self.commit_lock.lock();
```
rkuris: This should be a lock on something, here are some candidates:
`persist_worker` `current_revision`, `in_memory_revisions`

IMO the reason we need this check is because we need to be sure that the current revision is it's parent, and we can't allow someone else to start a commit unless that's correct, so that seems like the correct thing to lock.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+
+/// Error type for persistence operations.
+#[derive(Clone, Debug, thiserror::Error)]
+pub enum PersistError {
+    #[error("IO error during persistence: {0}")]
+    FileIo(#[from] Arc<FileIoError>),
+    #[error("RootStore error during persistence: {0}")]
+    RootStore(#[source] Arc<dyn std::error::Error + Send + Sync>),
+    #[error("Background thread panicked")]
+    ThreadPanic,
```
rkuris: Can we include the panic message here?

See https://medium.com/@bugsybits/what-happens-when-a-rust-thread-crashes-d82cb21ff691

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+    #[error("Failed to send message to background thread: channel disconnected")]
+    ChannelDisconnected,
```
rkuris, line 65: How does this happen?

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+    Persist(CommittedRevision),
+    /// A persisted revision to be reaped.
+    Reap(NodeStore<Committed, FileBacked>),
+    /// The background thread should shutdown.
+    Shutdown,
+}
+
+/// Handle for managing the background persistence thread.
+#[derive_where(Debug)]
+#[derive_where(skip_inner)]
```
rkuris: Feels like I might want `SharedState` at a minimum when rendering this in the debugger.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+/// The background persistence loop that runs in a separate thread.
+struct PersistLoop {
+    /// Channel for receiving messages from `PersistWorker`.
+    receiver: Receiver<PersistMessage>,
+    /// Persist every `sub_interval` commits.
+    sub_interval: usize,
+    /// Shared state for coordination with `PersistWorker`.
+    shared: Arc<SharedState>,
+    /// The commit number of the last successful persist (for calculating permits to release).
+    last_persisted_commit: usize,
```
rkuris:
```suggestion
    last_persisted_commit: Option<NonZeroUsize>,
```

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+            semaphore: PersistSemaphore::new(sub_interval),
+            root_store,
+            header: Mutex::new(header),
+        });
+
+        let persist_loop = PersistLoop {
+            receiver,
+            sub_interval,
+            shared: shared.clone(),
+            last_persisted_commit: 0,
```
rkuris: We didn't persist commit 0, so this should be `None` not 0. The compiler will do this work for you and ensure that you correctly handle "0" (None) correctly in all cases.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+            return Err(e);
+        }
+
+        // Release permits for all commits that were persisted
+        self.release_permits(num_commits);
+        Ok(())
+    }
+
+    /// Releases semaphore permits for commits up to `num_commits`.
+    fn release_permits(&mut self, num_commits: usize) {
```
rkuris: `num_commits` is also a `NonZeroUsize` I think.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+        // Persist the revision
+        let mut header = self.shared.header.lock();
+        let result = revision.persist(&mut header);
+        drop(header);
+
+        if let Err(e) = result {
+            error!("Failed to persist revision: {e}");
+
+            let err = PersistError::FileIo(Arc::new(e));
+            self.shared.error.set(err.clone()).expect("should be empty");
```
rkuris: Probably so. I'd need to think harder if we can eliminate this `expect` call somehow.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// Persists the given revision and releases semaphore permits.
+    fn persist(
+        &mut self,
+        revision: &CommittedRevision,
+        num_commits: usize,
+    ) -> Result<(), PersistError> {
+        // Persist the revision
+        let mut header = self.shared.header.lock();
+        let result = revision.persist(&mut header);
+        drop(header);
```
rkuris, line 369: It's unfortunate that we have to pass persist a guard instead of the header itself.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+        }
+
+        // Release permits for all commits that were persisted
+        self.release_permits(num_commits);
+        Ok(())
+    }
+
+    /// Releases semaphore permits for commits up to `num_commits`.
+    fn release_permits(&mut self, num_commits: usize) {
+        let permits_to_release = num_commits.saturating_sub(self.last_persisted_commit);
```
rkuris: why is this a saturating_sub?

rkuris (changes requested):
Needs another pass

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+    shared: Arc<SharedState>,
+}
+
+impl PersistWorker {
+    /// Creates a new `PersistWorker` and starts the background thread.
+    ///
+    /// Returns the worker for sending messages to the background thread.
+    #[allow(clippy::large_types_passed_by_value)]
+    pub(crate) fn new(header: NodeStoreHeader, root_store: Option<Arc<RootStore>>) -> Self {
+        let (sender, receiver) = channel::unbounded();
```
rkuris, line 98: why is this unbounded? Seems like we should only keep maybe 2 messages max and block otherwise? Worth thinking about in any case.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+        let sub_interval = 1;
```
rkuris: nit: naming: `sub_interval` -> `persist_commit_ratio` or `persist_rate` or `persist_interval`

firewood/src/persist_worker.rs
rkuris: This should also be a uint64 not a usize...

The size of usize varies on machines with different maximum addressable memory, which we don't want here. uint32 might be sufficient but uint64 is safer.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+
+        let shared = Arc::new(SharedState {
+            error: OnceLock::new(),
+            semaphore: PersistSemaphore::new(sub_interval),
```
rkuris: nit: naming: `commits_before_blocking` or something more creative. `semaphore` tells me nothing about what this is used for.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+        let handle = thread::spawn(move || persist_loop.run());
+
+        Self {
+            handle: Mutex::new(Some(handle)),
+            sender,
+            shared,
+        }
+    }
+
+    /// Sends `committed` to the background thread for persistence. This calls
```
rkuris:
```suggestion
    /// Sends `committed` to the background thread for persistence. This call
```

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,338 @@
...
+    ///   modulo `sub_interval` is zero.
+    fn run(mut self) -> Result<(), PersistError> {
+        let mut num_commits = 0usize;
+
+        loop {
+            // An error indicates that the channel is closed.
+            let message = self.receiver.recv().unwrap_or(PersistMessage::Shutdown);
+
+            match message {
+                PersistMessage::Shutdown => return Ok(()),
```
rkuris: nit: break might be more readable here, with `Ok(())` at the end of the function. Your choice though.

ffi/src/lib.rs
RodrigoVillar: Done: 4383901

firewood/src/persist_worker.rs
RodrigoVillar: Went with `persist_interval`: d50a03c

firewood/src/persist_worker.rs
RodrigoVillar: Good catch: 0e1d1d6

firewood/src/persist_worker.rs
RodrigoVillar: Ah this is a remnant from a prior version of this PR - applied `#[derive(Debug)]` to persistence types where needed: fb0576a

firewood/src/persist_worker.rs
RodrigoVillar: Going to stick with the return since it makes matching the documentation with the source code itself easier to match + read.

firewood/src/persist_worker.rs
RodrigoVillar: Changed to `NonZeroU64` here: 88760d6

firewood/src/persist_worker.rs
RodrigoVillar: Done: 88760d6

firewood/src/persist_worker.rs
RodrigoVillar: Done: 88760d6

firewood/src/persist_worker.rs
RodrigoVillar: This was to satisfy `clippy` - should I use `checked_sub()` instead?

RodrigoVillar:
@rkuris I moved the PR diagram to the top of `persist_worker.rs` as requested: b14a97c

firewood/src/db.rs
RodrigoVillar: Hmm could you elaborate on why we would want to implement `Deref` on `Db`?

rkuris:
You can use mermaid diagrams directly!

ffi/src/handle.rs
RodrigoVillar: Updated here: ea953d4. If `close()` returns an error, I'm not sure what else the user can do except reopen the database (which should still be in a consistent state). Would it perhaps to make sense to detail the expected recovery procedure here as well?

firewood/src/persist_worker.rs
RodrigoVillar: `PersistLoop` would be the sole owner of the `NodeStoreHeader` if the database didn't have the `check()` method on it... it might make sense to tag this in an issue of it's own as this would improve the performance of deferred persistence (no `Mutex` over the `NodeStoreHeader`)

firewood/src/persist_worker.rs
RodrigoVillar: Renamed to `commit_throttle` as this throttles the number of unpersisted commits: 028de38

firewood/src/persist_worker.rs
RodrigoVillar: This would happen if we persist/reap after closing the `RevisionManager`; I modified the error handling code to make this more clear here: 90e827f

firewood/src/persist_worker.rs
RodrigoVillar: Rather than encapsulating the panic message in an error, I went with throwing the panic back up the stack via `resume_rewind()`: 90e827f

firewood/src/manager.rs
RodrigoVillar: Created a PR for this here: https://github.com/ava-labs/firewood/pull/1661

firewood/src/manager.rs
RodrigoVillar: Created a PR for this here: https://github.com/ava-labs/firewood/pull/1661

firewood/src/persist_worker.rs
RodrigoVillar: The limit would be `2N+1` (`ReapMessage` => `CommitMessage` => `ReapMessage` => `BLOCK`) but regardless, I think this would make the code harder to read as the `PersistSemamphore` is the only source of backpressure that we care about. Open to setting the limit here if you feel strongly about it.

ffi/src/handle.rs
rkuris: Self is consumed so it is clear they can't do anything 🙂

rkuris: Self is consumed so it is clear they can't do anything 🙂

rkuris: Self is consumed so it is clear they can't do anything 🙂

firewood/src/db.rs
rkuris: Deref coersion is why

https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion

We wouldn't need this wrapper as we can call it directly on db.

Not a perfect fit here but worth thinking about.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// if archival mode is disabled. Otherwise, the `nodestore` is dropped.
+    pub(crate) fn reap(
+        &self,
+        nodestore: NodeStore<Committed, FileBacked>,
+    ) -> Result<(), PersistError> {
+        let is_persisted = nodestore
+            .root_as_maybe_persisted_node()
+            .is_some_and(|node| node.unpersisted().is_none());
+
+        if is_persisted && self.shared.root_store.is_none() {
```
rkuris: Let's reverse these two tests. First check that there isn't a root store, then convert the root to a MaybePersisted. This could involve a memory allocation which we can avoid if you're an archive node.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,408 @@
...
+            shared,
+        }
+    }
+
+    /// Sends `committed` to the background thread for persistence. This call
+    /// blocks if the limit of unpersisted commits has been reached.
+    pub(crate) fn persist(&self, committed: CommittedRevision) -> Result<(), PersistError> {
+        self.shared.commit_throttle.acquire();
+
+        // Check for errors after potentially blocking
```
rkuris: Is this test needed? Seems like if an error has occurred, the thread will have closed the channel, so this check probably isn't necessary. I don't think it hurts to keep it, but it could use a better comment.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,408 @@
...
+        self.check_error()?;
+
+        self.sender
+            .send(PersistMessage::Persist(committed))
+            .map_err(|_| self.resolve_send_error())
```
rkuris: I think we can do a little better on naming here, took me several reads to understand what this actually does.

Some ideas: `worker_error` or `stored_error` or `persist_thread_error_or_disconnected`.

Could also be that we have a function with undocumented side effects here. `resolve_send_error` waits for the worker thread (which is presumed dead), which isn't obvious from any of these function names.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Sends `nodestore` to the background thread for reaping if persisted and
+    /// if archival mode is disabled. Otherwise, the `nodestore` is dropped.
+    pub(crate) fn reap(
+        &self,
+        nodestore: NodeStore<Committed, FileBacked>,
+    ) -> Result<(), PersistError> {
+        let is_persisted = nodestore
+            .root_as_maybe_persisted_node()
+            .is_some_and(|node| node.unpersisted().is_none());
```
rkuris: Is an empty trie considered unpersisted? Should it be? Let's doc that if so.

firewood/src/persist_worker.rs
rkuris: Clippy is forcing you to think about how to handle the case where the value becomes negative. The question to answer here is, is it possible for this to be negative, and if so, what do you want it to do?

firewood/src/persist_worker.rs
RodrigoVillar: Done: e3dc113

firewood/src/persist_worker.rs
RodrigoVillar: Renamed to `resolve_worker_error` here: c2c0451. Regarding the undocumented side effect, IMO the documentation for `resolve_worker_error` makes it clear what the function does:

```
Joins the background thread and returns the error that caused it to exit.
```

firewood/src/persist_worker.rs
RodrigoVillar: Good point - removed unnecessary error check here: 75af470.

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+impl PersistWorker {
+    /// Creates a new `PersistWorker` and starts the background thread.
+    ///
+    /// Returns the worker for sending messages to the background thread.
+    #[allow(clippy::large_types_passed_by_value)]
+    pub(crate) fn new(header: NodeStoreHeader, root_store: Option<Arc<RootStore>>) -> Self {
+        let (sender, receiver) = channel::unbounded();
+        let persist_interval = NonZeroU64::new(1).expect("should be nonzero");
```
rkuris:
```suggestion
        let persist_interval = nonzero!(1u64);
```
See https://docs.rs/nonzero_ext/0.3.0/nonzero_ext/macro.nonzero.html for docs

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,416 @@
...
+    ///
+    /// Upon receiving a message, this can do one of three things:
+    /// - On `Shutdown` or channel close: exits gracefully.
+    /// - On `Reap`: drops the revision.
+    ///   If persisted, the revision's nodes are added to the free lists only if
+    ///   not running in archival mode.
+    /// - On `Persist`: persists the revision if the number of revisions received
+    ///   modulo `persist_interval` is zero.
+    fn event_loop(&mut self) -> Result<(), PersistError> {
+        let mut num_commits = NonZeroU64::new(1).expect("should be nonzero");
```
rkuris: same macro comment as earlier, use `nonzero!(1u64)`

firewood/src/persist_worker.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        loop {
+            // An error indicates that the channel is closed.
+            let message = self.receiver.recv().unwrap_or(PersistMessage::Shutdown);
```
rkuris: There are two ways to shut down the thread, one is to close the sender and the other is to send a Shutdown message. Are these 100% equivalent? I think so, especially since you're mapping the closed sender to Shutdown, and Shutdown can never fail.

Consider removing PersistMessage::Shutdown completely and just closing the send channel to shut down the receiver. In fact, it might automatically just close when it goes out of scope in the caller, which perhaps is even easier...

firewood/src/persist_worker.rs
RodrigoVillar: Done: c1630c0

firewood/src/persist_worker.rs
RodrigoVillar: Done: c1630c0.

firewood/src/persist_worker.rs
RodrigoVillar: The empty trie is considered unpersisted which is problematic in the following case:

1. We open a fresh instance of Firewood (trie is empty) 
2. We propose and commit a revision `R1` - the trie contains root node `A`
3. We then propose and commit another revision `R2` which reverts `R1` - the trie is now empty and node `A` is in the deleted list

With our previous code, if we called `reap(R2)`, then we would just drop the proposal as the `root_node` is `None`, even though we had a node to add to the free lists. I've updated the code so that we always reap here: aa5da68.

firewood/src/persist_worker.rs
RodrigoVillar: It's not possible for the value to be negative (or to underflow). I went ahead and used `check_sub()`/`expect()` here: 8a5a6bf.

firewood/src/db.rs
RodrigoVillar: I went ahead and added a warning message if we hang in `wait_persisted()`: 710248e.

firewood/src/persist_worker.rs
RodrigoVillar: Went ahead and chose to drop the channel when closing: b62295a (removes the `Shutdown` variant altogether).

RodrigoVillar:
Ran the reexecution test against this PR and `master` of AvalancheGo here: 

https://github.com/ava-labs/firewood/actions/runs/21884348961/job/63175273483

https://github.com/ava-labs/avalanchego/actions/runs/21884370329/job/63175334353

PR #1657, opened 2026-02-03
feat: defer persist every `N` commits

RodrigoVillar:
## Why this should be merged

Extends deferred persistence to allow for persists on every `N`th commit.

This PR precedes #1650 and #1656 (in order).

## How this works

Adds a new configuration option `deferred_persistence_commit_count` which controls the maximum number of unpersisted revisions.

  - `N = 1`: preserves current behavior (persist every commit)
  - `N > 1`: defers persistence until `N/2` commits accumulate (the "sub-interval")

Below is an example of when `deferred_persistence_commit_count = 10`:

```mermaid
sequenceDiagram                                                                                                                                                                  
      participant Caller                                                                                                                                                           
      participant Main as Main Thread                                                                                                                                                          
      participant BG as Background Thread                                                                                                                                                    
      participant Disk                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                      
      loop Commits 1-4                                                                                                                                                             
          Caller->>Main: commit()                                                                                                                                                    
          Main->>BG: queue revision                                                                                                                                       
          Note right of BG: Waiting...                                                                                                                             
      end                                                                                                                                                                          
                                                                                                                                                                                   
      Caller->>Main: commit() (5th)                                                                                                                                                  
      Main->>BG: queue revision                                                                                                                                           
      BG->>Disk: persist revision 5                                                                                                                                 
      Note right of Disk: Sub-interval (10/2) reached                                                                                                                              
                                                                                                                                                                                   
      loop Commits 6-8                                                                                                                                                             
          Caller->>Main: commit()                                                                                                                                                    
          Main->>BG: queue revision                                                                                                                                       
          Note right of BG: Waiting...                                                                                                                             
      end                                                                                                                                                                          
                                                                                                                                                                                   
      Caller->>Main: close()                                                                                                                                                         
      Main->>BG: shutdown + persist last committed revision                                                                                                                             
      BG->>Disk: persist revision 8                                                                                                                                 
      Note right of Disk: Latest committed revision is persisted        
```

On `close()`, the last committed revision is persisted to disk.

## How this was tested

Added UTs + CI.

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -113,14 +113,18 @@ type config struct {
 	rootStore bool
 	// expensiveMetricsEnabled controls whether expensive metrics recording is enabled.
 	expensiveMetricsEnabled bool
+	// deferredPersistenceCommitCount determines the maximum number of unpersisted
+	// revisions that can exist at a given time.
+	deferredPersistenceCommitCount uint64
 }
 
 func defaultConfig() *config {
 	return &config{
-		nodeCacheEntries:     1_000_000,
-		freeListCacheEntries: 1_000_000,
-		revisions:            100,
-		readCacheStrategy:    OnlyCacheWrites,
+		nodeCacheEntries:               1_000_000,
+		freeListCacheEntries:           1_000_000,
+		revisions:                      100,
+		readCacheStrategy:              OnlyCacheWrites,
+		deferredPersistenceCommitCount: 1,
 	}
 }
 
@@ -193,6 +197,15 @@ func WithExpensiveMetrics() Option {
 	}
 }
 
+// WithDeferredPersistenceCommitCount sets the maximum number of unpersisted revisions
+// that can exist at a time. Note: `commitCount` must be greater than 0.
+// Default: 1
+func WithDeferredPersistenceCommitCount(commitCount uint64) Option {
+	return func(c *config) {
+		c.deferredPersistenceCommitCount = commitCount
+	}
+}
+
 // A CacheStrategy represents the caching strategy used by a [Database].
 type CacheStrategy uint8
 
@@ -250,15 +263,16 @@ func New(dbDir string, nodeHashAlgorithm NodeHashAlgorithm, opts ...Option) (*Da
 	defer pinner.Unpin()
 
 	args := C.struct_DatabaseHandleArgs{
-		dir:                  newBorrowedBytes([]byte(dbDir), &pinner),
-		cache_size:           C.size_t(conf.nodeCacheEntries),
-		free_list_cache_size: C.size_t(conf.freeListCacheEntries),
-		revisions:            C.size_t(conf.revisions),
-		strategy:             C.uint8_t(conf.readCacheStrategy),
-		truncate:             C.bool(conf.truncate),
-		root_store:           C.bool(conf.rootStore),
-		expensive_metrics:    C.bool(conf.expensiveMetricsEnabled),
-		node_hash_algorithm:  C.enum_NodeHashAlgorithm(nodeHashAlgorithm),
+		dir:                               newBorrowedBytes([]byte(dbDir), &pinner),
+		cache_size:                        C.size_t(conf.nodeCacheEntries),
+		free_list_cache_size:              C.size_t(conf.freeListCacheEntries),
+		revisions:                         C.size_t(conf.revisions),
+		strategy:                          C.uint8_t(conf.readCacheStrategy),
+		truncate:                          C.bool(conf.truncate),
+		root_store:                        C.bool(conf.rootStore),
+		expensive_metrics:                 C.bool(conf.expensiveMetricsEnabled),
+		node_hash_algorithm:               C.enum_NodeHashAlgorithm(nodeHashAlgorithm),
+		deferred_persistence_commit_count: C.uint64_t(conf.deferredPersistenceCommitCount),
 	}
 
 	return getDatabaseFromHandleResult(C.fwd_open_db(args))
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -23,8 +23,9 @@ use firewood_storage::{
     CheckOpt, CheckerReport, Committed, FileBacked, FileIoError, HashedNodeReader,
     ImmutableProposal, NodeHashAlgorithm, NodeStore, Parentable, ReadableStorage, TrieReader,
 };
+use nonzero_ext::nonzero;
 use std::io::Write;
-use std::num::NonZeroUsize;
+use std::num::{NonZeroU64, NonZeroUsize};
 use std::path::Path;
 use std::sync::Arc;
 use thiserror::Error;
@@ -135,6 +136,9 @@ pub struct DbConfig {
     /// Whether to enable `RootStore`.
     #[builder(default = false)]
     pub root_store: bool,
+    /// The maximum number of unpersisted revisions that can exist at a given time.
+    #[builder(default = nonzero!(1u64))]
+    pub deferred_persistence_commit_count: NonZeroU64,
 }
 
 #[derive(Debug)]
@@ -182,6 +186,7 @@ impl Db {
             .create(cfg.create_if_missing)
             .truncate(cfg.truncate)
             .root_store(cfg.root_store)
+            .deferred_persistence_commit_count(cfg.deferred_persistence_commit_count)
             .manager(cfg.manager)
             .build();
         let manager = RevisionManager::new(config_manager)?;
@@ -337,9 +342,9 @@ impl Db {
 
     /// Closes the database gracefully.
     ///
-    /// This method shuts down the background persistence worker. If not called
-    /// explicitly, `Drop` will attempt a best-effort shutdown but cannot report
-    /// errors.
+    /// This method shuts down the background persistence worker and persists
+    /// the last committed revision. If not called explicitly, `Drop` will
+    /// attempt a best-effort shutdown but cannot report errors.
     pub fn close(mut self) -> Result<(), api::Error> {
         self.manager.close().map_err(Into::into)
     }
@@ -423,18 +428,19 @@ mod test {
     use core::iter::Take;
     use std::collections::HashMap;
     use std::iter::Peekable;
-    use std::num::NonZeroUsize;
+    use std::num::{NonZeroU64, NonZeroUsize};
     use std::ops::{Deref, DerefMut};
     use std::path::Path;
 
     use firewood_storage::{
         CheckOpt, CheckerError, HashedNodeReader, IntoHashType, LinearAddress, MaybePersistedNode,
-        NodeStore, RootReader, TrieHash,
+        NodeStore, TrieHash,
     };
+    use nonzero_ext::nonzero;
 
     use crate::db::{Db, Proposal, UseParallel};
     use crate::manager::RevisionManagerConfig;
-    use crate::v2::api::{Db as _, DbView, Proposal as _};
+    use crate::v2::api::{Db as _, DbView, HashKeyExt, Proposal as _};
 
     use super::{BatchOp, DbConfig};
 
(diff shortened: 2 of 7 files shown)


RodrigoVillar:
Ran a reexecution test from blocks 101-250k against the `master` branch of AvalancheGo here: 

https://github.com/ava-labs/firewood/actions/runs/21929042863/job/63328596761

https://github.com/ava-labs/avalanchego/actions/runs/21929048753/job/63328621624

ffi/firewood.go
```diff
@@ -113,14 +113,18 @@ type config struct {
 	rootStore bool
 	// expensiveMetricsEnabled controls whether expensive metrics recording is enabled.
 	expensiveMetricsEnabled bool
+	// deferredPersistenceCommitCount determines the maximum number of unpersisted
+	// revisions that can exist at a given time.
+	deferredPersistenceCommitCount uint
```
rkuris: Per the golang docs, "The int , uint , and uintptr types are usually 32 bits wide on 32-bit systems and 64 bits wide on 64-bit systems". However, on the rust side, we're expecting 64-bit values (C.uint64_t).
```suggestion
	deferredPersistenceCommitCount uint64
```

ffi/firewood.go
```diff
@@ -193,6 +197,15 @@ func WithExpensiveMetrics() Option {
 	}
 }
 
+// WithDeferredPersistenceCommitCount sets the maximum number of unpersisted revisions
+// that can exist at a time. Note: `commitCount` must be greater than 0.
+// Default: 1
+func WithDeferredPersistenceCommitCount(commitCount uint) Option {
```
rkuris: same typing issue uint->uint64

ffi/firewood.go
```diff
@@ -245,20 +258,24 @@ func New(dbDir string, nodeHashAlgorithm NodeHashAlgorithm, opts ...Option) (*Da
 	if conf.freeListCacheEntries < 1 {
 		return nil, fmt.Errorf("free list cache entries must be >= 1, got %d", conf.freeListCacheEntries)
 	}
+	if conf.deferredPersistenceCommitCount == 0 {
+		return nil, fmt.Errorf("deferred persistence commit count must be >= 1, got %d", conf.deferredPersistenceCommitCount)
```
rkuris: nit: since only zero triggers this error:
```suggestion
		return nil, fmt.Errorf("deferred persistence commit count must not be 0")
```

firewood/src/v2/api.rs
```diff
@@ -197,6 +197,9 @@ pub enum Error {
     // Feature not supported in this build
     #[error("feature not supported in this build: {0}")]
     FeatureNotSupported(String),
+
+    #[error("commit count must be positive")]
+    InvalidConversionToCommitCount,
```
rkuris: nit: naming: recommend "ZeroCommitCount"

firewood/src/db.rs
```diff
@@ -1392,6 +1398,195 @@ mod test {
...
+        let revision = db.view(root_hash).unwrap();
+        let new_value = revision.val(b"foo").unwrap().unwrap();
+
+        assert_eq!(value, new_value.as_ref());
+    }
+
+    #[test]
+    fn test_deferred_persist_after_reaching_sub_interval() {
+        const COMMIT_COUNT: NonZeroU64 = nonzero!(10u64);
+        const SUB_INTERVAL: u64 = COMMIT_COUNT.get() / 2;
```
rkuris: nit: This feels a little fragile to me, as it depends on the implementation. I think it would be safer to set SUB_INTERVAL to COMMIT_COUNT - 1 in case this changes later.

firewood/src/db.rs
```diff
@@ -1392,6 +1398,195 @@ mod test {
...
+            let proposal = db.propose(batch).unwrap();
+            proposal.commit().unwrap();
+        }
+
+        // Wait for the background thread to finish persisting
+        db.wait_persisted();
+
+        // Verify the root is now persisted
+        let revision = db.manager.current_revision();
+        let is_persisted = revision
```
rkuris: Observation only: This test really peeks under the covers. I wonder if there's a better approach.

firewood/src/db.rs
```diff
@@ -1392,6 +1398,195 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_deferred_persistence_root_store() {
+        const NUM_COMMITS: usize = 20;
+        const COMMIT_COUNT: NonZeroU64 = nonzero!(10u64);
+        const MAX_IN_MEMORY_REVISIONS: usize = 5;
+
+        // Revisions to verify after reopening (1-indexed commit numbers)
+        const CHECKPOINTS: [usize; 4] = [5, 10, 15, 20];
```
rkuris: Peeking a bit too much here as well. I wonder if we should be asking the revision manager this question via a test-only API.

firewood/src/db.rs
```diff
@@ -1392,6 +1398,195 @@ mod test {
...
+            let proposal = db.propose(batch).unwrap();
+            if i % 5 != 0 {
+                unpersisted_roots.push(proposal.root_hash().unwrap().unwrap());
+            }
+            proposal.commit().unwrap();
+        }
+
+        let db = db.reopen();
+
+        // Commits 1-4 and 6-9 were not persisted and should not be accessible
```
rkuris: If they actually are persisted, we're not violating the agreement, so I'm not sure we should be testing this.

firewood/src/persist_worker.rs
```diff
@@ -94,13 +106,18 @@ impl PersistWorker {
...
-    pub(crate) fn new(header: NodeStoreHeader, root_store: Option<Arc<RootStore>>) -> Self {
+    pub(crate) fn new(
+        commit_count: NonZeroU64,
+        header: NodeStoreHeader,
+        root_store: Option<Arc<RootStore>>,
+    ) -> Self {
         let (sender, receiver) = channel::unbounded();
-        let persist_interval = nonzero!(1u64);
+        let persist_interval =
+            NonZeroU64::new((commit_count.get() / 2).max(1)).expect("should be positive");
```
rkuris: Hmm, this means 1->1, 2->1, 3->1, 4->2, 5->2, which seems wrong.

What about `commit_count.get().div_ceil(2)`, which gives 1->1, 2->1, 3->2, 4->2, 5->3

Also "should be positive" becomes "a nonzero div_ceil(2) is always positive"

firewood/src/persist_worker.rs
```diff
@@ -140,9 +158,9 @@ impl PersistWorker {
         nodestore: NodeStore<Committed, FileBacked>,
     ) -> Result<(), PersistError> {
         if self.shared.root_store.is_none() {
-            // Always send the reap message, even for empty tries. A committed
-            // revision with no root can still carry deleted nodes from the
-            // previous revision that need their disk space freed.
+            // Always send the reap message, even for empty tries. Even if this
+            // revision wasn't committed, it could be the case that this
```
rkuris:
```suggestion
            // revision wasn't persisted, it could be the case that this
```

firewood/src/persist_worker.rs
```diff
@@ -350,10 +370,22 @@ impl PersistLoop {
...
+                    self.last_committed_revision = Some(revision);
                 }
             }
         }
 
+        if let Some(revision) = self.last_committed_revision.take() {
+            // `num_commits` was already incremented past the last commit,
+            // so subtract 1 to get the actual last commit number.
+            let last_commit_num =
+                NonZeroU64::new(num_commits.get() - 1).expect("at least one commit was processed");
```
rkuris: Is the expectation actually correct? I'm not fully convinced. What if `revision` is the first revision?

firewood/src/db.rs
RodrigoVillar: > If they actually are persisted, we're not violating the agreement

This is true. However, my intuition for adding this test was that if we persist more often than necessary, that this is not desirable as:

1. We'll be utilizing more disk space than needed
2. If we persist more often than needed, that this could trigger backpressure which reduces our performance.

rkuris: Discussed offline. We should test that there are 1-2 persists if the persist count is 10 and we do 10 commits, without caring exactly which ones are persisted.

It might be better to use the metric that says how many persist operations occurred (assuming there is one).

ffi/firewood.go
RodrigoVillar: Good call; done.

ffi/firewood.go
RodrigoVillar: Done.

ffi/firewood.go
RodrigoVillar: Done.

firewood/src/v2/api.rs
RodrigoVillar: Done.

firewood/src/persist_worker.rs

firewood/src/persist_worker.rs
RodrigoVillar: I interpreted this in two ways:

- If we haven't committed anything and we're shutting down: this code won't execute.
- If we've committed one revision and we're shutting down:
    - `num_commits` is initialized to `1` and we added `1` when we persisted the revision and so `num_commits` will always be at least 2.

That being said, this is a code smell that we can refactor this to remove the `expect`, which is what I did.

firewood/src/persist_worker.rs

firewood/src/db.rs
RodrigoVillar: I agree; modified this test to test that once least one revision was persisted (i.e relax restriction that revision had to be persisted at a specific checkpoint).

firewood/src/db.rs
RodrigoVillar: Updated this test to verify that we've persisted at most 2 revisions; this code also now uses the `revision_persist_status` method to avoid having to inspect the `nodestore` code.

firewood/src/db.rs
RodrigoVillar: Updated this code so that it uses the `revision_persist_status` method, which avoids having to peek into the `nodestore` logic.

firewood/src/db.rs
RodrigoVillar: Yeah this test is really trying to test "did we go more than `COMMIT_COUNT` revisions without persisting?" Refactored this test so that it is verifying the question above + is using the `revision_persist_status` method.

ffi/firewood.go
```diff
@@ -245,20 +259,24 @@ func New(dbDir string, nodeHashAlgorithm NodeHashAlgorithm, opts ...Option) (*Da
 	if conf.freeListCacheEntries < 1 {
 		return nil, fmt.Errorf("free list cache entries must be >= 1, got %d", conf.freeListCacheEntries)
 	}
+	if conf.deferredPersistenceCommitCount == 0 {
```
rkuris: nit: We should map the error so we're only checking for 0 on the rust side. Why? less code, less tests, one place to change if we ever allow 0 for some reason.

firewood/src/db.rs
```diff
@@ -1392,6 +1398,210 @@ mod test {
...
+            .iter()
+            .filter(|&hash| {
+                db.manager
+                    .revision_persist_status(hash.clone())
+                    .is_ok_and(|is_persisted| is_persisted)
+            })
+            .count();
+
+        assert!(
+            persisted_count <= 2,
```
rkuris: nit: I realize this is in another test, but it can't hurt to also disallow 0
```suggestion
            persisted_count > 0 && persisted_count <= 2,
```
or, if you'd prefer to be more rusty and perhaps less readable:
```
            1..=2.contains(&persisted_count),
```

The error message would help maintain readability, but this is really a nit so your choice.

firewood/src/manager.rs
```diff
@@ -547,6 +557,22 @@ mod tests {
...
             self.persist_worker.wait_persisted();
         }
+
+        /// Returns true if the root node (if it exists) of this revision is
+        /// persisted. Otherwise, returns false.
+        ///
+        /// ## Errors
+        ///
+        /// Returns an error if the revision does not exist.
+        pub(crate) fn revision_persist_status(
```
rkuris, line 567: I think this is only used in tests?

```suggestion
#[cfg(test)]
        pub(crate) fn revision_persist_status(
```

RodrigoVillar: This function is under `mod tests`, which already has the `#[cfg(test)]` attribute applied.

ffi/firewood.go
RodrigoVillar: Done.

firewood/src/db.rs
RodrigoVillar: Went with the first option.

PR #1659, opened 2026-02-05
fix(track-performance): fail hard on gh errors 

Elvis339:
## Why this should be merged
The benchmark script silently continued after gh CLI errors (e.g., HTTP 422), causing it to pick up unrelated workflow runs. The root cause was command substitution ($()) which creates a subshell where set -e doesn't propagate failures.

## How this works
Uses global `TRIGGER_RUN_ID` to avoid the `$()` subshell pattern:
```
# Before (broken): set -e doesn't catch gh failures in subshell
run_id=$(trigger_workflow "$test")

# After (fixed): direct call, set -e works properly  
trigger_workflow "$test"  # Sets TRIGGER_RUN_ID on success
run_id="$TRIGGER_RUN_ID"
```

Additional improvements:
- Fail fast after 5 consecutive gh run list failures (vs waiting 180s)
- Include actual error message in timeout output for debugging
- Add expected_end check for more accurate run matching
- Simplify check_status to avoid $() pattern

## How this was tested
```
gh workflow run "C-Chain Reexecution Performance Tracking" \
  --repo ava-labs/firewood \
  --ref es/perf-tracking-local-improvements \
  -f test="firewood-101-250k" \
  -f avalanchego="2f6b55f384af4e313ae4b2aa8a15f161552e2222" \
  -f runner="avalanche-avalanchego-runner-2ti"
```
https://github.com/ava-labs/firewood/actions/runs/21725593504/job/62666699097

Workflow correctly fails on the 422 error instead of continuing.

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -239,8 +239,20 @@ impl RevisionManager {
     #[fastrace::trace(short_name = true)]
     #[crate::metrics("proposal.commit", "proposal commit to storage")]
     pub fn commit(&self, proposal: ProposedRevision) -> Result<(), RevisionManagerError> {
+        // Hold a write lock on `in_memory_revisions` for the duration of the
+        // critical section (steps 1-5). This is necessary because:
+        // 1. Without the lock, two proposals with the same parent could pass the
+        //    commit check simultaneously, allowing both to commit.
+        // 2. New proposals rely on the latest committed revision via
+        //    `current_revision()`, which takes a read lock on `in_memory_revisions`.
+        //    The write lock here prevents proposals from being created against
+        //    an older revision while a newer revision is mid-commit.
+        let mut in_memory_revisions = self.in_memory_revisions.write();
+
         // 1. Commit check
-        let current_revision = self.current_revision();
+        let current_revision = in_memory_revisions
+            .back()
+            .expect("there is always one revision");
         if !proposal.parent_hash_is(current_revision.root_hash()) {
             return Err(RevisionManagerError::NotLatest {
                 provided: proposal.root_hash(),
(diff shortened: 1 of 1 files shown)


rkuris:
I am suspicious that this will not fix the problem, since the script should exit if these commands fail due to the shell directive at the beginning.

Elvis339:
> I am suspicious that this will not fix the problem, since the script should exit if these commands fail due to the shell directive at the beginning.

The issue was `run_id=$(trigger_workflow)` runs in a subshell where `set -e` doesn't propagate failures.
Fixed by calling `trigger_workflow` directly (no `$()`), so `set -e` now catches `gh` errors properly. Removed the defensive checks since they're redundant now.

Test: https://github.com/ava-labs/firewood/actions/runs/21725593504/job/62666699097

Elvis339:
Successful run:  https://github.com/ava-labs/firewood/actions/runs/21726798619/job/62670862604

```
 gh workflow run "C-Chain Reexecution Performance Tracking" \
  --repo ava-labs/firewood \
  --ref es/perf-tracking-local-improvements \
  -f test="firewood-101-250k" \
  -f runner="avalanche-avalanchego-runner-2ti"
```

scripts/bench-cchain-reexecution.sh
```diff
@@ -257,11 +280,89 @@ trigger_workflow() {
...
+
+# GitHub's API can return transient errors (502/503, 403 rate limits) during
+# long-running jobs. We retry gh run watch with exponential backoff, then fall
+# back to polling gh run view for final status. This prevents false failures
+# when the actual benchmark succeeds but API monitoring fails.
 wait_for_completion() {
     local run_id="$1"
-    log "Waiting for run $run_id"
+    local interval
+    interval=$(calculate_watch_interval)
```
rkuris:
```suggestion
    local interval=$(calculate_watch_interval)
```

Elvis339:
> IMO we should also use `set -e` to catch if someone adds something that

I'm not sure I understand, we have `set -euo pipefail` here.

https://github.com/ava-labs/firewood/blob/20b575de3a36967f7435e3a6277c8e72ec678bd1/scripts/bench-cchain-reexecution.sh#L2

PR #1662, opened 2026-02-09
feat(fwdctl/launch): add feature-gated fwdctl EC2 deploy workflow for benchmark setup (1/4)

### LINKED ISSUE #1318: feat: Migrate from aws-launch.sh to fwdctl launch

AminR443:
## Why this should be merged

This PR adds a new `launch` workflow to `fwdctl` for provisioning benchmark-ready EC2 instances with a consistent cloud-init pipeline. This gives us a repeatable way to spin up reexecution environments with less manual setup and better traceability.

It is useful for:
- Reproducible benchmark environment bootstrapping
- Faster PR/perf validation spins with branch-specific checkouts
- Standardizing launch scripts and execution stages in one config

## How this works

The PR introduces a feature-gated `fwdctl launch deploy` command and supporting modules for:
- Loading and processing stage-based launch config (shared stages + scenarios)
- Template interpolation for `variables.*`, `args.*`, and `branches.*`
- Cloud-init generation and base64 encoding for EC2 user-data
- Instance launch flow: architecture detection, latest Ubuntu AMI lookup, instance creation, wait-for-running, and IP reporting
- Launch metadata tags (including optional branch tags and custom tag)

The default stage config is embedded from `benchmark/launch/launch-stages.yaml`, with support for a user override.

Closes #1318.

## How this was tested
There is some UT. Also tested manually, launched few instances.

## Design Notes

1. **Feature-gated dependencies:** AWS SDK + async runtime are isolated behind `launch` so default `fwdctl` builds stay thin.
2. **Scenario/stage model:** Launch steps are YAML-driven and composable, so benchmark workflows can evolve without hardcoding all logic in Rust.

### DIFF
--- a/.github/workflows/track-performance.yml
+++ b/.github/workflows/track-performance.yml
@@ -5,8 +5,8 @@ name: C-Chain Reexecution Performance Tracking
 on:
   schedule:
     # These cron expressions are matched exactly in configure-benchmark to set parameters
-    # Daily at 05:00 UTC (00:00 ET) - 1M blocks: 40M → 41M
-    - cron: '0 5 * * *'
+    # Weekdays at 05:00 UTC (00:00 ET) - 1M blocks: 40M → 41M
+    - cron: '0 5 * * 1-5'
     # Weekly Saturday at 05:00 UTC (00:00 ET) - 10M blocks: 50M → 60M
     - cron: '0 5 * * 6'
   workflow_dispatch:
@@ -67,7 +67,7 @@ jobs:
 
             # Cron strings must match exactly from schedule definitions above
             case "${{ github.event.schedule }}" in
-              "0 5 * * *")  # Daily at 05:00 UTC: 1M blocks (40M → 41M)
+              "0 5 * * 1-5")  # Weekdays at 05:00 UTC: 1M blocks (40M → 41M)
                 echo "::notice::Schedule: daily-40m-41m (blocks 40000001-41000000)"
                 echo "name=daily-40m-41m" >> "$GITHUB_OUTPUT"
                 echo "start-block=40000001" >> "$GITHUB_OUTPUT"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1063,6 +1063,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1137cd7e7fc0fb5d3c5a8678be38ec56e819125d8d7907411fe24ccb943faca8"
+dependencies = [
+ "crossbeam-channel",
+ "crossbeam-deque",
+ "crossbeam-epoch",
+ "crossbeam-queue",
+ "crossbeam-utils",
+]
+
+[[package]]
+name = "crossbeam-channel"
+version = "0.5.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1077,6 +1099,15 @@ name = "crossbeam-epoch"
 version = "0.9.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
+name = "crossbeam-queue"
+version = "0.3.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0f58bbc28f91df819d0aa2a2c00cd19754769c2fad90579b3592b1c9ba7a3115"
 dependencies = [
  "crossbeam-utils",
 ]
@@ -1528,6 +1559,7 @@ dependencies = [
  "clap",
  "coarsetime",
  "criterion",
+ "crossbeam",
  "ctor",
  "derive-where",
  "env_logger",
@@ -1544,6 +1576,7 @@ dependencies = [
  "integer-encoding",
  "lender",
  "metrics",
+ "nonzero_ext",
  "parking_lot",
  "plain_hasher",
  "pprof",
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -415,7 +415,8 @@ func (db *Database) Revision(root Hash) (*Revision, error) {
 	return rev, nil
 }
 
-// Close releases the memory associated with the Database.
+// Close releases the memory associated with the Database and stops the
+// background persistence thread.
 //
 // This blocks until all outstanding keep-alive handles are disowned or the
 // [context.Context] is cancelled. That is, until all Revisions and Proposals
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -1275,6 +1275,8 @@ struct ValueResult fwd_change_proof_to_bytes(const struct ChangeProofContext *_p
 /**
  * Close and free the memory for a database handle
  *
+ * This also stops the background persistence thread.
+ *
  * # Arguments
  *
  * * `db` - The database handle to close, previously returned from a call to [`fwd_open_db`].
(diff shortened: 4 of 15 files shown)


benchmark/launch/launch-stages.yaml
```diff
@@ -0,0 +1,169 @@
...
+    name: "Downloading blocks"
+    commands:
+      - "sudo -u ubuntu mkdir -p {{ variables.nvme_base }}/ubuntu/exec-data/blocks"
+      - "s5cmd cp 's3://{{ variables.s3_bucket }}/cchain-mainnet-blocks-{{ args.nblocks }}-ldb/*' {{ variables.nvme_base }}/ubuntu/exec-data/blocks/ >/dev/null"
+      - "chown -R ubuntu {{ variables.nvme_base }}/ubuntu/exec-data"
+      - "chmod -R g=u {{ variables.nvme_base }}/ubuntu/exec-data"
+  init-state-dir:
+    name: "Initializing state directory"
+    commands:
+      - "sudo -u ubuntu mkdir -p {{ variables.nvme_base }}/ubuntu/exec-data/current-state"
```
RodrigoVillar, line 148: Rather than starting from genesis, could we add a parameter for users to define the S3 URI for the state they want to start off with (e.g. if I pass in `s3://avalanchego-bootstrap-testing/cchain-current-state-firewood-50m/`, the script should download this for me and set it as the current state)

fwdctl/src/launch/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, ValueEnum)]
+pub enum NBlocks {
+    #[value(name = "10k")]
+    TenK = 10_000,
+    #[value(name = "1m")]
+    OneM = 1_000_000,
+    #[value(name = "50m")]
+    FiftyM = 50_000_000,
```
RodrigoVillar, line 109: Rather than having the only options being executing the first 10k/1m/50m blocks, could we further parameterize this to allow users to specify the block range to execute (for example, a common workflow of mine is to execute from blocks 33m-33.5m)?

benchmark/launch/launch-stages.yaml
```diff
@@ -0,0 +1,169 @@
...
+    name: "Installing Go and Grafana"
+    commands:
+      - "bash {{ variables.nvme_base }}/ubuntu/firewood/benchmark/setup-scripts/install-golang.sh"
+      - "bash {{ variables.nvme_base }}/ubuntu/firewood/benchmark/setup-scripts/install-grafana.sh"
+  clone-repos:
+    name: "Cloning repositories"
+    commands:
+      - "snap install task --classic"
+      - "sudo -u ubuntu -D {{ variables.nvme_base }}/ubuntu git clone --depth 100 {{ branches.avalanchego }} {{ variables.avalanchego_repo }}"
+      - "sudo -u ubuntu -D {{ variables.nvme_base }}/ubuntu git clone --depth 100 {{ branches.libevm }} {{ variables.libevm_repo }}"
```
RodrigoVillar, line 107: q: what happens if a `libevm` branch isn't specified? Would we be using the `master`/`main` branch of `libevm` or would we not be overriding the `libevm` version specified in AvalancheGo/Coreth?

benchmark/launch/launch-stages.yaml
AminR443: You can simply add a scenario that does this, and that's why I didn't add this scenario by default. Next PRs simplify doing this by adding stage variables, I suggest you take a look at #1664 and what I did for replay log generation. I meant to add a follow up PR to allow custom block ranges, but got mixed up and forgot to add. I'll try to add it in the last PR or create a new follow up PR for it. I don't want to add too much customizing logic to the rust code, and would prefer to keep it in the yaml file so it's both readable and easily modifiable.

fwdctl/src/launch/mod.rs
AminR443: See above.

benchmark/launch/launch-stages.yaml
AminR443: This mirrors the behavior of aws-launch.sh and checks out master if not specified. We should probably not do that; if branch isn't specified I think the safest option would be going with the version specified in avalanchego/coreth.

fwdctl/src/launch/mod.rs
RodrigoVillar: > I meant to add a follow up PR to allow custom block ranges, but got mixed up and forgot to add. I'll try to add it in the last PR or create a new follow up PR for it.

Sounds good, being able to set custom block ranges will make this feature 10x more useful.

benchmark/launch/launch-stages.yaml
RodrigoVillar: I took a look at #1664 and I see the main use case of scenarios - to define complex behavior that a CLI can't encapsulate (e.g. `snapshotter`). However, I still don't see why I should need to write code (i.e. define a scenario) if I just want to override the starting state? For example, setting the current state can be as easy as calling:

```
./target/release/fwdctl launch deploy --firewood-branch foo --config firewood --start-block 10_000_000 --end-block 20_000_000 --current-state-dir <INSERT-S3-URI-HERE>
```

Given that this is overriding a dependency and not modifying the actual bootstrapping logic itself, having clients pass in the current state directory via an optional flag seems much simpler.

fwdctl/src/launch/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    let (public_ip, private_ip) = ec2_util::describe_ips(&ec2, &instance_id).await?;
+
+    info!("=== Instance Launched ===");
+    info!("Instance ID: {instance_id}");
+    if let Some(ip) = &public_ip {
+        info!("Public IP:   {ip}");
+    }
+    if let Some(ip) = &private_ip {
+        info!("Private IP:  {ip}");
```
rkuris: Nobody probably ever needs this, so perhaps this should be `debug!`

fwdctl/src/launch/stage_config.rs
```diff
@@ -0,0 +1,331 @@
...
+    ScenarioNotFound(String),
+
+    #[error("Unknown variable: {0}")]
+    UnknownVariable(String),
+}
+
+/// Root configuration structure.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct StageConfig {
+    #[serde(default)]
```
rkuris: whole lotta serde(default) statements here. Why not have it on the whole struct? Is `scenario` a hard requirement?

fwdctl/src/launch/stage_config.rs
```diff
@@ -0,0 +1,331 @@
...
+pub enum StageRef {
+    /// Simple reference by ID
+    Ref(String),
+    /// Reference with overrides, or fully inline stage
+    Override(StageOverride),
+}
+
+/// A stage reference with optional overrides, or a fully inline stage.
+///
+/// If `name` and `commands` are provided, this is a fully inline stage.
```
rkuris, line 86: If `name` is provided and `commands` is not, what happens?

fwdctl/src/launch/stage_config.rs
```diff
@@ -0,0 +1,331 @@
...
+/// Processes `{{ namespace.key }}` placeholders in a template string.
+fn process_template(template: &str, ctx: &TemplateContext) -> Result<String, ConfigError> {
+    let mut out = String::with_capacity(template.len());
+    let mut rest = template;
+
+    while let Some((before, after_open)) = rest.split_once("{{") {
+        out.push_str(before);
+        let Some((expr, after_close)) = after_open.split_once("}}") else {
+            // No closing braces: keep the remainder unchanged
+            out.push_str("{{");
```
rkuris: This should be an error I think.

fwdctl/src/launch/stage_config.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/// Template context for variable interpolation.
+///
+/// Variables are resolved in this order:
+/// 1. `variables.*` - from config file and stage-specific variables
+/// 2. `args.*` - from CLI arguments
+/// 3. `branches.*` - git branch overrides
+#[derive(Debug, Clone, Default)]
+pub struct TemplateContext {
```
rkuris, line 127: These are all <String, String> hashmaps, can this instead be:
`pub struct TemplateContext(HashMap<String, HashMap<String, String>>);`

benchmark/launch/launch-stages.yaml
AminR443: The reason is simply not to pollute arg space, as each scenario can have it's own variables. But I see how the feature you're describing is desirable. In the follow up PR that I mentioned I'll make, I'll add manual override support for variables and arguments so it covers your usecase.

fwdctl/src/launch/mod.rs
AminR443: Done.

fwdctl/src/launch/stage_config.rs
AminR443: Resolved. We have default now and validation later on the values.

fwdctl/src/launch/stage_config.rs
AminR443: Done. New error created for capturing this.

fwdctl/src/launch/stage_config.rs
AminR443: Added a new error.

fwdctl/src/launch/stage_config.rs
AminR443: I chose to keep this as is; It's much cleaner when they're used.

PR #1663, opened 2026-02-09
feat(fwdctl/launch): add SSM-based monitor/follow flow for cloud-init and bootstrap progress (2/4)

AminR443:
## Why this should be merged

This PR introduces launch monitoring for `fwdctl` using AWS SSM. It closes the visibility gap after instance startup by exposing cloud-init stage progress and optional bootstrap re-execution progress without requiring SSH access.

This is important for:
- Faster debugging when launch stages fail
- Better operator visibility during long-running setup/replay flows
- More deterministic failure context (stage, command index, exit code)

## How this works

The launcher now writes structured cloud-init state to `/var/log/cloud-init-state.json` and updates it after each stage/command via a small helper script. A new SSM monitor module polls this state and renders progress in real time.

## How this was tested

Manually.

## Design Notes

1. **SSM-first monitoring:** SSM is chosen over SSH so monitoring works with standard AWS controls and without opening additional access paths.
2. **Structured state file:** stage state is persisted as JSON in cloud-init, making progress/failure reporting deterministic instead of log-pattern-based.


fwdctl/src/launch/cloud_init.rs
```diff
@@ -142,11 +142,106 @@ impl CloudInitContext {
...
+const STATE_FILE_TMP: &str = "/var/log/cloud-init-state.json.tmp";
+const STATE_HELPER: &str = "/usr/local/bin/fwdctl-state";
+
+fn state_helper_script() -> String {
+    format!(
+        r#"#!/usr/bin/env bash
+set -euo pipefail
+
+STATE_FILE="{STATE_FILE}"
+STATE_FILE_TMP="{STATE_FILE_TMP}"
```
rkuris: nit: use mktemp instead to create a temporary file, preventing pollution of /var/log in the extremely unlikely event of a crash.

AminR443: Good note. Fixed.

PR #1664, opened 2026-02-09
feat(fwdctl/launch): add scenario-driven deploy, list/kill instance management, and dry-run planning (3/4)

AminR443:
## Why this should be merged

This PR upgrades `fwdctl launch` from deploy/monitor-only into a full lifecycle management tool. It adds planning (`--dry-run`), managed instance operations (`list`, `kill`), and scenario-based launches.

## How this works

The launch CLI now supports:
- `fwdctl launch list` to show `ManagedBy=fwdctl` instances (with `--running` / `--mine`)
- `fwdctl launch kill` to terminate a specific instance, your instances, or all managed instances
- `fwdctl launch deploy --scenario <name>` to pick a scenario from `benchmark/launch/launch-stages.yaml`
- `fwdctl launch deploy --dry-run [plan|plan-with-cloud-init]` to preview actions without creating resources
- `fwdctl launch deploy --follow [follow|follow-with-progress]` for log/progress streaming modes

## How this was tested
Some UT. mostly manually launching and ensuring all the cases work fine.

## Design Notes

1. `list`/`kill` are intentionally scoped to `ManagedBy=fwdctl` resources to avoid accidental operations on unrelated instances.
3. `plan-with-cloud-init` was added so generated cloud-init can be inspected/reviewed before launch.

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -28,7 +28,9 @@ typedef enum NodeHashAlgorithm {
 } NodeHashAlgorithm;
 
 /**
- * FFI context for a parsed or generated change proof.
+ * FFI context for a parsed or generated change proof. This change proof has not
+ * been verified. Calling `verify` on it will generate a `VerifiedChangeProofContext`
+ * and consume the `proof` and replacing it with None.
  */
 typedef struct ChangeProofContext ChangeProofContext;
 
@@ -56,11 +58,26 @@ typedef struct IteratorHandle IteratorHandle;
 typedef struct ProposalHandle ProposalHandle;
 
 /**
+ * FFI context for a proposed change proof. It is created from calling `propose`
+ * on a `VerifiedChangeProofContext` and stores the database and proposal handle.
+ * Calling `commit` on it will consume the proof.
+ */
+typedef struct ProposedChangeProofContext ProposedChangeProofContext;
+
+/**
  * FFI context for for a parsed or generated range proof.
  */
 typedef struct RangeProofContext RangeProofContext;
 
 typedef struct RevisionHandle RevisionHandle;
+
+/**
+ * FFI context for a verified change proof. It is created from calling `verify`
+ * on a `ChangeProofContext` and stores the parameters of that call in `params`.
+ * Calling `propose` on it will consume the proof to create a
+ * `ProposedChangeProofContext`.
+ */
+typedef struct VerifiedChangeProofContext VerifiedChangeProofContext;
 
 /**
  * A database hash key, used in FFI functions that require hashes.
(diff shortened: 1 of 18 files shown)


fwdctl/src/launch/ec2_util.rs
rkuris: move this to a const as it's likely to change later

rkuris: This should also be a const with a description so we know what this number represents

rkuris: Is this true for all AWS images?

rkuris: const

rkuris, line 156: Aren't these the defaults? I hope so!

rkuris, line 198: This needs more comments. What are these magic branch patterns?

benchmark/launch/launch-stages.yaml
rkuris, line 75: This is difficult or impossible to review since I can't compare the original script to what you've got here without doing a lot of work. Suggestions welcome.

fwdctl/src/launch/ec2_util.rs
```diff
@@ -18,9 +18,24 @@ use super::{DeployOptions, LaunchError};
 
 const WAIT_RUNNING_TIMEOUT: Duration = Duration::from_secs(300);
```
rkuris, line 20: Could use a doc comment, maybe "Time to wait after launch for the instance to transition to a RUNNING status". The others could use the same.

fwdctl/src/launch/ec2_util.rs
AminR443: Done.

AminR443: Done.

AminR443: Replaced with adding a function that gets the root volume name. Thanks for pointing this out. It'd be the same for ubuntu images but this could break on image change.

AminR443: Done.

AminR443: Yeah, these sound like to have defaults, but actually they are required. Not setting them will produce an explicit error when running; "The request must contain the parameter MaxCount" and "The request must contain the parameter MinCount".

AminR443: Explained more.

benchmark/launch/launch-stages.yaml
AminR443: I'd suggest doing a dry-run with cloud-init dump so you could inspect the resulting cloud init manually. It'll be a little bit hard still because of progress tracking commands, but you could remove them and then do the comparison.

To do this:
```bash
AWS_PROFILE=<your-profile> cargo run --bin fwdctl --features launch -- launch deploy --dry-run=plan-with-cloud-init
```

You can check this diff to that I generated:
https://www.diffchecker.com/GglAZd0v/

Note: The branch args are not present because I ran the default configuration.

AminR443: It's pretty easy to verify with the diff that the original commands are mostly untouched. The rust install changed because one of the original commands had non-successful exit code, couldn't write to the specified directory.

fwdctl/src/launch/ec2_util.rs
AminR443: Added this and now all the others have doc comments too.

fwdctl/src/launch/mod.rs
```diff
@@ -26,6 +26,9 @@ pub enum LaunchError {
     #[error("AWS SDK error: {0}")]
     AwsSdk(String),
 
+    #[error("{0}")]
```
rkuris: nit: shouldn't this be
```suggestion
    #[error("Validation failed: {0}")]
```

Otherwise maybe this should be `Other(String)`?

AminR443: Good catch, Fixed.

PR #1668, opened 2026-02-11
perf: Increase limits and monitor freelist cache size better

rkuris:
## Why this should be merged

The freelist cache is tiny and isn't preallocated, so setting it large doesn't hurt. Updated the default from 40K to 1M. Each entry is roughly 40 bytes, so this is about 40MB of memory to significantly reduce disk churn.

Also adds a new metric, cache.freelist.size (gauge) and associated documentation.

## How this works

The new metric provides visibility into actual number of entries in the LRU. If this count is the same as the max size, then it's probably too small, but we need some hard limit to avoid a possible unbounded memory growth.

Also removed a bunch of warnings regarding gauge usage, since the cast to f64 is expected and now it's all done inside the macro, we don't need to suppress the warnings.

## How this was tested

Unit tests. We need to test this further with the nightly re-execution tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1756,6 +1756,7 @@ dependencies = [
  "io-uring",
  "log",
  "lru",
+ "lru-mem",
  "metrics",
  "metrics-util",
  "nonzero_ext",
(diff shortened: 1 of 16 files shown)


rkuris:
> That "isn't preallocated" part is super important and I wasn't aware before, can we document this better?

I'll create a ticket to rename it to "freelist_memory_limit_kb" and change it to kbytes, with a default of 4096 or so, but that's an API change and we should save that for 0.2.0.

PR #1670, opened 2026-02-11
chore: add deferred persistence benchmark

RodrigoVillar:
## Why this should be merged

Having benchmarks to test the performance of deferred persistence across different commit counts is useful for observability.

## How this works

Adds a benchmark `bench_deferred_persistence()` to test deferred persistence with commit counts `1` and `100`. This benchmark is run daily along with our other benchmarks.

## How this was tested

CI

### DIFF
--- a/firewood/benches/defer_persist.rs
+++ b/firewood/benches/defer_persist.rs
@@ -1,0 +1,65 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use firewood::db::{BatchOp, Db, DbConfig};
+use firewood::v2::api::{Db as _, Proposal as _};
+use firewood_storage::NodeHashAlgorithm;
+use rand::{Rng, distr::Alphanumeric};
+use std::iter::repeat_with;
+use std::num::NonZeroU64;
+
+#[expect(clippy::unwrap_used)]
+fn bench_deferred_persistence<const N: usize, const COMMIT_COUNT: u64>(criterion: &mut Criterion) {
+    const KEY_LEN: usize = 4;
+    let rng = &firewood_storage::SeededRng::from_option(Some(1234));
+    let commit_count = NonZeroU64::new(COMMIT_COUNT).unwrap();
+
+    criterion
+        .benchmark_group("deferred_persistence")
+        .sample_size(20)
+        .bench_function(format!("commit_count_{COMMIT_COUNT}"), |b| {
+            b.iter_batched(
+                || {
+                    let batch_ops: Vec<_> =
+                        repeat_with(|| rng.sample_iter(&Alphanumeric).take(KEY_LEN).collect())
+                            .map(|key: Vec<_>| BatchOp::Put {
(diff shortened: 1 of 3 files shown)


firewood/benches/defer_persist.rs
```diff
@@ -0,0 +1,60 @@
...
+use firewood_storage::NodeHashAlgorithm;
+use rand::{Rng, distr::Alphanumeric};
+use std::iter::repeat_with;
+use std::num::NonZeroU64;
+
+#[expect(clippy::unwrap_used)]
+fn bench_deferred_persistence<const N: usize, const COMMIT_COUNT: u64>(criterion: &mut Criterion) {
+    const KEY_LEN: usize = 4;
+    let rng = &firewood_storage::SeededRng::from_option(Some(1234));
+    let commit_count = NonZeroU64::new(COMMIT_COUNT).unwrap();
```
rkuris, line 16: super nit: nonzero macro would have worked here since it's a compile time constant.

firewood/benches/defer_persist.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    criterion
+        .benchmark_group("deferred_persistence")
+        .sample_size(20)
+        .bench_function(format!("commit_count_{COMMIT_COUNT}"), |b| {
+            b.iter_batched(
+                || {
+                    let batch_ops: Vec<_> =
+                        repeat_with(|| rng.sample_iter(&Alphanumeric).take(KEY_LEN).collect())
```
rkuris, line 25: nit: looks a lot like other tests. Could we have consolidated code here?

firewood/benches/defer_persist.rs
```diff
@@ -0,0 +1,60 @@
...
+                },
+                criterion::BatchSize::SmallInput,
+            );
+        });
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default();
+    targets = bench_deferred_persistence::<1_000, 1>, bench_deferred_persistence::<1_000, 100>
```
rkuris: nit: comment explaining why 1, 100, and 1000 was chosen. I realize we never did this earlier, and maybe there aren't any good reasons :)

firewood/benches/defer_persist.rs
RodrigoVillar: Unfortunately, applying the `nonzero` macro here fails as seen below:

```
generic parameters may not be used in const operations
const parameters may only be used as standalone arguments here, i.e. `COMMIT_COUNT`
```

firewood/benches/defer_persist.rs
RodrigoVillar: Yes! However, refactoring IMO falls outside the scope of this PR, hence why I passed on it. Creating an issue for refactoring our benchmarks to reduce duplicate code.

firewood/benches/defer_persist.rs
RodrigoVillar: I chose these values on the fly. However, this does call out a good point - these benchmarks could do a better job in showing the performance gain (or loss) from choosing different commit counts. I went ahead and added more benchmark targets to get a performance curve from persisting every commit to persisting just once.

PR #1671, opened 2026-02-11
feat!: Remove error returned by Hash

alarso16:
## Why this should be merged

I realized that the error returned by the hash function can only happen if the db handle is empty. Otherwise, it will always be nil. Why even give the opportunity to return an error?

## How this works

Removes errors from the signature for everything hash-related.

## How this was tested

CI

### DIFF
--- a/benchmark/launch/launch-stages.yaml
+++ b/benchmark/launch/launch-stages.yaml
@@ -68,23 +68,31 @@ variables:
   libevm_repo: "https://github.com/ava-labs/libevm.git"
   s3_bucket: "avalanchego-bootstrap-testing"
 
+# Reviewability note:
+# The `stages` list below intentionally mirrors the cloud-init `runcmd` flow in
+# `benchmark/bootstrap/aws-launch.sh` (around lines 410-485). Keep stage order
+# aligned with that script so reviewers can compare stage-by-stage.
 stages:
+  # aws-launch.sh: install rust (runcmd lines 410-417)
   rust-install:
     name: "Installing Rust"
     commands:
       - "install -d -m 0755 -o ubuntu -g ubuntu /usr/local/rust"
       - "sudo -u ubuntu --login env RUSTUP_HOME=/usr/local/rust CARGO_HOME=/usr/local/rust sh -c 'curl https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path'"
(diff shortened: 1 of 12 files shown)


ffi/firewood.go
```diff
@@ -347,14 +347,15 @@ func (db *Database) Get(key []byte) ([]byte, error) {
...
 //
 // This function conflicts with all other calls that access the latest state of the database,
 // and will lock for the duration of this function.
-func (db *Database) Root() (Hash, error) {
+func (db *Database) Root() Hash {
 	db.handleLock.RLock()
 	defer db.handleLock.RUnlock()
 	if db.handle == nil {
-		return EmptyRoot, errDBClosed
+		return EmptyRoot
```
alarso16, line 372: This is the crux of the PR. There are no other possible errors, so it would be really nice to not have deal with this, but it's kind of strange behavior

ffi/src/handle.rs
```diff
@@ -180,9 +180,9 @@ impl DatabaseHandle {
     ///
     /// # Errors
     ///
-    /// An error is returned if there was an i/o error while reading the root hash.
+    /// Never errors.
     pub fn current_root_hash(&self) -> Result<Option<HashKey>, api::Error> {
```
demosdemon: can go ahead and remove the result here too. it's a bit weird to have "Errors: never errors". 

I suspect you didn't because of `invoke_with_handle`. The error you get is likely because there's no `From<Option<api::HashKey>>` for `HashResult`. Instead, there's this:

https://github.com/ava-labs/firewood/blob/99d5a3617cc4ab599fe9d82b9d0e5d93959a2b60/ffi/src/value/results.rs#L174-L182

Should be able to remove the result by adding an `impl From<Option<api::HashKey>> for HashResult`.

alarso16: That is why! I'll do this instead

demosdemon: Might need both `From` impls for the other api calls or might be able to replace this one.

alarso16: Is this the way you expected?

demosdemon:
Since this is a breaking change for the Go api, we should wait to merge until after 0.1.1 is released.

rkuris (approved):
Yeah there's no way to get an error from a revision since the hashes are known before the call is made.

This is a breaking change so let's not merge it until 0.2.0

PR #1675, opened 2026-02-12
refactor(perf): calculate AreaIndex in as_bytes methods

### LINKED ISSUE #1114: Node::as_bytes should fill in the area index byte
We shouldn't allow a serialized node to exist unless it also contains the area_size byte at the beginning.

In as_bytes, `T` could be bound to implement `IndexMut` and then calculate the `AREA_SIZE`, rather than expecting the caller to fill it in. You'd need to get the length of the underlying vec also, and there's no std trait for this (you could use the `len_trait` crate).

Or, get rid of the `ExtendableBytes` trait completely and just accept a Vec, which would make this a lot easier.

Then, `Node::as_bytes` could also return either the length or the area size.

rkuris:
## Why this should be merged

Refactored Node::as_bytes and added FreeArea::as_bytes to automatically calculate and return the AreaIndex instead of requiring callers to provide it. This eliminates the previous double-encoding pattern where callers had to encode twice to determine the correct area size.

## How this works

- Node::as_bytes and FreeArea::as_bytes now return a Result<AreaIndex, Error> and calculates the area index from encoded size automatically
- Added NodeAllocator::io_error helper method for error conversion
- Removed double-encoding logic from serialize_node_to_bump
- Pre-reserves exact buffer size in FreeArea::as_bytes
- Uses AsRef<[u8]> and IndexMut trait bounds for flexibility
- Single-pass encoding for all nodes

## How this was tested

TIP == test in production! Let's see if our perf numbers change after merging. The unit test benchmarks won't change, and in fact may be slightly worse since each call is doing more work, but we do avoid calling it twice in the happy path.

Resolves #1114


demosdemon (approved):

storage/src/node/mod.rs
```diff
@@ -311,6 +320,14 @@ impl Node {
                 encoded.extend_from_slice(&l.value);
             }
         }
+
+        // Calculate the area index from the encoded length
+        let area_index = AreaIndex::from_size(encoded.as_ref().len() as u64)?;
+
+        // Update the first byte with the correct area size index
+        encoded[area_size_index_position] = area_index.get();
```
bernard-avalabs, line 333: Given that an AreaIndex is computed from the encoded length, do we need to check that it can fit within a byte?

storage/src/node/mod.rs
```diff
@@ -311,6 +320,14 @@ impl Node {
                 encoded.extend_from_slice(&l.value);
             }
         }
+
+        // Calculate the area index from the encoded length
+        let area_index = AreaIndex::from_size(encoded.as_ref().len() as u64)?;
```
RodrigoVillar:
```suggestion
        let area_index = AreaIndex::from_size(area_size_index_position as u64)?;
```

nit/feel-free-to-ignore

storage/benches/serializer.rs
```diff
@@ -75,7 +75,7 @@ fn manual_deserializer(b: &mut Bencher, input: &Vec<u8>) {
 
 fn to_bytes(input: &Node) -> Vec<u8> {
     let mut bytes = Vec::new();
-    input.as_bytes(firewood_storage::AreaIndex::MIN, &mut bytes);
+    let _area_index = input.as_bytes(&mut bytes).expect("to serialize node");
```
RodrigoVillar, line 78:
```suggestion
    let _ = input.as_bytes(&mut bytes).expect("to serialize node");
```

nit/feel-free-to-ignore

storage/src/node/mod.rs
```diff
@@ -573,7 +589,7 @@ than 126 bytes as the length would be encoded in multiple bytes.
         use std::io::Cursor;
 
         let mut serialized = Vec::new();
-        node.as_bytes(AreaIndex::MIN, &mut serialized);
+        let _area_index = node.as_bytes(&mut serialized).unwrap();
```
RodrigoVillar, line 597:
```suggestion
        let _ = node.as_bytes(&mut serialized).unwrap();
```

nit/feel-free-to-ignore

storage/src/node/mod.rs
rkuris: Not sure i follow. AreaIndex is a NewType that only allows valid indexes into the AREA_INDEX array. This array can't have more than 255 entries.

I don't know that we enforce that anywhere but lots of things would break if that isn't followed. I can't imagine someone wanting that many area sizes though. We probably already have too many.

rkuris: Yeah that's pretty much impossible. The newtype itself is a wrapper around a u8, so it can't be larger than 255. The `get` method on an AreaIndex returns a u8, which can never be larger than a byte.

storage/src/node/mod.rs
rkuris: Actually that's incorrect, but it does call out a possible issue. `encoded` has changed since it was calculated earlier, so these should be different. Arguably we should be subtracting `area_size_index_position` from the current len() but we don't do that, which could result in a larger size than we need.

storage/benches/serializer.rs
rkuris: I like the idea that it's documenting what we're throwing away, so I'm going to leave this one alone.

storage/src/node/mod.rs
rkuris: Fixed in a685bc3af

storage/src/node/mod.rs
```diff
@@ -311,6 +320,19 @@ impl Node {
...
             }
         }
+
+        // Calculate the area index from the encoded length (subtract position to get just this node's size)
+        let node_size = encoded
+            .as_ref()
+            .len()
+            .checked_sub(area_size_index_position)
+            .expect("area_size_index_position should always be <= encoded length");
+        let area_index = AreaIndex::from_size(node_size as u64)?;
```
demosdemon, line 330: this is incorrect. the total size of the area needs to include the area size index, because the area size index controls the size the allocation. The allocation includes us writing the area size index as the first byte. Not including it means we may overwrite our allocation by one byte.

demosdemon: Ah, nevermind. I missed that you were subtracting from the initial len.

storage/src/node/mod.rs
```diff
@@ -311,6 +320,19 @@ impl Node {
                 encoded.extend_from_slice(&l.value);
             }
         }
+
+        // Calculate the area index from the encoded length (subtract position to get just this node's size)
+        let node_size = encoded
+            .as_ref()
+            .len()
+            .checked_sub(area_size_index_position)
+            .expect("area_size_index_position should always be <= encoded length");
```
demosdemon, line 329: this is logically impossible... why not use a wrapping sub?

RodrigoVillar: +1

storage/src/node/mod.rs
```diff
@@ -311,6 +320,19 @@ impl Node {
                 encoded.extend_from_slice(&l.value);
             }
         }
+
+        // Calculate the area index from the encoded length (subtract position to get just this node's size)
+        let node_size = encoded
```
RodrigoVillar, line 325: q: node size is consists of the `area_size` followed by the node serialization?

rkuris: Good question. A node consists of the area size byte followed by the serialized contents. The exact details are available here, including a discussion of the first byte; there are a few special cases there.

https://ava-labs.github.io/firewood/firewood_storage/node/enum.Node.html#method.as_bytes

storage/src/nodestore/persist.rs
```diff
@@ -168,9 +167,10 @@ fn serialize_node_to_bump<'a>(
     node_allocator: &mut NodeAllocator<'_, impl WritableStorage>,
 ) -> Result<(&'a [u8], crate::LinearAddress, usize), FileIoError> {
     let mut bytes = bumpalo::collections::Vec::new_in(bump);
-    shared_node.as_bytes(AreaIndex::MIN, &mut bytes);
-    let (persisted_address, area_size_index) = node_allocator.allocate_node(bytes.as_slice())?;
-    *bytes.get_mut(0).expect("byte was reserved") = area_size_index.get();
+    let area_size_index = shared_node
+        .as_bytes(&mut bytes)
+        .map_err(|e| node_allocator.io_error(e, 0, Some("allocate_node".to_owned())))?;
+    let (persisted_address, _) = node_allocator.allocate_node(bytes.as_slice())?;
```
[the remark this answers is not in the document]
RodrigoVillar: AFAIK, `serialize_node_to_bump()` is the only caller of `allocate_node()`, so it might make sense to remove returning the area index from `allocate_node()`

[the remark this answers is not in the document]
demosdemon: I think this qualifies as "newtype" territory under the general guidance of "parse, don't validate."

After we've serialized a node and calculate its AreaSize, we should wrap the `Vec<u8>`, `Box<[u8]>`, `&[u8]`, etc. in a new type that is aware that the first byte is a guaranteed valid `AreaSize` for the trailing bytes. We can pass this through to the allocator where it can trust the `AreaSize` it gets from the wrapper type.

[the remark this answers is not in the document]
demosdemon: Oh, but also, we could simply pass the `AreaSize` into `allocate_node` instead of the slice.

PR #1677, opened 2026-02-12
feat: node cache entry-based to memory-based LRU

rkuris:
## Why this should be merged

The lru-mem crate provides memory-based LRU eviction with per-entry heap size accounting. This improves cache effectiveness by evicting based on actual memory usage rather than entry count, preventing memory pressure from large nodes.

Additionally, the configuration provided from the EVM is in bytes, and we were doing some shenanigans to estimate the memory usage from that.

## How this works

- Replace entry-based LruCache with lru-mem's MemLruCache for memory-based eviction
- Implement HeapSize trait for all Node types (Path, LeafNode, BranchNode, Node)
- Create CachedNode wrapper around SharedNode to implement foreign traits
- Add dual configuration: deprecated node_cache_size and new node_cache_memory_limit
- Default to 192MB cache capacity (1,500,000 nodes × 128 bytes estimate)
- Add validation to prevent conflicting cache configuration
- Add cache memory metrics: node.cache.memory_used and node.cache.memory_limit
- Update metrics on every cache insert/remove operation (O(1) cost)

Note:
- RevisionManagerConfig::node_cache_size is now deprecated (use node_cache_memory_limit)
- Setting both fields results in ConflictingCacheConfig error

## How this was tested

Unit tests, both old and new!

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -47,7 +47,7 @@ pub struct RevisionManagerConfig {
     /// **Deprecated:** Use `node_cache_memory_limit` instead for memory-based sizing.
     /// If specified, this value is multiplied by 128 to estimate memory usage.
     /// Cannot be specified together with `node_cache_memory_limit`.
-    #[deprecated(since = "0.1.0", note = "Use node_cache_memory_limit instead")]
+    #[deprecated(since = "0.2.0", note = "Use node_cache_memory_limit instead")]
     #[builder(default, setter(strip_option))]
     node_cache_size: Option<NonZero<usize>>,
 
@@ -72,25 +72,27 @@ impl RevisionManagerConfig {
     ///
     /// Returns an error if both `node_cache_size` and `node_cache_memory_limit` are specified.
     #[expect(deprecated)]
-    pub(crate) const fn compute_node_cache_memory_limit(
+    pub(crate) fn compute_node_cache_memory_limit(
         &self,
     ) -> Result<NonZero<usize>, crate::v2::api::Error> {
+        // Convert entry count to memory: size × 128 bytes per node (estimate)
+        const BYTES_PER_NODE_ESTIMATE: usize = 128;
+        // Default: 192MB (equivalent to 1,500,000 nodes × 128 bytes)
+        const DEFAULT_MEMORY_LIMIT: usize = 192_000_000;
+
         match (self.node_cache_size, self.node_cache_memory_limit) {
             (Some(_), Some(_)) => Err(crate::v2::api::Error::ConflictingCacheConfig),
             (Some(size), None) => {
-                // Convert entry count to memory: size × 128 bytes per node (estimate)
-                const BYTES_PER_NODE_ESTIMATE: usize = 128;
+                warn!(
+                    "node_cache_size is deprecated as of 0.2.0; use node_cache_memory_limit instead"
+                );
                 Ok(
                     NonZero::new(size.get().saturating_mul(BYTES_PER_NODE_ESTIMATE))
                         .expect("non-zero size produces non-zero memory"),
                 )
             }
             (None, Some(limit)) => Ok(limit),
-            (None, None) => {
-                // Default: 192MB (equivalent to 1,500,000 nodes × 128 bytes)
-                const DEFAULT_MEMORY_LIMIT: usize = 192_000_000;
-                Ok(NonZero::new(DEFAULT_MEMORY_LIMIT).expect("default is non-zero"))
-            }
+            (None, None) => Ok(NonZero::new(DEFAULT_MEMORY_LIMIT).expect("default is non-zero")),
         }
     }
 }
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -85,7 +85,7 @@ pub type SharedNode = triomphe::Arc<Node>;
 /// This newtype allows implementing foreign traits while maintaining
 /// compatibility with the rest of the codebase that uses `SharedNode` directly.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct CachedNode(pub SharedNode);
+pub(crate) struct CachedNode(pub(crate) SharedNode);
 
 impl From<SharedNode> for CachedNode {
     fn from(node: SharedNode) -> Self {
@@ -125,7 +125,7 @@ impl CachedNode {
     /// with cache misses for oversized nodes.
     ///
     /// Updates cache memory metrics after insertion.
-    pub fn insert_into_cache(
+    pub(crate) fn insert_into_cache(
         self,
         cache: &mut lru_mem::LruCache<LinearAddress, Self>,
         addr: LinearAddress,
@@ -143,7 +143,7 @@ impl CachedNode {
         clippy::cast_precision_loss,
         reason = "Precision loss is acceptable for metrics; only affects values > 2^52 bytes"
     )]
-    pub fn update_cache_metrics(cache: &lru_mem::LruCache<LinearAddress, Self>) {
+    pub(crate) fn update_cache_metrics(cache: &lru_mem::LruCache<LinearAddress, Self>) {
         use firewood_metrics::firewood_set;
         firewood_set!(registry::CACHE_MEMORY_USED, cache.current_size() as f64);
         firewood_set!(registry::CACHE_MEMORY_LIMIT, cache.max_size() as f64);
--- a/storage/src/node/branch.rs
+++ b/storage/src/node/branch.rs
@@ -201,8 +201,8 @@ impl lru_mem::HeapSize for BranchNode {
             .sum();
         self.partial_path
             .heap_size()
-            .saturating_add(value_size)
-            .saturating_add(children_size)
+            .wrapping_add(value_size)
+            .wrapping_add(children_size)
     }
 }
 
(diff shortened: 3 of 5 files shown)


firewood/src/manager.rs
```diff
@@ -42,9 +42,21 @@ pub struct RevisionManagerConfig {
...
 
-    /// The size of the node cache
-    #[builder(default_code = "NonZero::new(1500000).expect(\"non-zero\")")]
-    node_cache_size: NonZero<usize>,
+    /// The size of the node cache (number of entries).
+    ///
+    /// **Deprecated:** Use `node_cache_memory_limit` instead for memory-based sizing.
+    /// If specified, this value is multiplied by 128 to estimate memory usage.
+    /// Cannot be specified together with `node_cache_memory_limit`.
+    #[deprecated(since = "0.1.0", note = "Use node_cache_memory_limit instead")]
```
demosdemon: nit: 0.1.1 or 0.2.0 as 0.1.0 has already been released.

firewood/src/manager.rs
```diff
@@ -53,6 +65,36 @@ pub struct RevisionManagerConfig {
...
+    ///
+    /// Returns an error if both `node_cache_size` and `node_cache_memory_limit` are specified.
+    #[expect(deprecated)]
+    pub(crate) const fn compute_node_cache_memory_limit(
+        &self,
+    ) -> Result<NonZero<usize>, crate::v2::api::Error> {
+        match (self.node_cache_size, self.node_cache_memory_limit) {
+            (Some(_), Some(_)) => Err(crate::v2::api::Error::ConflictingCacheConfig),
+            (Some(size), None) => {
+                // Convert entry count to memory: size × 128 bytes per node (estimate)
```
demosdemon: nit: I would recommend emitting a deprecating warning log message at runtime (maybe even guarded behind a Once) to make it easier to highlight the deprecated fallback behavior.

storage/src/node/mod.rs
```diff
@@ -44,6 +44,17 @@ pub enum Node {
     Leaf(LeafNode),
 }
 
+impl lru_mem::HeapSize for Node {
+    fn heap_size(&self) -> usize {
+        match self {
+            Node::Branch(branch) => {
+                std::mem::size_of::<BranchNode>().saturating_add(branch.heap_size())
```
demosdemon: `saturating_add` is a more expensive cpu operation (`add` + `mov` + `cmovae` on x86 and `adds` + `csinv` on aarch64) than `wrapping_add` (simple `add` for both archs) and `wrapping_add` should be preferred for values that cannot wrap.

Because you're counting the number of *allocated* bytes that you know are definitely *allocated*, the usize value *cannot* wrap because otherwise the allocator would have already aborted. In fact, any kind of wrapping here should cause the process to abort and not continue because it means we've somehow violated the rules of the allocator. But since we calculating from a prior allocation, that's not necessary.

storage/src/node/leaf.rs
```diff
@@ -15,6 +15,14 @@ pub struct LeafNode {
     pub value: Box<[u8]>,
 }
 
+impl lru_mem::HeapSize for LeafNode {
+    fn heap_size(&self) -> usize {
+        self.partial_path
+            .heap_size()
+            .saturating_add(self.value.len())
```
demosdemon: same here for `saturation_add` vs `wrapping_add`. saturating creates unnecessary work and is technically incorrect.

storage/src/node/branch.rs
```diff
@@ -179,6 +190,22 @@ pub struct BranchNode {
...
+        let children_size: usize = self
+            .children
+            .iter()
+            .filter_map(|(_, child)| child.as_ref())
+            .map(lru_mem::HeapSize::heap_size)
+            .sum();
+        self.partial_path
+            .heap_size()
+            .saturating_add(value_size)
+            .saturating_add(children_size)
```
demosdemon: same here for `saturation_add` vs `wrapping_add`. saturating creates unnecessary work and is technically incorrect.

storage/src/node/branch.rs
```diff
@@ -179,6 +190,22 @@ pub struct BranchNode {
...
 
+impl lru_mem::HeapSize for BranchNode {
+    fn heap_size(&self) -> usize {
+        let value_size = self.value.as_ref().map_or(0, |v| v.len());
+        let children_size: usize = self
+            .children
+            .iter()
+            .filter_map(|(_, child)| child.as_ref())
+            .map(lru_mem::HeapSize::heap_size)
+            .sum();
```
demosdemon, line 201: `sum` [doesn't check](https://doc.rust-lang.org/stable/src/core/iter/traits/accum.rs.html#70-78) for overflow (in release mode, it does in debug mode), so having overflow checks after this is not consistent (and wasteful as discussed elsewhere).

storage/src/node/branch.rs
```diff
@@ -68,6 +68,17 @@ pub enum Child {
     MaybePersisted(MaybePersistedNode, HashType),
 }
 
+impl lru_mem::HeapSize for Child {
+    fn heap_size(&self) -> usize {
+        match self {
+            Child::Node(node) => node.heap_size(),
+            Child::AddressWithHash(_, _) => 0,
+            // MaybePersisted contains Arc<Mutex>, we don't count shared data
```
demosdemon, line 76: This will potentially never count any child `SharedNode` that is not separately cached. Not sure if that's important.

storage/src/lib.rs
```diff
@@ -81,6 +81,75 @@ pub use test_utils::TestRecorder;
...
+    /// Update cache memory usage metrics.
+    ///
+    /// This is called after cache operations to keep metrics in sync.
+    /// The `current_size()` and `max_size()` methods are simple field accesses,
+    /// so this is very cheap to call frequently.
+    #[expect(
+        clippy::cast_precision_loss,
+        reason = "Precision loss is acceptable for metrics; only affects values > 2^52 bytes"
+    )]
+    pub fn update_cache_metrics(cache: &lru_mem::LruCache<LinearAddress, Self>) {
```
demosdemon: does this need to be `pub`?

storage/src/lib.rs
```diff
@@ -81,6 +81,75 @@ pub use test_utils::TestRecorder;
 /// A shared node, which is just a triophe Arc of a node
 pub type SharedNode = triomphe::Arc<Node>;
 
+/// A wrapper around `SharedNode` for use in memory-sized caches.
+/// This newtype allows implementing foreign traits while maintaining
+/// compatibility with the rest of the codebase that uses `SharedNode` directly.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CachedNode(pub SharedNode);
```
demosdemon: Does this need to be `pub`?

firewood/src/manager.rs
rkuris: It's tough to tell what release it should be. I'll go for 0.2.0 and we'll target this for next week.

storage/src/node/branch.rs
rkuris: I think since it might be shared among other nodes, we can't count it, but I'm open for ideas.

storage/src/lib.rs
rkuris: Changed to pub(crate)

storage/src/lib.rs
rkuris: pub(crate) is fine here too.

firewood/src/manager.rs
```diff
@@ -42,9 +42,21 @@ pub struct RevisionManagerConfig {
     #[builder(default = 128)]
     max_revisions: usize,
 
-    /// The size of the node cache
-    #[builder(default_code = "NonZero::new(1500000).expect(\"non-zero\")")]
-    node_cache_size: NonZero<usize>,
+    /// The size of the node cache (number of entries).
+    ///
+    /// **Deprecated:** Use `node_cache_memory_limit` instead for memory-based sizing.
+    /// If specified, this value is multiplied by 128 to estimate memory usage.
```
bernard-avalabs, line 48: nit: we don't know if 128 is a good estimate of the amount of memory used by a node. Might be better to reference the constant (BYTES_PER_NODE_ESTIMATE) in the comment.

rkuris:
> 
> ```
> +    describe_gauge!(
> +    describe_gauge!(
> ```
> 
> However, the dashboard was not modified.
> 
> You may need to update `benchmark/Grafana-dashboard.json` accordingly.
> 
> _This check is automated to help maintain the dashboard._

@Elvis339 I've ignored this and I'm considering removing it, but open for suggestions.

storage/src/linear/filebacked.rs
```diff
@@ -27,18 +27,19 @@ use std::os::unix::fs::FileExt;
...
 
 use super::{FileIoError, OffsetReader, ReadableStorage, WritableStorage};
 
 /// A [`ReadableStorage`] and [`WritableStorage`] backed by a file
 #[derive(Debug)]
 pub struct FileBacked {
     filename: PathBuf,
-    cache: Mutex<LruCache<LinearAddress, SharedNode>>,
-    free_list_cache: Mutex<LruCache<LinearAddress, Option<LinearAddress>>>,
+    cache: Mutex<MemLruCache<LinearAddress, CachedNode>>,
```
bernard-avalabs, line 41: Not specific to this PR, but it may be worth measuring the amount of contention we get from this lock on the cache (perhaps by looking at the tail access latency). I've seen this as a performance bottleneck in a number of other systems, and it is relatively easy to fix by sharding and having independent locks per shard.

bernard-avalabs (commented):
I've left a few minor comments, with the main one being to check for lock contention and sharding the cache if there is any.

PR #1680, opened 2026-02-13
chore: add RootStore metrics

RodrigoVillar:
## Why this should be merged

With the time-to-persist now decoupled from the time-to-commit, it would be great if we could capture data on how long it takes to persist + other similar processes.

## How this works

- Adds timer to track how long it takes to persist
- Adds timer to track how long it takes to persisting root addresses to `RootStore`
- Adds counter to track how many times we block on commit

## How this was tested

CI

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -56,7 +56,8 @@ use std::{
 };
 
 use firewood_storage::{
-    Committed, FileBacked, FileIoError, HashedNodeReader, NodeStore, NodeStoreHeader,
+    Committed, FileBacked, FileIoError, HashedNodeReader, LinearAddress, NodeStore,
+    NodeStoreHeader, TrieHash,
 };
 use parking_lot::{Condvar, Mutex, MutexGuard};
 
@@ -396,7 +397,7 @@ impl PersistLoop {
     ) -> Result<(), PersistError> {
         let result = self
             .persist_to_disk(revision)
-            .and_then(|()| self.save_to_root_store(revision));
+            .and_then(|()| self.maybe_save_to_root_store(revision));
         self.release_permits(commits_since_persist);
         result
     }
@@ -426,16 +427,25 @@ impl PersistLoop {
     }
 
     /// Saves the revision's root address to `RootStore` if configured.
-    fn save_to_root_store(&self, revision: &CommittedRevision) -> Result<(), PersistError> {
+    fn maybe_save_to_root_store(&self, revision: &CommittedRevision) -> Result<(), PersistError> {
         if let Some(ref store) = self.shared.root_store
             && let (Some(hash), Some(addr)) = (revision.root_hash(), revision.root_address())
         {
-            store.add_root(&hash, &addr).map_err(|e| {
-                error!("Failed to persist revision address to RootStore: {e}");
-                PersistError::RootStore(e.into())
-            })?;
+            save_to_root_store(store, &hash, &addr)?;
         }
 
         Ok(())
     }
+}
+
+#[crate::metrics("persist.root_store", "persist revision address to root store")]
+fn save_to_root_store(
+    store: &RootStore,
+    hash: &TrieHash,
+    addr: &LinearAddress,
+) -> Result<(), PersistError> {
+    store.add_root(hash, addr).map_err(|e| {
+        error!("Failed to persist revision address to RootStore: {e}");
+        PersistError::RootStore(e.into())
+    })
 }


RodrigoVillar:
Happy to add these metrics to our existing local dashboard if requested, but with @Elvis339 creating a new set of dashboards, it might make more sense to add these new metrics there rather than migrating.

firewood/src/persist_worker.rs
```diff
@@ -266,8 +267,11 @@ impl PersistSemaphore {
     #[inline]
     fn acquire(&self) {
         let mut permits = self.state.lock();
-        while *permits == 0 {
-            self.condvar.wait(&mut permits);
+        if *permits == 0 {
+            firewood_increment!(crate::registry::PERSIST_THROTTLE_BLOCKS, 1);
```
rkuris: Imo we should track how long we waited for. If the pause is 1ms it's probably not as big of a deal

firewood/src/persist_worker.rs
```diff
@@ -444,9 +449,17 @@ impl PersistLoop {
         if let Some(ref store) = self.shared.root_store
             && let (Some(hash), Some(addr)) = (revision.root_hash(), revision.root_address())
         {
-            store
+            let start = coarsetime::Instant::now();
```
rkuris: We have a macro you can just wrap a function call rather than doing the math here, which is a preferred way to do this.

https://ava-labs.github.io/firewood/firewood_macros/attr.metrics.html

firewood/src/registry.rs
```diff
@@ -35,6 +35,21 @@ pub const ACTIVE_REVISIONS: &str = "active_revisions";
...
 pub const MAX_REVISIONS: &str = "max_revisions";
 
+/// Number of times a commit blocked waiting for persistence to catch up.
+pub const PERSIST_THROTTLE_BLOCKS: &str = "persist.throttle_blocks";
+
+/// Number of persist operations.
+pub const PERSIST_TOTAL: &str = "persist.total";
+
+/// Persist latency in milliseconds.
+pub const PERSIST_TOTAL_MS: &str = "persist.total_ms";
```
Elvis339: Seems like this is a counter (cumulative), but for latency analysis we would want a histogram bucket. Without a histogram we can only compute average latency i.e., `rate(ersist.total_ms)  / rate(persist.tota)`, not p50/p99.

rkuris: I agree. Histograms should be behind the expensive metrics flag and we have support for them in our macros.

We should not be trying to do metrics differently and the macro was an attempt to address that. The macros could use more design work.

firewood/src/persist_worker.rs
Elvis339: We already have duration: when a commit blocks on `semaphore.acquire()`, that wait time is included in `ffi.commit_ms`. (correct me if I'm wrong here)

I'd replace `PERSIST_THROTTLE_BLOCKS` with a permits gauge:
- Gauge = 0 answers "is backpressure happening now?"
- Correlate with ffi.commit_ms p95/p99 to see "how bad is it?"

A counter catches both sustained backpressure and brief transient blocks that self-resolve. We probably only care about sustained - if backpressure persists long enough to matter, a gauge will eventually catch it. Count without duration isn't actionable signal.

firewood/src/persist_worker.rs
```diff
@@ -444,9 +449,17 @@ impl PersistLoop {
...
+
+            let result = store
                 .add_root(&hash, &addr)
-                .map_err(|e| PersistError::RootStore(e.into()))?;
+                .map_err(|e| PersistError::RootStore(e.into()));
+
+            let success = if result.is_err() { "false" } else { "true" };
+            firewood_increment!(crate::registry::PERSIST_ROOT_STORE, 1, "success" => success);
+            firewood_increment!(crate::registry::PERSIST_ROOT_STORE_MS, start.elapsed().as_millis(), "success" => success);
```
Elvis339: Is root store latency a known concern, or is this speculative instrumentation?

firewood/src/registry.rs
Elvis339: Re: persist.total / persist.total_ms

This function's latency is essentially `flush_nodes` + negligible overhead.

`flush_nodes` already captures the disk I/O time, which is the dominant cost here. `persist.total_ms` would show identical values with noise.

For count, we can derive from FFI counters: `rate(ffi.commit) + rate(ffi.batch) + rate(ffi.merge)` = persists/sec. (correct me if I'm wrong).

firewood/src/persist_worker.rs
RodrigoVillar: `RootStore` latency _could_ be a concern, so adding a tracker which measures the latency is an attempt to figure out if this concern is valid or not.

firewood/src/persist_worker.rs
RodrigoVillar: I looked at the macro; however, attaching the macro to this would mean that, if running a non-archival node, that this metric would still be collected as `RootStore` would be `None`. Manually writing out the metric code here comes with the benefit of not tracking this metric when running a non-archival node.

RodrigoVillar: I'm open to still using the macro, but wanted to make this side-effect known.

firewood/src/persist_worker.rs
RodrigoVillar: Went ahead and removed the counter for how many times we block.

firewood/src/registry.rs
RodrigoVillar: > For count, we can derive from FFI counters: rate(ffi.commit) + rate(ffi.batch) + rate(ffi.merge) = persists/sec. (correct me if I'm wrong).

Why would we need the batch and merge numbers? Alternatively, this might suggest that we should add a counter to match `FLUSH_NODES` as well.

firewood/src/persist_worker.rs
rkuris: You can get around that with some code restructuring.

Rename`save_to_root_store` -> `maybe_save_to_root_store` or `save_to_root_store_if_enabled`. Have that function do all the checks, then call another one called `save_to_root_store` with the appropriate annotations. Something like this:

```
    #[crate::metrics("persist.root_store", "persist revision address to root store")]
    fn save_to_root_store(
        store: &RootStore,
        hash: &TrieHash,
        addr: &LinearAddress,
    ) -> Result<(), PersistError> {
        store.add_root(hash, addr).map_err(|e| {
            error!("Failed to persist revision address to RootStore: {e}");
            PersistError::RootStore(e.into())
        })
    }
```

If the function is never called because the root store is disabled, the metric will never be emitted.

rkuris (changes requested):

firewood/src/persist_worker.rs
RodrigoVillar: Went with this approach.

Elvis339 (commented):
How does persist.total_ms differ from flush nodes? Looks like we're measuring the same disk I/O at two layers.

We already have a lot in the registry, so I want to make sure new metrics either give us something we can't derive, or enable a clear action. If the goal is catching backpressure, a permits gauge might be more direct - when permits hit zero, commits block.

What action are these meant to enable?

RodrigoVillar:
@Elvis339 I think we're talking about two things:

- `flush_nodes()` captures most of the time in `persist()` (I agree)
- `flush_nodes()` only tracks time, not count (we can add this in a separate PR)

Going to remove the `persist()` metric and going to follow up with adding a counter for `flush_nodes()`.

Elvis339:
> * `flush_nodes()` only tracks time, not count (we can add this in a separate PR)

Pushing back on adding count for flush_nodes. We're going to instrument disk I/O latency, a histogram makes more sense than count + ms.

flush_nodes is the fundamental disk write path. The average would still look fine while the tail is causing problems.
Count + ms only gives us average. A histogram would let us see p95/p99 where the actual pain hides.

PR #1681, opened 2026-02-13
feat(ffi): Remove explicit garbage collection call

demosdemon:
This removes the explicit `go runtime.GC()` in the database close method. It could be a red-herring from #1679 but it is probably not great that we do this anyways.

## Caveat Emptor

https://github.com/ava-labs/avalanchego/blob/a69e790ea9b6524117f3cae884e23def5084cfda/graft/evm/firewood/triedb.go#L274-L277

This 5 minute delay is likely to be hit and be exceeded. There is no guarantee that the finalizer on any object will be ran within that 5 minute window. In my experience, unless there is pressure on the GC it is unlikely to seek out finalizers until needed. Meaning, this can theoretically **never** succeed until every outstanding handle is explicitly closed or a `GC()` call is made. However, it appears *firewood* should not be the ones performing that GC.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,4 +16,16 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
-    open-pull-requests-limit: 0 # Disable non-security version updates
+    labels:
+      - "dependencies"
+      - "github-actions"
+      - "security"
+    commit-message:
+      prefix: "ci"
+      include: "scope"
+    # Group all GitHub Actions updates into a single PR to reduce noise
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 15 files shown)


demosdemon:
A thought post merge, the affect on AvalancheGo should theoretically be small. AvalancheGo is already periodically invoking `runtime.GC`. Both [coreth](https://github.com/ava-labs/avalanchego/blob/4473fe964c6428727adb35964ea6e365cfd3fea6/graft/coreth/plugin/evm/vm.go#L1088) and [subnet-evm](https://github.com/ava-labs/avalanchego/blob/4473fe964c6428727adb35964ea6e365cfd3fea6/graft/subnet-evm/plugin/evm/vm.go#L1248) spawn this continuous profiler with a default timeout of 15 minutes. The profiler triggers runtime.GC every time it takes a memory snapshot.

PR #1685, opened 2026-02-16
fix: Avoid holding the last_committed_revision arc

### LINKED ISSUE #1684: `TestRevisionOutlivesReaping` test fails in 3p validation
Since the run at 12:30am eastern on 2/14, we've been occasionally seeing `TestRevisionOutlivesReaping` fail. The only thing that looks like it merged on the 13th is https://github.com/ava-labs/firewood/pull/1657

```
=== RUN   TestRevisionOutlivesReaping
    firewood_test.go:1095: 
        	Error Trace:	/firewood/ffi/firewood_test.go:1095
        	Error:      	An error is expected but got nil.
        	Test:       	TestRevisionOutlivesReaping
--- FAIL: TestRevisionOutlivesReaping (0.01s)
```

Error Line: https://github.com/ava-labs/firewood/blob/7d4d171c8710c361faadc656836fa5c8ec32ba3a/ffi/firewood_test.go#L1094-L1095

---

Note that I have not see this error in our own CI but it was reported by third-party validation.

rkuris:
To reliably reproduce the original timing problem, sleep for 200ms or so during the call to `self.persist`.

## Why this should be merged

It's a bug.

## How this works

When persist() is slow, the persist worker holds the Arc reference through last_committed_revision for an extended period after persistence completes. This prevents the revision manager from reaping the revision because Arc::try_unwrap() fails (reference count > 1).

So, we clear the last_committed_revision the moment we've decided to persist it, which lowers the reference count faster.

## How this was tested

Added that sleep and reproduced the bug. Added the fix and left in the sleep and it worked great.

Fixes #1684

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -448,8 +448,6 @@ func (db *Database) Close(ctx context.Context) error {
 		return nil
 	}
 
-	go runtime.GC()
-
 	done := make(chan struct{})
 	go func() {
 		db.outstandingHandles.Wait()
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -156,7 +156,16 @@ func newTestDatabase(tb testing.TB, opts ...Option) *Database {
 	db, err := newDatabase(tb.TempDir(), opts...)
 	r.NoError(err)
 	tb.Cleanup(func() {
-		r.NoError(db.Close(oneSecCtx(tb)))
+		err := db.Close(oneSecCtx(tb))
+		if errors.Is(err, ErrActiveKeepAliveHandles) {
+			// force a GC to clean up dangling handles that are preventing the
+			// database from closing, then try again. Intentionally not looping
+			// since a subsequent attempt is unlikely to succeed if the first
+			// one didn't.
+			runtime.GC()
+			err = db.Close(oneSecCtx(tb))
+		}
+		r.NoError(err)
 	})
 	return db
 }
--- a/ffi/tests/eth/eth_compatibility_test.go
+++ b/ffi/tests/eth/eth_compatibility_test.go
@@ -6,10 +6,14 @@ package eth
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"math/rand"
+	"runtime"
 	"slices"
 	"testing"
+	"time"
 
+	"github.com/ava-labs/firewood-go-ethhash/ffi"
 	firewood "github.com/ava-labs/firewood-go-ethhash/ffi"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core/rawdb"
@@ -62,13 +66,33 @@ type merkleTriePair struct {
 	pendingFwdBatch  []firewood.BatchOp
 }
 
+// oneSecCtx returns `tb.Context()` with a 1-second timeout added. Any existing
+// cancellation on `tb.Context()` is removed, which allows this function to be
+// used inside a `tb.Cleanup()`
+func oneSecCtx(tb testing.TB) context.Context {
+	ctx := context.WithoutCancel(tb.Context())
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
+	tb.Cleanup(cancel)
+	return ctx
+}
+
 func newFirewoodDB(t *testing.T) *firewood.Database {
 	t.Helper()
+	r := require.New(t)
 
 	db, err := firewood.New(t.TempDir(), firewood.EthereumNodeHashing)
-	require.NoError(t, err, "firewood.New()")
+	r.NoError(err, "firewood.New()")
 	t.Cleanup(func() {
-		require.NoErrorf(t, db.Close(context.Background()), "%T.Close()", db)
+		err := db.Close(oneSecCtx(t))
+		if errors.Is(err, ffi.ErrActiveKeepAliveHandles) {
+			// force a GC to clean up dangling handles that are preventing the
+			// database from closing, then try again. Intentionally not looping
+			// since a subsequent attempt is unlikely to succeed if the first
+			// one didn't.
+			runtime.GC()
+			err = db.Close(oneSecCtx(t))
+		}
+		r.NoError(err, "%T.Close()", db)
 	})
 	return db
 }
--- a/ffi/tests/firewood/merkle_compatibility_test.go
+++ b/ffi/tests/firewood/merkle_compatibility_test.go
@@ -5,12 +5,16 @@ package firewood
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"testing"
+	"time"
 
+	"github.com/ava-labs/firewood-go/ffi"
 	firewood "github.com/ava-labs/firewood-go/ffi"
 
 	"github.com/ava-labs/avalanchego/database"
@@ -50,14 +54,34 @@ var stepMap = map[byte]string{
 	commitProposal:           "commitProposal",
 }
 
+// oneSecCtx returns `tb.Context()` with a 1-second timeout added. Any existing
+// cancellation on `tb.Context()` is removed, which allows this function to be
+// used inside a `tb.Cleanup()`
+func oneSecCtx(tb testing.TB) context.Context {
+	ctx := context.WithoutCancel(tb.Context())
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
+	tb.Cleanup(cancel)
+	return ctx
+}
+
 func newTestFirewoodDatabase(t *testing.T) *firewood.Database {
 	t.Helper()
+	r := require.New(t)
 
 	dbFile := filepath.Join(t.TempDir(), "test.db")
 	db, err := newFirewoodDatabase(dbFile)
-	require.NoError(t, err)
+	r.NoError(err, "firewood.New()")
 	t.Cleanup(func() {
-		require.NoError(t, db.Close(context.Background())) //nolint:usetesting // t.Context() will already be cancelled
+		err := db.Close(oneSecCtx(t))
+		if errors.Is(err, ffi.ErrActiveKeepAliveHandles) {
+			// force a GC to clean up dangling handles that are preventing the
+			// database from closing, then try again. Intentionally not looping
+			// since a subsequent attempt is unlikely to succeed if the first
+			// one didn't.
+			runtime.GC()
+			err = db.Close(oneSecCtx(t))
+		}
+		r.NoError(err, "%T.Close()", db)
 	})
 	return db
 }


rkuris:
@RodrigoVillar the risk here is that now the revision being persisted can vanish from the revision manager before the persist operation. Please confirm that this is okay before approving this PR.

rkuris:
> @RodrigoVillar the risk here is that now the revision being persisted can vanish from the revision manager before the persist operation. Please confirm that this is okay before approving this PR.

It's not 100% clear to me why we need to keep that reference at all.

RodrigoVillar:
> > @RodrigoVillar the risk here is that now the revision being persisted can vanish from the revision manager before the persist operation. Please confirm that this is okay before approving this PR.
> 
> It's not 100% clear to me why we need to keep that reference at all.

This is so that we persist the last committed revision in-memory. The flow for when this occurs is as follows:

```
- RevisionManager::close() is called
- This calls PersistWorker::close()
- Sender channel is closed, triggering the background thread to exit from it's `while` loop
- If the latest committed revision is not persisted, persist it
```

RodrigoVillar:
> @RodrigoVillar the risk here is that now the revision being persisted can vanish from the revision manager before the persist operation. Please confirm that this is okay before approving this PR.

Revisions vanish from the revision manager only on reaps. If a revision vanishes from the revision manager, this implies that the revision has already been persisted/has been ignored by the background loop. I'm not sure why this matters in the context of this PR?

rkuris:
> Revisions vanish from the revision manager only on reaps. If a revision vanishes from the revision manager, this implies that the revision has already been persisted/has been ignored by the background loop. I'm not sure why this matters in the context of this PR?

Ah and since the backgorund thread is the only one who can do the reaping, I think it's safe to remove the reference to the last persisted revision as done in this fix.

Why can't we, on shutdown, just get the most recent revision from the revision manager rather than trying to track the most recent revision in the worker thread?

firewood/src/persist_worker.rs
```diff
@@ -366,10 +366,14 @@ impl PersistLoop {
                         .get()
                         .is_multiple_of(self.persist_interval.get())
                     {
+                        // Clear last_committed_revision before persisting to avoid holding
+                        // an Arc reference during the persist operation.
+                        self.last_committed_revision = None;
                         self.persist(&revision, num_commits)?;
+                    } else {
+                        // Store the revision so we can persist it on shutdown if needed.
+                        self.last_committed_revision = Some((revision, num_commits));
```
RodrigoVillar: With your changes, `Some` implies the revision wasn't persisted while `None` implies the revision was persisted (assuming `self.persist()` returns `Ok`/the thread didn't crash). Therefore, there's no need to store `num_commits` within `last_committed_revision` (furthermore, we should probably rename `last_committed_revision` to `last_unpersisted_revision`).

rkuris:
This could use another review. I made some significant changes that I think improve readability and seemed to me to be more readable. I think num_commits is not a NonZero, you could have zero commits, and that's what it should start at, instead of starting it at 1 and post-incrementing.

I completely removed having a separate variable tracking the commit number of the most recent commit.

See the commit comments for more details.

Again, a full review and some manual stepping through would be helpful here.

firewood/src/persist_worker.rs
```diff
@@ -356,29 +356,30 @@ impl PersistLoop {
...
 
         while let Ok(message) = self.receiver.recv() {
             match message {
                 PersistMessage::Reap(nodestore) => self.reap(nodestore)?,
                 PersistMessage::Persist(revision) => {
-                    if num_commits
-                        .get()
-                        .is_multiple_of(self.persist_interval.get())
-                    {
+                    // wrapping_add is safe: even at 1ns per commit, u64 overflow would take ~584 years
```
demosdemon: why not reset to zero on persist so that overflow is never possible. Resetting to zero means `is_multiple_of` can be replaced with `>=`.

firewood/src/persist_worker.rs
```diff
@@ -420,12 +421,9 @@ impl PersistLoop {
...
     /// Releases semaphore permits for commits up to `num_commits`.
-    fn release_permits(&mut self, num_commits: NonZeroU64) {
-        let last = self.last_persisted_commit.map_or(0, NonZeroU64::get);
-        let permits_to_release = num_commits
-            .get()
-            .checked_sub(last)
-            .expect("should be positive");
+    fn release_permits(&mut self, num_commits: u64) {
+        let last = self.last_persisted_commit.unwrap_or(0);
+        let permits_to_release = num_commits.checked_sub(last).expect("should be positive");
```
demosdemon: I'm not sure when `count` would ever not be equal to `persist_interval` (except when shutting down). I don't see the point of storing `last_persisted_commit` and doing this subtraction if `num_commits` were reset to zero after every persist. In which case, `num_commits` should be always less than or equal to `persist_interval` when releasing permits.

rkuris: Agreed, this is even simpler

rkuris:
Attempt 3 to simplify the worker thread, this time focusing on separation of concerns.

firewood/src/persist_worker.rs
```diff
@@ -348,88 +348,63 @@ impl PersistLoop {
...
-    fn release_permits(&mut self, num_commits: NonZeroU64) {
-        let last = self.last_persisted_commit.map_or(0, NonZeroU64::get);
-        let permits_to_release = num_commits
-            .get()
-            .checked_sub(last)
-            .expect("should be positive");
-        if let Some(count) = NonZeroU64::new(permits_to_release) {
+    /// Releases semaphore permits for commits since last persist if any
+    fn release_permits(&mut self) {
+        if let Some(count) = NonZeroU64::new(self.commits_since_persist) {
```
RodrigoVillar: `count` will always be non-zero as:

- We first increment this value in `persist()`
- We then call `do_persist() => release_permits()` (only logic flow that calls `release_permits()`)

Would it make sense to use an `expect()` here?

rkuris: Good observation. Yes, you're correct. It's currently impossible for commits_since_persist to be 0 when we call this function, since the current set of code paths never call it when it's zero.

I think I like this behavior better than crashing though. If it happens that someone in the future changes the code so that release_permits is called when there aren't any to release, I'd find it okay that it doesn't do anything rather than panicking. The comment also says "if any"; if we did change it to expect we'd have to add some panic docs, and that (marginally) adds to cognitive overhead while reading.

RodrigoVillar (approved):

firewood/src/persist_worker.rs
```diff
@@ -279,8 +278,9 @@ impl PersistSemaphore {
     #[inline]
     fn release(&self, count: NonZeroU64) {
         let mut permits = self.state.lock();
+        // wrapping_add is safe here: even at 1ns per commit, u64 overflow would take ~584 years
         *permits = permits
-            .saturating_add(count.get())
+            .wrapping_add(count.get())
```
bernard-avalabs, line 282: Given that the modified semaphore is initialized with permits that can be acquired and later released but no new permits can be created after initialization, then having more permits than the initial amount should return an error (indicating a logic error in using the semaphore).

firewood/src/persist_worker.rs
bernard-avalabs, line 244: nit: Not directly related to this PR, but the relationship between commits, persists, and permits is not very clear. The purpose of the semaphore is the control the rate of commits. Does a permit allow one to commit, and a persist of a batch of commits will then release multiple permits?

firewood/src/persist_worker.rs
```diff
@@ -348,88 +348,63 @@ impl PersistLoop {
...
-
-            let err = PersistError::FileIo(Arc::new(e));
-            // Release permits even on error to unblock waiting threads
-            self.release_permits(num_commits);
-
-            return Err(err);
+    /// Handles a persist message: increments counter, decides whether to persist now or defer.
+    fn persist(&mut self, revision: CommittedRevision) -> Result<(), PersistError> {
+        // wrapping_add is safe: we will never exceed persist_interval
+        self.commits_since_persist = self.commits_since_persist.wrapping_add(1);
```
bernard-avalabs: nit: It seems odd that we increment the commits counter when we receive a persist message.

firewood/src/persist_worker.rs
```diff
@@ -348,88 +348,63 @@ impl PersistLoop {
...
+
+        if self.commits_since_persist >= self.persist_interval.get() {
+            // Clear persist_on_shutdown before persisting to avoid holding
+            // an Arc reference during the persist operation.
+            self.persist_on_shutdown = None;
+            self.do_persist(&revision)
```
bernard-avalabs: May be helpful to add a comment indicating that `release_permits` called by `do_persist` resets `commits_since_persist`.

firewood/src/persist_worker.rs
```diff
@@ -348,88 +348,63 @@ impl PersistLoop {
...
-            // Release permits even on error to unblock waiting threads
-            self.release_permits(num_commits);
-
-            return Err(err);
+    /// Handles a persist message: increments counter, decides whether to persist now or defer.
+    fn persist(&mut self, revision: CommittedRevision) -> Result<(), PersistError> {
+        // wrapping_add is safe: we will never exceed persist_interval
+        self.commits_since_persist = self.commits_since_persist.wrapping_add(1);
+
+        if self.commits_since_persist >= self.persist_interval.get() {
```
bernard-avalabs: Design may be a bit error prone since you have two related counters: `commits_since_persist` and the counter inside the semaphore. It is possible to deadlock by accident if these counters are not set correctly. Specifically, if the semaphore value hits zero before `commits_since_persist` >= persist interval.

We can avoid this if we modify the semaphore so we can inspect the semaphore value, and call do_persist when the semaphore value is zero. 

I'm haven't looked at how reaping is handled yet, but without reaping, the **textbook approach** for solving this synchronization problem is to do most of the work on the sender side using one lock, two condition variables (something like `commits_not_full` and `persist_ready`), some associated counters/flags and no channels. 

The main event loop would block on `persist_ready`, and the sender would acquire the lock, decrement commit counter, and store the revision in the the struct associated with the lock. If a sender performs the 1 to 0 transition for the counter, it would signal `persist_ready` to wake up the event loop. Additional senders would block on the `commits_not_full` condition. Upon waking up, the event thread would persist the revision, set the counter back to its initial value and call signal all on `commits_not_full` to allow the senders to continue. 

Again, I haven't looked at how reaping works, so maybe the textbook approach would need to be modified for reaping.

firewood/src/persist_worker.rs
rkuris: Yeah this is clearer if you read the design docs, and probably didn't make it into the code as well as it could have. The idea behind the permits is to allow only N commits before we stop allowing commits, but we start persisting at the halfway mark instead of when we are completely out to avoid blocking the thread when we're near full load.

firewood/src/persist_worker.rs
rkuris: Good suggestion. This method shouldn't be named 'persist' because it doesn't always persist. `maybe_persist`or just `commit` is better. I'll change it.

demosdemon: taking this out of the `event_loop` body makes this error prone. I would rather the mutable counter be defined within the event_loop and only be mutable within the event_loop.

firewood/src/persist_worker.rs
rkuris: I'm not following this. The code supports: take a permit, and release N permits. These are the only methods on this semaphore. "new permits can be created" doesn't make too much sense here.

You can think of the semaphore as a ticketing queue for a set of dining room tables. We support having a new guest and then dismissing groups of guests at once. I don't know what "new permits can be created" means in terms of this scenario.

I generally prefer not returning any errors and making it impossible to use something in a way that isn't supported, and I think we're doing that here. Either that or I'm not following.

firewood/src/persist_worker.rs
rkuris: The comment on `do_persist` does say that, but the method has side effects that isn't named. I think this can be fixed by renaming the method `do_persist` to `persist_and_release`, which is another good naming suggestion.

firewood/src/persist_worker.rs
rkuris: The approach taken here was a little different, since we prefer never to block unless we can't catch up. What was done here was each time a commit is wanted, the main code gets a permit from the semaphore, then sends a message to the worker thread. The worker then decides how often to persist. separating the concerns between the committer and the persister.

The semaphore is solely used for backpressure in case the worker can't catch up, which we expect to happen during bootstrapping for example.

firewood/src/persist_worker.rs
rkuris: Done. Nicer to not pass around a `&mut PersistLoop`

firewood/src/persist_worker.rs
```diff
@@ -348,88 +345,75 @@ impl PersistLoop {
...
+        *commits_since_persist = commits_since_persist.wrapping_add(1);
+
+        if *commits_since_persist >= self.persist_interval.get() {
+            // Clear persist_on_shutdown before persisting to avoid holding
+            // an Arc reference during the persist operation.
+            self.persist_on_shutdown = None;
+            self.persist_and_release(&revision, commits_since_persist)
+        } else {
+            // Store the revision so we can persist it on shutdown if needed.
+            self.persist_on_shutdown = Some(revision);
```
demosdemon, line 384: As an observation, the event loop could reset the counter if `persist_on_shutdown` is `None`. Alternatively, you could use an enum result to signal that to the event loop:

```rust
enum CommitResult {
    Queued,
    FlushedOk,
    FlushedErr(PersistError),
}
```

That way you also don't need to thread the `&mut u64` (instead, just a `u64`). Passing the mutable reference through the function is not exactly "only mutable within the event_loop".

firewood/src/persist_worker.rs
```diff
@@ -348,88 +345,75 @@ impl PersistLoop {
...
+    /// Performs the actual persistence and releases semaphore permits.
+    fn persist_and_release(
+        &mut self,
+        revision: &CommittedRevision,
+        commits_since_persist: &mut u64,
+    ) -> Result<(), PersistError> {
+        let result = self
+            .persist_to_disk(revision)
+            .and_then(|()| self.save_to_root_store(revision));
+        self.release_permits(commits_since_persist);
```
demosdemon, line 398: releasing the permits even if persisting errors is irrelevant. the `PersistSemaphore` needs a state to indicate aborted so that no new permits are ever acquired and no workers ever block again and that state should be triggered followed by a `notify_all` whenever the event loop exits.

demosdemon: Because there's a number of panic points in the code, the aborted signal also should be set by a drop guard ensuring that the flag is set even on unwind so that no threads deadlock due to a panic in the worker thread.

The semaphore is acquired before checking if the worker channel is broken, so the semaphore can cause a deadlock if the worker exits, even without panicking.

firewood/src/persist_worker.rs
bernard-avalabs: But the worker is still blocking on the channel. I'm not sure I understand the difference between blocking on a channel versus blocking on a condition variable.

In an effort of separating concerns that are coupled, the code introduces an implicit dependency between the counter in the semaphore and a secondary counter that can lead to a deadlock if there are any "math" errors. Instead of trying to catch this through comprehensive code review, it would be better to use a synchronization approach where there is a single counter (if possible).

firewood/src/persist_worker.rs
bernard-avalabs: In a normal semaphore, there is a corresponding V() (release) for each P() (acquire). In this modified approach, you pass a counter to release. Perhaps I'm missing something, but it seems the caller can arbitrary call release(LARGE_NUMBER) that would break the invariant of your abstraction. This is what I was calling out. Calling release with a number larger than the initial value would effectively create new permits. This code simply calls min on it and ignores a logic error from the caller.

firewood/src/persist_worker.rs
rkuris: Sure, but this PR doesn't change that part of the design. I think the design changes would be under a different PR. The question to ask here is if this is better than what's on main right now, which still has that same issue.

We can open a followup ticket and investigate alternative designs for this if you'd like, but I'm not convinced there is a much simpler approach yet, even after a side conversation.

firewood/src/persist_worker.rs
rkuris: I think this is a pre-existing bug. I'll create another issue for this. I didn't change this behavior.

firewood/src/persist_worker.rs
rkuris: Created #1694 for followups here.

firewood/src/persist_worker.rs
rkuris: Created #1693

firewood/src/persist_worker.rs
bernard-avalabs: Perhaps semaphores can be removed altogether in a followup PR.

PR #1690, opened 2026-02-17
refactor: stop keeping around the last committed revision

RodrigoVillar:
## Why this should be merged

Builds on the readability improvements of #1685 by no longer having the `PersistLoop` track what the last committed revision was. By never tracking what the latest committed revision, the `PersistLoop` can no longer be the source of contention for trying to gain ownership of a revision during reaping.

## How this works

Moves the `persist_on_shutdown` field from the `PersistLoop` to `SharedState`. This field is populated on `close()` and is read by the `PersistLoop` when the channel closes. 

Here are some alternative designs I considered:
- Having the `PersistWorker` persist the last committed revision - this would involve the worker having a reference to the `header`. I would prefer for the `PersisWorker` to be the only thing persisting.
- Adding a `Shutdown` message type - from reading online, dropping the channel is strongly preferred (and more idiomatic) versus having a shutdown signal via a `Shutdown` message. 

## How this was tested

CI

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -119,13 +119,13 @@ impl PersistWorker {
             commit_throttle: PersistSemaphore::new(commit_count),
             root_store,
             header: Mutex::new(header),
+            persist_on_shutdown: OnceLock::new(),
         });
 
         let persist_loop = PersistLoop {
             receiver,
             persist_interval,
             shared: shared.clone(),
-            persist_on_shutdown: None,
         };
 
         let handle = thread::spawn(move || persist_loop.run());
@@ -182,11 +182,16 @@ impl PersistWorker {
         }
     }
 
-    /// Close the persist worker.
-    ///
-    /// This method is idempotent for compatibility with `drop` (as `drop` is
-    /// called after this function).
-    pub(crate) fn close(mut self) -> Result<(), PersistError> {
+    /// Close the persist worker and persist `latest_committed_revision`.
+    pub(crate) fn close(
+        mut self,
+        latest_committed_revision: CommittedRevision,
+    ) -> Result<(), PersistError> {
+        self.shared
+            .persist_on_shutdown
+            .set(latest_committed_revision)
+            .expect("should be empty");
+
         // Drop the sender to close the channel, signaling the background
         // thread to exit.
         drop(self.sender.take());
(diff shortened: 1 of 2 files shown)


firewood/src/persist_worker.rs
```diff
@@ -186,7 +186,16 @@ impl PersistWorker {
...
-    pub(crate) fn close(&mut self) -> Result<(), PersistError> {
+    pub(crate) fn close(
+        &mut self,
+        latest_committed_revision: CommittedRevision,
+    ) -> Result<(), PersistError> {
+        // Ignore the error if already set.
+        let _ = self
+            .shared
+            .persist_on_shutdown
+            .set(latest_committed_revision);
```
RodrigoVillar: Using an `expect()` here fails because `drop()` is called after `close()` (which calls `close()` again). I still went with using a `OnceLock` here as it makes the intent clear that this value should be set once.

rkuris: There is some code smell here if it's possible to drop stuff twice. There are a few choices, many of them aren't all that great, but I think the easiest answer here is:

 - Database close tells the revision manager to send a shutdown message to the worker. That message can contain a reply one-time channel with the final error status of the completed persist operation, but it leaves the worker thread alive. If it gets a "channel closed" error, then you can assume the thread aborted and wait for the thread to exit, collecting the error.
 - Drop shouldn't bother sending the shutdown message. All our use cases we will call close() explicitly. Otherwise, the most recent revision isn't guaranteed to be on disk and/or we'll miss passing I/O errors back to the caller.
 
 I think this simplifies things and never leaves the RevisionManager in a state where it couldn't continue to operate.

RodrigoVillar: Good point; with #1701 merged in, this code smell has been removed altogether as `close()` now consumes itself.

firewood/src/persist_worker.rs
```diff
@@ -357,8 +362,10 @@ impl PersistLoop {
             }
         }
 
-        // Persist any unpersisted revision on shutdown
-        if let Some(revision) = self.persist_on_shutdown.take() {
+        // Persist the last unpersisted revision on shutdown
+        if let Some(revision) = self.shared.persist_on_shutdown.get().cloned()
```
bernard-avalabs, line 366: nit: can remove the `cloned` if `persist_on_shutdown` is a `Mutex<Option<T>>` instead of a `OnceLock<T>` (with some other small changes). Not sure if a `lock` here is cheaper than a `get` followed by a `cloned`. However, the system is shutting down at this point so performance doesn't actually matter.

rkuris (approved):

firewood/src/persist_worker.rs
RodrigoVillar: Yeah on the note of performance, the `cloned()` call is fine since we call it only once. Using a `OnceLock` here makes the intent of `persist_on_shutdown` easier to understand IMO rather than a `Mutex<Option<T>>`.

PR #1691, opened 2026-02-17
chore: github actions security audit

rkuris:
## Why this should be merged

Pin everything. Add dependabot config so we get updates too.

## How this works

There were some older versions with some minor security problems also fixed here.

## How this was tested

ci only


.github/workflows/metrics-check.yaml
```diff
@@ -55,15 +55,15 @@ jobs:
       - name: Look for previous comment
         if: github.repository == github.event.pull_request.head.repo.full_name && steps.check-metrics.outputs.metric_changes_found == 'true'
         id: find-comment
-        uses: peter-evans/find-comment@v3
+        uses: peter-evans/find-comment@d2dae40ed151c634f4189e0c8a6bb31de6e07b0e # v3.2.0
```
demosdemon: where did these hashes come from? There is no 3.2.0 and v3.1.0 is 3eae4d37986fb5a8592848f6a574fdf654e61f9e

.github/workflows/attach-static-libs.yaml
```diff
@@ -50,9 +50,9 @@ jobs:
     outputs:
       has_secrets: ${{ steps.check_secrets.outputs.has_secrets }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: dtolnay/rust-toolchain@7bb1479df4dc3e19ca9d98e09e189ab48a6ec762 # stable
```
demosdemon: the current `stable` is 631a55b12751854ce901bb631d5902ceb48146f7

.github/workflows/metrics-check.yaml
rkuris: Yeah sorry this was supposed to be a draft PR :/

rkuris: Okay I had AI write a script to verify that (a) all the hashes are reachable and (b) the comment matches the hash. That script is checked in. We may want to consider adding some additional functionality to it so it can maybe do upgrades for us, or at least verify after upgrading..

.github/workflows/attach-static-libs.yaml
```diff
@@ -51,8 +51,8 @@ jobs:
       has_secrets: ${{ steps.check_secrets.outputs.has_secrets }}
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
+      - uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # @stable
```
maru-ava, line 54: Am I missing something, or is `stable` a branch rather than a tag? I recommend either not pinning or pinning to a specific version (e.g. `v1.85.0`) rather than to a branch.

.github/workflows/attach-static-libs.yaml
```diff
@@ -51,8 +51,8 @@ jobs:
...
+      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # @v2 -> v2.8.2
```
maru-ava, line 55: (No action required) Not sure of the point of `@v2` given that it can be trivially inferred. Suggest just using the explicit version for consistency with avalanchego.

.github/workflows/track-performance.yml
```diff
@@ -124,7 +124,7 @@ jobs:
       # workflow via API and passes FIREWOOD_REF to it. AvalancheGo is responsible
       # for checking out and building Firewood at that ref.
       - name: Checkout Firewood
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
maru-ava: Can skip pinning for github-provided actions.

.github/dependabot.yml
```diff
@@ -16,4 +16,16 @@ updates:
...
+      - "security"
+    commit-message:
+      prefix: "ci"
+      include: "scope"
+    # Group all GitHub Actions updates into a single PR to reduce noise
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    open-pull-requests-limit: 10 # Enable GitHub Actions updates for security
```
maru-ava, line 31: (No action required) I'm not sure this is desirable, dependabot can be pretty noisy and action updates are notoriously difficult to review. avalanchego limits dependabot to security updates as a result.

scripts/verify-action-hashes.py
```diff
@@ -0,0 +1,268 @@
+#!/usr/bin/env python3
```
maru-ava: (No action required) I'm not sure I'd recommend the addition of a script like this, let alone one adding a dependency on a python runtime. There are established tools like https://github.com/mheap/pin-github-action that could be used without imposing a maintenance burden. Once this PR lands, dependabot maintains existing references, new references take 30s to correlate the desired version, and CI will fail if any references are invalid.

.github/workflows/attach-static-libs.yaml
rkuris: We are pinning to a sha now. The "# @stable" is only a comment, and that's what used to be there. I do like the comment because it shows what we intended to be there and could be used to prove that the sha of that branch suddenly changed.

.github/workflows/attach-static-libs.yaml
rkuris: We want v2, but the current version of v2 is 2.8.2. The comment is there mostly to show that we started with @v2 but this really means v2.8.2.

.github/dependabot.yml
rkuris: Let's try it. If it's too noisy we can always turn it off

scripts/verify-action-hashes.py
rkuris: Agreed, we should nuke it. I'll remove it before merging.

.github/workflows/track-performance.yml
rkuris: Fixed.

scripts/verify-action-hashes.py
rkuris: Fixed.

.github/workflows/metrics-check.yaml
demosdemon: For next time, I recommend letting CI pass before publishing the PR if CI is the only testing performed. It is not respectful of our time to publish a PR you didn't test or even write.

rkuris: As mentioned, this was supposed to be a draft PR. Totally agree with the sentiment. Apologies again.

PR #1697, opened 2026-02-19
chore: condvar WIP

rkuris:

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -12,9 +12,9 @@
 //! the number of unpersisted commits.
 //!
 //! The diagram below shows how commits are handled under deferred persistence.
-//! The main thread validates and updates in-memory state, then hands off the
-//! committed revision to the background thread for disk I/O. A semaphore provides
-//! backpressure when the background thread falls behind.
+//! The main thread updates shared state and signals the background thread via
+//! condition variables. Backpressure is enforced by waiting when all permits
+//! are exhausted.
 //!
 //! Below is an example when `commit_count` is set to 10:
 //!
@@ -27,24 +27,25 @@
 //!
 //!     loop Commits 1-4
 //!         Caller->>Main: commit()
-//!         Main->>BG: queue revision
-//!         Note right of BG: Waiting...
+//!         Main->>Main: store latest revision, count -= 1
+//!         Note right of BG: Waiting (count > threshold)
 //!     end
 //!
 //!     Caller->>Main: commit() (5th)
-//!     Main->>BG: queue revision
-//!     BG->>Disk: persist revision 5
+//!     Main->>Main: store latest revision, count -= 1
+//!     Main->>BG: notify persist_ready
+//!     BG->>Disk: persist latest revision
 //!     Note right of Disk: Sub-interval (10/2) reached
 //!
 //!     loop Commits 6-8
 //!         Caller->>Main: commit()
-//!         Main->>BG: queue revision
-//!         Note right of BG: Waiting...
+//!         Main->>Main: store latest revision, count -= 1
+//!         Note right of BG: Waiting (count > threshold)
 //!     end
 //!
 //!     Caller->>Main: close()
-//!     Main->>BG: shutdown + persist last committed revision
-//!     BG->>Disk: persist revision 8
+//!     Main->>BG: shutdown signal
+//!     BG->>Disk: persist last committed revision
 //!     Note right of Disk: Latest committed revision is persisted
 //! ```
 
@@ -61,7 +62,6 @@ use firewood_storage::{
 use parking_lot::{Condvar, Mutex, MutexGuard};
 
 use crate::{manager::CommittedRevision, root_store::RootStore};
-use crossbeam::channel::{self, Receiver, Sender};
 
 use firewood_storage::logger::error;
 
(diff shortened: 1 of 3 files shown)


firewood/src/persist_worker.rs
```diff
@@ -186,10 +192,22 @@ impl PersistWorker {
...
+        let _ = self
+            .shared
+            .persist_on_shutdown
+            .set(latest_committed_revision);
+
+        {
+            let mut state = self.shared.state.lock();
+            state.shutdown = true;
+            self.shared.persist_ready.notify_one();
+            self.shared.commit_not_full.notify_all();
```
RodrigoVillar, line 209: This is duplicated in `PersistLoop::run()`, maybe we should add a `shutdown` method (would also help in documenting what these steps do)?

firewood/src/persist_worker.rs
```diff
@@ -110,43 +96,61 @@ impl PersistWorker {
...
+        // Wake the persister once we reach the threshold (or on any reap).
+        if state.count <= self.shared.persist_threshold {
+            self.shared.persist_ready.notify_one();
+        }
 
-        self.sender
-            .as_ref()
-            .ok_or(PersistError::ChannelDisconnected)?
-            .send(PersistMessage::Commit(committed))
-            .map_err(|_| self.resolve_worker_error())
```
RodrigoVillar, line 149: One con of this new code is that if the background thread crashes, we won't know until we call `close()` on the `PersistWorker` (we'll instead get `PersistError::ChannelDisconnected` errors but this goes against the "fail fast" principle IMO).

firewood/src/persist_worker.rs
```diff
@@ -340,62 +283,85 @@ impl PersistLoop {
...
-            // Clear persist_on_shutdown before persisting to avoid holding
-            // an Arc reference during the persist operation.
-            self.persist_on_shutdown = None;
-            self.persist_and_release(&revision, commits_since_persist)
-        } else {
-            // Store the revision so we can persist it on shutdown if needed.
-            self.persist_on_shutdown = Some(revision);
-            Ok(())
+        // Persist the last unpersisted revision on shutdown
+        if let Some(revision) = self.shared.persist_on_shutdown.get().cloned()
```
bernard-avalabs, line 346: nit: It looks like shutdown is triggered from a call to close, which already sets the shared.shutdown flag to true and notifies all of the callers to persist. Releasing the permits seems unnecessary unless the underlying data structure will later be reused (which seems unnecessary).

Should be safe to just persist_to_disk and save to root_store and remove the block above to get the outstanding_commits unless we want to have that for logging.

firewood/src/persist_worker.rs
```diff
@@ -186,10 +192,22 @@ impl PersistWorker {
...
-        // thread to exit.
-        drop(self.sender.take());
+    pub(crate) fn close(
+        &mut self,
+        latest_committed_revision: CommittedRevision,
+    ) -> Result<(), PersistError> {
+        // Ignore the error if already set.
+        let _ = self
+            .shared
+            .persist_on_shutdown
```
bernard-avalabs, line 202: Is a separate persist_on_shutdown OnceLock necessary? Why not just write it into shared.state.latest_revision? Once the shutdown flag has been set to true, no other thread will replace it with a different revision. 

When the event loop thread breaks out of the loop, it can just persist the latest_revision instead of a separate revision stored in persist_on_shutdown.

firewood/src/persist_worker.rs
```diff
@@ -110,43 +96,61 @@ impl PersistWorker {
...
         Self {
             handle: Mutex::new(Some(handle)),
-            sender: Some(sender),
             shared,
         }
     }
 
     /// Sends `committed` to the background thread for persistence. This call
     /// blocks if the limit of unpersisted commits has been reached.
     pub(crate) fn persist(&self, committed: CommittedRevision) -> Result<(), PersistError> {
```
bernard-avalabs, line 134: This comment is not specific to this PR, but about the overall design.

Are we assuming there can be more than one thread calling persist? If yes, then the ordering in which they call persist can be arbitrary, and an older CommittedRevision can overwrite a newer one if we don't include a version number.  If not (single producer/single consumer), then this is fine.

firewood/src/persist_worker.rs
```diff
@@ -437,3 +409,24 @@ impl PersistLoop {
         Ok(())
     }
 }
+
+impl SharedState {
+    #[cfg(test)]
+    #[allow(clippy::arithmetic_side_effects)]
+    fn wait_all_released(&self) {
```
bernard-avalabs, line 416: It might be helpful even for a function used only for testing to add a comment stating that this can block forever if it is called before the number of commits reach the threshold, and no additional persists are performed.

firewood/src/persist_worker.rs
```diff
@@ -310,25 +241,37 @@ impl PersistSemaphore {
...
     root_store: Option<Arc<RootStore>>,
+    /// Unpersisted revision to persist on shutdown.
+    persist_on_shutdown: OnceLock<CommittedRevision>,
+    /// Coordination state for commit and persist operations.
+    state: Mutex<PersistState>,
+}
+
+#[derive(Debug)]
+struct PersistState {
+    count: u64,
```
bernard-avalabs, line 265: Maybe something more descriptive such as available_commit_slots?

firewood/src/persist_worker.rs
RodrigoVillar: Only one thread can be calling `commit()` (which then calls `persist()` at a time):

https://github.com/ava-labs/firewood/blob/f3a60b49c33fdd9fa488cac10f626fc6dbb57d7f/firewood/src/manager.rs#L306-L314

RodrigoVillar (changes requested):
Some initial comments, can discuss in our meeting later.

firewood/src/persist_worker.rs
rkuris: I wonder if a heartbeat could be added, or if we can use standard poison mutex semantics for this.

One potential problem is if the thread crash happens during I/O when we (hopefully) hold no locks.

rkuris:
This was just a checked-in prototype.

PR #1701, opened 2026-02-19
refactor: close consumes itself

### LINKED ISSUE #1699: Refactor `close()`
There is some code smell here if it's possible to drop stuff twice. There are a few choices, many of them aren't all that great, but I think the easiest answer here is:

 - Database close tells the revision manager to send a shutdown message to the worker. That message can contain a reply one-time channel with the final error status of the completed persist operation, but it leaves the worker thread alive. If it gets a "channel closed" error, then you can assume the thread aborted and wait for the thread to exit, collecting the error.
 - Drop shouldn't bother sending the shutdown message. All our use cases we will call close() explicitly. Otherwise, the most recent revision isn't guaranteed to be on disk and/or we'll miss passing I/O errors back to the caller.
 
 I think this simplifies things and never leaves the RevisionManager in a state where it couldn't continue to operate.

_Originally posted by @rkuris in https://github.com/ava-labs/firewood/pull/1690#discussion_r2824902228_

RodrigoVillar:
## Why this should be merged

Closes #1699.

## How this works

Follows the [RAII](https://doc.rust-lang.org/rust-by-example/scope/raii.html) design by having the revision manager consume itself upon closing. This eliminates the previous scenario where, after calling `close()`, the revision manager was still usable even though calls such as `commit()` would fail.

In particular, this PR does the following:

- Makes `RevisionManager::close()` consume itself + eliminates its `Drop` implementation
- Refactors `TestDb` to automatically close the inner database at the end of the test
- Adds `close()` calls to the `fwdctl` crate.

## How this was tested

CI.

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -141,8 +141,12 @@ pub struct DbConfig {
     pub deferred_persistence_commit_count: NonZeroU64,
 }
 
-#[derive(Debug)]
 /// A database instance.
+///
+/// Callers **must** call `close()` when they are done with the database
+/// to ensure the latest committed revision is persisted to disk. Dropping a
+/// database without calling `close()` may result in committed data being lost.
+#[derive(Debug)]
 pub struct Db {
     metrics: Arc<DbMetrics>,
     manager: RevisionManager,
@@ -342,10 +346,10 @@ impl Db {
 
     /// Closes the database gracefully.
     ///
-    /// This method shuts down the background persistence worker and persists
-    /// the last committed revision. If not called explicitly, `Drop` will
-    /// attempt a best-effort shutdown but cannot report errors.
-    pub fn close(mut self) -> Result<(), api::Error> {
+    /// Shuts down the background persistence worker and persists the latest
+    /// committed revision to disk. This method **must** be called before the
+    /// database is dropped as otherwise, any committed data may be lost.
+    pub fn close(self) -> Result<(), api::Error> {
         self.manager.close().map_err(Into::into)
     }
 }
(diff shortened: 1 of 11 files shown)


firewood/src/db.rs
```diff
@@ -1090,7 +1094,7 @@ mod test {
         const CHANNEL_CAPACITY: usize = 8;
 
         let testdb = TestDb::new();
-        let db = &testdb.db;
+        let db: &Db = &testdb;
```
demosdemon: nit:

```suggestion
        let db = &*testdb;
```

I find the explicit deref to be cleaner in cases like this and also automatically adapts if the deref target were to ever change.

firewood/src/db.rs
```diff
@@ -342,10 +346,10 @@ impl Db {
 
     /// Closes the database gracefully.
     ///
-    /// This method shuts down the background persistence worker and persists
-    /// the last committed revision. If not called explicitly, `Drop` will
-    /// attempt a best-effort shutdown but cannot report errors.
-    pub fn close(mut self) -> Result<(), api::Error> {
+    /// Shuts down the background persistence worker and persists the latest
+    /// committed revision to disk. This method **must** be called before the
+    /// database is dropped as otherwise, any committed data may be lost.
```
bernard-avalabs, line 351: Is there a need to call `close`, but then still access the Db struct afterwards? If we always drop the database shortly after calling `close`, then perhaps we should just move its functionality to the `drop` function in the `Drop` trait.

demosdemon: The main problem is `Drop` does not give you the ability to receive an error and handle it. `Drop` must be infallible or panic.

firewood/src/db.rs
RodrigoVillar: Done.

PR #1705, opened 2026-02-20
perf: Use `runtime.AddCleanup`

alarso16:
## Why this should be merged

It's better practice, supposedly. At least more difficult to misuse. Fun fact - they can run concurrently!

## How this works

Required a lot of refactoring, hopefully it's clearer. Didn't work on proofs yet though, not sure what the guarantees need to be

## How this was tested

CI


rkuris (commented):
Did this go through any perf testing? It does seem a bit heavier than the previous implementation, but I'd like to know how much this costs before we merge it.

rkuris:
> Did this go through any perf testing? It does seem a bit heavier than the previous implementation, but I'd like to know how much this costs before we merge it.

https://github.com/ava-labs/firewood/blob/main/METRICS.md#running-benchmarks-from-github-ui has instructions on how to do this. If there are any problems running this or you think it has a lot of resistance, please reach out to @Elvis339 for resolution.

alarso16:
https://github.com/ava-labs/firewood/actions/runs/22407811806
here's some result! It looks slightly slower, but is that just random variation?

ffi/firewood_test.go
```diff
@@ -165,7 +166,7 @@ func newTestDatabase(tb testing.TB, opts ...Option) *Database {
 			runtime.GC()
 			err = db.Close(oneSecCtx(tb))
 		}
-		r.NoError(err)
+		assert.NoError(tb, err) // don't mask any earlier error in the test
```
demosdemon, line 166: can you copy this change to the other tests as well?
- https://github.com/ava-labs/firewood/blob/39cba5afaf1f0eac280a5077948dfe0983240cd0/ffi/tests/eth/eth_compatibility_test.go#L95
- https://github.com/ava-labs/firewood/blob/39cba5afaf1f0eac280a5077948dfe0983240cd0/ffi/tests/firewood/merkle_compatibility_test.go#L84

discussion for a separate pr: should we move some repetitive test stuff into its own package for DRY?

demosdemon (approved):

PR #1706, opened 2026-02-21
chore: Created channel abstraction around lock and condvars

bernard-avalabs:
## Why this should be merged

Builds on Ron's draft PR (https://github.com/ava-labs/firewood/pull/1697). Creates a channel-like abstraction around the lock and condvars to make it easier to use. Purpose of PR is to illustrate the simpler channel-like abstraction. PR should not be merged directly

## How this works

Committer and persister use a shared channel abstraction for communication. Channel is implemented using locks and condition variables underneath. Revisions are pushed and popped. Number of available permits is automatically updated on the push and when the popped object is dropped. Condvars are called inside push and pop and drop. Note that reaps are only performed when a persist is triggered.

## How this was tested

CI

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -254,30 +254,32 @@ impl<T> PersistChannel<T> {
     }
 
     fn pop(&self) -> Result<PersistDataWrapper<'_, T>, PersistError> {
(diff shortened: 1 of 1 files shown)


rkuris:
Out of date PR

Issue #1694, opened 2026-02-17
Placeholder: investigate switching from a channel to locks and convars


rkuris:
But the worker is still blocking on the channel. I'm not sure I understand the difference between blocking on a channel versus blocking on a condition variable.

In an effort of separating concerns that are coupled, the code introduces an implicit dependency between the counter in the semaphore and a secondary counter that can lead to a deadlock if there are any "math" errors. Instead of trying to catch this through comprehensive code review, it would be better to use a synchronization approach where there is a single counter (if possible).

_Originally posted by @bernard-avalabs in https://github.com/ava-labs/firewood/pull/1685#discussion_r2818758724_

bernard-avalabs:
The current solution uses a combination of a modified semaphore and a channel to apply back pressure, with the goal of bounding the staleness of the most recent persisted revision. In general, mixing semaphores and channels (or any other synchronization primitive) can result in tricky and difficult to reason about solutions. In this case, there is an implicit dependence between the values of the counter inside the semaphore, and a separate counter managed by the event loop to determine when to "release permits" in the modified semaphore. This can lead to a deadlock if the semaphore counter reaches zero before the event loop counter falls below the threshold to reset the semaphore. The current solution may work, but it is fragile since a change to just one of the counter thresholds (max permits or count value before reset) could lead to a deadlock that might be difficult to find during standard tests.

This back pressure problem is actually very similar to the standard textbook bounded buffer problem, with the only difference relevant to synchronization being when to signal the consumer of the buffer, which is a common variant of the problem.

The textbook solution to the bounded buffer-based producer/consumer problem uses 1 lock, and 2 condition variables, which are synchronization primitives that are designed to work together. Here is C-like pseudo-code for the modified solution that meets the basic requirements of our problem:

```
// Assume 0 < THRESHOLD < MAX_COUNT. 
// Example values can be THRESHOLD = 2, MAX_COUNT = 4
int shared_count = MAX_COUNT
int shared_item;

void produce(int item) {
    lock(&mutex);
    while (shared_count == 0) {
        cond_wait(&commit_not_full, &mutex);
    }

    // Update shared item
    shared_item = item;
    
    // Decrement and check threshold trigger
    if (--shared_count <= THRESHOLD) {
        cond_signal(&persist_ready);  // Assuming just one consumer
    }
    unlock(&mutex);
}

int consume() {
    lock(&mutex);
    while (shared_count > THRESHOLD) {
[...]

PR #1710, opened 2026-02-24
refactor: replace `PersistSemaphore` with `PersistChannel`

RodrigoVillar:
## Why this should be merged

Quoting #1694:

> The current solution uses a combination of a modified semaphore and a channel to apply back pressure, with the goal of bounding the staleness of the most recent persisted revision. In general, mixing semaphores and channels (or any other synchronization primitive) can result in tricky and difficult to reason about solutions. In this case, there is an implicit dependence between the values of the counter inside the semaphore, and a separate counter managed by the event loop to determine when to "release permits" in the modified semaphore. This can lead to a deadlock if the semaphore counter reaches zero before the event loop counter falls below the threshold to reset the semaphore. The current solution may work, but it is fragile since a change to just one of the counter thresholds (max permits or count value before reset) could lead to a deadlock that might be difficult to find during standard tests.

## How this works

Replaces the `PersistSemaphore` + `crossbeam::channel` with `PersistChannel`, a channel-like abstraction around the locks/condvars mentioned in #1694. By using `PersistChannel`, we no longer have the issue of the `PersistWorker` and the `PersistLoop` having their own notions of progress as all progress is managed via `PersistChannelState`.

This PR also adds a drop guard to the `PersistLoop`, such that if the background thread exits for whatever reason, the system is marked as shutdown which prevents the `PersistWorker` from hanging.

## How this was tested

CI + existing deferred persistence UTs

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -73,8 +73,8 @@ pub enum PersistError {
     FileIo(#[from] Arc<FileIoError>),
     #[error("RootStore error during persistence: {0}")]
     RootStore(#[source] Arc<dyn std::error::Error + Send + Sync>),
-    #[error("Failed to signal background thread after shutdown")]
-    ChannelDisconnected,
+    #[error("Persist worker has shut down")]
+    Shutdown,
 }
 
 /// Handle for managing the background persistence thread.
@@ -99,7 +99,7 @@ impl PersistWorker {
     ) -> Self {
         let persist_interval = NonZeroU64::new(commit_count.get().div_ceil(2))
             .expect("a nonzero div_ceil(2) is always positive");
-        let persist_threshold = commit_count.get().saturating_sub(persist_interval.get());
+        let persist_threshold = commit_count.get().wrapping_sub(persist_interval.get());
 
         let shared = Arc::new(SharedState {
             error: OnceLock::new(),
@@ -109,11 +109,8 @@ impl PersistWorker {
             channel: PersistChannel::new(commit_count, persist_threshold),
         });
 
-        let persist_loop = PersistLoop {
-            shared: shared.clone(),
-        };
-
-        let handle = thread::spawn(move || persist_loop.run());
+        let bg_shared = shared.clone();
+        let handle = thread::spawn(move || PersistLoop { shared: bg_shared }.run());
 
         Self {
             handle: Mutex::new(Some(handle)),
@@ -123,28 +120,53 @@ impl PersistWorker {
 
     /// Sends `committed` to the background thread for persistence. This call
     /// blocks if the limit of unpersisted commits has been reached.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the background thread has shut down or if it
+    /// exited due to an earlier failure.
+    ///
+    /// ## Panics
+    ///
+    /// Propagates any panic from the background thread.
     pub(crate) fn persist(&self, committed: CommittedRevision) -> Result<(), PersistError> {
-        self.shared.channel.push(committed)
+        if self.shared.channel.push(committed).is_err() {
+            self.join_handle();
+            self.check_error()?;
+        }
+
+        Ok(())
     }
 
     /// Sends `nodestore` to the background thread for reaping if archival mode
     /// is disabled. Otherwise, the `nodestore` is dropped.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the background thread has shut down or if it
+    /// exited due to an earlier failure.
+    ///
+    /// ## Panics
+    ///
+    /// Propagates any panic from the background thread.
     pub(crate) fn reap(
         &self,
         nodestore: NodeStore<Committed, FileBacked>,
     ) -> Result<(), PersistError> {
-        if self.shared.root_store.is_none() {
-            self.shared.channel.reap(nodestore)?;
+        if self.shared.root_store.is_none() && self.shared.channel.reap(nodestore).is_err() {
+            self.join_handle();
+            self.check_error()?;
         }
+
         Ok(())
     }
 
-    /// Get a lock to the header of the database.
+    /// Gets a lock to the header of the database.
     pub(crate) fn locked_header(&self) -> MutexGuard<'_, NodeStoreHeader> {
         self.shared.header.lock()
     }
 
-    /// Check if the persist worker has errored.
+    /// Checks if the persist worker has errored.
     pub(crate) fn check_error(&self) -> Result<(), PersistError> {
         match self.shared.error.get() {
             Some(err) => Err(err.clone()),
@@ -152,14 +174,24 @@ impl PersistWorker {
         }
     }
 
-    /// Close the persist worker and persist `latest_committed_revision`.
+    /// Closes the persist worker and persists `latest_committed_revision`.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the background thread exited due to an earlier
+    /// failure.
+    ///
+    /// ## Panics
+    ///
+    /// Propagates any panic from the background thread.
     pub(crate) fn close(
-        mut self,
+        self,
         latest_committed_revision: CommittedRevision,
     ) -> Result<(), PersistError> {
         self.shared
             .persist_on_shutdown
-            .set(latest_committed_revision);
+            .set(latest_committed_revision)
+            .expect("should be empty");
 
         self.shared.channel.close();
         self.join_handle();
@@ -171,7 +203,7 @@ impl PersistWorker {
     /// This is a no-op if the handle was already taken (e.g., by a prior call
     /// to `close()`), which guarantees idempotency.
     ///
-    /// # Panics
+    /// ## Panics
     ///
     /// Propagates the panic if the background thread panicked.
     fn join_handle(&self) {
@@ -182,23 +214,25 @@ impl PersistWorker {
         }
     }
 
-    /// Wait until all pending commits have been persisted.
+    /// Waits until all pending commits have been persisted.
     #[cfg(test)]
     pub(crate) fn wait_persisted(&self) {
         self.shared.wait_all_released();
     }
 }
 
+/// Abstraction for coordinating between the main thread and the background
+/// persistence thread.
 #[derive(Debug)]
-struct PersistChannel<T> {
-    state: Mutex<PersistChannelState<T>>,
+struct PersistChannel {
+    state: Mutex<PersistChannelState>,
     /// Condition variable to wake blocked committers when space is available.
     commit_not_full: Condvar,
     /// Condition variable to wake the persister when persistence is needed.
     persist_ready: Condvar,
 }
 
-impl<T> PersistChannel<T> {
+impl PersistChannel {
     const fn new(max_permits: NonZeroU64, persist_threshold: u64) -> Self {
         Self {
             state: Mutex::new(PersistChannelState {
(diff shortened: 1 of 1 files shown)


firewood/src/persist_worker.rs
```diff
@@ -217,80 +183,114 @@ impl PersistWorker {
...
+
+        // Wake the persister once we reach the threshold.
+        if state.permits_available <= state.persist_threshold {
+            self.persist_ready.notify_one();
+        }
+        Ok(())
+    }
+
+    fn pop(&self) -> Result<PersistDataWrapper<'_, T>, PersistError> {
+        let mut state = self.state.lock();
```
bernard-avalabs:
```        
let (permits_to_release, pending_reaps, data) = {
    let mut state = self.state.lock();
    loop {
    ...
    }
};
```
I just noticed in my branch that swapping the block ordering will drop the lock marginally earlier.

firewood/src/persist_worker.rs
```diff
@@ -299,40 +367,102 @@ impl PersistSemaphore {
...
+/// Drop guard for closing the persist channel if the background thread panics.
+///
+/// This ensures blocked committers are woken up and see the shutdown
+/// state rather than blocking indefinitely.
+struct PanicGuard(Arc<SharedState>);
+
+impl Drop for PanicGuard {
+    fn drop(&mut self) {
+        if std::thread::panicking() {
+            self.0.channel.close();
```
demosdemon: doesn't need to be closed only when panicking. it should always close the channel once dropped because it means the loop has exited for any reason.

firewood/src/persist_worker.rs
```diff
@@ -217,80 +223,142 @@ impl PersistWorker {
...
+    /// ## Errors
+    ///
+    /// Returns an error if the channel has been shut down.
+    fn reap(&self, nodestore: NodeStore<Committed, FileBacked>) -> Result<(), PersistError> {
+        let mut state = self.state.lock();
+        if state.shutdown {
+            return Err(PersistError::ChannelDisconnected);
         }
+        state.pending_reaps.push(nodestore);
+
```
bernard-avalabs, line 270: I believe a `self.persist_ready.notify_one();` is necessary here after the push since the persister should be woken up to handle each reap.

firewood/src/persist_worker.rs
```diff
@@ -217,80 +223,142 @@ impl PersistWorker {
...
+            latest_committed,
+        })
+    }
+
+    /// Signals shutdown and wakes all waiting threads (both committers and
+    /// the background persister).
+    fn close(&self) {
+        let mut state = self.state.lock();
+        state.shutdown = true;
+        self.persist_ready.notify_all();
```
bernard-avalabs: If there is only one persister, then a `notify_one()` is sufficient. Or we can change all of the `notify_one()` to `notify_all()` to be consistent (there should be no performance penalty). I slightly prefer the latter.

firewood/src/persist_worker.rs
```diff
@@ -73,30 +73,16 @@ pub enum PersistError {
     FileIo(#[from] Arc<FileIoError>),
     #[error("RootStore error during persistence: {0}")]
     RootStore(#[source] Arc<dyn std::error::Error + Send + Sync>),
-    #[error("Failed to send message after background thread channel disconnected")]
+    #[error("Failed to signal background thread after shutdown")]
     ChannelDisconnected,
```
rkuris: nit: This name should be changed

firewood/src/persist_worker.rs
```diff
@@ -111,92 +97,112 @@ impl PersistWorker {
...
+    /// exited due to an earlier failure.
+    ///
+    /// ## Panics
+    ///
+    /// Propagates any panic from the background thread.
     pub(crate) fn persist(&self, committed: CommittedRevision) -> Result<(), PersistError> {
-        self.shared.commit_throttle.acquire();
+        if let Err(e) = self.shared.channel.push(committed) {
+            self.join_handle();
+            self.check_error()?;
```
rkuris, line 135: Right now push can only return `ChannelDisconnected` (which I hope you rename), so perhaps as long as it's this error, we can just return whatever `check_error` returns. But if someone changes `push` to return some other kind of unexpected error, we should prioritize that over whatever the thread thinks is the error.

If that's not what we want, we should call this out in the description of errors above.

firewood/src/persist_worker.rs
```diff
@@ -111,92 +97,112 @@ impl PersistWorker {
...
     /// blocks if the limit of unpersisted commits has been reached.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if the background thread has shut down or if it
+    /// exited due to an earlier failure.
+    ///
+    /// ## Panics
+    ///
+    /// Propagates any panic from the background thread.
```
rkuris, line 131: Is this true? Doesn't it convert the panic from the background thread into an error, or will it panic?

firewood/src/persist_worker.rs
RodrigoVillar: Went with replacing `notify_all()` with `notify_one()` in `close()`

firewood/src/persist_worker.rs
RodrigoVillar: Good catch! Fixed.

firewood/src/persist_worker.rs
RodrigoVillar: Good call; done.

firewood/src/persist_worker.rs
```diff
@@ -449,3 +549,10 @@ fn save_to_root_store(
         PersistError::RootStore(e.into())
     })
 }
+
+impl SharedState {
+    #[cfg(test)]
```
rkuris: nit: this could be on the whole impl block

firewood/src/persist_worker.rs
```diff
@@ -342,66 +471,42 @@ impl PersistLoop {
...
 
-    /// Processes messages until the channel is closed or an error occurs.
-    fn event_loop(&mut self) -> Result<(), PersistError> {
-        let mut commits_since_persist = 0u64;
+    /// Processes pending work until shutdown or an error occurs.
+    fn event_loop(&self) -> Result<(), PersistError> {
+        loop {
+            let Ok(mut persist_data) = self.shared.channel.pop() else {
+                break; // Channel failed due to a shutdown. Break to handle shutdown gracefully.
+            };
```
rkuris: nit: seems overly complicated
```suggestion
        while let Ok(mut persist_data) = self.shared.channel.pop() {
```

firewood/src/persist_worker.rs
```diff
@@ -342,66 +471,42 @@ impl PersistLoop {
...
 
-        while let Ok(message) = self.receiver.recv() {
-            match message {
-                PersistMessage::Reap(nodestore) => self.reap(nodestore)?,
-                PersistMessage::Commit(revision) => {
-                    self.commit(revision, &mut commits_since_persist)?;
-                }
+            for nodestore in std::mem::take(&mut persist_data.pending_reaps) {
+                self.reap(nodestore)?;
+            }
```
rkuris, line 475: nit: also seems overly complicated. We own persist_data so why must we mutate it?

```suggestion
            for nodestore in &persist_data.pending_reaps {
                self.reap(nodestore)?;
            }
```

firewood/src/persist_worker.rs
```diff
@@ -342,66 +471,42 @@ impl PersistLoop {
...
+            if let Some(revision) = persist_data.latest_committed.take() {
+                self.persist_to_disk(&revision)
+                    .and_then(|()| self.maybe_save_to_root_store(&revision))?;
             }
         }
 
         // Persist the last unpersisted revision on shutdown
         if let Some(revision) = self.shared.persist_on_shutdown.get().cloned()
-            && commits_since_persist > 0
+            && !self.shared.channel.empty()
```
rkuris, line 485: why is this test needed? If we have something in `persist_on_shutdown` isn't that enough?

firewood/src/persist_worker.rs
```diff
@@ -299,40 +368,100 @@ impl PersistSemaphore {
...
+
+/// Drop guard for closing the persist channel.
+///
+/// This ensures blocked committers are woken up and see the shutdown
+/// state rather than blocking indefinitely.
+struct DropGuard(Arc<SharedState>);
+
+impl Drop for DropGuard {
+    fn drop(&mut self) {
+        self.0.channel.close();
```
rkuris: Can't we move this to the drop of PersistChannel? Do we only want to do this when we drop the whole DropGuard?

firewood/src/persist_worker.rs
RodrigoVillar: This is the only method on `SharedState`.

firewood/src/persist_worker.rs
RodrigoVillar: This is true - any panic captured by the background thread will be rethrown by the caller of `close()`/`reap()`/`persist()`

firewood/src/persist_worker.rs
RodrigoVillar: Without this call, we would be persist the last revision twice:

- Persist `R` 
- Initiate shutdown (sets `R` as `persist_on_shutdown`)
- After exiting the `while` loop, since `R` is `Some`, we persist `R`

firewood/src/persist_worker.rs
rkuris: Yes, so why can't we just mark the whole impl block as `#[cfg(test)]`?

It's a nit so feel free to ignore this though.

firewood/src/persist_worker.rs
rkuris: Ah so it's `persist_on_shutdown_unless_it_is_already_persisted` :)

Maybe we should rename this to something more sane, since `persist_on_shutdown` says we'll always persist it on shutdown for the casual reader.

Even if you don't rename it, it could use a comment saying that this is ignored if it's already persisted.

firewood/src/persist_worker.rs
RodrigoVillar: Ah I read this as "move this method to the `SharedState` `impl` block", rather than "mark the whole `impl` block as test-only". Changed it so that the entire `impl` block is test-only.

firewood/src/persist_worker.rs
RodrigoVillar: Good call; done.

firewood/src/persist_worker.rs
RodrigoVillar: Added a comment here to make it clear that we persist `R` only if we haven't persisted it.

firewood/src/persist_worker.rs
RodrigoVillar: Moved the `Drop` impl to the `PersistLoop`

firewood/src/persist_worker.rs
RodrigoVillar: Changed to `Shutdown`, as this error message is returned when `state.shutdown` is true.

firewood/src/persist_worker.rs
RodrigoVillar: `nodestore.reap_deleted()` takes in `self` so we need ownership over the nodestores in `pending_reaps`. The `mut` is added by definition of `take()`:

```rust
take<T>(dest: &mut T) -> T
```

firewood/src/persist_worker.rs
RodrigoVillar: After some thought, returning `ChannelDisconnected` in both `push` and `reap` seems to be unreachable code so I agree that we can just return whatever `check_error` returns. Changed.

firewood/src/persist_worker.rs
```diff
@@ -111,92 +97,103 @@ impl PersistWorker {
         header: NodeStoreHeader,
         root_store: Option<Arc<RootStore>>,
     ) -> Self {
-        let (sender, receiver) = channel::unbounded();
         let persist_interval = NonZeroU64::new(commit_count.get().div_ceil(2))
             .expect("a nonzero div_ceil(2) is always positive");
+        let persist_threshold = commit_count.get().saturating_sub(persist_interval.get());
```
demosdemon: nits:

- the `uwnrap` should not lint anymore inside the `const { ... }` block and `div_ceil` is defined for `NonZero<_>` that returns a `NonZero<_>`.
- wrapping_sub is fine here because we logically know that `persist_interval` is strictly less than or equal to `commit_count` so the subtraction will never wrap.

```suggestion
        let persist_interval = commit_count.div_ceil(const { NonZeroU64::new(2).unwrap() });
        let persist_threshold = commit_count.get().wrapping_sub(persist_interval.get());
```

It's important to be aware of the mathematical operations you're doing and not blanket chose `saturating_`, `wrapping_`, or `checked_`. `saturating` and `checked` operations have overhead that we can and should skip if we know our logical operations won't trigger overflow.

firewood/src/persist_worker.rs
```diff
@@ -217,80 +214,143 @@ impl PersistWorker {
...
+        while state.permits_available == 0 && !state.shutdown {
+            self.commit_not_full.wait(&mut state);
+        }
+
+        if state.shutdown {
+            return Err(PersistError::Shutdown);
+        }
+
+        state.latest_committed = Some(revision);
+        state.permits_available = state.permits_available.saturating_sub(1);
```
demosdemon: another example here. The compiler might actually skip the extra work from `saturation` because we have an earlier check  that `permits_available` is not zero. However, that condition is not exclusive to letting control flow to here, so the compiler might conservatively check again. However, because we know that because `state.shutdown` is false at this point, we also know that `state.permits_available` is non-zero and saturation is extra work.

firewood/src/persist_worker.rs
RodrigoVillar: Thanks for the comments! Your first point about using a `const` block, while cleaner, seems to only be a unstable feature:

```
use of unstable library feature `unsigned_nonzero_div_ceil`
see issue #132968 <https://github.com/rust-lang/rust/issues/132968> for more information
```

As for the second point, I went ahead and changed this to a `wrapping_sub()`. Likewise, I went through the persistence code and identified that all `saturating_sub()` calls can be changed to `wrapping_sub()` as logically, those operations will not trigger an underflow.

firewood/src/persist_worker.rs

PR #1711, opened 2026-02-24
perf(ffi): clamp histogram buckets to linux timer resolution

rkuris:
## Why this should be merged

There is a ton of noise in histograms because the timer resolution is much larger, especially when the thing we are timing is very fast.

## How this works

This calls the OS to ask what the timer resolution is and adjusts the buckets accordingly.

## How this was tested

Unit tests and CI. We'll need to see how well this talks to the histogram dashboards and see if it eliminates noise.

### DIFF
--- a/ffi/src/metrics.rs
+++ b/ffi/src/metrics.rs
@@ -128,13 +128,16 @@ fn monotonic_coarse_resolution_ms() -> Option<f64> {
         unsafe_code,
         reason = "Required to call clock_getres for OS timer resolution"
     )]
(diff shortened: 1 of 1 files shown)


rkuris:
> Once we swap coarsetime to std::time, the timer has nanosecond precision and there's nothing to clamp. This mechanism becomes dead code that needs to be removed.

Yes, but this assumes there's no perf penalty for switching. We might need to support both for some use cases.

rkuris:
Closing for now, hoping the std::time duration doesn't create a perf difference.

PR #1712, opened 2026-02-24
refactor(metrics): migrate ffi histogram recording to macros

rkuris:
## Why this should be merged

Centralize the histogram code and follow the same pattern as used for the counters and timing metrics.

## How this works

Brand new macro. I kept "expensive" as a specific option because I think we might want some histograms even if expensive is disabled.

## How this was tested

Just unit tests and CI.

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -16,7 +16,9 @@ use firewood::{
 use crate::{BatchOp, BorrowedBytes, CView, CreateProposalResult, arc_cache::ArcCache};
 
 use crate::revision::{GetRevisionResult, RevisionHandle};
-use firewood_metrics::{MetricsContext, firewood_increment, firewood_record};
+use firewood_metrics::{
+    MetricsContext, firewood_increment, firewood_record, fwd_expensive_timed_result,
+};
 
 /// The hashing mode to use for the database.
 ///
@@ -215,7 +217,7 @@ impl DatabaseHandle {
         self.db.revision(root)?.val(key)
     }
 
-    /// Creates a proposal with the given values and returns the proposal and the start time.
+    /// Creates and commits a proposal with the given values.
     ///
     /// # Errors
     ///
@@ -224,19 +226,13 @@ impl DatabaseHandle {
         &self,
         values: impl AsRef<[BatchOp<'a>]> + 'a,
     ) -> Result<Option<HashKey>, api::Error> {
-        let CreateProposalResult { handle, start_time } =
-            self.create_proposal_handle(values.as_ref())?;
-
-        let root_hash = handle.commit_proposal(|commit_time| {
-            firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis());
-            firewood_record!(
-                crate::registry::COMMIT_MS_BUCKET,
-                commit_time.as_f64() * 1000.0,
-                expensive
-            );
-        })?;
-
-        let elapsed = start_time.elapsed();
+        let (root_hash_result, elapsed) =
+            fwd_expensive_timed_result!(crate::registry::BATCH_MS_BUCKET, {
+                let CreateProposalResult { handle } =
+                    self.create_proposal_handle(values.as_ref())?;
+                handle.commit_proposal()
+            });
+        let root_hash = root_hash_result?;
         firewood_increment!(crate::registry::BATCH_MS, elapsed.as_millis());
         firewood_increment!(crate::registry::BATCH_COUNT, 1);
         firewood_record!(
--- a/ffi/src/proposal.rs
+++ b/ffi/src/proposal.rs
@@ -4,7 +4,7 @@
 use firewood::v2::api::{self, BoxKeyValueIter, DbView, HashKey, IntoBatchIter, Proposal as _};
 
 use crate::{IteratorHandle, iterator::CreateIteratorResult, metrics::MetricsContextExt};
-use firewood_metrics::{firewood_increment, firewood_record};
+use firewood_metrics::{firewood_increment, firewood_record, fwd_expensive_timed_result};
 
 /// An opaque wrapper around a Proposal that also retains a reference to the
 /// database handle it was created from.
@@ -63,20 +63,11 @@ impl ProposalHandle<'_> {
 
     /// Consume and commit a proposal.
     ///
-    /// # Arguments
-    ///
-    /// - `token`: An callback function that will be called with the duration
-    ///   of the commit operation. This will be dropped without being called if
-    ///   the commit fails.
-    ///
     /// # Errors
     ///
     /// This function will return an error if committing the proposal fails or if the
     /// proposal is empty.
-    pub fn commit_proposal(
-        self,
-        token: impl FnOnce(coarsetime::Duration),
-    ) -> Result<Option<HashKey>, api::Error> {
+    pub fn commit_proposal(self) -> Result<Option<HashKey>, api::Error> {
         let ProposalHandle {
             hash_key,
             proposal,
@@ -89,14 +80,15 @@ impl ProposalHandle<'_> {
             _ = handle.get_root(hash_key.clone());
         }
 
-        let start_time = coarsetime::Instant::now();
-        proposal.commit()?;
-        let commit_time = start_time.elapsed();
+        let (commit_result, commit_time) =
+            fwd_expensive_timed_result!(crate::registry::COMMIT_MS_BUCKET, proposal.commit());
+        commit_result?;
 
         // clear the cached view so that it does not hold onto the proposal view
         handle.clear_cached_view();
 
-        token(commit_time);
+        firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis());
+        firewood_increment!(crate::registry::COMMIT_COUNT, 1);
 
         Ok(hash_key)
     }
@@ -119,7 +111,6 @@ impl ProposalHandle<'_> {
 #[derive(Debug)]
 pub struct CreateProposalResult<'db> {
     pub handle: ProposalHandle<'db>,
-    pub start_time: coarsetime::Instant,
 }
 
 impl<'db> CreateProposalResult<'db> {
@@ -127,9 +118,9 @@ impl<'db> CreateProposalResult<'db> {
         handle: &'db crate::DatabaseHandle,
         f: impl FnOnce() -> Result<firewood::db::Proposal<'db>, api::Error>,
     ) -> Result<Self, api::Error> {
-        let start_time = coarsetime::Instant::now();
-        let proposal = f()?;
-        let propose_time = start_time.elapsed();
+        let (proposal_result, propose_time) =
+            fwd_expensive_timed_result!(crate::registry::PROPOSE_MS_BUCKET, f());
+        let proposal = proposal_result?;
         firewood_increment!(crate::registry::PROPOSE_MS, propose_time.as_millis());
         firewood_increment!(crate::registry::PROPOSE_COUNT, 1);
         firewood_record!(
@@ -146,7 +137,6 @@ impl<'db> CreateProposalResult<'db> {
                 proposal,
                 handle,
             },
-            start_time,
         })
     }
 }
--- a/metrics/src/lib.rs
+++ b/metrics/src/lib.rs
@@ -58,6 +58,8 @@
 //! | `firewood_set!(name, value, expensive)` | Set only if expensive metrics enabled |
 //! | `firewood_record!(name, value)` | Always record to histogram |
 //! | `firewood_record!(name, value, expensive)` | Record only if expensive metrics enabled |
+//! | `fwd_timed_result!(name, expr)` | Time a `Result` expression and record histogram on `Ok` |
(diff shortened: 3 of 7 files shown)


ffi/src/proposal.rs
Elvis339: The #[histogram_timing] macro on line 76 wraps the entire function, so the histogram 

(ffi.commit_ms_bucket) measures: cached view promotion + proposal.commit() + cached view clearing + token callback.

But the manual timer on line 93 only measures proposal.commit(), and that's what gets passed to the counter (ffi.commit_ms).

After this change, the counter and histogram for the same "commit" operation measure different scopes. Deriving average latency from the counter versus the histogram's _sum / _count will get different numbers with no obvious explanation.

ffi/src/handle.rs
```diff
@@ -212,6 +212,7 @@ impl DatabaseHandle {
     /// # Errors
     ///
     /// An error is returned if the proposal could not be created.
+    #[firewood_macros::histogram_timing(crate::registry::BATCH_MS_BUCKET, expensive)]
     pub fn create_batch<'a>(
```
Elvis339, line 225: Same divergence as in `commit_proposal`

ffi/src/proposal.rs
rkuris: Ah actually I see what you're talking about, ignore my prior comment.

metrics/src/lib.rs
```diff
@@ -265,6 +267,41 @@ macro_rules! firewood_record {
...
+/// let (result, elapsed) = fwd_timed_result!(registry::LATENCY_MS_BUCKET, work());
+/// let value = result?;
+///
+/// let (result, elapsed) = fwd_timed_result!(registry::LATENCY_MS_BUCKET, expensive, work());
+/// let value = result?;
+/// ```
+#[macro_export]
+macro_rules! fwd_timed_result {
+    ($name:expr, expensive, $expr:expr) => {{
+        let __fwd_start = ::coarsetime::Instant::now();
```
Elvis339, line 283: Should we replace `coarsetime` with `std::time` this would partially solve. https://github.com/ava-labs/firewood/issues/1720

metrics/src/lib.rs
```diff
@@ -265,6 +267,41 @@ macro_rules! firewood_record {
...
+        let __fwd_result = $expr;
+        let __fwd_elapsed = __fwd_start.elapsed();
+        if __fwd_result.is_ok() {
+            $crate::fwd_histogram_record!($name, __fwd_elapsed.as_f64() * 1000.0, expensive);
```
Elvis339: What's the rationale for `__fwd_elapsed.as_f64() * 1000.0`?  Could we replace it with `__fwd_elapsed..as_millis()`?

metrics/src/lib.rs
```diff
@@ -265,6 +267,41 @@ macro_rules! firewood_record {
...
+/// # Usage
+/// ```ignore
+/// let (result, elapsed) = fwd_timed_result!(registry::LATENCY_MS_BUCKET, work());
+/// let value = result?;
+///
+/// let (result, elapsed) = fwd_timed_result!(registry::LATENCY_MS_BUCKET, expensive, work());
+/// let value = result?;
+/// ```
+#[macro_export]
+macro_rules! fwd_timed_result {
```
Elvis339, line 281: Would it make sense to split this into `fwd_timed_result!` (always records) and `fwd_expensive_timed_result!`? It's cleaner and no ambiguity about argument order.

firewood-macros/src/lib.rs
```diff
@@ -152,6 +184,83 @@ pub fn metrics(args: TokenStream, input: TokenStream) -> TokenStream {
...
+///
+/// # Usage
+/// ```rust,ignore
+/// #[firewood_macros::histogram_timing(crate::registry::PROPOSE_MS_BUCKET, expensive)]
+/// fn create() -> Result<(), Error> {
+///     Ok(())
+/// }
+/// ```
+#[proc_macro_attribute]
+pub fn histogram_timing(args: TokenStream, input: TokenStream) -> TokenStream {
```
Elvis339: Is this still needed? It seems like it isn’t used anywhere. 

If it's needed and you want to keep it, clarify in the docs that it measures the whole function body, not a specific sub-expression unlike fwd_timed_result!, which can wrap only the part you care about. That difference matters when deciding between the attribute and the macro.

ffi/src/proposal.rs
```diff
@@ -119,24 +122,18 @@ impl ProposalHandle<'_> {
...
         f: impl FnOnce() -> Result<firewood::db::Proposal<'db>, api::Error>,
     ) -> Result<Self, api::Error> {
-        let start_time = coarsetime::Instant::now();
-        let proposal = f()?;
-        let propose_time = start_time.elapsed();
+        let (proposal_result, propose_time) =
+            fwd_timed_result!(crate::registry::PROPOSE_MS_BUCKET, expensive, f());
+        let proposal = proposal_result?;
         firewood_increment!(crate::registry::PROPOSE_MS, propose_time.as_millis());
         firewood_increment!(crate::registry::PROPOSE_COUNT, 1);
```
Elvis339, line 125: We accidentally built a manual histogram. PROPOSE_MS accumulates total latency, PROPOSE_COUNT tracks observations that's exactly what _sum and _count [give us for free](https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations) on any histogram. 

Then we added an actual histogram and put it behind the expensive flag, so the worse version runs always and the better one would get dropped if expensive flag is off.

The histogram is strictly better because counters only give you averages, which hide tail latency. With the histogram we can answer questions like "what % of proposes exceeded X ms?"

metrics/src/lib.rs
```diff
@@ -246,12 +248,12 @@ macro_rules! firewood_gauge {
...
-/// firewood_record!(registry::LATENCY_MS, elapsed_ms);
-/// firewood_record!(registry::LATENCY_MS, elapsed_ms, "op" => "read");
-/// firewood_record!(registry::COMMIT_MS_BUCKET, elapsed_ms, expensive);
+/// fwd_histogram_record!(registry::LATENCY_MS, elapsed_ms);
+/// fwd_histogram_record!(registry::LATENCY_MS, elapsed_ms, "op" => "read");
+/// fwd_histogram_record!(registry::COMMIT_MS_BUCKET, elapsed_ms, expensive);
 /// ```
 #[macro_export]
-macro_rules! firewood_record {
+macro_rules! fwd_histogram_record {
```
Elvis339: This looks like an unnecessary abstraction over [metrics::histogram!](https://docs.rs/metrics/0.24.3/metrics/macro.histogram.html) and it drops label support in the process. 

Callsites could just use `metrics::histogram!` directly.

ffi/src/handle.rs
```diff
@@ -220,26 +220,17 @@ impl DatabaseHandle {
...
-            firewood_record!(
-                crate::registry::COMMIT_MS_BUCKET,
-                commit_time.as_f64() * 1000.0,
-                expensive
-            );
-        })?;
-
-        let elapsed = start_time.elapsed();
+        let (root_hash_result, elapsed) =
+            fwd_timed_result!(crate::registry::BATCH_MS_BUCKET, expensive, {
```
Elvis339: Same comment as in propsal.

ffi/src/proposal.rs
```diff
@@ -119,24 +122,18 @@ impl ProposalHandle<'_> {
...
 impl<'db> CreateProposalResult<'db> {
     pub(crate) fn new(
         handle: &'db crate::DatabaseHandle,
         f: impl FnOnce() -> Result<firewood::db::Proposal<'db>, api::Error>,
     ) -> Result<Self, api::Error> {
-        let start_time = coarsetime::Instant::now();
-        let proposal = f()?;
-        let propose_time = start_time.elapsed();
+        let (proposal_result, propose_time) =
+            fwd_timed_result!(crate::registry::PROPOSE_MS_BUCKET, expensive, f());
```
Elvis339: `_BUCKET` is redundant rename `PROPOSE_MS_BUCKET` to just `PROPOSE_MS` and remove `PROPOSE_COUNT`. See comment below for rationale.

rkuris:
I don't want this PR to turn into a "fix every metric issue" so I'm going to push back some here.

> A Prometheus histogram already provides _sum and _count, so the manual counters are redundant:
> 
> * rate(ffi_commit_ms_sum[5m]) replaces rate(ffi_commit_ms[5m])
> * rate(ffi_commit_ms_count[5m]) replaces rate(ffi_commit[5m])
> 
> COMMIT_MS, COMMIT_COUNT, PROPOSE_MS, PROPOSE_COUNT, BATCH_MS, BATCH_COUNT can all go, along with the token: impl FnOnce(Duration) callback that only exists to feed them.

This problem existed before this PR was started, and as such is unrelated to creating this macro. Removing these extra counters can (and should) be done under a separate PR.

> The bucket range also needs fixing. It caps at 50ms, but non-deferred commits run 280-360ms and proposes run 50-200ms everything piles into +Inf and the distribution tells us nothing.

Also a pre-existing problem. Looks like @PlatCore is looking to fix this in a separate PR.

> Since this PR is establishing the macro as the standard approach, worth getting the buckets right before this ships.

Let's separate the macroization from the fixes. I think the fixes will be easier once we have the macro in place.

> One naming note: the awkward ffi_commit_ms_bucket_sum (double _bucket) comes from the histogram being named ffi.commit_ms_bucket. Renaming it to something like ffi.commit_latency_ms would clean that up.

Also pre-existing.

metrics/src/lib.rs
rkuris: > This looks like an unnecessary abstraction over [metrics::histogram!](https://docs.rs/metrics/0.24.3/metrics/macro.histogram.html) and it drops label support in the process.
> 
> Callsites could just use `metrics::histogram!` directly.

This was just a rename, not a redesign of this macro. Changing this macro should be done under another PR.

Let me know and I can open an issue that describes all these problems for followups.

metrics/src/lib.rs
rkuris: This is done under another PR.

metrics/src/lib.rs
rkuris: This was pre-existing and probably broken. It could use another issue as well.

Elvis339: Seems like this is related: https://github.com/ava-labs/firewood/issues/1735

Elvis339:
> I don't want this PR to turn into a "fix every metric issue" so I'm going to push back some here.
> 
> > A Prometheus histogram already provides _sum and _count, so the manual counters are redundant:
> > 
> > * rate(ffi_commit_ms_sum[5m]) replaces rate(ffi_commit_ms[5m])
> > * rate(ffi_commit_ms_count[5m]) replaces rate(ffi_commit[5m])
> > 
> > COMMIT_MS, COMMIT_COUNT, PROPOSE_MS, PROPOSE_COUNT, BATCH_MS, BATCH_COUNT can all go, along with the token: impl FnOnce(Duration) callback that only exists to feed them.
> 
> This problem existed before this PR was started, and as such is unrelated to creating this macro. Removing these extra counters can (and should) be done under a separate PR.
> 
> > The bucket range also needs fixing. It caps at 50ms, but non-deferred commits run 280-360ms and proposes run 50-200ms everything piles into +Inf and the distribution tells us nothing.
> 
> Also a pre-existing problem. Looks like @PlatCore is looking to fix this in a separate PR.
> 
> > Since this PR is establishing the macro as the standard approach, worth getting the buckets right before this ships.
> 
> Let's separate the macroization from the fixes. I think the fixes will be easier once we have the macro in place.
> 
> > One naming note: the awkward ffi_commit_ms_bucket_sum (double _bucket) comes from the histogram being named ffi.commit_ms_bucket. Renaming it to something like ffi.commit_latency_ms would clean that up.
> 
> Also pre-existing.

I opened https://github.com/ava-labs/firewood/issues/1736, feel free to edit if needed. My intent is to have these things centralized for simpler tracking.

metrics/src/lib.rs
Elvis339: Can you link it here https://github.com/ava-labs/firewood/issues/1736 please? I couldn't find it.

metrics/src/lib.rs
Elvis339: Yes, let's add it here.

https://github.com/ava-labs/firewood/issues/1736

metrics/src/lib.rs
rkuris: Sorry I mis-spoke. I meant to say "should be done under another PR". I think we're looking for numbers before deciding to make this change though.

ffi/src/proposal.rs
rkuris: Resolved in latest push

ffi/src/proposal.rs
```diff
@@ -119,24 +117,18 @@ impl ProposalHandle<'_> {
 #[derive(Debug)]
 pub struct CreateProposalResult<'db> {
     pub handle: ProposalHandle<'db>,
-    pub start_time: coarsetime::Instant,
 }
```
demosdemon, line 114: You can get rid of this struct and return the `ProposalHandle` directly instead. The only reason I made a struct for this was to include the time field. But, that's no longer needed with this change.

PR #1713, opened 2026-02-24
feat(ffi)!: add optional node cache memory limit to open args

rkuris:
## Why this should be merged

Allows coreth to just pass the memory in

## How this was tested

unit tests

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -118,18 +118,17 @@ impl DatabaseHandleArgs<'_> {
 
         let memory_limit = NonZeroUsize::new(self.node_cache_memory_limit);
 
-        let config = match memory_limit {
-            Some(memory_limit) => RevisionManagerConfig::builder()
+        let config = {
+            let builder = RevisionManagerConfig::builder()
                 .max_revisions(self.revisions)
                 .cache_read_strategy(cache_read_strategy)
-                .free_list_cache_size(free_list_cache_size)
-                .node_cache_memory_limit(memory_limit)
-                .build(),
-            None => RevisionManagerConfig::builder()
-                .max_revisions(self.revisions)
-                .cache_read_strategy(cache_read_strategy)
-                .free_list_cache_size(free_list_cache_size)
-                .build(),
+                .free_list_cache_size(free_list_cache_size);
+
+            if let Some(memory_limit) = memory_limit {
+                builder.node_cache_memory_limit(memory_limit).build()
+            } else {
+                builder.build()
+            }
         };
 
         Ok(config)


ffi/src/handle.rs
```diff
@@ -61,9 +61,20 @@ pub struct DatabaseHandleArgs<'a> {
 
     /// The size of the node cache.
     ///
-    /// Opening returns an error if this is zero.
+    /// **Deprecated:** Prefer `node_cache_memory_limit`.
+    ///
+    /// If this and `node_cache_memory_limit` are both non-zero, opening
+    /// returns an error.
     pub cache_size: usize,
```
RodrigoVillar: With `v0.2.0` being the next release, wouldn't it make sense to remove this field altogether?

rkuris: We could, but I'll just do it under a separate PR. This means the avalanchego PR that upgrades to the release will be a little smaller since this could be done separately.

alarso16: I'll use this in the coreth release - I think it makes more sense to get rid of it, unless "number of nodes" is a reasonable method of having a cache (but I don't think it is)

ffi/firewood.go
```diff
@@ -252,7 +268,13 @@ func New(dbDir string, nodeHashAlgorithm NodeHashAlgorithm, opts ...Option) (*Da
...
 		return nil, fmt.Errorf("revisions must be >= 2, got %d", conf.revisions)
 	}
-	if conf.nodeCacheEntries < 1 {
+	if conf.nodeCacheEntries < 1 && conf.nodeCacheSizeInBytes < 1 {
+		return nil, errors.New("either node cache entries or node cache size in bytes must be >= 1")
+	}
+	if conf.nodeCacheEntries > 0 && conf.nodeCacheSizeInBytes > 0 {
+		return nil, errors.New("node cache entries and node cache size in bytes are mutually exclusive")
+	}
+	if conf.nodeCacheEntries < 1 && conf.nodeCacheSizeInBytes < 1 {
```
alarso16: If we are going to allow both options, we should add tests to ensure these errors can be hit.

ffi/firewood.go
```diff
@@ -146,6 +151,17 @@ func WithTruncate(truncate bool) Option {
...
+	}
+}
+
+// WithNodeCacheSizeInBytes sets the node cache memory limit in bytes.
+// When set to a non-zero value, this overrides node cache entry-based sizing.
+// Must be non-zero to enable memory-based sizing.
+func WithNodeCacheSizeInBytes(sizeInBytes uint) Option {
+	return func(c *config) {
+		c.nodeCacheSizeInBytes = sizeInBytes
+		c.nodeCacheEntries = 0
```
alarso16: This might add some unexpected behavior where it actually overwrites a previous `WithNodeCacheEntries`

ffi/firewood.go
alarso16: Should we still keep the default config in number of entries?

ffi/src/handle.rs
```diff
@@ -112,21 +113,25 @@ impl DatabaseHandleArgs<'_> {
...
+                .node_cache_memory_limit(memory_limit)
+                .build(),
+            None => RevisionManagerConfig::builder()
+                .max_revisions(self.revisions)
+                .cache_read_strategy(cache_read_strategy)
+                .free_list_cache_size(free_list_cache_size)
+                .build(),
+        };
+
         Ok(config)
```
RodrigoVillar:
```suggestion
        let mut config = RevisionManagerConfig::builder()
                .max_revisions(self.revisions)
                .cache_read_strategy(cache_read_strategy)
                .free_list_cache_size(free_list_cache_size);
        if let Some(memory_limit) = memory_limit {
            config.node_cache_memory_limit(memory_limit);
        }

        Ok(config.build())
```

nit: delegating the `memory_limit` case to its own branch is easier to follow IMO.

ffi/firewood.go
```diff
@@ -139,13 +139,12 @@ func WithTruncate(truncate bool) Option {
 	}
 }
 
-// WithNodeCacheEntries sets the number of entries in the node cache.
-// The node cache stores frequently accessed trie nodes to improve read performance.
+// WithNodeCacheSizeInBytes sets the node cache memory limit in bytes.
 // Must be non-zero.
-// Default: 1,000,000
-func WithNodeCacheEntries(entries uint) Option {
+// Default: 128,000,000
```
```suggestion
// Default: 128 MB
```
Do you agree?

ffi/src/handle.rs
rkuris: Fixed in a slightly different way. I hate `mut`.

rkuris:
> Should we remove this?
> 
> https://github.com/ava-labs/firewood/blob/94641ecefaf724dd76e8c3d839994dbb01d368a1/firewood/src/manager.rs#L42-L49

Yes, referenced in new issue. I'll get to this in a few days.

PR #1714, opened 2026-02-24
chore(ci): Add elvis when tagged observability

rkuris:
## Why this should be merged

Any observability changes should get reviewed by Elvis, this might do that.

## How this works

Supposedly this will auto assign when the tag is added, but I'm not 100% sure. Let's try it.

## How this was tested

I don't know how other than to merge it :(

### DIFF
--- a/.github/auto_assign.yml
+++ b/.github/auto_assign.yml
@@ -1,0 +1,4 @@
+addReviewers: true
+reviewers:
+  - Elvis339
+labelTrigger: "observability"


rkuris:
> > Supposedly this will auto assign when the tag is added, but I'm not 100% sure. Let's try it.
> 
> Do you know what GH action will read this file and do the automatic assign?

Not sure. Do you know how to do this?

The other option here is to do it where we are currently scanning for dashboard changes and maybe automatically tag it observability at the same time?

RodrigoVillar:
@rkuris from a quick search (https://share.google/aimode/tBbdxQXm0axPYqXLq), I'm pretty sure we'll need to add a GH action that will do the work of reading the YAML file and subsequently tagging Elvis as a reviewer.

rkuris:
Let's do this differently, maybe using the old "metrics changed" detection to add Elvis,

PR #1716, opened 2026-02-25
feat(proofs): 2/6 Range Proof Verification

rkuris:
Add in-memory branch proof-node reconciliation code

- Added reconcile_branch_proof_node for branch proof-node reconciliation.

How it works:
- reconcile_branch_proof_node converts proof-node key path components
  into nibble bytes and ensures the branch structure exists.
- If the proof node carries ValueDigest::Value:
  - existing conflicting branch value => ProofError::UnexpectedValue
  - existing matching branch value => no-op
  - missing branch value => inserts the provided value at that branch key
- If the proof node has no value or hash-only digest, value mutation is deferred
  for later hash-validation phases.
- Rejects odd-nibble keys that carry inline values with ProofError::ValueAtOddNibbleLength.

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -1360,10 +1360,18 @@ impl Merkle<NodeStore<Mutable<Propose>, MemStore>> {
 
     /// Reconciles a branch proof node against the in-memory proving merkle.
     ///
-    /// This helper never overwrites an existing branch value. It only:
-    /// - creates missing branch structure,
-    /// - inserts a value when the branch exists without one, and
-    /// - errors if an existing value conflicts with the proof.
+    /// This helper never overwrites an existing branch value. It only
+    /// creates missing branch structure and inserts a value when the
+    /// branch exists without one.
+    ///
+    /// ## Arguments
+    ///
+    /// * `proof_node` - A branch proof node containing the key (as nibble
+    ///   path components) and an optional value digest to reconcile.
+    ///
+    /// ## Errors
+    ///
+    /// Returns an error if an existing value conflicts with the proof.
     pub(crate) fn reconcile_branch_proof_node(
         &mut self,
         proof_node: &ProofNode,


firewood/src/merkle/mod.rs
```diff
@@ -1131,6 +1314,98 @@ impl<S: ReadableStorage> Merkle<NodeStore<MutableProposal, S>> {
...
+        {
+            return Err(ProofError::ValueAtOddNibbleLength);
+        }
+
+        self.insert_branch_from_nibbles(key_nibbles)?;
+
+        let Some(ValueDigest::Value(proof_value)) = proof_node.value_digest.as_ref() else {
+            // Hash-only value digests and absent values are validated in later
+            // proof-hash reconstruction steps.
+            return Ok(());
```
bernard-avalabs: It might be helpful to return a enum of different valid cases even if it is only for testing.

rkuris: Resolved in 0584121383afbc5188dc2f6656d3028dea20c785

firewood/src/merkle/mod.rs
```diff
@@ -1345,6 +1357,57 @@ impl Merkle<NodeStore<Mutable<Propose>, MemStore>> {
         *self.nodestore.root_mut() = root_node.into();
         Ok(())
     }
+
+    /// Reconciles a branch proof node against the in-memory proving merkle.
+    ///
+    /// This helper never overwrites an existing branch value. It only:
+    /// - creates missing branch structure,
+    /// - inserts a value when the branch exists without one, and
+    /// - errors if an existing value conflicts with the proof.
```
RodrigoVillar: nit: could we follow the existing convention in this file for describing errors? As an example:

```
/// Returns a proof that the given key has a certain value,
/// or that the key isn't in the trie.
///
/// ## Errors
///
/// Returns an error if the trie is empty or an error occurs while reading from storage.
pub fn prove(&self, key: &[u8]) -> Result<FrozenProof, ProofError> {
```

firewood/src/merkle/mod.rs
```diff
@@ -1345,6 +1357,57 @@ impl Merkle<NodeStore<Mutable<Propose>, MemStore>> {
...
+    ) -> Result<ReconcileResult, ProofError> {
+        let key_nibbles = proof_node
+            .key
+            .iter()
+            .map(|component| component.as_u8())
+            .collect::<Vec<_>>();
+        let key_nibbles = key_nibbles.as_slice();
+
+        if !key_nibbles.len().is_multiple_of(2)
+            && matches!(proof_node.value_digest, Some(ValueDigest::Value(_)))
```
alarso16, line 1387: What is the second part of this check? I don't understand the syntax (never seen the macro or the `_` value

firewood/src/merkle/tests/mod.rs
```diff
@@ -285,6 +304,72 @@ fn test_insert_branch_from_nibbles_splits_divergent_paths() {
     );
 }
 
+#[test]
```
alarso16, line 307: Unrelated to the code, but why are the tests in a separate package?

firewood/src/merkle/tests/mod.rs
```diff
@@ -285,6 +304,72 @@ fn test_insert_branch_from_nibbles_splits_divergent_paths() {
...
+    merkle.insert_branch_from_nibbles(&[0xa, 0xb]).unwrap();
+
+    let proof_node = test_branch_proof_node(&[0xa, 0xb], None);
+
+    let result = merkle.reconcile_branch_proof_node(&proof_node).unwrap();
+    assert_eq!(result, ReconcileResult::NoValue);
+
+    let node = merkle.get_node_from_nibbles(&[0xa, 0xb]).unwrap().unwrap();
+    let branch = node.as_branch().expect("expected branch node");
+    assert_eq!(branch.value.as_deref(), Some([3u8].as_slice()));
```
alarso16, line 353: Why don't you overwrite the merkle with `None`?

firewood/src/merkle/mod.rs
```diff
@@ -1306,6 +1306,18 @@ impl<K: MutableKind, S: ReadableStorage> Merkle<NodeStore<Mutable<K>, S>> {
     }
 }
 
+/// The outcome of reconciling a branch proof node against the in-memory trie.
+#[cfg(test)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum ReconcileResult {
```
alarso16, line 1312: Is this type only used for tests? I don't see it in the resulting PRs

firewood/src/merkle/mod.rs
rkuris: The [matches! macro](https://doc.rust-lang.org/std/macro.matches.html) checks to see if `proof_node.value_digest` is Some, and the contents of the Some is type `ValueDigest::Value` instead of one of the other `ValueDigest` types, or None.

firewood/src/merkle/tests/mod.rs
rkuris: Tests that tend to span across several functions (not unit tests) are typically downstream of the actual code. If the test is just testing a single function, then you're right in that it probably belongs closer to the code it is testing.

You can think of these more like functional tests instead of unit tests.

firewood/src/merkle/tests/mod.rs
rkuris: This is testing `reconcile_branch_proof_node` semantics, which is documented as follows:

```
This helper never overwrites an existing branch value. It only:

creates missing branch structure,
inserts a value when the branch exists without one, and
errors if an existing value conflicts with the proof.
```

I think that explains why it didn't overwrite the branch value, and this test verifies it.

firewood/src/merkle/mod.rs
rkuris: It's the type returned from reconcile_branch_proof_node, which currently is not reachable outside of tests. Without the `#[cfg(test)]` attribute you get warnings that this code is not actually used in the production code.

The annotation is removed once the functions are actually used in a later PR.

firewood/src/merkle/mod.rs
rkuris: Fixed in 6067c4e991c2548ddea0621530542dd28d29a84a

PR #1723, opened 2026-02-26
chore: Update pull request template to include breaking changes

demosdemon:
## Why this should be merged

By adding this section, the idea is that both pull request authors and reviewers would be more cognizant of breaking changes by identifying which packages would break.

## How this works

This adds a section to the PR template to select which package, if any, includes breaking changes.

## How this was tested

n/a

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,6 +8,6 @@
 
 - [ ] firewood
 - [ ] firewood-storage
-- [ ] firewood-ffi
-- [ ] firewood-go
+- [ ] firewood-ffi (C api)
+- [ ] firewood-go (Go api)
 - [ ] fwdctl


.github/pull_request_template.md
```diff
@@ -3,3 +3,11 @@
 ## How this works
 
 ## How this was tested
+
+## Breaking Changes
+
+- [ ] firewood
+- [ ] firewood-storage
+- [ ] firewood-ffi
+- [ ] firewood-go
```
RodrigoVillar: Hmm what's the difference between `firewood-ffi` and `firewood-go`?

demosdemon: `firewood-ffi` is the rust package whereas `firewood-go` is the go wrapper around the rust ffi api. Changing the `firewood-ffi` api in a breaking way doesn't necessarily break the `firewood-go` api.

RodrigoVillar: Ah understood.

alarso16 (approved):
I think this is a good non-technical fix for the easily forgetting the ! in the title

PR #1725, opened 2026-02-27
feat(reconstruction): 1/7 Scaffolding for Reconstructible

rkuris:
Scaffolding for linear reconstruction, without enabling full reconstruction behavior yet.

## Why this should be merged

Establishes the API for reconstruction so follow-up commits can implement behavior incrementally.
Adds a dedicated reconstructed view type that is explicit in the type system, instead of overloading proposal or committed flows.
Keeps reconstruction as a separate lifecycle from commit/persist paths.

## How this works

Added new trait Reconstructible to the public API
Added a user-visible Reconstructed db view wrapper and DbView implementation
Added a new storage-level Reconstructed nodestore kind and docs describing its linear flow
Exported Reconstructed from storage public exports
Added NodeReader support for reconstructed nodestores

## How this was tested

No new behavioral tests in this scaffolding PR.
Existing behavior is unchanged for committed/proposal workflows.
Follow-up PRs will add reconstruction behavior tests once TODO paths are implemented.

## Breaking Changes

None

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -21,7 +21,8 @@ use crate::manager::{ConfigManager, RevisionManager, RevisionManagerConfig};
 use firewood_metrics::firewood_counter;
 use firewood_storage::{
     CheckOpt, CheckerReport, Committed, FileBacked, FileIoError, HashedNodeReader,
-    ImmutableProposal, NodeHashAlgorithm, NodeStore, Parentable, ReadableStorage, TrieReader,
+    ImmutableProposal, NodeHashAlgorithm, NodeStore, Parentable, ReadableStorage,
+    Reconstructed as StorageReconstructed, TrieReader,
 };
 use nonzero_ext::nonzero;
 use std::io::Write;
@@ -54,9 +55,9 @@ impl std::fmt::Debug for DbMetrics {
     }
 }
 
-impl<P: Parentable, S: ReadableStorage> api::DbView for NodeStore<P, S>
+impl<P, S: ReadableStorage> api::DbView for NodeStore<P, S>
 where
-    NodeStore<P, S>: TrieReader,
+    NodeStore<P, S>: TrieReader + HashedNodeReader,
 {
     type Iter<'view>
         = MerkleKeyValueIter<'view, Self>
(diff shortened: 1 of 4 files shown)


firewood/src/v2/api.rs
```diff
@@ -373,6 +373,57 @@ pub trait DbView {
...
+/// A reconstructable database view.
+///
+/// This trait models linear reconstruction by consuming `self` and returning
+/// a new reconstructed view.
+pub trait Reconstructible: DbView {
+    /// The reconstructed output type.
+    type Reconstructed: DbView + Reconstructible<Reconstructed = Self::Reconstructed>;
+
+    /// Reconstruct a new view from this one by applying `data`.
+    #[expect(clippy::missing_errors_doc)]
```
rkuris, line 421: This gets filled in later

firewood/src/db.rs
```diff
@@ -54,9 +55,9 @@ impl std::fmt::Debug for DbMetrics {
     }
 }
 
-impl<P: Parentable, S: ReadableStorage> api::DbView for NodeStore<P, S>
+impl<P, S: ReadableStorage> api::DbView for NodeStore<P, S>
```
rkuris, line 58: Arguably this bound was incorrect before. We only really care if we have a HashedNodeReader.

firewood/src/v2/api.rs
```diff
@@ -373,6 +373,57 @@ pub trait DbView {
...
 }
 
+impl<T: DbView + ?Sized> DbView for &T {
+    type Iter<'view>
+        = T::Iter<'view>
+    where
+        Self: 'view;
+
+    fn root_hash(&self) -> Option<HashKey> {
+        (*self).root_hash()
```
demosdemon, line 383: `*self` is ambiguous and potentially not what you want. I'm actually surprised there isn't a recursion warning/error. Typically, these deref forwarders are implemented using `**self` because `self` is actually `&&T` therefore `(*self)` is calling `root_hash` on `&T` which is this implementation.

```suggestion
        (**self).root_hash()
```

(example: https://github.com/ava-labs/firewood/blob/c443887dcebbaf46a1f8dcd59b61210e29d797ed/ffi/src/metrics.rs#L27-L31)

rkuris: I was actually able to completely remove this. I'm going to close this PR and recreate it with some slightly simpler implementations.

rkuris:
Closed in favor of another stack

PR #1726, opened 2026-02-27
feat(reconstruction): 3/7 Add mutable constructors

rkuris:
## Why this should be merged

* Replaces reconstructible scaffolding TODO paths with a working linear reconstruction path.
* Introduces an explicit constructor for reconstruction that avoids overloading proposal-parent semantics.
* Provides a concrete bridge from committed/reconstructed views into mutable Merkle application and back to reconstructed output.

## How this works

* Adds a reconstruction-specific mutable constructor: NodeStore::new_for_reconstruction
* Implements reconstruction execution in db.rs via reconstruct_with_parent(...)
* Routes both reconstruct entrypoints to the shared helper &NodeStore<Committed, FileBacked>::reconstruct(...)]
* Applies BatchOp::{Put, Delete, DeleteRange} onto a mutable Merkle built from new_for_reconstruction
* Converts the resulting mutable nodestore back into a reconstructed view

## How this was tested

* Mostly type changes which are caught by the compiler
* No new dedicated behavioral tests are included in this commit, will be in followups

## Breaking Changes

None

### DIFF
--- a/ffi/src/reconstructed.rs
+++ b/ffi/src/reconstructed.rs
@@ -1,0 +1,101 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use firewood::v2::api::{
+    self, BoxKeyValueIter, DbView, HashKey, IntoBatchIter, Reconstructible as _,
+};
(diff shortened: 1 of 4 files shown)


rkuris:

PR #1727, opened 2026-02-27
feat(reconstruction): 4/7 Add mutable conversions

rkuris:
## Why this should be merged

* Implements the core reconstructed-to-mutable conversion path so reconstruction chains can reuse in-memory roots efficiently.
* Adds reconstruction behavior coverage so follow-up work can iterate on performance and correctness with safety nets.
* Improves node resolution behavior for reconstructed views by enabling move-if-unique semantics.

## How this works

* Add reconstruct_with_reconstructed, which converts Reconstructed → MutableProposal via TryFrom and applies the batch.
* Update Reconstructible::reconstruct on Reconstructed to use the new conversion path.
* Add TryFrom helpers for reconstructed nodes that try to move the root node when uniquely owned and clone otherwise.
* Extend node persistence helpers so MaybePersistedNode can resolve into a concrete Node efficiently.
* Add reconstruction chain tests and bring Reconstructible into scope in the DB tests.

## How this was tested

New test_reconstruct_reads_and_chains in the DB test suite.

## Breaking Changes

None

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -371,6 +371,7 @@ impl Db {
         batch: impl IntoBatchIter,
     ) -> Result<Reconstructed, api::Error>
     where
+        P: Parentable,
         NodeStore<P, FileBacked>: TrieReader + HashedNodeReader,
     {
         reconstruct_with_parent(parent, batch, self.manager.threadpool())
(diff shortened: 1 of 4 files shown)


rkuris:

PR #1728, opened 2026-02-27
feat(reconstruction): 5/7 Add ffi interface

rkuris:
## Why this should be merged

* Adds the Go FFI surface for reconstruction so Go callers can access the new reconstructed view and its lifecycle safely.
* Introduces reconstructed handle semantics (drop/invalidated-after-reconstruct) consistent with the Rust API.

## How this works

This mostly follows the design document for the ffi calls. Additional work was done on the Reconstructed type to make it reusable and mutable from golang's perspective, which required an additional api (fwd_reconstructed_root_hash) for lazy hash calculations.

* Adds a dedicated ReconstructedHandle in the Rust FFI with DbView impl and a reconstruct helper.
* Adds C-FFI APIs for reconstructed view access (fwd_get_from_reconstructed, fwd_iter_on_reconstructed, fwd_reconstructed_dump) plus a new fwd_reconstructed_root_hash.
* ReconstructedResult only carries the handle (no hash), and exposes reconstructed APIs through the FFI module.
* Updates Go Reconstructed to lazily compute and memoize root hash
* Reconstruct method mutates the receiver in place to point to a new handle.

## How this was tested

Added coverage and benchmarks for reconstruction flows from Go to validate behavior and concurrency.

At this point, the benchmark reports about 900ms/op. Further optimizations to the hash code are made in the next diff, which reduce the benchmark time down to about 350ms/op (an op being applying ~2048 reconstructed revisions of 100 k/v pairs).

## Breaking Changes

None

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -56,7 +56,7 @@ benchmark/            # Performance benchmarking suite
 ## Important Terminology
 
 - **Revision**: Historical point-in-time state of the trie
-- **View**: Interface to read from a Revision or Proposal
+- **View**: Interface to read from a Revision, Proposal, or Reconstructed view
 - **Node**: Portion of a trie that can point to other nodes and/or contain Key/Value pairs
 - **Hash/Root Hash**: Merkle hash for a node/root node
 - **Proposal**: Consists of base Root Hash and Batch, not yet committed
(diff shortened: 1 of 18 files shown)


rkuris:

Issue #1737, opened 2026-03-04
Chaos test unable to reopen database when `DeferredPersistenceCommitCount > 1`


RodrigoVillar:
This flake is observed with the following PR: https://github.com/ava-labs/avalanchego/pull/5022

Below are two instances of the chaos test failing: 
 
- https://github.com/ava-labs/avalanchego/actions/runs/22681775911/job/65753974195
- https://github.com/ava-labs/avalanchego/actions/runs/22579719448/job/65408176236

RodrigoVillar:
Call Stack:

```
  1. DatabaseHandle::new (ffi/src/handle.rs:160) — builds DbConfig from FFI args, calls:
  2. Db::new (firewood/src/db.rs:183) — builds ConfigManager, calls:
  3. RevisionManager::new (firewood/src/manager.rs:~188) — opens FileBacked storage, reads the header, then calls:
    - NodeStoreHeader::read_from_storage (storage/src/nodestore/header.rs:~320) — reads the persisted header from offset 0, which contains the root_address
    - Then calls:
  4. NodeStore::open (storage/src/nodestore/mod.rs:112) — if the header has a root_address but no root_hash (line 122-128), it needs to recompute the hash by reading the root node:
  5. NodeStore::read_node_from_disk (storage/src/nodestore/mod.rs:706) — called with the root_address from the header. Delegates to:
  6. NodeStore::read_node_with_num_bytes_from_disk (storage/src/nodestore/mod.rs:732) — adds 1 to skip the area size index byte (actual_addr = addr + 1), creates a stream from storage, calls:
  7. Node::from_reader (storage/src/node/mod.rs:339) — reads the first byte. If the byte is 0xFF (255), that's the sentinel for a freed area, and it returns:
  Err(Error::other("attempt to read freed area"))
```

RodrigoVillar:
My intuition as to why this bug occurs is as follows:

- We commit revision `R`, which is then persisted
- We continue committing revisions until the number of in-memory revisions is full
- `R` is reaped
- We crash
- On startup, the header points to `R`, which was reaped. 

Here's a test that reproduces the error:

```rust
/// Creates a committed revision with a single leaf node from the given parent.
    fn commit_leaf(
        parent: &NodeStore<impl firewood_storage::Parentable, FileBacked>,
        value: &[u8],
    ) -> NodeStore<firewood_storage::Committed, FileBacked> {
        let mut proposal = NodeStore::new(parent).unwrap();
[...]

PR #1739, opened 2026-03-04
fix: reap after persisting

RodrigoVillar:
## Why this should be merged

Fixes the bug described in #1737 by first persisting revisions before reaping nodestores in deferred persistence.

## How this works

- `reap()` no longer wakes up the background thread
- when `pop` returns in the `event_loop`, we first persist and then reap

## How this was tested

CI + added regression test

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -284,7 +284,10 @@ impl PersistChannel {
         state.permits_available == state.max_permits.get()
     }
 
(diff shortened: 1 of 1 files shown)


RodrigoVillar:
Closing in favor of #1749

Issue #1720, opened 2026-02-25
Clock source mismatch between Go and Rust FFI causes incorrect commit latency metrics


Elvis339:
The performance dashboard shows Rust FFI commit time consistently higher than Go commit time even though Go calls Rust synchronously via CGO (Go >= Rust, always).

This was surfaced without deferred persistence, where both sides register non-zero values and the impossible relationship is visible. With deferred persistence, commits drop below 1ms and Go's integer truncation records them as 0, hiding the mismatch entirely.

### Root Cause

1. Different clock sources where Rust uses `coarsetime::Instant` while Go uses `time.Now()`
2. Go integer truncation: `time.Since(start).Milliseconds()` returns `int64` any commit under 1ms records as 0.

### Background

`coarsetime` was chosen to avoid syscall overhead from `std::time`. Brandon confirmed this isn't a concern `std::time::Instant` uses `clock_gettime(CLOCK_MONOTONIC)` via vDSO, resolving entirely in userspace on any modern Linux kernel.

### Fix

Rust: Swap `coarsetime` to `std::time::Instant` in the `#[histogram_timing]` macro (after #1712 merges).

Go: Change `.Milliseconds()` to `.Nanoseconds()` in `triedb.go` (consistent with AvalancheGo conventions per Stephen), or use float seconds (Prometheus best practice). Same pattern exists in `blockchain.go` chain timers.

Elvis339:
There was concern that removing `coarsetime` might reintroduce syscall overhead. We ran a controlled comparison on a real EC2 instance (`i4i.2xlarge`, Ubuntu) executing blocks 101–250k with the Firewood config. 

## Results

| | coarsetime | std::time |
|---|---|---|
| `clock_gettime` syscalls | 0 | 0 |

`clock_gettime` does not appear in either strace output. The `std::time::Instant` resolves `clock_gettime(CLOCK_MONOTONIC)` via vDSO a kernel function mapped into userspace so it never enters the kernel and is invisible to strace. This confirms what Brandon noted in the Slack thread. 

- https://doc.rust-lang.org/std/time/struct.Instant.html#underlying-system-calls
- https://doc.rust-lang.org/std/time/struct.SystemTime.html#platform-specific-behavior

Brandon:
[...]

Issue #1735, opened 2026-03-04
Metrics should use base units, not ms


rkuris:
We have a number of metrics that do not use base units, but everything should. Let's audit the metrics generated and change them to use base units instead of non-standard ones.

References:
https://www.robustperception.io/who-wants-seconds/
Article by core committer saying everything should be in seconds

https://stackoverflow.com/questions/56121502/changing-the-base-time-unit-displayed-on-prometheus-endpoint#:~:text=1%20Answer&text=You%20are%20describing%20the%20very,regardless%20of%20the%20underlying%20implementation.
Discussion about some problems using ms.

https://prometheus.io/docs/practices/naming/
...SHOULD use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.

PR #1742, opened 2026-03-06
fix!: clock source miss-match between Rust and Go

Elvis339:
## Why this should be merged

coarsetime was introduced to avoid syscall overhead from `std::time`, but @demosdemon confirmed (and we verified empirically) that std::time uses `clock_gettime(CLOCK_MONOTONIC)` via vDSO on modern Linux which is resolving in userspace with no kernel entry. The two are equivalent in cost.

Removing coarsetime fixes the clock source mismatch between Rust `CLOCK_MONOTONIC_COARSE` and Go `CLOCK_MONOTONIC` that was causing the Firewood dashboard to show Rust commit time higher than Go commit time, which is physically impossible since Go wraps Rust synchronously via CGO.

## How this works

Replaces all `coarsetime::Instant` / `coarsetime::Duration` usages with`std::time::Instant` / `std::time::Duration`

## How this was tested

Full methodology in #1720.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [x] firewood-ffi (C api) — `commit_proposal` token closure parameter type changes from `coarsetime::Duration` to `std::time::Duration`
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/benchmark/src/main.rs
+++ b/benchmark/src/main.rs
@@ -55,8 +55,8 @@ struct GlobalOpts {
     batch_size: u64,
     #[arg(short, long, default_value_t = 1000)]
     number_of_batches: u64,
-    #[arg(short, long, default_value_t = NonZeroUsize::new(1500000).expect("is non-zero"))]
-    cache_size: NonZeroUsize,
+    #[arg(short, long, default_value_t = NonZeroUsize::new(192_000_000).expect("is non-zero"))]
+    cache_memory_limit: NonZeroUsize,
     #[arg(short, long, default_value_t = 128)]
     revisions: usize,
     #[cfg(feature = "prometheus")]
@@ -227,9 +227,8 @@ fn main() -> Result<(), Box<dyn Error>> {
     let prometheus_handle = spawn_prometheus_listener(args.global_opts.prometheus_port)
         .expect("failed to spawn prometheus listener");
 
-    #[expect(deprecated)]
     let mgrcfg = RevisionManagerConfig::builder()
-        .node_cache_size(args.global_opts.cache_size)
+        .node_cache_memory_limit(args.global_opts.cache_memory_limit)
         .free_list_cache_size(
             NonZeroUsize::new(4 * args.global_opts.batch_size as usize).expect("batch size > 0"),
         )
--- a/firewood/examples/insert.rs
+++ b/firewood/examples/insert.rs
@@ -31,8 +31,8 @@ struct Args {
     read_verify_percent: u16,
     #[arg(short, long)]
     seed: Option<u64>,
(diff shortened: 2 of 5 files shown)


ffi/src/handle.rs
```diff
@@ -233,11 +233,11 @@ impl DatabaseHandle {
                 handle.commit_proposal()
             });
         let root_hash = root_hash_result?;
-        firewood_increment!(crate::registry::BATCH_MS, elapsed.as_millis());
+        firewood_increment!(crate::registry::BATCH_MS, elapsed.as_millis() as u64);
         firewood_increment!(crate::registry::BATCH_COUNT, 1);
         firewood_record!(
             crate::registry::BATCH_MS_BUCKET,
-            elapsed.as_f64() * 1000.0,
```
RodrigoVillar, line 240: why do we call `as_secs_f64()` when we could just use `as_millis()`?

Elvis339: Good catch, I wanted to keep scope of this PR to only replace coarsetime with std::time. The broader unit change is described here and that's up next for a change: https://github.com/ava-labs/firewood/issues/1735

ffi/src/proposal.rs
```diff
@@ -87,7 +87,7 @@ impl ProposalHandle<'_> {
         // clear the cached view so that it does not hold onto the proposal view
         handle.clear_cached_view();
 
-        firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis());
+        firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis() as u64);
```
rkuris, line 90: nit: No need to fix this, since it's about to change to floating point anyway, just calling it out.

Overflow bugs are nasty. Generally we shouldn't just convert from u128 to u64. and use a saturating or checked conversion:

Saturating: `...as_millis().clamp(u64::MIN as u128, u64::MAX as u128) as u64`
Checked: `...as_millis().try_into().expect("<500m years")`

Cargo.toml
```diff
@@ -70,7 +70,6 @@ aquamarine = "0.6.0"
 bytemuck = { version = "1.25.0", features = ["const_zeroed"] }
 bytemuck_derive = "1.10.2"
 clap = { version = "4.5.60", features = ["derive"] }
-coarsetime = "0.1.37"
```
rkuris, line 73: 🏆

rkuris (approved):

ffi/src/proposal.rs
demosdemon: 2**64 milliseconds is over 584,600,000 Gregorian years. If this takes so long that millis wraps a u64, we have different problems. Handling it here is excessive.

PR #1745, opened 2026-03-07
refactor(reconstructed): 2/7 Extract parallel batch-apply code to separate function

rkuris:
## Why this should be merged

Extracts the parallel batch-apply pipeline into a reusable entry point so other flows can share it without duplicating logic.

No new code, just moves and calls.

## How this works|

Add ParallelMerkle::apply_to_mutable, which runs the existing prepare/split/merge/post-process pipeline on a supplied mutable nodestore.
Update create_proposal to call apply_to_mutable after creating the mutable nodestore, then convert to ImmutableProposal as before.
All logic in create_proposal remains functionally the same; it is just delegated.

## How this was tested

No new tests in this refactor.
Existing proposal paths are unchanged; create_proposal delegates to the new helper.

## Breaking Changes
None

### DIFF
--- a/firewood/src/merkle/parallel.rs
+++ b/firewood/src/merkle/parallel.rs
@@ -64,6 +64,125 @@ pub struct ParallelMerkle {
 }
 
 impl ParallelMerkle {
+    /// Applies a batch of operations to an existing mutable nodestore using the parallel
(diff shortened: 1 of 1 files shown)


rkuris:
Closed in favor of another stack

PR #1749, opened 2026-03-08
fix!: enforce `max_revisions` > `commit_count`

RodrigoVillar:
## Why this should be merged

During review of #1739, it was noted that we could prevent the last persisted revision from being persisted if we set `max_revisions > commit_count`. This is because if `max_revisions > commit_count`, then once the oldest revision is marked for reaping, then there has been at least a newer revision persisted as `commit_count` revisions have been committed.

## How this works

- During initialization, enforces that `max_revisions > commit_count`
- Moves `commit_count` from `ConfigManager` to `RevisionManagerConfig`
    - Having an invariant be linked between two different data structures seems incredibly fragile to me, so I decided to move `commit_count` to reduce any confusion.
- Updates Go + Rust docs to make this new invariant clear 

## How this was tested

CI + added UT

## Breaking Changes

- [x] firewood
- [ ] firewood-storage
- [x] firewood-ffi (C api)
- [x] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -105,7 +105,7 @@ type config struct {
 	// revisions is the maximum number of historical revisions to keep in memory.
 	// If rootStoreDir is set, then any revisions removed from memory will still be kept on disk.
 	// Otherwise, any revisions removed from memory will no longer be kept on disk.
-	// Must be >= 2.
+	// Must be >= 2 and > deferredPersistenceCommitCount.
 	revisions uint
 	// readCacheStrategy is the caching strategy used for the node cache.
 	readCacheStrategy CacheStrategy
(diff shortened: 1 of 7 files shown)


RodrigoVillar:
> I think this is a weird check to enforce just so this bug isn't hit. If we just want to fix it, that's cool. But we should probably find a better long term fix

@alarso16 This is definitely not a long-term fix (we will eventually revert this change). @rkuris can give more insight here but the true fix would be to refactor how we manage our free-lists/reaping logic. This is a (performance) feature of its own and I would prefer we patch this bug ASAP.

ffi/firewood.go
```diff
@@ -115,6 +115,7 @@ type config struct {
 	expensiveMetricsEnabled bool
 	// deferredPersistenceCommitCount determines the maximum number of unpersisted
 	// revisions that can exist at a given time.
+	// Note: revisions must be > deferredPersistenceCommitCount
```
alarso16, line 118: Will you put this comment on `WithPersistedCount` as well? This won't be exported in the godoc

RodrigoVillar: Done.

Issue #1717, opened 2026-02-25
Mermaid graphs don't render in documentation


RodrigoVillar:
Going to the documentation for the `persist_worker` module, we see:

<img width="793" height="857" alt="Image" src="https://github.com/user-attachments/assets/308257d6-cd61-4e78-8f08-28904e7a4713" />

There's some crates we can use for this:

https://crates.io/crates/aquamarine: straightfoward (just one macro) but doesn't work on docstrings
https://crates.io/crates/simple-mermaid/0.2.0: works on docstrings (I believe) but mermaid graph has to live in a separate file

demosdemon:
We do use aquamarine, but it requires decorating the diagram with the appropriate attribute. Example:
https://github.com/ava-labs/firewood/blob/94641ecefaf724dd76e8c3d839994dbb01d368a1/storage/src/checker/mod.rs#L771-L782

demosdemon:
to be fair, that is the only example in the code and it doesn't render because the it's buried deep in a `test` build. `cargo doc` doesn't render `test` docs by default.

PR #1753, opened 2026-03-09
chore(docs): render mermaid graphs with aquamarine

demosdemon:
## Why this should be merged

Closes: #1717

## How this works

We already had aquamarine imported and use it for one graph. However, other graphs were missing the annotation.

Using this annotation on the mod-level comment is not yet supported by `cargo doc`. However, the diagram where this was a problem was incorrect and stale, so I removed it instead of relocating.

This however means we must remove `--no-deps` from the `cargo doc` invocation. This is because `--no-deps` will cause the build of the documentation to not include the necessary static assets for rendering.

## How this was tested

image

## Breaking Changes

n/a

### DIFF
--- a/firewood/src/persist_worker.rs
+++ b/firewood/src/persist_worker.rs
@@ -70,6 +70,7 @@ pub enum PersistError {
 ///
 /// Below is an example when `commit_count` is set to 10:
 ///
+#[cfg_attr(doc, aquamarine::aquamarine)]
 /// ```mermaid
 /// sequenceDiagram
 ///     participant Caller
@@ -106,7 +107,6 @@ pub enum PersistError {
 ///     BG->>Disk: persist last committed revision
 ///     Note right of Disk: Latest committed revision is persisted
 /// ```
-#[cfg_attr(doc, aquamarine::aquamarine)]
 #[derive(Debug)]
 pub(crate) struct PersistWorker {
     /// The background thread responsible for persisting commits async.
--- a/storage/src/nodestore/mod.rs
+++ b/storage/src/nodestore/mod.rs
@@ -61,28 +61,6 @@ pub use primitives::{AreaIndex, LinearAddress};
 // Re-export types from header module
 pub use header::NodeStoreHeader;
 
-/// The [`NodeStore`] handles the serialization of nodes and
-/// free space management of nodes in the page store. It lays out the format
-/// of the [`PageStore`]. More specifically, it places a [`FileIdentifyingMagic`]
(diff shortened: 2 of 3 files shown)


firewood/src/persist_worker.rs
```diff
@@ -70,6 +70,7 @@ pub enum PersistError {
 ///
 /// Below is an example when `commit_count` is set to 10:
 ///
+#[cfg_attr(doc, aquamarine::aquamarine)]
```
RodrigoVillar, line 73: Is this meant to be repeated twice? Ref line 110

demosdemon: No, it needs to be above the triple-ticks. I missed that, but that is why this one was not rendering.

demosdemon:
> There's a high cost to not running `--no-deps`, particularly when the agent is doing it's checkin tests. Let's see if the benefits outweigh the cost.

That's fair. We can add the `--no-deps` back to agent instructions as the only negative consequence is the diagrams will fail to render when viewing the docs locally.

firewood/src/persist_worker.rs
RodrigoVillar: I'm not sure I follow - looking at our current docs, this mermaid graph is being rendered: https://ava-labs.github.io/firewood/firewood/persist_worker/struct.PersistWorker.html#sequence-diagram

demosdemon: Ah, I thought the attribute needed to decorate the diagram. The docs for aquamarine always does that but it's not an actual requirement.

storage/src/nodestore/mod.rs
```diff
@@ -61,28 +61,6 @@ pub use primitives::{AreaIndex, LinearAddress};
...
-/// flowchart TD
-/// subgraph subgraph["Committed Revisions"]
-/// L("Latest Nodestore&lt;Committed, S&gt;") --- |...|O("Oldest NodeStore&lt;Committed, S&gt;")
-/// end
-/// O --> E("Expire")
-/// L --> |start propose|M("NodeStore&lt;ProposedMutable, S&gt;")
-/// M --> |finish propose + hash|I("NodeStore&lt;ProposedImmutable, S&gt;")
-/// I --> |commit|N("New commit NodeStore&lt;Committed, S&gt;")
-/// style E color:#FFFFFF, fill:#AA00FF, stroke:#AA00FF
-/// ```
```
demosdemon, line 85: TBH, `cargo doc` or even `rustc` in general should error when trying annotate a `use` statement with a doc comment because there's nothing to attach the comment to.

PR #1765, opened 2026-03-11
feat(reconstruct): Framework for reconstruction 

rkuris:
## Why this should be merged

This change introduces a new nodestore type, a Reconstructed. The guts for this change are in the next PR, this just introduces all the types and conversions, the details of which are implemented in the next PR.

## How this works

STACKED: There are two commits, you may wish to review only the most recent commit.

Reconstructed also implements DbView. Previously, nodestores must be
Parentable to be a DbView, which included Historical and Proposed
nodestores, but not Mutable nodestores. With the addition of Reconstructed,
the bounds for DbView changed to not require Parentable, since
Reconstructed nodestores are not Parentable. Instead, they must implement
HashedNodeReader, which is implemented by all nodestores except Mutable, so
the bounds on DbView were changed accordingly.

This PR also introduces the conversion functions:
into_reconstruction_child converts a root node from a shared node into
a mutable root node by copying the node.

Additionally, the Reconstructible trait is defined.

## How this was tested

No significant tests, except for the compiler type checking.

## Breaking Changes

None

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -56,7 +56,7 @@ benchmark/            # Performance benchmarking suite
 ## Important Terminology
 
 - **Revision**: Historical point-in-time state of the trie
-- **View**: Interface to read from a Revision or Proposal
+- **View**: Interface to read from a Revision, Proposal, or Reconstructed view
 - **Node**: Portion of a trie that can point to other nodes and/or contain Key/Value pairs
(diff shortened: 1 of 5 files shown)


rkuris:
@RodrigoVillar pointed out that we should be using the type system better here. Will replace this with some additional type work before this.

PR #1782, opened 2026-03-12
chore: add additional metrics for `v0.3.0`

RodrigoVillar:
## Why this should be merged

Adds exploratory metrics prior to `v0.3.0` related to deferred persistence and the upcoming delete list feature.

## How this works

- Adds `PERMITS_AVAILABLE` and `MAX_PERMITS` metrics for tracking the number of outstanding permits and the number of maximum permits, respectively.
- Adds `COMMIT_BLOCKED` metric for counting how many times we block.
- Adds metric for tracking how it takes to reap the deleted nodes of a revision.
- Adds metric for tracking the length of the deleted list for each committed revision.

## How this was tested

CI + Grafana dashboards

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -24,7 +24,7 @@ use crate::merkle::Merkle;
 use crate::merkle::parallel::ParallelMerkle;
 use crate::persist_worker::{PersistError, PersistWorker};
 use crate::root_store::RootStore;
-use firewood_metrics::{firewood_increment, firewood_record, firewood_set};
+use firewood_metrics::{firewood_increment, firewood_set};
 pub use firewood_storage::CacheReadStrategy;
 use firewood_storage::{
     BranchNode, Committed, FileBacked, FileIoError, HashedNodeReader, ImmutableProposal, Mutable,
@@ -300,12 +300,7 @@ impl RevisionManager {
 
         let committed = proposal.as_committed();
 
-        #[expect(
-            clippy::cast_precision_loss,
-            reason = "deleted list length will never exceed 2^52"
-        )]
-        let deleted_list_len = committed.deleted_len() as f64;
-        firewood_record!(crate::registry::DELETED_LIST_LEN, deleted_list_len);
+        firewood_set!(crate::registry::DELETED_LIST_LEN, committed.deleted_len());
 
         // 3. Revision reaping
         // When we exceed max_revisions, remove the oldest revision from memory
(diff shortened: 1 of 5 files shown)


firewood/src/manager.rs
```diff
@@ -300,6 +300,10 @@ impl RevisionManager {
 
         let committed = proposal.as_committed();
 
+        #[expect(clippy::cast_precision_loss)]
```
rkuris: nit: This could be eliminated if the macro does the work -- clippy doesn't report problems on expanded macro code.

You'd just have to change metrics/src/lib.rs's firewood_record macro to add "as f64" everywhere value is found, then this simplifies to just `firewood_record!(..., committed.deleted_len())`

firewood/src/persist_worker.rs
```diff
@@ -416,6 +420,17 @@ struct PersistChannelState {
     latest_committed: Option<CommittedRevision>,
 }
 
+impl PersistChannelState {
+    /// Emits the current permit gauges.
+    fn emit_permits(&self) {
+        #[expect(clippy::cast_precision_loss)]
```
rkuris: gauge already has the optimization I mentioned before, so just drop the conversion here and it will work.

firewood/src/manager.rs
RodrigoVillar: Going to defer this to a separate issue as this would also clean up code in the `storage` crate.

firewood/src/persist_worker.rs
RodrigoVillar: Switched to using `firewood_set!()`, which does this for us.

firewood/src/persist_worker.rs
```diff
@@ -416,6 +420,14 @@ struct PersistChannelState {
     latest_committed: Option<CommittedRevision>,
 }
 
+impl PersistChannelState {
+    /// Emits the current permit gauges.
+    fn emit_permits(&self) {
+        firewood_set!(crate::registry::PERMITS_AVAILABLE, self.permits_available);
+        firewood_set!(crate::registry::MAX_PERMITS, self.max_permits.get());
```
Elvis339: Nit: what do you think about moving max_permits from a function that is always updating? The benefit would that we would not pay for ingestion for a constant that never changes.

RodrigoVillar: Do you mean having a function that calls `firewood_set!(crate::registry::MAX_PERMITS, self.max_permits.get());` just once (e.g. on startup)?

Elvis339: Yes.

PR #1785, opened 2026-03-13
feat(reconstruct): Framework for Reconstruction

rkuris:
## Why this should be merged

Framework for reconstruction, no implementations of some key functions yet (that's in the next PR), just the new types.

## How this works

This change introduces new types:
 - Reconstructed -- a new NodeStore<T> type, which implements DbView
 - ReconstructedRoot -- the type of a Reconstructed root
 - ReconstructedView -- Holds a reference to a Reconstructed nodestore

And some new traits:
 - Reconstructible -- A trait that allows reconstruct() calls
 - ReconstructionSource -- NodeStore<T> types that implement
   Reconstructible

A mermaid diagram that describes the lifecycle has been added to
NodeStore's doc comment and should be reviewed for context while
reviewing this diff.

Previously, nodestores must be Parentable to be a DbView, which included
Committed and Proposed nodestores, but not Mutable nodestores. With the
addition of Reconstructed, the bounds for DbView changed to not require
Parentable, since Reconstructed nodestores are not Parentable. Instead,
they must implement HashedNodeReader, which is implemented by all nodestores
except Mutable, so the bounds on DbView were changed accordingly.

Also included:
 - new_for_reconstruction converts a root node from a shared node into
   a mutable root node by copying the node.
 - into_reconstruction_child which creates a Mutable<Recon> from a
   ReconstructionSource.
 - Glossary of README.md and AGENTS.md updated

## How this was tested

CI

## Breaking Changes

None

### DIFF
--- a/storage/src/nodestore/mod.rs
+++ b/storage/src/nodestore/mod.rs
@@ -637,7 +637,7 @@ pub struct NodeStore<T, S> {
 /// The root hash is computed lazily on the first call to
 /// [`HashedNodeReader::root_hash`] and cached in a [`OnceLock`] for
 /// lock-free access on subsequent reads.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Reconstructed {
     /// The root node, wrapped in a [`SharedNode`] at construction.
     node: Option<SharedNode>,
@@ -724,20 +724,14 @@ impl<S: ReadableStorage> From<NodeStore<Mutable<Recon>, S>> for NodeStore<Recons
     }
 }
 
-/// Implement a very expensive clone for `NodeStore<Reconstructed, S>` that recursively clones
-/// the entire trie in memory. This is a bit cheaper if the nodestore has been hashed, since
-/// only the [`SharedNode`] arc is cloned rather than the full node tree.
+/// Implement clone for `NodeStore<Reconstructed, S>`.
+///
+/// This clones the [`SharedNode`] arc (cheap ref-count bump) and the
+/// [`OnceLock`] hash (cloned if already computed, empty otherwise).
 impl<S> Clone for NodeStore<Reconstructed, S> {
     fn clone(&self) -> Self {
-        let hash = OnceLock::new();
-        if let Some(h) = self.kind.hash.get() {
-            let _ = hash.set(h.clone());
-        }
         NodeStore {
-            kind: Reconstructed {
-                node: self.kind.node.clone(),
-                hash,
-            },
+            kind: self.kind.clone(),
             storage: self.storage.clone(),
         }
     }
@@ -848,7 +842,7 @@ impl<T, S: ReadableStorage> NodeReader for NodeStore<Mutable<T>, S> {
 
 impl<S: ReadableStorage> NodeReader for NodeStore<Reconstructed, S> {
     fn read_node(&self, addr: LinearAddress) -> Result<SharedNode, FileIoError> {
-        self.read_node_from_disk(addr, "recon-read")
+        self.read_node_from_disk(addr, ReadableNodeMode::ReconRead)
     }
 }
 
--- a/storage/src/linear/mod.rs
+++ b/storage/src/linear/mod.rs
@@ -119,6 +119,7 @@ impl Deref for FileIoError {
 pub enum ReadableNodeMode {
     Open,
     Read,
+    ReconRead,
     Write,
 }
 
@@ -127,6 +128,7 @@ impl ReadableNodeMode {
         match self {
             ReadableNodeMode::Open => "open",
             ReadableNodeMode::Read => "read",
+            ReadableNodeMode::ReconRead => "recon-read",
             ReadableNodeMode::Write => "write",
         }
     }


rkuris:
Extracted from mermaid diagram in generated documentation

Screenshot 2026-03-13 at 7 07 44 AM

firewood/src/api.rs
```diff
@@ -497,6 +499,28 @@ where
     }
 }
 
+/// A reconstructable database view.
```
demosdemon: nit: typo 

```suggestion
/// A reconstructible database view.
```

storage/src/nodestore/mod.rs
```diff
@@ -338,6 +345,61 @@ impl<K: MutableKind, S: ReadableStorage> NodeStore<Mutable<K>, S> {
...
+            Some(root)
+        } else {
+            None
+        };
+
+        Ok(NodeStore {
+            kind: Mutable { root, inner: Recon },
+            storage: self.storage.clone(),
+        })
+    }
```
demosdemon, line 390: This has the same body as `new_for_reconstruction`. One should call the other.

storage/src/nodestore/mod.rs
```diff
@@ -734,6 +966,54 @@ where
...
+        // Reconstructed views are never persisted.
+        None
+    }
+
+    // lazily compute the hash
+    // We don't use the parallel hash calculations here, again because
+    // reconstructed views don't benefit as much from spinning up multiple
+    // threads.
+    fn root_hash(&self) -> Option<TrieHash> {
+        let mut guard = self.kind.root.lock();
```
demosdemon: Will this ever be set back to `ReconstructedRoot::Node`?

If yes, should this use a `RWLock` instead of a `Mutex`?
If no, this should use a `OnceLock` instead of a `Mutext`.

README.md
```diff
@@ -46,7 +46,8 @@ as well as carefully managing the free list during the creation and expiration o
 - `Revision` - A historical point-in-time state/version of the trie. This
   represents the entire trie, including all `Key`/`Value`s at that point
   in time, and all `Node`s.
-- `View` - This is the interface to read from a `Revision` or a `Proposal`.
+- `View` - This is the interface to read from a `Revision`, `Proposal`, or
+  `Reconstructed` view.
```
RodrigoVillar:
```suggestion
- `View` - This is the interface to read from a `Revision`, `Proposal`, or
  a `Reconstructed.
```

nit: using `view` within the definition of `View` is circular

AGENTS.md
```diff
@@ -56,7 +56,7 @@ benchmark/            # Performance benchmarking suite
 ## Important Terminology
 
 - **Revision**: Historical point-in-time state of the trie
-- **View**: Interface to read from a Revision or Proposal
+- **View**: Interface to read from a Revision, Proposal, or Reconstructed view
```
RodrigoVillar: Same comment regarding using `view` in the definition of a `View`

storage/src/nodestore/mod.rs
rkuris: Great suggestion.

A OnceLock is a better choice here generally. An interesting bug was discovered while doing this too, in the failure path while computing the hash. If the hash_helper fails, we leave the root None.

Fixing this greatly simplifies the approach here too. We now wrap all roots in Arc<Node> but we end out doing that anyway in a number of places.

storage/src/nodestore/mod.rs
```diff
@@ -734,6 +966,54 @@ where
...
+        let root = guard.take();
+
+        let (hash, root) = match root {
+            None => (None, None),
+            Some(ReconstructedRoot::Node(node)) => {
+                #[cfg(feature = "ethhash")]
```
RodrigoVillar: (q/no-need-for-action): I wonder what the implications are of discarding the error (I understand that we're limited in returning an error with `HashedNodeReader`)

storage/src/nodestore/mod.rs
```diff
@@ -761,20 +716,27 @@ impl<S: ReadableStorage> From<NodeStore<Mutable<Recon>, S>> for NodeStore<Recons
...
 /// Implement a very expensive clone for `NodeStore<Reconstructed, S>` that recursively clones
-/// the entire trie in memory. This is a bit cheaper if the nodestore has been hashed.
+/// the entire trie in memory. This is a bit cheaper if the nodestore has been hashed, since
+/// only the [`SharedNode`] arc is cloned rather than the full node tree.
 impl<S> Clone for NodeStore<Reconstructed, S> {
     fn clone(&self) -> Self {
+        let hash = OnceLock::new();
+        if let Some(h) = self.kind.hash.get() {
+            let _ = hash.set(h.clone());
+        }
```
demosdemon: nit: `oncelock` on clone will clone if inhabited and skip if not

https://doc.rust-lang.org/stable/src/std/sync/once_lock.rs.html#628-637

storage/src/nodestore/mod.rs
rkuris: Yes, this isn't good. If an error happens, we end out mapping it to Option<TrieHash>. It's a limitation of the api trait. I'll open a new issue for this.

rkuris: Created https://github.com/ava-labs/firewood/issues/1791

storage/src/nodestore/mod.rs
```diff
@@ -642,34 +840,15 @@ impl<S: ReadableStorage> TryFrom<NodeStore<Mutable<Propose>, S>>
...
-                inner: Propose {
-                    deleted: Vec::default(),
-                    parent,
-                },
-            },
-            storage: val.storage,
-        }
+impl<T, S: ReadableStorage> NodeReader for NodeStore<Mutable<T>, S> {
+    fn read_node(&self, addr: LinearAddress) -> Result<SharedNode, FileIoError> {
+        self.read_node_from_disk(addr, "write")
```
RodrigoVillar:
```suggestion
        self.read_node_from_disk(addr, "read")
```

Isn't this supposed to be `read`?

README.md
rkuris: fixed in da67d9875ced819352aa65fe0ab38815090aa432

storage/src/nodestore/mod.rs
rkuris: I think it's a read for writing. You've got another PR in this area so let's address it there.

storage/src/nodestore/mod.rs
rkuris: https://github.com/ava-labs/firewood/pull/1785/changes/df8241da84778e015b108fa4b729302d4586fec0 fixes this by utilizing it, and fixes the associated comments too.

PR #1786, opened 2026-03-13
feat(reconstruct): Support Reconstructible revisions

rkuris:
## Why this should be merged

Completes the rust implementation for Reconstructible types.

## How this works

Implements the methods needed to fully implement reconstruction,
plus some integration tests.
 - Reconstructible implementation for ReconstructedView
 - RevisionManager::apply_batch_recon for applying batches to
   Reconstructed NodeStores

The trickiest code is in try_into_node which avoids cloning whenever possible.

A small optimization which avoids making an extra Arc clone and then copying it
was added to both proposals and Reconstructed types using unwrap_or_clone.

## How this was tested

CI
Benchmarks are in the next PR with some numbers (they look great)

## Breaking Changes

None

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -583,6 +583,9 @@ mod test {
         }
     }
 
+    /// Reconstruction always uses the serial path (`apply_batch_recon`), but the base
+    /// revision may have been built with parallel or serial proposals. This test confirms
+    /// both produce identical reconstruction results.
     #[test]
     fn test_reconstruct_deterministic() {
         let parallel_db = TestDb::new_with_config(


RodrigoVillar:
Could you rebase this branch on top of `rkuris-reconstruct-3`?

storage/src/nodestore/mod.rs
```diff
@@ -712,6 +716,23 @@ impl<S: ReadableStorage> From<NodeStore<Reconstructed, S>> for NodeStore<Mutable
...
+        // Fallback cost: when shared, we must clone the `NodeStore<Reconstructed, S>`, which
+        // clones the root `SharedNode` arc and then `Arc::unwrap_or_clone` extracts the `Node`.
+        // This shared-Arc fallback does not match our known reconstruction use cases, where the
+        // previous reconstructed handle is typically consumed before building the next one.
+        // We do this to avoid forcing callers to guarantee uniqueness while still exploiting the
+        // cheaper move path whenever possible.
+        Self::from(Arc::unwrap_or_clone(val))
+    }
+}
+
```
RodrigoVillar, line 741: nit: could we move this `impl` block back to where it was?

storage/src/nodestore/mod.rs
```diff
@@ -901,7 +904,6 @@ impl<S: ReadableStorage> RootReader for NodeStore<Reconstructed, S> {
         self.root_node().map(Into::into)
     }
 }
-
```
RodrigoVillar, line 904: nit: could we revert this? Likewise with all other instances in this file.

storage/src/nodestore/mod.rs
```diff
@@ -700,6 +700,10 @@ pub struct Mutable<Kind> {
     pub(crate) inner: Kind,
 }
 
+// given root from either a Committed or Reconstructed,
+// create a mutable root node for a new Reconstructed
+// with this root.
+// For reconstruct on reconstruct, this avoids cloning
```
RodrigoVillar:
```suggestion
/// Given root from either a Committed or Reconstructed,
/// create a mutable root node for a new Reconstructed
/// with this root.
/// For reconstruct on reconstruct, this avoids cloning
```

nit: feel free to push back on the `///` or `given => Given`

rkuris:
> Could you rebase this branch on top of `rkuris-reconstruct-3`?

It's rebased on main, or will be shortly.

firewood/src/db.rs
```diff
@@ -546,6 +664,47 @@ mod test {
...
+                historical.as_ref(),
+                vec![BatchOp::Put {
+                    key: b"base",
+                    value: b"v1",
+                }],
+            )
+            .unwrap();
+        let reconstructed_hash_1 = reconstructed.root_hash();
+        let reconstructed_hash_2 = reconstructed.root_hash();
+        assert_eq!(reconstructed_hash_1, reconstructed_hash_2);
```
RodrigoVillar: nit: not sure what we get from computing the `root_hash` twice here. I know this is to test that we `computed_root_hash == cached_root_hash` but I think we already established test coverage for that in the last PR.

storage/src/nodestore/mod.rs
```diff
@@ -1354,7 +1358,6 @@ mod tests {
 
         Ok(())
     }
-
```
RodrigoVillar, line 1357: Please revert this diff.

storage/src/nodestore/mod.rs
RodrigoVillar: I don't think this was addressed?

firewood/src/db.rs
```diff
@@ -521,6 +583,62 @@ mod test {
         }
     }
 
+    #[test]
+    fn test_reconstruct_parallel_vs_serial() {
```
RodrigoVillar: Is this meant to test that both parallel and serial databases use `apply_batch_serial`?

storage/src/nodestore/mod.rs
rkuris: Correct, since it was marked nit I chose to ignore it. Per the [code review guidelines](https://docs.google.com/document/d/1ptzmQjwqql4TWs1YLbwaM1vdMQFphefFAcEXaruypeo/edit?tab=t.0):
> If you’re going to nitpick, preface the comment with “nit:”. This means the author can choose to ignore the comment.

Fixed anyway in fae0a8377d4bde0adc6837149e31157364189df8

firewood/src/db.rs
rkuris: fixed in fae0a8377d4bde0adc6837149e31157364189df8

storage/src/nodestore/mod.rs
rkuris: Fixed in fae0a8377d4bde0adc6837149e31157364189df8

firewood/src/db.rs
rkuris: Fixed in 3e9ed19fdbfea17d69ad341bd8103c17e42e989a

firewood/src/db.rs
```diff
@@ -521,6 +583,62 @@ mod test {
         }
     }
 
+    #[test]
```
RodrigoVillar, line 589: I previously left a similar comment but what is `test_reconstruct_deterministic()` trying to test? Is it trying to test that both parallel and serial databases behave the same w.r.t reconstructed views?

rkuris: It changed a little from your original question. The test verifies that reconstruction works exactly the same regardless of UseParallel (well, it tests that it produces the same results regardless of that setting, it doesn't peek under the covers to check).

It's testing that reconstruction is deterministic regardless of how the base revision was built.

Added a doc comment in 948ce47822914c11446d126ca6f4e2113f04963a

RodrigoVillar: Gotcha, thanks for the clarification.

Issue #1683, opened 2026-02-16
Refactor benchmarks to reduce duplicate code


RodrigoVillar:
Yes! However, refactoring IMO falls outside the scope of this PR, hence why I passed on it. Creating an issue for refactoring our benchmarks to reduce duplicate code.

_Originally posted by @RodrigoVillar in https://github.com/ava-labs/firewood/pull/1670#discussion_r2812819269_

PR #1787, opened 2026-03-13
feat(reconstruct): Add benchmarks for rust reconstruction

rkuris:
## Why this should be merged

Add proposal_reconstruct benchmark comparing:
- Normal proposal chain vs reconstructed chain
- Serial vs parallel reconstruction  
- Includes flamegraph profiler support

## How this works

Standard criterion benchmark

## How this was tested

Performance on my macbook pro M3 shows reconstruction is 3x faster than
the proposal chain:

```
Benchmarking proposal_chain/nested: Warming up for 5.0000 s
proposal_chain/nested   time:   [950.85 ms 954.12 ms 956.44 ms]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

Benchmarking reconstructed_chain/nested: Collecting 10 samples in estimated 39.116 s (110 iteration
reconstructed_chain/nested
                        time:   [306.70 ms 308.26 ms 311.01 ms]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
```

## Breaking Changes

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1592,7 +1592,6 @@ dependencies = [
  "clap",
  "criterion",
  "ctor",
- "derive-where",
  "env_logger",
  "ethereum-types",
  "fastrace",
@@ -1600,7 +1599,6 @@ dependencies = [
  "firewood-metrics",
  "firewood-storage",
  "firewood-triehash",
- "fjall",
  "hash-db",
  "hex",
  "hex-literal",
@@ -1619,7 +1617,6 @@ dependencies = [
  "test-case",
  "thiserror",
  "typed-builder",
- "weak-table",
 ]
 
 [[package]]
@@ -1753,6 +1750,7 @@ dependencies = [
  "enum-as-inner",
  "fastrace",
  "firewood-metrics",
+ "fjall",
  "git-version",
  "hex",
  "indicatif",
@@ -1775,6 +1773,7 @@ dependencies = [
  "test-case",
  "thiserror",
  "triomphe",
+ "weak-table",
 ]
 
 [[package]]
--- a/firewood/Cargo.toml
+++ b/firewood/Cargo.toml
@@ -39,9 +39,6 @@ thiserror.workspace = true
 typed-builder = "0.23.2"
 rayon = "1.11.0"
 parking_lot.workspace = true
-fjall = "=2.11.2" # 3.0.0 has breaking changes
-derive-where = "1.6.1"
-weak-table = "0.3.2"
 lender = "0.6.2"
 nonzero_ext = "0.3.0"
 
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -528,10 +528,7 @@ mod test {
     use std::ops::{Deref, DerefMut};
     use std::path::Path;
 
-    use firewood_storage::{
-        CheckOpt, CheckerError, HashedNodeReader, IntoHashType, LinearAddress, MaybePersistedNode,
-        NodeStore, TrieHash,
-    };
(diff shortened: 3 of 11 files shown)


rkuris:
Flamegraph for the reconstructed chain benchmark attached

reconstructed_chain

Also tested performance of a proposal. There were a few performance gains there, mostly around not having to increment then drop the Arc pointer in some cases, but it's mostly the same speed.

Here's a flamegraph before and after the reconstruction changes:

proposal_chain_before

proposal_chain_after

firewood/benches/proposal_reconstruct.rs
```diff
@@ -0,0 +1,175 @@
...
+
+fn make_batch(rng: &firewood_storage::SeededRng, count: usize) -> BenchBatch {
+    repeat_with(|| {
+        let key: Vec<u8> = rng.sample_iter(&Alphanumeric).take(KEY_LEN).collect();
+        let value: Vec<u8> = rng.sample_iter(&Alphanumeric).take(VALUE_LEN).collect();
+        BatchOp::Put { key, value }
+    })
+    .take(count)
+    .collect()
+}
```
RodrigoVillar, line 75: Fine with merging this in but this should ideally be refactored in #1683

firewood/benches/proposal_reconstruct.rs
```diff
@@ -0,0 +1,175 @@
...
+        .collect();
+    (initial, batches)
+}
+
+#[expect(clippy::unwrap_used)]
+fn bench_proposal_chain(criterion: &mut Criterion) {
+    criterion
+        .benchmark_group("proposal_chain")
+        .sample_size(10)
+        .warm_up_time(std::time::Duration::from_secs(5))
```
RodrigoVillar, line 91: What's the rationale behind these values? IIRC, `criterion` prefers sample sizes of `30`.

firewood/benches/proposal_reconstruct.rs
```diff
@@ -0,0 +1,175 @@
...
+        .bench_function("nested", |b| {
+            b.iter_batched(
+                generate_batches,
+                |(initial, batches)| {
+                    let db_dir = TempDir::new().unwrap();
+                    let db_path = db_dir.path().join("benchmark_db");
+                    let cfg = DbConfig::builder()
+                        .node_hash_algorithm(NodeHashAlgorithm::compile_option())
+                        .truncate(true)
+                        .use_parallel(UseParallel::Never)
```
RodrigoVillar, line 140: Is this config necessary? I thought if when constructing a reconstructed view, the merkle ops are executed serially (i.e. `UseParallel::Never` is a no-op)

firewood/benches/proposal_reconstruct.rs
rkuris: This just makes it start faster. Since the loops are very small, a shorter warm up time is plenty.

firewood/benches/proposal_reconstruct.rs
rkuris: Explained in b3b809416417ab453ac732b3a44acf02c5915f01

When it creates that initial proposal, the flamegraphs look terrible due to the parallel threads -- each one gets a separate flamegraph, muddying up the whole point of the flamegraph to begin with.

demosdemon (approved):

firewood/benches/proposal_reconstruct.rs
RodrigoVillar: Ah sounds good.

firewood/benches/proposal_reconstruct.rs
RodrigoVillar: I would get @Elvis339's input here, but decreasing the sample size may increase the `p-value` of these benchmarks and lead to suboptimal results.

PR #1788, opened 2026-03-13
chore: remove nix flake and all references to it

demosdemon:
## Why this should be merged

We are no longer using the nix flake and do not wish to maintain it. We do not use it for development purposes and were mostly using the flake for reproducible builds. We have future plans to add Bazel build files therefore we no longer find that the flake is adding any value and is also causing me friction when adding a new `go generate` tool: https://github.com/ava-labs/firewood/actions/runs/23023292194/job/66864954453

## How this works

Removes the FFI nix flake, its lock file, the build equivalency test script, and the ffi-nix CI workflow. Updates justfile, scripts, and documentation to remove all nix-related recipes, fallbacks, and instructions.

## How this was tested

CI and review

## Breaking Changes

n/a

### DIFF
--- a/firewood/src/merkle/changes.rs
+++ b/firewood/src/merkle/changes.rs
@@ -6,7 +6,8 @@ use std::{cmp::Ordering, iter::once};
 
 use firewood_metrics::firewood_increment;
 use firewood_storage::{
-    Child, FileIoError, HashedNodeReader, Node, NodeReader, Path, SharedNode, TrieHash,
+    Child, FileIoError, HashedNodeReader, NibblesIterator, Node, NodeReader, Path, SharedNode,
+    TrieHash,
 };
 use lender::{Lender, Lending};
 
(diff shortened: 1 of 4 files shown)


justfile
```diff
@@ -20,68 +16,6 @@ check-clean-branch:
...
 
-# Check if the FFI flake is up-to-date (requires clean git tree)
-check-ffi-flake: check-nix
-    #!/usr/bin/env bash
-    set -euo pipefail
-    ./scripts/run-just.sh update-ffi-flake
-    ./scripts/run-just.sh check-clean-branch
-
-# Check if the golang version is set consistently (requires clean git tree)
-check-golang-version: check-nix
```
alarso16, line 31: We should still keep this to ensure all the testing go.mods match

demosdemon: Why do we even have separate `go.mod` files for those?

demosdemon: I'll put that in a new issue because this previously was only executed on the flake ci workflow which wasn't on every PR anyways.

alarso16: We have separate go.mod to avoid a circular reference with avalanchego. We don't actually need any of that garbage for anything besides tests. i think this is still the correct way to do it

Issue #1789, opened 2026-03-13
ci: verify `go` version in all `go.mod` files match


demosdemon:
This was previously done in the `ffi-nix` workflow which was removed in https://github.com/ava-labs/firewood/pull/1788.

We should remove the `toolchain` field and add a CI workflow step to ensure the `go` version matches in all files.

PR #1790, opened 2026-03-13
ci: verify go.mod files all have the same declared go version

demosdemon:
## Why this should be merged

Closes: #1789

## How this works

This adds a CI workflow to ensure the go version in `go.mod` is the same for all tracked files.

## How this was tested

CI: https://github.com/ava-labs/firewood/actions/runs/23058274067/job/66977621517

## Breaking Changes

None


.github/workflows/verify-go-versions.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install jq
+        run: |
+          if command -v jq >/dev/null 2>&1; then
+            # expected on most GHA runners
+            echo "jq is already installed"
+          else
+           sudo apt-get update && sudo apt-get install -y jq
+          fi
+
+      - name: Extract Go versions
```
alarso16, line 23: I think this would be a lot clearer as an extracted script in `.github/scripts`

ffi/go.mod
```diff
@@ -1,12 +1,10 @@
...
 
-go 1.25
-
-// Changes to the toolchain version should be replicated in:
+// Changes to the go version should be replicated in:
 //   - ffi/go.mod (here)
 //   - ffi/tests/eth/go.mod
 //   - ffi/tests/firewood/go.mod
-toolchain go1.25.8
+go 1.25.8
```
alarso16, line 7: Not super positive, but is this a surprise breaking change? Like if avalanchego was using 1.25.7, could they still import this?

demosdemon: avalanchego already has theirs set to `go 1.25.8`:

https://github.com/ava-labs/avalanchego/blob/c4486367dab5494a00a6f6a10ef277d07d3c549a/go.mod#L19

PR #1793, opened 2026-03-13
chore(release): prepare for v0.3.0

demosdemon:
## Why this should be merged

Next release.

## How this works

Bumps the versions and dependencies.

## How this was tested

CI

## Breaking Changes

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,12 +55,12 @@ cast_possible_truncation = "allow"
 
 [workspace.dependencies]
 # workspace local packages
-firewood = { path = "firewood", version = "0.2.0" }
-firewood-macros = { path = "firewood-macros", version = "0.1.0" }
-firewood-storage = { path = "storage", version = "0.2.0" }
-firewood-ffi = { path = "ffi", version = "0.2.0" }
-firewood-metrics = { path = "metrics", version = "0.2.0" }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -55,12 +55,12 @@ cast_possible_truncation = "allow"
...
 [workspace.dependencies]
 # workspace local packages
-firewood = { path = "firewood", version = "0.2.0" }
-firewood-macros = { path = "firewood-macros", version = "0.1.0" }
-firewood-storage = { path = "storage", version = "0.2.0" }
-firewood-ffi = { path = "ffi", version = "0.2.0" }
-firewood-metrics = { path = "metrics", version = "0.2.0" }
-firewood-replay = { path = "replay", version = "0.2.0" }
+firewood = { path = "firewood", version = "0.3.0" }
+firewood-macros = { path = "firewood-macros", version = "0.2.0" }
```
RodrigoVillar, line 59: q: why is this `0.2.0` instead of `0.3.0`?

demosdemon: `firewood-macros` did not receive any changes the last round so it was never bumped to 0.2.0. Eventually, all of the versions will diverge at some point as its likely that will happen again where some packages will have updates but not all.

Issue #1768, opened 2026-03-11
Bad block detection


RodrigoVillar:
Detect when read from disk is inconsistent with expected state.

PR #1798, opened 2026-03-15
chore: Rootstore cleanups

rkuris:
## Why this should be merged

While on a flight, was looking at the docs and noticed the root store docs were light at the top level. While documenting, I re-audited the code there for both efficiency and some future goal work.

I also think most of root_store should live in firewood_storage.

- Move the `with_root` method from firewood to storage, reducing the surface area of the main firewood crate and making `with_root` pub(crate) only
- Add metrics for reads and writes to RootStore, including stats on how often the cache is used
- Modify with_root to eagerly read the root node, including the possibility of an error
 - Always check the hash and return an error when there is a mismatch (this is a goal for firewood 1.0 for all of them, this just fixes this one case)

## How this was tested

CI

## Breaking Changes

None


firewood/src/db.rs
```diff
@@ -1235,9 +1232,11 @@ mod test {
         });
     }
 
+    /// Verifies that after reopening the database, an older revision can still
+    /// be retrieved via the root store.
```
RodrigoVillar, line 1395: This was meant to test `NodeStore::with_root()` rather than `RootStore` itself. I like this new test structure, but wondering if we should add a new test that tests just `NodeStore::with_root()`

storage/src/root_store.rs
```diff
@@ -10,12 +16,21 @@ use std::{
...
+use crate::registry;
+
+/// Type alias for a committed revision stored in the root store.
+pub type CommittedRevision = Arc<NodeStore<Committed, FileBacked>>;
 
+/// fjall uses partitions. We store all firewood-specific data in a
+/// partition named 'firewood'
 const FJALL_PARTITION_NAME: &str = "firewood";
 
+/// This structure hold everything related to an open root store
```
RodrigoVillar:
```suggestion
/// This structure holds everything related to an open root store
```

storage/src/nodestore/mod.rs
```diff
@@ -159,33 +159,38 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
...
+            nodestore.kind.root = Some(Child::AddressWithHash(root_address, root_hash));
+            Ok(nodestore)
+        } else {
+            Err(FileIoError::new(
+                std::io::Error::other("hash verification failed"),
+                None,
+                root_address.get(),
+                None,
+            ))
+        }
```
RodrigoVillar, line 193: Why did this move from running on debug builds to now always running? I'm also interested in the performance implications of this change.

storage/src/registry.rs
```diff
@@ -39,6 +39,12 @@ pub const IO_READ_COUNT: &str = "io.read";
 /// Number of proposals reparented to committed parent.
 pub const REPARENTED_PROPOSAL_COUNT: &str = "proposals.reparented";
 
+/// Fetch attempts from the rootstore
+pub const ROOTSTORE_GET: &str = "rootstore_get";
+
+/// Store operations to the rootstore
+pub const ROOTSTORE_PUT: &str = "rootstore_put";
```
RodrigoVillar: Adding this metric is great, as we'll be able to tell which archival queries came from `RootStore` vs the in-memory revisions vec.

storage/src/registry.rs
```diff
@@ -39,6 +39,12 @@ pub const IO_READ_COUNT: &str = "io.read";
 /// Number of proposals reparented to committed parent.
 pub const REPARENTED_PROPOSAL_COUNT: &str = "proposals.reparented";
 
+/// Fetch attempts from the rootstore
+pub const ROOTSTORE_GET: &str = "rootstore_get";
```
RodrigoVillar, line 43: What signal are we getting from this metric?

storage/src/root_store.rs
```diff
@@ -63,79 +78,91 @@ impl RootStore {
...
         address: &LinearAddress,
     ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         self.items.insert(**hash, address.get().to_be_bytes())?;
 
+        firewood_counter!(registry::ROOTSTORE_PUT).increment(1);
+
+        // flush the keyspace to protect against application crashes
+        //
+        // note that OS crashes or power failures may result in the loss
+        // of this revision in the root store
```
RodrigoVillar: I'm not sure what we gain from this comment; the documentation on `keyspace.persist()` and the assoc `PersistMode::Buffer` provide enough context IMO:

ref: 
- https://docs.rs/fjall/2.11.2/fjall/struct.Keyspace.html#impl-Keyspace
- https://docs.rs/fjall/2.11.2/fjall/enum.PersistMode.html

storage/src/nodestore/mod.rs
demosdemon: We do need some level of verification when reading nodes (https://github.com/ava-labs/firewood/issues/1768). Though, I haven't decided which nodes should be verified on read and when.

storage/src/root_store.rs
```diff
@@ -63,79 +78,91 @@ impl RootStore {
...
+    /// Returns the raw underlying database error if any
     pub fn get(
         &self,
         hash: &TrieHash,
     ) -> Result<Option<CommittedRevision>, Box<dyn std::error::Error + Send + Sync>> {
-        // Obtain the lock to prevent multiple threads from caching the same result.
+        // 1. Obtain the lock to prevent multiple threads from caching the same result
         let mut revision_cache = self.revision_cache.lock();
 
         // 1. Check if the committed revision is cached.
```
demosdemon: nit: two number ones

storage/src/nodestore/mod.rs
rkuris: These really do. These are older revisions that someone is fetching from disk that have been there for a while.

I'm not so worried about the performance of fetching older revisions and this check seems cheap and infrequent.

storage/src/registry.rs
rkuris: How many times we are fetching from the rootstore

storage/src/root_store.rs
rkuris: We could refer the reader to that documentation or just state it here. I picked the latter but easily could change it to the former.

storage/src/root_store.rs
rkuris: Not a nit, that's annoying. I'll fix it after I meet up with Rodrigo on this PR.

firewood/src/db.rs
rkuris: Fixed in fd78e204bc2b950e8723711333a4fba214a6eb5f

storage/src/root_store.rs
rkuris: Fixed in fd78e204bc2b950e8723711333a4fba214a6eb5f

storage/src/root_store.rs
rkuris: Fixed in fd78e204bc2b950e8723711333a4fba214a6eb5f

storage/src/registry.rs
RodrigoVillar: Ah I messed up - this question should've been about `ROOTSTORE_PUT`. Applying the feedback previously said about `ROOTSTORE_PUT` to `ROOTSTORE_GET`: 

> Adding this metric is great, as we'll be able to tell which archival queries came from RootStore vs the in-memory revisions vec.

storage/src/registry.rs
RodrigoVillar: Please refer to https://github.com/ava-labs/firewood/pull/1798#discussion_r2956794042, but what signal is this metric giving us? For example, if we're in archive mode then counting the number of times we `flush_nodes()` or something similar would give us the same data.

storage/src/root_store.rs
RodrigoVillar: Yeah "Flush the keyspace to the OS buffer ... please refer to Fjall for durabiltiy guarantees" or something similar would sufficient IMO.

storage/src/registry.rs
rkuris: Ah, I see that point. I'll remove the put parts. Since we always call flush_nodes every time we call put, it's definitely redundant.

rkuris: Fixed in 3089bc1f45f302b4fe3e9c2a6b120d592389c279

storage/src/root_store.rs
rkuris: Fixed in 5a65a27b352bbab4b4ce9c64262cd06e065df7d8

storage/src/nodestore/mod.rs
```diff
@@ -1353,6 +1358,89 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn with_root_success() -> Result<(), Box<dyn Error>> {
```
RodrigoVillar, line 1362: Educational rust question: if this returns `Err(Box<...>)`, the test fails, yes? (i.e. is this another idiomatic way of writing Rust tests)

rkuris: > Educational rust question: if this returns `Err(Box<...>)`, the test fails, yes? (i.e. is this another idiomatic way of writing Rust tests)

The Box part says that this is a heap allocated thing but the size of it is not known at compile time. Basically the function returns a pointer to that heap and transfers ownership so it will be freed when it goes out of scope.

It also says that the thing in the box implements the Error trait: https://doc.rust-lang.org/std/error/trait.Error.html so if you have a boxed error like this you can call display() on it.

There are conversions in the standard library like From<E> for Box<dyn Error> where E implements Error, or impl<'a> From<&str> for Box<dyn Error + 'a>, so you can use the ? operator to call these conversions automatically.

Err("blah")? works as a Box<dyn Error> because of the last impl.

Errors like this are more expensive because they allocate and free memory for the Box, but yeah they are useful in tests.

RodrigoVillar: Ah I was more-so talking about the function signature - most tests I've seen in Firewood follow:

```rust
#[test]
fn test_foo() {
    // Test here
}

rkuris: https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#tests-and-

PR #1801, opened 2026-03-17
docs(c-chain-benchmark): add BENCHMARK.md 

Elvis339:
## Why this should be merged
Running a benchmark required knowing which S3 buckets exist, which state snapshots pair with which block ranges, which runner to use and why, and what to do when the 6h monitoring limit is hit. None of this was written down. This document makes the benchmarking infrastructure self-serve.

## How this works
Adds `BENCHMARKS.md` at the repo root.

## How this was tested
Not needed

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,3 +2,4 @@
 # Note: more specific rules overrule wildcard
 *        @rkuris @demosdemon @RodrigoVillar @bernard-avalabs
 /ffi     @rkuris @demosdemon @alarso16 @RodrigoVillar @bernard-avalabs
+/fwdctl  @rkuris @demosdemon @RodrigoVillar @bernard-avalabs @aminR443
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2873,9 +2873,9 @@ dependencies = [
 
 [[package]]
 name = "lz4_flex"
-version = "0.11.5"
+version = "0.11.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08ab2867e3eeeca90e844d1940eab391c9dc5228783db2ed999acbc0a9ed375a"
+checksum = "373f5eceeeab7925e0c1098212f2fbc4d416adec9d35051a6ab251e824c1854a"
 
 [[package]]
 name = "maybe-dangling"
--- a/storage/src/nodestore/persist.rs
+++ b/storage/src/nodestore/persist.rs
@@ -183,7 +183,7 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
     ///
     /// Returns a [`FileIoError`] if any node cannot be written to storage.
     #[fastrace::trace(short_name = true)]
-    pub fn flush_nodes(&self, header: &mut NodeStoreHeader) -> Result<(), FileIoError> {
+    fn flush_nodes(&self, header: &mut NodeStoreHeader) -> Result<(), FileIoError> {
         let flush_start = Instant::now();
 
         let mut node_allocator = NodeAllocator::new(self.storage.as_ref(), header);


Elvis339:
> Could you coordinate with `dev-infra` to see what the timeline is on Firewood joining the monorepo? If it's the case that Firewood joins the monorepo, let's say this week, then I'm not sure what the value is of having this doc if it'll become stale in a week.

Sure, I'll check with dev-infra on the timeline.

Even when Firewood moves to the monorepo, the setup (runner selection, S3 data, Grafana) stays the same the operational knowledge in this doc doesn't become stale. Happy to update the doc paths when the move happens.

BENCHMARKS.md
```diff
@@ -0,0 +1,171 @@
+# Re-Execution Benchmarks
+
+Firewood measures performance by re-executing historical C-Chain blocks against
+a state snapshot. Results publish to GitHub Pages automatically.
+
+For how this benchmark work, what S3 data exists, and how to create snapshots, see
```
RodrigoVillar:
```suggestion
For how this benchmark works, what S3 data exists, and how to create snapshots, see
```

BENCHMARKS.md
```diff
@@ -0,0 +1,171 @@
...
+| --- | --- | --- |
+| Weekdays 05:00 UTC | `firewood-40m-41m` (blocks 40M–41M) | <2h |
+
+## Choosing a Test
+
+| Goal | Use |
+| --- | --- |
+| A/B test, daily-style run, quick regression check | `firewood-40m-41m` |
+| Quick smoke test | `firewood-101-250k` (~7 min) |
+| Deep regression | `firewood-33m-40m` (~7h, needs >6h path) |
```
RodrigoVillar: Why is this listed if it's not compatible with `just bench-cchain`?

BENCHMARKS.md
```diff
@@ -0,0 +1,171 @@
...
+**Pairing matters.** State snapshot must match the block immediately before
+`START_BLOCK`. Wrong pairings silently produce incorrect results.
+→ [How to create new snapshots](https://github.com/ava-labs/avalanchego/blob/master/tests/reexecute/c/README.md#how-to-create-and-use-a-re-execution-snapshot)
+
+## Gotchas
+
+1. **Push before triggering.** The workflow builds from the remote branch.
+   Unpushed commits benchmark the wrong code — the justfile catches this.
+
+2. **`GH_TOKEN` needs cross-repo access.** Requires a PAT for both
```
RodrigoVillar: What's a PAT?

BENCHMARKS.md
```diff
@@ -0,0 +1,171 @@
...
+   `ava-labs/firewood` and `ava-labs/avalanchego`. Default `gh auth login`
+   tokens scoped to one repo will fail silently.
```
RodrigoVillar: Also not sure why this is relevant for users of `just bench-cchain` (i.e. doesn't `gh auth login` solves this for me)?

BENCHMARKS.md
rkuris: Personal Access Token (PAT)

BENCHMARKS.md
rkuris: Because `gh auth login` scoped to one repo will fail silently, unfortunately. If you work on multiple repos you may be in for a surprise when you get throttled.

BENCHMARKS.md
Elvis339: Fair point, going to remove it.

rkuris:
> > Could you coordinate with `dev-infra` to see what the timeline is on Firewood joining the monorepo? If it's the case that Firewood joins the monorepo, let's say this week, then I'm not sure what the value is of having this doc if it'll become stale in a week.
> 
> Sure, I'll check with dev-infra on the timeline.
> 
> Even when Firewood moves to the monorepo, the setup (runner selection, S3 data, Grafana) stays the same the operational knowledge in this doc doesn't become stale. Happy to update the doc paths when the move happens.

Even if this is true (they are moving within a week) the project of editing this document should be part of that migration, so let's get it in now and have them note that it needs doing.

p.s. there's no way it's next week :)

rkuris:
> This is awesome but I'm concerned about adding another top level README. I'll create an issue to move this later.

Created #1808

BENCHMARKS.md
Elvis339: Valid point, narrowed the scope to CI only.

PR #1805, opened 2026-03-18
fix(fwdctl): use libevm only when specified

RodrigoVillar:
## Why this should be merged

Currently, when executing `fwdctl launch deploy`, the CLI will clone and use the latest version of `libevm` even though the client didn't specify a specific version of `libevm` to use (i.e. we should be using the version specified in `avalanchego`/`coreth`).

## How this works

If the `--libevm-branch` flag isn't specified, then:
1. We don't clone `libevm`
2. `avalanchego`/`coreth` will retain their version of `libevm`

## How this was tested

CI + ran `fwdctl launch deploy`

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


fwdctl/src/launch/mod.rs
```diff
@@ -302,7 +302,7 @@ pub struct DeployOptions {
     )]
     pub variable_overrides: Vec<VariableOverride>,
 
-    /// Enable metrics server during execution (`--metrics-server=false` disables)
+    /// Whether to enable the metrics server during execution.
```
RodrigoVillar, line 305: Changes `--help` message from:

``` 
--metrics-server <BOOL>       Enable metrics server during execution (`--metrics-server=false` disables) [default: true] [possible values: true, false]
```

to:

```
--metrics-server <BOOL>        Whether to enable the metrics server during execution [default: true] [possible values: true, false]
```

RodrigoVillar:
@AminR443 Yup I tested this fix with and without the `--libevm-branch` specified (with each option doing as expected)

PR #1811, opened 2026-03-18
feat: add devcontainer configuration for development environment

demosdemon:
## Why this should be merged

The previous Docker setup (README.docker.md) was outdated and required extensive manual tool installation. A devcontainer provides a reproducible, ready-to-use development environment with all required tooling pre-installed.

## How this works

- Custom Dockerfile based on Ubuntu 24.04 (parameterized for easy 26.04 upgrade) with Rust (stable + nightly), Go 1.25.8, and Nix (Determinate Systems installer with flakes enabled).
- sccache installed and globally configured as Cargo's rustc-wrapper, with a named volume for cache persistence across container rebuilds.
- Additional named volumes for cargo registry, cargo git deps, and Go module cache.
- Pre-installs cargo tools (ast-grep, cargo-nextest, just, starship, etc.), Go tools (shfmt), GitHub CLI, Claude Code CLI, and shellcheck.
- docker-in-docker devcontainer feature for building/testing images.
- Username inherited from the host via build-time ARG so git branch naming conventions (e.g. $USER/feature) work naturally.
- 17 VS Code extensions auto-installed including rust-analyzer, Go, and Claude Code.
- CLAUDE.md updated with conventional commit guidance and PR template recommendations for commit bodies.

## How this was tested

Launched devcontainer in VS Code, verified all tools are installed and working (rustc, cargo, go, nix, sccache).


README.md
rkuris, line 100: The badge was removed, so we should just have them click the link instead.

demosdemon: It's still there. 

Image

I moved it from the top to this section because the image is much larger than the other badged and was odd

rkuris: :shipit:

PR #1814, opened 2026-03-19
feat(reconstruction): Add Go bindings and tests

rkuris:
## Why this should be merged

Adds a go layer for reconstructed revisions

## How this works

Added reconstructed.go with Go wrapper for ReconstructedHandle

## How this was tested

Added reconstructed_test.go with comprehensive tests

## Breaking Changes

None, but adds new functionality

### DIFF
--- a/ffi/reconstructed_test.go
+++ b/ffi/reconstructed_test.go
@@ -99,24 +99,10 @@ func BenchmarkReconstructChain(b *testing.B) {
 		valueLen     = 32
 	)
 
-	makeBatch := func(rng *rand.Rand) []BatchOp {
-		batch := make([]BatchOp, 0, batchItems)
-		for range batchItems {
-			key := make([]byte, keyLen)
-			value := make([]byte, valueLen)
-			_, err := rng.Read(key)
-			r.NoError(err)
-			_, err = rng.Read(value)
-			r.NoError(err)
-			batch = append(batch, Put(key, value))
-		}
-		return batch
-	}
-
 	rng := rand.New(rand.NewSource(1234))
 	batches := make([][]BatchOp, 0, totalBatches)
 	for range totalBatches {
-		batches = append(batches, makeBatch(rng))
+		batches = append(batches, makeRandomBatch(b, rng, batchItems, keyLen, valueLen))
 	}
 
 	db := newTestDatabase(b)
@@ -145,6 +131,27 @@ func BenchmarkReconstructChain(b *testing.B) {
 	}
 }
 
+func TestReconstructedDump(t *testing.T) {
+	r := require.New(t)
+	db := newTestDatabase(t)
+
+	_, _, batch := kvForTest(4)
+	root, err := db.Update(batch[:2])
+	r.NoError(err)
+
+	rev, err := db.Revision(root)
+	r.NoError(err)
+	t.Cleanup(func() { r.NoError(rev.Drop()) })
+
+	reconstructed, err := rev.Reconstruct(batch[2:4])
+	r.NoError(err)
+	t.Cleanup(func() { r.NoError(reconstructed.Drop()) })
+
+	dot, err := reconstructed.Dump()
+	r.NoError(err)
+	r.Contains(dot, "digraph")
+}
+
 func TestReconstructedDropThenUse(t *testing.T) {
 	r := require.New(t)
 	db := newTestDatabase(t)
@@ -159,11 +166,6 @@ func TestReconstructedDropThenUse(t *testing.T) {
 
 	reconstructed, err := rev.Reconstruct(batch[2:4])
 	r.NoError(err)
-
-	// Dump succeeds on a live view.
-	dot, err := reconstructed.Dump()
-	r.NoError(err)
-	r.NotEmpty(dot)
 
 	// First Drop succeeds.
 	r.NoError(reconstructed.Drop())
--- a/ffi/firewood_test.go
+++ b/ffi/firewood_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -356,6 +357,25 @@ func sortKV(keys, vals [][]byte) error {
 		}
 	}
 	return nil
+}
+
+// makeRandomBatch generates a batch of count Put operations with random keys
+// and values of the specified lengths, using the provided rng.
+// makeRandomBatch generates a batch of count Put operations with random keys
+// and values of the specified lengths, using the provided rng.
+func makeRandomBatch(tb testing.TB, rng *rand.Rand, count, keyLen, valueLen int) []BatchOp {
+	tb.Helper()
+	batch := make([]BatchOp, 0, count)
+	for range count {
+		key := make([]byte, keyLen)
+		value := make([]byte, valueLen)
+		_, err := rng.Read(key)
+		require.NoError(tb, err)
+		_, err = rng.Read(value)
+		require.NoError(tb, err)
+		batch = append(batch, Put(key, value))
+	}
+	return batch
 }
 
 func makeBatch(keys, vals [][]byte) []BatchOp {


ffi/reconstructed.go
```diff
@@ -0,0 +1,241 @@
...
+	case C.ReconstructedResult_NullHandlePointer:
+		return nil, errDBClosed
+	case C.ReconstructedResult_Ok:
+		body := (*C.ReconstructedResult_Ok_Body)(unsafe.Pointer(&result.anon0))
+		reconstructed := &Reconstructed{
+			handle: body.handle,
+			root:   EmptyRoot,
+		}
+		reconstructed.keepAliveHandle.init(wg)
+		runtime.SetFinalizer(reconstructed, (*Reconstructed).Drop)
```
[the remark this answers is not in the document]
rkuris: Fixed in 6dd9f82467315ac2a894a25d53049cd1484d6151

ffi/reconstructed.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	r.keepAliveHandle.mu.Lock()
+	if wg != nil {
+		wg.Add(1)
+		r.keepAliveHandle.outstandingHandles = wg
+	}
```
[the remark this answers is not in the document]
rkuris: Fixed in 6dd9f82467315ac2a894a25d53049cd1484d6151

ffi/reconstructed.go
```diff
@@ -0,0 +1,212 @@
...
+
+	result := C.fwd_reconstruct_on_reconstructed(r.ptr, kvp)
+	// The old handle is consumed by the FFI call regardless of outcome.
+	r.ptr = nil
+
+	newHandle, err := getReconstructedHandleFromResult(result, nil)
+	if err != nil {
+		r.dropped = true
+		return err
+	}
```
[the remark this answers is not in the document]
alarso16: Copilot is on to something here. It's extremely weird that creating a child would corrupt this one, and I don't think this is dropped correctly

ffi/reconstructed.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Root returns the root hash of the reconstructed view.
+func (r *Reconstructed) Root() Hash {
+	r.rootMu.Lock()
+	if r.rootSet {
+		root := r.root
+		r.rootMu.Unlock()
+		return root
+	}
```
alarso16, line 79: this logic is shared at the bottom. I think this can be refactored so that it's something like:
```
if rootNotSet {
     get root
}
return root
```
The only thing this will complicate is lock order I think

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	for i := 8; i < len(keys); i++ {
+		got, err := reconstructed.Get(keys[i])
+		r.NoError(err)
+		r.Nil(got)
+	}
+
+	err = reconstructed.Reconstruct(batch[8:])
+	r.NoError(err)
```
alarso16: nit: 
```suggestion
	r.NoError(reconstructed.Reconstruct(batch[8:]))
```

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,242 @@
...
+		r.Equal(vals[i], got)
+	}
+
+	for i := 8; i < len(keys); i++ {
+		got, err := reconstructed.Get(keys[i])
+		r.NoError(err)
+		r.Nil(got)
+	}
+
+	err = reconstructed.Reconstruct(batch[8:])
```
alarso16: Consider checking that the root is updated (I think just `oldRoot != newRoot` is sufficient)

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,242 @@
...
+	close(start)
+	wg.Wait()
+
+	if reconstructErr == nil {
+		r.NoError(reconstructed.Drop())
+	} else {
+		r.ErrorIs(reconstructErr, ErrDroppedReconstructed)
+	}
+
+	if getErr != nil {
```
alarso16: What is the expected behavior? This is certainly a race, and this test acknowledges that there are multiple outcomes. I don't understand the point of this test, or the following one.

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,242 @@
...
+	r.NoError(err)
+
+	start := make(chan struct{})
+	var wg sync.WaitGroup
+	var getErr error
+	var getVal []byte
+	var reconstructErr error
+
+	wg.Add(2)
+	go func() {
```
alarso16: Consider `wg.Go`

ffi/revision.go
```diff
@@ -101,6 +104,35 @@ func (r *Revision) Iter(key []byte) (*Iterator, error) {
...
+	return getReconstructedFromResult(
+		C.fwd_reconstruct_on_revision(r.ptr, kvp),
+		r.keepAliveHandle.outstandingHandles,
+	)
+}
+
+// Drop releases the resources backed by the revision handle.
+//
+// It is safe to call Drop multiple times; subsequent calls after the first are no-ops.
+func (r *Revision) Drop() error {
```
alarso16: The pointer was embedded to expose the drop function. Why did you wrap it?

ffi/reconstructed.go
```diff
@@ -0,0 +1,212 @@
...
+	r.rootMu.Unlock()
```
RodrigoVillar: A comment/explanation on why we `unlock()`/subsequentially `lock()` the root mutex would be great here.

RodrigoVillar: If the `lock()` can be held for the lifetime of `Root()`, a `defer` would be nice to add here.

ffi/reconstructed.go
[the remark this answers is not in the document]
rkuris: Yep, I'm fixing it.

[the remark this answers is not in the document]
rkuris: Fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		r.dropped = true
+		return err
+	}
+
+	r.ptr = newHandle
+
+	// Safe to lock rootMu here: Root() takes rootMu then keepAliveHandle.mu.RLock(),
+	// so it cannot be between rootMu.Unlock() and keepAliveHandle.mu.RLock() while
+	// we hold the write lock.
```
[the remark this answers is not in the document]
rkuris: Fixed in 577c4cd3ce1d43f2a152989c908dae7076063625 and refined further in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/iterator.go
```diff
@@ -167,7 +167,7 @@ func (it *Iterator) Key() []byte {
 // If the iterator has not been advanced or is exhausted, it returns nil.
 //
 // If the iterator was advanced with [Iterator.NextBorrowed], the returned slice
-// borrows Rust memory and is only valid until the next advance or drop.
+// borrows Rust memory and is only valid until the next advance or release.
 func (it *Iterator) Value() []byte {
```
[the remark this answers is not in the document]
rkuris: Fixed in 577c4cd3ce1d43f2a152989c908dae7076063625

ffi/reconstructed_test.go
rkuris: fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed_test.go
rkuris: Agreed it's autogenerated trash and I didn't review it thoroughly. Thanks for catching it. Removed it in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed_test.go
rkuris: Fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/revision.go
rkuris: Good point. Fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed.go
rkuris: Fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed_test.go
rkuris: Fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed.go
rkuris: Yeah this was messy. Fixed in 704b247537f5cd85764b526a0fde6b68e18d1248

ffi/reconstructed.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if r.rootSet {
+		return r.root
+	}
+
+	// rootMu is held while calling into FFI so that concurrent Root() calls
+	// don't redundantly compute the hash. keepAliveHandle.mu is taken second;
+	// Reconstruct() acquires them in the opposite order but resets rootSet
+	// under its own write lock, so there is no deadlock.
+	r.keepAliveHandle.mu.RLock()
```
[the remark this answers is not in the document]
rkuris: Fixed in c61a34f20f5d3b411d4bc919d26d67063fb374f8 and audited for other deadlocks as well. New comments in 0ae3524d6155a8fcaf1beba188e3a6ace8ae655a explain the lock strategy.

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+//
+// Reconstructed handles must be released before the associated database is closed
+// by calling [Reconstructed.Drop]. A cleanup function is registered to call Drop
+// automatically if needed, but explicit calls are recommended.
+type Reconstructed struct {
+	*handle[*C.ReconstructedHandle]
+
+	root    Hash
+	rootMu  sync.Mutex
+	rootSet bool
```
RodrigoVillar:
```suggestion
    rootMu  sync.Mutex
	root    Hash
	rootSet bool
```

nit: from advice I got working in AvalancheGo, the mutex should go first followed by the fields that the mutex protects.

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+	var pinner runtime.Pinner
+	defer pinner.Unpin()
+
+	itResult := C.fwd_iter_on_reconstructed(r.ptr, newBorrowedBytes(key, &pinner))
+	return getIteratorFromIteratorResult(itResult)
+}
+
+// Reconstruct applies a new batch on top of this reconstructed view.
+//
+// On success, the receiver is updated to point at the newly reconstructed view.
```
RodrigoVillar, line 132: Nice!

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+	r.rootMu.Lock()
+	defer r.rootMu.Unlock()
+
+	if r.rootSet {
+		return r.root
+	}
+
+	if r.ptr == nil {
+		return EmptyRoot
+	}
```
RodrigoVillar, line 83: Should this be the first check?

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,191 @@
...
+		valueLen     = 32
+	)
+
+	makeBatch := func(rng *rand.Rand, count int) []BatchOp {
+		batch := make([]BatchOp, 0, count)
+		for range count {
+			key := make([]byte, keyLen)
+			value := make([]byte, valueLen)
+			rng.Read(key)
+			rng.Read(value)
```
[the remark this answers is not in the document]
rkuris: This seems to be an ongoing debate in golang, see https://www.reddit.com/r/golang/comments/1rgt8ps/do_you_actually_check_the_error_for_cryptorandread/#:~:text= for the most recent discussion.

I'm opting to ignore this request, as this is test code anyway, and I can't imagine why rng.Read might fail in such a way to provide enough signal that there is a real problem being masked by it.

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+	itResult := C.fwd_iter_on_reconstructed(r.ptr, newBorrowedBytes(key, &pinner))
+	return getIteratorFromIteratorResult(itResult)
+}
+
+// Reconstruct applies a new batch on top of this reconstructed view.
+//
+// On success, the receiver is updated to point at the newly reconstructed view.
+// On error, the receiver is no longer usable and its resources are fully released;
+// calling [Reconstructed.Drop] is not required (but safe as a no-op).
+func (r *Reconstructed) Reconstruct(batch []BatchOp) error {
```
RodrigoVillar, line 135: While this is idiomatic, being able to switch the `Reconstructed` view and the fact that `r` can have multiple owners makes me take Rust's ownership model for granted 😅

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,191 @@
...
+	t.Cleanup(func() {
+		r.NoError(rev.Drop())
+	})
+
+	reconstructed, err := rev.Reconstruct(batch[4:6])
+	r.NoError(err)
+
+	start := make(chan struct{})
+	var wg sync.WaitGroup
+	errCh := make(chan error, 32)
```
RodrigoVillar: Why `32`?

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	r.NoError(reconstructed.Drop())
+}
+
+func BenchmarkReconstructFromRevision(b *testing.B) {
+	r := require.New(b)
+	db := newTestDatabase(b)
+
+	_, _, batch := kvForTest(1024)
```
RodrigoVillar:
```suggestion
	_, _, batch := kvForTest(numKeys)
```

nit: this seems like a magic number

ffi/reconstructed_test.go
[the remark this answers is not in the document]
maru-ava: If choosing not to handle a given error, the error should be explicitly ignored (i.e. `_ = ...`) rather than implicitly ignored (i.e. `...`). Best practice for ignoring errors also suggests the addition of a comment capturing the rational (e.g. 'the error is not actionable' or 'the error is impossible for our range of input values').

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRevisionReconstructReadsAndChains(t *testing.T) {
+	r := require.New(t)
+	db := newTestDatabase(t)
+
+	keys, vals, batch := kvForTest(10)
```
maru-ava: (No action required) Maybe avoid the use of magic numbers (e.g. 10, 8, 5, etc) in preference for constants and/or relative offsets to simplify maintenance?

ffi/reconstructed.go
rkuris: I think we don't want to move this. If the caller drops it, the cached hash is actually still valid, and we can return it. Technically this is a caller mistake anyway, and isn't formally defined, so I think this doesn't matter much -- we don't return an error from this function to explain what the problem actually is.

This is also identical to what we do in propose, so if this really should return EmptyRoot we should fix this in both places (perhaps a separate bug, but I think this isn't important).

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	rev, err := db.Revision(root)
+	r.NoError(err)
+	t.Cleanup(func() {
+		r.NoError(rev.Drop())
+	})
+
+	reconstructed, err := rev.Reconstruct(batch[5:8])
+	r.NoError(err)
+
```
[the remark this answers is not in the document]
rkuris: Fixed in 82009612c824f290b3f62b217acf342c230a4516

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// Get retrieves the value for the given key in this reconstructed view.
+func (r *Reconstructed) Get(key []byte) ([]byte, error) {
+	r.keepAliveHandle.mu.RLock()
+	defer r.keepAliveHandle.mu.RUnlock()
+
+	if r.dropped {
+		return nil, ErrDroppedReconstructed
+	}
```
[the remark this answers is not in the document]
rkuris: Fixed in 82009612c824f290b3f62b217acf342c230a4516

ffi/reconstructed_test.go
rkuris: fixed in 29e946168510cfdcede319d0bcbd52bfe1dffae5

ffi/reconstructed_test.go
rkuris: fixed in 3324698267dec251f6a81ece4d4850c44814a8d7

ffi/reconstructed_test.go
rkuris: Answered in 29e946168510cfdcede319d0bcbd52bfe1dffae5

ffi/reconstructed.go
rkuris: Fixed in 29e946168510cfdcede319d0bcbd52bfe1dffae5. Why can't this be checked by the linter?

ffi/reconstructed_test.go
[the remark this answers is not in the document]
rkuris: Fixed in db349e3c6e00c5f2af2da85b5bcfce02e6ebbe59

ffi/reconstructed.go
rkuris: Think about the fact that each of these Reconstructed items are references counted and they act more like `Arc<T>` instead of just T, which includes all that overhead, makes me think I'm happy I'm a rustacean.

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,243 @@
...
+		<-start
+		errCh <- reconstructed.Drop()
+	})
+
+	close(start)
+	wg.Wait()
+	close(errCh)
+
+	for err := range errCh {
+		if err == nil {
```
maru-ava: Maybe simplify? `FailNow` is usually reserved for instances where a regular assertion isn't possible.

```golang
if err != nil {
    r.ErrorIs(err, ErrDroppedReconstructed, "unexpected concurrent error: %v", err)
}
```

rkuris: fixed in 2315ec33b8e0ae187c4ea7f36ca842bf9654d24d

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+	close(start)
+	wg.Wait()
+	close(errCh)
+
+	for err := range errCh {
+		if err != nil {
+			r.ErrorIs(err, ErrDroppedReconstructed, "unexpected concurrent error: %v", err)
+		}
+	}
```
RodrigoVillar: If you care only if one of the go-routines errors, using an `errgroup` would be prefered here: https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		current, err := rev.Reconstruct(batches[0])
+		r.NoError(err)
+		for _, batch := range batches[1:] {
+			err := current.Reconstruct(batch)
+			r.NoError(err)
```
RodrigoVillar:
```suggestion
			r.NoError(current.Reconstruct(batch))
```

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		_ = current.Root()
```
RodrigoVillar, line 129: A comment as to why we call `current.Root()` would be great here.

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+	rev, err := db.Revision(root)
+	r.NoError(err)
+	t.Cleanup(func() {
+		r.NoError(rev.Drop())
+	})
+
+	reconstructed, err := rev.Reconstruct(batch[committedKeys:firstBatchEnd])
+	r.NoError(err)
+	t.Cleanup(func() { _ = reconstructed.Drop() })
```
RodrigoVillar: Why not `r.NoError(reconstructed.Drop())`?

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+
+// Lock ordering for Reconstructed:
+//
+//	keepAliveHandle.mu (A)  before  rootMu (B)
+//
+// Every method that acquires both locks does so in A-then-B order:
+//
+//	Root()        — A.RLock  → B.Lock
+//	Reconstruct() — A.Lock   → B.Lock
+//
```
alarso16, line 40: This doesn't need to be in the godoc - locking is an implementation detail

ffi/reconstructed_test.go
rkuris: fixed in bf3fb425836e4100c9e312212aa41dd815010737

ffi/reconstructed_test.go
rkuris: fixed in bf3fb425836e4100c9e312212aa41dd815010737

ffi/reconstructed_test.go
rkuris: added in bf3fb425836e4100c9e312212aa41dd815010737

ffi/reconstructed_test.go
rkuris: fixed in bf3fb425836e4100c9e312212aa41dd815010737

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+	r.NoError(reconstructed.Reconstruct(batch[firstBatchEnd:secondBatchEnd]))
+	r.NotEqual(oldRoot, reconstructed.Root())
+
+	for i := range len(keys) {
+		got, err := reconstructed.Get(keys[i])
+		r.NoError(err)
+		r.Equal(vals[i], got)
+	}
+
+	r.NoError(reconstructed.Drop())
```
alarso16:
```suggestion
```
this is done in the cleanup

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+	// First Drop succeeds.
+	r.NoError(reconstructed.Drop())
+
+	// Second Drop is a no-op.
+	r.NoError(reconstructed.Drop())
+
+	// All operations return ErrDroppedReconstructed after Drop.
+	_, err = reconstructed.Get(keys[0])
+	r.ErrorIs(err, ErrDroppedReconstructed)
+
```
alarso16, line 179: This error doesn't need exported - but I see the other similar errors are. Should we be exporting this?

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+	r.NoError(err)
+	t.Cleanup(func() { r.NoError(rev.Drop()) })
+
+	reconstructed, err := rev.Reconstruct(batch[2:4])
+	r.NoError(err)
+
+	// Dump succeeds on a live view.
+	dot, err := reconstructed.Dump()
+	r.NoError(err)
+	r.NotEmpty(dot)
```
alarso16: I think this is out of scope for the test

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+	r.ErrorIs(err, ErrDroppedReconstructed)
+
+	_, err = reconstructed.Dump()
+	r.ErrorIs(err, ErrDroppedReconstructed)
+
+	err = reconstructed.Reconstruct(batch[:1])
+	r.ErrorIs(err, ErrDroppedReconstructed)
+}
+
+func TestReconstructedConcurrentGetAndDrop(t *testing.T) {
```
alarso16, line 193: Are you just testing that it doesn't panic?

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+	)
+
+	makeBatch := func(rng *rand.Rand, count int) []BatchOp {
+		batch := make([]BatchOp, 0, count)
+		for range count {
+			key := make([]byte, keyLen)
+			value := make([]byte, valueLen)
+			// an error is impossible as rng.Read is documented to never fail
+			_, _ = rng.Read(key)
+			_, _ = rng.Read(value)
```
RodrigoVillar: Why can't we use `r.NoError()` here? While these are anonymous functions, using `r` should be fine here as it's still within the scope of `BenchmarkReconstructChain()`.

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+// automatically if needed, but explicit calls are recommended.
+type Reconstructed struct {
+	*handle[*C.ReconstructedHandle]
+
+	rootMu  sync.Mutex
+	root    Hash
+	rootSet bool
+}
+
+// Root returns the root hash of the reconstructed view.
```
demosdemon, line 66: nit: Should document that root is still (intentionally) usable after the handle has been dropped since it deviates from others.

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,238 @@
...
+			batch = append(batch, Put(key, value))
+		}
+		return batch
+	}
+
+	generateBatches := func(seed int64) ([]BatchOp, [][]BatchOp) {
```
RodrigoVillar: This function signature could be simplified to `func(seed) []BatchOp { // ... }` as `initialItems == batchItems`. We could just index into the list.

ffi/reconstructed.go
```diff
@@ -0,0 +1,227 @@
...
+	newHandle, err := getReconstructedHandleFromResult(result, nil)
+	if err != nil {
+		// The old handle was consumed by the FFI call, so mark as dropped
+		// and disown the keep-alive lease so Database.Close is not blocked.
+		r.dropped = true
+		if r.keepAliveHandle.outstandingHandles != nil {
+			r.keepAliveHandle.outstandingHandles.Done()
+			r.keepAliveHandle.outstandingHandles = nil
+		}
+		return err
```
demosdemon, line 160: This re-implements a bit of the logic in `(*databaseKeepAliveHandle).disown(bool, func() error)`. But, I also understand why as it's only necessary to disown the handle if an error occurs whereas the other options are to always disown even if an error occurs and disown only if an error does not occur.

While my opinion on this is not strong, I would rather encapsulate this logic on the `databaseKeepAliveHandle` and not detached like this. I am also open to a more holistic combination of `databaseKeepAliveHandle` and `handle[T]`.

So, I'm approving since I have no other feedback and don't necessary want this to change right away but would re-review if it did.

ffi/reconstructed_test.go
rkuris: This just tests that Dump actually still produces some output on a reconstructed version. I think it's worth keeping.

ffi/reconstructed.go
rkuris: The generated godoc does not include this section, it starts on line 48 due to the blank line at line 47

Definitely an implementation detail but nevertheless helpful both from a review standpoint as well as notes on how the implementation actually works.

That said, if the convention is not to document this sort of thing, say the word and I'll nuke it.

ffi/reconstructed.go
rkuris: Created https://github.com/ava-labs/firewood/issues/1818 for tracking

ffi/reconstructed_test.go
rkuris: Not sure I'm following here. ErrDroppedReconstructed matches ErrDroppedRevision's exported convention as far as I can see?

ffi/reconstructed_test.go
rkuris: Yes. Documented in de6cef456e17a36ddeec71ddd9ba88deb9eb8202.

ffi/reconstructed_test.go
rkuris: We can, but IMO it's not needed, as the function can never fail and it reduces readability. Fixed in 0748378e484dc5d8c1543b600a1da1cbf8cb4960 anyway.

ffi/reconstructed.go
rkuris: Fixed in de6cef456e17a36ddeec71ddd9ba88deb9eb8202

ffi/reconstructed_test.go
rkuris: Changed in de6cef456e17a36ddeec71ddd9ba88deb9eb8202 with questionable improvement in maintainability and readability.

I typically prefer less DRY style coding in tests, see https://youtu.be/FcHUPqKRvxQ?si=rKeeS-v6o174miBR&t=802 for some reasonings and discussions.

ffi/reconstructed.go
alarso16: Ahhh I missed the space - my bad

ffi/reconstructed_test.go
alarso16: Why is it exported? Probably neither need to be exported. Do we want to allow consumers to depend on the error?

ffi/reconstructed_test.go
alarso16: Yeah, I just mean this test is checking that the drop functions work. I would make this a small separate test

ffi/reconstructed_test.go
```diff
@@ -0,0 +1,234 @@
...
+			key := make([]byte, keyLen)
+			value := make([]byte, valueLen)
+			_, err := rng.Read(key)
+			r.NoError(err)
+			_, err = rng.Read(value)
+			r.NoError(err)
+			batch = append(batch, Put(key, value))
+		}
+		return batch
+	}
```
RodrigoVillar: It may make sense to add this as a helper function in `firewood_test.go` for reusability purposes.

ffi/reconstructed_test.go
rkuris: Isn't this typically how it's handled? This means callers can use ErrorIs and friends to handle different error classes differently?

ffi/reconstructed_test.go
rkuris: fixed in 98cbfcd4b76e6331d40c377c4fcdbadf0a09008d

ffi/reconstructed_test.go
rkuris: done in 98cbfcd4b76e6331d40c377c4fcdbadf0a09008d

Issue #1763, opened 2026-03-11
Add keepAliveHandle for ChangeProof


bernard-avalabs:
ProposedChangeProof missing keepAliveHandle — unlike RangeProof, it holds a raw *Database pointer without incrementing outstandingHandles.

PR #1817, opened 2026-03-19
fix: Adding keepAlive to ProposedChangeProof

bernard-avalabs:
## Why this should be merged

Addresses https://github.com/ava-labs/firewood/issues/1763 by adding a keepAliveHandle in ProposedChangeProof in the same way as a RangeProof. Some unused pinners were also removed, and commitLock is held as part of CommitChangeProof.

## How this works

Creates a keepAliveHandle for ProposedChangeProof. This ensures the database is kept alive until the proof is committed or freed. 

## How this was tested

CI + 3 new keepalive-related tests.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -405,8 +405,8 @@ func (db *Database) root() Hash {
 }
 
 // LatestRevision returns a [Revision] representing the latest state of the database.
-// If the latest revision has root [EmptyRoot], it returns an error. The [Revision] must
-// be dropped prior to closing the database.
+// If the latest revision has root [EmptyRoot], it returns an error. The
+// [Revision] must be released with [Revision.Drop] before closing the database.
 //
 // This function conflicts with all other calls that access the latest state of the database,
 // and will lock for the duration of this function.
@@ -427,8 +427,9 @@ func (db *Database) LatestRevision() (*Revision, error) {
 }
 
 // Revision returns a historical revision of the database.
-// If the provided root does not exist (or is the [EmptyRoot]), it returns an error.
-// The [Revision] must be dropped prior to closing the database.
+// If the provided root does not exist (or is the [EmptyRoot]), it returns an
+// error. The [Revision] must be released with [Revision.Drop] before closing
+// the database.
 //
 // This function is thread-safe with all other operations.
 func (db *Database) Revision(root Hash) (*Revision, error) {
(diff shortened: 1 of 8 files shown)


ffi/proofs.go
```diff
@@ -427,13 +433,19 @@ func (db *Database) ProposeChangeProof(
...
-	return getProposedChangeProofFromProposedChangeProofResult(db, C.fwd_db_propose_change_proof(db.handle, args))
+
+	proposed, err := getProposedChangeProofFromProposedChangeProofResult(db, C.fwd_db_propose_change_proof(db.handle, args))
+	if err != nil {
+		return nil, err
+	}
+
+	// keep the database alive while the proof owns the embedded proposal
+	proposed.keepAliveHandle.init(&db.outstandingHandles)
+	runtime.SetFinalizer(proposed, (*ProposedChangeProof).Free)
```
alarso16, line 451: We've decided to attempt a `runtime.AddCleanup` in general. I didn't know the guarantees for range proofs when I added it to everything else (there's also a difference between `Free` and `Drop`), but IMO we should try and share the infrastructure.

we could punt this off to a TODO, but we should definitely try and fix this uniformity issue before v1.0.0

bernard-avalabs:
> Will you add a test?

Thanks for the suggestion. I just added 3 keepalive-related tests.

ffi/proofs.go
bernard-avalabs: I've added a TODO both here and in range proof to use runtime.AddCleanup in the future.

rkuris: Let's create an issue as well. TODO is for very low priority stuff.

ffi/proofs_test.go
```diff
@@ -660,6 +695,60 @@ func TestChangeProofFindNextKey(t *testing.T) {
...
+	r.ErrorIs(dbB.Close(oneSecCtx(t)), ErrActiveKeepAliveHandles)
+
+	// Free the proof (releases keep-alive)
+	r.NoError(proposed.Free())
+
+	// Database should now be closeable
+	r.NoError(dbB.Close(oneSecCtx(t)))
+}
+
+func TestProposedChangeProofCommitReleasesKeepAlive(t *testing.T) {
```
alarso16: This and the test above could be put in a table pretty easily

ffi/proofs.go
bernard-avalabs: Created issue: https://github.com/ava-labs/firewood/issues/1824

ffi/proofs_test.go
bernard-avalabs: Updated the tests to use a table.

Issue #1821, opened 2026-03-19
chore: Reduce churn on future PRs


rkuris:
Let's discuss how we can reduce churn on future golang PRs. Here's a claude analysis of a recent golang PR in firewood as a starting point.

# PR #1814 Review Analysis

PR [#1814](https://github.com/ava-labs/firewood/pull/1814) (feat(reconstruction): Add Go bindings and tests) required significant review effort — 44 reviewer comments across 36 threads, 40 author responses, all within a single day.

## Metrics by Reviewer

| Reviewer | Unique issues raised | Replies/follow-ups | Rounds to approval |
|---|---|---|---|
| **Copilot** | 9 | 0 | 5 passes (never approves) |
| **alarso16** | 11 | 4 | 3 (CHANGES_REQUESTED → APPROVED) |
| **RodrigoVillar** | 14 | 1 | 5 (4× COMMENTED → APPROVED) |
| **maru-ava** | 2 | 1 | 3 (COMMENTED only) |
| **demosdemon** | 2 | 0 | 1 (straight to APPROVED) |
| **rkuris (responses)** | — | 40 | — |

## Issue Classification

| Category | Count | Catchable by tooling? |
|---|---|---|
| Correctness / Bugs (finalizer panic, WaitGroup race, error-path leak) | 6 | Partially — `go vet`, `-race` |
| Concurrency / Lock ordering (deadlock risk, explain unlock/relock) | 4 | Partially — `-race`, `go-deadlock` |
| Code style / Idiomatic Go (`r.NoError`, `errgroup`, `defer`, struct ordering) | 8 | Partially — `revive`, team conventions |
| Error handling / Lint (errcheck, explicit `_ = ...`) | 5 | **Yes** — `errcheck`, `golangci-lint` |
| Magic numbers ("Why 32?", use constants) | 4 | **Yes** — `mnd` linter |
| Test design / Scope (unclear test purpose, missing cleanup, separate tests) | 7 | No — requires human judgment |
| Documentation / Comments (inaccurate comments, missing docs) | 5 | No — requires human judgment |
| Refactoring / DRY | 3 | No |
| API design (export decisions, wrapping vs embedding) | 2 | No |

Of the 44 comments, approximately 19 (43%) were catchable by existing or easily-added tooling. The remaining 25 required human judgment.

### A. Add `mnd` (magic number detector) to golangci-lint

Would catch 4 comments automatically. Add to `.golangci.yaml`:

```yaml
- mnd
```

[...]

PR #1822, opened 2026-03-19
chore(ci): Enable revive

rkuris:
## Why this should be merged

Enable the revive checker to help with churn. Related to https://github.com/ava-labs/firewood/issues/1821

## How this works

Added a new checker to improve coverage of magic numbers and other pedantic requirements for golang.

### Code fixes

proofs.go: Fixed inconsistent receiver names (proof → p) for inconsistent methods; added doc comments to exported CommitChangeProof and FindNextKey; wrapped long function signature

metrics.go: Added doc comments to exported Gatherer type and Gather method

reconstructed.go: Added doc comment to ErrDroppedReconstructed

revision.go: Added doc comments to exported error vars

iterator_test.go: i += 1 → i++

firewood_test.go: Renamed block_307_items/block_308_items to block307Items/block308Items, removed redundant type annotations, removed redundant defer in loop, added //revive:disable directives for hex test data

replay_test.go: Wrapped long line; added //revive:disable for complex but clear dispatch function

proofs_test.go: Split long hex string literal

eth_compatibility_test.go and merkle_compatibility_test.go: Removed duplicate imports, fixed ffi. → firewood. references

### Test file constants

added named constants across 8 test files:

```
firewood_test.go — 16 constants (counts, format strings, sentinel values)
proofs_test.go — 8 constants + testKey var (entry counts, byte values, key length)
iterator_test.go — 12 constants (per-test key counts, batch sizes, format string)
metrics_test.go — 2 constants (key count, timeout)
reconstructed_test.go — 5 constants (seed, key counts)
replay_test.go — 2 constants (max commits, key count)
eth_compatibility_test.go — 4 constants (balance values, fuzz params)
merkle_compatibility_test.go — 2 constants (fuzz steps, seeds)
```

## How this was tested

CI

## Breaking Changes

None

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -108,29 +108,6 @@ The `fwdctl` tool provides command-line operations on databases. See `fwdctl/REA
 
 For more information on coding conventions and constraints, please refer to [CONTRIBUTING.md](./CONTRIBUTING.md)
 
-### Go (FFI) Constants and Magic Numbers
-
-When extracting magic numbers or repeated string literals into named constants
-in Go code (especially to satisfy `revive` linter rules like `add-constant`):
-
-1. **Always replace the literal with the constant.** Defining a constant but
-   leaving the original magic number in place creates an unused-constant error
-   and defeats the purpose. After adding a constant, grep for every occurrence
-   of the literal value and replace them all.
-
-2. **Scope constants as narrowly as possible.** If a constant is only used
-   inside a single function, declare it as a `const` inside that function
-   rather than at package level. Package-level constants should be reserved
-   for values shared across multiple functions.
-
-3. **Avoid duplicate imports.** Never import the same package twice (e.g.
-   once bare and once with an alias). Use a single import with the alias
-   that the file actually needs, and update all references to match.
-
-4. **Run `gofumpt`** after modifying `const` blocks — alignment changes and
-   trailing blank lines inside `const ( ... )` will cause `gofmt`/`gofumpt`
-   lint failures.
-
 ## Commit and PR Title Convention
 
 Commit messages and PR titles **must** follow the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,29 @@ We generally follow the same rules that `cargo fmt` and `cargo clippy` will repo
 
 By default, we prohibit bare `unwrap` calls and index dereferencing, as there are usually better ways to write this code. In the case where you can't, please use `expect` with a message explaining why it would be a bug, which we currently allow. For more information on our motivation, please read this great article on unwrap: [Using unwrap() in Rust is Okay](https://blog.burntsushi.net/unwrap) by [Andrew Gallant](https://blog.burntsushi.net).
 
+### Go (FFI) Constants and Magic Numbers
+
+When extracting magic numbers or repeated string literals into named constants
+in Go code (especially to satisfy `revive` linter rules like `add-constant`):
+
+1. **Always replace the literal with the constant.** Defining a constant but
+   leaving the original magic number in place creates an unused-constant error
+   and defeats the purpose. After adding a constant, grep for every occurrence
+   of the literal value and replace them all.
+
+2. **Scope constants as narrowly as possible.** If a constant is only used
+   inside a single function, declare it as a `const` inside that function
+   rather than at package level. Package-level constants should be reserved
+   for values shared across multiple functions.
+
+3. **Avoid duplicate imports.** Never import the same package twice (e.g.
+   once bare and once with an alias). Use a single import with the alias
+   that the file actually needs, and update all references to match.
+
+4. **Run `gofumpt`** after modifying `const` blocks — alignment changes and
+   trailing blank lines inside `const ( ... )` will cause `gofmt`/`gofumpt`
+   lint failures.
+
 ## [Where can I ask for help]?
 
 If you have questions or need help, please post them as issues in the [issue tracker](https://github.com/ava-labs/firewood/issues). This allows the community to benefit from the discussion and helps us maintain a searchable knowledge base.


rkuris:
@copilot can you review this?

rkuris:
> @rkuris I've opened a new pull request, #1823, to work on those changes. Once the pull request is ready, I'll request review from you.

I closed this. I just added you as a reviewer, I thought mentioning you would work but no.

ffi/replay_test.go
```diff
@@ -280,6 +286,8 @@ type replayConfig struct {
 
 // applyReplayLogs applies replay logs to a database.
 // Returns the number of commits applied and any error encountered.
+//
+//revive:disable:cognitive-complexity // straightforward switch-case dispatch over operation types
 func applyReplayLogs(db *Database, logs []replayLog, cfg replayConfig) (int, error) {
```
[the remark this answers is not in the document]
rkuris: fixed in 8eeb848ef09bb499e83f34712220780fd886cd06

ffi/proofs.go
```diff
@@ -497,7 +499,7 @@ func (db *Database) VerifyAndCommitChangeProof(
 //
 // FindNextKey can only be called after a successful call to [*Database.VerifyChangeProof] or
 // [*Database.VerifyAndCommitChangeProof].
-func (p *ChangeProof) FindNextKey() (*NextKeyRange, error) {
+func (proof *ChangeProof) FindNextKey() (*NextKeyRange, error) {
 	return getNextKeyRangeFromNextKeyRangeResult(C.fwd_change_proof_find_next_key(p.handle))
```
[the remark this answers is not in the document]
rkuris: fixed in 8eeb848ef09bb499e83f34712220780fd886cd06

.github/workflows/ci.yaml
```diff
@@ -264,6 +264,11 @@ jobs:
         with:
           version: latest
           working-directory: ffi
+      - name: Run revive
+        working-directory: ffi
+        run: |
+          go install github.com/mgechev/revive@latest
```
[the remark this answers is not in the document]
rkuris: fixed in 8eeb848ef09bb499e83f34712220780fd886cd06 (used sha)

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -28,6 +27,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	initialAccountBalance = 100
```
alarso16: nit: put these closer to their uses

ffi/tests/firewood/merkle_compatibility_test.go
```diff
@@ -30,6 +29,8 @@ const (
 	valSize      = 4
 	maxBatchSize = 5
 	maxNumKeys   = 10_000
+	maxFuzzSteps = 100
+	numFuzzSeeds = 5
```
alarso16: same here

ffi/firewood_test.go
```diff
@@ -32,6 +32,26 @@ const (
...
+	numTestEntries        = 20
+	halfTestItems         = 5
+	numRevisionsToKeep    = 5
+	numTestBatchOps       = 3
+	numTestHandles        = 2
+	numKVParts            = 2
+	numReapingTestEntries = 40
+
+	// Format string constants for test assertions
+	fmtGet          = "Get(%d)"
```
alarso16: I do not think this improves readability

ffi/firewood_test.go
```diff
@@ -32,6 +32,26 @@ const (
 	emptyEthhashRoot  = "56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"
 	emptyFirewoodRoot = "0000000000000000000000000000000000000000000000000000000000000000"
 	errWrongParent    = "The proposal cannot be committed since it is not a direct child of the most recent commit. "
+
+	// Test size constants
+	numTestItems          = 10
```
alarso16: I don't think that dumping all the constants used in this file into a block is intuitive. Do all tests use the same number of keys? What's the difference between items and entries and batchops? The linter might be going overboard here

ffi/firewood_test.go
```diff
@@ -104,15 +124,19 @@ func TestMain(m *testing.M) {
 		case "cgocheck=1":
 			hasCgoCheck = true
 		case "cgocheck=0":
-			fmt.Fprint(os.Stderr, "GODEBUG=cgocheck=0; MUST be 1 for Firewood cgo tests")
+			if _, err := fmt.Fprint(os.Stderr, "GODEBUG=cgocheck=0; MUST be 1 for Firewood cgo tests"); err != nil {
```
alarso16: I don't think this will ever error unless you're doing something wrong piping your output or something.

ffi/metrics.go
```diff
@@ -27,8 +27,10 @@ import (
 
 var _ prometheus.Gatherer = (*Gatherer)(nil)
 
+// Gatherer implements [prometheus.Gatherer] by collecting metrics from the Firewood FFI layer.
```
alarso16: This could have been caught by the `godoc` linter in golangci.yaml

ffi/metrics_test.go
```diff
@@ -19,6 +19,11 @@ import (
 	dto "github.com/prometheus/client_model/go"
 )
 
+const (
+	metricsTestKeys      = 10
+	metricsClientTimeout = 10 * time.Second
```
alarso16: This is only used once, why not just make it a local constant for the test?

ffi/proofs.go
```diff
@@ -497,8 +499,8 @@ func (db *Database) VerifyAndCommitChangeProof(
 //
 // FindNextKey can only be called after a successful call to [*Database.VerifyChangeProof] or
 // [*Database.VerifyAndCommitChangeProof].
-func (p *ChangeProof) FindNextKey() (*NextKeyRange, error) {
-	return getNextKeyRangeFromNextKeyRangeResult(C.fwd_change_proof_find_next_key(p.handle))
+func (proof *ChangeProof) FindNextKey() (*NextKeyRange, error) {
```
alarso16: Using a single (or a couple) letter is conventional for `self`-style identifiers. Consider `p` everywhere, or `pr` or `cp`

.github/workflows/ci.yaml
```diff
@@ -264,6 +264,11 @@ jobs:
         with:
           version: latest
           working-directory: ffi
+      - name: Run revive
+        working-directory: ffi
+        run: |
+          go install github.com/mgechev/revive@815ffde4de8dcb31a2de91efd6c6795d75e24380 # v1.15.0
```
alarso16: revive is a standard go linter, that can also be configured in the `.golangci.yaml` file. From some very brief research (and looking at the new config file), their functionality is mostly the same

rkuris:
> Although the spirit of the PR is in the right place, I suspect that Claude may have been a little vigorous in moving things to global constants. I think adding a rule like that isn't actually a code improvement, but just another rule to follow. There's a lot of style changes in this PR, and I'm on board for most of them

Maybe, but there was churn in that PR requesting making constants by two reviewers. I would rather have the linter prohibit it than have reviewers sometimes decide to reject a PR on the basis that it wasn't done. The cost to the reviewer is much higher (second round of reviews) than this rule enforces.

ffi/firewood_test.go
rkuris: I feel very strongly that we need this change, and is perhaps the motivator for this entire PR.

This function has a higher risk of failing than the #1814's `rand.Read` method that three different reviewers commented should be handled, despite the function being documented as never erroring. Even a comment that says that was insufficient for one of the reviewers:

<details><summary>Details of PR 1814's churn over rand.Read</summary>
<p>
<a href="https://github.com/ava-labs/firewood/pull/1814#discussion_r2961404379">
github copilot's review</a><br/>
<a href="https://github.com/ava-labs/firewood/pull/1814#discussion_r2961496332">@maru-ava's review</a><br/>
<a href="https://github.com/ava-labs/firewood/pull/1814#discussion_r2962086982">@RodrigoVillar's review</a><br/>
</p>
</details>

I could also pull the 3 diffs I applied in the process.

ffi/firewood_test.go
rkuris: Different reviewers have different tolerances for magic numbers. Some are extremely strict and others are less strict. The major reason for this PR is to reduce churn.

If you'd like to see the pedantic school of thought that all constants should have a description, here's a video: https://www.youtube.com/watch?v=x9PFhEfIuE4

Since this opinion is shared by some reviewers, let's just make it a rule on the go code to avoid the chance of finding one of those reviewers later 😄

ffi/tests/eth/eth_compatibility_test.go
rkuris: Fixed in dd407d1b62a615f235b7a21eca1eeab09ceff81f

ffi/tests/firewood/merkle_compatibility_test.go
rkuris: Fixed in dd407d1b62a615f235b7a21eca1eeab09ceff81f

ffi/firewood_test.go
rkuris: I don't fully disagree, but other reviewers might. Let's keep this rule, as there is considerable precedent that it's worthwhile:

https://stackoverflow.com/questions/4234129/should-i-use-constants-instead-of-strings-even-if-the-strings-are-only-ever-used#:~:text=The%20advantage%20of%20using%20constants,likely%20to%20pick%20up%20certain

More references on request.

One could argue that `fmtGet` is insufficiently meaningful instead, however, and could use a longer more descriptive name, but I'm not convinced that will help much.

ffi/metrics.go
rkuris: But it didn't. Revive appears to be more thorough.

ffi/proofs.go
rkuris: The thing actually being fixed here was an inconsistent use of the receiver. Some callers used p and others used proof.

Do you find `p` more readable? I don't, but if you do, revive supports adding a rule to restrict receiver names via a regex to avoid future churn, which is a better fix.

.github/workflows/ci.yaml
rkuris: revive caught stuff that golangci didn't, so I'd rather try and use both to avoid churn. See the parent ticket for more details.

ffi/metrics_test.go
rkuris: Fixed in 699d72eb70872fef87814a35425922a6a42a36c5

For the more permanent fix, I didn't see a way to configure revive to catch this, but I can check whether a custom rule can catch the case where constants used only one time could be local constants inside their corresponding function, otherwise:

1. we should clearly document that this is a coding convention
2. mention this in AGENTS.md
3. fix all other instances of it

For the last one, there weren't any other instances of it in firewood so it does seem like we need a rule. Any thoughts?

ffi/proofs.go
rkuris: YAY revive does have that option. Fixed in f92e05a008c4e29c526a5226304179943f06f30d

ffi/proofs.go
```diff
@@ -731,7 +733,9 @@ func getVerifiedChangeProofFromVerifiedChangeProofResult(result C.VerifiedChange
 	}
 }
 
-func getProposedChangeProofFromProposedChangeProofResult(db *Database, result C.ProposedChangeProofResult) (*ProposedChangeProof, error) {
+func getProposedChangeProofFromProposedChangeProofResult(
+	db *Database, result C.ProposedChangeProofResult,
+) (*ProposedChangeProof, error) {
 	switch result.tag {
 	case C.ProposedChangeProofResult_NullHandlePointer:
```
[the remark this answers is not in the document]
rkuris: It isn't misformatted and passes CI, so I have no idea why this is considered incorrect. Looks like a false positive.

ffi/replay_test.go
```diff
@@ -318,7 +326,8 @@ func applyReplayLogs(db *Database, logs []replayLog, cfg replayConfig) (int, err
 			case op.ProposeOnProposal != nil:
 				parent, ok := proposals[op.ProposeOnProposal.ProposalID]
 				if !ok {
-					return totalCommits, fmt.Errorf("ProposeOnProposal: unknown parent proposal id %d", op.ProposeOnProposal.ProposalID)
+					return totalCommits, fmt.Errorf(
+						"ProposeOnProposal: unknown parent proposal id %d", op.ProposeOnProposal.ProposalID)
```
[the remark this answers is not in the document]
rkuris: It isn't misformatted and passes CI, so I have no idea why this is considered incorrect. Looks like a false positive.

ffi/firewood_test.go
```diff
@@ -240,11 +268,13 @@ func TestNodeHashAlgorithmValidation(t *testing.T) {
 	case ethereumOnly:
 		r.NoError(dbEth.Close(oneSecCtx(t)))
 		r.Error(errMerkledb)
-		r.ErrorContains(errMerkledb, "node store hash algorithm mismatch: want to initialize with MerkleDB, but build option is for Ethereum")
+		r.ErrorContains(errMerkledb,
+			"node store hash algorithm mismatch: want to initialize with MerkleDB, but build option is for Ethereum")
 	case merkledbOnly:
```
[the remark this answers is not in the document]
rkuris: It isn't misformatted and passes CI, so I have no idea why this is considered incorrect. Looks like a false positive.

.github/workflows/ci.yaml
alarso16: https://golangci-lint.run/docs/linters/configuration/#revive
You have to explicitly enable/disable the revive rules. It is integrated with golangci-lint as a sort-of megalinter. I don't think we need a separate CI job when you can add the additional config into the already-existing linter (these docs link to the repo above, it really is the same)

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -167,6 +166,8 @@ func (tr *merkleTriePair) commit() {
 	tr.require.NoError(err)
 }
 
+const initialAccountBalance = 100
```
alarso16: I still think this is scoped incorrectly. I would _totally_ keep this as a magic number, since it's completely arbitrary and doesn't affect anything, but if you wanted it defined as a constant, why not move it within the function that uses it? (This applies to a lot of the constants)

ffi/metrics.go
alarso16: https://golangci-lint.run/docs/linters/configuration/#godoclint

We just don't have it sufficiently configured. The basic set is pretty lame, doesn't do much

ffi/proofs_test.go
```diff
@@ -16,8 +16,29 @@ const (
...
+	numEntries       = 100
+	largeNumEntries  = 10000
+	mediumNumEntries = 50
+	smallNumEntries  = 10
+	// Entry count for the empty change proof range test; must be odd so
+	// the split at 5 leaves a non-empty second half.
+	emptyRangeNumEntries = 9
+
+	// Byte used to XOR-flip a root hash byte to produce an invalid root.
+	flipByte = 0xFF
```
alarso16: This isn't "magic" - decreasing readability moving here IMO

ffi/proofs_test.go
```diff
@@ -16,8 +16,29 @@ const (
...
+
+	// Length of a merkle-trie key in bytes.
+	keyLen = 32
+
+	// Proof length limit used in TestVerifyEmptyChangeProofRange.
+	emptyRangeProofLen = 5
```
alarso16: Only used once? I don't think it needs to be global (this applies below too

ffi/tests/eth/eth_compatibility_test.go
```diff
@@ -13,7 +13,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ava-labs/firewood-go-ethhash/ffi"
 	firewood "github.com/ava-labs/firewood-go-ethhash/ffi"
```
alarso16, line 16: Out of scope, but the `firewood` alias here doesn't do anything - it just might alleviate confusion for consumers. In avalanchego I actually alias it to ffi

ffi/metrics.go
rkuris: Fixed in fe3fe48688da770a4837108243a381e7acb02fd2

.github/workflows/ci.yaml
rkuris: 🥇 good suggestion

Fixed in fe3fe48688da770a4837108243a381e7acb02fd2

ffi/firewood_test.go
```diff
@@ -240,11 +268,13 @@ func TestNodeHashAlgorithmValidation(t *testing.T) {
...
 		r.NoError(dbMerkledb.Close(oneSecCtx(t)))
 		r.Error(errEth)
-		r.ErrorContains(errEth, "node store hash algorithm mismatch: want to initialize with Ethereum, but build option is for MerkleDB")
+		r.ErrorContains(errEth,
+			"node store hash algorithm mismatch: want to initialize with Ethereum, but build option is for MerkleDB")
```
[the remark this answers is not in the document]
rkuris: This is a false positive

ffi/proofs.go
```diff
@@ -397,7 +397,7 @@ func (db *Database) ChangeProof(
 
 // VerifyChangeProof verifies the provided change [proof] proves the changes
 // between [startRoot] and [endRoot] for keys in the range [startKey, endKey].
-func (proof *ChangeProof) VerifyChangeProof(
+func (p *ChangeProof) VerifyChangeProof(
 	startRoot, endRoot Hash,
 	startKey, endKey Maybe[[]byte],
```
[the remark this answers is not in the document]
rkuris: Edited PR description

ffi/tests/eth/eth_compatibility_test.go
rkuris: fixed in dd407d1b

ffi/proofs_test.go
rkuris: Fixed in dd407d1b

ffi/proofs_test.go
rkuris: Fixed in 7b3ab962a87fdddb130c95c670b65d17bfb1064a

rkuris:
> I still don't enjoy moving arbitrary, locally used test values into global constants (for the prod code, I think it's helpful). If other people disagree, I'll digress. You definitely don't need the extra CI job though

Fixed the extra CI job. Based on the review of my prior PR, other people do care, so let's just make it a rule to reduce churn. I think there is a longer discussion about test code quality which we can discuss offline or during our offsite/retreat.

ffi/tests/eth/eth_compatibility_test.go
rkuris: I think we have more than one `ffi` here so it's a slightly different use case. Yell if you want me to pick a different name though.

ffi/proofs_test.go
```diff
@@ -165,7 +180,7 @@ func TestRangeProofNonExistentRoot(t *testing.T) {
 	r.NoError(err)
 
 	// create a bogus root
-	root[0] ^= 0xFF
+	root[0] ^= 0xFF //nolint:revive // add-constant: 0xFF is self-documenting as a byte flip
```
maru-ava, line 182: (No action required) Why is it preferable to add an exclusion and a comment justifying that exclusion instead of a constant?

ffi/proofs_test.go
```diff
@@ -351,6 +366,7 @@ func TestRangeProofCodeHashes(t *testing.T) {
 			r.NoError(err, "%T.CodeHashes()", proof)
 			r.Equal(codeHash, h)
 		} else {
+			//nolint:forbidigo // checking error substring, not identity
```
maru-ava: Maybe provide a more substantive rational e.g. 'this call site doesn't return an error with an identify'?

ffi/proofs_test.go
rkuris: @alarso16 suggested that it is less readable

alarso16: Ugh yeah not great either

.github/workflows/ci.yaml
alarso16: Why is the CI job still present?

ffi/tests/eth/eth_compatibility_test.go
alarso16: I think you misunderstood. Why is it a GLOBAL constant?
```go
func (tr *merkleTriePair) createAccount() {
    const initialAccountBalance=100
```

ffi/proofs_test.go
JuanLeon2: FWIW I am with Austin that arbitrary numbers not receiving a name is documentation of sorts which signals the number is not special.  Especially when accompanied by a comment that has the word "bogus".

I realize that Ron wants to lay down a law that is decidable in advance (not subject to case-by-case interpretation).   Maybe the argument in the previous paragraph cannot be codified easily in linter.  In this case perhaps it is better to live with the law, but only if we expect litigation.  Perhaps we can change society to be less litigious.

ffi/tests/eth/eth_compatibility_test.go
rkuris: It's not global -- it would have to start with an uppercase letter, but it's package-level and could be scoped to the function.

Fixed in a837a561895dfc2a07e07da9156e6234051477da along with a few others that followed this pattern. I looked for a way for the linter to enforce this, but I couldn't find one.

Instead, I added instructions for AI agents to catch this in db5797219b2e5f315885499fa094fb7444d20690

ffi/metrics_test.go
rkuris: Fixed in db5797219b2e5f315885499fa094fb7444d20690

ffi/proofs_test.go
rkuris: Fixed in 52a3bb50087acdee44b6dc19a3af30772928c5bb

rkuris:
> Why are there no changes to `firewood_test.go`?

I'm not sure I'm fully following. I did notice we still had some leftover revive stuff in ci.yaml and fixed that up in 52a3bb50087acdee44b6dc19a3af30772928c5bb, but I'm not sure if that resolves your concern?

AGENTS.md
```diff
@@ -108,6 +108,25 @@ The `fwdctl` tool provides command-line operations on databases. See `fwdctl/REA
...
+
+2. **Scope constants as narrowly as possible.** If a constant is only used
+   inside a single function, declare it as a `const` inside that function
+   rather than at package level. Package-level constants should be reserved
+   for values shared across multiple functions.
+
+3. **Run `gofumpt`** after modifying `const` blocks — alignment changes and
+   trailing blank lines inside `const ( ... )` will cause `gofmt`/`gofumpt`
+   lint failures.
+
```
RodrigoVillar: Is this necessary? I would think any errors that the agent gets from running the linter would provide sufficient context for what needs to be fixed. Adding this section feels heavy-headed for a single rule.

ffi/firewood_test.go
```diff
@@ -31,7 +31,28 @@ const (
 	insert100Key      = "100"
 	emptyEthhashRoot  = "56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"
 	emptyFirewoodRoot = "0000000000000000000000000000000000000000000000000000000000000000"
-	errWrongParent    = "The proposal cannot be committed since it is not a direct child of the most recent commit. "
+	errWrongParent    = "The proposal cannot be committed since it is not a " +
```
RodrigoVillar, line 34: Why are we concatenating two strings here?

ffi/reconstructed_test.go
```diff
@@ -12,6 +12,24 @@ import (
...
+	firstReconstructedBatch = 2
+
+	// kvCountDump is the number of key-value pairs generated for the dump
+	// and drop-then-use tests.
+	kvCountDump = 4
+
+	// kvCountConcurrent is the number of key-value pairs generated for the
+	// concurrent get-and-drop test.
+	kvCountConcurrent = 8
+)
```
RodrigoVillar: Could these variables instead be scoped to the functions they're used in?

ffi/metrics_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	dto "github.com/prometheus/client_model/go"
 )
 
+const metricsTestKeys = 10
```
RodrigoVillar, line 22: Since this const is used twice, could we instead scope this value to the functions it's used in instead (it's only used twice, so I don't think the benefit of deduplication follows here).

ffi/tests/eth/eth_compatibility_test.go
RodrigoVillar: Why are we importing `github.com/ava-labs/firewood-go-ethhash/ffi` twice?

ffi/iterator_test.go
```diff
@@ -10,6 +10,20 @@ import (
...
+const (
+	iterTestBasicCount          = 100
+	iterTestOnRootCount         = 240
+	iterTestOnProposalCount     = 240
+	iterTestAfterCommitCount    = 10
+	iterTestUpdateCount         = 10
+	iterTestDoneCount           = 18
+	iterTestDoneBatchSize       = 5
+	iterTestOutlivesRevCount    = 30
+	iterTestOutlivesProposalCnt = 4
```
RodrigoVillar: I'm not sure how I feel about the usage of global test values. While I acknowledge their utility in minimizing duplication, writing new tests to use these values seems like an additional burden that I need to account for (i.e. I can imagine myself writing a new test where instead of the main focus being writing the test itself, I'm partially distracted by trying to get these test values to fit the test I'm writing).

RodrigoVillar (commented):
Left some comments, I like the direction that this PR is going in though!

.github/workflows/ci.yaml
rkuris: It was removed in this changeset: https://github.com/ava-labs/firewood/pull/1822/changes/52a3bb50087acdee44b6dc19a3af30772928c5bb

AGENTS.md
rkuris: Step 2 is the important one. @alarso16 prefers constants scoped as narrowly as possible. The other steps just support this step -- AI tends to create these lint failures and should be more careful to not fail that step when making suggestions. Step 1 is needed so it doesn't just suppress the warning. Step 3 makes it easier to know what needs to happen after making these const changes.

ffi/firewood_test.go
rkuris: I think you're asking "which linter rule forces us to break these long strings into separate concatenated strings"

It's this one:

```
+        - name: line-length-limit
+          disabled: false
+          arguments: [132]
```

But you might be asking:

"I would reject a PR that broke up a line like this, so this rule is stupid" -- in that case, we need to remove the lint rule and document that line length is never a valid reason to reject a PR.
"nobody has ever cared about go code line length, why are we caring now?" -- similar, also remove the lint rule and document (note that rust does care and the formatter will break these things up)
"I might reject a PR that had long lines that wasn't broken up like this" -- in this case no action is required, the rule is working as expected

ffi/reconstructed_test.go
rkuris: This ties directly into the AI instructions you commented about earlier:
> *Scope constants as narrowly as possible.** If a constant is only use inside a single function, declare it as a `const` inside that function rather than at package level. Package-level constants should be reserved for values shared across multiple functions.

And, of course, I didn't audit the code all the other cases of this. Good catch, fixed in 9faba24b4223b30ebc1b8c2eb06b4fa5679fa0ef.

ffi/metrics_test.go
rkuris: Is this something you'd typically reject a PR for? If so, we'll need some agent instructions to explain the exact limit instead of what was already added.

ffi/firewood_test.go
RodrigoVillar: I think context is important here - for example take this function:

```go
func (eth *Ethereum) StateAtNextBlock(ctx context.Context, parent *types.Block, nextBlock *types.Block, reexec uint64, base *state.StateDB, readOnly bool, preferDisk bool) (*state.StateDB, tracers.StateReleaseFunc, error)
```

In the above, I would almost certainly make a request to format it to:

```go
func (eth *Ethereum) StateAtNextBlock(
	ctx context.Context, 
	parent *types.Block, 
	nextBlock *types.Block, 
	reexec uint64, 
	base *state.StateDB, 
	readOnly bool, 
	preferDisk bool,
) (*state.StateDB, tracers.StateReleaseFunc, error)
```

In the latter, legibility is significantly better as I'm able to parse the function parameters (and its individual fields) from the return arguments. Whereas going from:

```go
errWrongParent    = "The proposal cannot be committed since it is not a direct child of the most recent commit. "
```

```go
errWrongParent    = "The proposal cannot be committed since it is not a " + 
		"direct child of the most recent commit. "
```

It's not clear to me what the improvement to legibility is. To answer your question though, I would say: "I might reject a PR that had long lines that wasn't broken up like this *unless doing so would be dubious*" .

ffi/tests/eth/eth_compatibility_test.go
rkuris: Fixed here and in merkle_compatibility_test.go, also updated agent instructions in c8099419a5c6c69b154340b42ef1deb60ff52d4c

ffi/iterator_test.go
rkuris: Any suggestions on how to make this better?

RodrigoVillar: I would remove these global variables and instead scope them to the function(s) they're used in.

ffi/.golangci.yaml
```diff
@@ -113,24 +113,92 @@ linters:
...
         - pattern: ^sort\.(Slice|Strings)$(# the slices package should be used instead)?
       # Exclude godoc examples from forbidigo checks.
       exclude-godoc-examples: false
     revive:
       rules:
+        # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#add-constant
+        - name: add-constant
+          disabled: false
+          arguments:
+            - allowInts: "0,1"
```
alarso16, line 137: I would personally use some regex to exclude test functions. IMO, test functions to use constants is typically, but not always, better practice. For example, the `testThreeKeys` constant is not better than putting `3`. I think that if we can very frequently ignore the linter, we shouldn't have the linter.

I promise this is the last time I'll fight for this change. If I'm the only one that feels this way, I'll digress

AGENTS.md
RodrigoVillar: Ah understood. In that case, could we move this to `CONTRIBUTING.md` (or even a new `STYLE.md` file)? This linting guideline is applicable to both AI agents and humans (Firewood team + external contributors). Furthermore, we already tell the agent to look to `CONTRIBUTING.md` for style:

https://github.com/ava-labs/firewood/blob/3da98c89bbd5912b22538f7a5ff6dbc488913f0e/AGENTS.md?plain=1#L109

ffi/.golangci.yaml
```diff
@@ -113,24 +113,92 @@ linters:
...
+              allowStrs: '"","%w: %w"'
         # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bool-literal-in-expr
         - name: bool-literal-in-expr
           disabled: false
+        # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#call-to-gc
+        - name: call-to-gc
+          # Test code intentionally calls runtime.GC() to exercise finalizers.
+          disabled: true
```
alarso16: By default, because you explicitly enabled some linters, all others are disabled:
https://golangci-lint.run/docs/linters/configuration/#revive
To be fair, this configuration defaults is giving me a headache. But the short answer is if you explicitly disabled something, you can remove it entirely, unless you want to keep the justification here for further consideration

ffi/metrics_test.go
RodrigoVillar: I would be okay with adding this rule to the style guide (i.e. arbitrary testing values should be scoped to the functions they're used in).

ffi/firewood_test.go
rkuris: > I'm able to parse the function parameters

I think this rule is covered by another linter rule so arguments always get split out like this.

> I might reject a PR that had long lines that wasn't broken up like this unless doing so would be dubious

OK so the lint rule makes sense then. If it's dubious for some reason, the author can add an lint exception and add an explanation as to why.

I think that means there is nothing actionable here, but tell me if that's wrong.

ffi/metrics_test.go
rkuris: Isn't this already covered with rule 2, or is there something else you think should be added?

> 2. **Scope constants as narrowly as possible.** If a constant is only used inside a single function, declare it as a `const` inside that function rather than at package level. Package-level constants should be reserved for values shared across multiple functions.

ffi/iterator_test.go
rkuris: This was fixed in 9faba24b4223b30ebc1b8c2eb06b4fa5679fa0ef I think, along with the instructions to the agent, unless I'm not following.

ffi/firewood_test.go
RodrigoVillar: > OK so the lint rule makes sense then. If it's dubious for some reason, the author can add an lint exception and add an explanation as to why.

This makes sense, thanks for clarifying!

ffi/.golangci.yaml
rkuris: Fair point. Having them here does raise the awareness that we could also add this, but the same information could be taken from the revive docs instead. Fixed in 238dd1cb.

ffi/metrics_test.go
RodrigoVillar: I would go further with rule 2 in that test-only package-level constants should be avoided at all costs (with the rationale explained in https://github.com/ava-labs/firewood/pull/1822#discussion_r2974710637).

ffi/firewood_test.go
```diff
@@ -1619,26 +1643,44 @@ func TestDump(t *testing.T) {
...
+	":f8450180a00000000000000000000000000000000000000000000000000000000000000000" +
+	"a0a33ec8100b06cfbacc612bc2baa4d36f01d5d52df97bcef04f54d5b8ceccbd9280\n" +
+	"f22bb46edf31af855938befaa870ed3d86a4ad93a9ecf7c63ceaa80daea9ac4d" +
+	":f84e808906348c71086eb3ac00" +
+	"a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421" +
+	"a0611445c10cb8404ad2103d510e139c3ace61b5acec80505bd1f5870528f587d780\n" +
+	"71645710251b71938795615f63f7bcf27fd5aa4ffbaeb7ace16cbd1c76a41e42" +
+	":f84d128865e452b75c31a000" +
+	"a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421" +
+	"a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a47080\n"
```
RodrigoVillar, line 1683: Could we revert this diff by adding a `nolint` comment?

ffi/.golangci.yaml
rkuris: The only thing I care about is no human linting, which includes "you added a magic number here, why is this not a constant" or "why did you pick this specific number? make it a constant and document it" which happened more than once over the past week and created some churn (examples on request).

Generally I agree with you in that tests shouldn't be so DRY and should not cause the reader to have to look up some refactored constant, but as mentioned many go reviewers feel otherwise.

One option is to take this out completely, another is to remove it only in test files, and a third is to just leave it as is. I'm fine either way, as long as we can have some agreement that a constant pull-out is usually not a valid cause for rejecting a PR and we document as such (perhaps in the linter file even).

ffi/firewood_test.go
```diff
@@ -31,7 +31,26 @@ const (
...
+	errWrongParent    = "The proposal cannot be committed since it is not a " +
+		"direct child of the most recent commit. "
+
+	// Numeric test constants
+	testNumKeys       = 10
+	testNumKeysDouble = 20
+	testNumKeysQuad   = 40
+	testFiveRevisions = 5
+	testTwoRevisions  = 2
+	testThreeKeys     = 3
```
RodrigoVillar, line 43: Similar to the concerns pointed out in https://github.com/ava-labs/firewood/pull/1822#discussion_r2974710637, could we remove these package-level constants?

alarso16, line 43: I think all these constants should be locally declared

ffi/firewood_test.go
```diff
@@ -31,7 +31,26 @@ const (
...
+
+	// Format string constants
+	getFmt            = "Get(%d)"
+	commitFmt         = "Commit(%d)"
+	proposeFmt        = "Propose(%d)"
+	baseGetFmt        = "base.Get(): %d"
```
alarso16, line 49: I think logs and error reports like this, it is super typical just use strings. Forcing these into globals feels weird, but I don't know why

ffi/firewood_test.go
```diff
@@ -31,7 +31,26 @@ const (
...
+	nonExistentKey    = "non-existent"
```
alarso16, line 50: When does this key not exist? I think the answer is that all the key generation functions we use will not generate this key, but there's nothing specific about this key that can't be added to the database

ffi/firewood_test.go
rkuris: Curious about the reasoning here. Seems like having this split into multiple lines makes it easier to read?

ffi/firewood_test.go
rkuris: Should this rule be "even if this means duplicating the constant, make a new, separate function-scoped constant for each test" so there is no sharing of the constant between tests. Is that accurate?

rkuris: Similar to @RodrigoVillar's comment. You want constants but have them redeclared in each function as function-scoped constants, correct?

If so, we need to change the instructions.

ffi/firewood_test.go
rkuris: 🤷 me neither. Personally I don't like const being used in tests at all, I prefer just having magic numbers in tests, but I think you and I are the exception.

ffi/firewood_test.go
rkuris: I think this is a pre-existing issue unrelated to this PR. Is that correct?

AGENTS.md
rkuris: Moved in 1f18514e5e76fef6a6fd17f31fcfba31fd0f6fec

ffi/.golangci.yaml
rkuris: Let's just leave this. If it turns into a huge burden we can always revert it later. I doubt this code will get many updates going forward, it's mostly just shim code.

ffi/firewood_test.go
RodrigoVillar: I may be missing context, but it seems that this is just raw block contents? If that's the case, I'm not sure what the legibility improvement is here (i.e. who is reading the literal raw contents?)

ffi/firewood_test.go
RodrigoVillar: IMO yes, and reiterating the point I made:


> I would prefer that we remove test-only package-level variables that cannot be treated as a black box (e.g. I don't know what testNumKeys does/signifies unless I look at its definition, while for emptyEthhashRoot, I can treat it as a black box [and so I'm fine with having it be a test-only package-level variable]).

rkuris:
Based on some long discussions with @JuanLeon2 and others, I'm closing this PR for a few reasons:

1. Focus right now should be on feature-completeness, not on long-term velocity improvements
2. There is some disagreement on exactly which lints should be enabled or not
3. This PR is too large, not in the sense of lines of code, but in the sense of enabling too many lints in one PR. If we want to continue with this approach, we should do this one lint at a time
4. This PR is experiencing a considerable amount of churn, probably due to the prior two reasons
5. There is currently no well defined process for PR review. Although one is documented, it isn't the guideline. This should get resolved before we proceed with additional lint rules
6. While my preference is fewer rounds of ReadyForReview->Reject->ReadyForReview, this isn't a goal for the organization, in that the organization's preference is for PRs that don't just make code better but are as good as they can be for a given change set, and back-and-forth is the way to make that happen

I'm going to keep the branch around in case we want to cherry pick it later, as well as the parent issue for a possible future discussion.

PR #1833, opened 2026-03-24
feat(range-proof-verification): 4/6 complete range proof verification

rkuris:
## Why this should be merged

Implements standalone range proof verification by reconstructing a partial trie from the proven key-value pairs and comparing its root hash against the expected one.

## How this works

The verifier builds an in-memory trie (via MemStore + new_empty_proposal), inserts the proven keys, then reconciles each proof node into the trie. Three new helper functions compute which child branches fall outside the proven range and must use the proof's original hashes:

- compute_outside_children: walks the proof edge path to determine which child indices are outside the proven range at each node
- mark_outside: marks children strictly less than (left edge) or greater than (right edge) the on-path nibble
- compute_root_hash_with_proofs: recursively hashes the trie, using proof sibling hashes for children marked as outside

Also un-gates several functions so they are available outside of test builds.

## How this was tested

Some tests were enabled here but more exhaustive testing is in the next PR of the series.

## Breaking Changes

None

### DIFF
--- a/ffi/metrics_test.go
+++ b/ffi/metrics_test.go
@@ -6,7 +6,6 @@ package ffi
 import (
 	"fmt"
 	"io"
-	"maps"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -22,12 +21,8 @@ import (
 var (
 	metricsPort     = uint16(3000)
 	expectedMetrics = map[string]dto.MetricType{
-		"ffi_batch":          dto.MetricType_COUNTER,
 		"proposal_commit":    dto.MetricType_COUNTER,
 		"proposal_commit_ms": dto.MetricType_COUNTER,
-		"ffi_propose_ms":     dto.MetricType_COUNTER,
-		"ffi_commit_ms":      dto.MetricType_COUNTER,
-		"ffi_batch_ms":       dto.MetricType_COUNTER,
 		"flush_nodes":        dto.MetricType_COUNTER,
 		"insert":             dto.MetricType_COUNTER,
 		"space_from_end":     dto.MetricType_COUNTER,
@@ -39,11 +34,6 @@ var (
 		"jemalloc_metadata_bytes":  dto.MetricType_GAUGE,
 		"jemalloc_mapped_bytes":    dto.MetricType_GAUGE,
 		"jemalloc_resident_bytes":  dto.MetricType_GAUGE,
-	}
-	expectedExpensiveMetrics = map[string]dto.MetricType{
-		"ffi_commit_ms_bucket":  dto.MetricType_HISTOGRAM,
-		"ffi_propose_ms_bucket": dto.MetricType_HISTOGRAM,
-		"ffi_batch_ms_bucket":   dto.MetricType_HISTOGRAM,
 	}
 	initMetrics   sync.Once
 	initLogs      sync.Once
@@ -104,24 +94,6 @@ func TestMetrics(t *testing.T) {
 	}
 }
 
-func TestExpensiveMetrics(t *testing.T) {
-	r := require.New(t)
-	db, _ := newDbWithMetricsAndLogs(t, WithExpensiveMetrics())
-	// batch update
-	_, _, batch := kvForTest(10)
-	_, err := db.Update(batch)
-	r.NoError(err)
-
-	// Close database to ensure background persistence completes before checking metrics.
-	// The flush_nodes metric is recorded during persistence, which happens asynchronously.
-	r.NoError(db.Close(t.Context()))
-
-	merged := make(map[string]dto.MetricType, len(expectedMetrics)+len(expectedExpensiveMetrics))
-	maps.Copy(merged, expectedMetrics)
-	maps.Copy(merged, expectedExpensiveMetrics)
-	assertMetrics(t, metricsPort, merged)
-}
-
 func assertNonEmptyFile(t *testing.T, path string) bool {
 	t.Helper()
 	f, err := os.ReadFile(path)
@@ -146,12 +118,9 @@ func assertMetrics(t *testing.T, metricsPort uint16, expected map[string]dto.Met
 	resp, err := client.Do(req)
 	r.NoError(err)
 
-	body, err := io.ReadAll(resp.Body)
+	_, err = io.ReadAll(resp.Body)
 	r.NoError(err)
 	r.NoError(resp.Body.Close())
-
-	// Check that batch op was recorded (no prefix)
-	r.NotContains(string(body), "ffi_batch 0")
 
 	g := Gatherer{}
 	metricsFamily, err := g.Gather()
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -17,9 +17,7 @@ use firewood_storage::{Committed, FileBacked, NodeStore};
 use crate::{BatchOp, BorrowedBytes, CView, CreateProposalResult, arc_cache::ArcCache};
 
 use crate::revision::{GetRevisionResult, RevisionHandle};
-use firewood_metrics::{
-    MetricsContext, firewood_increment, firewood_record, fwd_expensive_timed_result,
-};
+use firewood_metrics::{MetricsContext, firewood_increment};
 
 /// The hashing mode to use for the database.
 ///
@@ -227,22 +225,8 @@ impl DatabaseHandle {
         &self,
         values: impl AsRef<[BatchOp<'a>]> + 'a,
     ) -> Result<Option<HashKey>, api::Error> {
-        let (root_hash_result, elapsed) =
-            fwd_expensive_timed_result!(crate::registry::BATCH_MS_BUCKET, {
-                let CreateProposalResult { handle } =
-                    self.create_proposal_handle(values.as_ref())?;
-                handle.commit_proposal()
-            });
-        let root_hash = root_hash_result?;
-        firewood_increment!(crate::registry::BATCH_MS, elapsed.as_millis() as u64);
-        firewood_increment!(crate::registry::BATCH_COUNT, 1);
-        firewood_record!(
-            crate::registry::BATCH_MS_BUCKET,
-            elapsed.as_secs_f64() * 1000.0,
-            expensive
-        );
-
-        Ok(root_hash)
+        let CreateProposalResult { handle } = self.create_proposal_handle(values.as_ref())?;
+        handle.commit_proposal()
     }
 
     /// Returns an owned handle to the revision corresponding to the provided root hash.
--- a/ffi/src/metrics.rs
+++ b/ffi/src/metrics.rs
@@ -6,8 +6,8 @@ use std::net::Ipv6Addr;
 use std::sync::OnceLock;
 
 use crate::jemalloc_metrics;
-use firewood_metrics::{HistogramBucketConfig, MetricsContext};
-use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
+use firewood_metrics::MetricsContext;
+use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
 use oxhttp::Server;
 use oxhttp::model::{Body, Response, StatusCode};
 use std::net::Ipv4Addr;
@@ -47,28 +47,14 @@ impl<T: MetricsContextExt + ?Sized> MetricsContextExt for &mut T {
 /// This happens on a per-process basis, meaning that the metrics system cannot
 /// be initialized if it has already been set up in the same process.
 pub fn setup_metrics() -> Result<(), Box<dyn Error>> {
-    // Collect histogram bucket configurations from all crates
-    let mut histogram_configs: Vec<HistogramBucketConfig> = Vec::new();
-    crate::registry::register(&mut histogram_configs);
+    crate::registry::register();
     firewood::registry::register();
     firewood_storage::registry::register();
     #[cfg(feature = "block-replay")]
     firewood_replay::registry::register();
     jemalloc_metrics::register();
 
-    // Build the Prometheus exporter with bucket configurations from the registry
-    // TODO: Switch to Prometheus's native histograms
-    // they are cheaper, more efficient, and easier to configure (no predefined buckets)
-    // proper default support will start in prometheus v3.9 and v4.0; once our infra switches,
-    // we should switch too.
-    let mut builder = PrometheusBuilder::new();
-
-    // Apply bucket configurations from the global registry
-    for config in histogram_configs {
-        builder = builder
-            .set_buckets_for_metric(Matcher::Full(config.name.to_string()), config.buckets)?;
-    }
-
+    let builder = PrometheusBuilder::new();
     let handle = builder.install_recorder()?;
 
     RECORDER
--- a/ffi/src/proposal.rs
+++ b/ffi/src/proposal.rs
@@ -4,7 +4,6 @@
 use firewood::api::{self, BoxKeyValueIter, DbView, HashKey, IntoBatchIter, Proposal as _};
 
 use crate::{IteratorHandle, iterator::CreateIteratorResult, metrics::MetricsContextExt};
-use firewood_metrics::{firewood_increment, firewood_record, fwd_expensive_timed_result};
 
 /// An opaque wrapper around a Proposal that also retains a reference to the
 /// database handle it was created from.
@@ -80,15 +79,10 @@ impl ProposalHandle<'_> {
             _ = handle.get_root(hash_key.clone());
         }
 
-        let (commit_result, commit_time) =
-            fwd_expensive_timed_result!(crate::registry::COMMIT_MS_BUCKET, proposal.commit());
-        commit_result?;
+        proposal.commit()?;
 
         // clear the cached view so that it does not hold onto the proposal view
         handle.clear_cached_view();
-
-        firewood_increment!(crate::registry::COMMIT_MS, commit_time.as_millis() as u64);
-        firewood_increment!(crate::registry::COMMIT_COUNT, 1);
 
         Ok(hash_key)
     }
@@ -118,16 +112,7 @@ impl<'db> CreateProposalResult<'db> {
         handle: &'db crate::DatabaseHandle,
         f: impl FnOnce() -> Result<firewood::db::Proposal<'db>, api::Error>,
     ) -> Result<Self, api::Error> {
-        let (proposal_result, propose_time) =
-            fwd_expensive_timed_result!(crate::registry::PROPOSE_MS_BUCKET, f());
-        let proposal = proposal_result?;
-        firewood_increment!(crate::registry::PROPOSE_MS, propose_time.as_millis() as u64);
-        firewood_increment!(crate::registry::PROPOSE_COUNT, 1);
-        firewood_record!(
-            crate::registry::PROPOSE_MS_BUCKET,
-            propose_time.as_secs_f64() * 1000.0,
-            expensive
-        );
+        let proposal = f()?;
 
         let hash_key = proposal.root_hash();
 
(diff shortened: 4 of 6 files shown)


demosdemon:
> **Note:** This review was generated by Claude Code (Opus 4.6).

# Range Proof Verification Review

## Overview

This PR implements standalone range proof verification by building an in-memory trie from proven key-value pairs, reconciling proof nodes, computing which child branches fall outside the proven range, and recursively hashing while substituting proof hashes for outside children. The core algorithm is sound — I traced through several edge cases (shared prefix paths, terminal node handling, boundary divergence) and the approach is correct.

The architectural choice of computing hashes externally (via `compute_root_hash_with_proofs`) rather than mutating the trie (like Go's `addPathInfo`) is a good fit for Rust — it avoids storing `Child::Hash` variants in the in-memory trie and keeps the data model clean.

---

## Critical: All 14 ignored tests now pass — un-ignore them

I ran all 14 tests marked with `#[ignore = "https://github.com/ava-labs/firewood/issues/738"]` and **every single one passes**:

`test_bad_range_proof_malformed`, `test_range_proof_with_invalid_non_existent_proof`, `test_one_element_range_proof`, `test_all_elements_proof`, `test_empty_range_proof`, `test_gapped_range_proof`, `test_same_side_proof`, `test_single_side_range_proof`, `test_reverse_single_side_range_proof`, `test_both_sides_range_proof`, `test_empty_value_range_proof`, `test_all_elements_empty_value_range_proof`, `test_range_proof_keys_with_shared_prefix`, `test_bloadted_range_proof`

These should be un-ignored in this PR (or the next in the series with explicit justification). The whole point of the implementation is to make these pass, and leaving them ignored means CI doesn't validate the new code.

---

## High Severity

### 1. Missing validation: keys within `[start, end]` range

The Go reference (`verifySortedKeyChanges`) checks that all key-value pairs fall within `[start, end]`:
```go
if (hasLowerBound && bytes.Compare(keyChanges[i].Key, start.Value().Bytes()) < 0) ||
   (hasUpperBound && bytes.Compare(keyChanges[i].Key, end.Value().Bytes()) > 0) {
    return ErrStateFromOutsideOfRange
}
```

The Rust code (lines 472-482) only checks monotonicity but **never validates that keys are within the requested range**. A malicious peer could send key-value pairs outside `[first_key, last_key]` and the verifier would accept them if the root hash happens to match.

### 2. Missing validation: end proof required when key changes exist

The Go code requires an end proof whenever there are key changes:
```go
case len(endProof) == 0 && (endKey.HasValue() || len(keyChanges) > 0):
    return ErrNoEndProof
```

The Rust code skips edge verification for empty proofs (lines 497, 506) without checking whether the proof *should* be non-empty. This means a verifier could accept a range proof with key-value pairs but no end proof, potentially missing completeness guarantees.

### 3. Missing validation: start proof should be empty when start key is None

The Go code rejects `startKey.IsNothing() && len(startProof) > 0` with `ErrUnexpectedStartProof`. The Rust code doesn't have this check. While more of a structural validation issue than a soundness one, it could allow non-canonical proofs.

---

## Medium Severity

### 4. Proof node map collision is silently lossy

At lines 534-538, when start and end proofs share a node at the same key path, the end proof's `ProofNode` silently overwrites the start proof's in `proof_node_map`. For a valid proof both should be identical, but a malicious proof could exploit this — the end proof's child hashes would be used even if the start proof had different (correct) hashes.

Consider either:
- Asserting the proof nodes are equal when keys collide
- Using `.entry().or_insert()` instead of `.insert()`
- Explicitly documenting the overwrite behavior

### 5. `verify_range_proof` doesn't use `self`

The method creates its own `MemStore`-backed trie internally and never reads from `self`. This makes the API slightly misleading — callers must create a `Merkle<T>` instance they don't need. Pre-existing issue (the old TODO-stub also didn't use `self`), fine to defer, but worth tracking since it's now a complete implementation.

---

## Low Severity

### 6. PR title typo

"verificatoin" → "verification"

### 7. Docstring is stale

The `verify_range_proof` doc (lines 436-448) still describes only "structural validation" and "boundary proof verification", but the method now does full trie reconstruction + root hash comparison. Should be updated to reflect the actual verification process.

### 8. Pre-existing test bug in `test_range_proof` (range.rs line 39)

```rust
let end = rng.random_range(0..items.len() - start) + start - 1;
```

When `start = 0` and the random value is 0, this underflows (`0 + 0 - 1` on `usize`), producing `usize::MAX`. The `end <= start` guard doesn't catch it since `MAX > 0`, leading to a panic on `items[start..end]`. Not introduced by this PR, but worth fixing while you're here.

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+            parent.key.clone(),
+            on_path_nibble.as_u8(),
+            is_left_edge,
+        );
+    }
+
+    // Terminal node: derive the on-path nibble from the boundary key.
+    // If the boundary key diverges within the terminal node's partial path,
+    // either all or none of the children are outside the range.
+    if let (Some(terminal), Some(boundary)) = (proof_nodes.last(), boundary_key) {
```
alarso16, line 215: Is it rusty to early return?

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+        let divergence = terminal
+            .key
+            .iter()
+            .zip(boundary_nibbles.iter())
+            .position(|(tk, &bn)| tk.as_u8() != bn);
+
+        match divergence {
+            None if boundary_nibbles.len() > terminal.key.len() => {
+                // Boundary extends beyond terminal key and matches it fully.
+                // Use the next nibble from the boundary as the on-path nibble.
```
alarso16: Why?

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+                mark_outside(
+                    &mut result,
+                    terminal.key.clone(),
+                    on_path_nibble,
+                    is_left_edge,
+                );
+                // The on-path child itself also needs the proof hash because
+                // the proving trie only contains keys within the range, but
+                // the original subtree under this child may contain keys
+                // outside the range that contribute to its hash.
```
alarso16: I do not understand this comment. what's the "original subtree"?

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+                let entry = result.entry(terminal.key.clone()).or_insert([false; 16]);
```
alarso16: It was just inserted - why specify the default value? You know it's there

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+                // If boundary is "past" the terminal (left edge: B > terminal,
+                // right edge: B < terminal), all children are outside.
+                let all_outside = if is_left_edge {
+                    boundary_nibbles[d] > terminal.key[d].as_u8()
+                } else {
+                    boundary_nibbles[d] < terminal.key[d].as_u8()
+                };
+                if all_outside {
+                    let entry = result.entry(terminal.key.clone()).or_insert([false; 16]);
+                    *entry = [true; 16];
```
alarso16: Why not just insert `[true;16]`, overriding whatever is there (but it should be nothing)?

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+    proof_nodes: &[ProofNode],
+    boundary_key: Option<&[u8]>,
+    is_left_edge: bool,
+) -> HashMap<PathBuf, [bool; 16]> {
+    let mut result: HashMap<PathBuf, [bool; 16]> = HashMap::new();
+
+    // Non-terminal nodes: derive the on-path nibble from the next proof node
+    for window in proof_nodes.windows(2) {
+        let (parent, child) = (&window[0], &window[1]);
+        let Some(on_path_nibble) = child.key.get(parent.key.len()) else {
```
alarso16: When is this not true?

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+///
+/// The `boundary_key` (in bytes, not nibbles) is used to determine the on-path
+/// nibble at the terminal proof node, since there is no subsequent proof node
+/// to derive it from.
+#[expect(clippy::indexing_slicing)]
+fn compute_outside_children(
+    proof_nodes: &[ProofNode],
+    boundary_key: Option<&[u8]>,
+    is_left_edge: bool,
+) -> HashMap<PathBuf, [bool; 16]> {
```
alarso16: This `[bool; 16]` type is used a lot - should it be aliased? Not sure what's common in Rust. Looking at the type, it's not really clear what it means

firewood/src/merkle/mod.rs
```diff
@@ -313,23 +491,75 @@ impl<T: TrieReader> Merkle<T> {
...
+            .chain(proof.end_proof().as_ref())
+            .collect();
+
+        let mut proof_node_map: HashMap<PathBuf, &ProofNode> = HashMap::new();
+        for proof_node in &all_proof_nodes {
+            proving_merkle.reconcile_branch_proof_node(proof_node)?;
+            proof_node_map.insert(proof_node.key.clone(), proof_node);
+        }
+
+        // Compute which children at each edge node are outside the proven range
```
alarso16: this comment is clear from the first function calls, but I still don't know what's going on in the for loop. Your clarifying comment didn't clarify much for me

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+    let full_key: PathBuf = path_prefix
+        .iter()
+        .chain(branch.partial_path.as_components().iter())
+        .copied()
+        .collect();
+
+    let mut child_hashes: Children<Option<HashType>> = Children::new();
+
+    // For children outside the proven range, use proof node hashes
+    if let (Some(proof_node), Some(outside)) =
```
alarso16, line 324: This section is super clear, love it!

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+        (proof_nodes.get(&full_key), outside_children.get(&full_key))
+    {
+        for (nibble, hash) in proof_node.child_hashes.iter_present() {
+            if outside[nibble.as_u8() as usize] {
+                child_hashes[nibble] = Some(hash.clone());
+            }
+        }
+    }
+
+    // For children in the in-memory trie, compute hashes recursively
```
alarso16: I think you should add a "why" this doesn't overlap with the range above

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +172,182 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+    for (nibble, child_opt) in &branch.children {
+        if let Some(Child::Node(child_node)) = child_opt {
+            let mut child_prefix: PathBuf = full_key.iter().copied().collect();
```
alarso16: Is this call expensive? Seems like it would be better to make it once

rkuris:
Great idea to have claude check against the existing go code! Comments addressed below.

> > **Note:** This review was generated by Claude Code (Opus 4.6).
> 
> # Range Proof Verification Review
> ## Overview
> This PR implements standalone range proof verification by building an in-memory trie from proven key-value pairs, reconciling proof nodes, computing which child branches fall outside the proven range, and recursively hashing while substituting proof hashes for outside children. The core algorithm is sound — I traced through several edge cases (shared prefix paths, terminal node handling, boundary divergence) and the approach is correct.

Aw shucks AI. You're my friend too most of the time. ⭐ 

> The architectural choice of computing hashes externally (via `compute_root_hash_with_proofs`) rather than mutating the trie (like Go's `addPathInfo`) is a good fit for Rust — it avoids storing `Child::Hash` variants in the in-memory trie and keeps the data model clean.

🦀 

> ## Critical: All 14 ignored tests now pass — un-ignore them

As mentioned in the PR description, these tests are enabled and more exhaustive tests are added in the next PR of the series.

> ## High Severity
> ### 1. Missing validation: keys within `[start, end]` range
> The Go reference (`verifySortedKeyChanges`) checks that all key-value pairs fall within `[start, end]`:
> 
> ```go
> if (hasLowerBound && bytes.Compare(keyChanges[i].Key, start.Value().Bytes()) < 0) ||
>    (hasUpperBound && bytes.Compare(keyChanges[i].Key, end.Value().Bytes()) > 0) {
>     return ErrStateFromOutsideOfRange
> }
> ```

Fixed in 57b47e1c49cf3c46a4edcd50d98146ac169dbd76

> ### 2. Missing validation: end proof required when key changes exist

Fixed in 57b47e1c49cf3c46a4edcd50d98146ac169dbd76

> ### 3. Missing validation: start proof should be empty when start key is None

Fixed in 57b47e1c49cf3c46a4edcd50d98146ac169dbd76

(remaining issues addressed in next comment)

rkuris:
(additional comments and fixes below)

> ## Medium Severity
> ### 4. Proof node map collision is silently lossy
> At lines 534-538, when start and end proofs share a node at the same key path, the end proof's `ProofNode` silently overwrites the start proof's in `proof_node_map`. For a valid proof both should be identical, but a malicious proof could exploit this — the end proof's child hashes would be used even if the start proof had different (correct) hashes.
> 
> Consider either:
> 
> * Asserting the proof nodes are equal when keys collide
> * Using `.entry().or_insert()` instead of `.insert()`
> * Explicitly documenting the overwrite behavior

Fixed in bec9541ed5e587d6c15b219fe33b8f5bb78e9dcf. I would rather fail the proof early if this happens.

> ### 5. `verify_range_proof` doesn't use `self`
> The method creates its own `MemStore`-backed trie internally and never reads from `self`. This makes the API slightly misleading — callers must create a `Merkle<T>` instance they don't need. Pre-existing issue (the old TODO-stub also didn't use `self`), fine to defer, but worth tracking since it's now a complete implementation.

Yes, this created a pretty big diff but it's in bec9541ed5e587d6c15b219fe33b8f5bb78e9dcf as well.

> ## Low Severity
> ### 6. PR title typo
> "verificatoin" → "verification"

Fixed.

> ### 7. Docstring is stale
> The `verify_range_proof` doc (lines 436-448) still describes only "structural validation" and "boundary proof verification", but the method now does full trie reconstruction + root hash comparison. Should be updated to reflect the actual verification process.

Fixed in bec9541ed5e587d6c15b219fe33b8f5bb78e9dcf

> ### 8. Pre-existing test bug in `test_range_proof` (range.rs line 39)
> ```rust
> let end = rng.random_range(0..items.len() - start) + start - 1;
> ```
> 
> When `start = 0` and the random value is 0, this underflows (`0 + 0 - 1` on `usize`), producing `usize::MAX`. The `end <= start` guard doesn't catch it since `MAX > 0`, leading to a panic on `items[start..end]`. Not introduced by this PR, but worth fixing while you're here.

FIxed in bec9541ed5e587d6c15b219fe33b8f5bb78e9dcf

firewood/src/merkle/mod.rs
rkuris: I think this is okay and is idiomatic. YMMV.

firewood/src/merkle/mod.rs
rkuris: Explained better in 8e7cae63eefe71dface58633c7759a995abbe3d1. The comment was wrong which explains why you were confused.

The state here is that the boundary key is longer which guarantees the terminal node is an ancestor, so we need to get that nibble to find the right branch to proceed.

firewood/src/merkle/mod.rs
rkuris: Tricky code for sure. I made another attempt to explain this in 6b366077d9343331bdd01b4f56307b7c922fac9c.

Let me know if this is better.

firewood/src/merkle/mod.rs
rkuris: Fair point, there is a another way to do this but it does add an expect call.

Changed in bc73231ca776bb3bae54eaf2e573aadd0313c166. Both are kinda hokey.

firewood/src/merkle/mod.rs
rkuris: Fair. Fixed and easier to read in 620989018f6471932f7ef8fef3c330258c2448fc

firewood/src/merkle/mod.rs
rkuris: Fixed in e219ba73dce9674fcd7f068460791d16b215d3fe

firewood/src/merkle/mod.rs
rkuris: Trying again in c4557f4711d6f69f45de097992a25c8b43ce5c73

The reconciler was done under a different PR in this branch, but a casual reader of this function might not understand what the idea is here, so a longer comment is in fact warranted.

firewood/src/merkle/mod.rs
rkuris: Second attempt in f6bc8d2bcdae9d03d13e8bcbd2bdc44daa55e5c3

firewood/src/merkle/mod.rs
rkuris: yeah but this was easier to write :)

Fixed in c096a32e5b369ab43064100a630e13e98272e3a4

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +177,345 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+                // right edge: B < terminal), all children are outside.
+                let all_outside = if is_left_edge {
+                    boundary_nibbles[d] > terminal.key[d].as_u8()
+                } else {
+                    boundary_nibbles[d] < terminal.key[d].as_u8()
+                };
+                if all_outside {
+                    result.insert(terminal.key.clone(), [true; 16]);
+                }
+                // Otherwise (boundary is "before" terminal), no children are outside.
```
alarso16, line 260: This case is weird. So `d` is the key prefix that the boundary and the terminal proof node share. In this case, the first (or last) key of the proof is outside the proof, correct? How can we verify that key?

I'm pretty sure that this is verified earlier, but I just want to make sure I'm not misunderstanding something

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +177,345 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+fn compute_outside_children(
+    proof_nodes: &[ProofNode],
+    boundary_key: Option<&[u8]>,
+    is_left_edge: bool,
+) -> Result<HashMap<PathBuf, ChildMask>, ProofError> {
+    let mut result: HashMap<PathBuf, ChildMask> = HashMap::new();
+
+    // Non-terminal nodes: derive the on-path nibble from the next proof node
+    for window in proof_nodes.windows(2) {
+        let (parent, child) = (&window[0], &window[1]);
```
demosdemon, line 199: fun nit:

```suggestion
    for [parent, child] in proof_nodes.array_windows() {
```

firewood/src/merkle/mod.rs
```diff
@@ -173,6 +177,345 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+}
+
+/// Marks children at the given node as "outside" the proven range.
+#[expect(clippy::indexing_slicing)]
+fn mark_outside(
+    map: &mut HashMap<PathBuf, ChildMask>,
+    key: PathBuf,
+    on_path_nibble: u8,
+    is_left_edge: bool,
+) {
```
demosdemon, line 280: nit: this could return `&mut ChildMask` to avoid the unwrap above.

firewood/src/merkle/mod.rs
demosdemon: This is checking that the key in `boundary_nibbles`, if it were in the trie, would exist at some branch with a path `boundary_nibbles[..d]` on the inside edge of the proof relative to the branch leading toward `terminal`.

firewood/src/merkle/mod.rs
rkuris: I'm gonna pull out all this code in a different PR.

Issue #1655, opened 2026-02-03
perf: Add jemalloc metrics to firewood metrics set


rkuris:
**Is your feature request related to a problem? Please describe.**
We need to understand the memory requirements for firewood's components. This might also be useful to add some backpressure when we're using too much memory later.

**Describe the solution you'd like**
Use an existing rust library to collect all the metrics and run them down the pipe

**Describe alternatives you've considered**
We could inspect the metrics ourselves but I think there are crates to help us with the heavy lifting.

PR #1837, opened 2026-03-25
feat(metrics): Add jemalloc metrics

rkuris:
## Why this should be merged

We need jemalloc stats for monitoring

## How this works

Add the appropriate crates and call the methods to fetch jemalloc stats on each call to gather().

This works fine for ffi based applications, but not for general rust applications, but they might have different metric semantics anyway.

## How this was tested

Added a test to verify these metrics are emitted in the metrics_test.go code

## Breaking Changes

None. Adds new metrics.

### DIFF
--- a/ffi/src/jemalloc_metrics.rs
+++ b/ffi/src/jemalloc_metrics.rs
@@ -8,6 +8,7 @@
 //! exposes: active, allocated, metadata, mapped, resident, retained.
 
 use firewood_metrics::firewood_set;
+use firewood_storage::logger;
 use metrics::describe_gauge;
 use tikv_jemalloc_ctl::{epoch, stats};
 
@@ -64,13 +65,17 @@ pub fn register() {
 /// individual stats are silently ignored (the gauge simply won't update).
 pub fn refresh() {
     // Advance the epoch so jemalloc refreshes its cached stats.
-    if epoch::advance().is_err() {
+    if let Err(e) = epoch::advance() {
+        logger::warn!("jemalloc epoch advance failed: {e}");
         return;
     }
 
     for stat in JEMALLOC_STATS {
-        if let Ok(v) = (stat.read)() {
-            firewood_set!(stat.name, v);
+        match (stat.read)() {
+            Ok(v) => firewood_set!(stat.name, v),
+            Err(e) => {
+                logger::warn!("failed to read jemalloc stat {}: {e}", stat.name);
+            }
         }
     }
 }


rkuris:
Resolved #1655

ffi/src/jemalloc_metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/// Advances the jemalloc epoch and updates all gauge metrics with current values.
+///
+/// Call this before gathering metrics to ensure fresh data. Errors reading
+/// individual stats are silently ignored (the gauge simply won't update).
+pub fn refresh() {
+    // Advance the epoch so jemalloc refreshes its cached stats.
+    if epoch::advance().is_err() {
+        return;
```
Elvis339, line 70: Is this silent failure expected?

ffi/src/jemalloc_metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+    },
+    JemallocStat {
+        name: "jemalloc.retained_bytes",
+        description: "Bytes in virtual memory mappings retained by jemalloc",
+        read: stats::retained::read,
+    },
+];
+
+/// Registers all jemalloc metric descriptions.
+pub fn register() {
```
Elvis339, line 56: This can be called multiple times nothing prevents it from being called twice, which would double-register gauge descriptions.

ffi/Cargo.toml
```diff
@@ -29,6 +29,7 @@ parking_lot.workspace = true
 # Regular dependencies
 oxhttp = "0.3.1"
 tikv-jemallocator = "0.6.1"
+tikv-jemalloc-ctl = { version = "0.6.0", features = ["stats"] }
```
Elvis339: I'm unsure of the effect of this incosistency it could be minor, bringing it up for attention: tikv-jemallocator is 0.6.1 and tikv-jemalloc-ctl is 0.6.0

ffi/src/metrics.rs
```diff
@@ -106,9 +108,12 @@ pub fn setup_metrics_with_exporter(metrics_port: u16) -> Result<(), Box<dyn Erro
 }
 
 /// Returns the latest metrics for this process.
+///
+/// Refreshes jemalloc stats before rendering so the returned snapshot is current.
 pub fn gather_metrics() -> Result<String, String> {
     let Some(recorder) = RECORDER.get() else {
         return Err(String::from("recorder not initialized"));
     };
+    jemalloc_metrics::refresh();
```
Elvis339, line 118: This will only be exposed via `Gatherer` and not in server. I think you want this below line 87 as well.

ffi/src/jemalloc_metrics.rs
```diff
@@ -0,0 +1,76 @@
...
+use tikv_jemalloc_ctl::{epoch, stats};
+
+/// A jemalloc stat to expose as a gauge metric.
+struct JemallocStat {
+    name: &'static str,
+    description: &'static str,
+    read: fn() -> tikv_jemalloc_ctl::Result<usize>,
+}
+
+static JEMALLOC_STATS: &[JemallocStat] = &[
```
Elvis339, line 22: These names will conflict with any other registry that exposes the same jemalloc stats. This works today only because AvalancheGo's Gatherer interface adds a `firewood` prefix when scraping (see [vm.go L514–L517](https://github.com/ava-labs/avalanchego/blob/96d55fe91ab284b941485cb77e6de2cc02cd7fc9/graft/coreth/plugin/evm/vm.go#L514-L517)), so collision is avoided at the integration layer rather than at the source. This would surface if this was implemented https://github.com/ava-labs/firewood/pull/1837/changes#r2988647924

Note that adding a `firewood.` prefix here is not the fix you'd end up with `firewood_firewood_jemalloc_active_bytes` since AvalancheGo is already applying that prefix.

ffi/src/jemalloc_metrics.rs
rkuris: I don't think we have a good way of surfacing this error to anyone. It would be possible to create a metrics named "jemalloc.last_error" or something. Any suggestions?

ffi/src/jemalloc_metrics.rs
rkuris: This is safe, per the documentation: https://crates.io/crates/metrics-prometheus/0.9.0#:~:text=//%20Description%20can%20be%20changed%20multiple%20times%20and%20anytime.%0Ametrics%3A%3Adescribe_counter!(%22count%22%2C%20%22Another%20description.%22)%3B

ffi/src/jemalloc_metrics.rs
rkuris: I think this problem is true for any of our other metrics. We assume we own the namespace for all metrics we generate, and expect callers to prefix us with something. We document that we don't add the firewood prefix on anything in METRICS.md (at least for now)

rkuris:
> epoch::advance() runs on every gather_metrics() call

This is a good point. If you think it's very expensive we could add it to the expensive metrics set (which we do enable by default). I think we only call gather less than once per second from avalanchego, so perhaps this isn't a big problem. Suggestions welcome.

> Is this meant to run in production permanently,

I think we do to help diagnose memory leaks or growth especially coming from firewood.

ffi/Cargo.toml
rkuris: Fixed in 9d8b8d78f40710431874cb5d6ec35b495fcd4d16

ffi/src/metrics.rs
rkuris: Great catch. Fixed in 9d8b8d78f40710431874cb5d6ec35b495fcd4d16

ffi/src/jemalloc_metrics.rs
Elvis339: I think a simple log would be enough. Putting myself into debug mode and trying to figure out why this metric isn't being ingested would drive me crazy.

rkuris: Fixed in 275dc523b0b7b5d74e5ba9341677593d566717c3

ffi/src/jemalloc_metrics.rs
Elvis339: That is only true for [Gatherer](https://github.com/ava-labs/firewood/blob/b11fc0e575c87c6e89445724816d175ba3664d5a/ffi/src/metrics.rs#L109) but this breaks the moment someone start's using metrics server https://github.com/ava-labs/firewood/blob/b11fc0e575c87c6e89445724816d175ba3664d5a/ffi/src/metrics.rs#L82 

I think we should add this under experimental or expensive flag (as you noted below), I cannot think of a scenario where we'd be collecting jemalloc stats from the fleet of nodes in production, it would mostly be noise.

I think this is good segway to start separating experimental and production metrics. 
Not going to block on this but, it would be good to open a follow-up issue that would add this behind expensive metrics or maybe even introduce new flag.

ffi/src/jemalloc_metrics.rs
Elvis339: I see, it would panic. 👍

ffi/src/jemalloc_metrics.rs
rkuris: Created https://github.com/ava-labs/firewood/issues/1840 for followup

PR #1842, opened 2026-03-26
feat(cp-verification): add change proof verification types and verify_and_propose API

bernard-avalabs:
This is the first in a series of PRs for implementing change proof verification. 
The main new functions (verify_and_propose and verify_and_commit) are 
added in this PR but not used until a later PR.

## Reason for changing API

Previously the verification flow consists of **3 steps** after receiving a 
`ChangeProof`:

1) Calling `Verify` which consumes the FrozenChangeProof within the 
`ChangProofContext` to generate a `VerifiedChangeProof`

2) Calling `Propose` to get a `ProposedChangeProof`

3) Calling `Commit` to commit the proposal. 

This was an appropriate flow when only structural verifications were 
performed, such as checking `start_key` and `end_key` and checking if 
the keys in the `BatchOps` are sorted. However, for complete verification,  
it is necessary to create a proposal first before before performing 
additional verification.

This PR together with two followup PRs will replace the **3-step** flow
with a **2-step** flow consisting of the following after receiving a
`ChangeProof`:

1) Calling `VerifyAndPropose` which first performs the same structual
verification as `Verify`, creates a proposal, and then performs the 
remaining verification (consisting of verifying the child hashes of the 
proof nodes that are between the start and end keys matches the hashes
of the proposal).

2) Calling `Commit` to commit the proposal.

This PR just introduces `verify_and_propose` and `verify_and_commit` 
(which combines proposal and commit). The next PR will add the FFI support
for these functions while keeping the **3-step** flow to allow tests to pass.
And the PR after that will remove the **3-step** flow and modify everything
to use the **2-step** flow.

## Why this should be merged

  Lays groundwork for full change proof verification. Adds new ProofError
  variants for boundary proof validation, root hash mismatch, and structural
  checks. Adds Proposal::path_to_key for sub-trie hash verification.
  Introduces verify_and_propose/verify_and_commit methods on
  ChangeProofContext as internal Rust methods (not yet wired to FFI).

  Additionally, this PR aligns Firewood's change proof generator and structural verifier with AvalancheGo's conventions:                                                                                                                            
   
  - Generator convention: The end proof is now always built for the last batch_ops key when batch_ops is non-empty, matching AvalancheGo's GetChangeProof. Previously, non-truncated proofs used end_key, which created ambiguity — the verifier    
  couldn't determine which key the end proof was built for without a two-key trial. With this change, the verifier can always derive the end proof key deterministically from the proof content (batch_ops.last() or end_key when empty), enabling a
   simple single-key hash chain verification in downstream PRs.                                                                                                                                                                                     
  - MissingEndProof structural check: Rejects empty end_proof when end_key is provided or batch_ops is non-empty, matching AvalancheGo's ErrNoEndProof. The honest generator always produces an end proof in these cases. Without this check, a   
  malicious prover could force the verifier through expensive trie operations (proposal construction, root hash verification) before the proof is ultimately rejected.      

## How this works

  - Adds five new ProofError variants: EndRootMismatch, BoundaryValueMismatch,
    BoundaryProofUnverifiable, UnsupportedDeleteRange, MissingBoundaryProof.
  - Adds Proposal::path_to_key() for walking trie paths during verification.
  - Propagates inner ProofError message through api::Error display.
  - Adds ChangeProofContext::verify_and_propose() which takes ownership of the
    proof, performs structural checks, creates a proposal, and returns the
    proof on failure. verify_and_commit() composes verify_and_propose + commit.
  - Changes the change proof generator to always use last_op_key for the end proof when batch_ops is non-empty (previously used end_key for non-truncated proofs).                                                                                  
  - Adds MissingEndProof check to verify_proof_structure to reject empty end proofs early. 

## How this was tested

  - CI
  - Two new unit tests for path_to_key.

## Breaking Changes

  - None

### DIFF
--- a/ffi/src/proofs/change.rs
+++ b/ffi/src/proofs/change.rs
@@ -223,6 +223,16 @@ impl ChangeProofContext {
             return Err(api::Error::ProofError(ProofError::UnexpectedEndProof));
         }
 
+        // Reject empty end_proof when end_key is provided or batch_ops is
+        // non-empty. The honest generator always produces an end proof in
+        // these cases. Without this check, a malicious prover can force the
+        // verifier through expensive trie operations (proposal construction,
+        // root hash verification) before the proof is ultimately rejected.
+        // Matches AvalancheGo's ErrNoEndProof.
+        if proof.end_proof().is_empty() && (end_key.is_some() || !batch_ops.is_empty()) {
+            return Err(api::Error::ProofError(ProofError::MissingEndProof));
+        }
+
         Ok(())
     }
 
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -586,21 +586,21 @@ impl<T: HashedNodeReader> Merkle<T> {
             .take(limit.map_or(usize::MAX, NonZeroUsize::get))
             .collect::<Result<Box<_>, FileIoError>>()?;
 
-        let end_proof = if let Some(limit) = limit
-            && limit.get() <= batch_ops.len()
-            && iter.next().is_some()
-        {
-            // limit was provided, we hit it, and there is at least one more key
-            // end proof is for the last key provided
-            batch_ops.last().map(|largest_key| &**largest_key.key())
-        } else {
-            // limit was not hit or not provided, end proof is for the requested
-            // end key so that we can prove we have all keys up to that key
-            end_key
-        }
-        .map(|end_key| self.prove(end_key))
-        .transpose()?
-        .unwrap_or_default();
+        // The end proof is built for the last batch_op key when batch_ops
+        // is non-empty, or for end_key when batch_ops is empty (the "no
+        // changes in this range" case). When batch_ops is non-empty,
+        // end_key is not used for end proof generation — last_op_key
+        // always takes precedence. This matches AvalancheGo's convention
+        // and ensures the verifier can always determine the end proof's
+        // key deterministically from the proof content — no ambiguity, no
+        // two-key trial needed.
+        let end_proof = batch_ops
+            .last()
+            .map(|largest_key| &**largest_key.key())
+            .or(end_key)
+            .map(|end_key| self.prove(end_key))
+            .transpose()?
+            .unwrap_or_default();
 
         Ok(ChangeProof::new(start_proof, end_proof, batch_ops))
     }
--- a/firewood/src/proofs/types.rs
+++ b/firewood/src/proofs/types.rs
@@ -178,6 +178,12 @@ pub enum ProofError {
     /// Matches `AvalancheGo`'s `ErrUnexpectedEndProof`.
     #[error("unexpected non-empty end proof with no end key and no batch operations")]
     UnexpectedEndProof,
+
+    /// Empty end proof when `end_key` is set or `batch_ops` is non-empty.
+    /// The honest generator always produces an end proof in these cases.
+    /// Matches `AvalancheGo`'s `ErrNoEndProof`.
+    #[error("missing end proof: end_key is set or batch_ops is non-empty")]
+    MissingEndProof,
 }
 
 #[derive(Clone, PartialEq, Eq)]


ffi/src/proofs/change.rs
```diff
@@ -168,6 +168,106 @@ impl ChangeProofContext {
...
         }
     }
+
+    /// Verify the change proof and prepare a proposal against the given database
+    /// without committing it.
+    ///
+    /// On success, consumes `self` and returns a [`ProposedChangeProofContext`].
+    /// On failure, returns `self` back to the caller (along with the error) so
+    /// that the caller retains ownership of the unverified proof.
+    #[allow(dead_code, reason = "wired to FFI in a follow-up PR")]
```
bernard-avalabs, line 245: Uses #[allow(dead_code)] instead of #[expect(dead_code)] because the dead_code lint only fires under certain feature flag combinations (ethhash vs default). #[expect] requires the lint to fire in all configurations and produces an unfulfilled_lint_expectations warning otherwise. These unused functions will be used in a followup PR.

ffi/src/proofs/change.rs
```diff
@@ -168,6 +168,106 @@ impl ChangeProofContext {
...
+        max_length: Option<NonZeroUsize>,
+    ) -> Result<ProposedChangeProofContext<'db>, Box<(Self, api::Error)>> {
+        let Some(proof) = self.proof else {
+            return Err(Box::new((
+                Self { proof: None },
+                api::Error::ProofError(ProofError::ProofIsNone),
+            )));
+        };
+
+        let batch_ops = proof.batch_ops();
```
rkuris: This was difficult to review because it seems to do almost exactly what verify() did before and I can't tell what was changed.

I think the intent of this PR is to remove the old verify->propose->commit flow but that isn't prominent in the description of the PR. I'd have to look at future PRs to confirm your intentions.

The root cause here is that the function boundaries aren't right, which also makes this a bit harder to read.

Something like db7a8e2dbb85cc96bdfd181038ae817c3ac3e4ff would help here.

A better PR description might be:

> First step for better state tracking for change proofs. Previously, change proofs were changed in place. This lays the foundation for moving the proposal through various stages.
> The actual removal of verify() is in a later PR; this just moves the code and retains the checks.

The stages should probably be documented here as well, perhaps with a mermaid diagram. That would make some future reader and the current reviewer more comfortable with this change.

ffi/src/proofs/change.rs
```diff
@@ -168,6 +168,106 @@ impl ChangeProofContext {
...
+        {
+            return Err(Box::new((
+                Self { proof: Some(proof) },
+                api::Error::ProofError(ProofError::EndKeyLessThanLastKey),
+            )));
+        }
+
+        if !batch_ops
+            .iter()
+            .is_sorted_by(|a, b| b.key().cmp(a.key()) == Ordering::Greater)
```
rkuris: Pre-existing problem, but this was difficult to check for correctness and even figure out what this was doing. Only noticed because it was changed from verify() when I made the diff.

This could use a top level comment; I used "Verify the keys are in sorted order" in my refactor in db7a8e2dbb85cc96bdfd181038ae817c3ac3e4ff

ffi/src/proofs/change.rs
rkuris: nit: `#[cfg(test)]` might have been a better choice here, which would still guarantee you'll remove this warning once the implementation is complete, or it won't compile.

ffi/src/proofs/change.rs
bernard-avalabs: > I think the intent of this PR is to remove the old verify->propose->commit flow but that isn't prominent in the description of the PR. I'd have to look at future PRs to confirm your intentions.

This is indeed the intent. This PR replicates some of the functionality of existing functions so that a later PR can more cleanly switch from the old 3-step approach (validate->propose->commit) to a 2-step approach (validate and propose->commit) without breaking any of the existing tests.

This function will actually be rewritten in a later PR completely, but is needed in an intermediate PR. This is part of the reason why it is stripped of comments. The suggestion in your branch is certainly better, but that helper function will also be removed in a later PR.

I'll add a mermaid diagram to document the stages in the PR description.

ffi/src/proofs/change.rs
bernard-avalabs: There are comments in the rewritten version of this function that I will add in a later PR. But I can also add them here.

ffi/src/proofs/change.rs
rkuris: > that helper function will also be removed in a later PR

I'm not sure it needs to be. This separates the validation of the proof from the state changes needed, so a helper even if there is only one caller lowers the cognitive complexity of both functions.

ffi/src/proofs/change.rs
bernard-avalabs: Changing to `#[cfg(test)]` causes a number of warning amount unused `use` statements. I think it is simpler to keep it as `allow` since these functions will be used shortly in one of the follow up PRs.

ffi/src/proofs/change.rs
bernard-avalabs: I've pulled `verify_proof_structure` from a later PR and modify `verify` and `verify_and_propose` to call it. I contains all of the structural tests that was in verify plus several others to match the checks in AvalancheGo.

ffi/src/proofs/change.rs
bernard-avalabs: I've pulled `verify_proof_structure` from a later PR which is now called by `verify_and_propose`. It has each of the structural tests commented.

bernard-avalabs:
> Minimum change for approval: add a diagram explaining the desired flow. Mermaid preferred but you can use text to explain it otherwise.
> 
> Further suggestions in [db7a8e2](https://github.com/ava-labs/firewood/commit/db7a8e2dbb85cc96bdfd181038ae817c3ac3e4ff)

I've added a text description of the change from the 3-step API to the 2-step API in the PR description. It also includes the motivation for this change.

Update: I've now added generator changes to line up more closely with AvalancheGo's change proof generator which will allow for simpler end proof verification in a later PR.  I've also added more structural verification checks again to more closely match AvalancheGo's change proof verifier.

bernard-avalabs:
We are going to break the change proof verification changes up into a different series of PRs.

PR #1839, opened 2026-03-25
feat(change-proof-verification): consolidate change proof API with structural verification

bernard-avalabs:
This is the first of 3 PRs for implementing change proof verification. This first PR mainly
consists of refactoring the code for verification, and improving structural verification 
that can be done before applying the batchops and creating a proposal. The next PR
will perform verification after creating a proposal from the batchops.

## Why this should be merged

  The current change proof pipeline exposes a 3-step API (verify → propose → commit)
  where the verify step performs only cursory checks. This PR consolidates it into a
  2-step API (verify_and_propose → commit) and adds comprehensive structural and
  boundary verification so that invalid proofs are rejected before a proposal is created.

## How this works

  - Replaces `fwd_verify_change_proof` + `fwd_db_propose_change_proof` with
    `fwd_db_verify_and_propose_change_proof` and `fwd_db_verify_and_commit_change_proof`
  - Removes `VerifiedChangeProofContext` intermediate type; the new
    `ChangeProofContext::verify_and_propose()` goes directly to `ProposedChangeProofContext`
  - Adds `verify_proof_structure()` which checks: inverted ranges, `DeleteRange` rejection,
    key sort order, start/end key bounds, and missing boundary proofs
  - Adds `verify_start_proof()` and `verify_end_proof()` with fallback logic for truncated proofs
  - Adds `is_complete_proof()` with root hash comparison for full-range proofs
  - Adds `verify_boundary_values()` to check boundary key values match the proposal trie
  - `ProposedChangeProofResult` gains a `VerificationFailed` variant that returns the
    original `ChangeProofContext` on failure so the caller can retry or free it
  - Adds `Proposal::path_to_key()` for trie path walking
  - Makes `PrefixOverlap` and `PrefixOverlap::from()` public
  - Includes `api::Error::ProofError` display format change (`"proof error"` → `"proof error: {0}"`)
  - Updates Go wrapper (`proofs.go`) and existing Go tests for the new 2-step API
  - New `ProofError` variants: `EndRootMismatch`, `BoundaryValueMismatch`,
    `BoundaryProofUnverifiable`, `UnsupportedDeleteRange`, `MissingBoundaryProof`
  - Removes `ProofError::ProposalIsNone`

## How this was tested

  - Existing Go tests updated and passing. More tests will be added in the later PRs.

## Breaking Changes

  [x] firewood-ffi (C api)
  [x] firewood-go (Go api)

PR #1843, opened 2026-03-26
feat(cp-verification): add verify-and-propose change proof FFI functions

bernard-avalabs:
Builds on https://github.com/ava-labs/firewood/pull/1839. Adds the FFI  functions to make use of the verify_and_propose
and verify_and_commit functions added in the previous PR. The existing 3-step 
API has not been removed. It will be removed in the next PR. Tests will also be 
updated/added in the next PR once corresponding Go functions for the 2-step 
API are added.

## Why this should be merged

  Exposes the new 2-step change proof API (verify-and-propose → commit)
  through FFI alongside the existing 3-step API. This allows Go callers
  to adopt the new API incrementally.

## How this works

  - Adds fwd_db_verify_and_propose_change_proof FFI function which takes
    ownership of the proof and returns it on verification failure via a
    new VerificationFailed result variant.
  - Adds fwd_db_verify_and_commit_change_proof for direct verify+commit.
  - Adds fwd_proposed_change_proof_to_bytes for serializing proposed proofs.
  - Adds Go bindings: Database.VerifyAndProposeChangeProof() with ownership
    transfer, Database.VerifyAndCommitChangeProof(), and
    ProposedChangeProof.MarshalBinary().
  - Old 3-step API remains functional and unchanged.

## How this was tested

  - CI

## Breaking Changes

  - None

### DIFF
--- a/ffi/src/proofs/change.rs
+++ b/ffi/src/proofs/change.rs
@@ -241,6 +241,16 @@ impl ChangeProofContext {
             return Err(api::Error::ProofError(ProofError::UnexpectedEndProof));
         }
 
+        // Reject empty end_proof when end_key is provided or batch_ops is
+        // non-empty. The honest generator always produces an end proof in
+        // these cases. Without this check, a malicious prover can force the
+        // verifier through expensive trie operations (proposal construction,
+        // root hash verification) before the proof is ultimately rejected.
+        // Matches AvalancheGo's ErrNoEndProof.
+        if proof.end_proof().is_empty() && (end_key.is_some() || !batch_ops.is_empty()) {
+            return Err(api::Error::ProofError(ProofError::MissingEndProof));
+        }
+
         Ok(())
     }
 
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -586,21 +586,21 @@ impl<T: HashedNodeReader> Merkle<T> {
             .take(limit.map_or(usize::MAX, NonZeroUsize::get))
             .collect::<Result<Box<_>, FileIoError>>()?;
 
-        let end_proof = if let Some(limit) = limit
-            && limit.get() <= batch_ops.len()
-            && iter.next().is_some()
-        {
-            // limit was provided, we hit it, and there is at least one more key
-            // end proof is for the last key provided
-            batch_ops.last().map(|largest_key| &**largest_key.key())
-        } else {
-            // limit was not hit or not provided, end proof is for the requested
-            // end key so that we can prove we have all keys up to that key
-            end_key
-        }
-        .map(|end_key| self.prove(end_key))
-        .transpose()?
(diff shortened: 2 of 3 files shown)


ffi/src/proofs/change.rs
```diff
@@ -622,6 +638,74 @@ pub extern "C" fn fwd_change_proof_find_next_key_proposed(
...
+pub extern "C" fn fwd_db_verify_and_propose_change_proof<'db>(
+    db: Option<&'db DatabaseHandle>,
+    args: VerifyAndProposeArgs,
+) -> ProposedChangeProofResult<'db> {
+    let (Some(db), Some(proof)) = (db, args.proof) else {
+        return ProposedChangeProofResult::NullHandlePointer;
+    };
+    let start_key = args.start_key.into_option();
+    let end_key = args.end_key.into_option();
+    let _guard = firewood_metrics::set_metrics_context(db.metrics_context());
```
rkuris: This seems fragile. Why is this needed? What happens if it is left out?

I noticed it's only here and on fwd_db_verify_and_commit_change_proof. Is this required because these two methods might have side-effects of calling parallel inserters, or because they might generate metrics, or something else?

You could restructure this to use invoke_with_handle (looks like this is done elsewhere).

Another choice is an invoke_with_context.

Bare minimum here is a good comment that describes what's happening.

ffi/src/proofs/change.rs
```diff
@@ -622,6 +638,74 @@ pub extern "C" fn fwd_change_proof_find_next_key_proposed(
...
+
+/// Serialize a proposed `ChangeProof` to bytes.
+#[unsafe(no_mangle)]
+pub extern "C" fn fwd_proposed_change_proof_to_bytes(
+    proof: Option<&ProposedChangeProofContext>,
+) -> ValueResult {
+    crate::invoke_with_handle(proof, |ctx| {
+        let mut vec = Vec::new();
+        ctx.proof.write_to_vec(&mut vec);
+        Ok::<_, api::Error>(Some(vec.into_boxed_slice()))
```
rkuris:
```suggestion
        vec.into_boxed_slice()
```

There is already an impl that converts from a Box'ed u8 array to a ValueResult.

rkuris (approved):
Approving, but strongly suggest fixing the two reported issues.

I'm not convinced we need to worry so much about backwards compatibility yet since there are no callers, but since this is already done we might as well finish it.

The go side got a cursory review. Recommend a more thorough review from someone more go-literate.

I wanted to think more about the edge cases but didn't. Empty change proofs, empty databases, that kind of thing.

ffi/src/proofs/change.rs
bernard-avalabs: Thanks for catching this. I've restructured it to use invoke_with_handle in both places where I did the custom invoke previously.

PR #1844, opened 2026-03-26
refactor(cp-verification): migrate to verify-and-propose API, remove legacy 3-step API

bernard-avalabs:
Builds on https://github.com/ava-labs/firewood/pull/1843. Main purpose of this PR is to switch over to the 2-step 
API and remove all of the old 3-step API code. Most of the changes stem from 
removing about 900 lines of old code.

## Why this should be merged

     Completes the migration to the 2-step change proof API. Removes the
     intermediate VerifiedChangeProofContext type and the legacy verify →
     propose → commit FFI functions, simplifying the codebase.

## How this works

     Go:
     - Migrates all tests from VerifyChangeProof + ProposeChangeProof to
       VerifyAndProposeChangeProof.
     - Removes VerifiedChangeProof type, ChangeProof.VerifyChangeProof(),
       Database.ProposeChangeProof(), and related helpers.

     Rust:
     - Removes VerifiedChangeProofContext, VerificationParams, and legacy FFI
       functions (fwd_verify_change_proof, fwd_db_propose_change_proof,
       fwd_verified_change_proof_to_bytes, fwd_free_verified_change_proof).
     - Removes VerifiedChangeProofResult from result types.
     - Simplifies ChangeProofContext.proof from Option to direct ownership.
     - Removes ProposalIsNone error variant.

 ## How this was tested

    - CI with updated tests.

 ## Breaking Changes

     [ ] firewood
     [ ] firewood-storage
     [x] firewood-ffi (C api)
     [x] firewood-go (Go api)
     [ ] fwdctl

### DIFF
--- a/ffi/src/proofs/change.rs
+++ b/ffi/src/proofs/change.rs
@@ -202,6 +202,16 @@ impl ChangeProofContext {
             return Err(api::Error::ProofError(ProofError::UnexpectedEndProof));
         }
 
+        // Reject empty end_proof when end_key is provided or batch_ops is
+        // non-empty. The honest generator always produces an end proof in
+        // these cases. Without this check, a malicious prover can force the
+        // verifier through expensive trie operations (proposal construction,
+        // root hash verification) before the proof is ultimately rejected.
+        // Matches AvalancheGo's ErrNoEndProof.
+        if proof.end_proof().is_empty() && (end_key.is_some() || !batch_ops.is_empty()) {
+            return Err(api::Error::ProofError(ProofError::MissingEndProof));
+        }
+
         Ok(())
     }
 
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -586,21 +586,21 @@ impl<T: HashedNodeReader> Merkle<T> {
             .take(limit.map_or(usize::MAX, NonZeroUsize::get))
             .collect::<Result<Box<_>, FileIoError>>()?;
 
-        let end_proof = if let Some(limit) = limit
-            && limit.get() <= batch_ops.len()
-            && iter.next().is_some()
-        {
-            // limit was provided, we hit it, and there is at least one more key
-            // end proof is for the last key provided
-            batch_ops.last().map(|largest_key| &**largest_key.key())
-        } else {
-            // limit was not hit or not provided, end proof is for the requested
-            // end key so that we can prove we have all keys up to that key
-            end_key
-        }
-        .map(|end_key| self.prove(end_key))
-        .transpose()?
-        .unwrap_or_default();
+        // The end proof is built for the last batch_op key when batch_ops
+        // is non-empty, or for end_key when batch_ops is empty (the "no
+        // changes in this range" case). When batch_ops is non-empty,
+        // end_key is not used for end proof generation — last_op_key
+        // always takes precedence. This matches AvalancheGo's convention
+        // and ensures the verifier can always determine the end proof's
+        // key deterministically from the proof content — no ambiguity, no
+        // two-key trial needed.
+        let end_proof = batch_ops
+            .last()
+            .map(|largest_key| &**largest_key.key())
+            .or(end_key)
+            .map(|end_key| self.prove(end_key))
+            .transpose()?
+            .unwrap_or_default();
 
         Ok(ChangeProof::new(start_proof, end_proof, batch_ops))
     }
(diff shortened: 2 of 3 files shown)


ffi/src/proofs/change.rs
```diff
@@ -38,11 +35,11 @@ pub struct CreateChangeProofArgs<'a> {
     /// The root hash of the starting revision. This must be provided.
     /// If the root is not found in the database, the function will return
     /// [`ChangeProofResult::StartRevisionNotFound`].
-    pub start_root: HashKey,
+    pub start_root: crate::HashKey,
```
rkuris: Why did this change? Seems better before?

Might be because you're using the other HashKey from firewood::api::HashKey. If so, then I'd consider aliasing one or the other. Maybe `firewood::api::HashKey as ApiHashKey`

ffi/src/proofs/change.rs
```diff
@@ -56,166 +53,94 @@ pub struct CreateChangeProofArgs<'a> {
...
-    pub proof: Option<&'a mut ChangeProofContext>,
-    /// The root hash of the starting revision. This must match the starting
-    /// root of the proof.
-    pub start_root: HashKey,
-    /// The root hash of the ending revision. This must match the ending root of
-    /// the proof.
-    pub end_root: HashKey,
-    /// The lower bound of the key range that the proof is expected to cover. If
-    /// `None`, the proof is expected to cover from the start of the keyspace.
+    /// The change proof to verify. Ownership is transferred to the callee.
```
rkuris: nit: Not sure "ownership is transferred to the caller" adds much value. Made me look to see why it might not be obvious, but it does seem to be.

```suggestion
    /// The change proof to verify.
```

ffi/src/proofs/change.rs
```diff
@@ -56,166 +53,94 @@ pub struct CreateChangeProofArgs<'a> {
...
+    pub proof: Option<Box<ChangeProofContext>>,
+    /// The root hash of the starting revision.
+    pub start_root: crate::HashKey,
+    /// The root hash of the ending revision.
+    pub end_root: crate::HashKey,
+    /// The lower bound of the key range that the proof is expected to cover.
```
rkuris: I think the prior comment about None made sense. Why was it removed?

ffi/src/proofs/change.rs
```diff
@@ -286,101 +213,50 @@ impl ChangeProofContext {
...
-            db,
-            root_hash,
-            end_root: self.params.end_root,
-            end_key: self.params.end_key.clone(),
-            proposal: Some(proposal.handle),
-        })
+impl ProposedChangeProofContext<'_> {
+    /// Commit a previously proposed change proof. Consumes the proposal handle.
+    fn commit(&mut self) -> Result<Option<HashKey>, api::Error> {
+        let state = std::mem::replace(&mut self.proposal_state, ProposalState::Committed(None));
```
rkuris: This is a footgun. If you call commit twice, I think you'd get success with an empty hash.

One fix is to add another variant to ProposalState.

Absolute minimum is documenting this footgun.

ffi/src/proofs/change.rs
```diff
@@ -638,121 +545,52 @@ pub extern "C" fn fwd_change_proof_find_next_key_proposed(
...
-///   an error.
+///   method. If from a parsed proof, the proof will not be verified before
+///   serialization.
 ///
 /// # Returns
 ///
 /// - [`ValueResult::NullHandlePointer`] if the caller provided a null pointer.
 /// - [`ValueResult::Some`] containing the serialized bytes if successful.
-/// - [`ValueResult::Err`] if the proof has been consumed by verification.
+/// - [`ValueResult::Err`] if the caller provided a null pointer.
```
rkuris: Looks like a copy/paste error. What happens if the caller provides a null pointer? I think it's the first one.

ffi/src/proofs/change.rs
bernard-avalabs: I think this is only a problem if you call commit twice after the first commit returned an error.

I've added an extra terminal state to handle this case and return a CommitAlreadyFailed error.

ffi/src/proofs/change.rs
bernard-avalabs: I think claude removed it when I was doing some refactoring. I will add it back.

ffi/src/proofs/change.rs
bernard-avalabs: Fixed

ffi/src/proofs/change.rs
bernard-avalabs: Added back comments for start_key and end_key

ffi/src/proofs/change.rs
bernard-avalabs: Fixed.

bernard-avalabs:
> Want to see the fix around double commits. Please consider adding a test.

I've added an extra state to handle this case. This was only a problem if the caller calls commit a second time after the first one returned an error.

I'm group all of the tests in a later PR. I'll add a test for double commits in that PR.

PR #1847, opened 2026-03-26
refactor(metrics)!: remove histograms

Elvis339:
## Why this should be merged

We removed the histograms because they were semantically wrong and the instrumentation did not produce useful signal. We agreed to drop them for now and to add a clearer design later—both for how we instrument and for which metrics we expose.

## How this works

Removes code related to histograms.

## How this was tested

CI

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [x] firewood-go (Go api)
- [ ] fwdctl


ffi/src/registry.rs
```diff
@@ -3,62 +3,15 @@
 
 //! FFI layer metric definitions.
 
-use firewood_metrics::{HistogramBucketConfig, register_histogram_with_buckets};
 use metrics::describe_counter;
 
-pub const COMMIT_MS: &str = "ffi.commit_ms";
-pub const COMMIT_COUNT: &str = "ffi.commit";
```
Elvis339, line 10: Dropping these FFI commit/propose/batch metrics they were misleading as defined and didn’t give actionable signal; we’ll replace them with a deliberate instrumentation pass later.

PR #1848, opened 2026-03-27
fix(range-proof): manual audit bug fixes

rkuris:
## Why this should be merged

A code walkthrough resulted some fixes and code cleanup.

Proofs did not verify that the value provided on the edge matched what was in the proof. 

## How this works

Verifying the value wasn't too difficult.

Added a check for start > end with a new StartAfterend error.

Slight improvements to empty edge proof handling

## How this was tested

Added a test that fails without the fix.

- test_bad_range_proof_hidden_value_on_edge_path

More tests, mostly used to attempt to find problems
- test_bad_range_proof_truncated_{start,end}_proof
- test_bad_range_proof_value_at_odd_nibble
- test_bad_range_proof_truncated_non_existent_edge
- test_bad_range_proof_start_after_end

## Breaking Changes

None

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -139,8 +139,9 @@ impl<T> From<T> for Merkle<T> {
 
 /// Verify one edge (left or right) of a range proof.
 ///
-/// Checks that the requested bound is consistent with the edge key-value pair,
-/// then verifies the proof against the root hash.
+/// Empty proofs are accepted without verification. Otherwise, checks that the
+/// requested bound is consistent with the edge key-value pair, then verifies
+/// the proof against the root hash.
 fn verify_edge<H: ProofCollection + ?Sized>(
     requested_bound: Option<&[u8]>,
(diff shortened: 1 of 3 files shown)


firewood/src/merkle/mod.rs
```diff
@@ -384,6 +389,7 @@ fn compute_root_hash_with_proofs(
 /// Returns [`api::Error::ProofError`] if the proof is structurally invalid,
 /// keys are outside the requested range, boundary proofs fail verification,
 /// or the reconstructed root hash doesn't match.
+#[allow(clippy::too_many_lines)]
```
demosdemon, line 392: 1: prefer `expect` over allow
2: I'd rather not allow this one. can this be broken down? If not now, soon.

demosdemon (approved):

firewood/src/merkle/mod.rs
rkuris: Yes, next PR will break it down. This is bad.

PR #1849, opened 2026-03-27
refactor(proofs): replace ChildMask [bool; 16] with u16 newtype

rkuris:
## Why this should be merged

A ChildMask is much cleaner as a newtype and using bit manipulations. Compiler explorer shows that most of this reduces to a few bit operations.

Also verify_range_proof became enormous, so it was split up in this diff as well. Just created verify_proof_node_values and verify_root_hash as helpers.

Also, added some unit tests for outside_children. This allowed for some refactoring and testing at a higher level to detect problems a bit earlier.

## How this was tested

CI

## Breaking Changes

None

### DIFF
--- a/firewood/src/merkle/childmask.rs
+++ b/firewood/src/merkle/childmask.rs
@@ -1,0 +1,154 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+use firewood_storage::U4;
+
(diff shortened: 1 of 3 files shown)


demosdemon:
There is already a bitmap type used to serialize proof nodes: https://github.com/ava-labs/firewood/blob/c4fb05ff2e76a1029005d8a22f5757c5006069d1/firewood/src/proofs/childmap.rs#L34

This one is better as that one was written before we removed branch_factor_256 so it was trying to be generic to support both. We should replace that one with this one.

PR #1851, opened 2026-03-27
fix(range-proof): manual audit bug fixes

rkuris:
## Why this should be merged

A code walkthrough resulted some fixes and code cleanup. This is more of a smorgasbord than I would prefer; yell and I'll break it up better.

Proofs did not verify that the value provided on the edge matched what was in the proof. Added a test that fails without the fix.

Also added a check for start > end with a new StartAfterend error.

Slight improvements to empty edge proof handling, and broke up the complex range_proof_verify function.

## How this was tested

New tests, mostly used to attempt to find problems
- test_bad_range_proof_truncated_{start,end}_proof
- test_bad_range_proof_value_at_odd_nibble
- test_bad_range_proof_hidden_value_on_edge_path
- test_bad_range_proof_truncated_non_existent_edge
- test_bad_range_proof_start_after_end

## Breaking Changes

None

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -139,8 +139,9 @@ impl<T> From<T> for Merkle<T> {
 
 /// Verify one edge (left or right) of a range proof.
 ///
-/// Checks that the requested bound is consistent with the edge key-value pair,
-/// then verifies the proof against the root hash.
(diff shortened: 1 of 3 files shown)


rkuris:
All of these were already addressed elsewhere

PR #1879, opened 2026-04-03
feat(change-proofs): Add change proof verification

rkuris:
## Why this should be merged

Probably shouldn't quite yet. This branch is, however, what we are auditing.

## How this works

See the extensive doc comments inside the code.

## How this was tested

Fuzz tests. Tamper tests. Formal verifications. Etc.

## Breaking Changes

None


firewood/src/proofs/change.rs
```diff
@@ -0,0 +1,493 @@
...
+        });
+    }
+
+    // Validate boundary proof presence against batch_ops, start_key,
+    // and end_key. The honest generator follows strict rules about when
+    // proofs should be present. These O(1) checks reject malformed
+    // proofs before expensive O(n) scans.
+
+    // Non-empty start_proof without a start_key is unverifiable — the
+    // honest generator only produces start_proof when start_key is Some.
```
alarso16: I'm sure you'll go through and fix these Claude-y comments, but this one is especially tautological. "We check this because it's required". If I understand correctly, a start proof is only not supplied if `start_range` in the request was `nothing()`. What if the `start_key` is `0x1` and the first key in the database is `0x2`?

firewood/src/proofs/change.rs
```diff
@@ -0,0 +1,493 @@
...
+    if !proof.start_proof().is_empty() && start_key.is_none() {
+        return Err(api::Error::ProofError(ProofError::UnexpectedStartProof));
+    }
+
+    match (batch_ops.is_empty(), proof.end_proof().is_empty()) {
+        // batch_ops present but no end_proof — always an error, but
+        // distinguish "no boundary proofs at all" from "just missing end".
+        (false, true) => {
```
alarso16, line 322: Why is this an error? I guess really what I'm asking is, why do you need an end proof if you're proving through the end of the db? Is it just because we don't yet have the optimization to share the start and end proof?

firewood/src/proofs/change.rs
```diff
@@ -0,0 +1,493 @@
...
+        last_op = Some(op);
+    }
+
+    // Verify end boundary proof against end_root.
+    // The key used is last_op_key when batch_ops is non-empty, or end_key
+    // when batch_ops is empty. When both are None, this block is skipped
+    // — the structural match block above guarantees the end proof is also
+    // empty in that case (UnexpectedEndProof rejects non-empty end proofs
+    // when batch_ops is empty and end_key is None).
+    if let Some(key) = last_op.map(|op| op.key().as_ref()).or(end_key) {
```
alarso16: Can this be shared with the start boundary code? Very redundant

firewood/src/merkle/mod.rs
```diff
@@ -556,6 +566,183 @@ fn verify_root_hash<H: ProofCollection<Node = ProofNode>>(
...
+        if keys_match {
+            let entry = outside_children.entry(terminal.key.clone()).or_default();
+            *entry = ChildMask::ALL;
+        }
+    }
+
+    // Compute the hybrid root hash via `compute_root_hash_with_proofs`.
+    // In-range children are hashed from the proving trie. Out-of-range
+    // children use hashes from the proof nodes.
+    let Some(root_node) = proving_merkle.root() else {
```
alarso16: The commenting around here is a little weird. Couldn't this also happen if the entire range was deleted?

alarso16 (commented):
I don't see anything immediately wrong, but I didn't look at the branch collapsing stuff

firewood/src/proofs/change.rs
rkuris: partially fixed in 97955548 -- not all comments were reviewed, but this one was.

firewood/src/proofs/change.rs
rkuris: Explained in 97955548

firewood/src/proofs/change.rs
rkuris: Done in 97955548

firewood/src/merkle/mod.rs
rkuris: It's impossible for everything to be gone here. If any edge exists, the reshaping will force a root to exist, even if there isn't anything in it.

I fixed this in 97955548 and made it way more explicit (it can and should panic in this case)

rkuris:
Stale PR

PR #1883, opened 2026-04-06
chore: Trail of bits audit branch

rkuris:
## Why this should be merged

Adds change proof verification

## How this works

See docs

## How this was tested

unit tests and fuzz tests

## Breaking Changes

None


rkuris:
The audit is complete so I am closing this PR but retaining the branch.

PR #1901, opened 2026-04-07
refactor(ci): Speed up and move flake checks into main ci build

rkuris:
## Why this should be merged

CI nix flake checks are slow due to evaluating three separate nixpkgs instances and fetching latest rust-overlay master on every run.

## How this works

- Add `inputs.nixpkgs.follows = "nixpkgs"` to rust-overlay and golang inputs, deduplicating three nixpkgs evaluations down to one
- Pin Rust toolchain to MSRV 1.94.0 instead of floating `stable.latest` to avoid cache-busting rebuilds on rust-overlay updates
- Stop auto-updating rust-overlay in CI; the lock file pin is now authoritative (update manually with `nix flake update rust-overlay`)
- Move ffi-nix job from standalone workflow into ci.yaml and delete .github/workflows/ffi-nix.yaml

## How this was tested

- Verified flake.nix syntax
- CI will validate the nix build via the relocated ffi-nix job

## Breaking Changes

None


JuanLeon2:
How is consolidating ffi-nix into ci.yml expected to help?

ffi/flake.nix
```diff
@@ -9,10 +9,16 @@
 
   inputs = {
     nixpkgs.url = "https://flakehub.com/f/NixOS/nixpkgs/0.2505.*.tar.gz";
-    rust-overlay.url = "github:oxalica/rust-overlay";
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs.nixpkgs.follows = "nixpkgs";
```
JuanLeon2, line 14: Is this a no-op given that flake.lock lines 110 and 116 pin down two separate nixpkgs?

justfile
```diff
@@ -124,8 +124,8 @@ update-ffi-flake: check-nix
     set -euo pipefail
     cd ffi
 
-    echo "ensuring flake lock file is current for golang and rust-overlay"
-    nix flake update golang rust-overlay
+    echo "ensuring flake lock file is current for golang"
+    nix flake update golang
```
JuanLeon2, line 128: WHy are you taking out rust-overlay?

.github/workflows/ffi-nix.yaml
```diff
@@ -1,24 +0,0 @@
-name: ffi-nix
-
-on:
-  push:
```
JuanLeon2, line 4: Instead of deleting this file, can we say here:

on
  workflow_call:

and then on ci.yaml replace the copy of the below with

  uses: ./.github/workflows/ffi-nix.yaml

?

This way the behavior is the same but this workflow can be called separately (manually or from other contexts) and changes to this task remain localized.

.github/workflows/ci.yaml
```diff
@@ -358,3 +358,14 @@ jobs:
         run: cargo install cargo-machete
       - name: Check for stale dependencies
         run: cargo machete --with-metadata
+
+  ffi-nix:
+    runs-on: ubuntu-latest
```
JuanLeon2, line 363: Can we replace lines 363-371 with
    uses: ./.github/workflows/ffi-nix.yaml

and not delete the ffi-nix.yaml (instead replace its "on" clause with workflow_call: as I mention in a comment over there)?

JuanLeon2 (commented):
Why is consolidating ffi-nix into ci.yml supposed to help?

rkuris:
Stale PR

PR #1904, opened 2026-04-08
feat: add collapse functions for change proof verification

bernard-avalabs:
## Why this should be merged

The restructure approach for change proof verification requires collapsing trie nodes to fit the proof nodes created from the end trie. This PR adds these functions from the rkuris/cp-restructure-squashed branch with some minor changes.

## How this works

  - collapse_root_to_path and collapse_branch_to_path are entry points that reshape from the root or between two proof nodes
  - collapse_navigate walks the proving trie to the parent proof node
  - collapse_descend enters the on-path child from the parent proof node
  - collapse_strip does the actual stripping of off-path children and flattening of single-child branches
  - child_in_range and build_child_prefix are helpers for range checking during stripping
  - take_only_child on Children supports the flattening step
  - ProofNodeUnreachable error variant was added and replaces silent Ok returns in defensive paths

## How this was tested

 - 9 new collapse-specific tests covering stripping, flattening, range rejection, and error paths. Tests for private functions are in merkle/collapse.rs, and tests for pub(crate) functions are in merkle/tests/collapse.rs.
 - Integration tests will be added in a later PR once the rest of change proofs has been added 

## Breaking Changes

 - None


firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+/// Builds the accumulated nibble prefix for a child node by extending
+/// `prefix` with the descent nibble and the child's `partial_path`.
+fn build_child_prefix(prefix: &[u8], nibble: u8, node: &Node) -> Box<[u8]> {
+    let pp = &node.partial_path().0;
+    let capacity = prefix
+        .len()
+        .wrapping_add(mem::size_of::<u8>())
+        .wrapping_add(pp.len());
```
demosdemon: AI reviewers and some human reviewers may question this. We might want to add a comment but idk how many times it should be documented. However, this is guaranteed to **not overflow usize**. This is because the rust does not allow slices to contain more than `isize::MAX` elements which is `usize::MAX >> 1`. 

```rust
assert_eq!(
    (isize::MAX as usize) + (isize::MAX as usize) + 1_usize,
    usize::MAX,
);
```

Because the above expression is true and the constraints placed on slices by rust, this guarantees the two wrapping operations cannot overflow.

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+/// Used by [`Merkle::collapse_strip`] to distinguish in-range children
+/// (which indicate tampered `batch_ops`) from out-of-range children (which
+/// are stripped normally).
+fn child_in_range(
+    acc_prefix: &[u8],
+    nibble: PathComponent,
+    start_nib: &[u8],
+    end_nib: &[u8],
+) -> bool {
+    let mut child_path = Vec::with_capacity(acc_prefix.len().saturating_add(1));
```
demosdemon: we should use wrapping_add here for consistency with below, especially since we know it cannot overflow usize.

However, `Vec::with_capacity(x)` will panic if `x > (isize::MAX as usize)`. For path operations, I doubt that's worth caring about at this point. If someone managed to give us a path that causes that to happen, we've already exceeded the memory limits of any known computer.

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+    #[cfg_attr(not(test), expect(dead_code))]
+    pub(crate) fn collapse_root_to_path(
+        &mut self,
+        target: &[PathComponent],
+        range: Option<(&[u8], &[u8])>,
+    ) -> Result<(), api::Error> {
+        // The root's partial_path consumes some prefix of target.
+        // Only collapse if target extends beyond the root's partial_path.
+        let mut root_node = std::mem::take(self.nodestore.root_mut())
+            .expect("reconciliation guarantees a root node exists");
```
demosdemon: the `take` api on the `Option` is preferred here.

```suggestion
        let mut root_node = self.nodestore.root_mut()
            .take()
            .expect("reconciliation guarantees a root node exists");
```

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+    acc_prefix: &[u8],
+    nibble: PathComponent,
+    start_nib: &[u8],
+    end_nib: &[u8],
+) -> bool {
+    let mut child_path = Vec::with_capacity(acc_prefix.len().saturating_add(1));
+    child_path.extend_from_slice(acc_prefix);
+    child_path.push(nibble.as_u8());
+    child_path.as_slice() >= start_nib && child_path.as_slice() <= end_nib
+}
```
demosdemon: however, constructing a vector just to compare feels wasteful. This can be done without allocation.

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+    ) -> Result<(), api::Error> {
+        let suffix = to.strip_prefix(from).expect("to must start with from");
+        let parent_prefix: Box<[u8]> = from.iter().map(|c| c.as_u8()).collect();
+
+        let root = self.nodestore.root_mut();
+        let Some(root_node) = std::mem::take(root) else {
+            return Ok(());
+        };
+        let root_node = self.collapse_navigate(root_node, from, suffix, &parent_prefix, range)?;
+        *self.nodestore.root_mut() = root_node.into();
```
demosdemon: `take` here as well.

```suggestion
        let Some(root_node) = self.nodestore.root_mut().take() else {
            return Ok(());
        };
        let root_node = self.collapse_navigate(root_node, from, suffix, &parent_prefix, range)?;
        self.nodestore.root_mut() = Some(root_node);
```

This one is interesting cause I'm also compelled do do something like

```rust
let mut root = self.nodestore.root_mut();
let Some(root_node) = root.take() {
    return Ok(());
}
*root = Some(self.collapse_navigate(root_node, from, suffix, &parent_prefix, range)?);
```

but I suspect that might run into a mutable alias problem

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// `range`: `(start_nibbles, end_nibbles)` for the proven range.
+    /// In-range children that are also proposal-local trigger rejection.
+    #[cfg_attr(not(test), expect(dead_code))]
+    pub(crate) fn collapse_branch_to_path(
+        &mut self,
+        from: &[PathComponent],
+        to: &[PathComponent],
+        range: Option<(&[u8], &[u8])>,
+    ) -> Result<(), api::Error> {
+        let suffix = to.strip_prefix(from).expect("to must start with from");
```
demosdemon: `except` could return an error unless this is guaranteed. and if it's guaranteed, there should be a better way to represent this.

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+        if !deeper.is_empty() {
+            let child_prefix = build_child_prefix(acc_prefix, on_path.0.as_u8(), &child_node);
+            child_node = self.collapse_strip(child_node, deeper, &child_prefix, range)?;
+        }
+
+        branch.children[on_path] = Some(Child::Node(child_node));
+        debug_assert!(
+            branch.value.is_none(),
+            "intermediate branch value should have been cleared"
+        );
```
demosdemon: this is redundant. there is nothing between line 261 and here that could make this fail.

firewood/src/merkle/collapse.rs
```diff
@@ -0,0 +1,503 @@
...
+            // No child pointer exists on path to child proof node.
+            return Err(api::Error::ProofError(ProofError::ProofNodeUnreachable));
+        };
+
+        let mut child_node = self.read_for_update(child)?;
+
+        let child_prefix = build_child_prefix(acc_prefix, first.0.as_u8(), &child_node);
+
+        let child_pp_len = child_node.partial_path().len();
+        let after_child = remaining.get(child_pp_len..).unwrap_or_default();
```
demosdemon: this should verify that the `remaining` path actually begins with the `child_node`'s path. If the node's path is shorter than `remaining`, this unwrap_or_default will silently return an empty string adding information to the child's path that was not there before. If they're not equal up to this point, we've clobbered the path.

firewood/src/merkle/collapse.rs
bernard-avalabs: I can revert to the child_in_range in rkuris/cp-restructure-squashed which doesn't perform an allocation. It is just more complicated.

firewood/src/merkle/collapse.rs
bernard-avalabs: I changed it from an explicit check to a debug_assert mostly as documentation. But I agree it is better to just remove.

firewood/src/merkle/collapse.rs
bernard-avalabs: `from` and `to` are the PathComponents from consecutive proof nodes. If the boundary proof is malformed, then this invariant is not guaranteed. With that said, `verify_change_proof_structure` validates that each proof node's key is a prefix of the next before this function is called, so unless there is some refactoring, this should never fail.

I will change this to return an error.

firewood/src/merkle/collapse.rs
bernard-avalabs: Good catch. I think this also occurs in collapse_strip. I'll add checks for both.

firewood/src/merkle/collapse.rs
bernard-avalabs: I've removed the `wrapping_add`s and added a comment and a  #[allow(clippy::arithmetic_side_effects)].

firewood/src/merkle/collapse.rs
bernard-avalabs: I've replaced `child_in_range` with the version from rkuris/cp-restructure-squashed. This version does not have this code as it doesn't perform allocation.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed

firewood/src/merkle/collapse.rs
bernard-avalabs: I've replaced child_in_range with the version that doesn't perform allocation.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed. Now it returns an error.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed. Line removed.

firewood/src/merkle/collapse.rs
bernard-avalabs: I've added a helper function `consume_partial_path` to check that `remaining` begins with the child_node's path and returns only the suffix. It is now used in both `collapse_descend` and `collapse_strip`. I've also added a unit test for this case.

demosdemon (approved):

PR #1905, opened 2026-04-08
chore: remove unused mutable iter_present from Children

bernard-avalabs:
## Why this should be merged                                                                                                                                                                                                                         
                  
IterPresentMut and the mutable iter_present method on Children<&mut Option<T>> are dead code. Removing them simplifies the Children API. The lifetime on the remaining impl block is also simplified since the explicit 'a is no longer needed.

take_only_child will be added in https://github.com/ava-labs/firewood/pull/1904/
                                                                                                                                                                                                                                                    
## How this works                                                                                                                                                                                                                                    
   
  - Remove IterPresentMut type alias                                                                                                                                                                                                                
  - Remove iter_present method from impl Children<&'a mut Option<T>>
  - Simplify impl<'a, T> Children<&'a mut Option<T>> to impl<T> Children<&mut Option<T>>
                                                                                                                                                                                                                                                    
## How this was tested                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  CI                                                                                                                                                  
                  
## Breaking Changes                                                                                                                                                                                                                                  
                  
  - None.

### DIFF
--- a/storage/src/node/children.rs
+++ b/storage/src/node/children.rs
@@ -18,12 +18,6 @@ pub type IterPresentRef<'a, T> = std::iter::FilterMap<
     fn((PathComponent, &'a Option<T>)) -> Option<(PathComponent, &'a T)>,
 >;
 
-/// The type of iterator returned by [`Children::iter_present`].
-pub type IterPresentMut<'a, T> = std::iter::FilterMap<
-    ChildrenSlots<&'a mut Option<T>>,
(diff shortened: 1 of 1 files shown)


demosdemon:
I'd rather not remove this. `Children` is a generic type and I'd like to keep it usable for other use cases.

bernard-avalabs:
> I'd rather not remove this. `Children` is a generic type and I'd like to keep it usable for other use cases.

I'm fine with closing this PR if we want to keep the unused code for other use cases.

PR #1906, opened 2026-04-08
feat(storage): autogen area sizes hash and names

### LINKED ISSUE #1607: perf: sha256 hash of `AREA_SIZES` calculated every time a db header is crafted
This only occurs when the database header is crafted from thin air, which occurs at two points in time:

- Database Initialization; when creating the initial header (this does happen a lot in tests)
- RootStore reads; when acquiring the nodestore of a root no longer in memory

Because `AREA_SIZES` is a constant, we should calculate the hash at compile-time and make it constant as well.

Additionally, the algorithm is always sha256 even for ethhash mode, which uses keccak256 for everything else. This isn't necessarily a problem but the inconsistency may lead to confusion.

## Pointers
- <https://github.com/ava-labs/firewood/blob/953dad3100ee9118a1e0bb5324908ca37a009c56/storage/src/nodestore/primitives.rs#L54-L60>
- <https://github.com/ava-labs/firewood/blob/953dad3100ee9118a1e0bb5324908ca37a009c56/storage/src/nodestore/header.rs#L202-L205>

demosdemon:
## Why this should be merged

Closes #1607

## How this works

This uses a build script to auto-generate a source file with the area sizes, the string representation, and the pre-computed sha256 hash.

Example output: https://gist.github.com/demosdemon/3762331b0bfbe03ec9ee81f51ce14f03

## How this was tested

CI and `cargo nextest`

## Breaking Changes

None


storage/build.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// [`NodeStore`] divides the linear store into blocks of different sizes. This array
+/// contains every valid block size, in ascending order.
+///
+/// [`NodeStore`]: crate::nodestore::NodeStore
+const AREA_SIZES: [u64; NUM_AREA_SIZES] = [",
+    )?;
+
+    for (size, expr) in std::iter::zip(AREA_SIZES, AREA_EXPRS) {
+        writeln!(output, "    {expr}, // {size} bytes")?;
```
demosdemon, line 102: I intentionally chose to print the original expression and not just the calculated value because I liked carrying how the value was calculated through to the generated code.

storage/Cargo.toml
```diff
@@ -63,6 +63,10 @@ rand.workspace = true
 tempfile.workspace = true
 test-case.workspace = true
 
+[build-dependencies]
+bytemuck.workspace = true
```
rkuris, line 67: nit: making this a build-time dependency adds a dependency we could work around with the loop the old code used (to_ne_bytes).

Not a big deal though, maybe not even worth fixing.

storage/build.rs
```diff
@@ -0,0 +1,130 @@
...
+    1024 << 12,
+    1024 << 13,
+    1024 << 14,
+];
+
+fn main() {
+    let out_dir = std::env::var("OUT_DIR").expect("OUT_DIR not set");
+    let dest_path = std::path::Path::new(&out_dir).join("area_sizes.rs");
+    let contents = generate_area_sizes_rs().expect("Failed to generate area_sizes.rs");
+    std::fs::write(dest_path, contents).expect("Failed to write area_sizes.rs");
```
rkuris, line 74: Writing this every time means that on every compile, the mtime of this file will change, so anything that includes the file will be recompiled.

We should read the old file and only update it if it has changed.

storage/build.rs
```diff
@@ -0,0 +1,130 @@
...
+    1024 << 8,
+    1024 << 9,
+    1024 << 10,
+    1024 << 11,
+    1024 << 12,
+    1024 << 13,
+    1024 << 14,
+];
+
+fn main() {
```
rkuris, line 69: This should emit "cargo:rerun-if-changed=build.rs" otherwise this build script runs unnecessarily every time any file in the entire storage crate changes.

From https://doc.rust-lang.org/cargo/reference/build-scripts.html
> If the build script inherently does not need to re-run under any circumstance, then emitting cargo::rerun-if-changed=build.rs is a simple way to prevent it from being re-run (otherwise, the default if no rerun-if instructions are emitted is to scan the entire package directory for changes). Cargo automatically handles whether or not the script itself needs to be recompiled, and of course the script will be re-run after it has been recompiled. Otherwise, specifying build.rs is redundant and unnecessary.

rkuris (changes requested):
Minimum fix: rerun-if-changed
Also recommend the "only rewrite this file if it's different" but that's not a showstopper.

demosdemon:
> Minimum fix: rerun-if-changed Also recommend the "only rewrite this file if it's different" but that's not a showstopper.

doing one makes the other irrelvant

storage/Cargo.toml
demosdemon: bytemuck is already a hard dependency. this adds nothing

storage/build.rs
demosdemon: reading the file and comparing for differences is irrelevant if the build script only runs when the build script changes. and, if the build script is running then everything is already being recompiled

rkuris (approved):

PR #1910, opened 2026-04-09
fix(proofs): proofs of nodes that contain storage accounts are unverifiable

rkuris:
## Why this should be merged

Because we didn't actually store the value being hashed in the node,
when nodes are generated for proofs they contain sentinel values for
the account storageRoot. That means proofs generated that contain
accounts are unusable.

This relies on the fact that coreth always inserts zeroes as the hash
value for the storage root and expects firewood to calculate the hash
as necessary.

## How this works

This fix addresses the bug in a few ways. First, if a database is
created from scratch, it will calculate and store the roots along with
the node. This means that everything "just works", including even
re-hashing.

For existing databases, the problem is a bit more complicated. If a
node is found while traversing a proof, we check to see if there is a
sentinel value of all zeroes, and if so, we recompute the hash at
proof generation time. This applies both to ProofNode construction
(proofs/types.rs) and to range proof key_values emitted by the
key-value iterator (iter.rs).

The database version was updated so we can get a little more aggressive
about not checking or recalculating the hash for older databases later.

Also fixed: the merkle dump display for non-RLP values with ethhash
enabled. The write_attributes! macro called rlp::as_list() on non-list
values, which silently returns Ok(vec![]) due to an rlp crate quirk,
producing "rlp=[]" for plain values. Now checks the first byte before
attempting RLP list decode.

## How this was tested

- New test: `test_range_proof_accounts_have_computed_storage_root` verifies
  range proofs contain correct storageRoot for leaf and branch account nodes.
- New test: `test_range_proof_fixes_legacy_zeroed_storage_root` surgically
  clobbers a MemStore to simulate a pre-fix database, then verifies
  range_proof() still produces a valid proof.
- Fixed existing range proof tests (`test_reverse_single_side_range_proof`,
  `test_both_sides_range_proof`) to use committed values via `stored_key_values()`
  instead of raw input values, since ethhash hashing may rewrite account values.
- Full test suite: 559 passed, 0 failed.

## Breaking Changes

None

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,7 @@ metrics-util = "0.20.1"
 metrics-exporter-prometheus = { version = "0.18.1", default-features = false }
 nonzero_ext = "0.3.0"
 parking_lot = "0.12.5"
+rlp = "0.6.1"
 rand_distr = "0.6.0"
 sha2 = "0.10.9"
 smallvec = { version = "1.15.1", features = ["write", "union", "const_new"] }
--- a/firewood/Cargo.toml
+++ b/firewood/Cargo.toml
@@ -35,6 +35,7 @@ hex.workspace = true
 integer-encoding.workspace = true
 thiserror.workspace = true
 # Regular dependencies
+rlp = { workspace = true, optional = true }
 typed-builder = "0.23.2"
 rayon = "1.11.0"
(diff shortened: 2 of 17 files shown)


rkuris:
Closed in favor of a different design

PR #1911, opened 2026-04-10
refactor(metrics): add histograms and remove use wrappers

### LINKED ISSUE #1628: metrics: ensure all metrics generated by firewood are visible through the ffi
@alarso16 reported that some metrics are not visible on one of his bootstrap tests. We'll need to double check and ensure all the metrics are visible. Also, It'd be good to have some go tests to verify this.

demosdemon:
## Why this should be merged

Simplifies and unifies the metrics macro API by removing redundant thin-wrapper
macros and aligning all three metric types on the same call pattern.

Previously, `firewood_increment!` and `firewood_set!` were one-shot wrappers
that accepted a metric name as an arbitrary expression (e.g.
`crate::registry::FOO`) and called `.increment()` / `.set()` internally.
`firewood_histogram!` is new and uses a different, cleaner design: it accepts a
bare identifier and resolved the name as `crate::registry::NAME` in the
expansion — intentionally using `crate` (invocation-site) rather than `$crate`
(definition-site) so each crate's own `registry` module is used without any
path prefix at the call site.

This PR extends that design to all metric types and removes the extra macros.

## How this works

- **Removes `firewood_increment!` and `firewood_set!`**. Their call sites are
  inlined to `firewood_counter!(NAME).increment(value)` and
  `firewood_gauge!(NAME).set_integer(value)`.

- **Updates `firewood_counter!` and `firewood_gauge!`** to accept `$name:ident`
  instead of `$name:expr`, expanding to `crate::registry::$name` — matching the
  pattern already established by `firewood_histogram!`.

- **Documents `firewood_histogram!`** with a full doc comment, including an
  explanation of why `crate::registry` is used instead of `$crate::registry`.

- **Adds `GaugeExt` and `CounterExt` extension traits** to `firewood_metrics`.
  `Gauge::set` requires `f64` and casting integer values triggers
  `cast_precision_loss` at every call site. `GaugeExt::set_integer` accepts
  `u64` or `usize` and does the cast once in the trait impl (with a single
  lint suppression). `CounterExt::increment_f64` mirrors this for callers that
  compute measurements as `f64` and need to store them in a counter.

- **Suppresses `clippy::crate_in_macro_def`** on all three macros using
  `#[expect(..., reason = ...)]`, documenting the intentional design.

Depends On: #1867

## How this was tested

```
cargo doc --no-deps
```

All 529 tests pass, zero clippy warnings, docs build clean.

## Breaking Changes

N/A


ffi/src/jemalloc_metrics.rs
```diff
@@ -72,7 +72,7 @@ pub fn refresh() {
 
     for stat in JEMALLOC_STATS {
         match (stat.read)() {
-            Ok(v) => firewood_set!(stat.name, v),
+            Ok(v) => ::metrics::gauge!(stat.name).set_integer(v),
```
RodrigoVillar, line 75: q: why not call `firewood_gauge!()` instead?

RodrigoVillar (approved):
LGTM - had one question

ffi/src/jemalloc_metrics.rs
demosdemon: The way `firewood_gauge` is now written, the first parameter must be an identifier defined inside `crate::registry`. `stat.name` is a runtime variable and doesn't work. This is because of the way we have the jemalloc metrics specially setup in this module.

rkuris (approved):

PR #1912, opened 2026-04-10
feat(metrics)!: Overhaul metrics for Prometheus best practices

### LINKED ISSUE #983: Emit gauge with number of nodes that are deleted but not yet reclaimed on disk
Reclaiming these nodes after a crash is not possible unless we implement https://github.com/ava-labs/firewood/issues/832 (delete-node log).
We think the benefit of implementing the deleted-node log is small.   
This gauge will allow us to justify that we are right, and give us a warning if we are not.

### LINKED ISSUE #1301: chore: Add active revision counters to dashboard

### LINKED ISSUE #1628: metrics: ensure all metrics generated by firewood are visible through the ffi

### LINKED ISSUE #1672: rename ffi.* metrics
The metrics named ffi.* are misleading since they don't measure any ffi time. Additionally, they overlap very strongly with the rust metrics for proposal creation and committing. We should consolidate these with the actual rust metrics and drop the ffi naming.

### LINKED ISSUE #1700: chore: Metric naming
**Describe the bug**
Some metrics provide warnings in grafana because they do not follow the expected naming scheme. For example, avalanche_evm_firewood_cache_freelist does not end in _total, _sum, _count, nor _bucket. This generates a warning when the counter is used, that it might not be a counter.

**To Reproduce**
open grafana and create a graph using this variable, and observe the warning

**Expected behavior**
Our variables should follow the conventions

**Possible Solutions**
Audit all the metric names and macros

Note that this might be considered a breaking change as it will mess up our dashboards, so I'd like to get this in as soon as possible.

### LINKED ISSUE #1735: Metrics should use base units, not ms

demosdemon:
## Why this should be merged

Firewood's metrics were inconsistently named (dot-notation, no namespace prefix, abbreviated units like `_ms`/`_ns`), used counters to accumulate durations (preventing percentile calculation), and contained redundant or low-value entries. This brings all metrics into alignment with Prometheus conventions and adds coverage for previously uninstrumented paths.

## How this works

- Rename all metrics to `firewood_{subsystem}_{noun}_{unit}` with `_total` suffix on counters and base-unit suffixes (`_seconds`, `_bytes`) on histograms/gauges; no dots, no camelCase
- Replace counter-based duration accumulators (`_ms`, `_ns`) with proper histograms so percentiles (p50/p99) can be computed
- Remove the vestigial `DbMetrics` struct and `Db::metrics()` public API
- Add new counters: commits, I/O reads/writes/bytes, nodes allocated/wasted per free-list size class
- Add new gauges: database file size, free-list entries per size class
- Add new histograms: propose duration, persist-cycle duration, lock-wait times for `in_memory_revisions` and `by_hash` in `RevisionManager`, persist-worker submission duration
- Overhaul `#[crate::metrics]` proc macro to accept a registry `Ident` instead of a string literal; the compiler now validates that both the counter (`IDENT`) and histogram (`IDENT_DURATION_SECONDS`) constants exist in `crate::registry`, enabling bucket configuration via `define_metrics!`
- Add Go-side prometheus registry in `ffi/metrics.go` with histograms for proof marshal/unmarshal operations (`firewood_go_proof_marshal_duration_seconds`, `firewood_go_proof_unmarshal_duration_seconds`); merged into `GatherRenderedMetrics()` alongside Rust-recorder metrics
- Update `ffi/metrics_test.go` to use renamed metric names and the corrected `firewood_gather_duration_seconds` histogram name

The updated [METRICS](https://github.com/ava-labs/firewood/blob/51065a71352f227b286ba63a23f3a2600f1b4fb7/METRICS.md) file describes all of the currently defined metrics.

## How this was tested

- `cargo nextest run --workspace --features ethhash,logger --all-targets` (529/529 pass)
- `cargo clippy --workspace --features ethhash,logger --all-targets` (no warnings)
- `cargo doc --no-deps` (no warnings)
- `go build ./...` in `ffi/` (clean)

## Breaking Changes

All metric names have changed. Any existing dashboards, alerts, or recording rules referencing the old names (`proposals`, `insert`, `flush_nodes`, `io.read_ms`, `replay.propose_ns`, `ffi.gather_duration_seconds`, `jemalloc.active_bytes`, etc.) will need to be updated to the new `firewood_*` naming scheme.


rkuris (approved):
This looks fine to me. A few questions but let's get this merged and maybe we can address as followups.

I think there are a few other macros we could consider for timings.

We will need to work with one of the data team to get all these metrics migrated over. Can we generate an old->new naming guide for that effort?

PR #1914, opened 2026-04-11
refactor: update reconcile_branch_proof_node with callback interface

bernard-avalabs:
## Why this should be merged                                                                                             
                  
This PR continues the large change proof verification merge. In this PR, we are updating reconcile_branch_proof_node. The current implementation hardcodes conflict resolution (always reject value mismatches). Change proof  verification requires different strategies such as accepting proof values for out-of-range ancestors and clearing values for deleted keys. The callback interface lets the caller decide.

I've moved the reconcile code into its own reconcile.rs, and its tests to tests/reconcile.rs to move more functions out of mod.rs.          
                  
## How this works                                                                                                        
  
  - Add get_branch_from_nibbles_mut and helper to merkle/mod.rs — navigates to a branch node and returns a mutable      
  reference without disk I/O. Only works for in-memory nodes, which is guaranteed after insert_branch_from_nibbles.
  - Move reconcile_branch_proof_node to merkle/reconcile.rs with a new on_conflict callback parameter. When the proof node's value differs from the trie's value, the callback decides: Ok(Some(v)) to set, Ok(None) to clear, or Err(...) to reject.
  - Update verify_range_proof to pass a callback that accepts the proof's value.       
  - Remove ReconcileResult enum — no longer needed with the callback interface.                                         
  - Mark get_node_from_nibbles as #[cfg(test)] — no production callers remain.
 
### ValueDigest::Hash fix (merkledb mode)                                                                                                                                                                                                             
                                                                                                                                                                                                                                                    
In merkledb mode (non-ethhash), values >= 32 bytes are serialized as Hash(sha256(value)) in proof nodes. Two issues were identified during review:                                                                                                
                                                                                                                                                                                                                                                    
  1. reconcile_branch_proof_node: Hash was mapped to proof_value = None, which could trigger the callback incorrectly. Fixed by adding a Hash arm that verifies the hash against the branch value (fast path when they match) or falls through to let the comparison and callback handle it.
  2. compute_root_hash_with_proofs: Always used branch.value for hashing, but out-of-range branches may have no value while the proof node carries the Hash. Fixed by falling back to the proof node's value_digest when the branch has no value. The proof node's hash chain was verified by value_digest() during boundary proof validation, and reconcile_branch_proof_node verified the hash against the branch value when present.   


## How this was tested                                                                                                   
                                                                                                                        
  - 4 updated reconcile tests in tests/reconcile.rs:                                                                    
    - test_reconcile_branch_proof_node_creates_missing_branch_without_value — no value in proof, callback is never called                                                                                                                
    - test_reconcile_branch_proof_node_sets_missing_value_via_callback — callback accepts the proof's value
    - test_reconcile_branch_proof_node_clears_value_via_callback — callback clears the trie's value                     
    - test_reconcile_branch_proof_node_rejects_conflict_via_callback — callback rejects, value unchanged
  - 1 new test for get_branch_from_nibbles_mut — verifies found branch, missing path, and leaf (not branch) cases 
  - 2 merkledb-mode regression tests for ValueDigest::Hash:                                                                                                                                                                                         
    - test_range_proof_with_hashed_value: in-range prefix key with 32-byte value                                                                                                                                                                    
    - test_empty_range_proof_with_hashed_value: out-of-range with empty proving trie                                  
                                                                                                                        
  Breaking Changes                                                                                                      
                                                                                                                        
  - None.


firewood/src/merkle/mod.rs
```diff
@@ -539,7 +540,10 @@ fn verify_range_proof_root_hash<H: ProofCollection<Node = ProofNode>>(
     // Conflicting proof nodes (same key, different data) are rejected.
     let mut proof_node_map: HashMap<PathBuf, &ProofNode> = HashMap::new();
     for proof_node in all_proof_nodes {
-        proving_merkle.reconcile_branch_proof_node(proof_node)?;
+        proving_merkle.reconcile_branch_proof_node(proof_node, |pn| match &pn.value_digest {
+            Some(ValueDigest::Value(v)) => Ok(Some(v.clone())),
+            _ => Ok(None),
```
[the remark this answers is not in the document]
bernard-avalabs: The purpose of `reconcile_branch_proof_node` is to change the shape of the merkle trie so the hash computation is correct. 

Returning Ok(None) is correct behavior here.

firewood/src/merkle/reconcile.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        let proof_value = match proof_node.value_digest.as_ref() {
+            Some(ValueDigest::Value(v)) => Some(v.as_ref()),
+            _ => None,
+        };
+
+        if proof_value == branch.value.as_deref() {
+            return Ok(());
+        }
+
```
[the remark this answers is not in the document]
bernard-avalabs: Treating it as "no value" is intentional. We can't compare with the branch's value if it doesn't exist in the proof node.

firewood/src/merkle/reconcile.rs
```diff
@@ -0,0 +1,66 @@
...
+        // Values differ — let the caller decide what to do.
+        branch.value = on_conflict(proof_node)?;
```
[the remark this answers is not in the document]
bernard-avalabs: Odd nibble check already performed earlier in the function.

firewood/src/merkle/mod.rs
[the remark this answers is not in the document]
demosdemon: This contradicts your PR description:

> Update verify_range_proof to pass a callback that accepts the proof's value (preserving current behavior).


Case | Old behavior | New callback behavior
-- | -- | --
Proof has value Y, trie has different value X | Err(ProofError::UnexpectedValue) | Ok(Some(Y)) — silently overwrites with proof value
Proof has no value, trie has value X | Returns NoValue; X is preserved | Ok(None) — clears X from the branch

firewood/src/merkle/reconcile.rs
```diff
@@ -0,0 +1,67 @@
...
+///   path components) and an optional value digest to reconcile.
+/// * `on_conflict` - Called only when the proof node's value and the
+///   existing branch value conflict. Returning `Ok(Some(value))` stores
+///   that value in the branch, `Ok(None)` clears the branch value, and
+///   `Err(...)` aborts reconciliation with that error.
+///
+/// ## Errors
+///
+/// Returns an error if the proof is structurally invalid or if
+/// `on_conflict` returns `Err` while resolving a value conflict.
```
demosdemon, line 28: nit: move down onto the function. this doc won't surface anywhere.

firewood/src/merkle/mod.rs
```diff
@@ -540,7 +541,10 @@ fn verify_range_proof_root_hash<H: ProofCollection<Node = ProofNode>>(
     // Conflicting proof nodes (same key, different data) are rejected.
     let mut proof_node_map: HashMap<PathBuf, &ProofNode> = HashMap::new();
     for proof_node in all_proof_nodes {
-        proving_merkle.reconcile_branch_proof_node(proof_node)?;
+        proving_merkle.reconcile_branch_proof_node(proof_node, |pn| match &pn.value_digest {
+            Some(ValueDigest::Value(v)) => Ok(Some(v.clone())),
```
demosdemon, line 545: As mentioned by Copilot, this is ignoring when the value digest is `ValueDigest::Hash` which is true for proof nodes generated by the merkeldb hash method when the value is >= 32 bytes. Its behavior also contradicts your PR description. The behavior here is not the same as before. And, the removed tests specifically cover this case. Tests no longer cover any of these cases so it's unclear what the intent is.

firewood/src/merkle/reconcile.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        // insert_branch_from_nibbles guarantees a branch exists at this path
+        // and that all nodes along it are in-memory, so this cannot fail.
+        let branch = Self::get_branch_from_nibbles_mut(self.nodestore.root_mut(), &proof_node.key)
+            .ok_or(ProofError::NodeNotInTrie)?;
+
+        let proof_value = match proof_node.value_digest.as_ref() {
+            Some(ValueDigest::Value(v)) => Some(v.as_ref()),
+            _ => None,
+        };
```
demosdemon, line 57: this is ignoring `ValueDigest::Hash`, which will be set on merkledb proof nodes when the value is >= 32 bytes.

firewood/src/merkle/mod.rs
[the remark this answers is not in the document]
bernard-avalabs: PR description is definitely wrong. 

I'm digging into this more to see if we are handling this correctly.

firewood/src/merkle/mod.rs
bernard-avalabs: I'm digging into this more to see if we are handling this correctly.

bernard-avalabs: I think this is a bug for when merkledb hashing for values >= 32 bytes.

I suspect this change was made to support change proof verification, but it may be wrong for both when values >= 32 bytes. I'll update this PR with a fix and some new test cases.

bernard-avalabs:
@demosdemon I was able to confirm that the code did not correctly handle merkledb hashing for values >= 32 bytes.

I've committed a fix. The fix includes the following:

1. Short-circuits `reconcile_branch_proof_node` if the hash in the proof node matches the hash of the value in the proof trie.
2. For range proofs, the callback used in `reconcile_branch_proof_node` accepts the proof node's full value (ValueDigest::Value) when present. Otherwise, it will return UnexpectedValue.
3. For the case where the proof node has a hash, but the proof trie has no value, the callback is not called. `compute_root_hash_with_proofs` has been modified to accept the value digest of the proof node in this case.

firewood/src/merkle/tests/range.rs
```diff
@@ -1774,3 +1774,81 @@ fn test_right_edge_boundary_prefix_of_terminal() {
...
+/// and children (prefix key), the deserialized proof node carries Hash
+/// instead of Value. The reconcile step must not clear the branch value
+/// when the hash matches.
+///
+/// Setup: \x10 is a prefix of \x10\x20, making \x10 a branch with a value
+/// AND children. The 32-byte value at \x10 triggers `ValueDigest::Hash` after
+/// serialization round-trip.
+#[cfg(not(feature = "ethhash"))]
+#[test]
+fn test_range_proof_with_hashed_value() {
```
demosdemon, line 1756: Can you add an additional variant that tests a multi-level trie:

E.g.,

```text
  b"abc" => [0; 64],       // branch with value                                                               
  b"abc123" => [1; 64],    // leaf                                                                            
  b"abcdef" => [2; 64],    // branch with value                                                               
  b"abcdef123" => [3; 64], // leaf                                                                            
```

```text
  root
  └── "abc" (value: [0;64], Hash-encoded)     // branch with hashed value
      ├── "123" → leaf (value: [1;64])        // leaf with hashed value
      └── "def" (value: [2;64], Hash-encoded) // branch with hashed value
          └── "123" → leaf (value: [3;64])    // leaf with hashed value
```

and test with a range `["abc123"..)`. 

Or, some variation of this structure that contains multiple levels of branches that include hashed values tested against a proof that contains branches that won't include the values (e.g., `abc` above) and contains branches that will have both hash and value (e.g., `abcdef` above).

demosdemon (approved):

firewood/src/merkle/tests/range.rs
bernard-avalabs: I've added the test.

PR #1919, opened 2026-04-12
refactor(alloc): batch freelist writes per commit cycle and add free-block splitting

demosdemon:
## Why this should be merged

The previous implementation wrote freelist metadata to disk for each individual
allocation and deletion — one header write per freed node, and another per
`reap` call. Under workloads with many small nodes this caused redundant I/O
and held the header mutex for longer than necessary per commit cycle.

This refactor stages all freelist mutations in an in-memory `dirty_freelists`
buffer and flushes them in a single write at the end of each persist cycle,
regardless of how many nodes were freed. It also introduces free-block splitting
so that large free blocks can be subdivided to satisfy small allocations, reducing
unnecessary file growth.

## How this works

- `NodeAllocator` gains a `dirty_freelists: [Vec<LinearAddress>; NUM_AREA_SIZES]`
  field that accumulates blocks freed (or produced by splitting) during the current
  batch. Freed nodes are pushed onto this in-memory stack instead of immediately
  updating on-disk state.
- Allocation priority: (1) pop from `dirty_freelists`, (2) pop from the persisted
  freelist head in the header, (3) split a larger free block whose size is an exact
  multiple of the target (for targets below `AreaIndex::SPLITTABLE`), (4) extend the
  file.
- `NodeAllocator::finish()` writes all accumulated dirty freelist entries back to
  storage and flushes the header in one batched write.
- `PersistLoop` creates a single `NodeAllocator` at the top of each cycle, passes it
  through all pending reaps and the persist, then calls `finish()` once. Reap-only
  cycles (no revision to persist) also call `finish()` to flush any freelist changes
  and are counted by the new `REAP_ONLY_CYCLES` metric.
- `PersistWorker` now has a `Drop` impl that closes the channel and joins the
  background thread, ensuring deterministic cleanup of the `Arc<FileBacked>` reference
  before the worker is released.

## How this was tested

```bash
```

### Concurrent Benchmark

Concurrent benchmarks show file growth is slower and mgas/s is slightly higher. The average blocks/s with the change is higher as seen in the two grafana links.

**20m-30m**: https://avalabs.grafana.net/goto/efizdlm3vtm2oa?orgId=stacks-1371323

- 24321010812 - main (49b4187759ef789cc07afd0d55b37aad52742aac)
- 24321011304 - this pr (d1f60ab429abd051b37fcfb4207bcda566bab990)
- 24321011828 - with arena grow (c7d94011fc552e383063b1350bd7518720034ef9)

**1 to 30m**: https://avalabs.grafana.net/goto/dfizdntd1ucjka?orgId=stacks-1371323
 
- 24322573023 - main (49b4187759ef789cc07afd0d55b37aad52742aac)
- 24322573587 - with arena grow (c7d94011fc552e383063b1350bd7518720034ef9)

**NB**: file growth in the second link looks higher; however, the performance gain has caused the run with the change to process blocks faster than without and thus has progressed further than without. The first link show the database file size flat for a significant time while we drained the free lists.

## Breaking Changes

- [ ] firewood
- [x] firewood-storage — `WritableStorage::node_allocator()` is a new required trait
  method; `NodeAllocator::new()` is now `pub(crate)`; `NodeAllocator::allocate_node()`
  signature changed (takes `area_index` and `stored_area_size` separately)
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


storage/src/nodestore/primitives.rs
```diff
@@ -67,6 +67,10 @@ impl AreaIndex {
     /// The maximum area size available for allocation.
     pub const MAX_AREA_SIZE: u64 = AREA_SIZES[Self::NUM_AREA_SIZES - 1];
 
+    /// 768 and above are splittable, meaning they can be used to satisfy smaller
+    /// size requests by splitting them into multiple areas.
+    pub const SPLITTABLE: AreaIndex = AreaIndex(7);
```
demosdemon, line 72: I might consider making this configurable. I also noticed that we would benefit from sometimes splitting the 512 block; however, the 512 block also needs to split the 1024 if it can when we run out of 512 blocks.

storage/src/nodestore/alloc.rs
```diff
@@ -227,129 +254,322 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
...
+
+        Ok(next)
+    }
 
-            // Return the address of the newly allocated block.
-            trace!("Allocating from free list: addr: {address:?}, size: {index}");
-            return Ok(Some(address));
+    /// Attempts to satisfy an allocation for `target_index` by splitting a larger free block.
+    ///
+    /// Only operates on targets smaller than [`AreaIndex::SPLITTABLE`] (i.e. < 768 bytes);
```
rkuris, line 321: Why is this? Is it too expensive to split larger ones?

storage/src/nodestore/persist.rs
```diff
@@ -233,7 +231,8 @@ impl<S: WritableStorage> NodeStore<Committed, S> {
             // This isn't a guarantee that we won't exceed bump_size_limit
             // but it's a good enough approximation
             let might_overflow = allocated_len > bump_size_limit.saturating_sub(idx_size);
-            if might_overflow {
+            // let the arena grow if we still have dirty freelist entries
+            if might_overflow && !node_allocator.any_dirty_freelist() {
```
rkuris, line 235: I think this means that the bump arena might never get reset. This could make the size of the bump arena huge if we're doing a big tx.

Why doesn't this work correctly if `might_overflow` is set and there are dirty freelist entries?

storage/src/linear/mod.rs
```diff
@@ -251,6 +252,22 @@ pub trait WritableStorage: ReadableStorage {
...
+    /// Acquire a [`NodeAllocator`] bound to this storage and header.
+    ///
+    /// The allocator batches freelist writes and must be consumed by calling
+    /// [`NodeStore::persist`] (which calls [`NodeAllocator::finish`] internally)
+    /// or by calling [`NodeAllocator::finish`] directly after all
+    /// [`NodeStore::reap_deleted`] calls are complete.
+    ///
+    /// [`NodeStore::persist`]: crate::NodeStore::persist
+    /// [`NodeStore::reap_deleted`]: crate::NodeStore::reap_deleted
+    fn node_allocator<'a>(&'a self, header: &'a mut NodeStoreHeader) -> NodeAllocator<'a, Self>
```
rkuris, line 265: nit: Why is this part of the trait? Seems simpler if callers just call `NodeAllocator::new` themselves.

storage/src/nodestore/persist.rs
```diff
@@ -320,7 +325,9 @@ mod tests {
         header: &mut NodeStoreHeader,
     ) -> NodeStore<Committed, MemStore> {
         let ns = ns.as_committed();
-        ns.persist(header).unwrap();
+        let storage = ns.storage.clone();
```
rkuris, line 328: nit: Everywhere else I looked, we do `Arc::clone(&ns.storage)`, except here.

Technically no functional difference except the `Arc::clone` calls out the fact that this is super cheap.

rkuris (changes requested):
I'd like to better understand whether it's possible for this code to not to recycle the bumpalo buffer.

This PR adds some metrics too. We should either pull these out to a separate PR or clearly document them in the PR description. I didn't review them carefully.

demosdemon:
> I'd like to better understand whether it's possible for this code to not to recycle the bumpalo buffer.
> 
> This PR adds some metrics too. We should either pull these out to a separate PR or clearly document them in the PR description. I didn't review them carefully.

The added metrics are related to this PR. Why would I pull those out?

storage/src/nodestore/persist.rs
demosdemon: This is to prevent flushing free nodes that we would immediately pop. The metrics show this was significant in improving performance.

storage/src/linear/mod.rs
demosdemon: This is a convenient method for a fluid API.

storage/src/nodestore/persist.rs
demosdemon: We don't do it _everywhere_ but I can change this one.

storage/src/nodestore/alloc.rs
demosdemon: Yes. It is too expensive. Which is the reason splitting was removed in the first place.

rkuris:
This PR should either be reopened or an issue created. It was closed only because the author went on vacation.

PR #1922, opened 2026-04-13
test(proofs): add round-trip and FrozenChangeProof serialization tests

demosdemon:
## Why this should be merged

FrozenChangeProof serialization/deserialization had no test coverage, and there were no round-trip tests verifying that serialize(deserialize(bytes)) == bytes.

## How this works

Adds a create_valid_change_proof() helper that manually constructs a FrozenChangeProof covering all three BatchOp variants. Adds round-trip tests for both FrozenRangeProof and FrozenChangeProof. Mirrors the existing header/incomplete/ invalid-item parameterized tests for FrozenChangeProof, including an invalid BatchOp discriminant case.

## How this was tested

    cargo nextest run -p firewood --features ethhash,logger

## Breaking Changes

n/a

### DIFF
--- a/firewood/src/proofs/de.rs
+++ b/firewood/src/proofs/de.rs
@@ -168,8 +168,15 @@ impl Version0 for BatchOp<Key, Value> {
 
 impl Version0 for ProofNode {
     fn read_v0_item(reader: &mut V0Reader<'_>) -> Result<Self, ReadError> {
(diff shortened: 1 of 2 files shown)


rkuris (approved):

rkuris:
This is all getting replaced soon, so I'm not sure we need this, but since you already completed it we might as well merge it.

PR #1923, opened 2026-04-13
test(proofs): add edge case tests for valid proof structures

demosdemon:
## Why this should be merged

Several valid proof structures were never serialized in tests: nodes with children, full child bitmasks, value digests in hash mode, and individual BatchOp variants.

## How this works

Adds manually constructed ProofNodes targeting each structural edge case. The make_proof_node() helper uses PathComponent and Children from firewood_storage. Uses #[test_case] parameterization for BatchOp variants. The value-digest-hash test is gated on #[cfg(not(feature = "ethhash"))] since ValueDigest::Hash does not exist in that build. All tests verify round-trip byte equality.

## How this was tested

    cargo nextest run -p firewood --features ethhash,logger 
    cargo nextest run -p firewood --features logger  # ethhash=off path

## Breaking Changes

n/a

### DIFF
--- a/firewood/src/proofs/tests.rs
+++ b/firewood/src/proofs/tests.rs
@@ -6,6 +6,8 @@
 use integer_encoding::VarInt;
 use test_case::test_case;
 
+use firewood_storage::{Children, IntoHashType, PathComponent, TrieHash, ValueDigest};
+
 use super::{
     header::InvalidHeader,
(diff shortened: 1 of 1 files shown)


bernard-avalabs (approved):
Nice edge case serialization/deserialization tests!

PR #1924, opened 2026-04-13
test(proofs): add SeededRng fuzz tests for proof round-trips

demosdemon:
## Why this should be merged

Randomized testing exercises proof structures that hand-crafted tests miss, and ensures the parser never panics on malformed input. The seeded RNG allows deterministic reproduction of any failure.

## How this works

Adds helpers that generate random FrozenRangeProof and FrozenChangeProof structures using SeededRng. Three test_slow_ tests: round-trip fuzz for range proofs, round-trip fuzz for change proofs, and a malformed-data fuzz that randomly corrupts bytes and verifies the parser handles them gracefully. All iterations log the seed, proof structure, and hex bytes before asserting.

Corrupted proofs that accidentally parse as Ok are verified for idempotency (serialize(parse(serialize(parse(corrupted)))) == serialize(parse(corrupted))) rather than comparing against the corrupted bytes, avoiding false failures from non-minimal LEB128 varint encodings.

To reproduce a failure:

    FIREWOOD_TEST_SEED=<seed> cargo nextest run -p firewood \
        --features ethhash,logger --profile ci test_slow_range_proof_roundtrip_fuzz

## How this was tested

    cargo nextest run -p firewood --features ethhash,logger --profile ci proofs

## Breaking Changes

n/a

### DIFF
--- a/firewood/src/proofs/tests.rs
+++ b/firewood/src/proofs/tests.rs
@@ -6,7 +6,9 @@
 use integer_encoding::VarInt;
 use test_case::test_case;
 
-use firewood_storage::{Children, IntoHashType, PathComponent, TrieHash, ValueDigest};
+use firewood_storage::{
+    Children, IntoHashType, PathComponent, SeededRng, TrieHash, ValueDigest, logger::debug,
+};
 
 use super::{
     header::InvalidHeader,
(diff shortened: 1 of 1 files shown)


rkuris (approved):
nit: eprintln! should be logger.debug

PR #1926, opened 2026-04-13
fix(clippy): disallow panicking slice methods via clippy lint

rkuris:
## Why this should be merged

Add all stable panicking slice methods to `disallowed-methods` in clippy.toml so that every call site must carry an `#[expect]` with a safety comment explaining why the precondition is met.

Ideally all of these should be rewritten so the preconditions are not required to be checked and we use _checked variants of these methods when possible.

Disallowed methods by category:
- Out-of-bounds index: split_at, split_at_mut, swap, copy_within, rotate_left, rotate_right, select_nth_unstable{,_by,_by_key}
- Length mismatch: clone_from_slice, copy_from_slice, swap_with_slice
- Zero-size argument: windows, chunks, chunks_mut, chunks_exact, chunks_exact_mut, rchunks, rchunks_mut, rchunks_exact, rchunks_exact_mut

## How this works

Annotated all existing call sites with per-site `#[expect]` and safety comments.

## How this was tested

ci

## Breaking Changes

none


demosdemon (approved):
This wasn't as noisy as I expected.

Issue #1929, opened 2026-04-14
refactor: compute_right_edge_key can be simplified


bernard-avalabs:
The implementation of `comput_right_edge_key` in https://github.com/ava-labs/firewood/pull/1918, which was taken from the change proof audit branch, uses max_length as a heuristic for truncation.

This heuristic is unnecessary, and the function can be simplified. This refactor is deferred to facilitate the change proof verification merge.

bernard-avalabs:
Still an open issue that needs to be addressed. Change proof still relies on a `comput_right_edge_key` that uses `max_length` as a heuristic.

bernard-avalabs:
Addressed in #2188

PR #1931, opened 2026-04-14
fix: update change proof generator end proof key logic

bernard-avalabs:
## Why this should be merged                                                                                                                                                                                                                         

Continuation of the change proof verification merge. This change actually modifies the change proof generator. The change proof generator incorrectly determines the end proof key, creating a verification gap between last_proof_key and requested_end_key. An attacker could provide a valid end proof for last_proof_key while omitting changes in the (last_proof_key, requested_end_key] range — the hash verification would still pass because the "outside" hashes from the proof covered the gap.

## How this works  

  Previously, the generator only detected truncation when limit.get() <= batch_ops.len() && iter.next().is_some(). Now it checks iter.next().transpose()?.is_some() unconditionally to detect whether items remain, storing the result in hit_limit.
   
  End proof key selection:                                                                                                                                                                                                                          
  - Truncated (hit_limit): end proof is for last_batch_op_key (falling back to end_key) — the actual right edge of what was produced
  - Not truncated: end proof is for end_key (falling back to last_batch_op_key) — the full requested range                                                                                                                                          
                                                                                                                                   
                                                                                                                                                                                                                                                    
## How this was tested                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  - CI
                                                                                                                                                                                                                                                    
## Breaking Changes
                                                                                                                                                                                                                                                    
  - Proofs generated by the updated code may differ from those generated by the old code.

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -934,21 +934,22 @@ impl<T: HashedNodeReader> Merkle<T> {
             .take(limit.map_or(usize::MAX, NonZeroUsize::get))
             .collect::<Result<Box<_>, FileIoError>>()?;
 
-        let end_proof = if let Some(limit) = limit
-            && limit.get() <= batch_ops.len()
-            && iter.next().is_some()
-        {
-            // limit was provided, we hit it, and there is at least one more key
-            // end proof is for the last key provided
-            batch_ops.last().map(|largest_key| &**largest_key.key())
+        // Check whether the limit cut off remaining items.
+        let hit_limit = iter.next().transpose()?.is_some();
+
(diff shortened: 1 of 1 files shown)


rkuris:
Haven't started reviewing this yet. Is this related to https://github.com/ava-labs/firewood/issues/1929?

bernard-avalabs:
> Haven't started reviewing this yet. Is this related to #1929?

It is needed for the current change proof verification merge. It may need to be changed when we address #1929.

demosdemon:
This needs a test since it non-trivially changes the behavior of `change_proof` and declares breaking changes.

demosdemon (approved):
I see tests for this are in #1932

PR #1933, opened 2026-04-14
feat(dump): improve merkle dump display for ethhash RLP values

rkuris:
## Why this should be merged

Debugging incorrect storage roots inside account RLP values required being able to see decoded RLP fields in the merkle dump output. The existing dump code only displayed values as plaintext or hex, with no RLP awareness.

## How this works

Add RLP-aware value display via format_node_value() in firewood-storage. Values whose first byte is >= 0xc0 (RLP list prefix) are decoded and displayed as "rlp=[field0,field1,...]" with hex-encoded fields. Other values use the original plaintext-or-hex display.

The write_attributes! macro in merkle now delegates to this function, keeping RLP decoding in the storage crate where the rlp dependency already exists.

## How this was tested

Existing test suite passes. The Go TestDump test (which validates dump output contains plaintext values) continues to pass.

## Breaking Changes

None

### DIFF
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -441,7 +441,7 @@ pub fn format_node_value<W: std::io::Write + ?Sized>(
             let hex = hex::encode(item);
             write!(writer, "{hex:.12}")?;
             if hex.len() > 12 {
-                writer.write_all(b"..")?;
+                writer.write_all(b"...")?;
             }
         }
         write!(writer, "]")?;
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -407,3 +407,122 @@ impl From<CheckerError> for Vec<CheckerError> {
         vec![error]
     }
 }
+
+/// Write a human-readable representation of a node value to `writer`.
+///
+/// The caller should not pass an empty `value` — this function assumes at
+/// least one byte is present and will produce a spurious ` val=` prefix
+/// for empty input.
+///
+/// With ethhash enabled, values that look like non-empty RLP lists (first
+/// byte >= 0xc0) are decoded and displayed as ` rlp=[field0,field1,...]`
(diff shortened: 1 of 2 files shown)


demosdemon (approved):

storage/src/lib.rs
```diff
@@ -407,3 +407,122 @@ impl From<CheckerError> for Vec<CheckerError> {
...
+///
+/// # Errors
+///
+/// Returns an error if writing to `writer` fails.
+pub fn format_node_value<W: std::io::Write + ?Sized>(
+    value: &[u8],
+    writer: &mut W,
+) -> std::io::Result<()> {
+    #[cfg(feature = "ethhash")]
+    if value.first().is_some_and(|&b| b >= 0xc0)
```
RodrigoVillar, line 432: For the uninitiated, why `0xc0`?

storage/src/lib.rs
```diff
@@ -407,3 +407,122 @@ impl From<CheckerError> for Vec<CheckerError> {
...
+    {
+        write!(writer, " rlp=[")?;
+        for (i, item) in rlp_list.iter().enumerate() {
+            if i > 0 {
+                write!(writer, ",")?;
+            }
+            let hex = hex::encode(item);
+            write!(writer, "{hex:.12}")?;
+            if hex.len() > 12 {
+                writer.write_all(b"..")?;
```
RodrigoVillar:
```suggestion
                writer.write_all(b"...")?;
```

(nit) For consistency with the non-`ethhash` case.

RodrigoVillar (approved):

storage/src/lib.rs
rkuris: This is a shortcut that says "this is a list". https://ethereum.org/developers/docs/data-structures-and-encoding/rlp/ has more details.

PR #1934, opened 2026-04-15
test: add structural validation and partial proof tests for change proofs

bernard-avalabs:
## Why this should be merged                                                                                                                                                                                                                         
                  
Continuation of change proof validation merge. Stacks on https://github.com/ava-labs/firewood/pull/1932. Adds tests for structural validation (verify_change_proof_structure) and paginated/partial proof handling, covering error paths and the truncation logic.                                                                                 
                  
## How this works                                                                                                                                                                                                                                    
                  
  - Add tests/change/structural.rs — one test per error variant in verify_change_proof_structure: inverted range, unexpected start proof, unsorted/duplicate keys, start key > first key, end key < last key, max_length exceeded, missing boundary 
  proofs, unexpected end proof, wrong root hash, and DeleteRange rejection
  - Add tests/change/partial.rs — parameterized tests for limit/bound combinations (truncated, non-truncated, empty, bounded), a two-round paginated sync round trip, and truncation at a Delete boundary                                           
                                                                                                                                                                                                                                                    
## How this was tested
                                                                                                                                                                                                                                                                                                                                                                                                                   
  - 23 new tests (14 structural + 9 partial)

## Breaking Changes
                                                                                                                                                                                                                                                    
  - None


firewood/src/merkle/tests/change/partial.rs
```diff
@@ -0,0 +1,134 @@
...
+#[test_case(None,          Some(b"\x05"), Some(1),  0 ; "end before first change gives empty proof")]
+#[test_case(Some(b"\x20"), Some(b"\x40"), Some(1),  1 ; "bounded and limited")]
+#[test_case(Some(b"\x20"), Some(b"\x40"), None,     1 ; "bounded all fit")]
+fn test_partial_proof_verified(
+    start_key: Option<&[u8]>,
+    end_key: Option<&[u8]>,
+    limit: Option<usize>,
+    expected_ops: usize,
+) {
+    let (source, _dir_source) = setup_db![(b"\x10", b"v0"), (b"\x30", b"v1"), (b"\x50", b"v2")];
```
alarso16, line 33:
```suggestion
    let (source, _) = setup_db![(b"\x10", b"v0"), (b"\x30", b"v1"), (b"\x50", b"v2")];
```
Is this more conventional in Rust?

firewood/src/merkle/tests/change/partial.rs
```diff
@@ -0,0 +1,134 @@
...
+    let (target, _dir_target) = setup_db![(b"\x10", b"v0"), (b"\x30", b"v1"), (b"\x50", b"v2")];
+    let root1_target = target.root_hash().unwrap();
+    let (root1_source, root2) = setup_2nd_commit!(
```
alarso16, line 36: Unrelated to your change, but this macro has a weird name. You could use this multiple times, correct? Nothing "2nd" about it.

firewood/src/merkle/tests/change/partial.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Sync source→target using a partial proof of 1 key (limit=1), commit,
+/// then a proof of the remaining keys (2 passes).
+#[test]
+fn test_partial_proof_round_trip() {
+    let (source, _dir_source) = setup_db![(b"\x10", b"v0"), (b"\x20", b"v1"), (b"\x30", b"v2")];
+    let (target, _dir_target) = setup_db![(b"\x10", b"v0"), (b"\x20", b"v1"), (b"\x30", b"v2")];
+    let root1_target = target.root_hash().unwrap();
+
+    let (root1_source, root2) = setup_2nd_commit!(
```
alarso16, line 59:
```suggestio
    let (root1, root2) = setup_2nd_commit!(
```
These better not differ, right?

firewood/src/merkle/tests/change/structural.rs
```diff
@@ -0,0 +1,274 @@
...
+    assert!(matches!(
+        err,
+        api::Error::ProofError(crate::ProofError::UnexpectedStartProof)
+    ));
+}
+
+/// The `key <= prev.key()` check rejects both duplicate and reverse-ordered keys.
+#[test_case(b"\x50", b"\x50" ; "duplicate keys rejected")]
+#[test_case(b"\xa0", b"\x50" ; "reverse order rejected")]
+fn test_crafted_keys(key_a: &[u8], key_b: &[u8]) {
```
alarso16, line 56: This name is a little weird. Maybe `test_keys_out_of_order`?

firewood/src/merkle/tests/change/structural.rs
```diff
@@ -0,0 +1,274 @@
...
+    assert!(matches!(
+        err,
+        api::Error::ProofError(crate::ProofError::ProofIsLargerThanMaxLength)
+    ));
+}
+
+/// Non-empty `batch_ops` with `requested_start_key` and `requested_end_key`
+/// but both boundary proofs empty.
+#[test]
+fn test_crafted_missing_boundary_proof() {
```
alarso16, line 136: The `crafted` word again doesn't make sense to me. I assume you mean here that the proof is explicitly created in an invalid way

firewood/src/merkle/tests/change/structural.rs
```diff
@@ -0,0 +1,274 @@
...
+        err,
+        api::Error::ProofError(crate::ProofError::UnexpectedEndProof)
+    ));
+}
+
+/// Boundary proof hash chain was built for `root2` but verified against
+/// `HashKey::empty()` — the root hash in the proof nodes won't match,
+/// producing `UnexpectedHash`.
+///
+/// TODO: `ProofError::UnexpectedHash` is poorly named. The hash isn't unexpected, it's just wrong.
```
alarso16, line 213: Is this a good place for the TODO? Why not just rename it now?

alarso16 (commented):
One potentially missing edge case in most of the key-ordering tets is ensuring that keys with different lengths are handled correctly. Like what about the zero-length key? It's valid I think, but I don't know what the behavior should be.

firewood/src/merkle/tests/change/partial.rs
bernard-avalabs: My understanding is that _ drops the TempDir immediately, which would delete the temp directory and break the database.

firewood/src/merkle/tests/change/partial.rs
bernard-avalabs: That's a good point. setup_2nd_commit was added in a previous PR. I will change the name there.

firewood/src/merkle/tests/change/partial.rs
bernard-avalabs: I added an assert that they are equal.

firewood/src/merkle/tests/change/structural.rs
bernard-avalabs: Good call. I've changed the name to `test_crafted_out_of_order_or_dup_keys`

firewood/src/merkle/tests/change/structural.rs
bernard-avalabs: I agree that the word `crafted` here is a bit strange. But it is also fairly common terminology in security testing with the meaning that the data/packet/proof was explicitly created in an invalid way. I suggest we keep it, but please let me know if you think otherwise.

firewood/src/merkle/tests/change/structural.rs
bernard-avalabs: Good catch. I think `UnexpectedHash` is actually fine here, since we received an unexpected hash value. I have removed the TODO.

demosdemon (approved):

PR #1935, opened 2026-04-15
fix(proofs): 1/2 recompute storageRoot in ethhash proof generation

rkuris:
## Why this should be merged

Coreth inserts zeroed placeholder hashes for account storageRoot fields and expects firewood to compute the correct values. Currently, the computed hash is only used ephemerally during trie hashing but is never persisted in the stored node value. This means proofs generated from these nodes contain stale storageRoot values that fail verification.

## How this works

This fix recomputes the storageRoot from the node's children at proof generation time, in both ProofNode construction and the key-value iterator used by range proofs. A future change will persist the correct storageRoot during hashing so this recomputation becomes a no-op for new databases, but this fix will still be needed for compatibility with existing databases.

Also document test logging gotchas (RUST_LOG level, --nocapture) near the ctor::ctor init_logger in lib.rs.

## How this was tested

Add must_recompute_storage_hash() to the NodeStore/TrieReader layer so callers can query whether recomputation is needed based on database version. Currently returns true for all versions; a future database version bump will allow it to return false for databases that already store correct values.

Fix all proof and range proof tests to use values from the key-value iterator (which applies the same fix) rather than raw input values.

## Breaking Changes

None

### DIFF
--- a/firewood/src/iter.rs
+++ b/firewood/src/iter.rs
@@ -319,8 +319,7 @@ impl<T: TrieReader> Iterator for MerkleKeyValueIter<'_, T> {
                                 // no value, continue to next node
                                 return None;
                             };
-                            let must_recompute =
-                                self.iter.merkle.must_recompute_storage_hash();
+                            let must_recompute = self.iter.merkle.must_recompute_storage_hash();
                             let child_hashes = if must_recompute && key.len() == 32 {
                                 branch.children_hashes()
                             } else {
--- a/firewood/src/iter.rs
+++ b/firewood/src/iter.rs
@@ -319,12 +319,14 @@ impl<T: TrieReader> Iterator for MerkleKeyValueIter<'_, T> {
                                 // no value, continue to next node
                                 return None;
                             };
-                            fix_account_value(
-                                key,
-                                value,
-                                &branch.children_hashes(),
-                                self.iter.merkle.must_recompute_storage_hash(),
-                            )
+                            let must_recompute =
+                                self.iter.merkle.must_recompute_storage_hash();
+                            let child_hashes = if must_recompute && key.len() == 32 {
+                                branch.children_hashes()
+                            } else {
+                                firewood_storage::Children::new()
+                            };
+                            fix_account_value(key, value, &child_hashes, must_recompute)
                         }
                         Node::Leaf(leaf) => fix_account_value(
                             key,
--- a/firewood/src/iter.rs
+++ b/firewood/src/iter.rs
@@ -316,13 +316,22 @@ impl<T: TrieReader> Iterator for MerkleKeyValueIter<'_, T> {
                     match &*node {
                         Node::Branch(branch) => {
                             let Some(value) = branch.value.as_ref() else {
-                                // This node doesn't have a value to return.
-                                // Continue to the next node.
+                                // no value, continue to next node
                                 return None;
                             };
-                            Some((key, value.clone()))
+                            fix_account_value(
+                                key,
+                                value,
+                                &branch.children_hashes(),
+                                self.iter.merkle.must_recompute_storage_hash(),
+                            )
                         }
-                        Node::Leaf(leaf) => Some((key, leaf.value.clone())),
+                        Node::Leaf(leaf) => fix_account_value(
+                            key,
+                            &leaf.value,
+                            &firewood_storage::Children::new(),
+                            self.iter.merkle.must_recompute_storage_hash(),
+                        ),
                     }
                 })
                 .transpose()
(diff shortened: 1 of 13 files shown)


storage/src/nodestore/hash.rs
```diff
@@ -271,3 +271,58 @@ where
...
+        crate::TrieHash::from(Keccak256::digest(rlp::NULL_RLP))
+    } else {
+        let mut child_hashes = child_hashes.clone();
+        if let Some((_, child)) = child_hashes.take_only_child() {
+            match child {
+                HashType::Hash(hash) => hash,
+                HashType::Rlp(rlp_bytes) => {
+                    // Single storage child with small RLP — hash the raw bytes.
+                    crate::TrieHash::from(Keccak256::digest(&*rlp_bytes))
+                }
```
demosdemon: this appears incorrect and doesn't match what was previously hashed:

https://github.com/ava-labs/firewood/blob/1edde17f106d3ab356942a8913063bb310e03bf0/storage/src/hashers/ethhash.rs#L187-L197

firewood/src/proofs/types.rs
```diff
@@ -333,18 +333,58 @@ impl From<PathIterItem> for ProofNode {
...
+            .map(|value| ValueDigest::Value(Box::from(value)));
+
+        // For account-depth nodes on databases that need storageRoot
+        // recomputation, fix the value from the node's children.
+        #[cfg(feature = "ethhash")]
+        let value_digest = {
+            debug_assert!(
+                item.must_recompute_storage_hash,
+                "must_recompute_storage_hash should always be true"
+            );
```
demosdemon: this assertion feels unnecessary since the intent is to make this variable.

firewood/src/merkle/tests/ethhash.rs
```diff
@@ -118,6 +118,168 @@ fn make_key(hex_str: &str) -> Key {
...
+            (&storage_key_b, &storage_value_b),
+        ],
+    );
+
+    assert_ne!(
+        storage_root,
+        empty_trie_root().to_vec(),
+        "should not be empty trie root"
+    );
+}
```
demosdemon: These are not relevant to this PR and currently fail.

demosdemon (changes requested):

storage/src/nodestore/hash.rs
rkuris: > this appears incorrect and doesn't match what was previously hashed

I think the old code is incorrect, but it actually doesn't ever hit this case anyway. The problem here is that rlp_bytes is already prefixed with the appropriate nibble. The old code then incorrectly wrapped this in another extension node encoding.

It never hits this because we're always above a depth of 32 bytes (best case 33 bytes) and so we never use a HashType::Rlp here. I changed this to an unreachable block in a019385179ca8a37a4894d98c3305bf6ba62fd7f. It doesn't get hit in fuzz testing and I'll run another full bootstrap to verify it never hits this case.

firewood/src/proofs/types.rs
rkuris: Fixed in 681e7b2fb8d8b81ff65067f141da3aa3e9e0f8e9

firewood/src/merkle/tests/ethhash.rs
rkuris: Sorry about that, this was a copy/paste error just before committing and has been fixed.

firewood/src/proofs/types.rs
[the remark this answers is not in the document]
rkuris: This was fixed in 681e7b2fb8d8b81ff65067f141da3aa3e9e0f8e9

firewood/src/iter.rs
```diff
@@ -316,13 +316,22 @@ impl<T: TrieReader> Iterator for MerkleKeyValueIter<'_, T> {
...
+                                // no value, continue to next node
                                 return None;
                             };
-                            Some((key, value.clone()))
+                            fix_account_value(
+                                key,
+                                value,
+                                &branch.children_hashes(),
+                                self.iter.merkle.must_recompute_storage_hash(),
+                            )
```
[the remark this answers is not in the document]
rkuris: Fixed in 681e7b2fb8d8b81ff65067f141da3aa3e9e0f8e9

firewood/src/merkle/tests/proof.rs
```diff
@@ -125,10 +125,14 @@ fn test_proof() {
...
-    for (key, val) in items {
+    for (key, _val) in items {
         let proof = merkle.prove(key).unwrap();
         assert!(!proof.is_empty());
-        proof.verify(key, Some(val), &root_hash).unwrap();
+        // Verify as inclusion proof. We don't check the exact value because
+        // proof generation may rewrite account storageRoot fields, producing
+        // a different value than get_value() returns.
+        let digest = proof.value_digest(key, &root_hash).unwrap();
+        assert!(digest.is_some(), "should be inclusion proof for {key:?}");
```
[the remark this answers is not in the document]
rkuris: > test_proof no longer exercises Proof::verify at all

This is way overstated but putting this test back in isn't all that hard anyway. Fixed in 8c69952770ddd010372e8f66c3f4882d47a9c577

demosdemon (approved):

PR #1938, opened 2026-04-16
fix(storage): 2/2 persist computed storageRoot in account node values

### LINKED ISSUE #1266: bugish: firewood nor coreth write the account `storageRoot` to storage
Firewood does not rely on the ethereum `storageRoot` value when calculating the hash of the trie but instead replaces it with the calculated root for the subtrie beginning at the account node. However, firewood does not write this value to storage nor exposes this value to coreth; therefore; coreth does not write this value back to storage.

Why is this a problem?

1. Range Proofs will be unable verify the hash of the account node without rehashing the entire sub-trie and verifying the hash with its parent.

rkuris:
## Why this should be merged

When hashing in ethhash mode, the storageRoot field in Ethereum account nodes was only replaced ephemerally during hash computation but never written back to the node's stored value. This caused invalid eth_getProof responses and a potential DoS vector for proof verification.

## How this works

During the hashing phase in hash_helper_inner, compute the storage sub-trie root hash and update the account node's RLP-encoded value before the node is persisted to disk. The hash computation itself is unaffected since Preimage::write() idempotently re-computes the same replacement. The update logic is extracted into a standalone update_account_storage_root function for clarity.

For proofs generated from older databases (before firewood-v1-hfix), dynamically compute the correct storageRoot when building ProofNodes so that proofs are verifiable by downstream clients.

Also:
- Relax replace_hash to accept 3+ item RLP lists (coreth appends a trailing empty byte, producing 5-item account values)
- Bump database header version to firewood-v1-hfix so databases created with the fix can be distinguished from unfixed ones
- Export fix_account_storage_root_value for use by proof generation

## How this was tested

- Add test_autocompute_hash covering both 4-item and 5-item account RLP
- Add test_persisted_storage_root verifying one/two storage entries
- Add test_range_proof_fixes_legacy_zeroed_storage_root simulating a pre-fix database with clobbered storageRoot bytes

## Breaking Changes

None


firewood/src/merkle/tests/range.rs
```diff
@@ -25,28 +25,23 @@ fn proof_node(nibbles: &[u8]) -> ProofNode {
...
-                .next()
-                .expect("iterator should yield at least one item")
-                .expect("iterator should not error");
-            assert_eq!(found_key.as_ref(), key, "iterator returned wrong key");
-            (Box::from(key), val)
+            let val = merkle
+                .get_value(key)
+                .unwrap()
+                .unwrap_or_else(|| v.as_ref().to_vec().into_boxed_slice());
+            (key.to_vec().into_boxed_slice(), val)
```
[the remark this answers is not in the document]
rkuris: Fixed in bde746c1525bf55b1cb41b56dcc9eb0a9a6ae903

storage/src/nodestore/hash.rs
```diff
@@ -325,5 +337,82 @@ pub fn fix_account_storage_root_value(
...
+    use crate::node::BranchNode;
+    use rlp::RlpStream;
+    use sha3::{Digest, Keccak256};
+
+    // Both lengths are usize counts of nibbles in a trie path, so their
+    // sum cannot overflow on any platform firewood targets.
+    let total_depth = path_prefix
+        .0
+        .len()
+        .wrapping_add(node.partial_path().0.len());
```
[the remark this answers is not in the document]
rkuris: The two lengths being added are both usize counts of nibbles from trie nodes we constructed ourselves in memory, not from untrusted input. Overflow is physically impossible on 64-bit platforms (you'd need 2^64 bytes of path data), and even on 32-bit it would require exhausting memory first. The wrapping_add is just there to silence clippy's arithmetic_side_effects lint; it's not guarding against anything real.

We can't even construct paths longer than about 100k due to the recursive nature of node insertion.

firewood/src/proofs/types.rs
```diff
@@ -336,16 +336,13 @@ impl From<PathIterItem> for ProofNode {
...
+        // For account-depth nodes, ensure the storageRoot in the value reflects the
+        // actual storage sub-trie. Older databases may have a stale or zeroed
+        // storageRoot that was only replaced ephemerally during hashing.
         #[cfg(feature = "ethhash")]
-        let value_digest = if item.must_recompute_storage_hash {
-            fix_account_storage_root(value_digest, &item.key_nibbles, &child_hashes)
-        } else {
-            value_digest
-        };
+        let value_digest = fix_account_storage_root(value_digest, &item.key_nibbles, &child_hashes);
```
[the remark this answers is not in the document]
rkuris: Fixed in latest push

storage/src/nodestore/hash.rs
```diff
@@ -325,5 +337,82 @@ pub fn fix_account_storage_root_value(
...
+                            rlp.append_raw(rlp_bytes, 1);
+                        }
+                        None => {
+                            rlp.append_empty_data();
+                        }
+                    }
+                }
+                rlp.append_empty_data();
+                crate::TrieHash::from(Keccak256::digest(rlp.out()))
+            };
```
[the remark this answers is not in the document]
rkuris: Fixed in latest push

storage/src/nodestore/hash.rs
demosdemon: The copilot's point, we should be consistent with whether we use `saturating_add` or `wrapping_add` for this check. If we use `wrapping_add` for `update_account_storage_root`, then I suggest updating it here as well.

storage/src/nodestore/header.rs
demosdemon:
```suggestion
        if self.version.is_firewood_v1() {
```

This was an mistake on my part when I first wrote this. I missed replacing this with the function but now it's necessary.

firewood/src/proofs/types.rs
```diff
@@ -341,10 +341,12 @@ impl From<PathIterItem> for ProofNode {
         let value_digest = item
             .node
             .value()
-            .map(|value| ValueDigest::Value(Box::from(value)));
+            .map(|value| ValueDigest::Value(value.to_vec().into_boxed_slice()));
```
demosdemon:
```suggestion
            .map(|value| ValueDigest::Value(Box::from(value)));
```

`Box::from` is more idiomatic and is guaranteed to alloc only once whereas `to_vec().into_boxed_slice()` _might_ alloc twice. Changing this is unnecessary.

demosdemon (changes requested):
there are multiple places in `storage/src/nodestore/mod.rs` where `must_recompute_storage_hash` is hardcoded as true. These need some way of being replaced with the new flag. For this I'm specifically referring to `new_empty_committed` which affects in-memory databases.

firewood/src/proofs/types.rs
rkuris: Fixed in 4761fc75dad6. Good advice. `Box::from` should always be preferred over this chain. Perhaps we should add this to the github copilot review instructions.

storage/src/nodestore/header.rs
rkuris: Fixed in 4761fc75dad6. Nice catch. The inline equality also drifted from what `firewood_version_str()` actually populates, so using `is_firewood_v1()` ties the log line to the same gate as the data it prints.

storage/src/nodestore/hash.rs
rkuris: Fixed in 4761fc75dad6. Switched the depth check to `wrapping_add` so both account-depth sites use the same strategy.

rkuris:
Fixed in 4761fc75dad6. Nice catch on the in-memory case. `new_empty_committed`, `new_empty_proposal`, and `new_empty_recon` now derive `must_recompute_storage_hash` from `Version::new()`, so freshly created stores correctly say "no recomputation needed" since they're written by current code. `with_root` reads the actual on-disk header so revisions reconstructed from historical roots stay safe when the persisted data is pre-hfix.

firewood/src/merkle/tests/ethhash.rs
```diff
@@ -219,6 +219,132 @@ fn rlp_encode_storage(value: &[u8; 32]) -> Vec<u8> {
...
+        &account_value,
+        &dummy_storage_root,
+        &[(&storage_key, &storage_value)],
+    );
+
+    assert_ne!(
+        storage_root,
+        empty_trie_root().to_vec(),
+        "should not be empty trie root"
+    );
```
demosdemon: this is a weak assertion. a better assertion would be `assert_eq!(storage_root, expected_value)`. `test_range_proof_accounts_have_computed_storage_root` only indirectly verifies the value is correct but a more targeted test would be better.

firewood/src/merkle/tests/ethhash.rs
```diff
@@ -268,9 +395,22 @@ fn test_range_proof_accounts_have_computed_storage_root() {
...
+        .map(|(k, _)| {
+            let val = merkle
+                .get_value(k.as_ref())
+                .unwrap()
+                .expect("account key should exist");
+            (k.to_vec().into_boxed_slice(), val)
+        })
+        .collect();
+
+    let range_proof = RangeProof::new(start_proof, end_proof, key_values);
```
demosdemon: I'm curious why it was necessary to manually construct the proof instead of using `Merkle::range_proof`? Makes me concerned about test drift.

storage/src/nodestore/header.rs
```diff
@@ -106,16 +106,19 @@ impl Version {
     }
 
     const fn is_firewood_v1(self) -> bool {
```
demosdemon: claude made a good suggestion that we should rename this to something like `has_extended_header_fields`.

demosdemon (approved):

firewood/src/merkle/tests/ethhash.rs
rkuris: Fixed in 70feca32. Good point, there was another case of this I also fixed — strengthened both `test_persisted_storage_root_one_storage_entry` and `_two_storage_entries` to `assert_eq!` against the root hash of a standalone storage trie built via `init_merkle`.

storage/src/nodestore/header.rs
rkuris: Fixed in 70feca32. Great find — `has_extended_header_fields` is a much clearer name; it's exactly what the boolean gates (the extended fields like `cargo_version`, `git_describe`, `root_hash` that only exist on v1-format databases).

firewood/src/merkle/tests/ethhash.rs
rkuris: Fixed in d2584c9d. I have no idea either why it's manually built — refactored to use `Merkle::range_proof`. The storage entry sorts after `right_key` so the iterator naturally excludes it.

PR #1946, opened 2026-04-22
test: add edge case tests for change proof verification

bernard-avalabs:
## Why this should be merged                                                                                                                                                                                                                         
                  
Continuation of the change proof verification merge. Adds 21 edge case integration tests for the change proof verification pipeline, covering boundary conditions found during code review and TLA+ model checking.                                                                                    
                  
## How this works                                                                                                                                                                                                                                    
                  
  Add tests/change/edge_cases.rs with tests covering:                                                                                                                                                                                               
   
  - Empty batch ops with non-empty boundary proofs (out-of-range change)                                                                                                                                                                            
  - Odd nibble depth proof nodes (structural validation)
  - Inclusion proofs with children below — prefix keys where the boundary key is a branch with value and children (start and end sides)                                                                                                             
  - Key exhaustion — start key exhausted at parent branch, divergent node                                                                                                                                                                           
  - Exclusion proofs for deleted keys (start and end sides)                                                                                                                                                                                         
  - Disjoint proof — query range excludes all changes, out-of-range value skipping                                                                                                                                                                  
  - Divergent child gap — exclusion proof with sibling under shared branch                                                                                                                                                                          
  - Terminal divergence within partial path (compute_outside_children Case A)                                                                                                                                                                       
  - Terminal ancestor with all children outside (compute_outside_children Case B)                                                                                                                                                                   
  - Terminal divergent node with all children in range                                                                                                                                                                                              
  - Right edge boundary as prefix of terminal — end key is prefix of existing key                                                                                                                                                                   
  - Out-of-range value change adopted — reconcile callback adopts proof value for out-of-range ancestor                                                                                                                                             
  - Prefix key deleted in end_root — out-of-range deletion, callback clears stale value (regression)                                                                                                                                                
  - Single-point range — start_key == end_key                                                                                                                                                                                                       
  - Root compression — out-of-range deletion compresses root partial path (found by TLA+)                                                                                                                                                           
  - Root expansion — out-of-range insertion expands root partial path                                                                                                                                                                               
  - Root value change — empty key value delta
  - ValueDigest::Hash at out-of-range prefix key — merkledb-mode test verifying the Hash fix in reconcile_branch_proof_node and compute_root_hash_with_proofs (serialized/deserialized proof)                                                       
                                                                                                                                                                                                                                                    
## How this was tested                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  - CI
                                                                                                                                                                                                                                                    
## Breaking Changes
                                                                                                                                                                                                                                                    
  - None


firewood/src/merkle/tests/change/edge_cases.rs
```diff
@@ -0,0 +1,793 @@
...
+    let (target, _dir_target) = setup_db![(b"\x12", b"v0"), (b"\x13", b"v1"), (b"\x50", b"v2")];
+    let root1_target = target.root_hash().unwrap();
+
+    let (root1_source, root2) = setup_2nd_commit!(source, [(b"\x50", b"changed")]);
+
+    let proof = source
+        .change_proof(root1_source, root2.clone(), Some(b"\x14"), None, None)
+        .unwrap();
+    assert_eq!(proof.batch_ops().len(), 1);
+    // \x14 doesn't exist — start proof is exclusion at an odd nibble depth.
```
rkuris: Let's assert this rather than commenting.

firewood/src/merkle/tests/change/edge_cases.rs
```diff
@@ -0,0 +1,793 @@
...
+
+//! Edge cases: odd depths, divergence, exclusion, deletion, children.
+
+use super::*;
+
+#[test]
+fn test_empty_batch_ops_with_nonempty_proofs() {
+    let (source, _dir_source) = setup_db![(b"\x10", b"v0"), (b"\x20", b"v1"), (b"\x30", b"v2")];
+    let (target, _dir_target) = setup_db![(b"\x10", b"v0"), (b"\x20", b"v1"), (b"\x30", b"v2")];
+    let root1_target = target.root_hash().unwrap();
```
rkuris: nit: this identical code (lines 10-12) could use it's own macro, I spotted 4-5 tests that do exactly the same thing but there are probably more.

You don't have to fix this but it might make this easier to maintain (with usual caveats about not having too many helpers).

firewood/src/merkle/tests/change/edge_cases.rs
```diff
@@ -0,0 +1,793 @@
...
+    let proof = db
+        .change_proof(
+            root1.clone(),
+            root2.clone(),
+            Some(b"\x10".as_slice()),
+            None,
+            None,
+        )
+        .unwrap();
+    assert!(!proof.batch_ops().is_empty());
```
rkuris: This is true but could be stronger:
```suggestion
    assert_eq!(proof.batch_ops().len(), 1);
```

firewood/src/merkle/tests/change/edge_cases.rs
```diff
@@ -0,0 +1,793 @@
...
+        verify_change_proof_structure(&proof, root2, Some(b"\x20"), Some(b"\x20"), None).unwrap();
+    verify_and_check(&target, &proof, &ctx, root1_target).unwrap();
+}
+
+/// Out-of-range deletion compresses the root's `partial_path`. The proving
+/// trie's root (forked from the proposal) retains the old shape, and
+/// `collapse_root_to_path` must reshape it to match `end_root`.
+///
+/// Root1: keys at nibbles 0 and 1. Root at `[]`.
+/// Root2: `\x01` deleted → root compresses to `partial_path` `[1]`.
```
rkuris, line 607: This should be asserted in the test somewhere

rkuris (commented):
This could use a second pass. I think there are more cases where the assertions can be cleaned up more.

firewood/src/merkle/tests/change/edge_cases.rs
bernard-avalabs: Replaced the comment with an assertion.

firewood/src/merkle/tests/change/edge_cases.rs
bernard-avalabs: Added setup_source_target! macro and converted 14 call sites.

firewood/src/merkle/tests/change/edge_cases.rs
bernard-avalabs: Fixed

firewood/src/merkle/tests/change/edge_cases.rs
bernard-avalabs: Added assertion that the first proof node's key is [1] after the deletion.

rkuris (approved):

PR #1947, opened 2026-04-23
test: add adversarial attack tests for change proof verification

bernard-avalabs:
## Why this should be merged                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                    
Continuation of the change proof verification merge. Adds 18 adversarial tests that verify change proof verification correctly rejects tampered, forged, and malicious proofs. These tests exercise the security boundaries of the verification pipeline.                                              
                  
## How this works                                                                                                                                                                                                                                    
                  
  Add tests/change/attack.rs with tests covering:                                                                                                                                                                                                   
   
  - Omitted change — removing a batch op from a valid proof                                                                                                                                                                                         
  - Forged value — replacing a batch op's value with a different one
  - Spurious batch op — injecting an extra Put for a non-existent key                                                                                                                                                                               
  - Boundary mismatch (4 tests) — spurious Put/Delete at start/end boundaries where the op type contradicts the inclusion/exclusion proof (StartProofOperationMismatch, EndProofOperationMismatch)
  - Tampered start key value — changing the value at start_key in batch_ops                                                                                                                                                                         
  - Conflicting proof nodes — flipping a bit in the end proof's root node so start and end proofs disagree                                                                                                                                          
  - Value at odd nibble length — injecting a value at an odd-depth proof node                                                                                                                                                                       
  - Divergence at depth zero — crafting start/end proofs from disjoint bounded proofs with shared root removed                                                                                                                                      
  - Spurious delete in range — deleting an unchanged key within the proven range                                                                                                                                                                    
  - Spurious delete same subtree — delete target shares a first nibble with a changed key                                                                                                                                                           
  - Spurious delete with existing boundaries — same attack with inclusion boundary proofs                                                                                                                                                           
  - Swapped value in range — replacing a legitimate Put's value                                                                                                                                                                                     
  - Spurious put in range — injecting a Put for a non-existent in-range key                                                                                                                                                                         
  - Collapse root hides spurious key (TLA+ model) — injecting a key at a different first nibble that collapse would strip; range-safety check prevents this                                                                                         
  - Stripped divergent child (new) — removing a divergent sibling's hash from a proof node to break hash computation                                                                                                                                
                                                                                                                                                                                                                                                    
## How this was tested                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                    
  - CI
                                                                                                                                                                                                                                                    ## Breaking Changes                                                                                                                                                                                                                                  
   
  - None

### DIFF
--- a/firewood/src/merkle/tests/change/attack.rs
+++ b/firewood/src/merkle/tests/change/attack.rs
@@ -479,20 +479,13 @@ fn test_crafted_value_at_odd_nibble_length_rejected() {
 /// should detect the inconsistency.
 #[test]
 fn test_crafted_divergence_at_depth_zero() {
-    let (source, _dir_source) = setup_db![
+    let (source, target, root1_target, _ds, _dt) = setup_source_target![
         (b"\x10", b"v0"),
         (b"\x11", b"v1"),
         (b"\xa0", b"v2"),
         (b"\xa1", b"v3")
     ];
-    let root1_source = source.root_hash().unwrap();
-    let (target, _dir_target) = setup_db![
-        (b"\x10", b"v0"),
-        (b"\x11", b"v1"),
-        (b"\xa0", b"v2"),
-        (b"\xa1", b"v3")
-    ];
-    let root1_target = target.root_hash().unwrap();
+    let root1_source = root1_target.clone();
 
     let changes: Vec<BatchOp<&[u8], &[u8]>> = vec![
         BatchOp::Put {
@@ -571,20 +564,13 @@ fn test_crafted_divergence_at_depth_zero() {
 /// the proof node at [1, 0, 5] should cause rejection.
 #[test]
 fn test_crafted_stripped_divergent_child_rejected() {
-    let (source, _dir_source) = setup_db![
+    let (source, target, root1_target, _ds, _dt) = setup_source_target![
         (b"\x10\x50", b"a"),
         (b"\x10\x58", b"b"),
         (b"\x10\x5f", b"d"),
         (b"\x30\x00", b"c")
     ];
-    let root1_source = source.root_hash().unwrap();
-    let (target, _dir_target) = setup_db![
-        (b"\x10\x50", b"a"),
-        (b"\x10\x58", b"b"),
-        (b"\x10\x5f", b"d"),
-        (b"\x30\x00", b"c")
-    ];
-    let root1_target = target.root_hash().unwrap();
+    let root1_source = root1_target.clone();
 
     source
         .propose(vec![
--- a/firewood/src/merkle/tests/change/attack.rs
+++ b/firewood/src/merkle/tests/change/attack.rs
@@ -5,21 +5,32 @@
 
 use super::*;
 
+type OwnedBatchOps = Vec<BatchOp<Box<[u8]>, Box<[u8]>>>;
+
+/// Helper: returns true if the (possibly corrupted) proof is rejected by
+/// either the structural check or the root hash check.
+fn is_rejected(
+    db: &Db,
+    proof: &FrozenChangeProof,
+    end_root: api::HashKey,
+    start_key: Option<&[u8]>,
+    end_key: Option<&[u8]>,
+    start_root: api::HashKey,
+) -> bool {
+    match verify_change_proof_structure(proof, end_root, start_key, end_key, None) {
+        Err(_) => true,
+        Ok(ctx) => verify_and_check(db, proof, &ctx, start_root).is_err(),
+    }
+}
+
 #[test]
 fn test_crafted_omitted_change_detected() {
-    let (source, _dir_source) = setup_db![
+    let (source, target, root1_target, _ds, _dt) = setup_source_target![
         (b"\x10", b"v0"),
         (b"\x20", b"v1"),
         (b"\x30", b"v2"),
         (b"\x40", b"v3")
     ];
-    let (target, _dir_target) = setup_db![
-        (b"\x10", b"v0"),
-        (b"\x20", b"v1"),
-        (b"\x30", b"v2"),
-        (b"\x40", b"v3")
-    ];
-    let root1_target = target.root_hash().unwrap();
 
     let (root1_source, root2) =
         setup_2nd_commit!(source, [(b"\x20", b"changed1"), (b"\x30", b"changed2")]);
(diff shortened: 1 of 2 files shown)


firewood/src/merkle/tests/change/attack.rs
```diff
@@ -0,0 +1,1023 @@
...
+        crate::Proof::new(valid.end_proof().as_ref().into()),
+        ops.into_boxed_slice(),
+    );
+
+    // Structural check or root hash check catches this
+    let result = verify_change_proof_structure(&crafted, root2, None, None, None);
+    if let Ok(ctx) = result {
+        verify_and_check(&target, &crafted, &ctx, root1_target)
+            .expect_err("spurious batch op must be detected");
+    }
```
rkuris: We should fail if the `verify_change_proof_structure` fails, not just skip this additional step.

This problem seems repeated in several other tests (test_crafted_spurious_put_at_start_key_boundary for example)

```suggestion
      let ctx = result.expect("structural verification should succeed");
      verify_and_check(&target, &crafted, &ctx, root1_target)
            .expect_err("spurious batch op must be detected");
```

firewood/src/merkle/tests/change/attack.rs
```diff
@@ -0,0 +1,1023 @@
...
+/// branch. The verifier must reject the proof because the branch hash
+/// can't be computed correctly without the sibling's contribution.                                                                                                                                                                               
+///             
+/// Keys \x10\x50, \x10\x58, and \x10\x5f share a branch at [1, 0, 5].
+/// Deleting \x10\x50 makes the start proof an exclusion proof with
+/// divergent siblings at nibbles 8 and f. The branch survives deletion
+/// because two children remain. Stripping nibble 8's child hash from
+/// the proof node at [1, 0, 5] should cause rejection.                                                                                                                                                                           
+#[test]
+fn test_crafted_stripped_divergent_child_rejected() {
```
rkuris: This test is in the wrong place. See line 558 -- this should be before that.

firewood/src/merkle/tests/change/attack.rs
```diff
@@ -0,0 +1,1023 @@
...
+    let crafted = FrozenChangeProof::new(
+        crate::Proof::new(valid.start_proof().as_ref().into()),
+        crate::Proof::new(valid.end_proof().as_ref().into()),
+        shortened.into_boxed_slice(),
+    );
+
+    let ctx =
+        verify_change_proof_structure(&crafted, root2, Some(b"\x10"), Some(b"\x40"), None).unwrap();
+    verify_and_check(&target, &crafted, &ctx, root1_target)
+        .expect_err("omitted change must be detected");
```
rkuris, line 59: Yes but let's check that it reports the *correct* error using `assert!(matches!(...))`

firewood/src/merkle/tests/change/attack.rs
```diff
@@ -0,0 +1,1023 @@
...
+}
+
+#[test]
+fn test_crafted_forged_value_detected() {
+    let (source, _dir_source) = setup_db![(b"\x10", b"v0"), (b"\xa0", b"v1")];
+    let (target, _dir_target) = setup_db![(b"\x10", b"v0"), (b"\xa0", b"v1")];
+    let root1_target = target.root_hash().unwrap();
```
rkuris: I mentioned this in another PR but here are more cases of setting up two databases (I think it was in #1946).

firewood/src/merkle/tests/change/attack.rs
```diff
@@ -0,0 +1,1023 @@
...
+    for pc in firewood_storage::PathComponent::ALL {
+        if let Some(ref mut h) = root_node.child_hashes[pc] {
+            let trie_hash = h.clone().into_triehash();
+            let bytes: [u8; 32] = trie_hash.into();
+            let mut new_bytes = bytes;
+            new_bytes[0] ^= 1;
+            *h = firewood_storage::TrieHash::from(new_bytes).into_hash_type();
+            break;
+        }
+    }
```
rkuris: This seems a bit fragile. We need to be sure we tweaked the bit. Maybe something like this (untested):
```
let h = end_nodes[0]                            
      .child_hashes                               
      .iter_mut()                                 
      .find_map(|(_, opt)| opt.as_mut())          
      .expect("root node must have at least one child hash");                                   
                                                  
  let mut bytes: [u8; 32] = h.clone().into_triehash().into();               
  bytes[0] ^= 1;                                  
  *h = firewood_storage::TrieHash::from(bytes).into_hash_type();
```

rkuris (changes requested):

firewood/src/merkle/tests/change/attack.rs
bernard-avalabs: Fixed — replaced all if let Ok(ctx) patterns with expect_err + assert!(matches!(...)) to check the specific error variant.

firewood/src/merkle/tests/change/attack.rs
bernard-avalabs: Moved test_crafted_stripped_divergent_child_rejected and test_collapse_root_hides_spurious_key.

firewood/src/merkle/tests/change/attack.rs
bernard-avalabs: Added assert!(matches!(...)) to all expect_err calls — each now checks the specific error variant.

firewood/src/merkle/tests/change/attack.rs
bernard-avalabs: Added setup_source_target! macro and converted 5 tests.

firewood/src/merkle/tests/change/attack.rs
bernard-avalabs: Fixed based on your suggestion.

rkuris (approved):
LGTM. I think that macro can be used in a few other places, such as test_crafted_conflicting_proof_nodes_rejected and test_crafted_value_at_odd_nibble_length_rejected. There are probably others.

Not blocking; you can choose to merge as-is or add some additional ones and merge.

demosdemon (approved):

bernard-avalabs:
> LGTM. I think that macro can be used in a few other places, such as test_crafted_conflicting_proof_nodes_rejected and test_crafted_value_at_odd_nibble_length_rejected. There are probably others.
> 
> Not blocking; you can choose to merge as-is or add some additional ones and merge.

I believe these tests use a single db with setup_2nd_commit! rather than the source/target pair that setup_source_target! creates, so the macro doesn't quite fit. Merging as-is.

PR #1950, opened 2026-04-23
fix(proofs): reject truncated exclusion proofs with reachable children

bernard-avalabs:
## Why this should be merged

This is porting a fix from the rkuris/restructure-squashed branch that was accidentally omitted in the previous PRs. Interestingly, none of our unit/integration tests caught this missing fix. It was only discovered as I was porting over the fuzz tests. I've added a new regression test specifically for this case.

This PR fixes the following: An exclusion proof can be truncated by removing the terminal node. If the remaining terminal node is an ancestor of the proven key and has a child at the next nibble toward that key, the truncated proof is incomplete. The verifier cannot determine whether the key exists deeper in the trie. Without the check added in this PR, a truncated exclusion proof can pass both structural validation and root hash verification, allowing an attacker to mask tampered batch_ops in change proofs.

This check matches AvalancheGo's verifyProofPath behavior which returns ErrExclusionProofMissingEndNodes in the ancestor case.

## How this works

  - Add ExclusionProofMissingChild variant to ProofError
  - In Proof::value_digest(), before returning Ok(None) for exclusion proofs, check if the terminal node has a child at the next nibble toward the key. If it does, the proof is incomplete — return Err(ExclusionProofMissingChild)
  - Add regression test truncated_exclusion_proof_rejected that creates an exclusion proof, truncates the terminal node, and verifies rejection

## How this was tested

  - truncated_exclusion_proof_rejected — fails without fix (returns Ok(None)), passes with fix
  - CI

## Breaking Changes

  - None

### DIFF
--- a/firewood/src/proofs/types.rs
+++ b/firewood/src/proofs/types.rs
@@ -487,9 +487,6 @@ impl<T: ProofCollection + ?Sized> Proof<T> {
         // extends past or diverges from the key), this is the valid
         // "divergent child" case — the node occupies the position where
         // the key would be, proving it cannot exist.
-        //
-        // This matches AvalancheGo's verifyProofPath behavior which
-        // returns ErrExclusionProofMissingEndNodes in the ancestor case.
         if let Some(next_idx) = next_nibble(last_node.full_path(), key.as_components())
             && last_node.children()[next_idx].is_some()
         {
--- a/firewood/src/proofs/types.rs
+++ b/firewood/src/proofs/types.rs
@@ -239,6 +239,11 @@ pub enum ProofError {
     /// Proof node is unreachable in the proving trie during collapse                                                 
     #[error("proof node unreachable in proving trie")]
     ProofNodeUnreachable,
+
+    /// Exclusion proof is incomplete — the terminal node has a child at the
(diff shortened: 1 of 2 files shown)


firewood/src/proofs/types.rs
```diff
@@ -466,7 +471,31 @@ impl<T: ProofCollection + ?Sized> Proof<T> {
...
+        //
+        // When no child exists at the next nibble, the ancestor alone is
+        // sufficient: the absence of a child proves the key cannot exist.
+        //
+        // When the last node's key is NOT an ancestor of the key (it
+        // extends past or diverges from the key), this is the valid
+        // "divergent child" case — the node occupies the position where
+        // the key would be, proving it cannot exist.
+        //
+        // This matches AvalancheGo's verifyProofPath behavior which
```
rkuris: This part is irrelevant. We care about correctness, which may or may not match avalanchego.

rkuris (approved):

firewood/src/proofs/types.rs
bernard-avalabs: I've removed the section on matching AvalancheGo's behavior.

PR #1951, opened 2026-04-24
test: add fuzz tests for change proof verification

### LINKED ISSUE #2007: Change-proof fuzz test fails with ProofError::UnexpectedValue in ethhash mode
## Summary

  The unified change-proof fuzz test (test_slow_change_proof_fuzz in firewood/src/merkle/tests/change/fuzz.rs) reliably fails when compiled with the ethhash feature. The test passes in merkledb mode.

  The test is currently gated behind #[cfg(not(feature = "ethhash"))] to keep CI green; this issue tracks reinstating ethhash coverage.

## Reproducer
  
  Ungate the fuzz module in `firewood/src/merkle/tests/change/mod.rs` by
  dropping the `#[cfg(not(feature = "ethhash"))]` line above `mod fuzz;`,
  then run:
  
  ```sh
  cargo nextest run -p firewood --features ethhash,logger --profile ci \
    -E 'test(test_slow_change_proof_fuzz)'
 ```

  Failures occur within ~10–20 seconds (roughly 1 in 15 outer iterations). One observed failing case: seed 7509880878819395899, run 15, scenario 39 (start_nonexistent).

bernard-avalabs:
## Why this should be merged

Last of the series of change proof verification PRs not including FFI changes. This PR builds on https://github.com/ava-labs/firewood/pull/1948. This PR ports over the fuzz test from the rkuris/cp-restructure-squashed branch.

The test *was* previously gated to merkledb mode (#2007) and surfaced a soundness gap (#2008). Both of these issues have since been addressed (#2058, #2060, and #2064). The fuzz test now passes for both ethhash and merkledb.

Since #2066 is stale (which originally stacked on this PR), most of its changes have been cherry-picked and included in this PR. 

Soaking the fuzzer also surfaced #2136 — a false rejection of a valid change proof — which is still open; see "Known flake" below.

Closes #2007

## How this works

A single test_slow_change_proof_fuzz test replaces what was originally three (fixed-size / varlen / adversarial) tests. Each outer iteration builds a fresh fixture; each inner iteration generates a valid proof, verifies it passes, then mutates it and verifies the mutation is rejected.

### Cherry-picked from #2066

This branch folds in the test fixes from #2066 rather than merging that PR wholesale (its base predates a month of main, and merging it would revert review fixes already applied here):

- **M1/M2 omit-guards**: skip the omit mutation when nothing but deletes follow the dropped op. A trailing run of deletes is unprovable, so the omission produces a smaller-range proof that legitimately verifies — not a detectable tamper.
- **M3 value-swap fix**: the mutation now flips the first byte of the value (or writes `0x1` for an empty value). The previous random-mask XOR could leave short values unchanged, and the resulting no-op "swap" was correctly accepted by the verifier — tripping the must-reject assertion spuriously.
- **`FIREWOOD_TEST_SOAK_SECONDS=<secs>`**: soak mode that keeps generating fresh random seeds until the deadline, for long local runs.
- **Failure diagnostics**: on a must-reject failure, the panic now dumps the full start/end proofs and batch ops alongside the seed/scenario/mutation locator.
- **CI runtime scope-down**: 3 iterations in debug builds, 25 in release (was 25/250); deep coverage moves to soak runs.

Deliberately **not** taken from #2066:

- The `#[cfg(not(feature = "ethhash"))]` re-gating of the fuzz module: its TODO(#2007) premise is fixed on main, and the fuzzer passes under ethhash (soaked, see below).

### Known flake (#2136)

Soaking this test surfaced #2136: an honestly-generated change proof is rejected with `EndRootMismatch` when a key just past the end bound was deleted. The fuzzer's positive-verification step can randomly construct that shape in the `end_nonexistent`/`both_nonexistent` scenarios. Observed rate: once in ~95,000 fixtures — at 25 fixtures per CI run, roughly one failure per ~4,000 CI runs. When it fires, the panic reads `positive root hash check should pass: seed=… scenario=…(end_nonexistent)` and reproduces deterministically via `FIREWOOD_TEST_SEED`, so it is identifiable as #2136 rather than a new regression. The flake disappears once #2136 is fixed.

## How this was tested

  - Several successful fuzz test runs.
  - One-hour soak: ~68,000 fixtures (~33k ethhash + ~35k merkledb), ≈6.8M positive-verification and mutation-rejection checks — zero failures.
  - A guard-removal experiment validated that the M1/M2 guards are still needed: without them, a soak hits an accepted trailing-delete omission within ~20k fixtures.
  - cargo fmt / clippy (pinned nightly, default and maxperf profiles) / cargo doc, and the full workspace nextest run with the ci profile (860/860 including slow tests) all pass.

## Breaking Changes

  - None


firewood/src/merkle/tests/change/mod.rs
```diff
@@ -80,5 +80,10 @@ mod bounds;
 // root hash, so change proofs from an empty database can't be generated.
 #[cfg(feature = "ethhash")]
 mod empty;
+// Fuzz tests are disabled in ethhash mode because PR #1935 recomputes
+// storageRoot in proof generation but change proof verification doesn't
+// account for this yet, causing value mismatches during reconciliation.
```
rkuris: This should be a TODO with a referenced issue instead of just a comment here. We need to find and fix these problems.

I'm not convinced #1935 is causing this problem, what evidence do you have that it is? Does the test pass if you back out #1935?

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+        db.propose(start_batch).unwrap().commit().unwrap();
+        let root1 = db.root_hash().unwrap();
+
+        // Build the end trie: delete ~15% of keys, insert ~15% new keys.
+        let delete_step = (start_keys.len() / 7).max(1);
+        let mut end_batch: Vec<BatchOp<&[u8], &[u8]>> = Vec::new();
+
+        // Delete every delete_step-th key.
+        let mut deleted_indices = Vec::new();
+        for i in (0..start_keys.len()).step_by(delete_step + 1) {
```
rkuris: This doesn't seem very random; the first key is always deleted, and since start_keys is always sorted, this means the lowest key is always deleted.
```suggestion
        for i in (rng.random_range(delete_step)..start_keys.len()).step_by(delete_step + 1) {
```

Why is this `delete_step + 1`?

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+            .map(|(k, v)| BatchOp::Put {
+                key: k.as_ref(),
+                value: v.as_ref(),
+            })
+            .collect();
+        db.propose(start_batch).unwrap().commit().unwrap();
+        let root1 = db.root_hash().unwrap();
+
+        // Build the end trie: delete ~15% of keys, insert ~15% new keys.
+        let delete_step = (start_keys.len() / 7).max(1);
```
rkuris: Why is `max(1)` needed here? `start_keys.len()` will always be > 7.
Probably safe to just add an `assert!(delete_step > 0)` if that's what the intent was.

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+    };
+
+    for (run, &seed) in seeds.iter().enumerate() {
+        eprintln!("run {run}: seed={seed} (export FIREWOOD_TEST_SEED={seed} to reproduce)");
+        let rng = firewood_storage::SeededRng::new(seed);
+
+        // Build the start trie from 64-2048 random keys.
+        let key_count = rng.random_range(64..=2048_u32);
+        let start_data = fixed_and_pseudorandom_data(&rng, key_count);
+        let mut start_keys: Vec<[u8; 32]> = start_data.keys().copied().collect();
```
rkuris: Always using 32 byte keys isn't the best strategy here; we probably need a TODO to cover at least the 64 byte keys as well as the non-ethhash cases where the keys vary in size more significantly.

Ah nevermind, there is a later test that uses varlen.

rkuris (commented):
Partial review, will add more comments shortly.

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+//! produce an end trie, generates a change proof between them, then verifies
+//! the proof under five scenarios:
+//!
+//! 1. Both boundary keys are existing keys
+//! 2. Start boundary is a non-existent (decreased) key
+//! 3. End boundary is a non-existent (increased) key
+//! 4. Both boundaries are non-existent
+//! 5. No bounds (complete proof)
+//!
+//! Runs 100 independent iterations, each with a freshly seeded RNG. On failure,
```
rkuris: I don't think it runs 100 independent iterations; comment is out of date.

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+                    verify_and_check(&db, &proof, &ctx, root1.clone())
+                        .expect("verify should pass (scenario 0)");
+                }
+
+                // 20% — start boundary is a non-existent (decreased) key.
+                36..56 => {
+                    if end_keys.len() < 2 {
+                        continue;
+                    }
+                    let si = rng.random_range(1..end_keys.len() - 1);
```
rkuris: if end_keys.len() == 2 (which is okay because the test is only skipped if end_keys.len() < 2) this panics -- random_range is documented to panic.

Odds of this seem quite low but not impossible.

There are other cases of this; please audit for all of them.

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+/// - Prefix key relationships exist (key `b"\x10"` coexisting with
+///   `b"\x10\x50"`), creating branch nodes with values
+/// - Exclusion proofs with divergent children are common (exercises the
+///   `PathIterator` divergent child fix and `ExclusionProofMissingChild`
+///   check in `value_digest`)
+#[test]
+#[expect(clippy::too_many_lines)]
+fn test_slow_change_proof_fuzz_varlen() {
+    // When FIREWOOD_TEST_SEED is set, run only that single seed (for
+    // reproducing CI failures). Otherwise, run 100 random iterations.
```
rkuris: same as earlier comment

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+
+    for (run, &seed) in seeds.iter().enumerate() {
+        eprintln!("run {run}: seed={seed} (export FIREWOOD_TEST_SEED={seed} to reproduce)");
+        let rng = firewood_storage::SeededRng::new(seed);
+
+        // Build the start trie from variable-length random keys.
+        let key_count = rng.random_range(16..=512_u32) as usize;
+        let start_kvs = generate_random_kvs(&rng, key_count);
+        let mut start_keys: Vec<Vec<u8>> = start_kvs.iter().map(|(k, _)| k.clone()).collect();
+        start_keys.sort();
```
rkuris: nit: everywhere else uses sort_unstable, which is faster, especially since you're going to dedup afterwards anyway. Doesn't really need fixing but this seemed inconsistent.

firewood/src/merkle/tests/change/fuzz.rs
```diff
@@ -0,0 +1,1589 @@
...
+
+        // Build the sorted list of keys in the end state.
+        let mut end_keys: Vec<Vec<u8>> = start_keys
+            .iter()
+            .enumerate()
+            .filter(|(i, _)| !deleted_indices.contains(i))
+            .map(|(_, k)| k.clone())
+            .chain(new_keys.iter().cloned())
+            .collect();
+        end_keys.sort();
```
rkuris: nit: another case where sort_unstable should be used for consistency.

rkuris (changes requested):
This is a fairly big diff, mostly because it seems like most of the code does the same thing for varlen and fixed length keys.

It's kind of on the edge here (I am not a huge fan of test helpers) but since the two sections have diverged already it might be simpler if it were refactored.

If you feel strongly it should be left as-is, I'll take another pass and compare the two cases more carefully. I only spotted one discrepency.

firewood/src/merkle/tests/change/mod.rs
bernard-avalabs: I've changed it to a TODO that points to issue #2007. I confirmed that it fails without the gating, but I didn't try to back out of #1935 since that requires a lot of changes.

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Replaced the step_by(delete_step + 1) pattern with truly random per-key sampling.

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Done. max(1) and the surrounding delete_step variable are both gone

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: I've consolidated the three tests into a single test_slow_change_proof_fuzz.

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Done. Module docstring now says: "Runs 25 iterations in debug builds and 250 in release builds".

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Done. Audited every random_range call. The three sites that use (1..end_keys.len() - 1) (scenarios 1 and 3 in pick_proof_for_scenario, plus the combined-mutation block) all now have if end_keys.len() < 3 guards.

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: The varlen test was deleted as part of the consolidation, so this comment is no longer present. The remaining unified test has the correct iteration-count docstring.

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Done. All sorts use sort_unstable().

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Done. All sorts use sort_unstable().

bernard-avalabs:
> This is a fairly big diff, mostly because it seems like most of the code does the same thing for varlen and fixed length keys.
> 
> It's kind of on the edge here (I am not a huge fan of test helpers) but since the two sections have diverged already it might be simpler if it were refactored.
> 
> If you feel strongly it should be left as-is, I'll take another pass and compare the two cases more carefully. I only spotted one discrepency.

The three tests (test_slow_change_proof_fuzz, test_slow_change_proof_fuzz_varlen, test_slow_adversarial_change_proof_fuzz) are now a single test_slow_change_proof_fuzz. Each inner iteration runs both positive verification (with optional 10% serialize round-trip) and an adversarial mutation, asserting rejection.

The function is split into helpers (build_trie_fixture / build_primary_trie / build_cross_db_trie for setup, pick_proof_for_scenario for boundary selection, assert_proof_rejected for the rejection check). Repeated patterns inside the mutation logic are now macros: pick_op_index!, proof_parts!, clear_first_child_hash!, mutate_random_child_hash!.

One thing surfaced by the new test: it reliably finds a soundness bug — an attacker can drop an in-range Put from batch_ops and the verifier still accepts. It looks like the left-edge analog of the bug fixed in #1970. 

Reproducer: FIREWOOD_TEST_SEED=16855342578241928557 cargo nextest run -p firewood --features logger --profile ci -E 'test(test_slow_change_proof_fuzz)'. Created a separate issue for this potential bug (https://github.com/ava-labs/firewood/issues/2008).

rkuris:
We can't merge this until 2008 is fixed, so let's get it in draft and work on 2008.

bernard-avalabs:
> We can't merge this until 2008 is fixed, so let's get it in draft and work on 2008.

https://github.com/ava-labs/firewood/issues/2007 and https://github.com/ava-labs/firewood/issues/2008 have been fixed. I've re-enabled the fuzz test for both merkledb and ethhash. Re-opening this PR for review.

firewood/src/merkle/tests/change/fuzz.rs
rkuris: nit: the comment is tautological and sure to get out of date when someone tweaks the numbers later.

rkuris (commented):
The "how this was tested" section says you only ran CI. If that's correct, let's bake it for much longer before considering this for merging. If that's incorrect, please indicate how you actually tested this.

rkuris:
> > We can't merge this until 2008 is fixed, so let's get it in draft and work on 2008.
> 
> #2007 and #2008 have been fixed. I've re-enabled the fuzz test for both merkledb and ethhash. Re-opening this PR for review.

2008 is indeed fixed but the issue was open. Can you check on 2007 and close the issue if it's indeed fixed?

bernard-avalabs:
> > > We can't merge this until 2008 is fixed, so let's get it in draft and work on 2008.
> > 
> > 
> > #2007 and #2008 have been fixed. I've re-enabled the fuzz test for both merkledb and ethhash. Re-opening this PR for review.
> 
> 2008 is indeed fixed but the issue was open. Can you check on 2007 and close the issue if it's indeed fixed?

Yes, #2007  was addressed in #2060

bernard-avalabs:
> The "how this was tested" section says you only ran CI. If that's correct, let's bake it for much longer before considering this for merging. If that's incorrect, please indicate how you actually tested this.

I've updated the PR description. I've done multiple runs of the fuzz test, but I can run for a few hours to make sure I am not missing anything.

firewood/src/merkle/tests/change/fuzz.rs
bernard-avalabs: Fixed.

PR #1952, opened 2026-04-27
fix: resolve false rejection at straddling nibbles in collapse_strip

bernard-avalabs:
## Why this should be merged                                                                                                                                                                                                                         
                  
collapse_strip incorrectly rejects valid change proofs when an inherited out-of-range key sits at a nibble position that straddles the proof's boundary range. The nibble-level range check (child_in_range) cannot distinguish inherited         
out-of-range keys from in-range keys that indicate tampered batch_ops, causing legitimate proofs to fail with EndRootMismatch.
                                                                                                                                                                                                                                                      
## How this works                                                                                                                                                                                                                                    
                  
Renames the existing nibble-level check to nibble_in_range and introduces a new child_in_range method that recurses into the subtree when the nibble-level check is ambiguous (straddling). For leaves, the full key is compared directly against the range. For branches, children are pruned with nibble_in_range and checked recursively. The recursion terminates quickly because path compression keeps subtrees between proof nodes shallow.

child_in_range is read-only — it borrows the child via as_shared_node and inspects the subtree without mutating the nodestore. The caller in collapse_strip strips the child afterward.
                                                                                                                                                                                                                                                    
  - Out-of-range subtrees: stripped                                                                                                                                                                                             
  - In-range keys found: rejected with EndRootMismatch
                                                                                                                                                                                                                                                    
## How this was tested
                                                                                                                                                                                                                                                    
  - test_inherited_key_at_start_boundary_accepted — legitimate proof with only out-of-range keys under a straddling nibble (previously failed, now passes)                                                                                          
  - test_crafted_omitted_delete_at_straddling_nibble — tampered proof with in-range key hidden under the same straddling nibble (security test, passes)
  - New test_child_in_range unit test. Old one renamed test_nibble_in_range.
  -  CI 
                                                                                                                                                                                                                                                    
## Breaking Changes                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                    
  - None

### DIFF
--- a/firewood/src/merkle/collapse.rs
+++ b/firewood/src/merkle/collapse.rs
@@ -9,13 +9,15 @@ use firewood_storage::{
 
 use crate::{ProofError, api, merkle::Merkle};
 
-/// Returns `true` when a child at nibble `child_nib` under the accumulated
-/// prefix `acc_prefix` falls within `[start_nib, end_nib]` (inclusive).
+/// Returns `true` when a nibble at position `child_nib` under the accumulated
+/// prefix `acc_prefix` could contain keys within `[start_nib, end_nib]`.
 ///
-/// Used by [`Merkle::collapse_strip`] to distinguish in-range children
-/// (which indicate tampered `batch_ops`) from out-of-range children (which
-/// are stripped normally).
-fn child_in_range(
+/// This is a fast, nibble-level check used by [`Merkle::child_in_range`] as
+/// a first pass. It may return `true` for straddling nibbles — positions
+/// where the boundary passes through the subtree — even if no actual keys
+/// in the subtree are in range. `child_in_range` recurses into the subtree
+/// to resolve those cases.
+fn nibble_in_range(
     acc_prefix: &[u8],
     nibble: PathComponent,
     start_nib: &[u8],
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
         Ok(node)
     }
 
+    /// Returns `true` when the subtree rooted at `child` contains any key
+    /// within `[start_nib, end_nib]`.
+    ///
+    /// Reads the child node, computes its full nibble prefix, and checks
+    /// whether the key is in range. For leaves this is a direct comparison.
+    /// For branches, recurses into children whose nibble passes
+    /// [`nibble_in_range`] to resolve straddling positions.
+    ///
+    /// The caller is responsible for checking [`nibble_in_range`] and taking
+    /// the child from its slot before calling this method. The child is
+    /// consumed regardless of the result.
+    fn child_in_range(
+        &mut self,
+        child: Child,
+        acc_prefix: &[u8],
+        nibble: PathComponent,
+        start_nib: &[u8],
+        end_nib: &[u8],
+    ) -> Result<bool, api::Error> {
+        if !nibble_in_range(acc_prefix, nibble, start_nib, end_nib) {
+            return Ok(false);
+        }
+
+        let mut child_node = self.read_for_update(child)?;
+        let pfx = build_child_prefix(acc_prefix, nibble.0.as_u8(), &child_node);
+        let key_in_range = pfx.as_ref() >= start_nib && pfx.as_ref() <= end_nib;
+
+        let Some(branch) = child_node.as_branch_mut() else {
+            return Ok(key_in_range);
+        };
+
+        if key_in_range && branch.value.is_some() {
+            return Ok(true);
+        }
+
+        for (child_nibble, child_slot) in &mut branch.children {
+            let Some(inner_child) = child_slot.take() else {
+                continue;
+            };
+            if self.child_in_range(inner_child, &pfx, child_nibble, start_nib, end_nib)? {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
     /// Strip non-on-path children from an intermediate branch and recurse.
     /// Flattens single-child branches to match `end_root`'s path-compressed
     /// structure.
     ///
-    /// Children that are both in-range AND proposal-local (created by this
-    /// proposal's `batch_ops`) trigger `EndRootMismatch` — they indicate
-    /// tampered operations since `end_root` has no branch at this position.
-    /// In-range children inherited from the parent are legitimate unchanged
-    /// keys and are stripped normally (handled by proof hashes).
+    /// Non-on-path children containing in-range keys trigger
+    /// `EndRootMismatch` — they indicate tampered `batch_ops` since
+    /// `end_root` has no branch at this position. Non-on-path children
+    /// containing only out-of-range keys are stripped normally (accounted
+    /// for by proof hashes).
     fn collapse_strip(
         &mut self,
         mut node: Node,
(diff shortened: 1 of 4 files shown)


firewood/src/merkle/tests/change/attack.rs
```diff
@@ -555,6 +555,61 @@ fn test_crafted_divergence_at_depth_zero() {
...
+///
+/// `\x05` is in-range (>= `start_key` `\x01`) and shares nibble 0 with the
+/// out-of-range `\x00`. If the collapse step unconditionally strips all
+/// non-on-path children (without checking for in-range keys), the proving
+/// trie is flattened to match `end_root` and the tampered proof is accepted.
+///
+/// The verifier would then commit the incomplete `batch_ops`, leaving `\x05`
+/// in their state when `end_root` says it shouldn't exist.
+#[test]
+fn test_omitted_delete_at_straddling_nibble_rejected() {
```
rkuris: naming: all the other tests are named `test_crafted_*` when they craft proofs.

This lets you do stuff like `cargo nextest run --workspace --features ethhash,logger test_crafted_`

probably don't need "_rejected" at the end either; that convention is used int he adversarial naming section.

```suggestion
fn test_crafted_omitted_delete_at_straddling_nibble() {
```

firewood/src/merkle/collapse.rs
```diff
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+    /// within `[start_nib, end_nib]`.
+    ///
+    /// Reads the child node, computes its full nibble prefix, and checks
+    /// whether the key is in range. For leaves this is a direct comparison.
+    /// For branches, recurses into children whose nibble passes
+    /// [`nibble_in_range`] to resolve straddling positions.
+    ///
+    /// The caller is responsible for checking [`nibble_in_range`] and taking
+    /// the child from its slot before calling this method. The child is
+    /// consumed regardless of the result.
```
rkuris: This appears to be false. This function calls nibble_in_range and handles it. Not only that, but the one caller (collapse_strip) doesn't do this.

firewood/src/merkle/collapse.rs
```diff
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+    fn child_in_range(
+        &mut self,
```
rkuris: This shouldn't need to be `mut`

firewood/src/merkle/collapse.rs
```diff
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+        child: Child,
```
rkuris: Why is this not a reference?

firewood/src/merkle/collapse.rs
```diff
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+        acc_prefix: &[u8],
+        nibble: PathComponent,
+        start_nib: &[u8],
+        end_nib: &[u8],
+    ) -> Result<bool, api::Error> {
+        if !nibble_in_range(acc_prefix, nibble, start_nib, end_nib) {
+            return Ok(false);
+        }
+
+        let mut child_node = self.read_for_update(child)?;
```
rkuris: ah, here's why it had to be mut. We shouldn't be calling this here as we don't update the node. This causes some odd effects in the deleted node list for some mutable node types, which we don't want or need here.

firewood/src/merkle/collapse.rs
```diff
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+        let pfx = build_child_prefix(acc_prefix, nibble.0.as_u8(), &child_node);
+        let key_in_range = pfx.as_ref() >= start_nib && pfx.as_ref() <= end_nib;
+
+        let Some(branch) = child_node.as_branch_mut() else {
```
rkuris: `as_branch` here

firewood/src/merkle/collapse.rs
```diff
@@ -287,15 +289,61 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+            return Ok(key_in_range);
+        };
+
+        if key_in_range && branch.value.is_some() {
+            return Ok(true);
+        }
+
+        for (child_nibble, child_slot) in &mut branch.children {
+            let Some(inner_child) = child_slot.take() else {
```
rkuris: This also seems wrong. perhaps `child_slot.as_ref()`?

rkuris (changes requested):
There is a solid bug here: inspecting the node should not cause extra entries in the proposals future-delete log. To fix this, child_in_range should take &self, not &mut self. Also `collapse_strip` probably shouldn't use `slot.take()` (although it needs to set `*slot = None` somewhere)

firewood/src/merkle/tests/change/attack.rs
bernard-avalabs: Fixed — renamed to test_crafted_omitted_delete_at_straddling_nibble.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed — removed the false claims.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed — changed to &self. Uses child.as_shared_node(&self.nodestore) instead of read_for_update, so no mutation needed.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed — changed to &Child.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed — replaced with child.as_shared_node(&self.nodestore) which reads without adding to the deletion log.

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed — changed to as_branch().

firewood/src/merkle/collapse.rs
bernard-avalabs: Fixed.

bernard-avalabs:
> There is a solid bug here: inspecting the node should not cause extra entries in the proposals future-delete log. To fix this, child_in_range should take &self, not &mut self. Also `collapse_strip` probably shouldn't use `slot.take()` (although it needs to set `*slot = None` somewhere)

I've addressed the main problem by using `as_shared_node` instead of `read_for_update`. I've also rebased to main so this PR can be merged without merging the attack.rs PR first. This required that I move one helper function from attack.rs to this PR.

firewood/src/merkle/collapse.rs
```diff
@@ -9,13 +9,15 @@ use firewood_storage::{
...
 ///
-/// Used by [`Merkle::collapse_strip`] to distinguish in-range children
-/// (which indicate tampered `batch_ops`) from out-of-range children (which
-/// are stripped normally).
-fn child_in_range(
+/// This is a fast, nibble-level check used by [`Merkle::child_in_range`] as
+/// a first pass. It may return `true` for straddling nibbles — positions
+/// where the boundary passes through the subtree — even if no actual keys
+/// in the subtree are in range. `child_in_range` recurses into the subtree
+/// to resolve those cases.
```
rkuris, line 19: Unfortunately, there aren't any performance tests to see how much difference this makes. I don't think there's a good alternative but we should consider creating a new ticket to add some benchmarks in this area.

rkuris (approved):

firewood/src/merkle/collapse.rs
bernard-avalabs: Created issue: https://github.com/ava-labs/firewood/issues/1974

PR #1955, opened 2026-05-01
chore: export errRevisionNotFound

RodrigoVillar:
## Why this should be merged

In https://github.com/ava-labs/avalanchego/pull/5318, the block reexecution tests for using Firewood in archival mode check that the revision doesn't exist to verify that we're reexecuting blocks. We've had to define a new error `ErrNoRevisionFound` in Coreth for usage with `require.ErrorIs(...)` but Firewood already defines its own error for this: `errRevisionNotFound` - being able to use this error would be cleaner.

## How this works

Exports `errRevisionNotFound`

## How this was tested

CI

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -376,7 +376,7 @@ func (db *Database) Get(key []byte) ([]byte, error) {
 	val, err := getValueFromValueResult(C.fwd_get_latest(db.handle, newBorrowedBytes(key, &pinner)))
 	// The revision won't be found if the database is empty.
 	// This is valid, but should be treated as a non-existent key
-	if errors.Is(err, errRevisionNotFound) {
+	if errors.Is(err, ErrRevisionNotFound) {
 		return nil, nil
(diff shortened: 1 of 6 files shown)


alarso16 (commented):
I understand the merit of this PR, but I don't think this is a worthwile extension of the ffi API. What I mean is, is there any reason we would handle this error differently than any other error returned by `Revision`? In the avalanchego code, we could use the `*trie.MissingNodeError` to be shared with HashDB/PathDB. this error is also returned if the database is closed (for better or worse)

alarso16 (commented):
EDIT: GitHub was struggling, ignore

rkuris (approved):

demosdemon (approved):

PR #1956, opened 2026-05-03
fix(fwdctl): drop legacy rustls feature to clear GHSA-82j2-j2ch-gfr8

rkuris:
## Why this should be merged

GitHub Dependabot alert #33 (GHSA-82j2-j2ch-gfr8, CVSS 7.5 high) flags rustls-webpki < 0.103.13 for an out-of-bounds panic in `bit_string_flags()` when parsing a malformed CRL `issuingDistributionPoint` extension. fwdctl was pulling in the vulnerable rustls-webpki 0.101.7 transitively through the legacy `rustls` default feature on aws-sdk-{ec2,ssm,sts}.

## How this works

aws-sdk-{ec2,ssm,sts} default features include `rustls`, which activates `aws-smithy-runtime/tls-rustls` -> `aws-smithy-http-client/legacy-rustls-ring`, pinning rustls 0.21 / rustls-webpki 0.101. Their `default-https-client` feature uses the modern `rustls-aws-lc` stack instead, which resolves to rustls 0.23 / rustls-webpki 0.103.

Disable default features on these three SDK crates and re-select only `default-https-client` and `rt-tokio` (plus `sigv4a` for sts). aws-config's defaults already use the modern stack, so it is unchanged.

After the change Cargo.lock contains only rustls-webpki 0.103.13 (patched) and rustls 0.23.40; the legacy 0.21.12 / 0.101.7 pair is gone.

## How this was tested

- `cargo build -p firewood-fwdctl --features launch` succeeds.
- `cargo nextest run -p firewood-fwdctl --features ethhash,logger,launch`: 12/12 tests pass.
- `cargo clippy -p firewood-fwdctl --features ethhash,logger,launch --all-targets`: no warnings.
- Verified Cargo.lock no longer contains rustls-webpki 0.101.7.

## Breaking Changes

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -529,23 +529,17 @@ dependencies = [
  "aws-smithy-async",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
- "h2 0.3.27",
- "h2 0.4.13",
- "http 0.2.12",
+ "h2",
  "http 1.4.0",
- "http-body 0.4.6",
- "hyper 0.14.32",
- "hyper 1.9.0",
- "hyper-rustls 0.24.2",
- "hyper-rustls 0.27.9",
+ "hyper",
+ "hyper-rustls",
  "hyper-util",
  "pin-project-lite",
- "rustls 0.21.12",
- "rustls 0.23.40",
+ "rustls",
  "rustls-native-certs",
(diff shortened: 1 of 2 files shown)


AminR443 (approved):
LGTM.

PR #1958, opened 2026-05-03
perf: migrate firewood off the upstream `rlp` crate

### LINKED ISSUE #1944: perf: replace rlp crate with in-tree minimal RLP module
## Summary

Replace the `rlp = "0.6.1"` dependency in production code paths with a small, purpose-built module at `storage/src/rlp.rs`, exposed as `firewood_storage::rlp`. The module covers only the RLP shapes firewood actually uses.

## Motivation

The `rlp` crate is general-purpose and its decode API forces us through `Vec<Vec<u8>>` in the hot path (see `replace_hash` in `storage/src/hashers/ethhash.rs`, and `format_node_value` in `storage/src/lib.rs`). We only use a narrow slice of its functionality — building 2- and 17-element lists and extracting one field from an account record — so a bespoke module can both remove a dependency and eliminate the allocation overhead.

Key wins:

- Replace the `parse → Vec<Vec<u8>> → mutate → re-encode` pattern in the account storage-root rewrite with a single `replace_list_field` splice that does one allocation plus a memcpy.
- Decode returns borrowed `&[u8]` slices instead of owned `Vec<u8>`.
- Drop the `rlp` crate from `firewood-storage` (non-dev), `firewood-ffi`, and `firewood` (non-dev) dependency graphs.

## Scope

**Migrated:**

- `storage/src/hashers/ethhash.rs`
- `storage/src/nodestore/hash.rs` (`fix_account_storage_root_value`)
- `storage/src/lib.rs` (`format_node_value`)
- `ffi/src/proofs/change.rs` (account code-hash extraction)

**Not migrated:** `firewood-triehash` keeps its direct `rlp` dep (it is test-only inside this workspace but is a separately published crate; blast-radius reasons). The one test file at `firewood/src/merkle/tests/ethhash.rs` also stays on `rlp` alongside triehash.

## Full plan

The detailed plan — API surface, migration steps (one commit each), unit test checklist, and open items — is attached below.

<details>
<summary>RLP_MIGRATION_PLAN.md</summary>

# Plan: replace `rlp` crate with in-tree `storage/src/rlp.rs`

## Goal

Stop pulling `rlp = "0.6.1"` into production code paths. Replace it with a
small, purpose-built module living at `storage/src/rlp.rs` (exposed as
`firewood_storage::rlp`). The new module covers only the RLP shapes firewood
actually uses and returns borrowed `&[u8]` on decode instead of owned
`Vec<Vec<u8>>`.

## Scope: what gets migrated, what doesn't

**Migrated (production code):**

- `storage/src/hashers/ethhash.rs` — leaf/branch encoding + `replace_hash`
  decode-and-rewrite
- `storage/src/nodestore/hash.rs::fix_account_storage_root_value` — branch
  encoding for the account-depth storage root
- `storage/src/lib.rs::format_node_value` — decode-for-display of RLP node
  values
- `ffi/src/proofs/change.rs` — account code-hash field extraction

**Not migrated:**

- `firewood-triehash` keeps its direct `rlp = "0.6"` dep. It is a test-only
  crate inside this workspace (dev-dep of `firewood`, only used from
  `firewood/src/merkle/tests/ethhash.rs`, plus its own benches) but is
  published externally, so keeping its transitive `rlp` dependency minimizes
  blast radius.
- `firewood/src/merkle/tests/ethhash.rs` — stays on `rlp` crate (test-only,
  and already uses `firewood_triehash::trie_root`).
- `firewood/Cargo.toml` keeps `rlp = "0.6.1"` as a dev-dep for that one test
  file.
- `storage/Cargo.toml` keeps `rlp = "0.6.1"` as a **dev-dep** for the
  compatibility tests in the new module (see step 1).

Net effect: the production dependency graph loses `rlp` from `storage`
(non-dev), `ffi`, and `firewood` (non-dev). Dev/test still have it, which is
fine.

## API surface for `storage/src/rlp.rs`

Derived strictly from observed usage (see survey below). No generics beyond
what call sites need; no traits; no derives.

```rust
/// RLP encoding of an empty byte string — replaces `rlp::NULL_RLP`.
pub const EMPTY_STRING: &[u8] = &[0x80];

// ---------- errors ----------
#[derive(Debug, thiserror::Error, PartialEq, Eq)]
pub enum RlpError {
    #[error("RLP input truncated")]
    Truncated,
    #[error("RLP length prefix overflows")]
    Overflow,
    #[error("expected a byte string, found a list")]
    ExpectedBytes,
    #[error("expected a list, found a byte string")]
    ExpectedList,
    #[error("RLP index {0} out of range")]
    IndexOutOfRange(usize),
    #[error("trailing bytes after RLP item")]
    TrailingBytes,
}

// ---------- encoding ----------
// All encoders return `Box<[u8]>`: the output size is known at the moment
// of return and callers never need to grow the buffer further, so the
// extra `capacity` word in `Vec<u8>` is dead weight.

/// Encode a single byte-string item.
pub fn encode_bytes(b: &[u8]) -> Box<[u8]>;

/// Encode a `usize` as RLP (big-endian minimal byte-string).
pub fn encode_usize(n: usize) -> Box<[u8]>;

/// One child of an RLP list, as it arrives from the caller.
#[derive(Copy, Clone)]
pub enum RlpItem<'a> {
    /// A byte-string child; `encode_list` will add the correct RLP header.
    Bytes(&'a [u8]),
    /// A pre-encoded RLP item (used when a child was already hashed /
    /// encoded separately and its bytes should be copied verbatim).
    Raw(&'a [u8]),
    /// The empty byte-string marker (0x80).
    Empty,
}

/// Encode a slice of children as a single RLP list.
///
/// The output length is computed up-front from the sum of per-item lengths
/// plus the list header, so this performs exactly one allocation. Returns
/// a `Box<[u8]>` — callers never need to grow the result.
pub fn encode_list(items: &[RlpItem<'_>]) -> Box<[u8]>;

// ---------- decoding (zero-copy) ----------
/// A parsed RLP list, holding a slice into the caller's buffer.
pub struct RlpList<'a> { /* payload: &'a [u8] */ }

impl<'a> RlpList<'a> {
    /// Parse `bytes` as a single RLP list item. Errors if the input is
    /// a byte-string item, truncated, or has trailing bytes.
    pub fn parse(bytes: &'a [u8]) -> Result<Self, RlpError>;

    /// Iterate raw child items as `&[u8]` byte-string payloads.
    /// Yields `RlpError::ExpectedBytes` if a child is itself a list.
    pub fn iter(&self) -> RlpItemIter<'a>;

    /// Return the `i`-th child as a byte-string slice.
    pub fn nth_bytes(&self, i: usize) -> Result<&'a [u8], RlpError>;
}

pub struct RlpItemIter<'a> { /* ... */ }
impl<'a> Iterator for RlpItemIter<'a> { type Item = Result<&'a [u8], RlpError>; /* ... */ }

// ---------- in-place splice (fast path for account storageRoot rewrite) ----------
// Replaces a single byte-string field of an RLP list without materializing
// the other fields. Implemented as: parse the list header, walk item
// boundaries until reaching `field_index`, allocate an output buffer whose
// length equals the input's (when `replacement.len()` equals the original
// field payload length — the 32-byte hash case always satisfies this),
// copy the input bytes, and overwrite the field payload with `replacement`.
//
// If `replacement.len()` differs from the original, the list header and the
// replaced field's length prefix must be recomputed, so the implementation
// falls back to a single streaming copy (still no owned Vec<Vec<u8>>).
//
// Used by the ethhash account storageRoot rewrite; replaces the current
// `replace_hash` helper in `storage/src/hashers/ethhash.rs`.

/// Return a new RLP list identical to `input` except that the byte-string
/// at `field_index` is replaced by `replacement`.
///
/// Errors if `input` is not a well-formed RLP list, if it has fewer than
/// `field_index + 1` items, or if the target item is itself a list.
pub fn replace_list_field(
    input: &[u8],
    field_index: usize,
    replacement: &[u8],
) -> Result<Box<[u8]>, RlpError>;
```

### Deliberate omissions

- No `Encodable` / `Decodable` traits. Each call site knows its shape.
- No builder / streaming encoder. Every encode site in firewood knows its
  list length up front (always 2 or 17), so a one-shot `encode_list(&items)`
  with a fully-materialized slice is simpler than a mutable builder.
- No nested-list decoding API. Every decode site in firewood today reads a
  flat list of byte-string fields; if a future site needs nested decoding we
  add it then.
- No `append_bytes_from_iter` equivalent. The only caller that feeds bytes
  from an iterator (`hex_prefix_encode` in `triehash`) is not being
  migrated; every ethhash/nodestore call site already has a materialized
  `&[u8]` (from `nibbles_to_eth_compact`).
- No public `encode_usize`-on-a-list helper — only `triehash` encodes a
  usize, and triehash is not being migrated. `encode_usize` stays as a
  free function for completeness/tests but isn't wired into any call site
  today.
- No `RlpList::len()` — no call site needs it.

## Design wins vs. the `rlp` crate

1. **No `Vec<Vec<u8>>` on the decode path.** The current `replace_hash`
   does `Rlp::as_list::<Vec<u8>>() → mutate → re-encode`, allocating one
   owned `Vec<u8>` per child just to copy it back out again. The new code
   replaces this with `replace_list_field`, which parses only enough header
   bytes to locate field 2's byte offsets, allocates a single output buffer
   of the correct size, and writes through the input with the replacement
   field's bytes patched in. In the 32-byte hash case (the only one in
   firewood today) the output length equals the input length and the only
   bytes that differ are the 32-byte payload — effectively one allocation
   and one `memcpy` + one overwrite, instead of N allocations for the
   owned child vectors.
2. **Concrete types only.** Binary size and compile time both benefit from
   dropping the trait-based `Encodable`/`Decodable` machinery.
3. **Explicit error type.** `rlp::DecoderError` is broad; `RlpError` matches
   the small set of failure modes we actually care about.

## Migration steps (one commit per step)

1. **Land the module.** Add `storage/src/rlp.rs`, wire into `storage/src/lib.rs`
   as `pub mod rlp`. Unit tests cover:
   - Single byte < 0x80 (encoded as itself)
   - Byte strings of length 1, 55, 56, 255, 256, 2^16
   - Empty string (`EMPTY_STRING`)
   - Lists with payloads of length 0, 55, 56, 1024
   - `encode_usize` for 0, 1, 127, 128, 255, 256, `u32::MAX`, `usize::MAX`
   - Round-trip parity: encode with `encode_list`, decode with upstream
     `rlp::Rlp::as_list::<Vec<u8>>` and compare (uses `rlp` as a dev-dep
     in `storage/Cargo.toml`)
   - Reverse parity: encode with upstream `rlp::RlpStream::new_list(N)`,
     parse with `RlpList::parse` and compare
   - Decoder rejects trailing bytes, truncated headers, byte-string-where-list-expected
   - `replace_list_field` on a real 4-field account value: equal-length
     replacement (32-byte hash), shorter replacement, longer replacement;
     each result is compared against the upstream `rlp` crate's decoded
     view for byte-for-byte equivalence
2. **Migrate `storage/src/hashers/ethhash.rs`.** Replace each
   `RlpStream::new_list(N)` + appends + `.out()` sequence with a single
   `encode_list(&[RlpItem::Bytes(..), RlpItem::Raw(..), RlpItem::Empty, ..])`
   call. Replace the `replace_hash` helper with a direct call to
   `crate::rlp::replace_list_field(value, 2, &new_hash)` — the
   `Vec<Vec<u8>>` round-trip disappears entirely and the helper becomes a
   one-line wrapper (or the call sites use `replace_list_field` directly;
   decide during implementation based on readability). Also replace
   `rlp::NULL_RLP` with `crate::rlp::EMPTY_STRING`.
3. **Migrate `storage/src/nodestore/hash.rs::fix_account_storage_root_value`.**
   Same `encode_list` swap on the 17-element branch; encode-only site.
4. **Migrate `storage/src/lib.rs::format_node_value`.** Swap
   `rlp::Rlp::new(value).as_list::<Vec<u8>>()` for
   `RlpList::parse(value)?.iter()` — avoids allocating one `Vec<u8>` per
   displayed field.
5. **Migrate `ffi/src/proofs/change.rs`.** One line: the code-hash extractor
   becomes `RlpList::parse(value)?.nth_bytes(3)?`. Drop `rlp` from
   `ffi/Cargo.toml`.
6. **Remove `rlp` from `storage/Cargo.toml` non-dev deps.** It stays as a
   dev-dep for the compatibility tests from step 1.
7. **CI.** After each commit and finally in aggregate:

   ```bash
   cargo fmt
   cargo nextest run --workspace --features ethhash,logger --all-targets --profile ci
   cargo doc --no-deps
   ```

   (The `ci` profile is required because ethhash is exercised by
   `test_slow_*` paths per CLAUDE.md.)

## Usage survey (source of truth for the API)

- **Leaf encoding** — `storage/src/hashers/ethhash.rs:110-137` —
  `RlpStream::new_list(2)` → append path / bytes / empty → `.out()`
- **Branch encoding** — `storage/src/hashers/ethhash.rs:144-164` —
  `RlpStream::new_list(17)` → append hash / append_raw / append_empty /
  value → `.out()`
- **Nested leaf for single-child account** —
  `storage/src/hashers/ethhash.rs:190-194` —
  `new_list(2)` + `append_iter` + `append_raw`
- **Account wrapper re-encode** —
  `storage/src/hashers/ethhash.rs:223-233` —
  `new_list(2)` + append path + append bytes
- **`replace_hash` decode + rewrite** —
  `storage/src/hashers/ethhash.rs:242-265` —
  `Rlp::new(bytes).as_list()` → mutate index 2 →
  `RlpStream::new_list(list.len())` + per-item append
- **Storage-root branch hash** —
  `storage/src/nodestore/hash.rs:290-324` —
  `RlpStream::new_list(17)` + appends
- **Empty-trie constant** —
  `storage/src/nodestore/hash.rs:296` and `ethhash.rs:118` — `rlp::NULL_RLP`
- **Node value pretty-print** — `storage/src/lib.rs:435` —
  `rlp::Rlp::new(v).as_list::<Vec<u8>>()`
- **Account code-hash extract** — `ffi/src/proofs/change.rs:310` —
  extracts field 3 of the account list (`at(3).data()`)

## Open items (confirmed before this plan)

1. **Feature gate the new module?** No. Keep `storage::rlp` unconditional —
   the module is small and all current callers happen to be
   `#[cfg(feature = "ethhash")]`, but gating the module itself adds noise.
   Callers stay gated as-is.
2. **Scope of `replace_hash` refactor.** Do it as part of step 2 (same
   commit as the ethhash.rs migration), since the whole point of the
   migration is to kill the `Vec<Vec<u8>>` round-trip. Replaced in full by
   `replace_list_field` — no streaming-through-encoder intermediate.
3. **Triehash migration.** Deferred indefinitely. Triehash keeps `rlp`.
4. **Firewood's dev-dep on `rlp`.** Kept (used by `merkle/tests/ethhash.rs`
   alongside triehash).

</details>

rkuris:
TLDR: ~2.5% wall-clock speedup over main, consistent across first 1M blocks of C-chain. Probably faster when there are more account updates, later in the chain.

Second step of #1944. Stacked on top of #1957.

Replaces every production-code use of the upstream `rlp = "0.6"` crate
with calls into the in-tree `crate::rlp` module landed in #1957, then
removes `rlp` (and `bytes`, which was only used for the `BytesMut`
round-trip in the old `replace_hash` helper) from the production
dependency graph. Finally, tightens the visibility of `crate::rlp` so
only the items production callers actually need are part of the
public API.

## Migrated call sites

- `storage/src/hashers/ethhash.rs` — leaf encoding, 17-element branch
  encoding, single-child "fake root" encoding, and the account
  `storageRoot` splice. The old `replace_hash` helper (which round-tripped
  through `Vec<Vec<u8>>` to mutate one field) is removed entirely; both
  splice sites now call `replace_list_field` directly.

- `storage/src/nodestore/hash.rs::fix_account_storage_root_value` —
  the proof-time storage-root branch encoding swaps
  `RlpStream::new_list(17)` for `encode_list(&[RlpItem::*; 17])`, and
  the splice uses `replace_list_field`.

- `storage/src/lib.rs::format_node_value` — the debug-display path for
  RLP node values switches `Rlp::new(value).as_list::<Vec<u8>>()` (one
  `Vec<u8>` allocation per displayed field) for
  `RlpList::parse(value)?.fields()`.

- `ffi/src/proofs/change.rs` — `Rlp::new(value).at(3).and_then(|r| r.data())`
  becomes `RlpList::parse(value).and_then(|l| l.nth_bytes(3))`. Same
  validation surface, no allocation.

## Dependency cleanup

- `storage/Cargo.toml` — drops `rlp` and `bytes` from `[dependencies]`.
  The `ethhash` feature now only enables `dep:sha3`. `rlp` stays as a
  dev-dep for the parity tests landed in #1957.
- `ffi/Cargo.toml` — drops the optional `rlp` dep and removes it from
  the `ethhash` feature.
- `Cargo.lock` — auto-update.

After this PR, `rlp` and `bytes` are dev-dep only across the workspace;
`firewood`, `firewood-storage`, and `firewood-ffi` all ship without
them. `triehash` keeps its own `rlp` dep (test-only, published
externally — out of scope for this migration).

## Visibility tightening

Once #1957 lands and the migrate commit demonstrates the production
callers, this PR's final commit (*refactor(storage): tighten
`crate::rlp` visibility to crate-internal*) drops the `crate::rlp`
module to `pub(crate)` and re-exports only what's actually used:

- **Always public** (used cross-crate by `firewood-ffi`'s code-hash
  extractor): `NULL_RLP`, `RlpError`, `RlpList` and its `parse` /
  `nth_bytes` methods. Re-exported at `firewood_storage::*`.
- **Crate-internal only** (used by `ethhash`/`nodestore`/
  `format_node_value`): `RlpItem`, `encode_list`, `replace_list_field`.
- **Test-only**: `encode_bytes` moves into `mod tests` since the
  production encoders all flow through `encode_list`.

The benchmark target keeps cross-crate access via a gated re-export
behind the existing `test_utils` feature, matching the
`SeededRng`/`TestRecorder` pattern firewood uses elsewhere:

```rust
#[cfg(any(test, feature = "test_utils"))]
pub use rlp::{RlpItem, encode_list, replace_list_field};
```

The `rlp` benchmark target now requires `test_utils` to compile. To
reproduce the benchmark numbers from #1957:

```
cargo bench -p firewood-storage --bench rlp --features ethhash,logger,test_utils
```

## How this was tested

`cargo nextest run --workspace --features ethhash,logger --all-targets --profile ci`
— 699/699 passing. Account hashing and proof verification remain
bit-compatible with the upstream crate (verified by the parity tests
in #1957 and by the existing `merkle::tests::ethhash` fixtures).

## Breaking changes

None for end users.

Internal API surface changes (workspace-private):
- `firewood_storage::hashers::ethhash::replace_hash` is removed
  (was `pub(crate)`).
- `firewood_storage::rlp` is now `pub(crate)`. External callers of
  `firewood_storage::rlp::RlpList` etc. should use the top-level
  re-export `firewood_storage::RlpList` instead. Inside this PR,
  `ffi/src/proofs/change.rs` is updated accordingly.


storage/src/hashers/ethhash.rs
```diff
@@ -206,70 +200,59 @@ impl<T: Hashable> Preimage for T {
...
-                        rlp.append_empty_data();
-                    }
-                }
+            // For accounts, splice the empty-trie storage root hash into the
+            // account RLP. If the splice fails (malformed value), fall back
+            // to the raw value.
+            let empty_root = Keccak256::digest(NULL_RLP);
+            let account_value: Option<Box<[u8]>> = if is_account {
+                value_bytes
+                    .and_then(|bytes| replace_list_field(bytes, 2, empty_root.as_slice()).ok())
```
AminR443, line 212: perf/nit: do empty root hash computation only if it's an account. Can't we also make this a static const? prevent computing it each time.
```suggestion
            let account_value: Option<Box<[u8]>> = if is_account {
                let empty_root = Keccak256::digest(NULL_RLP);
                value_bytes
                    .and_then(|bytes| replace_list_field(bytes, 2, empty_root.as_slice()).ok())
```

AminR443 (approved):
LGTM.

PR #1960, opened 2026-05-04
perf(storage): reuse persist scratch storage across commits

AminR443:
## Why this should be merged

The background persist worker currently allocates fresh bump storage for each persist cycle. Repeated allocation churn is avoidable because the worker is long-lived and serializes persists on a single thread.

## How this works

- Adds `PersistScratch`, a reusable wrapper around the persistence bump allocator.
- Adds `NodeStore::persist_with_scratch` so callers can reuse scratch storage across commits.
- Updates the deferred persist worker to keep one `PersistScratch` for the lifetime of the background loop.
- Keeps the existing `persist` API unchanged by having it allocate temporary scratch internally.
- Adds a regression test covering reuse of the same scratch value across multiple persists.

## How this was tested

UT.

TBD.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/storage/src/nodestore/persist.rs
+++ b/storage/src/nodestore/persist.rs
@@ -46,6 +46,40 @@ use super::{Committed, NodeStore};
 #[cfg(not(test))]
 use super::RootReader;
 
+/// Reusable scratch state for node persistence.
+///
+/// The background persist worker can hold on to this across commits so `bumpalo`
+/// keeps its chunks instead of deallocating and reallocating them on every
+/// persist.
+#[derive(Debug)]
+pub struct PersistScratch {
+    bump: Bump,
+}
+
+impl PersistScratch {
+    /// Create scratch storage sized for the common node serialization batch.
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            bump: Bump::with_capacity(super::INITIAL_BUMP_SIZE),
+        }
+    }
+
+    const fn bump_mut(&mut self) -> &mut Bump {
+        &mut self.bump
+    }
+
+    fn reset(&mut self) {
+        self.bump.reset();
+    }
+}
+
+impl Default for PersistScratch {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl NodeStoreHeader {
     /// Persist this header to storage.
     ///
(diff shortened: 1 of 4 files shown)


storage/src/nodestore/persist.rs
```diff
@@ -46,6 +46,40 @@ use super::{Committed, NodeStore};
...
 
+/// Reusable scratch state for node persistence.
+///
+/// The background persist worker can hold on to this across commits so `bumpalo`
+/// keeps its chunks instead of deallocating and reallocating them on every
+/// persist.
+#[derive(Debug)]
+pub struct PersistScratch {
+    bump: Bump,
+}
```
rkuris, line 57: Nit
```suggestion
pub struct PersistScratch(Bump);
impl Deref for PersistScratch {
    type Target = Bump;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
```

This is just a newtype wrapper and `bump: Bump` stutters.

storage/src/nodestore/persist.rs
```diff
@@ -46,6 +46,40 @@ use super::{Committed, NodeStore};
...
+
+impl PersistScratch {
+    /// Create scratch storage sized for the common node serialization batch.
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            bump: Bump::with_capacity(super::INITIAL_BUMP_SIZE),
```
rkuris, line 64: nit: move this constant into PersistScratch

storage/src/nodestore/persist.rs
```diff
@@ -46,6 +46,40 @@ use super::{Committed, NodeStore};
...
+    const fn bump_mut(&mut self) -> &mut Bump {
+        &mut self.bump
+    }
+
+    fn reset(&mut self) {
+        self.bump.reset();
+    }
+}
+
+impl Default for PersistScratch {
```
rkuris, line 77: Why would someone need this, and if they do, why wouldn't they allocate the same size as INITIAL_BUMP_SIZE?

rkuris (changes requested):
Preliminary review.

This seems quite complicated and only saves one allocation per persist cycle. I can't imagine this improves performance, but it certainly increases code complexity. You'd have to convince me this is worth it through a noticeable improvement or a counterargument.

AminR443:
## Performance Report

Performance win is tiny (~1%), and only applies to scenario where dp=1. Closing the PR.

PR #1961, opened 2026-05-04
perf(storage): avoid SmallVec for inline RLP hashes

AminR443:
## Why this should be merged

Inline Ethereum RLP hash values are always 1..=31 bytes, but the current representation uses `SmallVec<[u8; 32]>`, which carries unnecessary length/capacity overhead in a hot hashing path.

## How this works

- Adds `RlpBytes`, a compact fixed-buffer representation for inline RLP bytes.
- Changes `HashType::Rlp` under `ethhash` to store `RlpBytes`.
- Streams RLP preimage bytes into a small inline collector and switches to Keccak hashing once the encoded node reaches hash length.

## How this was tested

UT.

Improves performance on read path, but causes a little regression on write path. With deferred persist commit set to 4k, experiences about 4% performance improvement. Will apply Ron's suggestions and mark it R4R.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


storage/src/hashers/ethhash.rs
```diff
@@ -124,6 +124,94 @@ impl HasUpdate for Keccak256 {
     }
 }
 
+struct RlpCollector {
+    len: u8,
+    buf: [u8; RlpBytes::MAX_LEN],
+    hasher: Option<Keccak256>,
+}
```
rkuris: This is a classic tagged enum case:

```suggestion
enum RlpCollector {
    Collecting {
        len: usize,
        buf: [u8; RlpBytes::MAX_LEN],
    },
    Hashing(Keccak256),
}
```

This makes the code a lot easier to read and will probably shorter. Clippy might complain because this has two large variants with different sizes, but that's actually better than what you had before, so I'd #expect that.

Also, there are a lot of conversions from u8 to usize because len is a u8. It probably will be easier to just save the usize -- fewer runtime checks. We have lots of extra bytes anyway now that it's an enum dominated by Keccak256's 300+ bytes. Those conversions aren't free, and storing a usize is just as fast as storing a u8. No alignment issues here fortunately.

rkuris (commented):
Preliminary comments. I think with the enum it might even be faster.

AminR443:
## Performance Report

Improves performance on read path, but causes a little regression on write path. With deferred persist commit set to 4k, experiences about 4% performance improvement. Will apply Ron's suggestions and mark it R4R.

storage/src/hashers/ethhash.rs
AminR443: updated to suggested appraoch. will measure performance and R4R.

AminR443:
After re-measuring this PR against the latest main, it no longer provides the same gains. Changes in #1957 and #1958 optimize RLP along a path that overlaps with this PR, so the additional benefit is now negligible.

PR #1962, opened 2026-05-04
perf(storage): replace node cache with quick_cache

AminR443:
## Why this should be merged

Replaces the memory-sized `lru-mem` node cache with `quick_cache`, removing the global cache mutex from the hot node read/write cache paths and simplifying cache accounting.

## How this works

- Switches node cache sizing from byte limits to entry counts via `node_cache_entries`.
- Replaces `lru-mem` with `quick_cache::sync::Cache`.

## How this was tested

UT.

TBD.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


rkuris (commented):
We can't use a count of entries here, it needs to be a memory size. Why? Some cached entries are bigger and some are smaller. Performance numbers you get here are going to be skewed based on a different cache size, so I'm not sure you can get a meaningful performance number from this diff.

Convince me otherwise.

AminR443:
## Performance Report

Quick cache improves performance between 2%-4% across the replay test, but as mentioned by Ron, we need a cache supporting memory size. Will explore it later (quickcache has weights, and that might be helpful), and maybe benchmark other caches too. Keeping this as draft.

PR #1963, opened 2026-05-04
perf(metrics): cache metric handles after recorder install

AminR443:
## Why this should be merged

Metric recording is on hot storage paths, and repeated recorder lookups add avoidable overhead once the global metrics recorder is installed.

## How this works

- Tracks when the process-global metrics recorder has been installed.
- Caches counter/gauge/histogram handles in per-callsite `OnceLock`s for no-label and literal-label metric callsites.
- Adds explicit cached counter arrays for storage cache and area-indexed metrics where labels are runtime-derived but bounded.
- Marks the recorder as installed from both FFI metrics setup and benchmark Prometheus setup.
- Allows Go metrics tests to skip cleanly in EPERM-restricted environments.

## How this was tested

UT.

TBD.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1773,6 +1773,7 @@ dependencies = [
  "fastrace",
  "fastrace-opentelemetry",
  "firewood",
+ "firewood-metrics",
  "firewood-storage",
  "hex",
  "log",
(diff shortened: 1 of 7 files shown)


rkuris (commented):
Preliminary review.

Not sure I'm crazy about requiring people to call `mark_global_recorder_install`. If they forget, they get the uncached stuff, but they might never know about it unless they run a perf test.

Consider wrapping the install in a helper, maybe `firewood_metrics::install_global_recorder(recorder)` that both installs the recorder and sets the bool to improve the odds someone doesn't mess up here.

PR #1964, opened 2026-05-04
perf(storage): fill predictive read buffer for unaligned reads

AminR443:
## Why this should be merged

PredictiveReader currently limits unaligned reads to the remaining bytes in the current 1 KiB boundary, which can cause smaller-than-necessary pread calls and reduce the benefit of predictive buffering.

## How this works

The file-backed predictive reader now always reads into the full predictive buffer when refilling, regardless of the current file offset. Logical reader offset accounting is unchanged and still advances based on bytes consumed by callers.

## How this was tested

UT.

This is tested with PREDICTIVE_READ_BUFFER_SIZE set to [512, 1024, 2048, 4096], the best option out of those are 1024. So, will just keep it to that, but however, removing the `bytes_left_in_page` logic still helps. So this is not dead.

Performance win in my experiment is around 10-12%, when deferred persist count is set to 4096. 

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


rkuris (commented):
If this provides a meaningful perf win, I'd suggest we change the PREDICTIVE_READ_BUFFER_SIZE to 2048 and see if that improves things.

rkuris:
I think this is dead?

AminR443:
This is tested with PREDICTIVE_READ_BUFFER_SIZE set to [512, 1024, 2048, 4096], the best option out of those are 1024. So, will just keep it to that, but however, removing the `bytes_left_in_page` logic still helps. So this is not dead.

Performance win in my experiment is around 10-12%, when deferred persist count is set to 4096.

rkuris:
Please update the description to include the benchmark results. It currently said the only testing was UT.

rkuris (approved):

demosdemon (approved):

PR #1965, opened 2026-05-04
perf(storage): cache node area index for deletion

AminR443:
## Why this should be merged

This avoids rereading persisted node metadata during deletion when the area index is already known from allocation. Deletion is a storage hot path, so carrying this metadata through the node persistence state removes unnecessary I/O while preserving existing address-based behavior.

## How this works

- Stores the allocated `AreaIndex` inside `MaybePersistedNode` while a node transitions from allocated to persisted.
- Adds internal `storage_info()` access so `NodeAllocator::delete_node` can use the cached area index when available, and fall back to reading from storage for older/address-only nodes.
- Keeps persisted-node equality based on address, so cached metadata does not change comparison semantics.

## How this was tested

UT.

TBD.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/storage/src/node/persist.rs
+++ b/storage/src/node/persist.rs
@@ -4,7 +4,7 @@
 use parking_lot::Mutex;
 use std::{fmt::Display, ops::Deref, sync::Arc};
 
-use crate::{FileIoError, LinearAddress, Node, NodeReader, SharedNode};
+use crate::{AreaIndex, FileIoError, LinearAddress, Node, NodeReader, SharedNode};
 
 /// A node that is either in memory or on disk.
 ///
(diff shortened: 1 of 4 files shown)


storage/src/node/persist.rs
```diff
@@ -225,10 +263,10 @@ impl Display for MaybePersistedNode {
         let guard = self.0.lock();
         match &*guard {
             MaybePersisted::Unpersisted(node) => write!(f, "M{:p}", (*node).as_ptr()),
-            MaybePersisted::Allocated(addr, node) => {
-                write!(f, "A{:p}@{addr}", (*node).as_ptr())
+            MaybePersisted::Allocated { address, node, .. } => {
+                write!(f, "A{:p}@{address}", (*node).as_ptr())
```
rkuris, line 267: Seems like we should dump the `area_index` for completion here

rkuris (commented):
Some preliminary comments

AminR443:
## Performance Report

Performance win is tiny (<1%), Closing the PR.

PR #1966, opened 2026-05-04
refactor(ffi): simplify change proof interface with unified Proposal type

bernard-avalabs:
## Why this should be merged

Closes https://github.com/ava-labs/firewood/issues/1893. The current FFI layer exposes a 3-step state machine (ChangeProofContext → VerifiedChangeProofContext → ProposedChangeProofContext) with a dedicated ProposedChangeProof Go type that duplicates Proposal. Go callers must manage two different proposal-like types depending on whether the proposal came from Database.Propose() or from change proof verification.

This PR unifies the interface: verification returns a standard Proposal, FindNextKey lives on the proof, and CommitWithRebase is a general Proposal capability.

## How this works

  Core library:
  - Db::verify_change_proof() — composes structural validation, batch op application, and root hash verification into a single call returning a standard Proposal
  - Proposal::commit_with_rebase() — commits a proposal, automatically rebasing via DiffMerkleNodeStream if the parent revision is stale. Holds the committed revisions write lock throughout to prevent interleaving.

  FFI layer:
  - fwd_db_verify_change_proof — verifies and returns a ProposalResult (same type as fwd_propose_on_db)
  - fwd_db_verify_and_commit_change_proof — convenience function that verifies and commits in a single FFI call. The proof is borrowed, not consumed. Ported from the rkuris/cp-restructure-squashed branch and adapted to use verify_change_proof + commit_proposal_with_rebase internally. 
  - fwd_change_proof_find_next_key — standalone on ChangeProofContext, takes end_key as parameter
  - fwd_commit_proposal_with_rebase — commits with auto-rebase, returns HashResult

  Go wrappers:
  - Database.VerifyChangeProof() → *Proposal
  - Database.VerifyAndCommitChangeProof() → (Hash, error)    
  - ChangeProof.FindNextKey(endKey) → *NextKeyRange
  - Proposal.CommitWithRebase() → (Hash, error)

  Removed: VerifiedChangeProofContext, ProposedChangeProofContext, VerificationParams, their FFI functions, Go types (VerifiedChangeProof, ProposedChangeProof), result enums, and DatabaseHandle::apply_change_proof_to_parent (dead code after old flow removed).

  Design doc deviations:

  - NextKeyRange.start_key type: Plan calls for Maybe<OwnedBytes>; kept as OwnedBytes since find_next_key always returns either a concrete key (the last batch op's key) or None (no more keys to fetch). The "start of keyspace" case is handled by returning None, not by a Nothing start key.                                         
  - Removed DatabaseHandle::apply_change_proof_to_parent: Not in the plan, but became dead code after VerifiedChangeProofContext::propose() was removed. DatabaseHandle::verify_change_proof replaces it with full      verification.   

  How this was tested

  - CI

  Breaking Changes

  - firewood-ffi (C api)
  - firewood-go (Go api)


ffi/proofs.go
```diff
@@ -418,85 +393,78 @@ func (db *Database) ChangeProof(
...
+	)
 }
 
-// ProposeChangeProof creates a proposal from a VerifiedChangeProof.
-func (db *Database) ProposeChangeProof(
-	proof *VerifiedChangeProof,
-) (*ProposedChangeProof, error) {
+// VerifyAndCommitChangeProof verifies the change proof and commits it in a
+// single call. The proof is not consumed — it remains available for
+// FindNextKey or serialization afterward.
```
alarso16:
```suggestion
// [ChangeProof.FindNextKey] or serialization afterward.
```
I think this fixes the link. You can verify by hovering over the text in your editor

ffi/proposal.go
```diff
@@ -154,6 +156,38 @@ func (p *Proposal) Commit() error {
...
+// CommitWithRebase commits the proposal, automatically rebasing if the
+// proposal's parent is no longer the latest revision. Returns the root hash
+// of the committed revision.
+//
+// This is safe to call only once; subsequent calls will return an error.
+//
+// During this function, the latest state of the database is locked, so other
+// operations that access it (such as [Database.Get] and [Database.Propose]) will
+// block until this function returns.
+func (p *Proposal) CommitWithRebase() (Hash, error) {
```
alarso16, line 171: What happens if we have `db -> p1 -> p2`? Only `p1` is rebased I think, so `p2` would also need to be committed with rebase. This usage is not clear from the doc

alarso16, line 171: You test this function as a part of the change proofs test, but it can be used independently. Can you verify that this works as expected if you create it with the method `Propose`?

ffi/src/proofs/change.rs
```diff
@@ -113,156 +66,39 @@ impl From<FrozenChangeProof> for ChangeProofContext {
...
-    /// Creates a proposal from the verified change proof context, and stores the change
-    /// proof, database handle, and proposal handle in a `ProposedChangeProofContext`.
-    /// Calling `propose` consumes the proof, and calling it again will return a
-    /// `ProofIsNone` error.
-    fn propose<'db>(
-        &'db mut self,
-        db: &'db DatabaseHandle,
-    ) -> Result<ProposedChangeProofContext<'db>, api::Error> {
-        let Some(proof) = self.proof.take() else {
+    /// Returns the next key range to fetch for truncated change proofs,
```
rkuris: This should work for non-truncated change proofs too, no?
```suggestion
    /// Returns the next key range to fetch for a change proof,
```

ffi/src/proofs/change.rs
```diff
@@ -113,156 +66,39 @@ impl From<FrozenChangeProof> for ChangeProofContext {
...
-    end_root: HashKey,
-    end_key: Option<Box<[u8]>>,
-    proposal: Option<crate::ProposalHandle<'db>>,
-}
-
-impl<'db> ProposedChangeProofContext<'db> {
-    fn find_next_key(&mut self) -> Result<Option<KeyRange>, api::Error> {
-        let Some(last_op) = self.proof.batch_ops().last() else {
-            // no BatchOps in the proof, so we are done
+        let Some(last_op) = proof.batch_ops().last() else {
```
rkuris, line 84: This is an interesting and odd case. If you ask for a change proof from A to B and there are NO changed keys in that range, you probably get a valid proof back but no keys. In this case, you don't want to return Ok(None) as you're not done.

ffi/src/proofs/change.rs
```diff
@@ -113,156 +66,39 @@ impl From<FrozenChangeProof> for ChangeProofContext {
...
 
-        if self.proof.end_proof().is_empty() {
-            // unbounded, so we are done
+        if proof.end_proof().is_empty() {
             return Ok(None);
         }
 
-        if let Some(ref end_key) = self.end_key
-            && **last_op.key() >= **end_key
+        if let Some(end_key) = end_key
```
rkuris: Quite confusing. This says "we asked for a proof from A to B but the last key returned is greater than B" which means the proof would fail structural verification. In that case Ok(None) is not the right return value.

ffi/src/proofs/change.rs
```diff
@@ -400,130 +236,100 @@ pub extern "C" fn fwd_db_change_proof(
...
-}
-
-/// Create a proposal from a change proof and return a `ProposedChangeProofResult`.
-///
-/// # Arguments
-///
-/// - `db` - The database to create the proposal.
-/// - `args` - The arguments for verifying the change proof.
+/// Performs structural validation, applies batch ops to the `start_root`
+/// revision, and verifies the root hash against `end_root`. The proof is
```
rkuris: Does it apply the proof to the `start_root` revision or to the latest stored revision? I think the latter.

firewood/src/db.rs
```diff
@@ -301,6 +303,33 @@ impl Db {
...
+        proof: &FrozenChangeProof,
+        start_root: HashKey,
+        end_root: HashKey,
+        start_key: Option<&[u8]>,
+        end_key: Option<&[u8]>,
+        max_length: Option<NonZeroUsize>,
+    ) -> Result<Proposal<'_>, api::Error> {
+        let verification =
+            verify_change_proof_structure(proof, end_root.clone(), start_key, end_key, max_length)?;
+        let parent = self.manager.revision(start_root)?;
```
rkuris: This is incorrect.

The verification model here is unnecessarily restrictive. A change proof from A to B can be applied to either the original hash or any other version that is derived from other change proofs applied to that hash. We can just apply this against the *latest* version and do the verification there.

If you do it this way, it has several problems:
 1 - The older revision may have already expired.
 2 - commit_with_rebase is forced to recreate the revision every time. Applying it to the latest revision makes this an odd case (only when there are multiple outstanding change proofs being applied simultaneously).

rkuris (changes requested):
Good start, but there are a few bugs here.

ffi/proposal.go
rkuris: This is a super interesting observation. If you have r1->p1->p2 and you call CommitWithRebase on p2, it probably just takes the diff from p1->p2 and applies it to the latest version. Assuming this is correct, we should doc this.

firewood/src/db.rs
bernard-avalabs: The implementation follows the design doc (the markdown file in https://github.com/ava-labs/firewood/issues/1893) that retrieves the parent using the start root (`let parent = &self.revision(start_root)?;` in `verify_change_proof`).

I'm happy to change it to use the latest revision.

rkuris: I think the design doc is wrong, but you should convince yourself it's wrong too before changing it. Do you think leaving it like this is better? If so, why?

bernard-avalabs: I don't think the implementation is **incorrect** the way it is. 

Our change proof verification supports your suggested change, which I agree is better for x/sync with change proofs since each truncated change proof touches a non-overlapping range. My comment was an explanation for the current implementation, which is based on your design doc.

As I said previously, I'm happy to make the change, but the implementation was not wrong. It just didn't have this **new** improvement that was not part of your original design.

ffi/proofs.go
bernard-avalabs: Fixed — changed to [ChangeProof.FindNextKey] for proper doc links.

ffi/proposal.go
bernard-avalabs: Added doc comment: "Only this proposal is rebased. Child proposals created via [Proposal.Propose] are not          
automatically reparented after a rebase and should not be used." Added TestCommitWithRebaseChildNotReparented to verify.

bernard-avalabs: Added TestCommitWithRebase (sibling rebase) and TestCommitWithRebaseChildNotReparented as standalone tests not tied to change proofs.

ffi/src/proofs/change.rs
bernard-avalabs: Fixed — "truncated change proofs" → "a change proof".

ffi/src/proofs/change.rs
bernard-avalabs: The logic was unchanged from the version in rkuris/cp-restructure-squashed. I've now changed it to the following: when end_key is set and there are no batch ops, find_next_key returns the next range starting from end_key to continue from end_key. When unbounded, returns Ok(None).

ffi/src/proofs/change.rs
bernard-avalabs: The logic was unchanged from the version in rkuris/cp-restructure-squashed. It now splits the >= check:    last_key == end_key returns Ok(None) (reached the end), last_key > end_key returns Err(EndKeyLessThanLastKey) as a defensive check. Structural validation should prevent this case.

firewood/src/db.rs
bernard-avalabs: Changed to use current_revision() instead of revision(start_root). Removed start_root parameter from verify_change_proof across all layers as it is no longer needed. Also fixed a bug where CommitWithRebase failed for proposals whose parent was the initial empty trie. Added CommittedParentHash enum to distinguish "committed with no hash" from "not committed."

ffi/src/proofs/change.rs
bernard-avalabs: Fixed.

rkuris (approved):

PR #1970, opened 2026-05-05
fix(range-proof): use last_kv_key as right-edge anchor for inclusion-style end_proofs

rkuris:
## Why this should be merged

Valid truncated range proofs fail verification in many cases.

## How this works

Pass the right-edge anchor key (rather than the caller's requested bound) to verify_edge and verify_range_proof_root_hash when the end_proof is an inclusion proof of the last key-value entry. This covers two cases:

  * Truncated proofs, where a max_length was hit before reaching bound.
  * Complete proofs whose bound exists in the trie (bound == last_kv_key).

In both cases the end_proof's terminal node has a value digest and its full path equals nibbles(last_kv_key); using bound would walk a different descent path and trip ShouldBePrefixOfNextKey from compute_outside_children. The exclusion case (bound not in trie) keeps using bound as the anchor.

## How this was tested

Adds a regression test that constructs a truncated bounded range proof and verifies it against the originally requested (start, end). Also adds a custom Debug impl on Proof that indexes proof nodes, since this was needed to diagnose the bug; ChangeProof and RangeProof get matching manual Debug impls because their derived Debug bound on inner Proof<H> would otherwise be unsatisfiable.

## Breaking Changes

None

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -417,6 +417,38 @@ pub fn verify_range_proof<H: ProofCollection<Node = ProofNode>>(
     let left_edge = left_edge_key.map(|(k, v)| (k.as_ref(), v.as_ref()));
     let right_edge = right_edge_key.map(|(k, v)| (k.as_ref(), v.as_ref()));
 
+    // Determine the right-edge anchor key.
+    //
+    // The end_proof is one of two shapes:
+    //
+    //   * Inclusion proof of `last_kv_key`. This happens both when the proof
+    //     is truncated (a `max_length` was hit before reaching `bound`) and
+    //     when `bound` itself exists in the trie (so `bound == last_kv_key`).
+    //     In both cases the proof anchors at `last_kv_key` and must be
+    //     verified against it; using `bound` would walk a different path.
+    //
+    //   * Exclusion proof of some key past `last_kv_key` — produced for a
+    //     complete proof of `[start, bound]` when `bound` is not in the trie.
+    //     The proof anchors at `bound` (the requested upper boundary).
+    //
+    // We detect inclusion-of-last-kv by examining the terminal node: if it has
+    // a value digest and its full nibble path equals `last_kv_key`'s nibbles,
+    // it is the inclusion case and the anchor must be `last_kv_key`.
+    let right_edge_anchor: Option<&[u8]> = if let (Some(_bound), Some((last_kv_k, _))) =
+        (last_key_bytes, key_values.last())
+        && let Some(terminal) = proof.end_proof().as_ref().last()
+        && terminal.value_digest.is_some()
+        && terminal
+            .key
+            .iter()
+            .map(|c| c.as_u8())
+            .eq(NibblesIterator::new(last_kv_k.as_ref()))
+    {
+        Some(last_kv_k.as_ref())
+    } else {
+        last_key_bytes
+    };
+
     if !proof.start_proof().is_empty() {
         verify_edge(
             first_key_bytes,
@@ -428,7 +460,7 @@ pub fn verify_range_proof<H: ProofCollection<Node = ProofNode>>(
     }
     if !proof.end_proof().is_empty() {
         verify_edge(
-            last_key_bytes,
+            right_edge_anchor,
             right_edge,
             proof.end_proof(),
             root_hash,
@@ -455,7 +487,7 @@ pub fn verify_range_proof<H: ProofCollection<Node = ProofNode>>(
         key_values,
         proof,
         first_key_bytes,
-        last_key_bytes,
+        right_edge_anchor,
         root_hash,
     )
 }
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -496,10 +496,11 @@ pub(crate) fn reject_odd_nibble_value_digests(proof_nodes: &[ProofNode]) -> Resu
 ///   `Inclusive(Some(last_kv))`. The end-proof anchors directly at
 ///   `last_kv`. Covered by
 ///   `test_dropped_trailing_key_accepted_as_partial_coverage`.
-/// * **Everything else** — terminal is a strict prefix of `last_kv` with
-///   an on-path child set, terminal has no value digest, `end_proof`
-///   empty, or `key_values` empty — → `Inclusive(fallback_last)`,
-///   anchored at the caller's requested bound. Covered by
+/// * **Everything else** — terminal value-node with `K < last_kv` (whether
+///   `K` is a strict prefix of `last_kv` or a divergent lex-less node),
+///   terminal has no value digest, `end_proof` empty, or `key_values`
+///   empty — → `Inclusive(fallback_last)`, anchored at the caller's
+///   requested bound. Covered by
 ///   `test_terminal_strict_prefix_of_last_kv_verifies` and
 ///   `test_tampered_in_range_value_rejected`.
 ///
@@ -551,8 +552,8 @@ pub(crate) fn right_edge<'a>(
     };
 
     // Terminal anchors at a real byte key only when it carries a value
-    // digest. (Callers reject end_proof terminals whose nibble path is
-    // odd-length but carry a value before reaching here — see
+    // digest. (Callers reject *any* end_proof node whose nibble path is
+    // odd-length but carries a value before reaching here — see
     // [`reject_odd_nibble_value_digests`] — so we don't re-check parity.)
     if terminal.value_digest.is_some() {
         let nibs: Vec<u8> = terminal.key.iter().map(|c| c.as_u8()).collect();
@@ -583,10 +584,11 @@ pub(crate) fn right_edge<'a>(
 ///    `last_key`; partial coverage is a valid outcome, not an error.
 /// 3. **Boundary proof verification**: Cryptographically verify the start
 ///    proof against `first_key`. The right edge is verified against the
-///    *proof-derived* anchor — which may be `last_kv` or the caller's
-///    `last_key`, depending on whether the right boundary is inclusive
-///    or exclusive. For the exclusive case only, the cryptographic check
-///    is folded into the hash reconstruction below instead.
+///    *proof-derived* right boundary — for the inclusive case via
+///    `verify_edge` against either `last_kv` or the caller's `last_key`
+///    (whichever the proof anchors at), and for the exclusive case via
+///    the hash reconstruction below, where the anchor is the end-proof
+///    terminal's full key.
 /// 4. **Trie reconstruction**: Build an in-memory Merkle trie from the key-value pairs
 ///    and reconcile it with proof nodes
 /// 5. **Root hash comparison**: Verify the reconstructed trie's root hash matches the
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -187,27 +187,32 @@ pub(crate) enum EdgeBoundary<'a> {
 
 /// The right edge of a range proof.
 ///
-/// `Exclusive` holds a `Cow` because the only path that produces it —
+/// `OutOfRange` holds a `Cow` because the only path that produces it —
 /// `right_edge` for a terminal value-node whose full key exceeds
 /// `last_kv` — derives the boundary by reconstructing the terminal's
-/// full byte key from its nibble path, which is owned. `Inclusive`
+/// full byte key from its nibble path, which is owned. `InRange`
 /// always borrows from caller-supplied slices.
+///
+/// Note: "in-range" / "out-of-range" describes whether the boundary
+/// key itself belongs to the proven range. This is independent from
+/// the inclusion-proof vs. exclusion-proof distinction at the proof
+/// layer (which is about whether a key exists in the trie).
 #[derive(Debug)]
 pub(crate) enum RightBoundary<'a> {
-    /// Inclusive at `end_key` (or no upper bound if `None`). `end_key`
-    /// itself is part of the proven range.
-    Inclusive(Option<&'a [u8]>),
-    /// Exclusive at `end_key`. `end_key` always exists for this variant
-    /// — it sits at the end-proof's terminal node — and is *not* part
-    /// of the proven range.
-    Exclusive(Cow<'a, [u8]>),
+    /// The boundary key (or no upper bound if `None`) is part of the
+    /// proven range — the right end is "closed" at `end_key`.
+    InRange(Option<&'a [u8]>),
+    /// The boundary key is *not* part of the proven range — the right
+    /// end is "open" at `end_key`. `end_key` always exists for this
+    /// variant; it sits at the end-proof's terminal node.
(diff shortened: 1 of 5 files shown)


firewood/src/merkle/mod.rs
```diff
@@ -437,6 +437,38 @@ pub fn verify_range_proof<H: ProofCollection<Node = ProofNode>>(
...
+    //
+    //   * Exclusion proof of some key past `last_kv_key` — produced for a
+    //     complete proof of `[start, bound]` when `bound` is not in the trie.
+    //     The proof anchors at `bound` (the requested upper boundary).
+    //
+    // We detect inclusion-of-last-kv by examining the terminal node: if it has
+    // a value digest and its full nibble path equals `last_kv_key`'s nibbles,
+    // it is the inclusion case and the anchor must be `last_kv_key`.
+    let right_edge_anchor: Option<&[u8]> = if let (Some(_bound), Some((last_kv_k, _))) =
+        (last_key_bytes, key_values.last())
```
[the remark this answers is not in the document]
rkuris: I don't think we actually want this, but I'm open for a second opinion. I would rather have the change proof stand by itself rather than relying on what was requested, except that if a change proof contains keys outside of what was requested it's clearly malicious. An exclusion proof of the right edge provided says "we know there are no keys all the way to here from the last one provided" which is sufficient.

One could argue that if there is an exclusion proof provided, it should be of the requested right edge, but I think it's okay for the generator to have generated an edge proof beyond the last key provided; it doesn't change correctness.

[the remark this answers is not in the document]
bernard-avalabs: I wasn't sure if the right-edge anchor heuristic covers all of the cases so I wrote a TLA+ model to verify it. TLC found a counterexample where the heuristic picks the wrong anchor.

The heuristic can't distinguish "inclusion proof of last_kv_key" from "exclusion proof of bound that terminates at  last_kv_key's node." This happens when bound doesn't exist in the trie and the exclusion proof's terminal node has a value matching last_kv_key. For example, when last_kv_key is an ancestor or divergent neighbor of bound.

I constructed an attack that exploits this: keys \x05, \x10, \x10\x10, \x10\x50. Range proof for [\x05, \x10\x30]. The end_proof is an exclusion proof of \x10\x30, but the heuristic picks \x10 as anchor. An attacker removes \x10\x10 from key_values. The edge check only verifies inclusion of \x10 and never notices the omitted key. Below is the test: it currently fails, meaning verification passes for the tampered proof.

```rust
#[test]
// Attack: remove an in-range key (\x10\x10) from key_values. The heuristic
// picks \x10 as anchor (wrong — should be \x10\x30). The edge check only
// verifies \x10 (inclusion) and never checks that \x10\x10 should be
// covered by the proof range up to \x10\x30.
fn test_wrong_anchor_omitted_key_attack() {
    let items: Vec<(&[u8], &[u8])> = vec![
        (b"\x05", b"before"),
        (b"\x10", b"parent"),
        (b"\x10\x10", b"in_range"),
        (b"\x10\x50", b"child"),
    ];
    let merkle = init_merkle(items.into_iter());
    let root_hash = merkle.nodestore().root_hash().unwrap();

    let honest = merkle
        .range_proof(Some(b"\x05".as_slice()), Some(b"\x10\x30".as_slice()), None)
        .unwrap();

    // Honest proof should include \x05, \x10, \x10\x10
    assert_eq!(honest.key_values().len(), 3);

    // Remove \x10\x10 — last_kv_key stays \x10
    let tampered_kvs: Vec<(Box<[u8]>, Box<[u8]>)> = honest
        .key_values()
        .iter()
        .filter(|(k, _)| k.as_ref() != b"\x10\x10")
        .map(|(k, v)| (k.as_ref().into(), v.as_ref().into()))
        .collect();
    assert_eq!(tampered_kvs.len(), 2);

    let tampered = RangeProof::new(
        crate::Proof::<Box<[ProofNode]>>::new(honest.start_proof().as_ref().into()),
        crate::Proof::<Box<[ProofNode]>>::new(honest.end_proof().as_ref().into()),
        tampered_kvs.into_boxed_slice(),
    );

    let result = verify_range_proof(
        Some(b"\x05".as_slice()),
        Some(b"\x10\x30".as_slice()),
        &root_hash,
        &tampered,
    );
    assert!(
        result.is_err(),
        "omitted key \\x10\\x10 should be rejected"
    );
}
```

bernard-avalabs (changes requested):
I believe I found a case where the right-edge anchor heuristic is incorrect, which can allow a tampered proof to be accepted. Please see my comment for the test case.

firewood/src/merkle/mod.rs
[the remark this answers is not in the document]
rkuris: Good catch. I think this also means that we have to verify that the exclusion proof on the right edge actually proves the non-existence of the requested right edge, and I'm not sure we're doing that either.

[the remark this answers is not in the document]
rkuris: Interestingly, there's no reason why this proof isn't a valid truncated proof, so I'm not sure this is actually a bug any more. We should discuss it.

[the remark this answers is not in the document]
bernard-avalabs: I think it is problematic that the caller doesn't know from the proof that the gap between \x10 and \x10\x30 is not included.

firewood/src/merkle/mod.rs
```diff
@@ -335,30 +434,165 @@ fn compute_root_hash_with_proofs(
...
-/// 2. **Boundary proof verification**: Cryptographically verify start/end proofs against
-///    the provided root hash
-/// 3. **Trie reconstruction**: Build an in-memory Merkle trie from the key-value pairs
+/// 2. **Right-edge anchor derivation**: Determine, from the proof's structure,
+///    the actual bound the end-proof anchors at. This may be smaller than
+///    `last_key`; partial coverage is a valid outcome, not an error.
+/// 3. **Boundary proof verification**: Cryptographically verify the start
+///    proof against `first_key`. The right edge is *not* re-verified against
+///    the caller's `last_key`; its cryptographic check is folded into the
+///    hash reconstruction below via the proof-derived anchor.
```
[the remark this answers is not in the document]
rkuris: Fixed in dbf3026c7feb

firewood/src/merkle/mod.rs
```diff
@@ -335,30 +434,165 @@ fn compute_root_hash_with_proofs(
...
 
+/// Reject any proof node that carries a `Value` digest at an odd-length
+/// nibble path. Byte keys are always even-nibble, so a value at an
+/// odd-length path can't correspond to any real key in the trie. This is
+/// a cheap structural sanity check that callers run on a proof's nodes
+/// before any anchoring decision is made — see e.g. the call before
+/// [`right_edge`] in `verify_range_proof`.
+pub(crate) fn reject_odd_nibble_value_digests(proof_nodes: &[ProofNode]) -> Result<(), ProofError> {
+    for node in proof_nodes {
+        if !node.key.len().is_multiple_of(2) && node.value_digest.is_some() {
```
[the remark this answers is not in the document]
rkuris: Fixed in dbf3026c7feb

firewood/src/proofs/change.rs
```diff
@@ -15,13 +15,28 @@ use crate::{
 /// - A start proof: proves that the smallest key does/doesn't exist
 /// - An end proof: proves the the largest key does/doesn't exist
```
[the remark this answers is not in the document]
rkuris: Fixed in dbf3026c7feb

firewood/src/merkle/mod.rs
```diff
@@ -335,30 +434,171 @@ fn compute_root_hash_with_proofs(
...
 
+/// Reject any proof node that carries a value digest (either
+/// [`ValueDigest::Value`] or [`ValueDigest::Hash`]) at an odd-length
+/// nibble path. Byte keys are always even-nibble, so a value at an
+/// odd-length path can't correspond to any real key in the trie — this
+/// is the same invariant [`Proof::value_digest`] enforces while walking
+/// a proof. This is a cheap structural sanity check that callers run on
+/// a proof's nodes before any anchoring decision is made — see e.g. the
+/// call before [`right_edge`] in `verify_range_proof`.
+pub(crate) fn reject_odd_nibble_value_digests(proof_nodes: &[ProofNode]) -> Result<(), ProofError> {
```
demosdemon: This is `pub(crate)` but is currently only used within this module. Let's remove the visibility unless you are going to use it in a future change.

firewood/src/merkle/mod.rs
```diff
@@ -335,30 +434,171 @@ fn compute_root_hash_with_proofs(
...
+///
+/// There is no `LeftExclusive` counterpart to `RightBoundary::Exclusive`.
+/// The mirror construction (`terminal_full_key < first_kv` via a divergent
+/// leaf in `prove(first_key)`) is empirically handled by the existing
+/// [`EdgeBoundary::Left`] path; see `test_divergent_terminal_before_first_kv`.
+/// The asymmetry holds because the in-range data on the left side is, by
+/// construction, present in `key_values`, so there is no off-path single-node
+/// subtree past `first_kv` that could synthesize incorrectly during hash
+/// reconstruction.
+pub(crate) fn right_edge<'a>(
```
demosdemon: same here, this is also only used within this module.

firewood/src/merkle/mod.rs
```diff
@@ -335,30 +434,171 @@ fn compute_root_hash_with_proofs(
...
+///
+/// In the inclusive-at-`last_kv` shape (case 2), the proven range may be
+/// strictly narrower than what the caller requested — the dropped-trailing-
+/// key scenario (`test_dropped_trailing_key_accepted_as_partial_coverage`)
+/// is the canonical case, since the attacker shrunk it on purpose. This is
+/// a *valid* outcome, not a verification error. The caller is expected to
+/// compare `key_values.last()` against their requested `last_key`; if the
+/// proven range is narrower, they re-request `(last_kv, last_key]`. No data
+/// is hidden — at most one extra round trip is induced, and the attacker
+/// gains nothing.
```
demosdemon, line 522: This section, and all of the docs on this method, will not surface on public docs. A section about partial coverage would be helpful on `verify_range_proof` and should explain what to do (e.g., compare `key_values.last()` against what they know to be the last requested key).

firewood/src/proofs/range.rs
```diff
@@ -76,13 +76,29 @@ use super::types::{Proof, ProofCollection};
...
     start_proof: Proof<H>,
     end_proof: Proof<H>,
     key_values: Box<[(K, V)]>,
 }
 
+impl<K, V, H> std::fmt::Debug for RangeProof<K, V, H>
+where
+    K: std::fmt::Debug,
+    V: std::fmt::Debug,
+    H: crate::proofs::types::ProofCollection,
```
demosdemon: nit: 

```suggestion
    H: ProofCollection,
```

ProofCollection is already imported.

demosdemon (approved):

firewood/src/merkle/mod.rs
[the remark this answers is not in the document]
rkuris: I think we talked about this in person, but i'm not sure. A hostile generator could generate proofs of length 1 and we can't tell we might falsely think there are no more keys in this range. To do this, though, the end key would have to be longer than the length of last_key. For a syncer, this is not going to happen.

firewood/src/merkle/mod.rs
bernard-avalabs, line 628: Should we consider returning the RightBoundary so it can be used later in `find_next_key`?

bernard-avalabs (approved):
LGTM. One minor concern is the naming of the RightBoundary cases. Inclusive and Exclusive are similar to Inclusion and Exclusion, but they don't have the same meaning here. There is also no test for the edge case where the terminal node for an exclusion proof has a value. Maybe something like this can be added:

```rust
#[test]
// Edge case: the end_proof is an exclusion proof whose terminal node has a
// value. The trie has keys \x05 and \x10\x30. A range proof for
// [\x05, \x10\x35] returns both keys. The end_proof for \x10\x35 diverges
// at \x10\x30's node (compressed path <3,0> disagrees with <3,5>). The
// terminal node holds \x10\x30's value despite being an exclusion proof —
// the value belongs to the divergent node, not to the target key \x10\x35.
fn test_exclusion_divergent_landing_on_last_kv() {
    let items: &[(&[u8], &[u8])] = &[
        (b"\x05", b"first"),
        (b"\x10\x30", b"last_in_range"),
    ];
    let merkle = init_merkle(items.iter().copied());
    let root_hash = merkle.nodestore().root_hash().unwrap();

    let range_proof = merkle
        .range_proof(Some(b"\x05".as_slice()), Some(b"\x10\x35".as_slice()), None)
        .unwrap();

    // Both keys are in range and returned (exhaustive).
    assert_eq!(range_proof.key_values().len(), 2);
    assert_eq!(range_proof.key_values().last().unwrap().0.as_ref(), b"\x10\x30");

    // Verification passes.
    verify_range_proof(
        Some(b"\x05".as_slice()),
        Some(b"\x10\x35".as_slice()),
        &root_hash,
        &range_proof,
    )
    .unwrap();

    // The end_proof terminal is at \x10\x30 (divergent from \x10\x35),
    // and it has a value — an exclusion proof whose terminal holds a value.
    let terminal = range_proof.end_proof().as_ref().last().unwrap();
    assert!(
        terminal.value_digest.is_some(),
        "terminal node should have a value (exclusion-divergent at a valued node)"
    );

    // right_edge sees: terminal has value, terminal_key == last_kv.
    // Returns Inclusive(last_kv), not Inclusive(e).
    let last_kv = range_proof.key_values().last().map(|(k, _)| k.as_ref());
    let boundary = right_edge(
        range_proof.end_proof().as_ref(),
        last_kv,
        Some(b"\x10\x35".as_slice()),
    );
    assert!(
        matches!(boundary, RightBoundary::Inclusive(Some(k)) if k == b"\x10\x30"),
        "expected Inclusive(\\x10\\x30) but got {boundary:?}"
    );
}
```

firewood/src/merkle/mod.rs
rkuris: Let's leave that for a change for the find_next_key PR.

firewood/src/merkle/mod.rs
rkuris: fixed in d28f4a1e

firewood/src/merkle/mod.rs
rkuris: fixed in d28f4a1e

firewood/src/merkle/mod.rs
rkuris: fixed in d28f4a1e

firewood/src/proofs/range.rs
rkuris: fixed in d28f4a1e

rkuris:
Renamed Inclusive/Exclusive to InRange/OutOfRange in d28f4a1e per your suggestion.

test_exclusion_divergent_landing_on_last_kv is a good idea, let's do this as a separate PR.

PR #1971, opened 2026-05-05
feat(ffi): implement range proof verification (Trail of Bits finding TOB-FIREWOOD-2)

rkuris:
## Why this should be merged

Replace the stub `RangeProofContext::verify` in the FFI with a real call to `firewood::verify_range_proof` plus a `max_length` check using `ProofError::ProofIsLargerThanMaxLength`. Removes the `warn!("range proof verification not yet implemented")` log line.

Re-enables the wrong-root assertion in `TestRangeProofVerify` (drops the `TODO(#738)` workaround that swallowed the verification failure).

Adjacent supporting changes:

  * firewood/src/merkle/mod.rs: accept an empty `end_proof` when the caller passes `last_key = None`. Honest unbounded full-range proofs produced by `Merkle::range_proof` carry no end_proof by construction; the previous check rejected every such proof with `NoEndProof`. The root-hash reconstruction at the end of `verify_range_proof` is the safety net against tampering. Includes `test_full_range_proof_round_trip` which round-trips an unbounded full-range proof through generation and verification.

  * firewood/src/api.rs: change `#[error("proof error")]` to `#[error("proof error: {0}")]` so the inner `ProofError` variant surfaces in error messages instead of being swallowed.

## How this was tested

CI tests all pass

## Breaking Changes

None

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -417,6 +417,38 @@ pub fn verify_range_proof<H: ProofCollection<Node = ProofNode>>(
     let left_edge = left_edge_key.map(|(k, v)| (k.as_ref(), v.as_ref()));
     let right_edge = right_edge_key.map(|(k, v)| (k.as_ref(), v.as_ref()));
 
+    // Determine the right-edge anchor key.
+    //
+    // The end_proof is one of two shapes:
+    //
+    //   * Inclusion proof of `last_kv_key`. This happens both when the proof
+    //     is truncated (a `max_length` was hit before reaching `bound`) and
+    //     when `bound` itself exists in the trie (so `bound == last_kv_key`).
+    //     In both cases the proof anchors at `last_kv_key` and must be
+    //     verified against it; using `bound` would walk a different path.
+    //
+    //   * Exclusion proof of some key past `last_kv_key` — produced for a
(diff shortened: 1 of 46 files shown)


firewood/src/merkle/mod.rs
AminR443: Is it possible to have a truncated proof that leads to a case that end proof is present, but last_key_bytes is None? This can lead to wrong right_edge_anchor. Need to take a deeper look though.

I wrote a failing test but not really sure if this is a real bug ([amin/ffi-range-proof-test](https://github.com/ava-labs/firewood/tree/amin/ffi-range-proof-test)):
https://github.com/ava-labs/firewood/blob/3e7024077ca5397f58875e0bd37d679f02573ae9/firewood/src/merkle/tests/range.rs#L277-L318

AminR443 (changes requested):

firewood/src/merkle/mod.rs
rkuris: This shouldn't be possible and should be rejected in the structural checks. If this is true though, this is a separate bug (that is, nothing to do with this PR). Please click the "reference in new issue" button if you think there's something here worth investigating further.

AminR443 (approved):

demosdemon (approved):

PR #1983, opened 2026-05-09
chore: remove weekly 50m-60m job

### LINKED ISSUE #1980: Weekly Firewood reexecution test is broken
In https://ava-labs.github.io/firewood/bench/, we have a weekly test that is supposed to reexecute blocks 50-60m. However, the last (successful?) test run of this was back in February. Given that this has been broken for over 3 months now and that no one has been relying on this, the best course of action here would be to remove this scheduled test.

RodrigoVillar:
## Why this should be merged

Closes #1980.

While investigating https://ava-labs.github.io/firewood/bench/, I noticed that the 50m-60m reexecution benchmark has only one published data point. The scheduled Firewood workflow for this benchmark has not produced usable published results for roughly three months now. We could try to fix it, but given that there's no evidence anyone has been relying on this signal, it's clear that this test is just a broken, unowned benchmark. I've therefore opted to remove it.

## How this works

Removes the weekly 50m-60m reexecution test from the list of reexecution cron jobs.

## How this was tested

CI

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/.github/workflows/track-performance.yml
+++ b/.github/workflows/track-performance.yml
@@ -7,8 +7,6 @@ on:
     # These cron expressions are matched exactly in configure-benchmark to set parameters
     # Weekdays at 05:00 UTC (00:00 ET) - 1M blocks: 40M → 41M
     - cron: '0 5 * * 1-5'
-    # Weekly Saturday at 05:00 UTC (00:00 ET) - 10M blocks: 50M → 60M
-    - cron: '0 5 * * 6'
   workflow_dispatch:
     inputs:
       firewood:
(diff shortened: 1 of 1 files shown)


rkuris (approved):
🏆 

Presumably the weekdays one has the same problem -- when it fails nothing is notified of the failure. Let's add an issue that says "report results to slack" or something so we can better monitor it.

RodrigoVillar:
@rkuris every weekday, we're notified in the Firewood notifications channel if the benchmark results for the day are posted. From that, absence of these results on any given weekday can be a clue that something has gone wrong.

Issue #1985, opened 2026-05-09
Concurrent commit_with_rebase produces freelist/index corruption


rkuris:
## Summary

Multiple threads each calling `Proposal::commit_with_rebase` concurrently against the same database produce on-disk corruption: `InvalidKey` errors throughout the trie and freelist `AreaLeaks`. In some downstream reports, this also surfaces as "Invalid \`FreeArea\` marker" errors at close.

## How to reproduce

`firewood/src/db/tests/concurrent_rebase.rs::test_concurrent_commit_with_rebase_no_freelist_corruption`. Eight threads each propose a disjoint 16-key batch and call `commit_with_rebase`. With no external mutex serializing the commits:

- The trie checker reports >100 `InvalidKey` errors (stored node keys don't match the index path that points to them).
- `AreaLeaks` are present.
- The corruption persists across reopen.

## Why this matters

`commit_with_rebase` is the documented API for proposing-and-committing under contention. Its contract is that only one commit can win directly and the others rebase onto the winner — the result should be equivalent to running the same commits sequentially in some order. Today that contract is violated under concurrency: callers either need to serialize externally (defeating the API's purpose) or end up with a corrupted database.

## Hypothesis

`commit_with_rebase` (`firewood/src/db.rs:525`) takes the `in_memory_revisions` write lock for the rebase + commit critical section, which should serialize commits. But stale proposals built against a parent that has since been committed/reaped may carry allocator state from that old parent; when re-proposed under the new current via `propose_with_parent`, the diff stream + rebuild path may leak old allocations into the new state. Worth investigating in conjunction with the open PRs touching this area (#1981 commit_cleanup TOCTOU, #1970 range-proof anchor) — either or both might shift the failure mode.

## Tests

`firewood/src/db/tests/concurrent_rebase.rs` — currently `#[ignore]`'d pending fix.

PR #1986, opened 2026-05-09
test: pin concurrent commit_with_rebase corruption (issue #1985)

rkuris:
## Why this should be merged

Pins #1985 (concurrent `commit_with_rebase` corruption) as an `#[ignore]`'d regression test so the failure mode is captured in the test corpus while we work on the fix.

Without an external mutex, eight threads each calling `commit_with_rebase` with disjoint batches produce >100 `InvalidKey` errors and freelist `AreaLeaks`. Corruption persists across reopen.

(An earlier version of this PR also pinned what looked like a verification gap in `verify_change_proof`. On further analysis the verifier's contract is "in-range slice matches end_root" — analogous to range-proof verification — and the apparent gap was the verifier being asked for a stronger guarantee than it promises. Those tests were removed from this PR.)

## How this works

`firewood/src/db/tests/concurrent_rebase.rs::test_concurrent_commit_with_rebase_no_freelist_corruption`:

- Eight threads, each proposes a 16-key disjoint batch and calls `commit_with_rebase`.
- After all join: assert all keys present in the latest revision.
- Run `check()` with `hash_check=true`; expect no errors (excluding `AreaLeaks` for now).
- Reopen the DB; re-check; same assertions.

Currently `#[ignore]`'d under #1985 pending fix.

## How this was tested

- `cargo nextest run --workspace --features ethhash,logger --all-targets` → 664 passed, 13 skipped (the new test + 12 pre-existing).
- `cargo nextest run -p firewood --features ethhash,logger --run-ignored only concurrent_rebase` → reproduces the bug.
- `cargo clippy --workspace --features ethhash,logger --all-targets` → clean.

## Breaking Changes

None.

### DIFF
--- a/firewood/src/db/tests.rs
+++ b/firewood/src/db/tests.rs
@@ -3,6 +3,7 @@
 
 #![expect(clippy::unwrap_used)]
 
+mod concurrent_rebase;
 mod merge;
 
 use super::test::TestDb;
--- a/firewood/src/db/tests/concurrent_rebase.rs
+++ b/firewood/src/db/tests/concurrent_rebase.rs
@@ -1,0 +1,143 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! Concurrent `commit_with_rebase` correctness.
+//!
+//! Multiple threads each propose a disjoint batch off the same parent and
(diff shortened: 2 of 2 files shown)


rkuris:
There's quite a bit wrong with this (too much artificial, not enough intelligence)

PR #1987, opened 2026-05-10
fix(manager): use CommittedId for parent identity, skip trivial commits

rkuris:
## Why this should be merged

A `commit` on a proposal that is hash-equivalent with the current root_hash corrupts
the database. Assume R1, R2, R3 have H(R1) == H(R3), and P1 is based off R1. Calling
`commit` on this proposal passes the simple hash commit check, therefore the nodes
are written as-is to the database. However, the database may have been restructured
by R2, so writing the merkle will result in an inconsistent database.

A more trivial instance of this was hit by the current avalanchego syncer which uses
`commit_with_rebase` resulting in a similar issue.

Two related bugs in revision-manager parent identity:

1. **Hash-only parent check is too weak.** A trivial commit (proposal whose
   root hash equals the current revision's) was pushed as a new revision
   sharing the same root address. A subsequent proposal built off the
   original parent then satisfied the hash-based parent check against the
   trivial sibling — letting two siblings of the same parent both commit,
   each adding the parent's root address to its deleted list. On reap,
   `delete_node` ran twice on that address and produced a self-cycle in
   the freelist.

2. **A->B->A round trips alias by hash alone.** Two committed revisions
   can share a root hash but always have distinct identities (and distinct
   addresses). A proposal built off the first A must not be commitable
   against the second A — its deleted list still references the first A's
   nodes, which may already have been freed and reallocated.

## How this works

- Introduces `CommittedId` in `firewood-storage`: a process-monotonic
  identifier assigned synchronously to every `Committed` revision via a
  global `AtomicU64`. Unlike root address (assigned async by the persist
  worker), `CommittedId` does not race with proposal creation.
- `NodeStoreParent::Committed` now stores `{ hash, id }`. `parent_id_is`
  checks identity by id alone; the hash is retained for the rebase path,
  which still looks up the old parent revision by hash. The id alone is
  unique; hash equality is not.
- Adds a trivial-commit fast path to
  `RevisionManager::commit_critical_section`: if `proposal.root_hash() ==
  current.root_hash()` after the parent check, return without pushing a
  new revision. Children of the trivial proposal reparent onto the current
  revision (logically equivalent).
- `commit_critical_section` now returns the resulting `CommittedId`, which
  `commit_cleanup` threads into `commit_reparent` so descendant proposals
  get the right id.

## How this was tested

Four regression tests in `firewood/src/db/tests/concurrent_rebase.rs`:

- `test_concurrent_commit_with_rebase_no_freelist_corruption` — eight
  threads, each running 32 rounds of `commit_with_rebase` with a disjoint
  16-key batch. Verifies no lost updates, no checker errors, and a clean
  freelist after reopen.
- `test_concurrent_rebase_overlapping_paths` — eight threads barrier-sync
  to build proposals off the *same* parent and commit them concurrently
  with random keys whose tries share intermediate branches. `max_revisions`
  is small so reaping fires during the concurrent phase.
- `test_trivial_rebase_succeeds_without_corruption` — two proposals making
  the same `Put` race to commit. Verifies the trivial-commit fast path
  absorbs the second commit without pushing a duplicate revision.
- `test_round_trip_rejects_stale_parent` — `Put → Delete → Put` reaches the
  same root hash twice with different `CommittedId`s. Verifies a proposal
  pinned to the first "A" rebases rather than committing directly against
  the second "A".

Plus:

- Full workspace test suite (`cargo nextest run --workspace --features
  ethhash,logger --all-targets`) passes.
- `cargo clippy --workspace --features ethhash,logger --all-targets`:
  clean for new code.
- `cargo doc --no-deps`: builds.

## Breaking Changes

- [ ] firewood
- [x] firewood-storage — `NodeStoreParent::Committed` variant shape
      changed; `Committed` gained an `id` field; `parent_hash_is` renamed
      to `parent_id_is(id)`; `commit_reparent` now takes a `CommittedId`.
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -529,7 +529,7 @@ impl Proposal<'_> {
         let mut revisions_guard = db.manager.lock_committed_revisions();
         let current = revisions_guard.back().expect("always one revision");
 
-        if !nodestore.parent_hash_is(current.root_hash()) {
+        if !nodestore.parent_id_is(current.committed_id()) {
             // Parent is stale — rebase onto the current revision.
             let old_parent = match nodestore.committed_parent_hash() {
                 CommittedParentHash::NotCommitted => {
@@ -555,10 +555,11 @@ impl Proposal<'_> {
         }
 
         let hash = api::DbView::root_hash(&*nodestore);
-        db.manager
+        let new_id = db
+            .manager
             .commit_critical_section(&nodestore, &mut revisions_guard)?;
         drop(revisions_guard);
-        db.manager.commit_cleanup(&nodestore);
+        db.manager.commit_cleanup(&nodestore, new_id);
         Ok(hash)
     }
 
--- a/firewood/src/db.rs
+++ b/firewood/src/db.rs
@@ -535,18 +535,23 @@ impl Proposal<'_> {
                 CommittedParentHash::NotCommitted => {
                     return Err(api::Error::ParentNotCommitted);
                 }
-                CommittedParentHash::Hash(Some(hash)) => db.manager.revision(hash)?,
-                CommittedParentHash::Hash(None) => {
-                    // Empty trie parent — use the initial empty revision.
-                    revisions_guard
-                        .front()
-                        .expect("always one revision")
-                        .clone()
-                }
+                CommittedParentHash::Hash(Some(hash)) => Some(db.manager.revision(hash)?),
+                CommittedParentHash::Hash(None) => None,
             };
             db.manager.remove_proposal(&nodestore);
 
-            let diff = DiffMerkleNodeStream::new(&*old_parent, &*nodestore, Box::default())?;
+            // Empty trie parent — diff against a synthetic empty view.
+            // Avoids depending on the manager's revisions queue, whose front
+            // entry is not guaranteed to still be empty once `max_revisions`
+            // has been exceeded.
+            let empty_parent;
+            let old_parent_ref: &NodeStore<Committed, FileBacked> = if let Some(rev) = &old_parent {
+                rev
+            } else {
+                empty_parent = nodestore.empty_committed_sibling();
+                &empty_parent
+            };
+            let diff = DiffMerkleNodeStream::new(old_parent_ref, &*nodestore, Box::default())?;
             let batch_ops: Vec<BatchOp<crate::merkle::Key, crate::merkle::Value>> =
                 diff.collect::<Result<_, _>>()?;
 
--- a/firewood/src/db/tests/concurrent_rebase.rs
+++ b/firewood/src/db/tests/concurrent_rebase.rs
@@ -60,7 +60,12 @@ const COMMITS_PER_THREAD: usize = 32;
 /// concurrent rebases interleave.
 const KEYS_PER_BATCH: usize = 16;
 
-/// Tighten `max_revisions` so reaping kicks in during the test.
+/// Tighten `max_revisions` so reaping kicks in during the concurrent tests.
+///
+/// Not shared with [`test_empty_parent_rebase_after_reap`], which uses a
+/// smaller constant locally: the concurrent tests' persist worker can't
+/// keep up if this is dropped below ~16, so a single shared low value is
+/// not viable.
 const MAX_REVISIONS: usize = 16;
 
 // Keys must be 32 bytes (64 nibbles) so the checker accepts them under
@@ -92,8 +97,13 @@ fn test_db_with_tight_revisions() -> TestDb {
     )
 }
 
-/// Run `db.check()` and assert no errors other than `AreaLeaks`.
-fn assert_check_clean(db: &TestDb, label: &str) {
+/// Run `db.check()` and assert no errors other than `AreaLeaks` (which is
+/// noisy and unrelated). `tolerate_unpersisted` allows callers running
+/// against an in-memory state — where the persist worker may still be
+/// behind the latest committed root — to ignore [`CheckerError::UnpersistedRoot`].
+/// After [`TestDb::reopen`], pass `false`: a reopened db must have a
+/// persisted root.
+fn assert_check_clean(db: &TestDb, label: &str, tolerate_unpersisted: bool) {
     let report = db.check(CheckOpt {
         hash_check: true,
         progress_bar: None,
@@ -102,6 +112,7 @@ fn assert_check_clean(db: &TestDb, label: &str) {
         .errors
         .iter()
         .filter(|e| !matches!(e, CheckerError::AreaLeaks(_)))
+        .filter(|e| !(tolerate_unpersisted && matches!(e, CheckerError::UnpersistedRoot)))
         .collect();
     assert!(real_errors.is_empty(), "{label}: {real_errors:?}");
 }
@@ -172,12 +183,16 @@ fn test_concurrent_commit_with_rebase_no_freelist_corruption() {
     });
 
     assert_disjoint_keys_present(&db, "post-concurrency");
-    assert_check_clean(&db, "post-concurrency");
+    // Pre-reopen check runs against the in-memory state; the persist worker
+    // may still be behind the latest committed root, so tolerate
+    // `UnpersistedRoot`.
+    assert_check_clean(&db, "post-concurrency", true);
 
-    // Reopen to surface any deferred-persistence corruption.
+    // Reopen to surface any deferred-persistence corruption. After reopen,
+    // `UnpersistedRoot` would indicate a real bug.
     let db = db.reopen();
     assert_disjoint_keys_present(&db, "post-reopen");
-    assert_check_clean(&db, "post-reopen");
+    assert_check_clean(&db, "post-reopen", false);
 }
 
 /// Many concurrent proposals built off the same parent, each modifying
@@ -265,7 +280,7 @@ fn test_concurrent_rebase_overlapping_paths() {
     // Close + reopen forces deferred persistence to flush; freelist
     // corruption surfaces here.
     let db = db.reopen();
-    assert_check_clean(&db, "post-reopen");
+    assert_check_clean(&db, "post-reopen", false);
 }
 
 /// Two proposals built off the same parent that produce the same `Put`. The
@@ -312,7 +327,7 @@ fn test_trivial_rebase_succeeds_without_corruption() {
     // Reopen forces the persist worker to flush before we run check();
     // otherwise the checker may race the worker and report `UnpersistedRoot`.
     let db = db.reopen();
-    assert_check_clean(&db, "post-trivial-rebase");
+    assert_check_clean(&db, "post-trivial-rebase", false);
 }
 
 /// A→B→A: commit `Put(K, V)`, then `Delete(K)`, then `Put(K, V)`. The first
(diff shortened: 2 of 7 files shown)


firewood/src/db.rs
```diff
@@ -529,7 +529,7 @@ impl Proposal<'_> {
         let mut revisions_guard = db.manager.lock_committed_revisions();
         let current = revisions_guard.back().expect("always one revision");
 
-        if !nodestore.parent_hash_is(current.root_hash()) {
+        if !nodestore.parent_id_is(current.committed_id()) {
             // Parent is stale — rebase onto the current revision.
             let old_parent = match nodestore.committed_parent_hash() {
                 CommittedParentHash::NotCommitted => {
```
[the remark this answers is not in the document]
rkuris: Fixed in 5ac3fd10fbd182e949a0c6c17c1e07f33dea7b68: the empty-parent branch now diffs against a synthetic empty `Committed` view constructed via a new `empty_committed_sibling()` helper, so it no longer depends on what `revisions_guard.front()` happens to be. Added `test_empty_parent_rebase_after_reap` as a regression: a proposal off the initial empty revision is committed via rebase after enough unrelated commits to reap that revision, and every key (proposal's plus all fillers) must survive. Passes under both default and `ethhash` features.

The same class of bug also exists for the `Hash(Some)` branch (non-empty parent reaped before `commit_with_rebase` looks it up via `db.manager.revision(hash)?`). That one can't use the synthetic-view trick — the diff needs the parent trie's actual contents — and the proper fix is to pin the parent revision Arc in `NodeStoreParent::Committed` (which currently holds only `hash`/`id` while `NodeStoreParent::Proposed` holds an Arc). Tracked separately in #1992 to keep this PR focused.

storage/src/nodestore/mod.rs
```diff
@@ -250,19 +264,28 @@ impl Parentable for Arc<ImmutableProposal> {
 
 impl<S> NodeStore<Arc<ImmutableProposal>, S> {
     /// When an immutable proposal commits, we need to reparent any proposal that
-    /// has the committed proposal as it's parent
-    pub fn commit_reparent(&self, other: &NodeStore<Arc<ImmutableProposal>, S>) {
-        other.kind.commit_reparent(&self.kind);
+    /// has the committed proposal as it's parent.
```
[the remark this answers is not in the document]
rkuris: Fixed in 5ac3fd10fbd182e949a0c6c17c1e07f33dea7b68

storage/src/nodestore/mod.rs
```diff
@@ -573,11 +635,21 @@ pub struct ImmutableProposal {
...
-    /// Returns true if the parent of this proposal is committed and has the given hash.
+    /// Returns true if the parent of this proposal is the committed revision
+    /// identified by `id`.
+    ///
+    /// `id` is fully unique per committed revision, so hash comparison would
+    /// be redundant. (We still keep the hash in [`NodeStoreParent::Committed`]
+    /// for the rebase path, which looks up the old parent revision by hash.)
+    /// Hash equality alone is not sufficient: a database can produce two
+    /// committed revisions with the same root hash (an A→B→A round trip),
+    /// and a proposal built off the first "A" must not be commitable against
```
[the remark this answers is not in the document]
rkuris: Fixed in 5ac3fd10fbd182e949a0c6c17c1e07f33dea7b68

storage/src/nodestore/mod.rs
```diff
@@ -419,7 +442,14 @@ impl<S: WritableStorage> NodeStore<Mutable<Propose>, S> {
...
                 inner: Propose {
                     deleted: Vec::default(),
-                    parent: NodeStoreParent::Committed(None),
+                    parent: NodeStoreParent::Committed {
+                        hash: None,
+                        // Sentinel id used only for the empty-trie placeholder
+                        // returned by `new_empty_proposal`; this proposal has
+                        // no real parent revision to identify, so it can never
+                        // commit through `commit_critical_section`.
+                        id: CommittedId::next(),
```
demosdemon, line 451: This would be better represented by an actual sentinel value instead of allocating an ID. Since the ids already start with one, you could make the `u64` a `NonZeroU64` and assert that this is not none during commit. Or, otherwise assert that the value is not the sentinel (zero is a good choice). The `next` method in a realistic system should never ever wrap and if it does, we should panic.

storage/src/nodestore/mod.rs
```diff
@@ -819,10 +902,19 @@ impl<S: WritableStorage> From<NodeStore<ImmutableProposal, S>> for NodeStore<Com
 }
```
demosdemon: This converter is wholly unused. Let's remove it now to minimize the number of places that allocate a new committed id.

storage/src/nodestore/mod.rs
```diff
@@ -123,6 +134,7 @@ impl<S: ReadableStorage> NodeStore<Committed, S> {
             kind: Committed {
                 deleted: Box::default(),
                 root: None,
+                id: CommittedId::next(),
```
demosdemon, line 138: Similarly here as for `NodeStoreParent::Committed` below, it appears that if the root hash is none then the committed id is irrelevant. Perhaps this means the committed id should be pushed into the option with the root hash?

demosdemon (approved):

firewood/src/manager.rs
RodrigoVillar, line 327: A comment on what `commit_critical_section()` returns here would be great.

firewood/src/registry.rs
```diff
@@ -21,6 +21,9 @@ firewood_metrics::define_metrics! {
         COMMITS_TOTAL          = "firewood_commits_total",
         /// Number of proposal commit operations
         PROPOSAL_COMMITS       = "firewood_proposal_commits_total",
+        /// Number of trivial proposal commits (proposal root hash equal to current
+        /// revision's root hash) that were skipped
```
RodrigoVillar:
```suggestion
        /// Number of proposal commits absorbed by the trivial fast path
        /// (proposal root hash equals current revision's; no new revision pushed)
```

Even in the case of trivial proposal commits, there is still some work to be done (proposal cleanup and reparenting). The current spec for this metric implies (IMO) that trivial proposal commits are a no-op.

firewood/src/db/tests/concurrent_rebase.rs
```diff
@@ -0,0 +1,472 @@
...
+/// proposal commits the front entry is no longer empty — the diff would
+/// then mis-compute the rebased batch (treating other revisions' keys as
+/// deletions). With the fix, an empty `Committed` view is constructed
+/// on the fly, so the diff yields exactly the proposal's puts and the
+/// rebased commit preserves all prior state.
+#[test]
+fn test_empty_parent_rebase_after_reap() {
+    // Tighten max_revisions so a small number of commits suffices to
+    // overflow and reap the initial empty revision.
+    const TIGHT_MAX_REVISIONS: usize = 4;
```
RodrigoVillar, line 447: q: would it suffice to set `MAX_REVISIONS = 4` so as to get rid of this constant?

firewood/src/db/tests/concurrent_rebase.rs
```diff
@@ -0,0 +1,472 @@
...
+                }
+            }));
+        }
+        for h in handles {
+            h.join().expect("thread panicked");
+        }
+    });
+
+    assert_disjoint_keys_present(&db, "post-concurrency");
+    assert_check_clean(&db, "post-concurrency");
```
RodrigoVillar: q: wouldn't this call be flaky as we could get an `UnpersistedRoot` error?

RodrigoVillar (approved):
Some comments but LGTM.

storage/src/nodestore/mod.rs
rkuris: Fixed in 5b0d474b. `CommittedId` now wraps `NonZeroU64`; `next()` builds via `NonZeroU64::new(...).expect("CommittedId counter overflowed u64")`, so an overflow panics rather than silently wrapping into a colliding id.

storage/src/nodestore/mod.rs
rkuris: Fixed in 5b0d474b — removed the `From<NodeStore<ImmutableProposal, S>> for NodeStore<Committed, S>` impl. All call sites go through the `Arc<ImmutableProposal>` path; the bare-`ImmutableProposal` form had no users.

storage/src/nodestore/mod.rs
rkuris: Not done — wanted to check the semantics with you first. The `id` distinguishes hash-equal empty parents in ∅→A→∅ sequences: a proposal pinned to the first empty must not commit against the second hash-equal empty, even though both have `hash: None`. Folding `id` into the `Option` (so empty parents have no id) would change `parent_id_is` to accept any empty as a match for any other empty — that's correct *only* if empty parents can never have a non-empty `deleted` list, which is true today but is an invariant I'd rather not pin down structurally in this PR. Happy to do it as a follow-up if you want — it'd let empty-parent proposals skip rebase entirely. Let me know.

firewood/src/manager.rs
rkuris: Fixed in 5b0d474b — doc now spells out: returns the `CommittedId` of the revision the proposal now lives in (either the freshly-inserted revision, or the current latest's id for the trivial fast path), which the caller passes to `commit_cleanup` to reparent siblings.

firewood/src/registry.rs
rkuris: Fixed in 5b0d474b. Took your wording: "Number of proposal commits absorbed by the trivial fast path (proposal root hash equals current revision's; no new revision pushed, but proposal cleanup and reparenting still run)."

firewood/src/db/tests/concurrent_rebase.rs
rkuris: Tried it (good idea in principle) — turns out the concurrent tests hang at `MAX_REVISIONS = 4`: with 8 threads × 32 commits each on a 4-revision cap, the persist worker can't keep up and `commit_with_rebase` blocks waiting for a persist permit. So the concurrent tests need ≥~16 and the new test stays on its own local `TIGHT_MAX_REVISIONS = 4`. Updated the constant's doc comment to record this.

firewood/src/db/tests/concurrent_rebase.rs
rkuris: Half-fix in 5b0d474b. The call you flagged (`assert_disjoint_keys_present`) only goes through `Db::revision`/`val`, which never raises `UnpersistedRoot` — that's a `CheckerError` from `db.check()`. The actually-racy line is `assert_check_clean(&db, "post-concurrency")` immediately after, which runs the checker on the in-memory state while the persist worker may still be behind. Gave `assert_check_clean` a `tolerate_unpersisted` flag and passed `true` for the pre-reopen call; post-reopen still passes `false` so a real `UnpersistedRoot` there would still fail.

storage/src/nodestore/mod.rs
demosdemon: agreed. I am okay with making the empty state generational for that future case.

PR #1988, opened 2026-05-10
fix(ffi): return None from find_next_key when no keys remain in requested range

rkuris:
## Why this should be merged

`FindNextKey` returns `end_key` as the next start key even if the next key is outside
the range requested.

Suppose a proof is generated that contains all keys in the range A-D, and that only
A-C and E-Z exist in the database. The returned right edge proof is likely to be an
exclusion proof of D, indicating that no keys to the right of C in the requested range
exist.

The current scheme always returns C as the "find_next_key" value. A subsequent
call for a range proof from C-D will again return a proof that C exists and will only
contain C. The caller has no way of knowing if the returned proof was just truncated
(that is, only one key was provided) or there are no other keys between C and D.

## How this works

When the caller asks for a bounded range and the proof confirms there are no
more keys to fetch within that range, `find_next_key` should return `None`.

Two paths produced the buggy behavior:

- **Change proof** (`ffi/src/proofs/change.rs`): when `batch_ops` was empty
  and the request was bounded, the function returned `(end_key, None)`. An
  empty diff over the requested range means the proof is complete — return
  `None`.

- **Range proof** (`ffi/src/proofs/range.rs`): a bounded range proof always
  carries a non-empty `end_proof` for `end_key` (even when the limit isn't
  hit), so `end_proof().is_empty()` can't distinguish "truncated" from
  "exhausted within range." Use `max_length` as the unambiguous truncation
  signal: if `key_values.len() < max_length` (or no limit was specified),
  the proof is exhaustive within the requested range and there's nothing
  more to fetch.

## How this was tested

- `cargo fmt`
- `cargo clippy --workspace --features ethhash,logger --all-targets` — clean
- `cargo nextest run --workspace --features ethhash,logger --all-targets` — 664 passed
- `cargo doc --no-deps` — clean
- `go test ./...` in `ffi/` — passes

Three Go FFI regression tests added in `ffi/proofs_test.go`:
- `TestChangeProofFindNextKeyBoundedNoDiffs` — bounded request over a range with zero diffs
- `TestChangeProofFindNextKeyEndAtRightmost` — bounded request where last_op equals end_key
- `TestRangeProofFindNextKeyEndPastTrie` — end_key past the rightmost trie key

## Breaking Changes

`firewood-ffi`/`firewood-go`: `FindNextKey` now returns `nil` rather than a
zero-progress continuation in the cases described above.

### DIFF
--- a/ffi/proofs_test.go
+++ b/ffi/proofs_test.go
@@ -670,6 +670,100 @@ func TestChangeProofFindNextKey(t *testing.T) {
 	r.NoError(nextRange.Free())
 }
 
+// TestChangeProofFindNextKeyBoundedNoDiffs verifies that when a bounded
+// change-proof request covers a range with zero diffs, FindNextKey returns
+// nil rather than echoing the requested end key as the next start key. The
+// caller would have no forward progress otherwise.
+func TestChangeProofFindNextKeyBoundedNoDiffs(t *testing.T) {
+	r := require.New(t)
+	dbA := newTestDatabase(t)
+	dbB := newTestDatabase(t)
+
+	keys, _, batch := kvForTest(100)
+	rootA, err := dbA.Update(batch[:50])
+	r.NoError(err)
+	_, err = dbB.Update(batch[:50])
+	r.NoError(err)
+
+	// All diffs between rootA and rootAUpdated live in keys[50:].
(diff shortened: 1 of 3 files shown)


rkuris:
This will get subsumed by a PR that handles all cases correctly, and eliminates the max_length heuristics.

PR #1991, opened 2026-05-11
feat: add ability to clone reconstructed revisions

RodrigoVillar:
## Why this should be merged

While benchmarking our Firewood archival nodes, we came across an error when making an `eth_estimateGas` API call that required block reexecution. After some investigation, the root cause was determined to be triggered by the following code:

https://github.com/ava-labs/avalanchego/blob/c5d4e6c31db451fb2e6d8a7fd178c19004985765/graft/evm/firewood/reconstructed_state.go#L73-L77

We currently return `nil` because reconstructed revisions cannot be shared while `state.Database` (the interface that `reconstructedStateAccessor` satisfies) has the following documentation for `CopyTrie`:

```go
// CopyTrie returns an independent copy of the given trie.
CopyTrie(Trie) Trie
```

Given that `eth_estimateGas` and similar methods call on `CopyTrie()`, there should be a way to create a new `reconstructedAccountTrie` while also not having two tries share the same reconstructed revision.

## How this works

Adds a `Clone()` method to the reconstructed pointer type, which returns a new, independent reconstructed revision. This independent reconstructed revision can then be used for the implementation of `CopyTrie()`.

*Note*: an alternative approach that was suggested was to just return a new instance of `reconstructedAccountTrie` via copy-by-reference. While this approach also solves the bug fix, this explicitly violates the invariant set in `CopyTrie()`, setting up a footgun I would **strongly prefer** not to introduce into AvalancheGo.

## How this was tested

CI + added additional UTs in Go and Rust verifying that cloned reconstructed revisions are independent of the reconstructed revision they are cloned from.

I also have a PoC in AvalancheGo which utilizes this fix, and have verified that all usages of `CopyTrie()` are read-only.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/ffi/reconstructed.go
+++ b/ffi/reconstructed.go
@@ -13,6 +13,8 @@ package ffi
 // #cgo nocallback fwd_iter_on_reconstructed
 // #cgo noescape fwd_reconstruct_on_reconstructed
 // #cgo nocallback fwd_reconstruct_on_reconstructed
+// #cgo noescape fwd_clone_reconstructed
+// #cgo nocallback fwd_clone_reconstructed
 // #cgo noescape fwd_reconstructed_dump
 // #cgo nocallback fwd_reconstructed_dump
 // #cgo noescape fwd_free_reconstructed
@@ -169,6 +171,21 @@ func (r *Reconstructed) Reconstruct(batch []BatchOp) error {
 	r.rootMu.Unlock()
 
 	return nil
+}
+
+// Clone returns an independent [Reconstructed] handle that shares the
+// underlying view with the receiver. The two handles can be used and
+// freed independently.
+func (r *Reconstructed) Clone() (*Reconstructed, error) {
+	r.keepAliveHandle.mu.RLock()
+	defer r.keepAliveHandle.mu.RUnlock()
+
+	if r.dropped {
+		return nil, ErrDroppedReconstructed
+	}
+
+	result := C.fwd_clone_reconstructed(r.ptr)
+	return getReconstructedFromResult(result, r.keepAliveHandle.outstandingHandles)
 }
 
 // Dump returns a DOT (Graphviz) representation of this reconstructed view.
--- a/ffi/reconstructed.go
+++ b/ffi/reconstructed.go
@@ -176,12 +176,6 @@ func (r *Reconstructed) Reconstruct(batch []BatchOp) error {
 // Clone returns an independent [Reconstructed] handle that shares the
 // underlying view with the receiver. The two handles can be used, reconstructed,
 // and freed independently.
-//
-// Calling [Reconstructed.Reconstruct] on either handle while another handle
-// shares the same reconstructed state is supported but expensive: it forces a
-// deep clone of the root node instead of the zero-copy move used when the
-// handle is uniquely owned. Drop unneeded clones before reconstructing when
-// write performance matters.
 func (r *Reconstructed) Clone() (*Reconstructed, error) {
 	r.keepAliveHandle.mu.RLock()
 	defer r.keepAliveHandle.mu.RUnlock()
--- a/ffi/reconstructed.go
+++ b/ffi/reconstructed.go
@@ -174,14 +174,14 @@ func (r *Reconstructed) Reconstruct(batch []BatchOp) error {
 }
 
 // Clone returns an independent [Reconstructed] handle that shares the
-// underlying view with the receiver. The two handles can be used and
-// freed independently.
+// underlying view with the receiver. The two handles can be used, reconstructed,
+// and freed independently.
 //
-// Callers should only perform read operations on the clone and drop it
-// before performing further writes. Calling [Reconstructed.Reconstruct]
-// on the clone (or on any sibling) while another clone is alive is
-// expensive: it forces a deep clone of the root node instead of the
-// zero-copy move that the uncloned path takes.
+// Calling [Reconstructed.Reconstruct] on either handle while another handle
+// shares the same reconstructed state is supported but expensive: it forces a
+// deep clone of the root node instead of the zero-copy move used when the
+// handle is uniquely owned. Drop unneeded clones before reconstructing when
+// write performance matters.
 func (r *Reconstructed) Clone() (*Reconstructed, error) {
 	r.keepAliveHandle.mu.RLock()
 	defer r.keepAliveHandle.mu.RUnlock()
--- a/ffi/reconstructed.go
+++ b/ffi/reconstructed.go
@@ -176,6 +176,12 @@ func (r *Reconstructed) Reconstruct(batch []BatchOp) error {
 // Clone returns an independent [Reconstructed] handle that shares the
 // underlying view with the receiver. The two handles can be used and
 // freed independently.
+//
+// Callers should only perform read operations on the clone and drop it
+// before performing further writes. Calling [Reconstructed.Reconstruct]
+// on the clone (or on any sibling) while another clone is alive is
+// expensive: it forces a deep clone of the root node instead of the
+// zero-copy move that the uncloned path takes.
 func (r *Reconstructed) Clone() (*Reconstructed, error) {
 	r.keepAliveHandle.mu.RLock()
 	defer r.keepAliveHandle.mu.RUnlock()
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -625,12 +625,6 @@ pub extern "C" fn fwd_reconstruct_on_reconstructed<'db>(
 /// view with the input handle. Both handles can be used, reconstructed, and
 /// freed independently.
 ///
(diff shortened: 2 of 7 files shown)


ffi/reconstructed.go
```diff
@@ -171,6 +173,27 @@ func (r *Reconstructed) Reconstruct(batch []BatchOp) error {
 	return nil
 }
 
+// Clone returns an independent [Reconstructed] handle that shares the
+// underlying view with the receiver. The two handles can be used and
+// freed independently.
+//
+// Callers should only perform read operations on the clone and drop it
+// before performing further writes. Calling [Reconstructed.Reconstruct]
```
alarso16: What's the point of returning a different pointer if all you can do is read?

RodrigoVillar: I'm not sure what you mean - it's possible to write to a cloned reconstructed pointer. However, the documentation here is making clear that writing to a cloned reconstructed pointer (or its siblings) is expensive for the reasons outlined. Happy to update the documentation though if it currently sounds like clones are read-only.

rkuris (commented):
We should make sure the thing being cloned has been hashed before allowing it to be cloned. This is fairly easy; just call root_hash() on the Reconstructed version and discard the hash.

Here is Claude's explanation as to why (slightly incorrect but still gets the point across; first noticed error is that it implies *all* children are of type `Child::Node`, which is incorrect). There are other mistakes so YMMV, but the conclusion is still correct.

```
  Why cloning a NodeStore is expensive before hashing

  The cost comes from the Child enum in storage/src/node/branch.rs:59:

  pub enum Child {
      Node(Node),                                  // owned, unhashed
      AddressWithHash(LinearAddress, HashType),    // persisted
      MaybePersisted(MaybePersistedNode, HashType) // Arc-shared
  }

  Before hashing, branch children are Child::Node(Node) — each one
  owns its entire subtree inline. Cloning a NodeStore in that state
  recursively deep-clones every descendant: every BranchNode, every
  Box/SmallVec for partial paths, every value buffer, all the way down
   to the leaves. Cost is O(total nodes × per-node copy) plus a fresh
  allocation per node. For a Reconstructed nodestore mid-build, this
  is the whole working set.

  After hashing (storage/src/nodestore/hash.rs:222), hash_helper_inner
   rewrites each visited child as:

  *child = Some(Child::MaybePersisted(child_node, child_hash));

  where child_node came from SharedNode::new(node).into() at line 247
  — i.e., the Node got wrapped in an Arc. Now cloning a child is one
  atomic refcount bump, O(1), zero allocation. The whole trie becomes
  structurally shared between the original and the clone.

  The type itself bakes this in: Child::Node's heap_size walks the
  subtree, while MaybePersisted's is hardcoded to 0 because the data
  is shared (branch.rs:71-79).

  TL;DR: pre-hash clone = deep copy of the entire owned trie.
  Post-hash clone = a pile of Arc refcount increments. Hash first,
  then clone is essentially free.
```

ffi/reconstructed.go
alarso16:
```
// Callers should only perform read operations on the clone and drop it
// before performing further writes.
```
I though this said that the clone is read only, and only after dropping the clone can you write to the original. This documentation was not clear to me at all

RodrigoVillar: Updated the documentation to make it clear that writing to cloned reconstruction revisions (or its siblings) is supported, but incurs a performance hit here: [a6d5ab2](https://github.com/ava-labs/firewood/pull/1991/commits/a6d5ab2d69cb0a9d48d8ed7f1b8101cd3426254c)

RodrigoVillar:
Spoke offline with @rkuris - we agreed that with the current implementation of `root_hash`, there are no performance/memory benefits to calling `root_hash()` before cloning because we throw away the newly computed root node. @rkuris is working on a fix that will enable us to use the newly computed root node rather than throwing it away (this would therefore, reduce the performance/memory hit of cloned reconstructed revisions).

ffi/src/lib.rs
```diff
@@ -621,6 +621,45 @@ pub extern "C" fn fwd_reconstruct_on_reconstructed<'db>(
...
 
+/// Produces an independent reconstructed handle that shares the underlying
+/// view with the input handle. Both handles can be used, reconstructed, and
+/// freed independently.
+///
+/// Calling [`fwd_reconstruct_on_reconstructed`] on either handle while another
+/// handle shares the same reconstructed state is supported but expensive: it
+/// forces a deep clone of the root node instead of the zero-copy move used when
+/// the handle is uniquely owned. Free unneeded clones before reconstructing
+/// when write performance matters.
```
rkuris: This could probably use a rewrite. There is only a performance problem if you hang on to a cloned reconstructed handle and either of them decide to advance forward.

ffi/src/lib.rs
```diff
@@ -621,6 +621,45 @@ pub extern "C" fn fwd_reconstruct_on_reconstructed<'db>(
...
+///
+/// # Arguments
+///
+/// * `handle` - The reconstructed handle returned by [`fwd_reconstruct_on_revision`] or
+///   [`fwd_reconstruct_on_reconstructed`].
```
rkuris: Seems like a test where we pass in something else is a good idea.

ffi/reconstructed.go
rkuris: I think you can be even stronger here. It's possible to *fork* reconstructed revision chains, and it's actually not so terrible provided you calculate the hash before the fork. I almost wonder if we should do that work automatically at fork time.

ffi/src/lib.rs
RodrigoVillar: Hmm I'm not sure how the current documentation doesn't make this clear:

```rs
/// Calling [`fwd_reconstruct_on_reconstructed`] on either handle while another
/// handle shares the same reconstructed state is supported but expensive:**
```

ffi/reconstructed.go
RodrigoVillar: Regarding the first point (forking reconstructed revision chains), I think the first few lines of the documentation for `Clone()` make this clear:

```rs
/// Produces an independent reconstructed handle that shares the underlying
/// view with the input handle. Both handles can be used, reconstructed, and
/// freed independently.
///
/// Calling [`fwd_reconstruct_on_reconstructed`] on either handle while another
/// handle shares the same reconstructed state is supported but expensive
```

Regarding the second point (perf hit of writing to cloned reconstructed revisions), I see you have https://github.com/ava-labs/firewood/pull/2005 open - if it's true that the perf hit of writing to a reconstructed revision becomes almost nil in this, I'm happy to remove all references of cloned reconstructed revisions being expensive in PR if #2005 is merged in first!

ffi/src/lib.rs
RodrigoVillar: An implicit callout here is that `fwd_reconstruct_on_reconstructed` can also take in a handle tied to a cloned reconstructed revision - updated the documentation for that here: [2e403d1](https://github.com/ava-labs/firewood/pull/1991/commits/2e403d11c1d5137c6cd9955b191f014c218b77f0) and added a UT that tests this edge case. 

Regarding: 

> Seems like a test where we pass in something else is a good idea.

We already check for `nil`/non-`nil` pointers via `invoke_with_handle`. If we want to check for runtime errors, I'm open to this but don't think this PR is the place for that, considering that all other functions in the FFI library don't test for this.

rkuris (commented):
Some of the comments could use some improvement before merging.

ffi/src/lib.rs
rkuris: Specifically, I'm looking for a test that passes a proposal handle made from outside the reconstructed flow instead of a handle from a previously reconstructed revision.

rkuris (approved):
Rodrigo points out that a ProposalHandle is not the same thing as a ReconstructedHandle. Our current convention in the FFI layer is to call attention to the fact that some rogue C programmer might decide to pass a raw pointer to something else despite the signature provided in the FFI call. Generally, this is going to create at a minimum a type cast warning even in C so I'm not sure of the merit, but that isn't a problem with this specific PR, as we do that in other places too. Might be a good followup.

The performance problems mentioned in the comments depend on factors that may change, particularly if we land something like #2005. Because of this, we don't need to describe the performance picture at the API layer and should leave performance decisions to standard practices of benchmarking API calls before releasing software.

Approved on conditions that we just remove these comments.

RodrigoVillar:
Removed all performance references in the documentation here: [95d7ad7](https://github.com/ava-labs/firewood/pull/1991/commits/95d7ad78999642b36a64f694d7c149cc612084bc). Furthermore, I've removed the debug log introduced in this commit ([cfeec9a](https://github.com/ava-labs/firewood/pull/1991/commits/cfeec9aa6e2cca103d58487705bf05746d327845)) as this was a signal related to performance.

PR #1993, opened 2026-05-12
chore: add firewood review skill and copilot instructions

demosdemon:
## Why this should be merged

A dedicated skill for firewood review will allow us to accelerate feedback and iteration and ideally reduce the time outstanding PRs spend in review. The copilot instructions will help onboard new contributors and ensure they have the necessary information to get started.

## How this works

This commit adds a new skill for Claude to review firewood PRs and provides detailed instructions for using GitHub Copilot to assist with contributions.

## How this was tested

This skill was used to review a sample of open PRs locally.

Reviewed skill against spec: <https://github.com/agentskills/agentskills> and ran the skills-ref validator for simple checks.

## Breaking Changes

None.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
+# Firewood — GitHub Copilot Instructions
+
+Firewood is an embedded key-value store for Merkleized blockchain state, written in Rust
+(edition 2024, MSRV 1.94.0) with a Go FFI layer (`ffi/`). It uses the trie structure
```
rkuris: Concerned about having another place to edit when we change MSRV and/or the rust edition.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+## Code Review
+
+When reviewing pull requests or suggesting code changes, apply the firewood-specific
+checks below on top of your built-in review. These supplement — not replace — your
+standard review.
+
+Claude users: a dedicated `firewood-review` skill is available at
+`.claude/skills/firewood-review/SKILL.md`. It orchestrates parallel review agents
+covering the Rust toolchain, Go FFI, CI workflows, and code quality.
+
```
rkuris: If this is instructions for copilot, this is perhaps the wrong place to make this reference.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+- **Comments and messages**: Every comment, log message, and error string must be
+  accurate, useful, and free of tautologies. Log and error messages must be unique enough
+  to locate their source quickly.
+- **Atomic operations**: Verify correct memory ordering (`Ordering`) for every load,
+  store, and read-modify-write operation.
+- **Locks**: Check for potential deadlocks, excessively long critical sections, and
+  inappropriate `RwLock` vs `Mutex` choices given the actual access patterns.
+- **Panic potential**: Flag unvalidated array indexing, unchecked arithmetic, and
+  `unreachable!()` in paths that could actually be reached.
```
rkuris: This should (mostly) get caught by clippy. I think a more useful comment here would be that often array indexing means the code is not idiomatic rust and to avoid the use of indexing in general. Obviously there are some valid use cases so I'm not sure how to word this correctly.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+- **Anti-patterns**: Recommend modern idioms. Avoid `ToString::to_string` — prefer
+  `to_owned()` for `&str → String` conversions. Flag unnecessary string allocations.
```
rkuris: Feels like we could/should add this as a clippy rule rather than letting github copilot catch it later.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+- **Visibility**: Reject unnecessary `pub` exports. Prefer the tightest visibility
+  (`pub(crate)`, `pub(super)`) that satisfies the interface requirements.
+- **Error handling**: Reject silently discarded errors, incorrect error types, and errors
+  returned without sufficient context.
```
rkuris: This feels duplicative to what was said before under "error propogation"

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+- **Tests**: New non-trivial code requires tests. Tests must be targeted (not covering
+  too broad a scope), DRY, and cover edge cases. Scrutinize test names for clarity.
```
rkuris: Test ordering matters; similar tests should be grouped together

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+| (none)                      | ✓     | ✓     |
+| `--no-default-features`     | ✓     | ✓     |
+| `--features ethhash,logger` | ✓     | ✓     |
+| `--all-features`            | ✗     | ✓     |
+
+## Commit and PR Conventions
+
+Commit messages and PR titles must follow [Conventional Commits](https://www.conventionalcommits.org/).
+Allowed types: `build`, `chore`, `ci`, `docs`, `feat`, `fix`, `perf`, `refactor`,
+`style`, `test`.
```
rkuris: I'd prefer referencing the file .github/workflows/pr-title.yaml in case we change the list.

Edited to add: we probably don't need this here as CI will catch this violation.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Additional Context
+
+See `AGENTS.md` for full project context, architecture principles, and the complete
+code review guidelines. See `CONTRIBUTING.md` for the contribution workflow and style
+guide.
```
rkuris: AGENTS.md already refers to CONTRIBUTING.md.

AGENTS.md
```diff
@@ -223,6 +223,38 @@ Key dependencies are centrally managed in workspace `Cargo.toml`:
 
 7. **Workspace Awareness**: This is a multi-crate workspace. Changes may affect multiple crates. Check `Cargo.toml` for workspace structure.
 
+## Code Review Guidelines
+
+Apply these firewood-specific checks whenever reviewing code, suggesting changes, or providing inline feedback — in addition to general best practices.
+
+### Rust
+
+- **Overflow arithmetic**: Reject `saturating_*` or `checked_*` for operations that are provably impossible to overflow — they add runtime cost for no benefit. Scrutinize `wrapping_*` for operations where wrapping would silently produce incorrect results.
```
rkuris: Is there any way we can reduce duplication here? We just said this (and several things after this) in copilot-instructions.md. It would be trivial to get these out of date going forward.

No further reviews of this file. I care mostly about the github copilot file; the rest we can iterate on.

rkuris (approved):

rkuris:
Approved anyway. This feels highly duplicative and somewhat tautological. For example, I doubt saying "Every comment, log message, and error string must be accurate, useful, and free of tautologies" is likely to change the agent's behavior; this is something it seems to do fine without this prompting.

.github/copilot-instructions.md
demosdemon: added with https://github.com/ava-labs/firewood/pull/1993/changes/4dcd2736f5b7f0531149042a70df25cb023986d4

PR #1996, opened 2026-05-13
ci: Update workflow to use macOS 26, drop x86_64 mac target

demosdemon:
## Why this should be merged

The `macos-latest` runners are having an issue that prevents them from correctly building. The `macos-26` runners do not. Additionally, the x86_64 mac build hasn't been supported for a while.

## How this works

This updates the os targets for the one workflow to use the `ubuntu-24.04`, `ubuntu-24.04-arm` (there is no ubuntu-latest-arm variant), and `macos-26` runners and drops the x86_64 run as its no longer supported. Consistency with the `ubuntu-*` runners in the `attach-static-libs` was preferred over using `ubuntu-latest` and `ubuntu-24.04-arm`.

## How this was tested

CI

### DIFF
--- a/.github/workflows/attach-static-libs.yaml
+++ b/.github/workflows/attach-static-libs.yaml
@@ -27,7 +27,6 @@ concurrency:
 # Supported architectures are:
 # - x86_64-unknown-linux-gnu
 # - aarch64-unknown-linux-gnu
-# - x86_64-apple-darwin
 # - aarch64-apple-darwin
 jobs:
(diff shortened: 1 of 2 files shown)


demosdemon:
The repo CI settings need to be updated as they will require a job that no longer runs after this merge.

RodrigoVillar (approved):
LGTM - I assume this is the failing job?: https://github.com/ava-labs/firewood/actions/runs/25818964569/job/75855259053

rkuris (approved):

PR #1998, opened 2026-05-14
ci: disable bin cache for rust-cache

demosdemon:
## Why this should be merged

This is a temporary workaround until the upstream issue is resolved or until alternative cache is setup.

- https://github.com/Swatinem/rust-cache/issues/341

## How this works

This applies the recommended change to disable `cache-bin`. This will cause CI to take longer as any installed bins will need to be rebuilt each time.

## How this was tested

CI

## Breaking Changes

n/a


demosdemon:
this doesn't appear necessary, as I have noted here: https://github.com/Swatinem/rust-cache/issues/341#issuecomment-4461991963

I purged all of our caches and PRs are passing again (except for the attach-static-libs workflow which is an unrelated credentials problem).

PR #1999, opened 2026-05-15
perf(storage): retain hashed root in Reconstructed for cheap forks

rkuris:
## Why this should be merged

`HashedNodeReader::root_hash` on a `Reconstructed` nodestore called `hash_helper`, which builds a fully-hashed in-memory tree (every `Child::Node` rewritten to `Child::MaybePersisted` over an `Arc`-shared subtree), and then threw that tree away. Only the resulting `TrieHash` was cached in a `OnceLock`. The next fork-and-reconstruct off the same `Reconstructed` still saw `Child::Node` children and paid for a full deep clone of the trie structure.

Now the hashed tree is kept: `root_hash` swaps the new `SharedNode` into the root via `ArcSwapAny`, so any subsequent
`From<NodeStore<Reconstructed>> for NodeStore<Mutable<Recon>>` walks `Arc`-shared `MaybePersisted` children rather than recursively copying `Child::Node` subtrees. The unhashed `M -> R -> M -> R` reconstruction chain is unaffected — it never invokes `root_hash`, so the swap never fires.

## How this works

- `Reconstructed::root` changes from `Option<SharedNode>` to `Option<ArcSwapAny<TriompheArc<Node>>>`. The `OnceLock<TrieHash>` next to it is unchanged.
- A new sealed module `storage/src/arc_swap_triomphe.rs` provides `TriompheArc<T>`, a `#[repr(transparent)]` newtype around `triomphe::Arc<T>`. It implements `arc_swap::RefCnt` by forwarding to `triomphe::Arc`'s raw-pointer API. This is the only place in the storage crate that opts out of `deny(unsafe_code)`; the unsafe surface is three forwarded methods with documented soundness.
- `root_hash` now stores the hashed `SharedNode` (extracted from the `MaybePersistedNode` returned by `hash_helper`) back into the `ArcSwap`, then caches the `TrieHash` in the existing `OnceLock`.
- `From<NodeStore<Mutable<Recon>>> for NodeStore<Reconstructed>` wraps the root in an `ArcSwapAny` without computing the hash, preserving the cheap-chain path.
- `RootReader::root_node` and the `Clone` impl for `Reconstructed` load from the `ArcSwap` rather than cloning a stored `SharedNode`.

## A note on arc-swap

We've used `arc-swap` in firewood before, on hot paths with hundreds of concurrently-swapping nodes, and it caused performance problems there. This use is different: a `Reconstructed` has exactly one `ArcSwapAny`, the swap fires at most once per `Reconstructed` lifetime (when `root_hash` is first computed), and the number of live `Reconstructed` instances is small. None of the conditions that hurt us last time apply.

## How this was tested

- `cargo nextest run --workspace --features ethhash,logger --all-targets` (710 tests, all pass).
- `cargo clippy --workspace --features ethhash,logger --all-targets` (clean).
- `cargo doc --no-deps` (clean).
- New unit test `reconstructed_root_hash_rewrites_root_children` builds a `Reconstructed` whose root is a branch with `Child::Node` children, calls `root_hash`, then asserts the root branch returned by `RootReader::root_node` no longer contains any `Child::Node` child. `hash_helper` works bottom-up, so the root assertion implies the whole subtree was rewritten.

## Breaking Changes

None. The change is internal to `firewood-storage`; no public API signatures change.

### DIFF
--- a/storage/src/arc_swap_triomphe.rs
+++ b/storage/src/arc_swap_triomphe.rs
@@ -1,0 +1,91 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! `RefCnt` bridge so that `arc_swap` can swap `triomphe::Arc` values.
+//!
+//! `arc-swap` ships `RefCnt` impls for `std::sync::Arc` and `std::rc::Rc` but
+//! not for `triomphe::Arc`, which is what `SharedNode` is built on. The orphan
+//! rule prevents us from implementing `RefCnt` for `triomphe::Arc` directly,
+//! so this module exposes a transparent newtype, [`TriompheArc`], that wraps
+//! it and implements `RefCnt` itself.
+//!
+//! This module is the *only* place in the storage crate that opts out of
+//! `deny(unsafe_code)`. The unsafe surface is intentionally minimal: each
+//! `RefCnt` method forwards to the matching `triomphe::Arc` raw-pointer API.
+//!
+//! Soundness rests on three properties that `triomphe::Arc` shares with
+//! `std::sync::Arc`:
+//!
+//! 1. `into_raw` consumes the `Arc` without touching the refcount and returns
+//!    a stable pointer to the inner `T`.
+//! 2. `from_raw(p)` rehydrates ownership of the same refcount slot, given a
+//!    pointer that came from a prior `into_raw`.
+//! 3. `as_ptr` returns the same pointer `into_raw` would, without affecting
+//!    the refcount.
+//!
+//! The pointer is non-null for any sized `T` we use here (`Node` is non-ZST),
+//! which is what `arc_swap`'s `RefCnt for Option<T>` relies on to distinguish
+//! `Some` from `None`.
+
+#![allow(
+    unsafe_code,
+    reason = "arc_swap::RefCnt requires unsafe trait methods; sealed bridge for triomphe::Arc"
+)]
+
+use std::ops::Deref;
+
+use arc_swap::RefCnt;
+
+/// Transparent wrapper around `triomphe::Arc<T>` that implements
+/// [`arc_swap::RefCnt`] so the value can live inside an `ArcSwapAny`.
+#[repr(transparent)]
+pub(crate) struct TriompheArc<T>(triomphe::Arc<T>);
+
+impl<T> TriompheArc<T> {
+    pub(crate) const fn new(arc: triomphe::Arc<T>) -> Self {
+        Self(arc)
+    }
+
+    pub(crate) fn into_inner(self) -> triomphe::Arc<T> {
+        self.0
+    }
+}
+
+impl<T> Clone for TriompheArc<T> {
+    fn clone(&self) -> Self {
+        Self(triomphe::Arc::clone(&self.0))
+    }
+}
+
+impl<T> Deref for TriompheArc<T> {
+    type Target = triomphe::Arc<T>;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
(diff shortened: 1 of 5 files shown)


storage/src/arc_swap_triomphe.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl<T: std::fmt::Debug> std::fmt::Debug for TriompheArc<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+// SAFETY: forwards to triomphe::Arc's raw-pointer API, which provides the same
+// refcount-stable, ABI-stable guarantees that std::sync::Arc does (see module docs).
+unsafe impl<T> RefCnt for TriompheArc<T> {
```
RodrigoVillar, line 75: From the top-level docs:

```rs
//! The pointer is non-null for any sized `T` we use here (`Node` is non-ZST),
//! which is what `arc_swap`'s `RefCnt for Option<T>` relies on to distinguish
//! `Some` from `None`.
```

Would it make sense to bound `T` with `Sized`?

storage/src/nodestore/mod.rs
```diff
@@ -1003,22 +1029,33 @@ where
...
     }
 
-    // Lazily compute and cache the hash. We don't use parallel hash
-    // calculations here because reconstructed views don't benefit as much
-    // from spinning up multiple threads.
+    // Lazily compute and cache the hash. The hashed root (with children rewritten
+    // to `Child::MaybePersisted`) is swapped back into `self.kind.root` so any
+    // subsequent fork-then-reconstruct walks Arc-shared children instead of
+    // deep-cloning every `Child::Node` subtree. We don't parallelise the hash
+    // because reconstructed views don't benefit much from multiple threads.
```
RodrigoVillar, line 1135: nit:
```suggestion
    // deep-cloning every `Child::Node` subtree. We don't parallelise the hash
    // calculations because reconstructed views don't benefit much from multiple threads.
```

storage/src/nodestore/mod.rs
```diff
@@ -1003,22 +1029,33 @@ where
...
-        let (_, hash) =
+        let (hashed_mp, hash) =
             NodeStore::<Mutable<Propose>, S>::hash_helper(node_val, Path::new()).ok()?;
+        // Extract the in-memory SharedNode from the freshly-built MaybePersistedNode
+        // (always the `Unpersisted` variant here, so this is a Mutex lock + Arc clone,
+        // no I/O). Replace the unhashed root with the fully-hashed one. If another
+        // thread raced us, the loser's swap overwrites with an equivalent tree; both
+        // contain the same trie content, only the children's form differs
+        // (`Child::Node` vs `Child::MaybePersisted`) and both are valid reads.
+        let hashed_shared = hashed_mp.as_shared_node(self).ok()?;
```
RodrigoVillar, line 1154: q: I wonder if it makes sense to `debug!()`/`debug_assert!()` the error if there is one.

RodrigoVillar (approved):
I would like another approval on this before merging in.

storage/src/arc_swap_triomphe.rs
rkuris: In Rust, every type parameter has an implicit Sized bound unless you add `?Sized`.

storage/src/nodestore/mod.rs
rkuris: I can't imagine this showing up as an actual problem. `debug_assert` can cause a panic if this race happens. I think it's just the cost of doing it efficiently here. Perhaps a debug log can substitute for the comment though.

demosdemon (approved):

PR #2002, opened 2026-05-15
ci: use deploy key instead of pat

demosdemon:
## Why this should be merged

The github PAT we used for this workflow has expired. Security recommended we switch to a deploy key, which has the advantage of not expiring (though, it should still be rotated periodically).

## How this works

This updates the workflow to use the deploy key instead of the pat.

## How this was tested

The workflow executed successfully: https://github.com/ava-labs/firewood/actions/runs/25934283918?pr=2002

## Breaking Changes

N/A


rkuris (approved):
LGTM. Can you put instructions on how to rotate it somewhere?

PR #2003, opened 2026-05-17
refactor(storage): lift nibbles_to_eth_compact into eth_encoding module

### LINKED ISSUE #2001: feat: Implement rust-side eth_getProof

**Additional context**

rkuris:
## Why this should be merged

Move the hex-prefix (compact) path encoder out of the `ethhash`-gated
hasher and into a new unconditionally compiled `storage::eth_encoding`
module. The function is pure and has no feature-specific dependencies;
its previous gating was incidental to where it lived.

This is groundwork for an `eth_getProof`-compatible proof emitter, which
needs to emit canonical MPT-RLP node bytes regardless of which hasher
the underlying trie was built with. Forcing the emitter to live behind
`#[cfg(feature = \"ethhash\")]` just to reach this helper would add walls
we'd then have to tear down.

It also lines up with the broader direction of replacing the `ethhash`
compile-time feature flag with a runtime mode knob: pulling pure
encoding helpers out of the gated module shrinks what the feature
actually has to guard, so the eventual flag removal is a smaller change.

## How this works

- New `storage/src/eth_encoding.rs` holds `nibbles_to_eth_compact` and
  its existing unit tests.
- `storage/src/lib.rs` registers `pub mod eth_encoding`.
- `storage/src/hashers/ethhash.rs` imports the helper from its new
  location and drops the local copy and tests.

No behavior change.

## How this was tested

- ``cargo build -p firewood-storage`` (default features)
- ``cargo build -p firewood-storage --features ethhash``
- ``cargo nextest run -p firewood-storage --features ethhash,logger eth_encoding`` — all 6 lifted test cases pass.
- ``cargo clippy -p firewood-storage --features ethhash,logger --all-targets`` — clean.
- ``cargo fmt --check`` — clean.

## Breaking Changes

None.

### DIFF
--- a/storage/src/eth_encoding.rs
+++ b/storage/src/eth_encoding.rs
@@ -1,0 +1,85 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! # Ethereum MPT path encoding
+//!
+//! Encoding primitives for the Ethereum Modified Merkle Patricia Trie that are
+//! needed by both the ethhash hasher and the `eth_getProof`-compatible proof
+//! emitter. These helpers are unconditionally compiled so the proof emitter
+//! (which is itself always compiled) can call them without pulling in the
+//! `ethhash` feature gate.
+
+use crate::TriePath;
+use bitfield::bitfield;
+use smallvec::SmallVec;
+use std::iter::once;
+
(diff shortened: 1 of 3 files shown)


rkuris:
Auto-merge is set. This is a straight move so it should be easy to review and can merge immediately.

RodrigoVillar (approved):

PR #2009, opened 2026-05-20
fix(proofs): simple find_next_key for range proofs

bernard-avalabs:
## Why this should be merged

  `find_next_key_after_range_proof` was a placeholder returning
  `Some((last_key, end_key))` whenever the proof had any keys and a non-empty
  `end_proof`. The returned cursor was `last_key` itself, so when a prover
  re-served the same first key in response to the new range, the syncer
  looped indefinitely.

  This PR replaces the placeholder with a simple, correct, conservative
  implementation. Range proofs only; the change-proof counterpart will
  follow in a separate PR.
  
  ## How this works

  The implementation enumerates five structural cases on the proof's
  `end_proof` terminal node, returning `None` whenever the proof structure
  proves the range is exhaustively covered:

  - Empty `key_values` (Case 1) → `None`.
  - `last_key == end_key` (Case 2) → `None`.
  - Empty `end_proof` (preflight) → `None`.
  - Terminal has no value (Case 3, exclusion proof) → `None`.
  - Terminal has a value at a key ≠ `last_key` (Case 4) → `None`.

  For the one structurally ambiguous case (Case 5 — terminal carries a
  value at exactly `last_key`'s path), the implementation cannot
  distinguish a truncated inclusion of `last_key` from an exhaustive
  exclusion of `end_key` whose terminal happens to lie at `last_key`'s
  node. It conservatively returns
  `Some((lex_successor(last_key), end_key))`, where `lex_successor`
  appends `0x00`. The cursor strictly advances past `last_key`, so the
  syncer cannot loop on identical prover responses — guaranteed forward
  progress.

  ## How this was tested

  - **Unit tests** (`firewood/src/proofs/range.rs`, `firewood/src/proofs/tests.rs`):
    7 hand-built proof tests covering every algorithm branch, plus 5 tests
    for the new `lex_successor` helper.
  - **Integration tests** (`firewood/src/merkle/tests/range.rs`): 2 real-proof
    tests using `init_merkle` + `range_proof`. One exercises the Case 5
    ambiguous shape end-to-end; the other is a two-round syncer flow that
    confirms forward progress.

  ## Breaking Changes

  None. `find_next_key_after_range_proof`'s signature is unchanged; only
  its body changes. The FFI wrapper compiles and runs unchanged.
  `lex_successor` is a new `pub(crate)` helper, so no external API
  surface change.

### DIFF
--- a/firewood/src/proofs/range.rs
+++ b/firewood/src/proofs/range.rs
@@ -58,6 +58,8 @@
 
 use std::num::NonZeroUsize;
 
+use firewood_storage::{Hashable, NibblesIterator, Path, TriePath};
+
 use crate::api::{self, FrozenRangeProof, HashKey};
 use crate::merkle::verify_range_proof;
 use crate::proofs::ProofError;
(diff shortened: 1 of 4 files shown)


firewood/src/proofs/range.rs
```diff
@@ -125,37 +128,94 @@ pub fn verify_range_proof_structure(
...
         return Ok(None);
     }
 
-    if let Some(ref end_key) = verification.end_key
-        && **last_key >= **end_key
+    // Case 4: terminal's nibble path differs from `last_key` unpacked
+    // into nibbles → exhaustive.
+    if !terminal
+        .full_path()
+        .path_eq(&PathBuf::path_from_packed_bytes(last_key))
```
rkuris, line 207: nit: Seems like pulling this into a helper would make sense, especially after looking briefly at #2010.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,44 @@ fn test_slow_malformed_proof_fuzz() {
         }
     }
 }
+
+#[test]
```
rkuris, line 920: This file already uses #[test_case] extensively, and this is another case where it makes sense. Something like (untested):

```
#[test_case(b"",  &[0x00]; "empty key")]
#[test_case(b"\x10",   &[0x10, 0x00]; "single byte")]
#[test_case(b"\xff",   &[0xff, 0x00]; "max byte (no overflow)")]
```

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,44 @@ fn test_slow_malformed_proof_fuzz() {
...
+fn test_lex_successor_max_byte() {
+    // No overflow: `succ([0xff])` is just `[0xff, 0x00]`.
+    assert_eq!(lex_successor(b"\xff").as_ref(), [0xff, 0x00].as_slice());
+}
+
+#[test]
+fn test_lex_successor_is_strictly_greater() {
+    let keys: &[&[u8]] = &[b"", b"\x00", b"\x10", b"\xff", b"abc"];
```
rkuris, line 938: Same problem here:
```suggestion
#[test_case(b""; "empty key")]
#[test_case(b"\x10"; "single byte")]
#[test_case(b"\xff"; "overflow")]
fn test_lex_successor_is_strictly_greater(key: &[u8]) {
```

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,44 @@ fn test_slow_malformed_proof_fuzz() {
...
+    for k in keys {
+        let s = lex_successor(k);
+        assert!(s.as_ref() > *k, "succ({k:?}) = {s:?} is not > input");
+    }
+}
+
+#[test]
+fn test_lex_successor_no_key_in_between() {
```
rkuris, line 946: This test is AI noise. It doesn't actually test that there is no key in between the provided key and the generated key (it could, but I think the claim is obvious from the results).

These comments might be valuable as doc comments on `lex_successor` or might just be more noise.

firewood/src/proofs/range.rs
```diff
@@ -125,37 +128,94 @@ pub fn verify_range_proof_structure(
 
 /// Determine the next key range to fetch after this range proof.
 ///
-/// Returns `None` when the originally-requested range is fully accounted
-/// for; otherwise returns `Some((last_key, end_key))`.
+/// Returns `None` when the proof's structure proves the requested range is
```
rkuris, line 131: This is incredibly noisy and conflates what should be implementation details and inline comments with the documentation for this function. Maybe something more like this:

```
/// Determine the next key range to fetch after this range proof, by
/// inspecting the proof's structure.
///
/// Assumes `proof` has been verified by the caller; this function does
/// not itself perform cryptographic verification.
///
/// # Returns
///
/// - `Ok(None)` when the proof's structure proves the requested range
///   is exhaustively covered.
/// - `Ok(Some((cursor, end_key)))` when a continuation may be needed.
///   This is conservative — a follow-up fetch may return no keys —
///   but `cursor` is always strictly past `last_key`, so repeated calls
///   make guaranteed forward progress.
///
/// # Errors
///
/// - [`ProofError::EndKeyLessThanLastKey`] — `last_key > end_key`. A
///   malformed-proof condition that verification should have caught
///   upstream; surfaced here as defense in depth.
```

I'm worried we're going to have to make much bigger changes if we make this method better in the future, so let's just focus on what the contract is and leave the implementation details in the function. Good idea normally, especially important here.

rkuris (changes requested):

PR #2013, opened 2026-05-22
feat(iter): Implement cycle detection in trie traversal

### LINKED ISSUE #1334: If there is a loop in the merkle trie, firewood can hang
Needed for hardening.

It seems like if we've traversed some maximum depth, we should stop with an error. I recommend a maximum depth of 1024.

demosdemon:
Fixes #1334

## Why this should be merged

A corrupted or adversarially-crafted trie could contain back-edges (child disk addresses
that point to an ancestor node), causing `MerkleNodeIter` and `PathIterator` to loop
infinitely. This change turns that silent hang into a recoverable `CycleDetected` error,
keeping the database usable and allowing callers to surface the corruption rather than
spinning forever.

A secondary `MAX_TRIE_DEPTH = 1024` guard also caps descent into pathologically deep
(but not strictly cyclic) tries, covering all child edge types including in-memory
`Child::Node` edges that have no disk address and cannot be tracked by the visited-address map.

## How this works

- **`storage/src/cycle.rs`** — new `CycleDetected` error type carrying the back-edge disk
  address and the nibble-paths to both the first visit and the revisit, for diagnosability.
- **`storage/src/linear/mod.rs`** — `FileIoError::from_cycle` wraps a `CycleDetected` as an
  `std::io::Error`; `FileIoError::as_cycle_detected` recovers it via `downcast_ref`, letting
  callers distinguish cycle errors from ordinary I/O failures. `FileIoError::Display` suppresses
  the `"at offset X of file 'Y'"` clause when no file context is present (e.g. cycle errors),
  avoiding misleading `"at offset 0 of file '[unknown]'"` output.
- **`firewood/src/iter.rs`** (`MerkleNodeIter` + `PathIterator`) — each iterator maintains a
  `visited: HashMap<LinearAddress, PathBuf>` populated before every `read_node` call. The root
  address is pre-recorded before descent begins so a direct root self-loop is caught immediately.
  The depth guard in `load_and_track_child` covers all child arms including in-memory
  `Child::Node` and in-memory `Child::MaybePersisted` edges. A separate depth guard in the seek
  phase (`get_iterator_initial_state`) prevents unbounded descent on adversarial keys. Both
  iterators transition to a terminal state after any error so subsequent `next()` calls return
  `None`.

## How this was tested

- **Unit tests in `storage/src/linear/mod.rs`**: `CycleDetected` roundtrips correctly through
  `FileIoError`; regular I/O errors do not downcast as cycles.
- **Unit tests in `firewood/src/iter.rs`**:
  - Self-loop (root → root), two-node cycle, and three-node cycle in `MerkleNodeIter`
  - Cycle path context (`path_to_first_visit` / `path_to_revisit`) verified with exact path lengths
  - Cycle detection fires during the seek phase (`get_iterator_initial_state`) with a non-empty key
  - Cycle detection works through `Child::MaybePersisted` edges (not only `AddressWithHash`)
  - `PathIterator` detects cycles and includes correct path context
  - No false positives on acyclic two-child tries (both iterators)
  - Depth guard fires at exactly `MAX_TRIE_DEPTH` for both iterators
  - Regression: branch at seek depth `MAX_TRIE_DEPTH` is yielded before the depth guard fires
  - Post-error termination: both iterators return `None` after a cycle error or depth-exceeded error

## Breaking Changes

None. The new error is additive and the iterator changes are all internal.

### DIFF
--- a/firewood/src/iter.rs
+++ b/firewood/src/iter.rs
@@ -5,11 +5,19 @@ use crate::api::{KeyType, KeyValuePair};
 use crate::merkle::{Key, Value};
 
 use firewood_storage::{
-    BranchNode, Child, FileIoError, NibblesIterator, Node, PathBuf, PathComponent, PathIterItem,
-    SharedNode, TriePathFromUnpackedBytes, TrieReader,
+    BranchNode, Child, CycleDetected, FileIoError, LinearAddress, NibblesIterator, Node, PathBuf,
+    PathComponent, PathIterItem, SharedNode, TriePathFromUnpackedBytes, TrieReader,
 };
 use std::cmp::Ordering;
+use std::collections::HashMap;
 use std::iter::FusedIterator;
+
+/// Maximum trie depth allowed during iteration.
+///
+/// This acts as a secondary defense against pathologically deep (but not cyclic) tries,
+/// preventing stack exhaustion or runaway allocation. A well-formed Firewood trie should
+/// never approach this depth.
+const MAX_TRIE_DEPTH: usize = 1024;
 
 /// Represents an ongoing iteration over a node and its children.
 enum IterationNode {
@@ -65,6 +73,11 @@ enum NodeIterState {
 pub struct MerkleNodeIter<'a, T> {
     state: NodeIterState,
     merkle: &'a T,
+    /// Maps each visited disk address to the trie nibble-path at which it was
+    /// first seen. Checked before every `read_node` call so that back-edges
+    /// (cycles) are detected and returned as [`FileIoError`]s rather than
+    /// causing infinite iteration.
+    visited: HashMap<LinearAddress, PathBuf>,
 }
 
 impl From<Key> for NodeIterState {
@@ -80,6 +93,7 @@ impl<'a, T: TrieReader> MerkleNodeIter<'a, T> {
         Self {
             state: NodeIterState::from(key),
             merkle,
+            visited: HashMap::new(),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


firewood/src/iter.rs
```diff
@@ -584,11 +754,7 @@ pub(crate) fn key_from_nibble_iter<Iter: Iterator<Item = u8>>(mut nibbles: Iter)
     let mut data = Vec::with_capacity(nibbles.size_hint().0 / 2);
 
     while let (Some(hi), Some(lo)) = (nibbles.next(), nibbles.next()) {
-        let byte = hi
-            .checked_shl(4)
-            .and_then(|v| v.checked_add(lo))
-            .expect("Nibble overflow while constructing byte");
-        data.push(byte);
+        data.push((hi << 4) | lo);
```
rkuris, line 757: Why was this change needed?

rkuris (commented):
This is a comment on an incomplete review before a long weekend. Do not assume this review is complete; providing early feedback based on my first look at this PR and have some questions.

Let's assume we're walking the entire trie using `MerkleNodeIter` (either fwdctl's dump or some other db.iter call) and we're using ethereum hashing, with 32 or 64 byte keys (most are 32). I think these steps are taken:

1. In the iterator, at the point when we hit a visited node (lines around 140-148) we pop the `Visited` entry off the stack and pull the next child. `key` at this point is the full nibble path (1 byte per nibble) from this root to the parent.
2. We build a fresh `PathBuf` of size `key.len() + 1`.
3. build_child_path_refix collects these nibbles. For ethereum's 64-nibble smaller keys this exceeds the SmallVec that PathBuf uses (32 nibbles) so we make an allocation on the heap.
4. This newly allocated heap is then passed into `resolve_child` by value (moved), who then stores it in the HashMap as the value based on the LinearAddress. This is used solely for the diagnostic error when we hit the failure case.

This is by inspection, and I haven't run the code, but it does seem like this allocates a ton of memory as we're traversing -- one LinearAddress->PathBuf entry for every key. In a large trie, this means the memory requirements for the checker or the dump utility have increased significantly.

I'm probably wrong about something here, but at a minimum this has an extra hard allocation and uses more stack during node traversal so it would warrant a benchmark run.

If you disagree, please let me know what assumption here is wrong.

PR #2026, opened 2026-05-26
build(ffi): vendor golangci-lint as go tool, drop third-party action

rkuris:
## Why this should be merged

Two goals:

1. **Allow local reproduction of the CI golangci-lint step.** The `golangci/golangci-lint-action` GitHub Action precludes this. Vendoring the linter as a Go tool dependency lets a contributor run the exact same lint locally with `./ffi/scripts/lint.sh`.
2. **Stop pulling third-party TypeScript** (the action) into our supply chain, which we are poorly equipped to audit.

Bonus: the current step is `version: latest`, so the linter binary (and its transitive deps) silently floats from one run to the next. Vendoring pins it.

This follows the avalanchego pattern:
- [`tools/external/go.mod`](https://github.com/ava-labs/avalanchego/blob/master/tools/external/go.mod#L22) — tool deps live in a separate module to avoid polluting the main module's dependency graph.
- [`scripts/lint.sh`](https://github.com/ava-labs/avalanchego/blob/master/scripts/lint.sh#L48) / [`scripts/run_tool.sh`](https://github.com/ava-labs/avalanchego/blob/master/scripts/run_tool.sh) — thin wrappers around `go tool -modfile=...`.

Also tightens `.github/dependabot.yml` to **security-only** updates for `github-actions` (mirroring the existing cargo policy), matching avalanchego's approach where dependabot only opens PRs for security fixes.

## How this works

- New module `ffi/tools/external/go.mod` pinning `github.com/golangci/golangci-lint/v2 v2.9.0` as a `tool` dependency (same version avalanchego is on).
- `ffi/scripts/run_tool.sh` — `GOWORK=off go tool -modfile=ffi/tools/external/go.mod "$@"`. `GOWORK=off` because `-modfile` is incompatible with workspace mode, and Go walks parent dirs looking for `go.work` — so we set the env var defensively, not because there's a `go.work` today.
- `ffi/scripts/lint.sh` — `./scripts/run_tool.sh golangci-lint run --config .golangci.yaml ./...`, scoped to `ffi/`.
- `.github/workflows/ci.yaml`: the `Run golangci-lint` step is now `run: ./ffi/scripts/lint.sh` instead of `uses: golangci/golangci-lint-action@...`. The existing `actions/setup-go` step picks up `ffi/tools/external/go.sum` for module caching alongside `ffi/go.sum`.
- `.github/dependabot.yml`: `github-actions` `open-pull-requests-limit: 10` → `0`. Dependabot still raises advisories; routine bump PRs go away.
- `.claude/skills/firewood-review/SKILL.md`: reviewer skill now invokes `./scripts/lint.sh` instead of `golangci-lint` from `PATH`; compatibility precondition swaps `golangci-lint` for `go`.

## How this was tested

- `./ffi/scripts/lint.sh` locally → `0 issues.`
- Confirmed `go.mod` builds and resolves under `GOWORK=off`.

## First-run cost

Fresh checkouts (and fresh CI caches) pay a ~2–4 min one-time cost to compile `golangci-lint` from source. Subsequent runs hit `~/.cache/go-build` and are equivalent to running the prebuilt binary directly. CI's `actions/setup-go` cache keeps it warm across runs.


alarso16 (approved):
Seems good to me, but I'm no bash expert

.github/dependabot.yml
```diff
@@ -28,4 +28,4 @@ updates:
       github-actions:
         patterns:
           - "*"
-    open-pull-requests-limit: 10 # Enable GitHub Actions updates for security
+    open-pull-requests-limit: 0 # Disable non-security version updates
```
demosdemon, line 31: this will require us to manually bump the `dtolnay/rust-toolchain` action after every major stable rust release. the `stable` branch and all of the version branches are floating commit are always one commit ahead from `main`.

demosdemon (approved):

ffi/scripts/run_tool.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
maru-ava, line 1: Given that you're only running golangci-lint maybe skip this script and run `go tool` directly in lint.sh?

ffi/scripts/run_tool.sh
```diff
@@ -0,0 +1,9 @@
...
+
+set -euo pipefail
+
+FFI_PATH="$(cd "$( dirname "${BASH_SOURCE[0]}" )"; cd .. && pwd )"
+
+# GOWORK=off is required because -modfile is incompatible with workspace mode,
```
maru-ava, line 7: GOWORK is only required if the repo has a go.work file, which this repo does not.

maru-ava (commented):
Functionally looks fine, just a couple of script suggestions.

ffi/scripts/run_tool.sh
rkuris: I thought about this, but I think I'd rather do things closer to how avalanchego does them. If you feel strongly about this, I can definitely change it.

ffi/scripts/run_tool.sh
rkuris: I don't know this area well, but Claude seems to want to defend it. If you have a go.work file anywhere up the tree, it will look at it, as well as future-proofing anyone who decides to make a workspace later.

LMK if you think it should be changed before I merge, I don't think it's that big a deal either way.

.github/dependabot.yml
rkuris: Agreed but we've been watching them like a hawk anyway.

ffi/scripts/run_tool.sh
maru-ava: I think it's strange to be this defensive unless a repo is actually using go.work. An external consumer can always GOWORK=off outside the script. We didn't set this until we adopted go.work to enable bazelification. But sure.

maru-ava (approved):

PR #2028, opened 2026-05-26
fix(proofs): annotate edge-proof hash mismatches with Left/Right edge

rkuris:
## Why this should be merged

Previously, any hash mismatch during proof verification surfaced as the bare error "proof error: unexpected hash" with no detail. That's useless for diagnosing a tampered or mis-rooted proof — you can't tell which hash didn't match, or, for range/change proofs, which edge tripped.

## How this works

- `ProofError::UnexpectedHash` is now a struct variant carrying `expected: HashType` and `actual: HashType`. The Display impl renders both.
- New public enum `ProofEdge { Left, Right }`.
- New variant `ProofError::EdgeProofHashMismatch { edge, expected, actual }` for range/change proof edge failures.
- `verify_edge` (range proofs) and `verify_boundary_proof` (change proofs) intercept `UnexpectedHash` from the inner `value_digest` walk and remap it to `EdgeProofHashMismatch` with the correct edge tag.
- The reconstructed-root check in `verify_range_proof_root_hash` continues to return plain `UnexpectedHash` (no edge applies there) with the populated expected/actual fields.

## How this was tested

- `cargo nextest run -p firewood --features ethhash,logger --all-targets` — 395 passed.
- `cargo clippy -p firewood --features ethhash,logger --all-targets` — clean.
- Existing test `test_boundary_proof_rejects_wrong_root` updated to match on the new edge-annotated variant; existing range-proof tests that matched on `UnexpectedHash` now match the struct shape.

## Breaking Changes

- `ProofError::UnexpectedHash` changed from a unit variant to a struct variant. External matchers on this variant must add `{ .. }` or destructure the fields.
- New variant `ProofError::EdgeProofHashMismatch` — already `#[non_exhaustive]`, so no break for exhaustive matches, but callers that previously caught "hash mismatch on edge proof" via `UnexpectedHash` will now see the new variant instead.


firewood/src/proofs/mod.rs
```diff
@@ -308,7 +308,7 @@ pub use self::range::{
 };
 pub use self::reader::ReadError;
 pub use self::types::{
-    EmptyProofCollection, Proof, ProofCollection, ProofError, ProofNode, ProofType,
+    EmptyProofCollection, Proof, ProofCollection, ProofEdge, ProofError, ProofNode, ProofType,
```
demosdemon, line 311: `ProofEdge` is re-exported from the `proofs` submodule here, but not from the crate root in `firewood/src/lib.rs`. Every sibling proof type in that block is re-exported at the crate root for ergonomic access:

```rust
// firewood/src/lib.rs around line 157
pub use proofs::{
    ChangeProof, ChangeProofVerificationContext, EmptyProofCollection, InvalidHeader, KeyRange,
    Proof, ProofCollection, ProofError, ProofNode, ProofType, RangeProof,
    ...
};
```

The new test at `firewood/src/merkle/tests/change/structural.rs:232` already pays for this asymmetry — it spells `crate::proofs::ProofEdge::Left` instead of `crate::ProofEdge::Left`. External callers that pattern-match on `ProofError::EdgeProofHashMismatch { edge, .. }` will need a second import path.

Suggest adding `ProofEdge` to the `pub use proofs::{...}` list in `lib.rs`.

firewood/src/merkle/tests/change/structural.rs
```diff
@@ -228,7 +228,10 @@ fn test_boundary_proof_rejects_wrong_root() {
     .unwrap_err();
     assert!(matches!(
         err,
-        api::Error::ProofError(crate::ProofError::UnexpectedHash)
+        api::Error::ProofError(crate::ProofError::EdgeProofHashMismatch {
+            edge: crate::proofs::ProofEdge::Left,
+            ..
+        })
```
demosdemon, line 258: Test coverage gap: only the `Left` edge of `EdgeProofHashMismatch` is exercised, and only through the change-proof path. The `Right` arm and the range-proof's `verify_edge` (in `firewood/src/merkle/mod.rs`) have **zero** assertions on the new variant.

`rg 'EdgeProofHashMismatch' firewood/src/merkle/tests` returns just this one site. A regression that swaps `ProofEdge::{Left,Right}` at either `proofs/change.rs:411,453` or flips the `bound_is_lower` mapping at `merkle/mod.rs:160-164` would not be caught by any test.

Two additions worth considering:

1. A `test_boundary_proof_rejects_wrong_root` companion that tampers the end proof and asserts `edge: ProofEdge::Right`.
2. A range-proof test that tampers each edge in turn (start vs end proof hash) and asserts the matching `ProofEdge` value through `verify_range_proof`. `firewood/src/merkle/tests/range.rs` already has tampering helpers (`test_tampered_in_range_value_rejected`, etc.) that could be patterned.

Without a Right-edge test, the whole point of the annotation — telling which side tripped — is unverified by the suite.

firewood/src/merkle/mod.rs
```diff
@@ -156,13 +157,33 @@ fn verify_edge<H: ProofCollection + ?Sized>(
...
+    let annotate = |e: api::Error| match e {
+        api::Error::ProofError(ProofError::UnexpectedHash { expected, actual }) => {
+            api::Error::ProofError(ProofError::EdgeProofHashMismatch {
+                edge,
+                expected,
+                actual,
+            })
+        }
+        other => other,
+    };
```
demosdemon: Two related cleanups here.

**(a)** The `bool` → `ProofEdge` conversion is plumbing the wrong type through `verify_edge`. The sibling `verify_boundary_proof` in `proofs/change.rs` already takes `edge: ProofEdge` directly (see line 230). Lifting `ProofEdge` into `verify_edge`'s own signature deletes lines 160-164 entirely and lets both call sites in `firewood/src/merkle/mod.rs` (~718, ~740) pass `ProofEdge::Left` / `ProofEdge::Right` explicitly instead of `true` / `false`.

**(b)** `annotate` takes `api::Error` but every caller wraps a `ProofError` into `api::Error::ProofError(_)` just so the closure can destructure it back out. Taking `ProofError` directly mirrors how the equivalent intercept in `verify_boundary_proof` is written and is materially clearer.

Combined suggestion (applies to lines 160-174):

```suggestion
    // Any `UnexpectedHash` bubbling up from this edge's `verify()` walk is
    // by construction an edge-proof failure — re-stamp it with which edge.
    let annotate = |e: ProofError| match e {
        ProofError::UnexpectedHash { expected, actual } => {
            ProofError::EdgeProofHashMismatch { edge, expected, actual }
        }
        other => other,
    };
```

…paired with `bound_is_lower: bool` → `edge: ProofEdge` in the signature, and the call-site map_errs becoming `.map_err(|e| api::Error::ProofError(annotate(e)))?`.

firewood/src/proofs/types.rs
```diff
@@ -55,6 +55,24 @@ use thiserror::Error;
 
 use crate::merkle::Value;
 
+/// Which edge of a range proof was being verified when an error occurred.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProofEdge {
+    /// The start (lower-bound) proof.
```
demosdemon, line 62: Minor: consider deriving `Hash` and marking the enum `#[non_exhaustive]` for parity with sibling proof types (`ProofError` is `#[non_exhaustive]`, `ProofType` too). `Hash` is cheap on a 2-variant POD enum and lets callers key collections by edge if they ever need to. `#[non_exhaustive]` is free forward-compat if a third edge concept (e.g. `Both` for symmetric failures) ever shows up.

```suggestion
/// Which edge of a range proof was being verified when an error occurred.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
#[non_exhaustive]
pub enum ProofEdge {
```

firewood/src/proofs/types.rs
```diff
@@ -63,9 +81,28 @@ pub enum ProofError {
     #[error("non-monotonic range increase")]
     NonMonotonicIncreaseRange,
 
-    /// Unexpected hash
-    #[error("unexpected hash")]
-    UnexpectedHash,
+    /// A node's hash did not match the hash expected by its parent (or the
+    /// root hash for the first proof node).
+    #[error("hash mismatch (expected {expected}, was {actual})")]
```
demosdemon, line 87: Doc drift: the rustdoc on `Proof::value_digest` (around line 474 in this file, unchanged in the diff) still lists `ProofError::UnexpectedHash` without acknowledging that boundary callers now re-stamp it as `EdgeProofHashMismatch`. A reader of `value_digest`'s docs would still expect `UnexpectedHash` for a tampered edge proof.

Suggested addition near that doc block:

> Note: callers verifying a range or change proof's edge proof (e.g.
> `verify_edge`, `verify_boundary_proof`) intercept `UnexpectedHash`
> from this walk and surface `ProofError::EdgeProofHashMismatch` instead,
> annotated with which edge tripped.

firewood/src/proofs/change.rs
```diff
@@ -226,10 +227,18 @@ fn verify_boundary_proof<C: ProofCollection>(
...
     let result = match proof.value_digest(key, end_root) {
         Ok(result) => result,
         Err(ProofError::Empty) => None,
+        Err(ProofError::UnexpectedHash { expected, actual }) => {
+            return Err(api::Error::ProofError(ProofError::EdgeProofHashMismatch {
+                edge,
+                expected,
+                actual,
+            }));
+        }
```
demosdemon, line 243: Minor: a one-line invariant pin would help future readers understand why this intercept is safe. The transformation is correct because *any* `UnexpectedHash` produced by `proof.value_digest(key, end_root)` is, by definition, a failure at this specific boundary — but that's only obvious if you've traced the call graph. A comment removes the burden.

```suggestion
        // Any `UnexpectedHash` from this boundary `value_digest` walk is by
        // construction an edge-proof failure — re-stamp it with which edge.
        Err(ProofError::UnexpectedHash { expected, actual }) => {
            return Err(api::Error::ProofError(ProofError::EdgeProofHashMismatch {
                edge,
                expected,
                actual,
            }));
        }
```

demosdemon (approved):

firewood/src/merkle/mod.rs
rkuris: Fixed in a53fb08d. I also hate bare booleans, great suggestion — `verify_edge` now takes `ProofEdge` directly and the bool→enum mapping is gone, and `annotate` takes `ProofError` like `verify_boundary_proof`.

firewood/src/merkle/tests/change/structural.rs
rkuris: Fixed in dcaecc84. Great find — the Right arm and range-proof `verify_edge` had zero assertions. Added a change-proof Right companion (`test_boundary_proof_rejects_wrong_root_right_edge`) plus `test_range_proof_{left,right}_edge_hash_mismatch` that tamper one edge proof's terminal node and assert the matching `ProofEdge`.

firewood/src/proofs/mod.rs
rkuris: Fixed in 91a28c24. `ProofEdge` is now re-exported from the crate root in lib.rs, and the test sites drop `crate::proofs::ProofEdge` for `crate::ProofEdge`.

firewood/src/proofs/types.rs
rkuris: Fixed in 91a28c24. Added `Hash` and `#[non_exhaustive]`.

firewood/src/proofs/types.rs
rkuris: Fixed in 91a28c24. Added a `# Note` on `value_digest` documenting that edge callers intercept `UnexpectedHash` and surface `EdgeProofHashMismatch`.

firewood/src/proofs/change.rs
rkuris: Fixed in 91a28c24.

PR #2029, opened 2026-05-27
feat: add update convenience API

dvansari65:
## Why this should be merged
Race condition fix: The old way (propose → commit → get root) allowed other threads to sneak in and commit between your commit and root read, giving you the wrong root. Now it's all one call—no interruption

What We Did
Added Db::update(batch) method that applies a batch, commits it, and returns the root hash in one atomic operation.

## How this works
The update() method captures the root hash before committing the proposal. Here's the flow:

propose(data) - Creates a proposal with the batch
DbView::root_hash(&proposal) - Captures the root hash from the proposal (before commit)
proposal.commit() - Commits the changes
Returns the captured root hash
This ensures the returned root is always from the revision you just committed, even if other threads commit concurrently.

## How this was tested
Three test cases cover:

test_update_commits_batch_and_returns_root - Verifies basic functionality: batch is committed and returned root matches db.root_hash(). Also tests sequential updates produce different roots and historical revisions remain unchanged.

test_update_returns_empty_root_after_delete_all - Tests edge case: deleting all entries returns the default/empty root hash.

Both tests verify data integrity across multiple revisions using db.revision().

## Breaking Changes
None. This is a new convenience method that doesn't modify existing APIs—purely additive.

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/api.rs
+++ b/firewood/src/api.rs
@@ -307,6 +307,19 @@ pub trait Db {
     ///   operations to apply
     #[expect(clippy::missing_errors_doc)]
     fn propose(&self, data: impl IntoBatchIter) -> Result<Self::Proposal<'_>, Error>;
+
+    /// Apply a batch to the latest revision, commit it, and return the committed root hash.
+    ///
+    /// This is a convenience method for the common one-shot write path. It returns the
+    /// proposal's root hash instead of reading [`Db::root_hash`] after commit, so callers
+    /// receive the root for the revision this method committed.
+    #[expect(clippy::missing_errors_doc)]
+    fn update(&self, data: impl IntoBatchIter) -> Result<Option<HashKey>, Error> {
+        let proposal = self.propose(data)?;
+        let root_hash = DbView::root_hash(&proposal);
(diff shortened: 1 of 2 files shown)


demosdemon:
Hi @dvansari65, this is a great addition. Could you update the PR description and [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

demosdemon (approved):

rkuris:
side note that you might have instead wanted commit_with_rebase which lets you do more things in parallel here -- it's an atomic "merge this and rebase it because I know it doesn't overlap" and provides better concurrency when people are working on independent revisions.

With this fix, you can end up in a case where just before you start executing things someone else committed and you didn't see those changes, so you're applying this batch blindly against some revision you know nothing about. That was the whole point of commit_with_rebase, and allows for parallel committers without holding that global lock while you're preparing the proposals.

PR #2030, opened 2026-05-27
fix(proofs): hash single storage child proofs as root

JonathanOppenheimer:
## Why this should be merged

**Note: This probably should not be merged, and is most likely AI slop. I hope it's still useful to some extent. @claude  wrote the below and the code, but I had fun with this haha.**

In ethhash mode, an account branch with exactly one storage child is hashed using the "fake root" trick documented in `storage/src/hashers/ethhash.rs`: the account-child nibble folds into the child's partial path, and the child is re-hashed as a standalone storage-trie root. Live hashing applies this in `nodestore::hash::hash_helper_inner`, but range/change proof verification did not — so when a range proof was truncated such that only a single storage child of an account remained visible, the computed root mismatched the on-disk root and verification spuriously failed with `UnexpectedHash`.

This surfaces on real C-Chain workloads any time a `range_proof` `limit` lands inside an account whose storage subtree (post-truncation) collapses to a single child.

## How this works

- `compute_root_hash_with_proofs` now takes a `fake_root` flag. Before recursing into each branch child, `single_effective_account_child` checks whether the parent is a depth-64 (account) branch whose in-range live children plus outside-range proof-node children together resolve to exactly one child; if so, that one child is recursed into with `fake_root = true`.
- The new `hash_proof_node_parts` helper centralizes the leaf/branch hash construction and, when `fake_root` is set, splits the last nibble off `path_prefix` and folds it into the partial path via `JoinedPath` before hashing — exactly mirroring `hash_helper_inner`'s `fake_root_extra_nibble` path.
- `Proof::verify` does the analogous rewrite on the linear proof walk: after stepping through a depth-64 node whose own child mask is a singleton, the next proof node is hashed via `proof_node_hash(..., fake_root = true)`. The `is_single_child_account_node` predicate counts the node's own children directly because the proof walk sees the authoritative child mask at each step (no range-visibility bookkeeping needed, unlike the recursive computation above).
- Both behaviors are `#[cfg(feature = "ethhash")]`-gated; in non-ethhash builds `fake_root` is threaded through but never set, and the hash math is byte-identical to before.

## How this was tested

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets -- -D warnings` (no features and `--features ethhash,logger`)
- `cargo doc --no-deps --workspace --features ethhash,logger`
- Two new range-proof regression tests in `firewood/src/merkle/tests/ethhash.rs`:
  - `test_limit_truncated_range_proof_inside_single_storage_child_account` — limit lands inside an account with one storage slot; verifies both the live proof and a serialization round-trip.
  - `test_limit_truncated_range_proof_inside_multi_storage_child_account` — limit lands inside an account with two storage slots, confirming the fake-root rewrite does *not* kick in when it shouldn't.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -369,9 +369,18 @@ fn compute_root_hash_with_proofs(
     path_prefix: &[PathComponent],
     proof_nodes: &HashMap<PathBuf, &ProofNode>,
     outside_children: &HashMap<PathBuf, ChildMask>,
+    fake_root: bool,
 ) -> HashType {
     let branch = match node {
-        Node::Leaf(_) => return HashableShunt::from_node(path_prefix, node).to_hash(),
+        Node::Leaf(leaf) => {
+            return hash_proof_node_parts(
+                path_prefix,
+                leaf.partial_path.as_components(),
+                Some(ValueDigest::Value(&leaf.value)),
+                Children::new(),
+                fake_root,
+            );
+        }
         Node::Branch(branch) => branch,
     };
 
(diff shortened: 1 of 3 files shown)


JonathanOppenheimer:
Superseded by https://github.com/ava-labs/firewood/pull/2039

PR #1976, opened 2026-05-06
feat(ffi): Close option to force-close outstanding FFI handles for avalanchego async execution engine

rkuris:
## Why this should be merged

Two things, one of which surfaced while doing the other:

**Pre-existing use-after-free in `Iterator`** — `Iterator` was not participating in the keep-alive infrastructure at all. `db.Close(ctx)` could silently succeed with live iterators, after which `Iterator.Next` was dereferencing freed Rust memory. This PR closes that gap by bringing iterators under the same keep-alive registry as `Proposal`, `Revision`, and `Reconstructed`.

**`WithForceCloseHandles` option for `Database.Close`** — `Close` currently fails with `ErrActiveKeepAliveHandles` whenever any view-like handle has not been explicitly released. The avalanchego async execution engine tracks handles on the Go side (or via GC) and wants an opt-in to have `Close` actively release them, rather than erroring out.

## How this works

`Database.Close` becomes variadic: `Close(ctx context.Context, opts ...CloseOption) error`. The signature is source-compatible with every existing `db.Close(ctx)` call. `WithForceCloseHandles()` makes Close forcibly drop every outstanding handle.

### Keep-alive registry

A `keepAliveRegistry` lives on `Database` and tracks every outstanding handle that holds a borrow on the DB. It owns three pieces of state:

- An outstanding-handle **count**, plus a `[]chan struct{}` of **waiters**. This replaces `sync.WaitGroup`. `WaitGroup.Wait` has no context, so the conventional ctx-aware wrapper spawns a goroutine per `Close`; if `ctx` fires first the goroutine leaks until every outstanding handle eventually releases — which is exactly the failure mode `Close` is trying to surface. The explicit count drains via per-call channels that are unregistered on `ctx` cancellation, so a failed-and-retried `Close` costs nothing.
- A **map of drop callbacks** keyed by lease pointer. `WithForceCloseHandles` walks this map and invokes each callback.
- A **`closed` flag**. Once set (under the registry mutex, atomically with snapshotting the drop map), `lease.attach` returns `errDBClosed` and runs the dropFn itself, so a derived-handle constructor that races with `Close` cannot leak its C handle.

### `lease` and `handle[T]`

The per-handle bookkeeping is a `lease` (a back-pointer to the registry, plus a `RWMutex` that every public method on the wrapping handle takes for read). `lease.attach(registry, dropFn)` is the single atomic activation point — count++ and map insert happen under the same registry critical section, so there is no window where a handle holds the count but isn't drop-callable.

[...]

PR #2032, opened 2026-05-28
fix(ffi): avoid false close error and stranded lease on panic

powerslider:
## Why this should be merged

Should be merged in #1976 or parts of it that resolve correctness issues and/or introduce better coverage for regressions.

## How this works

Resolve correctness issues found while reviewing the force-close work:

- `Database.Close` no longer returns `ErrActiveKeepAliveHandles` when force-closing with no live handles under an already-cancelled context. `waitForKeepAlives` is now the sole check for remaining handles and returns nil once the count drains, regardless of ctx state.
- `lease.release` releases the lease even when its callback panics, so a panicking free can no longer strand the lease and block `Close` forever.
- `iteratorHandle.Drop` marks the handle dropped before freeing, matching `handle[T].Drop`, so a panic in free cannot leave a retry able to double-free.
- `removeAndDecr` defers its unlock, matching `attachUnregistered` and preventing a stranded registry lock on panic.
- Restore `fmt.Errorf` for `errFreeingValue` wraps where the inner error is the cause rather than an independent error.
- Document the `Iterator` single-reader contract and that force-close does not auto-drop verified `RangeProof`s.

## How this was tested

Add tests for the force-close edge cases, panic-safe release, closed-registry attach, `Iterator`/`Reconstructed` finalizer back-stops, a live `RangeProof` under force-close, and partial-drain-then-retry.

## Breaking Changes
none

Signed-off-by: Tsvetan Dimitrov (tsvetan.dimitrov@avalabs.org)

### DIFF
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -485,9 +485,10 @@ type closeConfig struct {
 // [RangeProof.Free] before Close can complete.
 //
 // Each handle is dropped once even if its underlying free errors, so a
-// failing free cannot stall the close. Any drop errors are joined and
-// returned, and the underlying database close is still attempted on a
-// best-effort basis.
+// failing free cannot stall the close. Drops run sequentially under each
+// handle's lock, so an in-flight reader can delay the close past the
+// deadline. Drop errors are joined and returned, and the database close
+// is still attempted on a best-effort basis.
 //
 // While force-close is in progress, attempts to construct derived handles
 // (for example via [Proposal.Propose] or [Revision.Iter]) fail with an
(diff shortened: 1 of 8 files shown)


ffi/firewood.go
```diff
@@ -538,16 +539,11 @@ func (db *Database) Close(ctx context.Context, opts ...CloseOption) error {
 	var forcedDropErr error
 	if cfg.forceCloseHandles {
 		forcedDropErr = db.keepAlives.closeAndForceDrop(ctx)
-		if ctx.Err() != nil {
```
powerslider, line 541: Force-close under a cancelled context returns `ErrActiveKeepAliveHandles` even when no handles are outstanding. An extra `ctx.Err()` check fires before `waitForKeepAlives`, which already returns nil when nothing is left to wait for. That is why I removed the check and let `waitForKeepAlives` decide.

ffi/keepalive.go
```diff
@@ -339,11 +340,19 @@ func (l *lease) attachUnregistered(registry *keepAliveRegistry) {
...
-	err := attemptDisown()
+	// Release on panic too, so a panicking callback cannot strand the
+	// lease and block Close forever.
+	defer func() {
+		if p := recover(); p != nil {
+			l.releaseLocked()
+			panic(p)
+		}
+	}()
+	err = attemptDisown()
```
powerslider, line 355: If `attemptDisown` panics, `releaseLocked` never runs, so the lease is left registered with its count still held, and any later graceful close waits on it forever. The fix is to release the lease on the panic path as well and then re-raise. That follows the reasoning the PR already applies elsewhere: a Close that hangs is worse than a free that errored. Let me know if this line of thought makes sense.

rkuris (approved):
Reviewed all six fixes — the concurrency reasoning checks out (waitForKeepAlives as sole drain authority with the dual-select, panic-safe release defer ordering, dropped-before-free matching handle[T].Drop, deferred unlock in removeAndDecr, and the %w: %w restoration). Lint and `go test -race` are clean locally. Thanks for catching these!

Two non-blocking notes, fine as follow-ups:
- `TestCloseForceDoesNotDropRangeProof` waits out the full 1s context on every run since the RangeProof is never auto-dropped; a ~1ms timeout ctx would assert the same failure without the wall-clock cost.
- A one-line comment on why that test uses `newVerifiedRangeProof` + `db.VerifyRangeProof` (to put a lease on the proof) would help readers — especially since calling `VerifyRangeProof` twice on the same proof panics in `attachUnregistered`.

PR #2033, opened 2026-05-28
feat(ffi): code-hash iterator for change proofs

### LINKED ISSUE #1598: Change Proofs: eth compat needs ability to get code hashes from applied proof
#1157 adds for Range Proofs. After Change Proofs are implemented, they should provide code hashes as well.

bernard-avalabs:
Closes #1598.

  ## Why this should be merged

  The `(*ChangeProof).CodeHashes()` Go method was a stub returning an
  empty iterator. Syncers need this to discover which bytecode blobs to
  fetch from peers in parallel with state-trie reconstruction. It plays the 
  same role as the existing range-proof iterator.

  ## How this works
  
  The data is already in the proof: `batch_ops` carries the RLP-encoded
  account values for every `Put`. The iterator parses field [3] out of
  each one, filters the empty code hash sentinel, and yields the rest.
  `Delete` and `DeleteRange` ops carry no value and are skipped.

  No protocol, serialization, or verification changes. The C ABI is
  strictly additive (one new export, `fwd_change_proof_code_hash_iter`).

  Two notes on design choices:

  - **`Box<dyn Iterator>` over an enum.** `CodeIteratorHandle` needs to be
    one concrete type behind cbindgen's opaque pointer but accept both
    range-proof key/value slices and change-proof batch-op slices.
    `IteratorHandle` already uses this pattern via `BoxKeyValueIter`;
    following it keeps the FFI surface singular. Per-`next()` vtable
    lookup is invisible next to RLP parsing.
  - **New `code_hash.rs` module.** `CodeIteratorHandle` previously lived
    in `change.rs` even though range proofs were its only user. Looking at
    code-hash logic for range proofs shouldn't require opening
    `change.rs`. The new module owns the shared bits; per-proof-type entry
    points stay in their respective files.

  ## How this was tested

  - CI

  ## Breaking Changes

  - None. ABI is strictly additive


rkuris (approved):
Looks fine, but let's create an issue to reduce the number ethhash-enabled compile time switches.

bernard-avalabs:
> Looks fine, but let's create an issue to reduce the number ethhash-enabled compile time switches.

https://github.com/ava-labs/firewood/issues/2040

Issue #2008, opened 2026-05-19
Change-proof verification accepts proof with omitted in-range Put (possibly left-edge analog of #1970)


bernard-avalabs:
**Summary**

The unified change-proof fuzz test (test_slow_change_proof_fuzz in firewood/src/merkle/tests/change/fuzz.rs) in https://github.com/ava-labs/firewood/pull/1951 finds cases where an attacker drops an in-range Put from batch_ops and the verifier still accepts the proof. Both the structural check and the root-hash check pass.

This may be related to PR #1970 (which fixed a similar issue on the right edge of the proven range); the cases seen so far all have a non-existent start boundary.

**Reproducer** 

  FIREWOOD_TEST_SEED=16855342578241928557 cargo nextest run -p firewood \
    --features logger --profile ci -E 'test(test_slow_change_proof_fuzz)'

  The failure is deterministic for this seed at run 0.

PR #2039, opened 2026-06-01
fix(proofs): correct ethhash range/change proof verification

bernard-avalabs:
## Why this should be merged
  
Avalanche C-Chain state sync against firewood v0.5.0 fails with `ProofError::UnexpectedHash` on honest range proofs. Root cause: *bug 1* ethhash proof verification diverges from live hashing whenever a verifier has to recompute the hash of a "fake-root" boundary node (an account at depth 64 with exactly one storage child). A second, related defect surfaces in bounded change proofs as `UnexpectedValue` whenever an account's storage children are only partially in range *bug 2*.
  
PR #2030 was correct for bug 1 and fixed this in a different way, by duplicating some code. This fix eliminates that duplication.
  
**Bug 1** — single-storage-child fake-root divergence. Live hashing applies a fold at account-storage boundaries (depth 64, accounts with exactly one storage child) that proof verification didn't. Reproducible without truncation. 

Fix: one shared helper used by every hashing site (live hashing, checker, range/change-proof verification, single-key proof). Previously the fold lived inline in two places and the verifier didn't have it at all.
  
**Bug 2** — partial-storage change proofs. reconcile_branch_proof_node byte-compared the storageRoot field, which Preimage::write recomputes from children at hash time. The check rejected valid proofs because the proposal's recomputed storageRoot differed from the source's full-state storageRoot when batch_ops covered only a subset of an account's storage children. 

Fix: ignore that field when comparing account values at depth 64. The final root-hash check (computed_root == claimed_end_root) is unchanged and remains the security gate. The storageRoot bytes are functionally write-only on the wire because Preimage::write strips and recomputes them.

Range proofs aren't affected: their proving trie is built fresh from key_values, so reconcile compares source bytes against source bytes. Only change proofs fork an already-hashed proposal where update_account_storage_root has already mutated the storageRoot field.
 
## Out of scope
  
  - #2007 (UnexpectedValue in change-proof reconciliation): the documented seed no longer reproduces on current branches, so the relationship can't be confirmed. Bug 2 may or may not be this issue.
  - #2008-class (false-acceptance of dropped batch ops): different code path; one instance (M2_omit_delete) reproduced during this investigation's fuzz runs on main.
  
## Follow-ups to file
  
  1. Strip storageRoot field from account values in proofs (wire-format proposal — coordinated with avalanchego state syncer).
  2. Unify the recursive trie walkers (hash_helper_inner and compute_root_hash_with_proofs are conceptually the same walk over different child resolvers; a ChildResolver trait would close the remaining structural duplication).
  
## Breaking changes
  
  - None. Unit tests pass under both feature configs. The original symptom is only observable against C-Chain reexecution, so state-sync validation against production shape is needed before merge.

### DIFF
--- a/firewood/src/merkle/mod.rs
+++ b/firewood/src/merkle/mod.rs
@@ -30,6 +30,10 @@ use firewood_storage::{
     NibblesIterator, Node, NodeStore, Path, PathBuf, PathComponent, Propose, ReadableStorage,
     SharedNode, TrieHash, TrieReader, U4, ValueDigest,
 };
+#[cfg(feature = "ethhash")]
+use firewood_storage::{
+    hash_node_as_storage_trie_root_for_node, hash_node_as_storage_trie_root_parts,
+};
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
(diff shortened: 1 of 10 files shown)


bernard-avalabs:
Superseded by [2058](https://github.com/ava-labs/firewood/pull/2058) and [2060](https://github.com/ava-labs/firewood/pull/2060)

PR #2043, opened 2026-06-02
chore!: clarify deferred persistence terminology

### LINKED ISSUE #2027: Update deferred persistence terminology
The documentation for the deferred persistence commit count is as follows:

https://github.com/ava-labs/firewood/blob/7db31110ac62e3256fd230b669a3bc49afe8fe5f/ffi/firewood.go#L219-L227

However, this has become a common misnomer when Coreth/Subnet-EVM is introduced into the mix. The EVM has it's own commit-interval field which we use to set the deferred persistence commit count in Firewood, and it's assumed that `commit-interval == persist-interval`. In reality though, setting `commit-interval = N` means that we will persist to disk every `N / 2` commits.

RodrigoVillar:
## Why this should be merged

`deferred_persistence_commit_count` suggests a fixed persistence interval. The setting actually bounds how many committed revisions callers must go back from the latest committed state to reach a persisted state. Renaming it to `max_persistence_gap` makes that contract clearer.

## How this works

Renames the configuration across Rust, C, and Go, including validation errors, internal variables, tests, and benchmarks:

```diff
- deferred_persistence_commit_count
+ max_persistence_gap

- WithDeferredPersistenceCommitCount
+ WithMaxPersistenceGap
```

This PR also updates the documentation to distinguish committing from persisting and explain the existing guarantees:

- The last persisted state is at most `max_persistence_gap` committed revisions behind the latest committed state.
- Commits may wait for persistence to maintain this bound, but a successful commit does not guarantee that its revision has been persisted.
- Setting the gap to 1 ensures every committed revision is persisted before the next state-changing commit completes. Larger values allow intermediate revisions to be skipped.
- Only commits that change the current root count toward the bound; uncommitted proposals do not.
- A successful explicit close persists the latest committed state.

The default remains 1, and the value must remain positive and less than the number of retained revisions. Persistence behavior and the database format are unchanged.

## How this was tested

CI.

## Breaking Changes

- [x] firewood
- [ ] firewood-storage
- [x] firewood-ffi (C api)
- [x] firewood-go (Go api)
- [ ] fwdctl

Rust and C callers must replace `deferred_persistence_commit_count` with `max_persistence_gap`. Go callers must replace `WithDeferredPersistenceCommitCount` with `WithMaxPersistenceGap`.

Rust error matches must replace `ZeroCommitCount` with `ZeroMaxPersistenceGap` and the `InsufficientRevisions::commit_count` field with `max_persistence_gap`.

Existing configuration values remain valid. The C field retains its type and position, and database files require no migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ To do this, a new root is always created for each revision that can reference ei
 When creating a revision,
 a list of nodes that are no longer needed are computed and saved to disk in a future-delete log (FDL) as well as kept in memory.
 When a revision expires, the nodes that were deleted when it was created are returned to the free space.
-Firewood also supports archival mode via `RootStore`, which retains all historical revisions and allows lookup of any past state by its root hash.
+Firewood also supports archival mode via `RootStore`, which retains persisted historical revisions for lookup by root hash.
 
 Hashes are not used to determine where a node is stored on disk in the database file.
 Instead space for nodes may be allocated from the end of the file,
@@ -43,6 +43,11 @@ as well as carefully managing the free list during the creation and expiration o
 
 ## Terminology
 
+- `Commit` - Makes a proposal's state the latest committed state. Success does
+  not guarantee that the state has been persisted.
+- `Persist` - Writes committed state to the database files.
+- `Maximum persistence gap` - The maximum number of committed revisions to go
+  back from the latest committed state to reach the last persisted state.
 - `Revision` - A historical point-in-time state/version of the trie. This
   represents the entire trie, including all `Key`/`Value`s at that point
   in time, and all `Node`s.
@@ -81,8 +86,6 @@ as well as carefully managing the free list during the creation and expiration o
     uncommitted branches and do not participate in proposal-parent branching.
 - `Reconstructible` - Either a `Historical` or `Reconstructed` state, which supports
   building new `Reconstructed` states by applying a `Batch`.
-- `Commit` - The operation of applying one or more `Proposal`s to the most recent
-  `Revision`.
 
 ## Metrics
 
--- a/ffi/firewood.go
+++ b/ffi/firewood.go
@@ -124,7 +124,7 @@ type config struct {
 	// revisions is the maximum number of historical revisions to keep in memory.
 	// If rootStoreDir is set, then any revisions removed from memory will still be kept on disk.
 	// Otherwise, any revisions removed from memory will no longer be kept on disk.
-	// Must be >= 2 and > deferredPersistenceCommitCount.
+	// Must be >= 2 and > maxRevisionRecoveryLag.
 	revisions uint
 	// readCacheStrategy is the caching strategy used for the node cache.
 	readCacheStrategy CacheStrategy
@@ -132,19 +132,19 @@ type config struct {
 	rootStore bool
 	// expensiveMetricsEnabled controls whether expensive metrics recording is enabled.
 	expensiveMetricsEnabled bool
-	// deferredPersistenceCommitCount determines the maximum number of unpersisted
-	// revisions that can exist at a given time.
-	// Note: revisions must be > deferredPersistenceCommitCount
-	deferredPersistenceCommitCount uint64
+	// maxRevisionRecoveryLag determines the maximum number of latest committed
+	// revisions that may need to be recommitted after a crash.
+	// Note: revisions must be > maxRevisionRecoveryLag
+	maxRevisionRecoveryLag uint64
 }
 
 func defaultConfig() *config {
 	return &config{
-		nodeCacheSizeInBytes:           128_000_000,
-		freeListCacheEntries:           1_000_000,
-		revisions:                      100,
-		readCacheStrategy:              OnlyCacheWrites,
-		deferredPersistenceCommitCount: 1,
+		nodeCacheSizeInBytes:   128_000_000,
+		freeListCacheEntries:   1_000_000,
+		revisions:              100,
+		readCacheStrategy:      OnlyCacheWrites,
+		maxRevisionRecoveryLag: 1,
 	}
 }
 
@@ -181,7 +181,7 @@ func WithFreeListCacheEntries(entries uint) Option {
 // WithRevisions sets the maximum number of historical revisions to keep in memory.
 // If RootStoreDir is set, then any revisions removed from memory will still be kept on disk.
 // Otherwise, any revisions removed from memory will no longer be kept on disk.
-// Must be >= 2 and > WithDeferredPersistenceCommitCount.
+// Must be >= 2 and > WithMaxRevisionRecoveryLag.
 // Default: 100
 func WithRevisions(revisions uint) Option {
 	return func(c *config) {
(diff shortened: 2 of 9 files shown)


ffi/firewood.go
```diff
@@ -242,13 +240,34 @@ func WithMetricsTag(tag string) Option {
...
-func WithDeferredPersistenceCommitCount(commitCount uint64) Option {
+// WithMaxPersistenceGap sets the maximum number of committed revisions by which
+// the latest persisted state may lag behind the latest committed state.
+//
+// Committing makes a revision current, while persistence writes its state
+// to the database files asynchronously, without a fixed schedule. A commit
+// can therefore return before its state is persisted, even when this value
+// is 1, and subsequent commits may wait for persistence to maintain the
+// configured bound. Only commits that change the current root count toward
+// this limit; uncommitted proposals do not.
```
alarso16, line 251: I personally would delete this entire paragraph, and the examples. It's a little verbose, since if someone doesn't know what a revision is, I don't kno wif this will necessarily clarify anything. I would replace it specifically with disk-level guarantees.

RodrigoVillar: I went ahead and deleted the examples section. From talking offline, this paragraph makes clear the difference between committing and persisting revisions (we can work on making this distinction more clear in a follow-up PR). I've also added additional documentation regarding which revisions will be on-disk: [83ab18e](https://github.com/ava-labs/firewood/pull/2043/commits/83ab18e9251384d3941479ef0623c2e122d01e67).

ffi/firewood.go
```diff
@@ -242,13 +240,29 @@ func WithMetricsTag(tag string) Option {
...
+// Set this value to 1 to persist every committed revision before the next
+// state-changing commit completes. Values greater than 1 allow persistence
+// to skip intermediate revisions, so not every revision is guaranteed to be
+// written to disk.
+//
+// A successful [Database.Close] persists the latest committed state. If the
+// process exits without a successful Close, the recovered state is at most
+// maxPersistenceGap commits behind the latest committed revision.
+//
+// Defaults to 1. Must be positive and less than the value set by [WithRevisions].
```
alarso16, line 250:
```suggestion
// Defaults to 1. Must be non-zero and less than the value set by [WithRevisions].
```
nit, opinion, feel free to disagree

ffi/src/handle.rs
```diff
@@ -113,10 +113,28 @@ pub struct DatabaseHandleArgs<'a> {
...
+    /// state-changing commit completes. Values greater than 1 allow persistence
+    /// to skip intermediate revisions, so not every revision is guaranteed to be
+    /// written to disk.
+    ///
+    /// A successful explicit close persists the latest committed state. If the
+    /// process exits without a successful close, the recovered state is at
+    /// most `max_persistence_gap` commits behind the latest committed
+    /// revision.
+    ///
+    /// Must be positive and less than `revisions`.
```
rkuris, line 136: This is a design discussion that should go elsewhere. A reader that cares about understanding max_persistence_gap shouldn't care much about many of these points here. Here's what they need to know.

"max_persistence_gap" does the heavy lifting now.


```suggestion
    /// Block commits when this many revisions have not been persisted to disk.
    ///
    /// Setting this value to 1 persists every revision and will block if another
    /// commit is issued before the persistence completes.
    ///
    /// Must be positive and less than `revisions`.
```

ffi/firewood.go
rkuris: I'd much prefer to talk about the *effect* of this instead. See my comment above for some simplifications. I'm with Austin that this is too wordy.

ffi/firewood.go
```diff
@@ -242,13 +240,29 @@ func WithMetricsTag(tag string) Option {
...
+// this limit; uncommitted proposals do not.
+//
+// Set this value to 1 to persist every committed revision before the next
+// state-changing commit completes. Values greater than 1 allow persistence
+// to skip intermediate revisions, so not every revision is guaranteed to be
+// written to disk.
+//
+// A successful [Database.Close] persists the latest committed state. If the
+// process exits without a successful Close, the recovered state is at most
+// maxPersistenceGap commits behind the latest committed revision.
```
rkuris, line 260: This is tautological if we describe maxPersistenceGap clearly. See my earlier discussion of this. If the invariant is "we never are in a state where we are behind by max_persistence_gap or we block" then we don't need to specifically talk about the close case.

firewood/src/manager.rs
```diff
@@ -63,11 +63,29 @@ pub struct RevisionManagerConfig {
...
+    /// state-changing commit completes. Values greater than 1 allow persistence
+    /// to skip intermediate revisions, so not every revision is guaranteed to be
+    /// written to disk.
+    ///
+    /// A successful explicit database close persists the latest committed
+    /// state. If the process exits without a successful close, the recovered
+    /// state is at most `max_persistence_gap` commits behind the latest
+    /// committed revision.
+    ///
+    /// Defaults to 1. Must be positive and less than `max_revisions`.
```
rkuris, line 86: Interesting that there are three of these, all basically the same. Can we do something about that? Certainly this is a big block, maybe we can refer to a central location if someone wants more information.

README.md
```diff
@@ -43,6 +43,11 @@ as well as carefully managing the free list during the creation and expiration o
 
 ## Terminology
 
+- `Commit` - Makes a proposal's state the latest committed state. Success does
+  not guarantee that the state has been persisted.
+- `Persist` - Writes committed state to the database files.
+- `Maximum persistence gap` - The maximum number of committed revisions to go
+  back from the latest committed state to reach the last persisted state.
```
rkuris, line 50: This is misleading. Maybe:

```suggestion
- `Maximum persistence gap` - The maximum number of committed but unpersisted revisions to allow to remain in memory before blocking additional ones.
```

ffi/src/handle.rs
RodrigoVillar: > this many revisions have not been persisted to disk.

This implies that commits are unblocked when all prior revisions are persisted to disk, which is not true.

README.md
RodrigoVillar: https://github.com/ava-labs/firewood/pull/2043/changes#r3983622356

ffi/src/handle.rs
rkuris: I'm not following. If I say "the system is blocked when this many revisions have not been persisted" and the variable is named "max_something" then one could assume as long as the number of revisions persisted is less than that, the system is not blocked.

With reversing the logic, maybe "As long as the number of unpersisted revisions is less than this number, the system is unblocked" seems like a double negative and is harder to parse, but I'd be okay with that language too.

README.md
rkuris: Must be me today :)

Let's assume it is "the maximum number of revisions to go back from the latest committed state to reach the last persisted state"

Example scenario: maximum persistence gap is set to 1000. If we're at commit 1234, then the last persisted state is not 234, it's probably 500. If we're at commit 789 and the persist finished at 500, "going back" 1000 revisions isn't even possible. There's no relationship between this variable and how to "reach the last persisted state".

RodrigoVillar: Ah I think I see what the confusion is here - the qualifier for *last persisted state* should be "a" instead of "the". We'd have then:

> the maximum number of revisions to go back from the latest committed state to reach *a* persisted state

alarso16 (approved):
I'm good with this

ffi/firewood.go
RodrigoVillar: Removed this paragraph here: ea349ff971bf72436645fc8d723b0d0e7c8351fe.

README.md
RodrigoVillar: Resolved by [a8535f8](https://github.com/ava-labs/firewood/pull/2043/commits/a8535f8b8561036d8d19187d566015decab89b39).

firewood/src/manager.rs
RodrigoVillar: These docs are duplicated just twice now (once in Go and once in Rust): ea349ff971bf72436645fc8d723b0d0e7c8351fe.

ffi/src/handle.rs
RodrigoVillar: Resolved by [a8535f8](https://github.com/ava-labs/firewood/pull/2043/commits/a8535f8b8561036d8d19187d566015decab89b39).

ffi/firewood.go
RodrigoVillar: Resolved by [a8535f8](https://github.com/ava-labs/firewood/pull/2043/commits/a8535f8b8561036d8d19187d566015decab89b39).

ffi/firewood.go
RodrigoVillar: Going to keep using _positive_ rather that _non-zero_ as the former is a stronger qualifier (e.g. `-1` is non-zero but is not a valid persistence gap.

rkuris (approved):

PR #2044, opened 2026-06-02
chore: add TODO checker

### LINKED ISSUE #1603: All TODOs should have an associated issue
We have several TODOs in our code, but no attached GH issues with them. IMO, this is problematic for several reasons:

- No incentive structure: with GH issues, we have regular backlog grooming and we strive to keep the issue list as small as possible. With TODOs, there's no such structure.
- Lost context: without an attached GH issue, TODOs become lost and even more importantly, the context as to why the TODO exists become more vague over time (to play devil's advocate, `git blame` can partially mediate this).

RodrigoVillar:
## Why this should be merged

Closes #1603 by adding a TODO/FIXME checker to CI.

## How this works

This PR consists of two commits:

- [d200ead](https://github.com/ava-labs/firewood/pull/2044/commits/d200eadbbf9b7da3c837286c038cb82707e93c3a): adds the TODO/FIXME checker. Running `./scripts/check-todos.sh --help` provides a good summary of what this tool does:
```
Usage: scripts/check-todos.sh [-h|--help]

Scans tracked .rs and .go files for TODO/FIXME markers and the Rust todo!()
macro, and reports any that are not annotated with an owner or GitHub issue.

A marker is valid when it is immediately followed by a non-empty parenthesized
annotation:

  TODO(owner or GH issue)   e.g. TODO(#1603), TODO(@foo), FIXME(rust-lang/rust#143874)
  todo!("...")              e.g. todo!("#1603: implement this")

Exit status: 0 when no violations, 1 otherwise.
```

- [eaed37d](https://github.com/ava-labs/firewood/pull/2044/commits/eaed37de51398ecd43bdb644d4caf3415ffe7e3d): satisfies the linter by adding issues/owners to all TODOs/FIXMEs. The owner of an issue was determined via `git log` with the intent being to attach the owner who introduced the issue in-code. For owners not on the Firewood team, I deferred to creating issues for these issues.

In the future, we can add stricter rules to this check (e.g. all TODOs/FIXMEs must have a GH issue attached to them); however, this is a good start to making sure all TODOs/FIXMEs in the Firewood repository have sufficient context (either via an issue or an owner).

## How this was tested

Case where there are TODO violations: https://github.com/ava-labs/firewood/actions/runs/26830137698/job/79108776074?pr=2044

Case where there are no violations: https://github.com/ava-labs/firewood/actions/runs/26837932459/job/79136793249?pr=2044

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/scripts/check-todos.sh
+++ b/scripts/check-todos.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Enforces that every TODO/FIXME has an attached owner or issue.
+#
+# Motivation (https://github.com/ava-labs/firewood/issues/1603): bare TODOs lose
+# their context over time and have no backlog/grooming pressure behind them.
+# Requiring each one to name an owner or link an issue keeps that context alive.
+#
+# Rule: a TODO or FIXME marker in a .rs or .go file must be annotated with an
+# owner or GitHub issue inside parentheses immediately after the marker:
+#
+#   TODO(owner or GH issue) - e.g. TODO(#1603), TODO(@foo), FIXME(rust-lang/rust#143874)
+#
+# The Rust todo!() macro is held to the same rule: it must carry a non-empty
+# message naming an owner or issue, e.g. todo!("#1603: implement this"). A bare
+# todo!() is a violation.
+#
+# Any marker without a non-empty parenthesized annotation is a violation, e.g.:
+#   TODO:  /  TODO foo:  /  TODO[foo]:  /  //TODO  /  todo!()
+#
+# The contents of the parentheses are not validated beyond being non-empty; the
+# syntax can be tightened later. Comment markers are matched as whole, uppercase
+# words anywhere on a line; the todo!() macro is matched in lowercase.
+#
+# Exit status: 0 when clean, 1 when any violation is found.
+
+usage() {
+    cat <<'EOF'
+Usage: scripts/check-todos.sh [-h|--help]
+
+Scans tracked .rs and .go files for TODO/FIXME markers and the Rust todo!()
+macro, and reports any that are not annotated with an owner or GitHub issue.
+
+A marker is valid when it is immediately followed by a non-empty parenthesized
+annotation:
+
+  TODO(owner or GH issue)   e.g. TODO(#1603), TODO(@foo), FIXME(rust-lang/rust#143874)
+  todo!("...")              e.g. todo!("#1603: implement this")
+
+Exit status: 0 when no violations, 1 otherwise.
+EOF
+}
+
+case "${1:-}" in
+    -h | --help)
+        usage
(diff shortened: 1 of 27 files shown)


rkuris:
Brilliant idea, just happened to spot this. What about todo!("something")? Is this exempt or should it follow the same rules? Either way is fine, but we should be explicit. These are usually markers that mean we're doing some separated development; they panic when they are hit but are useful for breaking up WIP.

RodrigoVillar:
> Brilliant idea, just happened to spot this. What about todo!("something")? Is this exempt or should it follow the same rules? Either way is fine, but we should be explicit. These are usually markers that mean we're doing some separated development; they panic when they are hit but are useful for breaking up WIP.

@rkuris added an extra check in the script for `todo!()`. Similar to the existing rules, this passes if `todo!()` has a non-empty message, and fails otherwise.

scripts/check-todos.sh
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
```
demosdemon, line 1: one caveat is that we don't have an escape hatch for the literal `TODO` or `FIXME` in a string literal, comment, url, etc.

Currently doesn't affect us but may bite us in the future. We would need a mechanism to ignore an individual item or file. Not blocking for this PR but is a nice to have.

scripts/check-todos.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# The Rust todo!() macro, held to the same rule. The marker is a lowercase
+# todo! invocation; it is valid only when followed by a non-empty parenthesized
+# message (which should name an owner or issue). A bare todo!() is a violation.
+macro_marker_re='(^|[^A-Za-z0-9_])todo!'
+macro_valid_re='(^|[^A-Za-z0-9_])todo![[:space:]]*\([^)]*[^)[:space:]][^)]*\)'
+
+# git grep skips target/, untracked, and gitignored files for us. It exits 1
+# when there are no matches, which is not an error for us.
+candidates=$(git grep -nE 'TODO|FIXME|todo!' -- '*.rs' '*.go' || true)
```
demosdemon, line 79: Would like to see this expanded to include other files. We may put TODOs in shell scripts, toml, yaml, and markdown files as well (nonblocking).

demosdemon (approved):

scripts/check-todos.sh
RodrigoVillar: Created #2062 to handle in a follow-up.

scripts/check-todos.sh
RodrigoVillar: Creating #2061 to handle this in a follow-up.

PR #2053, opened 2026-06-02
feat(proofs): 5/7 add eth_get_proof entry point producing eth_getProof-compatible proofs

### LINKED ISSUE #2001: feat: Implement rust-side eth_getProof

rkuris:
## Why this should be merged

The high-level entry point that turns firewood's native proofs into an
`eth_getProof`-shaped result: the four account scalars plus RLP-encoded MPT
nodes for the account and each requested storage slot.

## How this works

- New `firewood::eth_proof` module with `eth_get_proof`, `EthProof`, and
  `EthStorageProof`. Failures surface as `api::Error`; there is no
  module-specific error type.
- Gates at runtime on `NodeHashAlgorithm::compile_option().is_ethereum()`,
  returning `Error::FeatureNotSupported` under merkledb mode.
- Absent accounts (including the empty-trie `ProofError::Empty` case) surface
  as zero fields plus the empty-code and empty-trie-root hashes, with the
  exclusion-proof bytes when one exists — matching geth's `eth_getProof` shape
  (no `AccountNotFound` error). `EthProof::default` carries the absent-account
  constants (`KECCAK_EMPTY` for `code_hash`; `HashKey::default_root_hash` for
  `storage_hash`) so the function builds its value in a single expression.
- Multi-storage-child accounts: `storage_proofs[*].proof[0]` is the account
  node's 17-element branch RLP (via the new
  `proofs::eth::account_storage_root_rlp` helper), whose keccak matches the
  storageRoot spliced into the account leaf.
- Single-storage-child accounts: `storage_proofs[*].proof` is the synthesized
  leaf bytes via `synth_storage_leaf_rlp`; an extra `single_key_proof` descent
  is issued when the requested slot's first storage nibble differs from the
  stored child's.
- Proof byte arrays use `Box<[Box<[u8]>]>` (built once via `collect`, never
  mutated) rather than `Vec`, saving the capacity field and signalling intent
  to FFI/callers.

## How this was tested

7 unit tests covering: mode-gate present, empty-trie absent account, exclusion
proof on a non-empty trie, present account with no storage, single-slot
inclusion + exclusion, multi-child branching, and an end-to-end MPT verifier
walking from `root_hash` to the account leaf. Full workspace nextest + clippy +
doc with `--features ethhash,logger`.

## Breaking Changes

None. Purely additive to firewood (new `eth_proof` module and re-exports).


firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! High-level `eth_getProof`-compatible proof export.
+//!
+//! Wraps [`crate::api::DbView::single_key_proof`] and converts the resulting
+//! firewood `ProofNode`s into the canonical Ethereum MPT-RLP byte form an
```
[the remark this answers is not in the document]
rkuris: Meh. Both seem fine to me. Someone yell if they disagree.

firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
...
+//! Absent accounts return zero account fields plus the exclusion proof
+//! bytes; callers translate that into the JSON-RPC default-fields shape
+//! themselves. This matches go-ethereum's behavior and keeps a single
+//! error type on the public API.
+
+use firewood_storage::{
+    NodeHashAlgorithm, PackedPathRef, PathComponent, TriePathFromPackedBytes, ValueDigest,
+};
+
+use crate::api::{DbView, Error, HashKey, HashKeyExt};
```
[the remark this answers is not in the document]
rkuris: This is incorrect. We need HashKeyExt or it won't compile the extension trait it uses.

firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
...
+            RlpItem::Bytes(nonce_min),
+            RlpItem::Bytes(balance_be),
+            RlpItem::Bytes(&[0u8; 32]), // storageRoot — fixed up by firewood
+            RlpItem::Bytes(&KECCAK_EMPTY),
+        ])
+    }
+
+    /// Confirm a non-ethhash build refuses to emit eth proofs. (We're under
+    /// `cfg(feature = "ethhash")`, so the gate's *negative* path is exercised
+    /// only by the workspace's non-ethhash test profile.)
```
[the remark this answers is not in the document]
rkuris: I also think this is okay as written.

firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
...
+///
+/// # Errors
+///
+/// - [`Error::FeatureNotSupported`] if the database is not running in
+///   ethereum hash mode.
+/// - Any [`Error`] surfaced by [`DbView::single_key_proof`] while fetching
+///   the underlying firewood proofs.
+/// - [`Error::InternalError`] wrapping an account-RLP decode error if the
+///   on-disk account value cannot be parsed.
+pub fn eth_get_proof<V>(
```
RodrigoVillar, line 117: > *Note*: this comment is AI-generated

`eth_get_proof` currently treats `account_node.child_hashes.count() == 1`
  as the single-storage-slot case. That assumption is incorrect.

  If two storage slots share the first storage nibble, Firewood path-compresses
  them under one account child. That account child is itself a storage subtree,
  not a single storage leaf.

  The implementation then enters the synthesized-leaf path, looks only at the
  first storage node below the account, and fails to reach the actual slot leaf.
  As a result, present slots can be returned as absent (`value: None`) and the
  emitted proof is not the storage proof rooted at the account’s `storageHash`.

Below is a regression test I wrote to test this case: 

  ```rust
  #[test]
  fn present_account_two_slots_shared_first_nibble() {
      // Step 1: Create one account with two storage slots whose first storage
      // nibble is the same. This makes the account node have exactly one child.
      let account_key = [0x55u8; 32];
      let acc_val = account_rlp(3, &[]);

      let slot_a = [0x00u8; 32];
      let mut slot_b = [0x00u8; 32];
      slot_b[0] = 0x01;

      assert_eq!(
          slot_a[0] >> 4,
          slot_b[0] >> 4,
          "slots must share their first storage nibble"
      );

      const VAL_A: &[u8] = b"val-a";
      const VAL_B: &[u8] = b"val-b";

      // Step 2: Build Firewood's full storage keys:
      // account_key || slot_key.
      let mut full_a = account_key.to_vec();
      full_a.extend_from_slice(&slot_a);
      let mut full_b = account_key.to_vec();
      full_b.extend_from_slice(&slot_b);

      // Step 3: Insert the account and both storage slots.
      let merkle = init_merkle([
          (account_key.as_slice(), acc_val.as_ref()),
          (full_a.as_slice(), VAL_A),
          (full_b.as_slice(), VAL_B),
      ]);

      // Step 4: Request eth_getProof-style storage proofs for both slots.
      let proof = eth_get_proof(merkle.nodestore(), &account_key, &[slot_a, slot_b]).unwrap();
      assert_eq!(proof.storage_proofs.len(), 2);

      // Step 5: Both slots are present, so both storage proofs must be
      // inclusion proofs with the stored values.
      assert_eq!(
          proof.storage_proofs[0].value.as_deref(),
          Some(VAL_A),
          "slot_a is present and must yield an inclusion proof"
      );
      assert_eq!(
          proof.storage_proofs[1].value.as_deref(),
          Some(VAL_B),
          "slot_b is present and must yield an inclusion proof"
      );

      // Step 6: Each storage proof must be rooted at the account's storageHash.
      for entry in &proof.storage_proofs {
          assert!(
              !entry.proof.is_empty(),
              "present slot must carry a non-empty storage proof"
          );
          assert_eq!(
              keccak(&entry.proof[0]),
              proof.storage_hash,
              "first storage proof node must hash to storageRoot"
          );
      }
  }
```

This test fails with the following error message:

```
       FAIL [   0.019s] firewood eth_proof::tests::present_account_two_slots_shared_first_nibble
  stdout ───

    running 1 test
    test eth_proof::tests::present_account_two_slots_shared_first_nibble ... FAILED

    failures:

    failures:
        eth_proof::tests::present_account_two_slots_shared_first_nibble

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 420 filtered out; finished in 0.00s

  stderr ───

    thread 'eth_proof::tests::present_account_two_slots_shared_first_nibble' (789544) panicked at firewood/src/eth_proof.rs:669:11:
    assertion `left == right` failed: slot_a is present and must yield an inclusion proof
      left: None
     right: Some([118, 97, 108, 45, 97])
```

File: firewood/src/eth_proof.rs [hunk 5]
```diff
@@ -0,0 +1,628 @@
...
+
+/// Nibble depth at which a node represents a storage slot leaf
+/// (account 64 nibbles + slot 64 nibbles).
+const STORAGE_LEAF_DEPTH_NIBBLES: usize = ACCOUNT_DEPTH_NIBBLES + 64;
+
+/// `keccak256("")` — the `codeHash` value for accounts with no contract code.
+const KECCAK_EMPTY: [u8; 32] = [
+    0xc5, 0xd2, 0x46, 0x01, 0x86, 0xf7, 0x23, 0x3c, 0x92, 0x7e, 0x7d, 0xb2, 0xdc, 0xc7, 0x03, 0xc0,
+    0xe5, 0x00, 0xb6, 0x53, 0xca, 0x82, 0x27, 0x3b, 0x7b, 0xfa, 0xd8, 0x04, 0x5d, 0x85, 0xa4, 0x70,
+];
```
RodrigoVillar, line 36: Maybe we should add a test to verify that `keccak256("") == KECCAK_EMPTY`?

firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
...
+        assert_eq!(proof.storage_hash, keccak(NULL_RLP));
+        assert!(proof.storage_proofs.is_empty());
+        // Account proof bytes for an empty trie is just the root's exclusion
+        // proof. The non-empty case is exercised below.
+    }
+
+    #[test]
+    fn present_account_no_storage_decodes_fields() {
+        let key = [0x22u8; 32];
+        let value = account_rlp(7, &[0x12, 0x34]);
```
RodrigoVillar:
```suggestion
        let balance = [0x12, 0x34]
        let value = account_rlp(7, &balance);
```

Allows for better legibility in the `assert_eq!(proof.balance[30..], [0x12, 0x34])` check later down the test.

RodrigoVillar (commented):
Few comments - the comment about `eth_get_proof` was flagged by multiple AI models.

firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
...
+///
+/// Trie keys here are the already-keccak-hashed 32-byte forms (firewood
+/// stores accounts at `keccak256(address)` and slots at
+/// `keccak256(address) ++ keccak256(slot_key)` — callers do the hashing).
+///
+/// # Returns
+///
+/// An [`EthProof`] with one [`EthStorageProof`] per `storage_keys` entry,
+/// in input order. Absent accounts come back with zero account scalars
+/// plus the empty-code and empty-trie hashes; the proof bytes themselves
```
demosdemon, line 106: This branch keys the proof shape off `account_node.child_hashes.count()` (line 61), but `count()` is the number of distinct first storage nibbles, not the number of slots. Two slots that share their first storage nibble path-compress under a single account child, so `count() == 1` even though that lone child is an interior branch at depth ~65, not a leaf.

When that happens, this single-slot path is wrong in two ways:

- the inclusion filter at lines 96-98 requires `STORAGE_LEAF_DEPTH_NIBBLES` (128), so the interior child is dropped and a present slot comes back as `value: None` (this is the `left: None` Rodrigo saw in the line-118 thread);
- `synth_storage_leaf_rlp` (lines 90-91) is handed an interior node with no `value_digest`, so it emits a proof that isn't rooted at `storageHash`.

The decision should key off the storage-node depth rather than the account child count: a lone storage node at depth 128 takes the synth-leaf path, otherwise fall through to the multi-child descend-and-collect. A `0xa0..`/`0xa1..`-style regression test (two slots sharing the first storage nibble) would fail today and should land with the fix. Note the end-to-end go-ethereum verification that would catch this lives in #2055, and its slot keys don't collide either.

firewood/src/eth_proof.rs
```diff
@@ -0,0 +1,628 @@
...
+    // Inclusion: terminal node sits at account depth and carries a value.
+    // Anything else is exclusion (terminal is a strict ancestor, or
+    // a depth-64 branch with no account value).
+    let account_terminal = account_nodes
+        .last()
+        .filter(|n| n.key.as_ref().len() == ACCOUNT_DEPTH_NIBBLES && n.value_digest.is_some());
+
+    let fields = account_terminal
+        .and_then(|n| n.value_digest.as_ref().and_then(ValueDigest::value))
+        .map(AccountFields::from_rlp)
```
demosdemon, line 156: When `AccountFields::from_rlp` fails on a malformed account RLP, the error ends up as `api::Error::InternalError`, whose `Display` is `#[error("Internal error")]` with no `{0}` — so the decode detail is lost except via `source()`. Worth mapping decode failures to `Error::IO(InvalidData, ...)` (`#[error("IO error: {0}")]`) instead, inside `from_rlp`. This is the same dropped-message pattern Copilot flagged on `to_trie_key` in #2054; nice to fix both consistently.

firewood/src/eth_proof.rs
demosdemon: This comment says the gate's negative path is exercised by the non-ethhash profile, but the whole `tests` module is behind `cfg(feature = "ethhash")`, so a non-ethhash build runs zero tests from this file — the negative `FeatureNotSupported` path isn't actually covered anywhere. Either add a `cfg(not(feature = "ethhash"))` test asserting `FeatureNotSupported` and reword this, or just reword it to match reality. This same gap recurs at the FFI and Go layers (#2054/#2055).

firewood/src/eth_proof.rs
demosdemon: agreed

demosdemon (changes requested):

firewood/src/eth_proof.rs
rkuris: Fixed in 0722bc9. The single-child-nibble case now re-emits the account's lone child as the storage root — folding back the account-branch nibble that firewood consumes at the account→child edge — instead of synthesizing a leaf, so a subtree shared under one first nibble is no longer dropped. The dispatch keys off the child-nibble count, not the slot count.

Added `present_account_two_slots_shared_first_nibble` (inclusion + both exclusion paths), and switched the storage-proof tests to verify each entry end-to-end against `storageHash` via the MPT walker — the old `keccak(proof[0])` spot-check passed the broken proof because the root hash matched while the interior was dropped.

firewood/src/eth_proof.rs
rkuris: Added in 7fabb01f — `keccak_empty_matches_keccak_of_empty_input` asserts the hardcoded `KECCAK_EMPTY` equals `keccak256("")`.

firewood/src/eth_proof.rs
rkuris: Done in 7fabb01f — bound the balance bytes to a named local so the later `assert_eq!(proof.balance[30..], balance)` checks against the same value.

firewood/src/eth_proof.rs
rkuris: Fixed in 0722bc9c. Great find — though the fix keys off the same `count() == 1` branch rather than storage-node depth: the synth-leaf path is gone entirely, and the single-child arm now re-emits the account's lone child (whether leaf or interior subtree root) with the consumed account-edge nibble folded back, then chains the rest of the descent. `present_account_two_slots_shared_first_nibble` is the shared-first-nibble regression test; it asserts both failure modes you describe are gone — present slots verify end-to-end with their values, and the proof roots at `storageHash`.

firewood/src/eth_proof.rs
rkuris: Fixed in 29c56d69. Decode failures now map to `Error::IO(InvalidData, ...)` inside the `From` impl, and the unit test asserts the detail actually survives `Display`. The `to_trie_key` counterpart lives in #2054's diff and will be fixed there.

firewood/src/eth_proof.rs
rkuris: Fixed in 58137122. Added `merkledb_gate_tests::eth_get_proof_rejected_without_ethhash` — compiles only without `ethhash` and asserts `FeatureNotSupported` — and reworded the comment to point at it instead of overclaiming.

demosdemon (approved):

PR #2055, opened 2026-06-02
feat(ffi): 7/7 add Go EthGetProof wrapper and go-ethereum verification test

### LINKED ISSUE #2001: feat: Implement rust-side eth_getProof

rkuris:
## Why this should be merged

Completes the `eth_getProof` export work: gives Go callers an idiomatic entry
point over the step-6 FFI surface and proves the emitted proofs verify under
go-ethereum, so Avalanche clients can answer Ethereum's `eth_getProof`
JSON-RPC against a firewood revision.

## How this works

- `ffi/eth_proof.go`: `Revision.EthGetProof(accountKey, slotKeys)` returns
  `*EthAccountProof` (plus one `EthStorageProof` per slot). It builds a
  `BorrowedBytes2D` for the slot keys, decodes the `EthProofResult` tagged
  union, deep-copies every buffer out of Rust memory, and frees the
  `EthProofOwned` with a single `fwd_free_eth_proof` call so the result
  outlives the revision and database. `NotSupported` maps to
  `ErrEthProofNotSupported`. Keys must be 32-byte trie keys; the caller
  keccak-hashes addresses/slots, and JSON marshalling is left to the caller.

## How this was tested

- `ffi/tests/eth/get_proof_test.go` materializes state into both a libevm trie
  (for canonical roots / account RLPs) and firewood (flat keccak-prefixed
  keys), asserts the roots agree, then verifies the returned account and
  storage proofs with go-ethereum's `trie.VerifyProof` — covering inclusion,
  exclusion, and the storage-root bridge (StorageHash vs the canonical storage
  trie root).
- `go test ./...` (ffi and tests/eth); golangci-lint clean on the ffi module.

## Breaking Changes

None. Purely additive to firewood-go (new `Revision.EthGetProof` and
`EthAccountProof`/`EthStorageProof` types); no existing API changes.


ffi/eth_proof.go
```diff
@@ -0,0 +1,183 @@
...
+	cSlices := make([]C.BorrowedBytes, len(slices))
+	for i, s := range slices {
+		cSlices[i] = newBorrowedBytes(s, pinner)
+	}
+
+	ptr := unsafe.SliceData(cSlices)
+	pinner.Pin(ptr)
+
+	return C.BorrowedBytes2D{
+		ptr: ptr,
```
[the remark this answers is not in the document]
rkuris: This refers to the old rule from go-1.21 (go memory passed to C must not contain go pointers). That rule has been relaxed, so now memory passed to C may contain Go pointers as long as they are pinned, which is exactly what `newBorrowedBytes2D` does.

References:
  Sources:
  - [cmd/cgo](https://pkg.go.dev/cmd/cgo#hdr-Passing_pointers) — Passing pointers
  - [runtime.Pinner / Pin](https://pkg.go.dev/runtime#Pinner)

ffi/eth_proof.go
```diff
@@ -0,0 +1,183 @@
...
+	"errors"
+	"fmt"
+	"runtime"
+	"unsafe"
+)
+
+// ErrEthProofNotSupported is returned by [Revision.EthGetProof] when the
+// database is not running in Ethereum hash mode and therefore cannot produce
+// an eth_getProof-compatible proof.
+var ErrEthProofNotSupported = errors.New("eth_getProof requires ethereum hash mode")
```
alarso16, line 24: I think you should only export this if you're going to check for the error in `ffi/tests/firewood`

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,229 @@
...
+	root firewood.Hash
+	// accountRLP maps account hash -> the RLP we stored for that account.
+	accountRLP map[common.Hash][]byte
+}
+
+// buildState materializes specs into both a libevm state trie (to compute the
+// canonical Ethereum storage roots and account RLPs) and a firewood database
+// (flat keccak-prefixed keys). It asserts the two roots agree, so the firewood
+// proofs can be checked against go-ethereum's verifier.
+func buildState(t *testing.T, specs []accountSpec) builtState {
```
alarso16, line 51: This might be a good candidate for fuzzing. If you're not familiar with go fuzzing, I'm happy to help

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,229 @@
...
+	t.Helper()
+	r := require.New(t)
```
alarso16, line 53: You don't need to change this, since it's been a norm in Firewood, but this is an anti-pattern

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,229 @@
...
+		accountRLP[accHash] = encodedAcc
+		batch = append(batch, firewood.Put(accHash[:], encodedAcc))
+	}
+
+	ethRoot, set, err := accountTrie.Commit(true)
+	r.NoError(err)
+	if set != nil {
+		r.NoError(merged.Merge(set))
+	}
+	r.NoError(tdb.TrieDB().Update(ethRoot, types.EmptyRootHash, 0, merged, nil))
```
alarso16: This will fail if `ethRoot == types.EmptyRootHash`

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/ava-labs/libevm/crypto"
+	"github.com/ava-labs/libevm/ethdb/memorydb"
+	"github.com/ava-labs/libevm/rlp"
+	"github.com/ava-labs/libevm/trie"
+	"github.com/ava-labs/libevm/trie/trienode"
+	"github.com/ava-labs/libevm/triedb"
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/require"
+
+	firewood "github.com/ava-labs/firewood-go-ethhash/ffi"
```
alarso16:
```suggestion
"github.com/ava-labs/firewood-go-ethhash/ffi"
```
Common practice is to only alias if necessary

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,229 @@
...
+func buildState(t *testing.T, specs []accountSpec) builtState {
+	t.Helper()
+	r := require.New(t)
+
+	db := newFirewoodDB(t)
+	tdb := state.NewDatabaseWithConfig(rawdb.NewMemoryDatabase(), triedb.HashDefaults)
+	accountTrie, err := tdb.OpenTrie(types.EmptyRootHash)
+	r.NoError(err)
+
+	merged := trienode.NewMergedNodeSet()
```
alarso16, line 60: The easier thing than managing this all yourself would be to create a `statedb.New` and update the statedb, and call `Commit` to push to the triedb

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+func TestEthGetProofStorageInclusionAndExclusion(t *testing.T) {
+	r := require.New(t)
+
+	addr := common.HexToAddress("0x00000000000000000000000000000000000000c3")
+	slots := []slot{
+		{key: common.HexToHash("0x01"), val: common.HexToHash("0x1111")},
+		{key: common.HexToHash("0x02"), val: common.HexToHash("0x2222")},
+		{key: common.HexToHash("0x03"), val: common.HexToHash("0x3333")},
+	}
```
demosdemon, line 185: These three fixed slot keys are stored at `keccak256(slotKey)` (line 195), and the keccak digests of `0x01`/`0x02`/`0x03` don't share a first nibble, so the storage trie root is always a multi-child branch. That means this test only validates the multi-child path of #2053's `build_storage_proof` — the single-account-child-hiding-a-subtree path (two or more slots whose storage trie keys share a first nibble) is never exercised, and that's the path that returns a wrong/absent value with a proof not rooted at `storageHash`.

Because the real go-ethereum `trie.VerifyProof` runs here, this is the natural place to catch the bug. Could you add a storage case whose slot trie keys collide in the first nibble — easiest to construct at the firewood key level, `Put(accHash ++ slotHash)` for two `slotHash` sharing nibble 0 — and assert both value correctness and `trie.VerifyProof` against `StorageHash`? That test fails against the current #2053 code and is what makes the stack's headline claim actually true.

ffi/eth_proof.go
demosdemon: Worth keeping exported (it's useful to consumers), but it's never asserted in a test. A non-ethhash-mode test that calls `EthGetProof` and checks `errors.Is(err, ErrEthProofNotSupported)` would cover it — and it fills the negative-gate gap that's untested at the Rust and FFI layers too (#2053/#2054).

ffi/tests/eth/get_proof_test.go
demosdemon: agreed

ffi/tests/eth/get_proof_test.go
demosdemon: `TrieDB().Update` fails when `ethRoot == types.EmptyRootHash`, so this helper can't build an empty-state spec. Nothing triggers it today, but it's a cheap guard for future callers:

```suggestion
	if ethRoot != types.EmptyRootHash {
		r.NoError(tdb.TrieDB().Update(ethRoot, types.EmptyRootHash, 0, merged, nil))
	}
```

demosdemon (approved):

alarso16 (changes requested):
I think it's worth blocking merge on simplifying the data generation by using a `state.StateDB` if possible

ffi/tests/eth/get_proof_test.go
rkuris: Fixed in 7da00e9. Great find — added `TestEthGetProofStorageSharedFirstNibble`, which builds two slots whose keccak trie keys collide in the first nibble and runs go-ethereum's `trie.VerifyProof` against `StorageHash`. #2058 (now on main) fixed the underlying bug, so this locks it in as regression coverage.

ffi/tests/eth/get_proof_test.go
rkuris: Fixed in 7da00e9. Good call — guarded the `TrieDB().Update` so the helper skips it when `ethRoot == types.EmptyRootHash`, keeping it usable for future empty-state specs.

ffi/tests/eth/get_proof_test.go
rkuris: Fixed in 7da00e9. Dropped the alias and use the bare `ffi` package name. I left `eth_compatibility_test.go` alone since it came in via an unrelated merged PR (#1705) and Go allows per-file import specs — happy to clean it up separately if you'd like.

ffi/eth_proof.go
rkuris: I dug into this and a Go-layer test isn't feasible: the mode gate is compile-time (`cfg!(feature = "ethhash")` via `NodeHashAlgorithm::compile_option()`), and the Go library is always built with ethhash, so the `NotSupported` path is unreachable from Go. The Rust negative path is now covered by `merkledb_gate_tests::eth_get_proof_rejected_without_ethhash` in `eth_proof.rs` (added in #2058). Keeping the symbol exported since it maps a real FFI tag and is useful to consumers.

ffi/tests/eth/get_proof_test.go
rkuris: Good idea — I'd like to take you up on that. Filing a follow-up to add Go fuzzing here rather than expanding this PR's scope.

ffi/tests/eth/get_proof_test.go
rkuris: Makes sense — leaving as-is to match the existing Firewood test norm, but noted for a future cleanup pass.

ffi/tests/eth/get_proof_test.go
rkuris: I'd prefer to keep the explicit trie construction here: it builds both the libevm storage trie (for the canonical root) and the flat firewood keys in one place, and asserts the two roots agree, which is the property these tests rely on. A `statedb.New`+`Commit` path would hide that cross-check. Open to revisiting if it becomes unwieldy.

ffi/tests/eth/get_proof_test.go
rkuris: Created #2094

ffi/eth_proof.go
```diff
@@ -0,0 +1,183 @@
...
+// The proof byte slices are canonical RLP-encoded Ethereum MPT nodes, suitable
+// for verification with go-ethereum's trie.VerifyProof against the revision
+// root (account proof) or against StorageHash (storage proofs). JSON
+// marshalling that matches the eth_getProof JSON-RPC response is left to the
+// caller.
+type EthAccountProof struct {
+	// Nonce is the account transaction count.
+	Nonce uint64
+	// Balance is the account balance, zero-padded big-endian.
+	Balance [32]byte
```
alarso16, line 39: We have a constant for this: `Hash` and `RootLength`. You should probably use it everywhere. There's an argument for not doing it for `Balance` though

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,310 @@
...
+// findSlotsSharingFirstNibble returns n storage slots whose trie keys
+// (keccak256 of the slot key) all share the same first nibble, with distinct
+// non-zero values. Firewood path-compresses such slots beneath a single
+// account child, so the storage trie root is a real subtree (extension or
+// branch) rather than one synthesized leaf — the multi-slot-under-one-child
+// path that a trie with distinct first nibbles never reaches.
+func findSlotsSharingFirstNibble(t *testing.T, n int) []slot {
+	t.Helper()
+	buckets := make(map[byte][]slot)
+	for i := uint64(1); i < 1_000_000; i++ {
```
alarso16, line 246: nit: this constant is arbitrary, but It is probably better to have a bound than endlessly search. Consider using a `const`

ffi/tests/eth/get_proof_test.go
```diff
@@ -0,0 +1,310 @@
...
+		var key, val common.Hash
+		binary.BigEndian.PutUint64(key[24:], i)
+		binary.BigEndian.PutUint64(val[24:], i)
+		nibble := crypto.Keccak256(key[:])[0] >> 4
+		buckets[nibble] = append(buckets[nibble], slot{key: key, val: val})
```
alarso16, line 251: This line is confusing to me. Don't key and val match?

ffi/tests/eth/get_proof_test.go
alarso16: I think that's reasonable, there's a lot of weird invariants that maybe you don't want to worry about.

alarso16 (approved):
Approving to prevent my blocking, but you should still address the types

PR #2057, opened 2026-06-03
fix(proofs): change-proof verification rejects honest partial-storage…

dvansari65:
… proofs (ethhash)

## Why this should be merged
In ethhash mode, bounded change-proof verification rejected honest proofs through an account at depth 64 whenever the proof's boundary cut through that account's storage trie, leaving only a subset of its storage children in range.

## How this works
Relaxes the branch proof node value comparison at depth 64 (account level) to ignore the storageRoot field. This is sound because storageRoot is recomputed from the storage children at hash time, and the final root-hash comparison remains the primary integrity check.

## How this was tested
Added test cases in empty.rs and reconcile.rs verifying that change proof verification succeeds when storage children are partially in range.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/merkle/reconcile.rs
+++ b/firewood/src/merkle/reconcile.rs
@@ -3,6 +3,9 @@
 
 use firewood_storage::{Mutable, NodeStore, Propose, ReadableStorage, ValueDigest};
 
+#[cfg(feature = "ethhash")]
+use firewood_storage::RlpList;
+
 use crate::{ProofError, ProofNode, Value, merkle::Merkle};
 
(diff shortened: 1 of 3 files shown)


bernard-avalabs:
Thanks for this, @dvansari65. A similar fix landed in #2060 (now merged), taking the same approach of ignoring `storageRoot` at depth 64, so I'll be closing this PR.

dvansari65:
Ok @bernard-avalabs

PR #2058, opened 2026-06-03
fix(proofs): hash single storage child as storage-trie root

### LINKED ISSUE #2035: fix(proofs): proof verification diverges from live hashing (ethhash)
**Describe the bug**

In ethhash mode, range/change proof verification fails for honest proofs through any account at depth 64 with exactly one storage child. Surfaces in production as state-sync failure on v0.5.0.

**To Reproduce**
  
 ```rust
  #[test]
  fn test_full_range_proof_single_storage_child_account_no_truncation() {
      let dummy_storage_root = [0u8; 32];
      let account_key: Box<[u8]> = [0x10u8; 32].into();
      let storage_key: Box<[u8]> = [account_key.as_ref(), &[0xAAu8; 32]].concat().into();
      let account_value = rlp_encode_account(1, 100, &dummy_storage_root, &empty_code_hash());
      let storage_value: Box<[u8]> = rlp_encode_storage(&[0x42u8; 32]).into();

      let merkle = init_merkle([
          (account_key.as_ref(), account_value.as_ref()),
          (storage_key.as_ref(), storage_value.as_ref()),
      ]);
      let root_hash = merkle.nodestore().root_hash().unwrap();
      let range_proof = merkle.range_proof(None, None, None).unwrap();
      verify_range_proof::<_>(None::<&[u8]>, None::<&[u8]>, &root_hash, &range_proof).unwrap();
  }
```

 Bug fires whenever (1) an account has exactly one storage child live, and (2) that child is in-range during  verification. Truncation is not required.

**Expected behavior**
  
  Verification succeeds for honest proofs that match the on-disk root.

**Possible Solutions**
  
PR #2030 demonstrates a working transformation and is state-sync-validated end-to-end. Worth using as the diagnostic baseline; the production fix likely wants a cleaner restructure.

### PR DESCRIPTION
**Pull request by bernard-avalabs**
Closes #2035. Replaces #2030 (fold-only fix). Split out of #2039.

  ## Why this should be merged
  
  Under `ethhash`, range- and change-proof verification rejected valid proofs
  with `UnexpectedHash` whenever a depth-64 account had exactly **one in-range
  storage child**.
  
  Live hashing writes that lone child as a *standalone storage-trie root*: it
  folds the account branch's nibble onto the front of the child's partial path
  and hashes it as a root (the `fake_root_extra_nibble` path). Proof
  verification recomputes that child's hash from scratch but did **not**
  replicate the fold, so the rebuilt hash diverged from the on-disk hash and the
  proof was wrongly rejected.

  ## How this works

  - The "hash a lone storage child as a standalone storage-trie root" logic now
    lives in **one shared helper**, instead of three near-identical copies (live
    hashing, the checker, and proof verification).
  - Proof verification — for both range and change proofs — now uses that helper
    when an account has a single in-range storage child, so the hash it rebuilds
    matches what live hashing originally wrote.
  - A malformed proof (a storage-child node claiming an empty parent prefix) is
    rejected cleanly instead of panicking.

  ## How this was tested

  - New range-proof fold matrix in `tests/ethhash.rs` (A1–A8): single- vs
    multi-storage-child accounts across unbounded / right-limit / left-start /
    both-bound shapes — including the account reconstructed from a start
    (exclusion) proof, the multi-child "union recompute, do **not** fold" case,
    and the zero-in-range "lone out-of-range child is copied, not folded" guard.
  - Regression test that a malformed proof with `partial_len == 0` at the
    storage-child position is rejected cleanly (no panic).
  - Passes CI

  ## Breaking Changes

  - None.

### CODE REVIEW

File: firewood/src/proofs/types.rs [hunk 1]
```diff
@@ -532,10 +586,21 @@ impl<T: ProofCollection + ?Sized> Proof<T> {
...
+                let children = node.children();
+                expected_hash = children[key_nibble]
                     .as_ref()
                     .ok_or(ProofError::NodeNotInTrie)?
                     .clone();
+
+                // If this node is a depth-64 account branch with exactly one
+                // storage child, the next node IS that storage child and must
+                // be hashed as a standalone storage-trie root on the next
+                // iteration.
```
rkuris: Highly repetitive ai-slop. This is what the function is described as doing at the top. Remove this or simplify it.

firewood/src/proofs/types.rs
```diff
@@ -814,6 +879,45 @@ mod tests {
         }
     }
 
+    /// Regression: a storage-child proof node with `partial_len == 0` (empty
```
rkuris: Reword this. This is a regression *test* not a "Regression" at best. Not sure why that's important anyway.

firewood/src/proofs/types.rs
```diff
@@ -814,6 +879,45 @@ mod tests {
...
+    /// parent prefix) at the single-storage-child fold position must be
+    /// rejected cleanly, not panic. `partial_len` is an independent wire field,
```
rkuris: panic?

firewood/src/proofs/types.rs
bernard-avalabs: Removed.

firewood/src/proofs/types.rs
bernard-avalabs: Removed regression. Shortened comment.

firewood/src/proofs/types.rs
bernard-avalabs: Removed.

rkuris (approved):

PR #2059, opened 2026-06-04
feat: support eth_getProof for reconstructed revisions

RodrigoVillar:
## Why this should be merged

`eth_getProof` is currently only available against committed revisions (`fwd_eth_get_proof` / `Revision.EthGetProof`). For archival nodes, we need the ability to call `eth_getProof` against reconstructed revisions as well. This PR exposes the same proof capability over a reconstructed revision.

## How this works

The Rust entry point is generalized over the handle type. A private helper `eth_get_proof_on<H: api::DbView + MetricsContextExt + ?Sized>` holds the key-parsing, null-handle, and proof logic; both `fwd_eth_get_proof` and the new `fwd_eth_get_proof_on_reconstructed` funnel through it and return the same `EthProofResult`. `ReconstructedHandle` already implements `DbView`, so the core `firewood::eth_get_proof` is reused unchanged.

On the Go side, an `ethGetProofCall` helper marshals the keys (pinned for the call), performs exactly one `fwd_eth_get_proof*` invocation, and decodes the result. `Revision.EthGetProof` and the new `Reconstructed.EthGetProof` both delegate to it.

## How this was tested

The existing `eth_getProof` tests were parameterized over both a committed revision and an equivalent reconstructed view, so each runs the full go-ethereum `trie.VerifyProof` verification against both:

- `TestEthGetProofAccountInclusionAndExclusion()`
- `TestEthGetProofStorageInclusionAndExclusion()`

A drop-then-use assertion for `Reconstructed.EthGetProof` was added to `TestReconstructedDropThenUse()`.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


ffi/src/proofs/eth.rs
```diff
@@ -92,6 +93,30 @@ fn to_trie_key(bytes: &[u8]) -> Result<[u8; 32], api::Error> {
...
+/// Shared body for the `fwd_eth_get_proof*` entry points: parse the keys and
+/// call the core proof function against `handle`. Returns
+/// [`EthProofResult::NullHandlePointer`] when `handle` is `None`.
+fn eth_get_proof_on<H>(
+    handle: Option<&H>,
+    account_key: BorrowedBytes<'_>,
+    storage_keys: BorrowedBytes2D<'_>,
+) -> EthProofResult
+where
+    H: api::DbView + crate::MetricsContextExt + ?Sized,
```
rkuris: I don't think `?Sized` is needed here (all callers are Sized)

ffi/src/proofs/eth.rs
```diff
@@ -135,17 +160,54 @@ pub extern "C" fn fwd_eth_get_proof(
...
+/// - `storage_keys` - An array of 32-byte slot trie keys (`keccak256(slot)`),
+///   in the order the proofs should be returned.
+///
+/// Callers are responsible for keccak-hashing addresses and slots into their
+/// 32-byte trie-key forms before calling this function.
+///
+/// # Returns
+///
+/// - [`EthProofResult::NullHandlePointer`] if `reconstructed` is null.
+/// - [`EthProofResult::NotSupported`] if the database is not running in
```
rkuris: Wall-o-text too big for me to pick it all. A one-line function means nobody will ever read this (including me). Let's trim this down and hand-write it please; nobody will read this and it can go stale super fast. Reference the function it calls too for parameters and such.

```suggestion
  /// Produce an `eth_getProof`-compatible proof against a reconstructed view 
  /// rather than a committed revision.
  ///
  /// See [`fwd_eth_get_proof`] for the proof format, arguments, return values,
  /// and key-encoding requirements.
  ///
  /// # Safety
  ///
  /// As [`fwd_eth_get_proof`], except `reconstructed` must be a valid pointer to
  /// a [`ReconstructedHandle`].
```

ffi/eth_proof.go
```diff
@@ -82,13 +85,53 @@ func (r *Revision) EthGetProof(accountKey []byte, slotKeys [][]byte) (*EthAccoun
...
+// caller is responsible for keccak-hashing addresses and storage slots into
+// their trie-key forms (firewood stores accounts at keccak256(address) and
+// slots at keccak256(address) ++ keccak256(slot)).
+//
+// Absent accounts come back with zero account scalars plus the empty-code and
+// empty-trie hashes; the proof bytes themselves distinguish inclusion from
+// exclusion to a verifier.
+//
+// It returns [ErrEthProofNotSupported] if the database is not in Ethereum hash
+// mode, and [ErrDroppedReconstructed] if this view has already been released.
```
rkuris: 🤖 This function does not do all this, it just calls something else that does, so move these docs there (except they are already there it seems, but please check)

What's wrong with:
```
// EthGetProof is [Revision.EthGetProof] evaluated against this reconstructed
// view. It returns [ErrDroppedReconstructed] if the view has been released.
```

If you want to be pedantic and add the `EthErrProofNotSupported` I won't complain much.

ffi/src/proofs/eth.rs
RodrigoVillar: Not needed! Removed here: [c6ddce2](https://github.com/ava-labs/firewood/pull/2059/commits/c6ddce24494a46367fae24ae3f38b6a7115c5fc9).

ffi/eth_proof.go
RodrigoVillar: I went with the doc comment you suggested here: [c6ddce2](https://github.com/ava-labs/firewood/pull/2059/commits/c6ddce24494a46367fae24ae3f38b6a7115c5fc9) (I did add the `EthErrProofNotSupported` error).

ffi/src/proofs/eth.rs
RodrigoVillar: Thank you for the feedback here! I went with your approach here: [c6ddce2](https://github.com/ava-labs/firewood/pull/2059/commits/c6ddce24494a46367fae24ae3f38b6a7115c5fc9).

rkuris (approved):

PR #2060, opened 2026-06-04
fix(proofs): reconcile partial-storage account values

### LINKED ISSUE #2042: fix(proofs): change-proof verification rejects honest partial-storage proofs (ethhash)
**Describe the bug**

In ethhash mode, **bounded change-proof** verification rejects honest proofs through an account at depth 64 whenever the proof's boundary cuts through that account's storage trie, leaving only a *subset* of its storage children in range.

`reconcile_branch_proof_node` byte-compares the account value from the proof against the value reconstructed from the proposal. Because `Preimage::write` recomputes the account's `storageRoot` (RLP field 2) from the storage children present at hash time, a proposal built from a partial set of storage children carries a *partial* storageRoot, while the proof carries the account's *full* on-disk value. Byte equality then fails and verification rejects an otherwise-valid proof with `ProofError::UnexpectedValue`.

**Trigger conditions**

1. An account at depth 64 has **K ≥ 2** storage children.
2. A change-proof boundary lands *between* two of that account's storage slots, so only a subset of the children are in range.
3. **Right-edge (end-bound) only** — storage children always sort *after* the account key, so only an end boundary can sit between an account and its storage. A left-edge cut leaves the account out-of-range (an ancestor `< start_key`), which reconcile already accepts.

**To Reproduce**

`test_change_proof_partial_storage_children_against_empty` exercises this: the `left_half_end_bound_only` and `left_edge_both_bounds` cases fail without the fix; `right_half_start_bound_only` passes (confirming the right-edge-only characterization).

**Expected behavior**

Verification succeeds for honest bounded change proofs that match the on-disk root, regardless of where the boundary falls relative to an account's storage children.

**Possible Solutions**

Ignore RLP field 2 (`storageRoot`) when comparing account values at depth 64 in `reconcile_branch_proof_node`. This is sound: `storageRoot` is write-only on the wire (always recomputed from children at hash time), and the final root-hash check remains the real gate, so relaxing the byte-compare to "all fields except storageRoot" adds no forgery surface. PR #2039 demonstrates this fix (its "bug 2") alongside the depth-64 fold fix (#2035).

Related: this is the completeness (false-reject) counterpart to the depth-64 fold defect in #2035; the two are independent and live in different code paths.

bernard-avalabs:
Closes #2042. Second half of the #2039 split.

  ## Why this should be merged

  Under `ethhash`, change-proof verification rejected valid proofs with
  `UnexpectedValue` whenever a depth-64 account's storage was only partially
  covered by the proof's range. A proposal built from a subset of an account's
  storage children gets a *partial* `storageRoot` spliced in by live hashing,
  while the proof carries the value over the account's *full* on-disk storage
  set. The two account values differ in exactly that one field, so the byte
  comparison in `reconcile_branch_proof_node` fails even though both hash
  identically.

  ## How this works

  - Relaxes `reconcile_branch_proof_node`: at account depth (64 nibbles), two
    account values that differ only in `storageRoot` (RLP field 2) reconcile as
    equal. `account_values_equal_except_storage_root` zeros field 2 on both
    sides and compares; a difference in any other field still conflicts.
  - Sound because `Preimage::write` always recomputes `storageRoot` from the
    node's current children at hash time, so the stored field is never
    authoritative on its own and the final root-hash check is what gates
    acceptance.
  - Right-edge only, and only for accounts with K≥2 storage children (K is the
    storage-child count): storage children sort after the account key, so only an
    end boundary can leave an in-range account with a subset of its storage; a
    single child (K=1) can't be a partial subset. Range proofs are unaffected:
    they rebuild the full in-range subtree and never byte-compare a subset-derived
    account value.

  ## How this was tested
  
  - New `tests/change/account.rs`: a change-proof matrix over a depth-64 account
    with four storage children, covering right-edge truncation (triggers),
    left-edge start bound (guard), both-bounds-inside-storage (guard),
    zero-in-range storage (the zero-storage extreme), and arbitrary batch
    ordering.
  - New contract test in `tests/reconcile.rs` pinning the relaxation to
    storageRoot-only diffs.
  - Verified by disabling the relaxation: the triggering cases fail with
    `UnexpectedValue`, the guards pass either way.
  - Passes CI.

  ## Breaking Changes
  
  None.


firewood/src/merkle/reconcile.rs
```diff
@@ -73,8 +76,50 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+        // full on-disk value. Both produce the same final hash because
+        // `Preimage::write` always recomputes storageRoot from the current
+        // children at hash time, but byte equality fails.
+        if cfg!(feature = "ethhash")
+            && proof_node.key.len() == ACCOUNT_DEPTH_NIBBLES
+            && let (Some(pv), Some(bv)) = (proof_value, branch.value.as_deref())
+            && account_values_equal_except_storage_root(pv, bv)
+        {
+            return Ok(());
+        }
```
demosdemon, line 97: This early-return fires for *any* depth-64 account pair that matches after zeroing `storageRoot` — on both start- and end-proof reconciliation, regardless of edge or storage-child count. The "right-edge only, K≥2" reasoning explains *why the divergence arises in practice*; it is not a guard the code enforces.

That's fine — it's safe because `verify_change_proof_root_hash` always re-checks the computed root against `end_root` (the `EndRootMismatch` check), so a spurious relaxation can never let a bad proof through. But that's the actual invariant the soundness rests on, and the comment doesn't say so. A future reader could mistake the scenario description for the safety boundary.

Suggest appending a sentence, e.g.:

```rust
// This early-return is safe regardless of how the divergence arose: the
// caller (`verify_change_proof_root_hash`) still gates acceptance on the
// final root-hash check, so relaxing here only avoids a spurious per-node
// `UnexpectedValue` — it never widens what proofs are accepted.
```

firewood/src/merkle/tests/change/account.rs
```diff
@@ -0,0 +1,286 @@
...
+/// key, so only an end boundary can truncate an account's storage while the
+/// account stays in range.
+/// - `left_half_end_bound_only` and `left_edge_both_bounds` keep the account in
+///   range with a storage subset → trigger the defect (fail pre-fix).
+/// - `right_half_start_bound_only` puts the account out of range via the start
+///   bound, so its value is never compared: a guard that passes with and
+///   without the fix.
+#[test_case(Bound::None, Bound::MidStorage ; "left_half_end_bound_only")]
+#[test_case(Bound::AtAccount, Bound::MidStorage ; "left_edge_both_bounds")]
+#[test_case(Bound::MidStorage, Bound::None ; "right_half_start_bound_only")]
```
demosdemon, line 94: The K=1 claim is asserted in prose but never exercised. The PR description argues "a single child (K=1) can't be a partial subset," but every case here uses the four-storage-child fixture (K=4). A regression that made the relaxation misbehave for single-child accounts wouldn't be caught by this matrix.

Risk is low given the root-hash backstop, but a one-line `#[test_case]` with a K=1 account whose single storage child is truncated out of range would lock the claim down and document the boundary.

firewood/src/merkle/reconcile.rs
```diff
@@ -73,8 +76,50 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+
+/// Two ethhash account RLP values are equivalent for hashing if they agree
+/// on every field except `storageRoot` (index 2). `Preimage::write` always
+/// recomputes that field from the current children, so the on-disk byte
+/// difference is invisible in the final hash.
+///
+/// Logs a warning when either side fails to parse as account RLP. The
+/// resulting `false` return falls through to `on_conflict`, which surfaces
+/// the conflict as `UnexpectedValue`. The warning gives operators a clearer
+/// signal that the underlying cause is malformed data, not a value mismatch.
```
demosdemon, line 118: nit: worth one more sentence stating this is **not** a general account-equality predicate.

The comparison is only valid for *hashing-equivalence*, and only because `Preimage::write` recomputes `storageRoot` from current children. Calling this in any context that doesn't re-derive `storageRoot` would silently treat two different accounts as equal. A short caveat here ("must not be used outside a context that re-derives `storageRoot`") guards against future misuse.

firewood/src/merkle/tests/change/account.rs
```diff
@@ -0,0 +1,286 @@
...
+/// The defect is **right-edge only**: storage children sort after the account
+/// key, so only an end boundary can truncate an account's storage while the
+/// account stays in range.
+/// - `left_half_end_bound_only` and `left_edge_both_bounds` keep the account in
+///   range with a storage subset → trigger the defect (fail pre-fix).
+/// - `right_half_start_bound_only` puts the account out of range via the start
+///   bound, so its value is never compared: a guard that passes with and
+///   without the fix.
+#[test_case(Bound::None, Bound::MidStorage ; "left_half_end_bound_only")]
+#[test_case(Bound::AtAccount, Bound::MidStorage ; "left_edge_both_bounds")]
```
demosdemon: nit: `left_edge_both_bounds` reads as "both bounds at the left edge," but this case is start `AtAccount` / end `MidStorage` — i.e. start exactly at the account, end mid-storage. It sits confusingly close to `test_change_proof_both_bounds_inside_account_storage`, which is the real both-bounds case. Consider `start_at_account_end_mid_storage` for clarity.

demosdemon (approved):

firewood/src/merkle/reconcile.rs
bernard-avalabs: Added.

firewood/src/merkle/reconcile.rs
bernard-avalabs: Added the extra sentence.

firewood/src/merkle/tests/change/account.rs
bernard-avalabs: Renamed.

firewood/src/merkle/tests/change/account.rs
bernard-avalabs: Addressed with a separate test (`test_change_proof_single_storage_child_truncated`) rather than a `#[test_case]`. The existing matrix can't express K=1: its cases all share `source_with_four_storage_children`. Adding a parameter for the number of children means abandoning the `setup_2nd_commit!` macro (it takes a literal list of puts), which is probably not worth it in this case.

PR #2064, opened 2026-06-07
fix(proofs): reject change proof that drops an in-range Put behind a left-edge exclusion boundary

rkuris:
Fixes the soundness gap tracked in #2008.

## Why this should be merged

Change-proof verification *accepted* a tampered proof that drops an in-range `Put` from `batch_ops` when the dropped key sits at the left-edge exclusion boundary. Both the structural and root-hash checks passed, so a malicious peer could omit a real mutation and still produce a proof that verifies against the honest root.

## How this works

In merkledb mode, values ≥ 32 bytes are serialized as `ValueDigest::Hash`. `reconcile_branch_proof_node`'s Hash-digest arm fell through to the `proof_value == branch.value` (`None == None`) short-circuit when the branch held no value, returning `Ok(())` without consulting `on_conflict`. A dropped in-range `Put` leaves the proving trie's branch empty while the boundary proof still carries the omitted key's hash, so the in-range guard (`node_nibbles >= start_key → UnexpectedValue`) was bypassed.

Fix: when the proof asserts a (hashed) value the branch does not satisfy — including an empty branch — defer to `on_conflict` instead of treating it as agreement. `on_conflict` now also receives the branch's current value (`Option<&[u8]>`) so the range-proof closure can still distinguish an out-of-range node (no branch value → accept the proof's stored hash) from a real mismatch (reject).

ethhash and small-value merkledb were already safe — they reach `on_conflict` via the `Value` digest arm.

## How this was tested

- `cargo nextest run --workspace --features ethhash,logger --all-targets` — 752 pass.
- `cargo clippy --workspace --features ethhash,logger --all-targets` — clean.
- `cargo doc --no-deps` — builds.
- The original soundness repro (merkledb fuzz seed `16855342578241928557`, scenario `M1_omit_put`) now rejects. The fuzzer that exercises it lives on #1951's branch and is not yet on `main`, so this PR carries the targeted reconcile unit tests; #1951 proves it end-to-end and lands separately.

## Breaking Changes

None. `reconcile_branch_proof_node` is `pub(crate)`; the `on_conflict` signature gained a parameter but has no external callers.


bernard-avalabs (changes requested):
The fix looks good to me. Two concerns that I think might be worth addressing before merging.

1. I suggest adding a deterministic regression test to exercise this case. Here is one that I believe works. start_key 0x1230 is absent and shares a prefix with the new key 0x1234, so the start exclusion proof descends to 0x1234, making the terminal proof node carry a hash digest. Dropping Put 0x1234 leaves the proving branch empty while that boundary node still carries the hash, so the old reconcile silently accepted it. The second op (0x90) keeps the batch non-empty so the tamper reaches reconcile instead of the structural check. I have verified that this test returns Ok(()) (false-accept) with the reconcile change reverted, and UnexpectedValue with the change.

```rust
#[test]
fn test_dropped_in_range_put_hashed_behind_left_exclusion_rejected() {
    let big = [0xabu8; 40]; // >= 32 bytes => Hash after serialization
    // start: 0x90 only. end: add NEW key 0x1234 (large value) AND change 0x90
    // (a second in-range op so the batch isn't empty after dropping 0x1234).
    let (source, target, root1_target, _ds, _dt) = setup_source_target![(b"\x90", b"anchor")];
    let (root1_source, root2) = setup_2nd_commit!(
        source,
        [
            (b"\x12\x34" as &[u8], big.as_slice()),
            (b"\x90", b"changed")
        ]
    );

    let valid = source
        .change_proof(root1_source, root2.clone(), Some(b"\x12\x30"), None, None)
        .unwrap();

    // Serialize so 0x1234's boundary node carries a Hash digest (value >= 32 bytes).
    let mut buf = Vec::new();
    valid.write_to_vec(&mut buf);
    let valid = crate::api::FrozenChangeProof::from_slice(&buf).unwrap();

    // Tamper: drop the in-range Put of 0x1234, keeping the 0x90 change.
    let mut ops: Vec<BatchOp<Key, Value>> = valid.batch_ops().to_vec();
    ops.retain(|op| !matches!(op, BatchOp::Put { key, .. } if key.as_ref() == b"\x12\x34"));
    let crafted = FrozenChangeProof::new(
        crate::Proof::new(valid.start_proof().as_ref().into()),
        crate::Proof::new(valid.end_proof().as_ref().into()),
        ops.into_boxed_slice(),
    );

    assert!(
        is_rejected(
            &target,
            &crafted,
            root2,
            Some(b"\x12\x30"),
            None,
            root1_target
        ),
        "dropped in-range Put (hashed) behind a left exclusion must be rejected"
    );
}
```

2.  The fix applies to both edges: `reconcile_branch_proof_node` routes the empty-branch-vs-`Hash` case to `on_conflict` regardless of which boundary proof it came from. Is it needed on the right edge? Perhaps adding a right edge test to demonstrate that it is needed or that this case is never reached would be helpful.

rkuris:
Both suggestions addressed in d47315ae:

1. Added your regression test (thanks for writing it!) as `test_crafted_dropped_hashed_put_behind_left_exclusion_rejected`, renamed to match the file's `test_crafted_*` convention. Also tightened it to assert the exact failure mode: `verify_change_proof_structure` must pass and `verify_and_check` must fail with exactly `UnexpectedValue`.

2. Good question on the right edge — it turns out the fix **is** needed there too. Added the mirrored `test_crafted_dropped_hashed_put_behind_right_exclusion_rejected`: `end_key 0x1238` is absent and shares a prefix with the new key `0x1234`, so the end exclusion proof descends to `0x1234` and carries its hash digest; dropping that Put leaves the proving branch empty. I verified empirically that with the reconcile change reverted, *both* tests false-accept (return `Ok`), and with it both reject with `UnexpectedValue` — so the empty-branch-vs-`Hash` routing to `on_conflict` is load-bearing on both edges.

bernard-avalabs (approved):

PR #2066, opened 2026-06-08
test(proofs): change-proof fuzz fixes and CI runtime scope-down

rkuris:
## Why this should be merged

Stacked on top of #1951 (`bernard/move-fuzz-test-fix-2`). Fixes several issues in the change-proof fuzz test that caused spurious failures, and brings its runtime down to something CI-appropriate.

## How this works

Four commits on top of Bernard's fuzz test:

1. **dump full proof on failure + `FIREWOOD_TEST_SOAK_SECONDS`** — on a failed assertion, dump the full proof (start/end proofs + batch ops) for reproduction, and add a soak knob that keeps generating fresh random seeds until a deadline for long local runs.
2. **M3 value-swap always changes the value** — the value-swap mutation could produce an identical value (e.g. empty old value), which the verifier correctly accepts, tripping the must-reject assertion spuriously. Now it always flips a bit (or produces a non-empty value).
3. **don't omit trailing change-proof ops** — a change proof's provable range is anchored by the last `Put`; a trailing run of deletes is unprovable (two deletes sharing a prefix are indistinguishable from a smaller valid range). The M1/M2 omit-mutations now skip when no `Put` follows, so dropping a trailing op yields a still-valid proof over a smaller range rather than a false "tamper". (Same gap affects `find_next_key` — noted as separate follow-up.)
4. **scope down iterations for CI** — the test ran ~15s in a debug build (25 fixtures × 50 inner scenarios). Cut the outer fixture count from 25→3 in debug (250→25 release), ~8–10× faster (~2s), keeping the 50 inner scenarios per fixture. `FIREWOOD_TEST_SOAK_SECONDS` remains for thorough local soaking.

## How this was tested

```
cargo nextest run --workspace --features logger --profile ci -E 'test(test_slow_change_proof_fuzz)'
```

Passes; runtime down from ~14.7s to ~2.4s on a fast laptop. Note: the fuzz module is `#[cfg(not(feature = "ethhash"))]`, so it must be run **without** `ethhash`.


bernard-avalabs (approved):
The changes look good to me. It fixes a few false positives that can cause flaky tests. The commit history seems a bit confused perhaps because it is a stacked PR but was merged with main.

PR #2076, opened 2026-06-09
chore(deps): bump Rust dependencies and fix breaking changes

RodrigoVillar:
## Why this should be merged

During #2075, I came across some breaking changes when trying to bump our Rust dependencies. As a result, I deferred bumping to this PR.

## How this works

Bumps the Rust dependency set skipped during the `v0.6.0` release; this handles the breaking changes that would've come with it:

- Bumped `sha2`/`sha3` and migrated `digest::generic_array::GenericArray` to digest `0.11`'s `array::Array`.
- `ctor`: the test logger initializer now needs #[ctor(unsafe)].
- `cbindgen`: regenerated `ffi/firewood.h `(adds C23 enums with a backward-compatible fallback).

This PR also keeps the metrics-related crates pinned to the `demosdemon/metrics` fork versions (bumping past them disables the patch), and moves the benchmark's `opentelemetry` pins to caret `0.32` to stay unified with `fastrace-opentelemetry` `0.18`.  

## How this was tested

CI

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/lib.rs
+++ b/firewood/src/lib.rs
@@ -178,7 +178,7 @@ mod batch_op;
 pub use firewood_storage::logger;
 
 #[cfg(all(test, feature = "logger"))]
-#[ctor::ctor]
+#[ctor::ctor(unsafe)]
 /// `ctor` will ensure this function is invoked before any tests are run so we
 /// can initialize the logger consistently across all tests without having to
 /// manually call it in each test.
--- a/firewood/src/lib.rs
+++ b/firewood/src/lib.rs
@@ -178,6 +178,8 @@ mod batch_op;
 pub use firewood_storage::logger;
 
 #[cfg(all(test, feature = "logger"))]
+// SAFETY: Called pre-main in test builds only. Does not panic (try_init + ok),
+// uses thread-safe logger initialization, and only reads env vars (safe pre-main).
 #[ctor::ctor(unsafe)]
 /// `ctor` will ensure this function is invoked before any tests are run so we
 /// can initialize the logger consistently across all tests without having to
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,9 +171,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "askama"
-version = "0.15.6"
+version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9b8246bcbf8eb97abef10c2d92166449680d41d55c0fc6978a91dec2e3619608"
+checksum = "f1bf825125edd887a019d0a3a837dcc5499a68b0d034cc3eb594070c3e18addc"
 dependencies = [
  "askama_macros",
  "itoa",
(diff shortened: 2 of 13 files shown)


rkuris:
A few things to fix here, nothing too major.The biggest benefit of digest::array::Array over the old type is that it implements From and Into for standard Rust arrays seamlessly. You no longer need to use .as_slice() hacks to get a normal array out of it.

firewood/src/lib.rs
```diff
@@ -178,7 +178,7 @@ mod batch_op;
 pub use firewood_storage::logger;
 
 #[cfg(all(test, feature = "logger"))]
-#[ctor::ctor]
+#[ctor::ctor(unsafe)]
```
rkuris, line 183: All usage of unsafe should have a safety clause explaining why this is safe. In this case, it's pretty safe, but the rules as defined by the docs are:

* Must not panic
* Must be safe to call from any thread
* Must not rely on other statics being initialized
* Must not cause recursive initialization

Maybe just a simple comment here up front at a minimum; we could also clean up the comments below some.

```suggestion
// SAFETY: Called pre-main in test builds only. Does not panic (try_init + ok),
// uses thread-safe logger initialization, and only reads env vars (safe pre-main).
#[ctor::ctor(unsafe)]
```

rkuris (changes requested):
The demosdemon/metrics fork is getting pretty stale, now 32 commits (2 months) behind upstream and they haven't accepted it yet. We'll need to upgrade and maintain this fork, but I think nobody has permission to do this except @demosdemon right now. Perhaps we should move this fork to ava-labs and vendor the fork. I pinged on the original PR https://github.com/metrics-rs/metrics/pull/686 for attention there, maybe we'll get a response.

The other thing that needs fixing here are all the unnecessary .as_slice() calls after migrating to digest::array::Array. AI analysis spotted these, could use verification (I checked on the second one -- it's legit and a perf win)

Lines 64, 73, 88 in storage/src/hashtype/ethhash.rs:
- Change .as_slice() → .as_ref() for consistency with Rust idioms

Lines 462, 621 in firewood/src/merkle/tests/ethhash.rs:
- Change Box::from(digest.as_slice()) → Box::from(digest.into())
- This is more efficient (no intermediate slice borrow)

Lines 162, 235, 263 in storage/src/hashers/ethhash.rs:
- Current usage is fine, but could change to .as_ref() for consistency
- Not incorrect, just style preference

firewood/src/lib.rs
RodrigoVillar: Done: 05e5c8f

RodrigoVillar:
Discussed offline:

- Lines 64, 73, 88 in storage/src/hashtype/ethhash.rs: not a clean fix as this requires type annotations
- Lines 462, 621 in firewood/src/merkle/tests/ethhash.rs: there is no performance win
- Lines 162, 235, 263 in storage/src/hashers/ethhash.rs: applied!
- Created an issue regarding the metrics crate here: #2077

rkuris (approved):

PR #2078, opened 2026-06-10
perf: skip using FDL when running in archival mode

### LINKED ISSUE #2015: perf(storage): skip future-delete log when root_store (archival mode) is enabled

In archival mode (`ConfigManager.root_store = true`), the future-delete log (FDL) — `Propose::deleted` carried forward into `Committed::deleted` — is built on every commit, cloned into every committed revision, retained in memory for `max_revisions` revisions, and then **discarded without ever being consumed**.

`PersistWorker::reap` already detects this and short-circuits when `root_store` is set:

```rust
// firewood/src/persist_worker.rs:194-204
pub(crate) fn reap(&self, nodestore: NodeStore<Committed, FileBacked>) -> Result<(), PersistError> {
    if self.shared.root_store.is_none() && self.shared.channel.reap(nodestore).is_err() {
        self.join_handle();
        self.check_error()?;
    }
    Ok(())
}
```

When `root_store` is `Some(_)`, the reaped `NodeStore` (and its `deleted` list) is dropped on the floor. No free-list maintenance is performed because archival mode preserves old nodes on disk for historical queries.

For archive-node workloads (e.g. C-Chain reexecution), where every block COW-replaces thousands of trie nodes, this is non-trivial memory:

- ~48 B per FDL entry × deletions-per-block × `max_revisions`
- e.g. 5,000 deletions/block × 1,024 revisions ≈ ~250 MB of FDL never read.

It also costs `Arc<Mutex<MaybePersisted>>` allocations and `Vec` growth on the commit hot path.

Skip FDL tracking entirely when the database was opened with `root_store` enabled, so `Propose::deleted` (and the cloned `Committed::deleted`) stays empty.

The same pattern already exists for reconstruction nodestores. `MutableKind::track_deleted` (`storage/src/nodestore/mod.rs:885`) is a push for `Propose` and a no-op for `Recon`:

```rust
impl<S> MutableKind for Recon<S> {
    fn track_deleted(&mut self, _node: MaybePersistedNode) {
        // Reconstruction views are ephemeral; no delete list to maintain.
    }
}
```

We want `Propose::track_deleted` (and the explicit `delete_node` / `delete_nodes` helpers at `storage/src/nodestore/mod.rs:339-352`) to behave the same way when archival mode is enabled.

**Audit of `deleted` consumers** (to confirm nothing else breaks):

- `as_committed` — clones `Propose::deleted` into `Committed::deleted` (would clone empty).
- `deleted_len()` — feeds the `DELETED_LIST_LEN` gauge (would always report 0 in archival; acceptable / arguably correct).
- `deleted()` — `#[cfg(test, feature = "test_utils")]` accessor.
- `reap_deleted` — invalidates cached nodes and returns nodes to the free list. **Already unreachable when `root_store.is_some()`.**

No proof, diff, or view code reads `deleted`.

1. **Status quo** — leave it. Simple, but wastes memory proportional to `max_revisions`, which is exactly the knob archive operators want to crank.
2. **Drop `deleted` only at the `Committed` boundary** (i.e. clear it during `as_committed` in archival mode). Saves the in-memory retention cost but still pays the allocation/push cost on every mutation during proposal building. Strictly worse than skipping at the source.
3. **Skip the channel send in `PersistWorker::reap` more aggressively** — already done; this issue is about the upstream cost, not the reap path.

**Additional context**

Suggested implementation sketch:

1. Carry an "archival" flag on `FileBacked` (or thread it onto `Propose` at `NodeStore::new` time from the parent storage).
2. In `Propose::track_deleted` / `delete_node` / `delete_nodes`, short-circuit to a no-op when the flag is set.
3. Add a unit test in `firewood/src/manager.rs`: open a DB with `root_store=true`, commit a proposal that mutates an existing key, assert `committed.deleted_len() == 0`.
4. No change required to `reap_deleted` — it remains unreachable in archival mode.

Stretch: the `invalidate_cached_nodes` call inside `reap_deleted` is also archival-irrelevant, but that already falls out for free since `reap_deleted` is not invoked when `root_store.is_some()`.

Risk: low. The reap path is already gated on `root_store`; this change moves the gate one layer earlier so the unused list is never built.

RodrigoVillar:
## Why this should be merged

When archival mode is enabled, the future-delete log is built on every proposal, cloned into each committed revision, and retained for max_revisions but is never consumed: `PersistWorker::reap` short-circuits when `RootStore` is present, since old nodes must be preserved on disk for historical queries. This is wasted work (an allocation/push per replaced node on the hot proposal path) and wasted memory (the deleted lists pinned by every retained revision).

## How this works

`NodeStore<T, S>` gains a `deleted_node_tracking: DeletedNodeTracking` field, a two-variant enum (`Enabled` = build the future-delete log, `Disabled` = skip it). It is set at the root constructors and propagated by every derived constructor, so the whole nodestore lineage of a database shares one value. The manager passes `DeletedNodeTracking::Disabled` when opening the database with `root_store` enabled, and `Enabled` otherwise.

When tracking is `Disabled`, all four FDL population sites become no-ops: 
- `read_for_update`
- `delete_node`
- `delete_nodes` 
- `NodeStore::new`

Each guard branches on an immutable flag, so it is perfectly predicted and effectively free next to the adjacent node-cache lookup.

## How this was tested

Added two UTs to test if the future-delete log is built in archival and non-archival mode, respectively:
- `test_no_fdl_when_root_store_enabled()`
- `test_fdl_tracked_when_root_store_disabled()`

## Breaking Changes

- [ ] firewood
- [X] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -27,9 +27,9 @@ use firewood_metrics::{GaugeExt, firewood_counter, firewood_gauge, firewood_hist
 pub use firewood_storage::CacheReadStrategy;
 use firewood_storage::RootStore;
 use firewood_storage::{
-    BranchNode, Committed, CommittedId, FileBacked, FileIoError, HashedNodeReader,
-    ImmutableProposal, Mutable, MutableKind, NodeHashAlgorithm, NodeStore, NodeStoreHeader,
-    Propose, Recon, TrieHash,
+    BranchNode, Committed, CommittedId, DeletedNodeTracking, FileBacked, FileIoError,
+    HashedNodeReader, ImmutableProposal, Mutable, MutableKind, NodeHashAlgorithm, NodeStore,
+    NodeStoreHeader, Propose, Recon, TrieHash,
 };
 
 pub(crate) const DB_FILE_NAME: &str = "firewood.db";
@@ -196,12 +196,17 @@ impl RevisionManager {
             Err(err) => return Err(err.into()),
         };
         // `root_store` enables archival mode: old nodes are preserved on disk
-        // for historical queries, so disable delete tracking and skip the
-        // future-delete log (#2015).
+        // for historical queries, so disable delete node tracking and skip the
+        // future-delete log.
+        let deleted_node_tracking = if config.root_store {
+            DeletedNodeTracking::Disabled
+        } else {
+            DeletedNodeTracking::Enabled
+        };
         let nodestore = Arc::new(NodeStore::open(
             &header,
             storage.clone(),
-            !config.root_store,
+            deleted_node_tracking,
         )?);
         let root_store = config
             .root_store
@@ -546,9 +551,14 @@ impl RevisionManager {
         //    against the file backing carried by the latest committed revision.
         if HashKey::default_root_hash().as_ref() == Some(&root_hash) {
             let storage = self.current_revision().storage().clone();
+            let deleted_node_tracking = if self.root_store.is_some() {
+                DeletedNodeTracking::Disabled
+            } else {
+                DeletedNodeTracking::Enabled
+            };
             return Ok(Arc::new(NodeStore::new_empty_committed(
                 storage,
-                self.root_store.is_none(),
+                deleted_node_tracking,
             )));
         }
 
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -195,9 +195,7 @@ impl RevisionManager {
             }
             Err(err) => return Err(err.into()),
         };
-        // `root_store` enables archival mode: old nodes are preserved on disk
-        // for historical queries, so disable delete node tracking and skip the
-        // future-delete log.
+
         let deleted_node_tracking = if config.root_store {
             DeletedNodeTracking::Disabled
         } else {
--- a/firewood/src/manager.rs
+++ b/firewood/src/manager.rs
@@ -195,7 +195,14 @@ impl RevisionManager {
             }
             Err(err) => return Err(err.into()),
         };
-        let nodestore = Arc::new(NodeStore::open(&header, storage.clone())?);
+        // `root_store` enables archival mode: old nodes are preserved on disk
+        // for historical queries, so disable delete tracking and skip the
+        // future-delete log (#2015).
+        let nodestore = Arc::new(NodeStore::open(
+            &header,
+            storage.clone(),
+            !config.root_store,
+        )?);
         let root_store = config
             .root_store
             .then(|| {
@@ -539,7 +546,10 @@ impl RevisionManager {
         //    against the file backing carried by the latest committed revision.
         if HashKey::default_root_hash().as_ref() == Some(&root_hash) {
             let storage = self.current_revision().storage().clone();
-            return Ok(Arc::new(NodeStore::new_empty_committed(storage)));
+            return Ok(Arc::new(NodeStore::new_empty_committed(
+                storage,
+                self.root_store.is_none(),
+            )));
         }
 
         // 3. Check `RootStore` (if it exists).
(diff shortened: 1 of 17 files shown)


rkuris:
Should be an enum instead of bare bools

RodrigoVillar:
@rkuris refactored this PR to use an enum (`DeletedNodeTracking`) enum, PTAL.

firewood/src/merkle/collapse.rs
```diff
@@ -408,13 +408,14 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
-    use firewood_storage::MemStore;
+    use firewood_storage::{DeletedNodeTracking, MemStore};
 
     use super::*;
 
     fn create_test_merkle() -> Merkle<NodeStore<Mutable<Propose>, MemStore>> {
         let memstore = MemStore::default();
-        let nodestore = NodeStore::new_empty_proposal(memstore.into());
+        let nodestore =
+            NodeStore::new_empty_proposal(memstore.into(), DeletedNodeTracking::Enabled);
```
rkuris, line 418: nit: arguably most tests would be fine with `DeletedNodeTracking::Disabled` and some might even benefit from being run with both modes.

firewood/src/manager.rs
```diff
@@ -195,7 +195,19 @@ impl RevisionManager {
             }
             Err(err) => return Err(err.into()),
         };
-        let nodestore = Arc::new(NodeStore::open(&header, storage.clone())?);
+        // `root_store` enables archival mode: old nodes are preserved on disk
+        // for historical queries, so disable delete node tracking and skip the
+        // future-delete log.
```
rkuris: nit: This is a bit tautological and is easily inferred by reading the code; this comment could be more terse.

firewood/src/manager.rs
```diff
@@ -539,7 +551,15 @@ impl RevisionManager {
         //    against the file backing carried by the latest committed revision.
         if HashKey::default_root_hash().as_ref() == Some(&root_hash) {
             let storage = self.current_revision().storage().clone();
-            return Ok(Arc::new(NodeStore::new_empty_committed(storage)));
+            let deleted_node_tracking = if self.root_store.is_some() {
+                DeletedNodeTracking::Disabled
+            } else {
+                DeletedNodeTracking::Enabled
+            };
```
rkuris, line 556: This is soo much more readable than what was there before! ⭐

fwdctl/src/check.rs
```diff
@@ -72,7 +73,9 @@ pub(super) fn run(opts: &Options) -> Result<(), api::Error> {
     };
 
     let mut header = NodeStoreHeader::read_from_storage(storage.as_ref())?;
-    let nodestore = NodeStore::open(&header, storage)?;
+    // Open with delete tracking enabled: the checker repairs free lists, which
+    // requires it regardless of how the database is normally run.
```
rkuris: I think this is a false statement. The checker doesn't make a mutable merkle that uses a delete log so this actually doesn't matter, and if it ever does, it probably needs to make a decision at the callsite on whether to delete an updated node.

Being a bit more concrete here, let's say the checker decides "oh this node is wrong, it should contain a new value starting at this revision". Older revisions might still point to the old node, and the checker may not know this or care to find out, so it's probably safest to set this based on whether root_store is enabled or not.

demosdemon (approved):

rkuris (approved):

firewood/src/merkle/collapse.rs
RodrigoVillar: Created #2079.

firewood/src/manager.rs
RodrigoVillar: Removed this comment.

fwdctl/src/check.rs
RodrigoVillar: > probably safest to set this based on whether root_store is enabled or not.

I went with this approach here: [4852d42](https://github.com/ava-labs/firewood/pull/2078/commits/4852d42f739206844be9591585f872a3c1ab847a)

Issue #352, opened 2023-11-16
Make RangeProof return the next key as part of the return value


rkuris:
Consider adding an Option<Vec<[u8]>> to RangeProof

rkuris:
This would need to bubble back through grpc

demosdemon:
Provide next key in result to allow range scan to resume from last key.

rkuris:
We're removing the grpc layer, but we could still use this call through ffi.

demosdemon:
After applying a range proof, firewood should be able to inspect the applied proof and determine the "holes" in the current database compared to the peer that provided the range proof. Where a hole is a minkey/maxkey range where we identified _a_ difference in our state from the range-proof state and the final max key in the set of holes may be unbounded to represent beyond all known states.

Additionally, we should be able to identify many ranges and not just one allowing synchronizers to parallelize peer requests for each detected hole.

From an unknown state, the request for a complete range is unlikely to generate a complete range proof. But, it includes a proof identifying all of the partial paths of each child of the root node. This applies recursively for each discovered node in the trie. If we have a matching hash for the node, we can skip descending as we have all data and keys under that node. This is true even if the key proof does not directly include these extra keys because a matching hash validates the state.

demosdemon:
Status: there is currently a `find_next_key` method and it returns a technically correct value; however, the value it returns is not optimal for the sync algorithm.

rkuris:
Closed in favor of some new ideas.

PR #2081, opened 2026-06-13
fix(proofs): implement proper next key backtracking for range proofs

dvansari65:
Implements proper backtracking logic for find_next_key_after_range_proof to calculate the exact next branch prefix based on the end proof nodes, resolving issue #352.

## Why this should be merged
This PR resolves Issue #352 by replacing the naive implementation of find_next_key_after_range_proof, which merely returned the last key in the range, with a precise calculation of the next key. The previous approach was inefficient as it forced the client to redundantly fetch the same key again or blindly guess the next boundary, leading to overhead during state synchronization.

## How this works
The implementation now uses a robust backtracking algorithm directly on the end_proof nodes provided by the range proof:

It traverses the end_proof nodes backwards (from the deepest node to the root) to locate the deepest common ancestor with last_key.
Once the divergence point is found, it scans the node's child_hashes to identify the very first populated branch strictly to the "right" of the path last_key took.
It reconstructs the prefix for this rightward branch. If the prefix has an odd nibble length, it pads it with a 0x0 nibble. This perfectly aligns with Firewood's inclusive start_key semantics, as it formulates the absolute lower-bound byte sequence for that branch.
Finally, it validates that the newly computed next_key has not exceeded end_key (if provided) before returning it to the client. If the trie is exhausted, it correctly returns None.
## How this was tested
The changes were verified locally by running the entire test and fuzzer suite across all workspace targets with the required environment features enabled:
cargo test --features ethhash,logger --all-targets

All 403 firewood tests, as well as the remaining workspace crates (712 tests total), passed successfully. This confirms that the logic correctly navigates the exclusion/inclusion proofs of the trie without breaking existing Merkle tree verification logic.

## Breaking Changes
None

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


firewood/src/proofs/range.rs
```diff
@@ -155,7 +153,67 @@ pub fn find_next_key_after_range_proof(
...
+        if last_key_nibbles.starts_with(&node_nibbles) {
+            let next_nibble = last_key_nibbles.get(node_nibbles.len());
+            let start_child_idx = match next_nibble {
+                Some(&n) => n + 1,
+                None => 0, // If last_key matches perfectly, next keys are in its children
+            };
+
+            for idx in start_child_idx..16 {
+                let component = firewood_storage::PathComponent::try_new(idx)
+                    .expect("valid child index between 0 and 15");
```
demosdemon: This `expect` is avoidable. `PathComponent` has an [ALL](https://ava-labs.github.io/firewood/firewood_storage/struct.PathComponent.html#associatedconstant.ALL) associated constant.

firewood/src/proofs/range.rs
```diff
@@ -155,7 +153,67 @@ pub fn find_next_key_after_range_proof(
...
+                
+                if node.child_hashes[component].is_some() {
+                    let mut next_nibbles = node_nibbles.clone();
+                    next_nibbles.push(idx);
+
+                    // Pad odd nibble length with 0 (a 0x0 nibble) to align to a full byte
+                    if next_nibbles.len() % 2 != 0 {
```
demosdemon: this emits a clippy warning

```suggestion
                    if !next_nibbles.len().is_multiple_of(2) {
```

firewood/src/proofs/range.rs
```diff
@@ -155,7 +153,67 @@ pub fn find_next_key_after_range_proof(
...
+        firewood_storage::NibblesIterator::new(last_key.as_ref()).collect();
+
+    for node in end_nodes.iter().rev() {
+        let node_nibbles: Vec<u8> = node.key.iter().map(|c| c.as_u8()).collect();
+
+        // Find the deepest node that is an ancestor of last_key
+        if last_key_nibbles.starts_with(&node_nibbles) {
+            let next_nibble = last_key_nibbles.get(node_nibbles.len());
+            let start_child_idx = match next_nibble {
+                Some(&n) => n + 1,
```
demosdemon: the +1 triggers a lint warning. wrapping_add is safe here because `c.as_u8()` guarantees the value is in the range `0..16`.

demosdemon (changes requested):
There are CI failures for formatting issues and lint warnings; these should be addressed.

There are no tests covering the new algorithm. A suggested test matrix: (a) next key is a deeper child of `last_key`'s node; (b) next key is a right-sibling requiring one walk-up; (c) `last_key` at nibble 15 forcing multi-level walk-up; (d) computed next key > `end_key` → `None`; (e) trie exhausted → `None`; (f) odd-nibble branch prefix padding.

dvansari65:
@demosdemon  review this pull request

dvansari65:
hey @demosdemon  , this is so inappropriate to say but  i emailed you few weeks ago with email id : dvansari360@gmail.com, can you check it out

PR #2082, opened 2026-06-14
feat(fwdctl): add import command

dvansari65:
## Why this should be merged

There is no way to rebuild or restore a database from exported CSV dumps. This adds the `fwdctl import` subcommand to reconstruct databases from dumps produced by `fwdctl dump`.

## How this works

Adds `fwdctl import` which streams CSV data, batches it into configurable chunks (default 10,000 records), and commits them. Features:

- Configurable `--batch-size` and `--status-interval` (default 1m)
- Optional `--hex` flag for hex-encoded keys/values
- Reads from a file (`--input-file-name`) or stdin
- Skips malformed rows with warnings; aborts after 100 consecutive errors
- Reports final summary: `imported N keys (skipped M malformed rows) in T (R keys/s)`

JSON support is deferred to a follow-up PR.

## How this was tested

Integration tests in `fwdctl/tests/cli.rs`:
- `test_slow_fwdctl_import_csv` — round-trip export/import
- `test_slow_fwdctl_import_csv_hex` — hex-encoded round-trip
- `test_slow_fwdctl_import_csv_malformed` — strict 2-column constraint and hex-parse error recovery
- `test_slow_fwdctl_import_large_random_database` — export, import, and compare a 100k-key random database

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [x] fwdctl (adds `import` subcommand)


fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+        value_name = "INPUT_FILE_NAME",
+        help = "Input file name of database dump."
+    )]
+    pub input_file_name: PathBuf,
+
+    #[arg(short = 'x', long, help = "Parse the keys and values as hex format.")]
+    pub hex: bool,
+}
+
+const BATCH_SIZE: usize = 1000;
```
rkuris: This is a great default but this should be a command line option. I bet performance improves if you increase this to 10k or 100k, or even as high as 1m. The option can just get added to the block above this.

firewood/src/proofs/range.rs
```diff
@@ -136,8 +136,6 @@ pub fn find_next_key_after_range_proof(
     proof: &FrozenRangeProof,
     verification: &RangeProofVerificationContext,
 ) -> Result<Option<KeyRange>, api::Error> {
-    // TODO(#352): proper implementation, this naively returns the last key
```
rkuris, line 139: Looks like some of the other PR you're submitting bled into this one. You can change the base if they are dependent, or just start off main again with this one, but these changes should not be part of the change set.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // Use a closure or macro to push ops and flush batches.
+    let mut push_op = |key: Vec<u8>, value: Vec<u8>| -> Result<(), api::Error> {
+        batch_ops.push(BatchOp::Put { key, value });
+        if batch_ops.len() >= BATCH_SIZE {
+            let mut batch = Vec::with_capacity(BATCH_SIZE);
+            std::mem::swap(&mut batch_ops, &mut batch);
+            let proposal = db.propose(batch)?;
+            proposal.commit()?;
+            total_imported = total_imported.saturating_add(BATCH_SIZE);
```
rkuris: This can never overflow (adding "usize" items to firewood would take forever), let's use wrapping_add which is faster, along with a comment.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+            log::info!("Imported {total_imported} records so far...");
```
rkuris: This seems spammy to me. Let's gate this message on a progress interval timer so that we only get it every minute. Ideally this is also configured (1m is the default). Consider using the humantime crate for parsing and rendering the duration.

Rate info here would also be awesome -- maybe "{total_imported} keys in {duration} ({rate} keys/s)"

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[arg(short = 'x', long, help = "Parse the keys and values as hex format.")]
+    pub hex: bool,
+}
+
+const BATCH_SIZE: usize = 1000;
+
+pub(super) fn run(opts: &Options) -> Result<(), api::Error> {
+    log::debug!("import database {opts:?}");
+
+    if opts.input_format == OutputFormat::Dot || opts.input_format == OutputFormat::Stdout {
```
rkuris: clap supports checking this for you. If you use clap, it will provide a better usage clause.

Also not opposed to creating a separate `InputFormat` type here, which would be a bit cleaner.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    let mut batch_ops: Vec<BatchOp<Vec<u8>, Vec<u8>>> = Vec::with_capacity(BATCH_SIZE);
+    let mut total_imported: usize = 0;
+
+    // Use a closure or macro to push ops and flush batches.
+    let mut push_op = |key: Vec<u8>, value: Vec<u8>| -> Result<(), api::Error> {
+        batch_ops.push(BatchOp::Put { key, value });
+        if batch_ops.len() >= BATCH_SIZE {
+            let mut batch = Vec::with_capacity(BATCH_SIZE);
+            std::mem::swap(&mut batch_ops, &mut batch);
```
rkuris: Let's avoid allocating and deallocating memory on each batch (performance). Call db.propose with a borrowed batch, then call batch.clear() to reset it without deallocating the memory.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+        default_value_t = OutputFormat::Json,
+        help = "Input format of database dump, default to json. CSV and JSON formats are available."
+    )]
+    pub input_format: OutputFormat,
+
+    /// The input file name of database dump.
+    #[arg(
+        short = 'f',
+        long,
+        required = true,
```
rkuris: Can we make this optional? If not present, read from stdin.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+    };
+
+    match opts.input_format {
+        OutputFormat::Csv => {
+            let mut rdr = csv::ReaderBuilder::new()
+                .has_headers(false)
+                .from_reader(file);
+
+            for result in rdr.records() {
+                let record = result.map_err(|e| api::Error::InternalError(Box::new(e)))?;
```
rkuris: This is pretty harsh; a single malformed row means we abort the entire input, even the last N rows not yet committed. Consider logging an error and skipping malformed rows.

One minus to this is that if there are a lot of malformed rows you might get a wall of errors; consider some limit, especially if no successful rows are being processed (input from /dev/urandom for example).

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+                if record.len() >= 2 {
+                    let key = parse_string(record.get(0).expect("len checked"), opts.hex)?;
+                    let value = parse_string(record.get(1).expect("len checked"), opts.hex)?;
+                    push_op(key, value)?;
+                }
+            }
```
rkuris: This silently skips rows with fewer columns, and ends out using expect way more than needed here.

Consider something like this:
```
for result in rdr.deserialize() {
    let (key, value): (String, String) =
        result.map_err(|e| api::Error::InternalError(Box::new(e)))?;
    let key = parse_string(&key, opts.hex)?;
    let value = parse_string(&value, opts.hex)?;
    push_op(key, value)?;
}
```

Same caveat about InternalError here; we probably should just skip bad rows.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+        OutputFormat::Json => {
+            let reader = BufReader::new(file);
+            for line in reader.lines() {
```
rkuris: This is hand-parsing JSON line-by-line, which breaks the moment the input isn't formatted exactly as expected: a single-line object, escaped quotes inside a value ("\""), unicode escapes, whitespace variations, numbers/booleans/nulls as values, nested objects. It's effectively a fragile regex pretending to be a parser.

Probably better to split this into a separate PR adding support for JSON. When you do that, you'll probably want serde's StreamDeserializer, quite a bit more complex than what you've hand-rolled here.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // Insert remaining ops
+    if !batch_ops.is_empty() {
+        total_imported = total_imported.saturating_add(batch_ops.len());
+        let proposal = db.propose(batch_ops)?;
+        proposal.commit()?;
+    }
+
+    log::info!("Successfully imported {total_imported} records.");
+    println!("Successfully imported {total_imported} records.");
```
rkuris: This is likely to produce two lines of output

fwdctl/tests/cli.rs
```diff
@@ -397,3 +398,220 @@ fn test_slow_fwdctl_check_db_with_data() {
             .success();
     });
 }
+
+#[test]
```
rkuris, line 402: We need some negative tests -- malformed input, maybe even /dev/urandom.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,151 @@
...
+        short = 'f',
+        long,
+        required = true,
+        value_name = "INPUT_FILE_NAME",
+        help = "Input file name of database dump."
+    )]
+    pub input_file_name: PathBuf,
+
+    #[arg(short = 'x', long, help = "Parse the keys and values as hex format.")]
+    pub hex: bool,
```
rkuris, line 56: This should probably be the default. Almost all databases will contain non-hex data.

rkuris (changes requested):
Required for next review:
 - Remove unrelated changes from this PR
 - Remove json, just focus on CSV in this PR (too many issues with the json for a solid review here)
 - Improve malformed input handling (no columns or 1 column shouldn't abort the dump in the middle, and 3+ columns shouldn't be silently ignored)
 - Make the batch size configurable
 - Reuse memory on each batch
 - No log spamming
 - Malformed input testing

Recommended, but could be skipped:
 - Create a random binary database (maybe one with ~100k rows) export then import it, and compare the results.
 - Rate information status updates

firewood/src/proofs/range.rs
rkuris: Still seeing the `find_next_key_after_range_proof` rewrite and its test in this changeset. Can you drop these before we go further? Easiest is to re-base this branch on a fresh `main` so only the `fwdctl` files remain — happy to help if the rebase is fiddly.

fwdctl/src/import.rs
rkuris: Still doubled — `log::info!("{final_msg}")` and `println!("{final_msg}")` both fire, so with the `logging` feature on you get two identical lines. Keep the `println!` (user-facing) and drop the `log::info!`.

fwdctl/src/import.rs
rkuris: Both `total_imported` sites are still `saturating_add`. `usize` overflow is unreachable here (you'd be importing for centuries), so `wrapping_add` with a one-line comment is the convention we use elsewhere.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,214 @@
...
+                .from_reader(reader);
+
+            let mut consecutive_errors = 0;
+            const MAX_CONSECUTIVE_ERRORS: usize = 100;
+
+            for (row_idx, result) in rdr.deserialize().enumerate() {
+                let (key_str, value_str): (String, String) = match result {
+                    Ok(r) => r,
+                    Err(e) => {
+                        consecutive_errors += 1;
```
rkuris: This `consecutive_errors += 1; if … > MAX { return Err(…) }` block is now copy-pasted three times (deserialize / key-parse / value-parse). Pull it into one helper so the limit logic lives in a single place.

fwdctl/src/import.rs
```diff
@@ -0,0 +1,214 @@
...
+    let total_duration = start_time.elapsed();
+    let duration_secs = total_duration.as_secs_f64();
+    let final_rate = if duration_secs > 0.0 {
+        (total_imported as f64 / duration_secs) as usize
+    } else {
+        0
+    };
+    let formatted_total_duration = format_duration(Duration::from_secs(total_duration.as_secs()));
+
+    let final_msg = format!(
```
rkuris, line 116: Since malformed rows are skipped, consider reporting the skip count here too (`imported N, skipped M`) — otherwise a half-garbage dump loads silently with no signal that data was dropped.

fwdctl/src/import.rs
rkuris: Nice — the `#[derive(clap::ValueEnum)] InputFormat` enum is exactly what I had in mind, and clap now handles the validation and usage clause. Resolved.

rkuris:
Thanks for the updates — most of the earlier feedback is addressed (InputFormat enum, configurable batch size + status interval, stdin input, malformed-row skipping, hex defaulting off, JSON deferred). Nice work.

**Open before merge:**

1. **Drop the `firewood/src/proofs/range.rs` changes.** The `find_next_key_after_range_proof` rewrite + test are unrelated to `fwdctl import` and shouldn't be in this PR — please re-base on a fresh `main` so only the `fwdctl` files remain.
2. **Final output prints twice** — both `log::info!` and `println!` emit `final_msg`; keep the `println!` and drop the `log::info!`.
3. **`saturating_add` → `wrapping_add`** (with a comment) for the two `total_imported` sites; overflow is unreachable.

**Nice-to-have:**

4. De-duplicate the triplicated `consecutive_errors` limit block into one helper.
5. Report the skip count in the final summary (`imported N, skipped M`) so a partially-bad dump doesn't load silently.

(1)–(3) are the blockers; (4)–(5) are cleanups. See the inline threads for specifics.

rkuris (changes requested):

dvansari65:
@rkuris  I've rebased this on a fresh main to completely drop the range.rs changes. I also added the skip count to the final output, abstracted the malformed row error handling into a single DRY closure, swapped saturating_add to wrapping_add, and dropped the duplicate log::info!

rkuris:
@copilot resolve the merge conflicts in this pull request

rkuris (approved):

PR #2089, opened 2026-06-18
ci: serve the mdBook at the GitHub Pages site root

demosdemon:
### Why this should be merged

The GitHub Pages deployment currently lands on raw rustdoc. This PR — the second of
three stacked changes, on top of the documentation-site foundation — makes the mdBook
the front door of the published site: generated rustdoc moves under `/rustdoc/`, the Go
docs and benchmark dashboards remain reachable from the book, and a post-deploy smoke
check guards the result. It also relocates the architecture diagram into the book
source, atomically with removing the workflow step that used to copy it.

### How this works

Output assembly in `.github/workflows/gh-pages.yaml` moves from the cargo-owned
`target/doc/` into a `site/` staging directory, so mdBook owns the root:

- `mdbook build docs` renders the book; because a second renderer (`linkcheck2`) is
  enabled, HTML lands in `docs/book/html/`, which is copied into `site/`. The build
  asserts `site/index.html` exists before upload.
- Rustdoc relocates to `site/rustdoc/` (`mv target/doc/* site/rustdoc/`) with a
  redirect index pointing at the `firewood` crate.
- Go docs build into `site/ffi` (`doc2go`, pinned to `v0.12.2`).
- Benchmark history merges from the `benchmark-data` branch, scoped per-path to
  `bench`/`dev` so the extract cannot clobber the book root.
- The mdBook toolchain installs from pinned sources: `mdbook` and `mdbook-mermaid` via
  `taiki-e/install-action` (pinned by commit SHA), `mdbook-linkcheck2` via
  `cargo binstall`.
- The hardcoded `ref: main` becomes a PR-head-aware checkout, and `docs/**` is added to
  the `pull_request.paths` filter so documentation PRs validate the full assembly.
  Deploy stays gated to non-`pull_request` events on the canonical repository; a `smoke`
  job runs after deploy and `curl --fail`s `/`, `/rustdoc/`, `/rustdoc/firewood/`,
  `/ffi/`, and `/bench/`.
- The architecture diagram moves to `docs/src/assets/`; the workflow's "Copy static
  assets" and hand-written root-redirect steps are removed (mdBook now provides the root
  `index.html` and copies `src/` assets itself).

### How this was tested

- `actionlint` reports no issues on the rewritten workflow.
- The book build and `site/` staging were verified locally on the foundation branch; the
  complete Pages assembly (cargo doc, doc2go, benchmark merge, mdBook build) runs in CI
  on this PR.
- Note: `/bench/` resolves only once the `benchmark-data` branch is populated, which it
  is in the canonical repository.

### Breaking Changes

- **External deep links to `/firewood/<crate>/` rustdoc paths now 404.** Rustdoc moves
  under `/rustdoc/`, with a redirect index at `/rustdoc/`. This is an accepted trade-off
  for the cleaner root URL; per-crate redirect stubs at the old paths remain a documented
  future option.
- No Rust, Go, or `fwdctl` code changes.

> Known limitation (pre-existing, not introduced here): the `build` job's canonical-repo
> guard skips pull requests opened from forks, so documentation PRs from forks do not run
> this workflow. Relaxing this requires a `pull_request_target` redesign with its own
> security review; tracked as a follow-up.


rkuris (approved):
I think this means pretty much every PR now triggers the full pages rebuild. Probably not the long pole anyway though.

PR #2095, opened 2026-06-19
fix: prevent OOM DoS in Box<[T]>::read_v0_item via untrusted num_items

dvansari65:
## Why this should be merged

The existing code reads `num_items` from untrusted input without validation and uses it to drive `(0..num_items).map(...).collect()`. While `.collect()` short-circuits on failure and does not actually allocate `usize::MAX`, this still allows a malicious peer to control loop bounds with an unvalidated value. A malformed proof with an absurd `num_items` causes the node to iterate over a massive range before failing, wasting CPU cycles and complicating error handling.

This PR adds a conservative early check: `num_items` cannot exceed the bytes remaining in the stream, since every non-ZST item must consume at least one byte. This rejects malformed proofs immediately with a clear error instead of relying on mid-loop EOF failure.

## How this works

The fix adds a bounds check before the existing `.collect()`:

```rust
let num_items = reader.read_item::<usize>()?;

## How this was tested

- `test_dos_array_length_bounds`: Corrupts a real proof to claim 10,000,000 items in `start_proof`. Verifies the guard rejects with `InvalidItem { item: "array length", .. }` and the correct error message.
- `rejects_usize_max_items_via_length_guard`: Sends `usize::MAX` as item count. Asserts `InvalidItem { item: "array length", .. }` to prove the new guard caught it, not incidental EOF.
- `rejects_eof_mid_array`: Sends 5 items with only 64 bytes (2 complete Hash32 items). Guard doesn't trigger, but loop fails safely on EOF.
- `rejects_var_len_item_with_huge_claimed_length`: Sends 1 item whose internal length field claims `usize::MAX`. Verifies nested amplification is caught.
- `accepts_empty_array`: Sends 0 items. Verifies empty arrays parse correctly.
- `boundary_exact_fit_passes`: Sends 50 items with exactly 50 bytes (`T = u8`). Verifies `num_items == remainder().len()` passes.
- `boundary_one_over_fails`: Sends 51 items with 50 bytes (`T = u8`). Asserts `InvalidItem { item: "array length", .. }` to verify `num_items > remainder().len()` rejects.


## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### CODE REVIEW

File: firewood/src/proofs/de.rs [hunk 1]
```diff
@@ -70,10 +70,20 @@ impl<T: Version0> Version0 for Box<[T]> {
             .read_item::<usize>()
             .map_err(|err| err.set_item("array length"))?;
 
-        // FIXME(demosdemon): we must somehow validate `num_items` matches what is expected
-        // An incorrect, or unexpectedly large value could lead to DoS via OOM
-        // or panicing
-        (0..num_items).map(|_| reader.read_v0_item()).collect()
+        if num_items > reader.remainder().len() {
```
rkuris: This bounds check is a great way to fix this problem. Each item costs at least 1 byte, so `num_items > remainder().len()` is a sound conservative guard.

firewood/src/proofs/de.rs
```diff
@@ -70,10 +70,20 @@ impl<T: Version0> Version0 for Box<[T]> {
...
+            return Err(reader.invalid_item(
+                "array length",
+                "length less than or equal to the remaining bytes",
+                format!("{} > {}", num_items, reader.remainder().len()),
+            ));
+        }
+
+        let mut items = Vec::new();
```
rkuris: The `Vec::new()` + `push` loop is an unnecessary de-optimization. Once the bounds check above caps `num_items` at the remaining bytes, the original `(0..num_items).map(...).collect()` is safe (it short-circuits and never pre-allocates `usize::MAX`). Please revert to `.collect()`.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+                found,
+                format!("{} > {}", malicious_num_items, remainder_len)
+            );
+        }
+        other => panic!("Expected ReadError::InvalidItem for DoS vector, got: {other:?}"),
+    }
+}
+
+#[cfg(test)]
+mod box_array_deserialization_tests {
```
rkuris, line 962: `#[cfg(test)]` here is redundant since the parent module is already `#[cfg(test)]`.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+        }
+    }
+
+    fn v0_reader(data: &[u8]) -> V0Reader<'_> {
+        let inner = ProofReader::new(data);
+        V0Reader::new(inner, Header::from(ProofType::Range))
+    }
+
+    #[test]
+    fn rejects_usize_max_items_without_oom() {
```
rkuris: This passes against the unpatched code too. `.collect()` short-circuits on the first failing read, so it never OOMs and `is_err()` is already true. Assert the specific `ReadError::InvalidItem { item: "array length", .. }` so it actually pins the new guard.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+        let mut data = Vec::new();
+        data.extend_from_slice(&usize::MAX.encode_var_vec());
+        let mut reader = v0_reader(&data);
+        let result: Result<Box<[Hash32]>, _> = reader.read_v0_item();
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn rejects_when_item_count_exceeds_remaining_bytes() {
```
rkuris: `50 > 50` is false, so this never reaches the new guard. It's exercising the pre-existing per-item EOF path, and duplicates `rejects_when_items_need_more_bytes_than_available` and `rejects_eof_mid_array`. Collapse these three into one.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+        let mut data = Vec::new();
+        data.extend_from_slice(&50usize.encode_var_vec());
+        data.extend_from_slice(&[0u8; 50]);
+        let mut reader = v0_reader(&data);
+        let result: Result<Box<[Hash32]>, _> = reader.read_v0_item();
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn rejects_when_items_need_more_bytes_than_available() {
```
rkuris: Duplicate of the per-item-EOF case above (`10 > 100` is false, so the guard isn't hit).

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+        data.extend_from_slice(&1usize.encode_var_vec());
+        data.extend_from_slice(&usize::MAX.encode_var_vec());
+        data.extend_from_slice(&[0u8; 10]);
+        let mut reader = v0_reader(&data);
+        let result: Result<Box<[VarLenVec]>, _> = reader.read_v0_item();
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn rejects_eof_mid_array() {
```
rkuris: Third copy of the same per-item-EOF path (`5 > 64` is false). Keep one of these three.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+        let mut data = Vec::new();
+        data.extend_from_slice(&0usize.encode_var_vec());
+        let mut reader = v0_reader(&data);
+        let result: Result<Box<[Hash32]>, _> = reader.read_v0_item();
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap().len(), 0);
+    }
+
+    #[test]
+    fn rejects_one_item_with_no_data() {
```
rkuris: Asserting only `is_err()` doesn't distinguish the new guard from an incidental downstream read error. Assert the `InvalidItem { item: "array length", .. }` variant.

firewood/src/proofs/tests.rs
```diff
@@ -916,3 +916,161 @@ fn test_slow_malformed_proof_fuzz() {
...
+        let mut data = Vec::new();
+        data.extend_from_slice(&1usize.encode_var_vec());
+        let mut reader = v0_reader(&data);
+        let result: Result<Box<[Hash32]>, _> = reader.read_v0_item();
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn accepts_large_valid_array() {
```
rkuris: Good positive test, but `1000` vs `32000` has too much slack to catch a `>=` or `*item_size` regression. Add a boundary pair (`num_items == remainder().len()` passes, `+1` fails).

rkuris (changes requested):
Thanks for this! The bounds check at `de.rs:73` is the right fix and the added coverage is welcome. A couple of notes.

`.collect()` into `Result<Box<[T]>, _>` doesn't pre-allocate from `num_items` (its `size_hint` lower bound is 0), so it short-circuits on the first failing read rather than OOMing. The `Vec::with_capacity(usize::MAX)` premise in the description isn't quite right.

I think the other tests will pass with or without this fix. Would be good to call this out in the PR description if they're just increasing coverage. Only `test_dos_array_length_bounds` actually fails against the unpatched code.

**Required for next review:**
- Revert the `Vec::new()`/`push` loop back to `.collect()` (keep the bounds check). The rewrite is an unnecessary de-optimization.
- De-duplicate the three EOF tests (`rejects_when_item_count_exceeds_remaining_bytes`, `rejects_when_items_need_more_bytes_than_available`, `rejects_eof_mid_array`). They exercise the same pre-existing per-item path, so collapse them to one.
- Make the guard-targeting tests assert the specific `ReadError::InvalidItem { item: "array length", .. }` instead of `is_err()`, so they're real regression tests.
- Add the off-by-one boundary tests (`num_items == remainder().len()` passes, `+1` fails) to pin `>` vs `>=`.

**Recommended, but could be skipped:**
- Add a change-proof DoS test (the `Box<[T]>` impl is shared, so this is low-risk but symmetric).
- Drop the redundant inner `#[cfg(test)]` on `box_array_deserialization_tests`.
- Update the PR description to drop the immediate-OOM claim.

firewood/src/proofs/de.rs
demosdemon: As I note in #2099, which this does not resolve, each `ProofNode` is a minimum of 5 bytes on the wire. Let's strengthen this check by defining a `MIN_BYTES_PER_ITEM` const set to 5 and return an error if `num_items > reader.remainder().len() / MIN_BYTES_PER_ITEM`.

dvansari65: ok @demosdemon , looking good

dvansari65: > As I note in #2099, which this does not resolve, each `ProofNode` is a minimum of 5 bytes on the wire. Let's strengthen this check by defining a `MIN_BYTES_PER_ITEM` const set to 5 and return an error if `num_items > reader.remainder().len() / MIN_BYTES_PER_ITEM`.

I implemented MIN_BYTES_PER_ITEM = 5, but it causes valid test suites (like full_range_proof and delete) to fail.

The issue is that impl<T: Version0> Version0 for Box<[T]> is a generic implementation used for multiple different array types, not just ProofNode.

While it's true that the absolute minimum encoded size of a ProofNode is exactly 5 bytes (1 byte key len + 1 byte partial_len + 1 byte value_digest discriminant + 2 bytes ChildMask), Box<[T]> is also used to deserialize:

Box<[BatchOp]>: An empty BatchOp::Delete is only 2 bytes (1 byte discriminant + 1 byte empty key length).
Box<[(Box<[u8]>, Box<[u8]>)]> (key_values): An empty key-value pair is only 2 bytes (1 byte empty key length + 1 byte empty value length).
By hardcoding / 5 in the generic Box<[T]> block, we accidentally break perfectly valid range proofs and change proofs that contain many small BatchOps or key_values. For example, in full_range_proof, it tries to parse an array of 256 key_values from a 1020-byte remainder. 256 items easily fit in 1020 bytes (since they only need 512 bytes minimum), but the / 5 check evaluates 256 > 1020 / 5 -> 256 > 204 and rejects it!

dvansari65:
@demosdemon  added const MIN_BYTES_PER_ITEM: usize as an associated constant on the Version0 trait itself, allowing ProofNode to specify 5 and BatchOp to specify 2

firewood/src/proofs/de.rs
```diff
@@ -146,6 +167,8 @@ impl Version0 for FrozenChangeProof {
 }
 
 impl Version0 for BatchOp<Key, Value> {
+    const MIN_BYTES_PER_ITEM: usize = 2;
```
rkuris: Let's use a `std::num::NonZeroUsize` to avoid the calls to `checked_div`. You can just divide without warning then, since a `NonZeroUsize` can't be zero.

firewood/src/proofs/de.rs
```diff
@@ -70,9 +70,30 @@ impl<T: Version0> Version0 for Box<[T]> {
...
+        let max_items = reader
+            .remainder()
+            .len()
+            .checked_div(T::MIN_BYTES_PER_ITEM)
+            .ok_or_else(|| {
+                reader.invalid_item(
+                    "array length",
+                    "valid minimum bytes per item",
+                    "MIN_BYTES_PER_ITEM is zero",
+                )
```
rkuris: This code is not needed if we use a non zero type. See below for specifics.

rkuris (changes requested):

demosdemon (approved):

rkuris (approved):

PR #2096, opened 2026-06-19
docs: explain how to sign commits in CONTRIBUTING

rkuris:
## Why this should be merged

Outside contributors often don't sign their commits. This documentation explains to them how to do that.

## How this works

RTFM

## How this was tested

md lints cleanly

## Breaking Changes

None

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,51 +42,18 @@ so you should expect a response during the week within 24 hours.
 
 ## Signing your commits
 
-Every commit in a PR must be cryptographically signed so that GitHub shows it
-as **Verified**. CI rejects PRs that contain unsigned commits, so set this up
-before you open one.
+CI rejects PRs that contain unsigned commits, so configure commit signing
+before you open one — GitHub should then show every commit as **Verified**.
 
-You can sign with either a GPG key or an SSH key. SSH is the quickest path if
-you already push over SSH.
-
-### Option A: Sign with your existing SSH key
+The quickest setup is signing with the SSH key you already push with:
 
     git config --global gpg.format ssh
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -39,6 +40,56 @@ Also, if you update any versions of packages, notably the MSRV (Minimum Supporte
...
+    git log --show-signature -1
+
+If you already have unsigned commits on your branch, re-sign them and
+force-push:
+
+    git rebase --exec 'git commit --amend --no-edit -S' -i <base>
+    git push --force-with-lease
+
+GitHub's [signing your commits][gh-signing] guide has more detail, including
+key setup on Windows and macOS.
```
RodrigoVillar: I think it would be sufficient to point contributors to Github's documentation on how to sign commits rather have an entire section in `CONTRIBUTING.md` on how to sign commits. IMO, we shouldn't duplicate documentation and instead point contributors to the source-of-truth for information like this (in this case, point contributors to GH).

RodrigoVillar (changes requested):

CONTRIBUTING.md
demosdemon: Agreed. And using GPG isn't even recommended anymore unless you have a reason.

rkuris: Fixed in 9ade337c. Good point, trimmed the section down to the one firewood-specific fact (CI rejects unsigned commits), a minimal SSH snippet, and a link to GitHub's signing guide for everything else. Also led with SSH per @demosdemon's note that GPG isn't the default recommendation anymore.

demosdemon (approved):

RodrigoVillar (approved):

PR #2101, opened 2026-06-22
feat(storage): add HashMode trait and route ethhash gates through it [2/9]

RodrigoVillar:
## Why this should be merged

Firewood needs a common vocabulary for the behavior that differs between the Ethereum and MerkleDB hashing schemes before the selected scheme can move from a compile-time feature to a per-database choice.

In the stack, this introduces the abstraction that the remaining PRs build on without changing how databases are selected or represented yet. PR 3 moves the mode-specific codecs and hashing behind this trait, and PR 6 eventually
threads the trait through the typed database stack.

## How this works

This change introduces the `HashMode` trait and the zero-sized `EthHash` and `MerkleDbHash` implementations. The trait initially owns the selected `NodeHashAlgorithm`, empty-root behavior, and valid-key policy. A temporary `DefaultHashMode` alias preserves the existing `ethhash` feature behavior while current feature-gated decisions are routed through the new abstraction. It does not yet thread a hash-mode type through the database stack or compile both implementations into normal execution paths.

## How this was tested

Added tests covering the compile-feature-to-`DefaultHashMode` bridge, mode-specific empty roots, and the valid-key rules for both hashing schemes.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


demosdemon (approved):

demosdemon (approved):
diff is unchanged since my last review

PR #2102, opened 2026-06-22
refactor(storage): unconditional HashType enum + per-mode codecs [3/9]

RodrigoVillar:
## Why this should be merged

Runtime selection is not possible while core node and proof types change shape at compile time. Both hashing schemes need to share one in-memory data model while continuing to read and write their existing, different byte encodings.

In the stack, this turns the PR 2 abstraction into a common low-level data and codec layer. That shared representation is required before PR 5 can parse both proof formats in one binary and before PR 6 can instantiate the database stack
with either hash mode.

## How this works

This change makes `HashType` an unconditional enum that can contain either a 32-byte hash or inline RLP, and makes the hashed arm of `ValueDigest` unconditional. It expands `HashMode` with mode-specific hashing, preimage, and child-codec behavior so Ethereum retains its tagged hash/inline-RLP encoding while MerkleDB retains its bare 32-byte hash encoding.

Proof serialization also dispatches through the selected mode without changing either wire format. A defensive reconciliation guard rejects hashed value digests when the effective mode is Ethereum before trie mutation or callbacks can observe invalid state.

## How this was tested

Added codec round-trip tests for Ethereum hashes, Ethereum inline RLP, and MerkleDB hashes; rejection tests for inline RLP in MerkleDB mode; cross-mode leaf-preimage tests; an exact Ethereum child-hash serialization golden test;
and a regression test for the defensive Ethereum value-digest guard.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/firewood/src/merkle/reconcile.rs
+++ b/firewood/src/merkle/reconcile.rs
@@ -60,11 +60,12 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
 
         let proof_value = match proof_node.value_digest.as_ref() {
             Some(ValueDigest::Value(v)) => Some(v.as_ref()),
-            #[cfg(not(feature = "ethhash"))]
             Some(digest @ ValueDigest::Hash(_)) => {
                 // In merkledb mode, large values (>= 32 bytes) are stored as
                 // hashes in serialized proofs. If the branch's value hashes to
                 // the same value, the proof and trie agree — no conflict.
+                // (Dead but harmless under the Ethereum scheme, which never
+                // emits a `Hash` digest.)
                 if matches!(branch.value.as_deref(), Some(v) if digest.as_ref().verify(v)) {
                     return Ok(());
                 }
--- a/firewood/src/merkle/reconcile.rs
+++ b/firewood/src/merkle/reconcile.rs
@@ -51,6 +51,12 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
             return Err(ProofError::ValueAtOddNibbleLength);
         }
 
+        if DefaultHashMode::ALGORITHM.is_ethereum()
+            && matches!(proof_node.value_digest, Some(ValueDigest::Hash(_)))
+        {
+            return Err(ProofError::UnexpectedValueDigest);
+        }
+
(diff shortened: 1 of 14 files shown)


firewood/src/merkle/reconcile.rs
demosdemon, line 88: your added comment here makes me want to add a defensive guard and return an error if we get an unexpected ValueDigest instead of doing the wrong thing.

demosdemon: Ah, I know what I want to do. HashType is already a trait with the algo as an associated constant. We can add an associated type for ValueDigest implementing a new trait and have a distinct ValueDigest type for each hash mode. Only MerkleDB needs the enum whereas Ethereum can be simply represented by a Value.

RodrigoVillar: Hmm good callout - rather than adding an associated type here, I instead added a defensive guard in `reconcile_branch_proof_node()` and `value_digest()` that returns `ProofError::UnexpectedValueDigest` if the value digest is of variant `Hash` and we're in the Ethereum hashing scheme. 

https://github.com/ava-labs/firewood/blob/173716f0f0e0071acfe15e93586b3cff2747dcb2/firewood/src/merkle/reconcile.rs#L54-L58

https://github.com/ava-labs/firewood/blob/173716f0f0e0071acfe15e93586b3cff2747dcb2/firewood/src/proofs/types.rs#L578-L584

I tried out adding the associated type as you suggested above, but found that adding the runtime guards was simpler than adding a generic here and having to propagate it upwards (I also added `test_reconcile_branch_proof_node_rejects_hashed_value_digest()` and `eth_hashed_value_digest_is_rejected()` to test the runtime guard). Happy to go with the associated-type approach though if you prefer.

demosdemon: An assoc type would be invasive. So, I'm inclined to wait until the full stack is finished to make that switch after measuring the performance implications.

demosdemon (approved):

RodrigoVillar:
Rebased to handle merge conflicts related to the `DenseChildren` PRs which got merged in.

PR #2103, opened 2026-06-24
test(proofs): expand snapshot tests to cover header, key-values, and batch ops

demosdemon:
Extends the insta snapshot test suite started in #2100 to lock
down the remaining V0 wire-format sections that were not yet snapshot-tested.

## Why this should be merged

The earlier PR pinned ProofNode encoding. The proof header, key-value pair
encoding, and BatchOp encoding were left without snapshot coverage, meaning an
accidental format change in those paths could go undetected until a round-trip
test caught the mismatch at runtime.

## How this works

Adds two helper functions (`range_proof_bytes`, `change_proof_bytes`) that
serialize minimal proofs with empty node lists, then adds three new test
modules:

- `header` — snapshots the full 32-byte header for Range and Change proof
  types; cfg-gated into `merkledb`/`ethhash` submodules because the
  `hash_mode` byte differs between SHA-256 and Keccak-256 modes.
- `key_values` — snapshots `bytes[32..]` for range proofs carrying 0, 1, and
  2 key-value pairs; feature-independent, no cfg-gating.
- `batch_ops` — snapshots `bytes[32..]` for change proofs carrying a single
  `Put`, `Delete`, `DeleteRange`, and all three in sequence; also
  feature-independent.

Updates the `justfile` recipe comment to reflect the broader coverage.

## How this was tested

`just snapshot-proof-nodes` writes all 11 new snapshot files. `cargo nextest
run -p firewood --features logger --all-targets` passes 415 tests.

### DIFF
--- a/firewood/src/proofs/snapshot_tests.rs
+++ b/firewood/src/proofs/snapshot_tests.rs
@@ -36,6 +36,27 @@
 //! `single_child` and `all_children` already cover the complete child
 //! encoding format. The large-value → hash behaviour is format-agnostic and
 //! is fully covered by the `merkledb` variant.
+//!
+//! ## Additional coverage
+//!
+//! The following tests lock down wire-format sections beyond individual node encoding.
+//! The payload snapshots (everything after the 32-byte header) are feature-independent
+//! for key-values and batch ops; the header snapshots are cfg-gated because the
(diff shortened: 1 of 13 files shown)


demosdemon:
merging with #2104

PR #2104, opened 2026-06-24
test: add insta snapshot tests for proof and node wire serialization

demosdemon:
Establishes binary-format snapshot tests using `insta` for both the
proof serialization layer (`firewood`) and the node serialization layer
(`firewood-storage`). These tests lock down wire-format bytes so that
any accidental encoding change is caught as a snapshot mismatch rather
than surfacing only as a round-trip equality failure.

Extracting these tests into their own PR also unblocks #2100
(`dense-children-array`): that PR can show format-change diffs
explicitly when it lands on top of this one.

## Why this should be merged

Without snapshot tests, wire-format regressions are silent: a varint
encoding tweak, a discriminant byte flip, or a field reordering only
fails if the mutated bytes happen to produce an invalid round-trip.
Snapshot tests detect changes at the byte level and make intentional
format changes visible in the diff.

## How this works

Each suite groups structurally-identical cases into `test_case`
parametrized functions. Two constraints shape the design:

- **Explicit snapshot names are required.** `insta` keys an auto-named
  snapshot on the enclosing function, so every case of one parametrized
  function would resolve to the same name and collide. Each case passes
  an explicit name as its first argument.
- **Hash-mode splits are encoded in the name, not the module tree.**
  Tests whose bytes differ between SHA-256 and Keccak-256 build in both
  modes from a single function; a `hash_mode_name()` helper prefixes the
  snapshot name with `merkledb__` / `ethhash__` so each mode keeps its
  own file. `cfg_attr` is reserved for the one case that exists in a
  single mode. Snapshot names are also keyed by module path (not
  function), so feature-agnostic groups carry a `key_values__` /
  `batch_ops__` prefix (applied once in the function body) to stay
  unique within the flattened module.

### `firewood` — proof wire format (`firewood/src/proofs/snapshot_tests.rs`)

19 snapshot files covering `ProofNode` encoding, the 32-byte proof
header, key-value pair encoding, and `BatchOp` encoding:

- `proof_node` (3): `empty`, `leaf_with_value`, `partial_path` —
  no children, so hash-mode-independent (one shared file each).
- `branch_node` (3 merkledb / 2 ethhash): `single_child`,
  `all_children` in both modes (prefixed), plus a merkledb-only
  `large_value_becomes_hash` via `cfg_attr`.
- `proof_header` (2): 32-byte header for range and change proof types,
  per hash mode (prefixed).
- `key_value` (3): empty sequence, single pair, multiple pairs.
- `batch_op` (4): `Put`, `Delete`, `DeleteRange`, and all three in
  sequence.

### `firewood-storage` — node wire format (`storage/src/node/snapshot_tests.rs`)

13 snapshot files covering `Serializable` trait implementations and
`Node::as_bytes` / `Node::from_reader`:

- `trie_hash` (2): zero hash, sequential-byte hash.
- `free_area` (3): `None` next pointer, single-byte varint (42),
  multi-byte varint (1024).
- `leaf` (2): `with_value`, `long_partial_path` —
  hash-mode-independent.
- `branch` (2): one child and value+child, in both modes (prefixed) —
  flat 32-byte SHA-256 child hashes vs Keccak-256 discriminant encoding.
- `hash_or_rlp` (2, ethhash-only): `HashOrRlp::Hash` and
  `HashOrRlp::Rlp` directly.

### Infrastructure

- `insta` promoted to `[workspace.dependencies]` rather than declared
  separately in each crate; suites are parametrized with the existing
  `test-case` workspace dependency.
- `snapshot-all` `just` recipe regenerates all suites in one command.
- `.github/check-license-headers.yaml` excludes `**/*.snap` files from
  the license header check.

## How this was tested

```
just snapshot-all
cargo nextest run --workspace --features logger --all-targets
cargo clippy --workspace --features logger --all-targets
cargo fmt --check
```

Snapshot suites pass in both hash modes (storage 9/11, proofs 15/14);
the full workspace suite passes with no clippy warnings.


firewood/src/proofs/snapshot_tests.rs
```diff
@@ -0,0 +1,374 @@
...
+fn empty() {
+    let node = make_node(&[], 0, None, &[]);
+    insta::assert_snapshot!(hex::encode(node_bytes(&node)));
+}
+
+/// Leaf with a short value (`b"hello"`, 5 bytes). Values shorter than 32 bytes
+/// are encoded as `ValueDigest::Value` (discriminant `0x00`) under both hash modes.
+#[test]
+fn leaf_with_value() {
+    let node = make_node(&[1, 2, 3], 0, Some(Box::from(b"hello".as_slice())), &[]);
```
rkuris: Use`#[test_case]` here since we're doing the exact same tests with different data (this is case 2 of at least 3)

firewood/src/proofs/snapshot_tests.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        /// Change proof header: same as range but `proof_type` `0x02` (change).
+        #[test]
+        fn change() {
+            insta::assert_snapshot!(hex::encode(change_header()));
+        }
+    }
+
+    #[cfg(feature = "ethhash")]
+    mod ethhash {
```
rkuris: this block looks identical to the prior `mod merkledb`. Is this for future-proofing when we make ethhash runtime?

rkuris (approved):

firewood/src/proofs/snapshot_tests.rs
demosdemon: `insta` creates the snapshot based on the module path. The separate module here is needed to make the file name distinct.

demosdemon: actually, this goes away after refactoring to use test_case.

Issue #804, opened 2025-02-24
Don't serialize or deserialize nodes


rkuris:
To read data, we currently read nodes (starting at the root) and for each node we construct a BranchNode object. Then we only read one child address out of that. It would be more efficient to read the node serialized and have a method that figures out from the serialized object what the child address is at a given offset.

One issue with this today is that the node is compacted very tightly. It might make sense to relax this compaction some so that we don't have to pay the cost of serializing. For example, the branch nodes could have the list of children as an array at the end rather than packing empty items together.

It might also be good for the actual data for the node to be in a [Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html) which could facilitate using memory mapped files for nodes stored on disk and a heap arena for nodes stored in memory.

demosdemon:
Suppose we stored node metadata and node values (and perhaps even node paths) as separate objects in the object store. We can then define a fixed allocation for all nodes. This does add a separate read in order to fetch the overflowing path or actual value; but that might be a good thing as we will no longer be reading and deserializing the value unnecessarily. We could then keep track of duplicate values by their hash and then reuse the storage for values with duplicate hashes. Because nodes will have a fixed size, we can quickly index to the specific 8 bytes for the address we are looking for. With branch factor of 16, it does not appear worth while to handle a dynamic vector for pointers to children but if we increase the branch factor then the extra overhead would be worthwhile.

```rust
const BRANCH_FACTORS: [u32; 8] = [
    1 << 1,
    1 << 2,
    1 << 3,
    1 << 4,
    1 << 5,
    1 << 6,
    1 << 7,
    1 << 8,
];

fn main() {
    for branch_factor in BRANCH_FACTORS {
        let bitmap_size = branch_factor.div_ceil(8);
        let array_size = branch_factor * 8;
        let total_size = bitmap_size.min(8) // pad to alignment
            + 8 // for inline path (<=63 bits)
                // or overflow to 63-bit pointer
                // which is sufficient for the majority of platforms
                // as only 48-bits are typically used for addressing
            + 8 // for value pointer
            + array_size // for children pointers
            + 32 // for branch hash
            ;

        println!("branch factor = {branch_factor}");
        println!("bitmap size   = {bitmap_size}");
        println!("array size    = {array_size}");
        println!("branch total  = {total_size}");

        println!();
    }
}
```

branch factor = 2
bitmap size   = 1
array size    = 16
branch total  = 65

branch factor = 4
bitmap size   = 1
[...]

PR #2107, opened 2026-06-26
docs: add v2 on-disk file format proposed design

demosdemon:
Add the proposed design for a version 2 (v2) Firewood on-disk file format at
`docs/src/designs/proposed/0002-v2-file-format.md`, tracking issue #804.

## Why this should be merged

The v1 format serializes each node with a variable-length encoding parsed
field-by-field on every read and bakes the area-size table into the binary (a
mismatch is rejected at open). Path traversal is Firewood's dominant operation, and
a production state file is already in the 2 TiB range, where a format migration is
extremely expensive. v2 removes per-node deserialization from the hot path and is
future-proofed to evolve within major version 2 without forcing another migration.

## How this works

v2 is a clean break from v1 (no v1 backward compatibility) designed for
zero-deserialization reads — `pread` into an over-aligned buffer in v2.0 (one copy,
then valid `bytemuck::from_bytes` casts), true zero-copy via `mmap` in v2.1 (the read
copy elided). A database is a small **control file** (the double-buffered header plus
a **manifest**) and one or more **data files** of areas; a node's address is its byte
offset in a single global data address space, and the manifest maps that space to
physical data files. The format is frozen for all of major version 2; within major
version 2 any reader reads any file in both directions with no byte-format change,
deferring the next migration as far as possible.

- **Hot/cold node split.** Hot, contiguous child addresses; cold child hashes and
  values, faulted only for hashing, proofs, or value reads.
- **Self-describing areas.** 8-byte-aligned, size-classed areas indexing a
  header-stored size-class table; an XXH3-64-checksummed, double-buffered header with
  a header-recorded page size and page-aligned slots.
- **Live manifest indirection.** The read path resolves every address through the
  manifest to a `(data file, intra-file offset)` — the identity map with one data
  file in v2.0. Paying that lookup now keeps the address file-agnostic and makes
  multi-file data a reader-transparent, write-gated addition later, not a v3 break.
- **Crash safety.** A two-`fsync` commit (data files in Phase 1, control file in
  Phase 2) resting on one recoverability invariant; recovery probes candidate header
  strides for a checksum-valid slot, opens the data files the manifest lists, runs a
  total rebuild-scan predicate after an unclean open, and returns reserved areas to
  the free lists on clean close. Crash leaks are bounded; no redo/undo log. v2 is
  durable-on-commit (machine-crash safe), a deliberate strengthening over v1's
  OS-level-only recovery.
- **Within-major compatibility on two axes.** A byte-format axis (bumping it is a v3
  migration) and a reader-capability axis (`min_reader_version`; bumping it makes old
  readers refuse but migrates nothing), over a frozen hot core with
  `required_writer_features`, a feature-bit, and a framed cold-TLV channel. The
  `AreaHeader.kind` vocabulary is an extensible registry, so a new on-disk object
  category (e.g. the durable-revision-history `fdl` log) lands behind the write gate
  without a v3 migration.
- **Runtime hash-mode selection.** merkledb (SHA-256) vs ethhash (Keccak-256)
  selected from the header field at open, over a frozen 32-byte hash geometry.

The document embeds pre-implementation merge gates split into freeze gates (which
block locking any byte layout) and implementation gates. Durable revision history and
multi-file data (split + freeze + whole-file reclaim + copy-on-read) are deferred, but
their on-disk hooks — reserved header fields and the live manifest — ship in v2.0 so
each lands as a v2.x addition.

## How this was tested

Documentation-only change; `markdownlint-cli2` passes. The design itself defines the
round-trip, crash-consistency, differential-vs-v1 consensus, and zero-copy gates that
block the implementation.

## Breaking changes

None in this PR (design document only). The proposed format is a clean break from v1
with no backward compatibility; the implementation lands separately behind the gates
the document defines.

### DIFF
--- a/.devcontainer/features/firewood-tools/install.sh
+++ b/.devcontainer/features/firewood-tools/install.sh
@@ -38,8 +38,10 @@ rustup toolchain install nightly --profile minimal
 rustup component add clippy rustfmt rust-src rust-docs llvm-tools --toolchain nightly
 
 # 3. cargo-binstall (prebuilt) so the remaining tools install without compiling.
+#    Pin the installer to a release commit (v1.21.0) instead of the floating `main`
+#    branch, matching the repo convention of pinning third-party sources by commit SHA.
 curl -L --proto '=https' --tlsv1.2 -sSf \
-  https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+  https://raw.githubusercontent.com/cargo-bins/cargo-binstall/ead08b90bd7b2e6d81963fb9cf0b7239f66d5db4/install-from-binstall-release.sh | bash
 
 # 4. sccache, then wire it in as the global rustc wrapper.
 cargo binstall --no-confirm --locked sccache
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,32 @@ The `fwdctl` tool provides command-line operations on databases. See `fwdctl/REA
 
 For more information on coding conventions and constraints, please refer to [CONTRIBUTING.md](./CONTRIBUTING.md)
 
+## Design Documents
+
+Firewood records designs as living documentation in a single flat directory,
+`docs/src/designs/` (rendered in the mdBook site). Every design is
+`YYYY-MM-DD-<slug>.md`, where the date is the day the design was written; it is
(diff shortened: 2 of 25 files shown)


rkuris:
Skimmed the design doc. I think we should attempt this on a separate dev branch and measure the performance gains (if any) before deciding whether this is worth doing. WDYT?

PR #2111, opened 2026-06-30
refactor(ffi)!: remove direct storage dependency

### LINKED ISSUE #1972: refactor(ffi): reduce direct dependencies on firewood_storage

The `firewood-ffi` crate reaches into `firewood_storage` directly for several behaviors and types. The FFI should depend on `firewood` as the API surface — types as parameters are acceptable, but direct function/macro calls into `firewood_storage` are a layering leak that makes the storage crate part of FFI's contract.

Current direct calls into `firewood_storage` from FFI:

1. **`ffi/src/metrics.rs:51`** — `firewood_storage::registry::register()`
   `setup_metrics` concatenates histogram configs from three layers: `crate::registry`, `firewood::registry`, and `firewood_storage::registry`. The two crate-level `register()` functions aren't duplicates — `firewood::registry` defines higher-level metrics (proposals, commits, revisions, change-proof iterations) while `firewood_storage::registry` defines storage metrics (free-list bytes, node reads/writes, IO uring counters, cache sizes, flush/reap durations).

2. **`ffi/src/jemalloc_metrics.rs:11, 69, 77`** — `use firewood_storage::logger;` and `logger::warn!`. Note that `firewood::logger` is already re-exported (`firewood/src/lib.rs:177`).

Type usages from `firewood_storage` in FFI:

| Type | Usage | Notes |
|---|---|---|
| `TrieHash` | `From<Option<TrieHash>> for HashResult` (`value/results.rs`); `TrieHash::try_from`/`from` constructors (`proofs/change.rs`) | Not re-exported from `firewood` |
| `NodeHashAlgorithm` | FFI enum → storage enum `From` impl (`handle.rs:33-37`) | Not re-exported from `firewood` |
| `NodeStore<Committed, FileBacked>` (with `Committed`, `FileBacked`) | `&NodeStore` parameter to `Db::reconstruct_from_view`; `Arc<NodeStore<...>>` stored in `RevisionHandle.historical` (`handle.rs:14, 252`; `revision.rs:10, 16, 25`) | Real abstraction leak — names storage internals in `firewood`'s own public API |

- Re-export `TrieHash` and `NodeHashAlgorithm` from `firewood`.
- Switch the FFI's `logger` import to `firewood::logger`.
- Roll storage metrics into `firewood::registry::register()` (or add `firewood::registry::register_all()`) so the FFI calls a single entry point.
- Refactor `Db::reconstruct_from_view` so it doesn't expose `NodeStore<Committed, FileBacked>` in its public signature — accept an opaque handle (e.g. `&dyn ReconstructSource` or reuse `ArcDynDbView`).
- Once the above lands, drop `firewood-storage` from `firewood-ffi`'s `Cargo.toml`.

- Re-exporting `NodeStore`/`Committed`/`FileBacked` from `firewood` without changing `reconstruct_from_view`. This removes the FFI's direct `firewood_storage` dep but doesn't fix the underlying leak in the `firewood` API surface — those internals would just be re-exported instead of imported. Acceptable as a stopgap if the API refactor is deferred.
- Leaving the metrics split as-is. The current shape works, but it requires every embedder (not just our FFI) to know the layered structure and call multiple `register()` functions.

**Additional context**

The `NodeStore<Committed, FileBacked>` case is the only one where the `firewood` crate's *own* public API leaks storage types. The other items are FFI-side cleanups that don't require API changes.

RodrigoVillar:
## Why this should be merged

Closes #1972.

`firewood-ffi` currently reaches directly into `firewood-storage` for metrics registration, logging, hash types, code-hash extraction details, and reconstruction parent types. That makes storage internals part of the FFI crate's contract and means storage-layer refactors can become FFI-breaking. This PR routes those dependencies through the `firewood` crate API instead.

## How this works

`firewood` now re-exports the public hash algorithm/hash types that FFI needs and exposes `registry::register_all()` as the single metrics registration entry point for embedders that want full Firewood database metrics, including storage metrics. FFI metrics setup now calls that firewood-level entry point, and jemalloc metrics logging uses `firewood::logger`.

Reconstruction now uses a firewood-owned boundary type. `Db::committed_view()` returns an opaque `CommittedView` for committed revisions, and `Db::reconstruct_from_view()` accepts `&CommittedView` instead of exposing `NodeStore<Committed, FileBacked>` and storage reconstruction traits in the public signature. The FFI `RevisionHandle` stores this opaque committed parent only when the requested root is committed; proposal roots remain readable via `Db::view()` but are not reconstructable.

Ethereum code-hash extraction moved behind `firewood::account_code_hash()`. The FFI iterator preserves its previous behavior for wrong-length `codeHash` fields by skipping that account and continuing to later entries, while malformed account RLP still surfaces as an iterator error. The account-format error now carries the underlying RLP decoding source.

## How this was tested

CI + added targeted tests added in this PR:
- `account_code_hash_decodes_all_outcomes`
- `register_all_includes_storage_metrics`
- `test_reconstruct_from_committed_view_uses_requested_revision`
- `malformed_account_value_is_returned_as_error`
- `wrong_length_code_hash_is_skipped`

## Breaking Changes

- [x] firewood
  - `Db::reconstruct_from_view` now accepts `&CommittedView` instead of a generic storage `NodeStore` source.
  - Malformed Ethereum account values now surface as `ProofError::InvalidAccountValueFormat` with the underlying RLP source instead of the removed `InvalidValueFormat` variant.
- [ ] firewood-storage
- [x] firewood-ffi (C api)
  - No C ABI change is intended, but malformed Ethereum account values returned through the code-hash iterator now include the more specific `invalid Ethereum account value format: ...` error text.
- [ ] firewood-go (Go api)
- [ ] fwdctl


firewood/src/eth_proof.rs
demosdemon, line 37: let's include the `c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470` comment that was removed from the other const to make it easier to [grep](https://github.com/search?q=c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470+org%3Aava-labs&type=code)

demosdemon (approved):

firewood/src/eth_proof.rs
RodrigoVillar: Good call - done here: 8a0617451e37aadf83f078879d046bbe1e60cfe1.

PR #2112, opened 2026-06-30
docs(storage): document FileBacked struct and its fields

RodrigoVillar:
## Why this should be merged

While looking at the caching policy for trie nodes, I noticed that the `FileBacked` type was lacking documentation - given that important Firewood operations funnel down to this type, I figured we should document it better.

## How this works

Adds documentation to the `FileBacked` type and its fields.

## How this was tested

CI.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/storage/src/linear/filebacked.rs
+++ b/storage/src/linear/filebacked.rs
@@ -35,17 +35,56 @@ use crate::{CacheReadStrategy, CachedNode, LinearAddress, MaybePersistedNode, Sh
 
 use super::{FileIoError, OffsetReader, ReadableStorage, WritableStorage};
 
-/// A [`ReadableStorage`] and [`WritableStorage`] backed by a file
+/// A [`ReadableStorage`] and [`WritableStorage`] backed by a single on-disk file.
+///
+/// This is the persistent storage backend for a node store. It owns the open
+/// file handle — which can be advisory-locked via [`Self::lock`] to stop another
+/// process from opening the same database — along with two in-memory LRU caches
+/// that sit in front of it: one for trie nodes and one for free-list links.
+///
+/// Reads and writes go through positioned (`pread`/`pwrite`) syscalls, so they
+/// share no file cursor and take no per-handle lock; the caches absorb most
+/// reads so that the hot path rarely touches disk. The only serialization
+/// points are the two cache mutexes and, on the `io-uring` path, the ring's
+/// internal lock.
 #[derive(Debug)]
 pub struct FileBacked {
+    /// Path of the on-disk database file.
     filename: PathBuf,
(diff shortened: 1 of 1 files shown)


storage/src/linear/filebacked.rs
```diff
@@ -35,17 +35,56 @@ use crate::{CacheReadStrategy, CachedNode, LinearAddress, MaybePersistedNode, Sh
...
+/// This is the persistent storage backend for a node store. It owns the open
+/// file handle — which can be advisory-locked via [`Self::lock`] to stop another
+/// process from opening the same database — along with two in-memory LRU caches
+/// that sit in front of it: one for trie nodes and one for free-list links.
+///
+/// Reads and writes go through positioned (`pread`/`pwrite`) syscalls, so they
+/// share no file cursor and take no per-handle lock; the caches absorb most
+/// reads so that the hot path rarely touches disk. The only serialization
+/// points are the two cache mutexes and, on the `io-uring` path, the ring's
+/// internal lock.
```
rkuris, line 49: I'm waffling between "this is too much documentation about how FileBacked works which belongs in the methods that do the work" and "this is a good overview". I think it's fine.

storage/src/linear/filebacked.rs
```diff
@@ -35,17 +35,56 @@ use crate::{CacheReadStrategy, CachedNode, LinearAddress, MaybePersistedNode, Sh
...
+    /// avoiding a disk read to walk the free list. A lookup is destructive: a
+    /// cache hit pops the entry, since allocating from the free list consumes it.
     free_list_cache: Mutex<EntryLruCache<LinearAddress, Option<LinearAddress>>>,
+    /// Policy controlling which reads get inserted into [`Self::cache`]
+    /// (no reads, branch reads only, or all reads). Writes are always cached
+    /// regardless of this setting.
     cache_read_strategy: CacheReadStrategy,
+    /// The node hashing algorithm (MerkleDB or Ethereum) this file was opened
+    /// with. Reported via [`ReadableStorage::node_hash_algorithm`] so callers
+    /// can detect a mismatch between the file and the running build.
```
rkuris, line 75: This appears incorrect. I don't think we use this field to check for mismatches. We set this based on what was in the header of the DB.

rkuris (approved):

storage/src/linear/filebacked.rs
RodrigoVillar: Created a follow-up here: https://github.com/ava-labs/firewood/issues/2114.

PR #2115, opened 2026-07-06
ci: run clippy on a pinned nightly, gate maxperf, and add a weekly nightly lint

demosdemon:
## Why this should be merged

Clippy gated on the floating `@stable` toolchain, so a Rust/clippy release could change the enabled-lint surface and break CI unrelated to any code change. This pins the clippy jobs to a fixed nightly for a deterministic surface, adds a weekly floating-`nightly` run to catch drift out of band, and gates the `maxperf` profile — built and tested in CI but never linted (its release settings turn off `debug-assertions`/`overflow-checks`). Groundwork for adopting stricter clippy lint groups (nursery, then `restriction`) next.

## How this works

Only the clippy jobs move to the pinned nightly; every other job stays on `stable`. See the commit history for specifics.

## How this was tested

`actionlint` clean; `cargo +nightly-2026-07-05 clippy … -D warnings` clean across the host-runnable configs (`--all-features`/io-uring is Linux-only → CI); `cargo fmt --check` clean; `cargo nextest` 810/810.

## Operational notes

- New status check `rust-lint (maxperf-ethhash-logger)` — add it to the branch-protection required checks.
- `lint-nightly` is `schedule`/`workflow_dispatch`-only (no PR run) — `workflow_dispatch` it once after merge.
- Local clippy now needs `rustup toolchain install nightly-2026-07-05 --component clippy`.

## Breaking Changes

None — CI/tooling only.

### DIFF
--- a/.github/workflows/attach-static-libs.yaml
+++ b/.github/workflows/attach-static-libs.yaml
@@ -66,8 +66,10 @@ jobs:
             pre-build-cmd: echo "MACOSX_DEPLOYMENT_TARGET=15.0" >> "$GITHUB_ENV"
     steps:
       - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # @stable
-      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # @v2 -> v2.9.1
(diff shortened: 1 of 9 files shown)


rkuris (approved):

rkuris:
There's new skew in CONTRIBUTING.md and .github/copilot-instructions.md. They still say plain "cargo clippy". This is really a separate issue that is pre-existing and could use a new issue for cleanup if you agree.

PR #2116, opened 2026-07-06
docs: document the lint-suppression policy

demosdemon:
## Why this should be merged

The rules for *when* and *how* to suppress a clippy/rustc lint have lived only in reviewers' heads. This writes them down as an explicit policy so the follow-up cleanup of existing suppressions has a documented standard — and it corrects `clippy.toml`, whose `disallowed-methods` guidance recommended `#[allow]` rather than the project's preferred reasoned `#[expect]`.

## How this works

Documentation and lint-neutral config only — no source changes and no change to which lints fire. Stacked on #2115; the machinery that mechanically enforces "no local `#[allow]`" lands in a later PR.

## How this was tested

markdownlint clean. Lint-neutral: the `clippy.toml` `reason` field is diagnostic-only and `CODE_REVIEW.md` isn't compiled, so `cargo clippy` behavior is unchanged.

## Breaking Changes

none.


CODE_REVIEW.md
```diff
@@ -5,20 +5,83 @@ providing inline feedback — in addition to general best practices.
...
+  `#![expect(unsafe_code, ...)]` as a documented FFI-boundary carve-out.
+
+### `unsafe_code` exemption
+
+`#[expect(unsafe_code)]` needs **no** `reason`. For a per-block `unsafe { ... }` the
+`// SAFETY:` comment (required by `undocumented_unsafe_blocks`) is the justification; for
+`#[unsafe(no_mangle)]`/`extern "C"` functions and `unsafe impl` bridges the item's own
+doc-comment is. An existing reason is fine but not required.
+
+### Per-lint guidance
```
rkuris, line 68: Let's move this into Cargo.toml instead of here. Example:
```rust
indexing_slicing = { level = "warn", reason = "Use checked operations like .get() or .get_mut() instead of direct indexing to prevent panics" }
```

These explanations are fine for AI but they're still too verbose for manual work. If you disagree, I think it's possible to add a "see <link>" to the reason field for the longer version.

rkuris (commented):
First pass review, would like to see us change the warning itself rather than burying the reasoning in a doc file.

CODE_REVIEW.md
demosdemon: This isn't a thing afaict <https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section>

The docs only mention `level` and `priority`. `reason` is a attribute field not a Cargo.toml field.

With the `reason` set, `cargo` spews:

```
warning: .../storage/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../triehash/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../firewood-macros/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../firewood/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../benchmark/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../metrics/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../ffi/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../replay/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
warning: .../fwdctl/Cargo.toml: unused manifest key: `lints.clippy.unwrap_used.reason`
```

rkuris: Nice catch — you're right, `reason` isn't a manifest key. Verified locally on cargo 1.97.0: it's silently ignored with `warning: unused manifest key`. Doc it is.

rkuris (approved):
LGTM. Verified the workspace-allow and ffi crate-root claims against the codebase, and confirmed locally that Cargo's `[lints]` table doesn't support a `reason` key.

PR #2121, opened 2026-07-07
refactor(proofs): dispatch proof parsing on the self-describing header byte [5/9]

RodrigoVillar:
## Why this should be merged

Proofs already identify their hashing scheme in the header, but parsing still depends on the binary's compile-time mode. That can decode the proof body with the wrong grammar and prevents one process from safely parsing proofs from both schemes.

In the stack, this is the first runtime dispatch boundary: proof bytes select their parser from their own header, using the common representation established by PRs 3 and 4. It deliberately stops short of runtime database selection; PR 6
makes the typed database core mode-generic, and PR 7 exposes that capability to mode-agnostic callers.

## How this works

This change makes header validation resolve a `NodeHashAlgorithm` and threads that value into a body-phase `ProofReader`. Reads of mode-dependent values then dispatch on the algorithm carried by the reader. Unknown header values are rejected, while verification APIs take the caller's expected algorithm and fail with `HashModeMismatch` before hashing when it differs from the proof's self-described mode.

This PR is intentionally limited to parsing and early validation. Trie hashing continues to use the compile-selected implementation until the typed stack is genericized in the next PR. Consequently, a proof can be parsed regardless of its header mode in this PR, but it can only be cryptographically validated when that mode matches the binary's compile-time mode. The database and FFI call sites therefore pass the compile-selected mode as their expectation, and the mismatch guard rejects an opposite-mode proof before it can be hashed with the wrong algorithm.

PR 6 removes this binary-level restriction by dispatching proof hashing and the typed database stack through the supplied runtime algorithm. From that point onward, the remaining mismatch guard compares the proof header with the validation context: standalone verification can use the proof's own mode, while database-bound change-proof verification requires the proof mode to match that database's mode. PR 7 then exposes both typed implementations
through one runtime-selected, object-safe binary.

## How this was tested

Added tests for accepting both known header modes, rejecting unknown modes, decoding `HashType` with the threaded runtime mode, rejecting range- and change-proof mode mismatches, and parsing hashed value digests independently of the compile-time feature.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/ffi/src/proofs/range.rs
+++ b/ffi/src/proofs/range.rs
@@ -4,7 +4,7 @@
 use std::num::NonZeroUsize;
 
 use firewood::{
-    KeyRange, ProofError, RangeProofVerificationContext,
+    KeyRange, NodeHashAlgorithm, ProofError, RangeProofVerificationContext,
     api::{self, DbView, FrozenRangeProof, HashKey},
 };
 use firewood_metrics::{MetricsContext, firewood_counter};
@@ -129,11 +129,13 @@ impl<'db> RangeProofContext<'db> {
 
         debug_assert!(self.verification.is_none());
 
+        // Expected mode is compile-time default - any proof with a different mode is rejected up front.
         self.verification = Some(firewood::verify_range_proof_structure(
             &self.proof,
             root,
             start_key,
             end_key,
+            NodeHashAlgorithm::compile_option(),
             max_length,
         )?);
         Ok(())
--- a/firewood/src/merkle/tests/change/fuzz_common.rs
+++ b/firewood/src/merkle/tests/change/fuzz_common.rs
@@ -9,7 +9,7 @@ use crate::api::{BatchOp, FrozenChangeProof, FrozenRangeProof, HashKey};
 use crate::db::Db;
 use crate::merkle::{Key, Value};
 use crate::{ChangeProof, Proof, ProofNode, verify_change_proof_structure};
-use firewood_storage::SeededRng;
+use firewood_storage::{NodeHashAlgorithm, SeededRng};
 
 /// Build a `FrozenChangeProof` from mutated parts.
 pub(in crate::merkle::tests) fn build_change_proof(
@@ -47,7 +47,16 @@ pub(in crate::merkle::tests) fn change_proof_rejected(
     first: Option<&[u8]>,
     last: Option<&[u8]>,
 ) -> bool {
-    match verify_change_proof_structure(proof, end_root.clone(), first, last, None) {
+    // This module is ethhash-gated (only the ethhash-shaped fuzzer uses it),
+    // so every proof it sees is Ethereum-mode.
+    match verify_change_proof_structure(
+        proof,
+        end_root.clone(),
+        first,
+        last,
+        NodeHashAlgorithm::Ethereum,
+        None,
+    ) {
         Err(_) => true,
         Ok(ctx) => verify_and_check(db, proof, &ctx, start_root.clone()).is_err(),
     }
--- a/firewood/src/proofs/change.rs
+++ b/firewood/src/proofs/change.rs
@@ -3,6 +3,8 @@
 
 use std::{fmt::Debug, num::NonZeroUsize};
 
+use firewood_storage::{DefaultHashMode, HashMode, NodeHashAlgorithm};
+
 use crate::{
     Proof, ProofCollection, ProofError,
     api::{self, FrozenChangeProof, HashKey},
@@ -20,6 +22,11 @@ pub struct ChangeProof<K: AsRef<[u8]> + Debug, V: AsRef<[u8]> + Debug, H> {
     start_proof: Proof<H>,
     end_proof: Proof<H>,
     batch_ops: Box<[BatchOp<K, V>]>,
+    /// The hash algorithm this proof was constructed or parsed with. For proofs
+    /// built in this binary it is the compile default; for proofs parsed via
+    /// [`FrozenChangeProof::from_slice`](crate::api::FrozenChangeProof::from_slice)
+    /// it is resolved from the self-describing header byte.
+    hash_mode: NodeHashAlgorithm,
 }
 
 impl<K, V, H> std::fmt::Debug for ChangeProof<K, V, H>
@@ -34,6 +41,7 @@ where
             .field("start_proof", &self.start_proof)
             .field("end_proof", &self.end_proof)
             .field("batch_ops", &self.batch_ops)
+            .field("hash_mode", &self.hash_mode)
             .finish()
     }
 }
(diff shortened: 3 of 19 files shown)


ffi/src/proofs/range.rs
```diff
@@ -129,11 +129,13 @@ impl<'db> RangeProofContext<'db> {
 
         debug_assert!(self.verification.is_none());
 
+        // Expected mode is compile-time default - any proof with a different mode is rejected up front.
         self.verification = Some(firewood::verify_range_proof_structure(
             &self.proof,
             root,
             start_key,
             end_key,
+            NodeHashAlgorithm::compile_option(),
```
demosdemon, line 138: The proof object will need to record which algorithm it uses. The proof already serializes that based on `DefaultHashMode` but really it should be a field on the proof.

https://github.com/ava-labs/firewood/blob/e7c2aecd2c1bb77cea26ea6e2312c7bc6ce5597b/firewood/src/proofs/ser.rs#L225-L250

firewood/src/proofs/change.rs
```diff
@@ -51,14 +59,42 @@ where
...
         end_proof: Proof<H>,
         key_values: Box<[BatchOp<K, V>]>,
+    ) -> Self {
+        // Proofs built in this binary carry the compile-default mode; the parse
+        // path stamps the resolved header mode via `new_with_hash_mode`.
+        Self::with_hash_mode(
+            start_proof,
+            end_proof,
+            key_values,
+            DefaultHashMode::ALGORITHM,
```
demosdemon, line 69: there's inconsistent use of `DefaultHashMode::ALGORITHM` vs `NodeHashAlgorithm::compile_option`. They're they same thing, so what's the intent?

firewood/src/proofs/header.rs
```diff
@@ -50,19 +52,26 @@ impl From<ProofType> for Header {
...
-    /// possible reasons.
+    /// Returns:
+    /// - [`InvalidHeader`] if the header is invalid. See the enum variants for
+    ///   possible reasons.
+    /// - [`InvalidHeader::UnsupportedHashMode`] if the header's `hash_mode` byte is unrecognized.
     pub(super) fn validate(
         &self,
         expected_type: Option<ProofType>,
-    ) -> Result<ProofType, InvalidHeader> {
+    ) -> Result<(ProofType, NodeHashAlgorithm), InvalidHeader> {
```
demosdemon: this is kludgey and because the proof object doesn't record the algorithm type. I'd rather the validation for this be done inside this method and not returned separately.

demosdemon (changes requested):

firewood/src/proofs/reader.rs
```diff
@@ -21,17 +23,69 @@ pub(super) trait Version0: Sized {
...
+pub(super) struct NoAlgorithm;
+
+pub(super) struct ProofReader<'a, M = NodeHashAlgorithm> {
     data: &'a [u8],
     offset: usize,
+    /// The hash algorithm the proof being read was encoded with, resolved from
+    /// the self-describing `hash_mode` header byte. During the header phase
+    /// this is [`NoAlgorithm`]: no placeholder mode ever exists.
+    mode: M,
 }
```
demosdemon, line 38: I like the design of this but unfortunately it is also an artifact of the Proof object not recording its algorithm. This shouldn't be necessary if it did.

ffi/src/proofs/range.rs
demosdemon: I see. I misinterpreted this field. This should be taken from the database and not minted from compile-time options.

demosdemon: But, there is no database for this method. That's the kicker. 

Unfortunately, when this moves to runtime we can't assume any option. The compile-time default may or may not be valid. We need a configuration for this.

demosdemon: TBH, I think I'd rather get rid of this method. What do you think @alarso16?

firewood/src/merkle/tests/change/fuzz_common.rs
```diff
@@ -47,7 +47,16 @@ pub(in crate::merkle::tests) fn change_proof_rejected(
     first: Option<&[u8]>,
     last: Option<&[u8]>,
 ) -> bool {
-    match verify_change_proof_structure(proof, end_root.clone(), first, last, None) {
```
demosdemon, line 50: should this use the `verify_change_proof_structure` from firewood/src/merkle/tests/change/mod.rs instead?

firewood/src/proofs/reader.rs
demosdemon: ok now that I realize you did add the field to the proofs, I was confused by this and the fact that `ser.rs` doesn't serialize the value from the proof object.

firewood/src/proofs/header.rs
demosdemon: I understand better now. This should return a struct and not a tuple. It seems likely we would need to observe more validated fields from the header in the future.

firewood/src/merkle/tests/change/fuzz_common.rs
demosdemon: nvm, missed it using `Ethereum` and not the default.

firewood/src/proofs/change.rs
```diff
@@ -20,6 +22,11 @@ pub struct ChangeProof<K: AsRef<[u8]> + Debug, V: AsRef<[u8]> + Debug, H> {
     start_proof: Proof<H>,
     end_proof: Proof<H>,
     batch_ops: Box<[BatchOp<K, V>]>,
+    /// The hash algorithm this proof was constructed or parsed with. For proofs
+    /// built in this binary it is the compile default; for proofs parsed via
+    /// [`FrozenChangeProof::from_slice`](crate::api::FrozenChangeProof::from_slice)
+    /// it is resolved from the self-describing header byte.
+    hash_mode: NodeHashAlgorithm,
```
demosdemon, line 29: my adhd review glazed over this and similarly for range. The big problem is we're not writing this out in `ser.rs`.

RodrigoVillar: After some reflection, I realize that PR 6 solves this. However, we shouldn't be breaking invariants (in this case, roundtripping) between PRs in a stack, so I went ahead and moved those changes here: [4904fd6](https://github.com/ava-labs/firewood/pull/2121/commits/4904fd6d1604c7662eb6f3064b56f96a266f6b23).

firewood/src/proofs/header.rs
RodrigoVillar: Good call - done here: [0441ae7](https://github.com/ava-labs/firewood/pull/2121/commits/0441ae7bf565164db936aacb643c42a7848af7a5)

firewood/src/proofs/change.rs
RodrigoVillar: Good catch - I went ahead and switched to using `DefaultHashMode::ALGORITHM` for the most part here: [b4fe0b9](https://github.com/ava-labs/firewood/pull/2121/commits/b4fe0b9a18a31a31d53a4de33e301e7191710ec5). One note: `NodeHashAlgorithm::compile_option()` is still used in the `ffi` package as using `DefaultHashMode` would reintroduce the `firewood-storage` dep into the `ffi` package, which was recently removed. This, however, is removed in a follow-up PR in this stack.

demosdemon (approved):

PR #2122, opened 2026-07-07
feat!: runtime hash-mode selection [7/9]

RodrigoVillar:
## Why this should be merged

The typed stack can represent either hashing scheme, but callers still need a single stable type that can open and operate on databases whose mode is chosen at runtime. The FFI, CLI, and other mode-agnostic consumers should not need to duplicate every operation over `Db<EthHash>` and `Db<MerkleDbHash>`.

This is the point in the stack where a single Firewood binary can open and operate on Ethereum and MerkleDB databases at the same time. Each database still uses exactly one mode, recorded in its header, but the mode is no longer fixed for the process as a whole.

In the stack, this is the user-visible payoff of the preceding refactors: it erases the typed implementation from PR 6 behind a runtime-selected interface used by the FFI, CLI, and other mode-agnostic callers. PR 8 then broadens
dual-mode test coverage, and PR 9 removes the now-obsolete compile-time feature and compatibility bridge.

## How this works

This change adds object-safe database, proposal, committed-revision, and reconstruction traits, plus owned batch and boxed iterator boundaries where generic methods are not object-safe. The new `firewood::open` entry point reads `DbConfig::node_hash_algorithm`, constructs the corresponding typed database, and returns `Box<dyn DynDb>`. Existing database headers are validated against the requested mode before use.

The FFI, `fwdctl`, replay, benchmark, proof, and reconstruction paths are moved onto the erased boundary. Runtime mode information remains available on views, and Ethereum-only operations explicitly reject MerkleDB databases rather than relying on compile-time availability.

The `ethhash` feature remains temporarily in this PR as a compatibility bridge; it no longer prevents the binary from serving both modes. The following test PR exercises both implementations in one binary, and the final PR removes the
feature and the remaining compile-time compatibility shims entirely.

## How this was tested

Database tests open Ethereum and MerkleDB databases side by side, verify empty and committed roots, check proof metadata, and confirm persisted headers restore the selected mode after reopening. Reconstruction is tested through `DefaultHashMode` under both CI feature configurations. Unit tests also cover configured hash-mode mismatches and header validation.

## Breaking Changes

_Note_: `fwdctl` create now requires an explicit `--hash-mode <merkle-db|ethereum>`. Every other database command requires an existing database and detects its hashing mode from the persisted header. In particular, `import` and `replay` no longer create databases; run `fwdctl create` first. `replay` still truncates the existing database by default while preserving its selected hashing mode.

- [x] firewood
- [x] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [x] fwdctl

### DIFF
--- a/storage/src/hashmode.rs
+++ b/storage/src/hashmode.rs
@@ -82,10 +82,14 @@ mod tests {
 
     #[test]
     fn default_hash_mode_matches_compile_option() {
-        assert_eq!(
-            DefaultHashMode::ALGORITHM,
-            NodeHashAlgorithm::compile_option()
-        );
+        // `DefaultHashMode` must stay coupled to the `ethhash` feature: it is
+        // the cfg-selected default `H` until the flag is removed (PR 6).
+        let expected = if cfg!(feature = "ethhash") {
+            NodeHashAlgorithm::Ethereum
+        } else {
+            NodeHashAlgorithm::MerkleDB
+        };
+        assert_eq!(DefaultHashMode::ALGORITHM, expected);
     }
 
     #[test]
--- a/storage/src/hashmode.rs
+++ b/storage/src/hashmode.rs
@@ -72,24 +72,23 @@ pub type DefaultHashMode = EthHash;
 #[cfg(not(feature = "ethhash"))]
 pub type DefaultHashMode = MerkleDbHash;
 
+const _: () = {
+    // `DefaultHashMode` must stay coupled to the `ethhash` feature: it is
+    // the cfg-selected default `H` until the flag is removed.
+    let expected = if cfg!(feature = "ethhash") {
+        NodeHashAlgorithm::Ethereum
+    } else {
+        NodeHashAlgorithm::MerkleDB
+    };
+    assert!(DefaultHashMode::ALGORITHM as u8 == expected as u8);
+};
+
 #[cfg(test)]
 mod tests {
     use super::*;
 
     fn path_of_len(len: usize) -> Path {
         Path((0..len).map(|_| 0u8).collect())
-    }
-
-    #[test]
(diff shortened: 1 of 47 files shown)


storage/src/hashmode.rs
```diff
@@ -82,10 +82,14 @@ mod tests {
...
-            NodeHashAlgorithm::compile_option()
-        );
+        // `DefaultHashMode` must stay coupled to the `ethhash` feature: it is
+        // the cfg-selected default `H` until the flag is removed.
+        let expected = if cfg!(feature = "ethhash") {
+            NodeHashAlgorithm::Ethereum
+        } else {
+            NodeHashAlgorithm::MerkleDB
+        };
+        assert_eq!(DefaultHashMode::ALGORITHM, expected);
```
demosdemon: if you want a compile time assertion, you can do this:

```rust
const _: () = {
    // `DefaultHashMode` must stay coupled to the `ethhash` feature: it is
    // the cfg-selected default `H` until the flag is removed.
    let expected = if cfg!(feature = "ethhash") {
        NodeHashAlgorithm::Ethereum
    } else {
        NodeHashAlgorithm::MerkleDB
    };
    assert!(DefaultHashMode::ALGORITHM as u8 == expected as u8);
};
```

demosdemon (approved):
This is really clean 🥇. The merge conflict is trivial.

storage/src/hashmode.rs
RodrigoVillar: Great suggestion - done here: [bc5dc80](https://github.com/ava-labs/firewood/pull/2122/commits/bc5dc803eb2b952b5db42c0e50db73c484f12864).

PR #2125, opened 2026-07-08
test: run every applicable test under both hash modes [8/9]

RodrigoVillar:
## Why this should be merged

Removing the feature matrix would otherwise leave many mode-agnostic tests running against only one concrete implementation. Both runtime-selectable paths need direct coverage in the same binary before the compile-time feature is removed.

In the stack, this is the confidence layer between enabling runtime selection in PR 7 and deleting the compile-time fallback in PR 9. It adds no new production selection mechanism; instead, it demonstrates that shared behavior
really works through both implementations and keeps single-mode assumptions explicit in tests.

## How this works

This change extends the argument-free `#[firewood_macros::hash_mode]` test attribute to generate Ethereum and MerkleDB wrappers for a generic test. It supports `test_case`, forwards test attributes and return types, and rejects
arguments so single-mode tests stay explicit: they use `#[test]` and name the concrete mode or pass the runtime algorithm directly.

Mode-agnostic helpers and tests are genericized and run under both modes. Genuinely mode-specific vectors remain explicitly pinned, while existing cross-mode and golden-format tests keep their purpose. `fwdctl` also gains a
MerkleDB create/insert/get/root smoke path.

## How this was tested

Added compile-pass coverage for the bare macro and `test_case` composition, compile-fail coverage for macro arguments and invalid function signatures, and the MerkleDB `fwdctl` smoke test. The defensive value-digest regression also explicitly verifies that the runtime algorithm, not the structural trie type, controls the guard.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1931,6 +1931,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+ "test-case",
  "trybuild",
 ]
 
@@ -1973,6 +1974,7 @@ dependencies = [
  "derive-where",
  "enum-as-inner",
  "fastrace",
+ "firewood-macros",
  "firewood-metrics",
  "fjall",
  "git-version",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1927,6 +1927,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+ "test-case",
  "trybuild",
 ]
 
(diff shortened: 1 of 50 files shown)


RodrigoVillar:
Removing this from the [v0.10.0](https://github.com/ava-labs/firewood/milestone/28) milestone since there's additional follow-ups for removing the `ethhash` compile-time flag.

demosdemon (approved):

PR #2126, opened 2026-07-09
feat: add firewood_build_info metric with git sha and tag

rkuris:
## Why this should be merged

Operators currently have no way to tell from metrics which firewood build a node is running. A constant build-info gauge is the standard Prometheus pattern for joining version info onto other metrics.

## How this works

A new `build.rs` captures the git commit sha and any exact-match tag at build time via rustc-env, with rerun-if-changed hooks on the git dir so values stay fresh across checkouts. `Db::new` sets the `firewood_build_info` gauge to a constant 1 with `git_sha` and `git_tag` labels when a database is opened. Falls back to `sha=unknown` / empty tag when built outside a git checkout (e.g. from a published crate).

## How this was tested

`cargo nextest run --workspace --features ethhash,logger --all-targets`; `cargo +nightly-2026-07-05 clippy` with `-D warnings` (debug and maxperf) per the updated PR gate; `cargo doc --no-deps`; `markdownlint-cli2` on METRICS.md.

## Breaking Changes

none.


demosdemon (changes requested):
We already have version info sourced via the db header. We should use the [same library](https://docs.rs/git-version/latest/git_version/macro.git_version.html) for that. 

https://github.com/ava-labs/firewood/blob/c88f2d8f85a7ead09145d3740a64f1753ece7ee4/storage/src/nodestore/header.rs#L158-L239

rkuris:
> We already have version info sourced via the db header. We should use the [same library](https://docs.rs/git-version/latest/git_version/macro.git_version.html) for that.

Good callout. Yeah this was pretty miserable as originally coded. It's much better now. That plus fighting the rebase and it's ready for review again.

demosdemon (approved):

PR #2134, opened 2026-07-10
test(ffi): add C-Chain replay test and CI job

### LINKED ISSUE #2131: Add replay test to CI to reexecute C-Chain blocks
As a signal to check if a PR could break Firewood on the C-Chain, we should add the test that does the following:

- Pulls a replay log from a set of C-Chain blocks (e.g. blocks 50m-50.1m)
- Reexecutes the replay log
- Verifies that merkle root hash is as expected

Using the replay test here would avoid having to import Coreth/SAEVM.

We run reexecution tests for the C-Chain for every PR in AvalancheGo, but this is flawed as any changes in Firewood that could break the C-Chain would only be caught once we make a version bump. As a signal to check if a PR could break Firewood on the C-Chain, we should add a test that does the following:

- Pulls a replay log from a set of C-Chain blocks (e.g. blocks 50m-50.1m)
- Reexecutes the replay log
- Verifies that merkle root hash is as expected

Using the replay test here would allow us to dogfood our code while also avoiding having to import Coreth/SAEVM.

RodrigoVillar:
## Why this should be merged

Closes #2131.

Changes to Firewood that would break the C-Chain are currently only caught after a version bump into AvalancheGo, when its reexecution benchmarks run. This PR adds a replay test that runs on every Firewood PR (and on `main`), catching C-Chain-breaking regressions before they ship — without importing Coreth/SAEVM.

## How this works

`TestCChainReplay` (in `ffi`) replays a recorded log of the FFI operations issued while executing C-Chain mainnet blocks 0–250,000 against a fresh Firewood database built with `ethhash`, and verifies state roots along the way:

- Configuration is env-driven: `CCHAIN_REPLAY_REQUIRED` gates the test (unset → skip, so the regular `go test ./...` runs are unaffected; set → the test must run and any missing/malformed configuration is a hard failure, so CI can never silently skip). `CCHAIN_REPLAY_LOG_PATH` and `CCHAIN_REPLAY_EXPECTED_ROOT` supply the artifact and its expected final root.

Replay artifacts are matrix entries (`name` / `replay-log` / `expected-root`), so additional ranges — e.g. a mid-chain window replayed on top of a database snapshot — are one-row additions.

Prerequisites for the job to go green (in progress):
- [x] Upload `cchain-mainnet-250k.replay` to `s3://firewood-cchain-replay-logs`
- [x] `AWS_S3_READ_ONLY_ROLE` secret on this repo (SecOps request pending)

## How this was tested

- Full local run against the 0–250,000 artifact: 250,001 commits
- New unit tests for the env contract (`TestCChainReplayParamsFromEnvNotRequired` / `Valid` / `RootErrors`).

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


AminR443 (changes requested):
A replay log already tracks root hashes returned by `Commit()` and if `VerifyHashes` is set when replayed, it'll make sure that the roots match. I think this PR introduces a bit of redundancy as there is a test present in `replay_test.go`, playing a very similar role already:
https://github.com/ava-labs/firewood/blob/405c26e8d4e7c54069eec58c76f6db02e1d6aa69/ffi/replay_test.go#L93-L99
This applies to `CCHAIN_REPLAY_EXPECTED_ROOT` variable as well. If we're trusting that replay log is obtained from a correct execution, there is no need to perform that check. However, you might argue that this additional check could be useful.

I think better direction would be to add a new env variable `REPLAY_EXPECTED_ROOT` to `TestReplayLogExecution` and add the root hash check to the end of it. Then this PR will become a tidy ci file that just reuses the test. Feel free to re-request review after making the changes and I'll accept.

Also I forgot to mention that the second version of replay logs is also coming soon, which records more data and will ease catching bugs and mismatches.

AminR443 (approved):
Looks great. Excited to see this merge when we get S3 access.

RodrigoVillar:
Replay log CI run: https://github.com/ava-labs/firewood/actions/runs/30318353312/job/90149332389?pr=2134

PR #2135, opened 2026-07-12
refactor: thread HashMode through the typed database stack [6/9]

RodrigoVillar:
## Why this should be merged

Parsing a proof by runtime mode is not enough if database, trie, storage, and reconstruction operations still hash through a compile-selected default. The selected mode must be represented throughout the typed implementation before a safe object-erased runtime entry point can be added.

In the stack, this is the typed core of the migration. It consumes the low-level mode abstraction, shared representation, codecs, and proof dispatch from PRs 2 through 5, then makes every database operation consistently generic over one hash mode. PR 7 can therefore add runtime selection as a thin object-safe boundary instead of duplicating database logic.

## How this works

This change threads an explicit `H: HashMode` through `Db`, `Proposal`, `RevisionManager`, persistence, `Merkle`, proof generation and verification, `NodeStore`, checker paths, and reconstruction state. Mode-dependent hashing
and account behavior dispatch through `H` rather than feature gates.

Proof reconciliation also receives the explicit runtime `NodeHashAlgorithm`. That value controls the defensive value-digest guard and the Ethereum account-value relaxation because a range-proving trie can have a structural type that differs from the proof mode. This keeps policy tied to the proof being processed rather than accidentally inferring it from a helper type.

This PR remains a typed refactor: production construction still uses the temporary compile-selected bridge, and object-safe runtime selection is added separately.

## How this was tested

CI.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1963,7 +1963,6 @@ dependencies = [
  "bumpalo",
  "bytemuck",
  "bytemuck_derive",
- "cfg-if",
  "criterion",
  "derive-where",
  "enum-as-inner",
--- a/benchmark/Cargo.toml
+++ b/benchmark/Cargo.toml
@@ -57,6 +57,7 @@ features = ["mio", "net", "parking_lot", "rt", "time"]
 
 [features]
 default = ["prometheus"]
+ethhash = ["firewood/ethhash"]
 logger = ["firewood/logger"]
 prometheus = [
      "dep:metrics-exporter-prometheus",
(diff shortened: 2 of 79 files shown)


RodrigoVillar:
Putting back in draft as there are several merge conflicts that need to be resolved.

demosdemon (approved):

PR #2140, opened 2026-07-17
fix(ffi): serialize RangeProof handle methods against Free

### LINKED ISSUE #2137: Segmentation Violation in`FindNextKey`
Found in an avalanchego [PR](https://github.com/ava-labs/avalanchego/actions/runs/29498132954/job/87619936393?pr=5635)

Relevant snippets:

Actual error:
```
austin.larson  [9 minutes ago](https://platwizards.slack.com/archives/C075H83CMSQ/p1784208338664989?thread_ts=1784208253.117069&cid=C075H83CMSQ)
SIGSEGV: segmentation violation
PC=0x183876f50 m=9 sigcode=2 addr=0x1ca0c8146d89a83
signal arrived during cgo execution

goroutine 537 gp=0xc000093880 m=9 mp=0xc00040a008 [syscall]:
runtime.cgocall(0x105720114, 0xc0000d9778)
	/Users/runner/hostedtoolcache/go/1.25.10/arm64/src/runtime/cgocall.go:167 +0x58 fp=0xc0000d9740 sp=0xc0000d9700 pc=0x1049fb238
github.com/ava-labs/firewood-go-ethhash/ffi._Cfunc_fwd_range_proof_find_next_key(0x303fab380)
	_cgo_gotypes.go:1525 +0x40 fp=0xc0000d9770 sp=0xc0000d9740 pc=0x104eb4e40
github.com/ava-labs/firewood-go-ethhash/ffi.(*RangeProof).FindNextKey.func1(...)
```

Caller code in `database/merkle/firewood/syncer/syncer.go`:
```go
_, err := db.db.VerifyAndCommitRangeProof(proof.rp, start, end, ffi.Hash(proof.root), uint32(proof.maxLength))
if err != nil {
	return maybe.Nothing[[]byte](), err
}

nextKeyRange, err := proof.rp.FindNextKey() // segfault here
if err != nil {
	return maybe.Nothing[[]byte](), err
}
```

demosdemon:
## Why this should be merged

Fixes #2137 — an intermittent `SIGSEGV: ... signal arrived during cgo execution` inside `fwd_range_proof_find_next_key`, hit during avalanchego state sync.

`RangeProof.FindNextKey` (and `Verify`, `MarshalBinary`) read `p.handle` and pass it into a cgo call **without** holding `p.lease.mu`, so they are not serialized against `(*RangeProof).Free`, which frees the underlying Rust `RangeProofContext` under that lock. In production the racing `Free` is the **GC finalizer** `RangeProof` registers: once the proof becomes unreachable — which it is for the duration of the cgo call, since these methods never touch `p` again after loading the handle — the finalizer runs concurrently and frees the context out from under the in-flight call → use-after-free.

`RangeProof` was the *only* handle type missing the `lease.mu.RLock` guard that `Proposal`, `Revision`, and `Iterator` already hold across every cgo call, and it intentionally stays out of the keep-alive registry so its finalizer can run — which is exactly what made the race reachable.

## How this works

Hold `p.lease.mu.RLock()` across the cgo call in `FindNextKey`, `Verify`, and `MarshalBinary`. The read lock both serializes against `Free` (which takes `lease.mu.Lock`) and keeps `p` reachable for the duration of the call, closing the finalizer window. The invariant is documented on the `handle` field so future methods preserve it.

## How this was tested

| Check | Result |
| --- | --- |
| `TestRangeProofMethodFreeRace` (races each method vs `Free`), **unfixed**, `go test -race` | data race on `p.handle` → FAIL |
| same, **with the fix** | PASS |
| full `ffi` package, `go test -race` + `GOEXPERIMENT=cgocheck2` (ethhash) | PASS |
| `gofmt` / `golangci-lint` | clean / 0 issues |

CI already runs the ffi suite under `-race` for both hash modes, so `TestRangeProofMethodFreeRace` guards this regression there.

## Breaking Changes

No breaking changes — internal locking only; the C and Go APIs and their observable behavior are unchanged.

---

> [!NOTE]
> Same-class issues intentionally left out of this focused fix (distinct shapes; to be handled separately):
>
> - **`RangeProof.CodeHashes`** — its Rust iterator borrows from the proof and is consumed across multiple calls, so it needs an iteration-spanning guard rather than the single-call one used here.
> - **`ChangeProof` method family** — the type has no `lease` field and needs its own keep-alive mechanism (it is slated for redesign).
> - **`UnmarshalBinary` then `VerifyRangeProof`** panics with `runtime.SetFinalizer: finalizer already set` — a separate latent bug surfaced while building the reproducer.

### DIFF
--- a/ffi/proofs_test.go
+++ b/ffi/proofs_test.go
@@ -6,6 +6,7 @@ package ffi
 import (
 	"encoding/hex"
 	"runtime"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


ffi/proofs_test.go
```diff
@@ -321,6 +322,90 @@ func TestRangeProofFindNextKey(t *testing.T) {
...
+// The finalizer's exact timing cannot be forced from Go — the proof must be
+// reachable to call a method on it, yet unreachable for the finalizer to run —
+// so this test drives the identical concurrent code paths directly: the method
+// and Free (the very method the finalizer calls) race on the same proof. Under
+// -race — which CI runs the ffi suite under — the unsynchronized p.handle
+// access is reported deterministically. Without -race the same race can surface
+// as the reported SIGSEGV, but the timing window is narrow and does not
+// reliably crash in a bounded run, so -race is the signal this test relies on.
+// Guarding each method with p.lease.mu fixes all of them.
+func TestRangeProofMethodFreeRace(t *testing.T) {
```
rkuris, line 348: 🏆 super nifty test

rkuris (approved):

PR #2143, opened 2026-07-17
fix(ffi): keep RangeProof alive during CodeHashes iteration

demosdemon:
## Why this should be merged

The Rust code-hash iterator *borrows* the proof for its whole lifetime —
`CodeIteratorHandle<'p>` wraps a `Box<dyn Iterator + 'p>` built from
`key_values.iter()`. `RangeProof.CodeHashes` referenced `p` only long enough to
create the iterator, so once the closure stopped mentioning it the GC could
reclaim the proof *mid-iteration* and run its finalizer, freeing the key-values
the iterator was still reading. Same use-after-free family as #2137, but
spanning the whole loop rather than a single call.

## Notes

`Next()` and `Free()` operate on the iterator's own handle, not `p.handle`, and
are deliberately left unlocked. The involuntary threat is the finalizer, which
the keepalive already prevents; a per-`Next` lock could not close a concurrent
explicit `Free` anyway, since that can still land between `Next`s. Concurrently
calling `Free` during iteration remains unsupported, as elsewhere on
`RangeProof`. This is why `codeIterator` differs from the registered `Iterator`,
which *does* lock per `Next`: that one is force-droppable via
`WithForceCloseHandles`.

Finalizer timing can't be forced, and `iter.Pull2` would *mask* the bug by
retaining the closure, so the regression test drops every reference to the proof
except the one `CodeHashes` captures and uses a `runtime.AddCleanup` canary to
assert the proof survives the iteration. Stashing the fix turns it red
("garbage-collected mid-iteration").

## Breaking Changes

None — internal lifetime fix; public API and behavior unchanged.

### DIFF
--- a/ffi/proofs_test.go
+++ b/ffi/proofs_test.go
@@ -5,6 +5,7 @@ package ffi
 
 import (
 	"encoding/hex"
+	"iter"
 	"runtime"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


ffi/proofs_test.go
```diff
@@ -499,6 +500,59 @@ func TestRangeProofVerifyTwiceIdempotent(t *testing.T) {
...
+	r := require.New(t)
+	mode, err := inferHashingMode(t.Context())
+	r.NoError(err)
+	if mode != ethhashKey {
+		t.Skip("code hashes are only produced in ethhash mode")
+	}
+
+	db := newTestDatabase(t)
+	key := [32]byte{0x12, 0x34, 0x56} // account key must be length 32
+	val, err := hex.DecodeString("f8440164a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d")
```
RodrigoVillar, line 522: q: is there a reason why we choose `f844...` as the value here? If so, would it make sense to add a short comment describing this value/rename the variable name?

ffi/proofs_test.go
```diff
@@ -499,6 +500,59 @@ func TestRangeProofVerifyTwiceIdempotent(t *testing.T) {
...
+		r.NoError(p.UnmarshalBinary(proofBytes)) // arms the Free finalizer
+		runtime.AddCleanup(p, func(struct{}) { close(collected) }, struct{}{})
+		return p.CodeHashes()
+	}()
+
+	for _, err := range seq {
+		r.NoError(err)
+		// Mid-iteration: try hard to reclaim the proof. If CodeHashes stopped
+		// referencing it after creating the iterator, it is collectible here.
+		for range 50 {
```
RodrigoVillar, line 542: q: is there a reason why we choose `50` for the number of iterations here?

ffi/proofs_test.go
```diff
@@ -499,6 +500,59 @@ func TestRangeProofVerifyTwiceIdempotent(t *testing.T) {
...
+	seq := func() iter.Seq2[Hash, error] {
+		p := new(RangeProof)
+		r.NoError(p.UnmarshalBinary(proofBytes)) // arms the Free finalizer
+		runtime.AddCleanup(p, func(struct{}) { close(collected) }, struct{}{})
+		return p.CodeHashes()
+	}()
+
+	for _, err := range seq {
+		r.NoError(err)
+		// Mid-iteration: try hard to reclaim the proof. If CodeHashes stopped
```
RodrigoVillar, line 540: I had AI spike a version of this test which was more deterministic (i.e. from my understanding, the logic below can yield false positives if, for example, it takes `51` iterations to yield a instance of the proof being GC'd). The first attempt though instead replaced it with a timer, which did not resolve the underlying problem. Not a blocker, but I wonder if it makes sense to revisit this at some point to make the test more robust.

RodrigoVillar (approved):
LGTM, mostly comments falling under the style umbrella.

ffi/proofs_test.go
demosdemon: This is a string used multiple times in this file. I'm just continuing the trend without really thinking about it. This could be a const but I don't want to refactor unrelated things in this diff. I'll sweep that in a new PR.

ffi/proofs_test.go
demosdemon: 50 is a magic number that was successful in my attempts to reproduce. As I'll note below, this would be better with a structural change.

ffi/proofs_test.go
demosdemon: I figured out a way to resolve this deterministically. It requires some structural changes to the code I'll stack on top of this and the changeproof pr.

demosdemon: covered in #2209

ffi/proofs_test.go
demosdemon: also covered in #2209

ffi/proofs_test.go
demosdemon: This is an encoded account node where the magic has meaning, documented in #2209.

Issue #1978, opened 2026-05-07
ffi: decouple RangeProof from constructing database's keep-alive bookkeeping


rkuris:
**Is your feature request related to a problem? Please describe.**

The Go FFI's `RangeProof` is registered against the constructing database's keep-alive `WaitGroup` (via `proof.keepAliveHandle.init(db.keepAlives)` in `ffi/proofs.go:212`). This blocks legitimate workflows that the Rust side actually permits.

Concrete use case that does not work today:

1. Open database A.
2. `proof, _ := A.RangeProof(...)` — increments A's keep-alive WaitGroup.
3. `A.Close(ctx)` — blocks on `wg.Wait()` because `proof` is live. `WithForceCloseHandles` does not help: `RangeProof` is not in the live-handle registry (no `register` call), so `forceDropAll` cannot drop it either. Eventually ctx fires, `Close` returns `ErrActiveKeepAliveHandles`, and `fwd_close_db` is never called. **A stays open.**
4. Workaround: `proof.MarshalBinary()`, then `proof.Free()`, then close A, then open B, then `UnmarshalBinary` into a fresh `RangeProof`, then apply. This forces a byte round-trip the Rust side does not require.

The Rust type genuinely permits the "carry across `Close`" case. `fwd_db_range_proof` (`ffi/src/proofs/range.rs:318`) is declared `RangeProofResult<'static>` with the comment:

```
// static lifetime is safe because the returned `RangeProofResult` does not
// retain a reference to the provided database handle.
```

`RangeProofContext<'db>` only holds a real database borrow after `verify_and_propose` (via `ProposalState::Proposed(ProposalHandle<'db>)`), and that borrow lives inside the apply call, not across `Close`. A freshly-created or freshly-deserialized `RangeProof` holds zero database references.

`ChangeProofContext` has no lifetime parameter at all (`ffi/src/proofs/change.rs:56`), and the Go `ChangeProof` correctly has no `keepAliveHandle`. `RangeProof` is the inconsistent one.

**Describe the solution you'd like**

Decouple `RangeProof` from the constructing database's keep-alive bookkeeping:

- Remove `proof.keepAliveHandle.init(db.keepAlives)` from `RangeProof` constructors.
- Remove the `keepAliveHandle` field from the `RangeProof` struct (or make `Free` a plain free with no disown).
- Update `Database.Close` and `WithForceCloseHandles` docstrings: neither proof type belongs in the lists, because neither holds a database lease.

[...]

PR #2144, opened 2026-07-17
fix(ffi): fix ChangeProof use-after-free and double-SetFinalizer bugs

demosdemon:
## Why this should be merged

`ChangeProof` carries the same lifetime bugs `RangeProof` had (#2140–#2143) with
none of the guards:

- **Double `SetFinalizer` panic.** `Database.ChangeProof` and `UnmarshalBinary`
  both register `Free`, and `Free` doesn't clear it, so `create` →
  `UnmarshalBinary` (or `UnmarshalBinary` twice) reaches
  `fatal error: runtime.SetFinalizer: finalizer already set`.
- **Finalizer use-after-free** in `FindNextKey`, `MarshalBinary`, and
  `CodeHashes` — the receiver is dead once its handle has been read into the cgo
  call, and for `CodeHashes` the Rust iterator borrows the proof's data for its
  whole lifetime.

## Notes

Per #1978 and the existing `ChangeProof.Free` doc, `ChangeProof` deliberately
holds no database reference, so every fix here is `runtime.KeepAlive` rather
than a lease read-lock. **No database handle is introduced** and no struct field
is added — `ChangeProof` stays pure proof data. That is the one place this
diverges from the `RangeProof` fixes it otherwise mirrors.

`FindNextKey` and `MarshalBinary` get the keepalive but no dedicated regression
test: a single fast cgo call leaves no forceable window for the finalizer, and
with no lock involved there is no data race for `-race` to catch. Concurrently
calling `Free` during a call or iteration remains unsupported, as elsewhere.

The `proofLen` parameter drop in `newSerializedChangeProof` isn't cleanup for its
own sake — `unparam` flags it once the new callers land.

## Breaking Changes

None — internal lifetime fixes; public C and Go APIs and behavior unchanged.

### DIFF
--- a/ffi/proofs_test.go
+++ b/ffi/proofs_test.go
@@ -100,11 +100,10 @@ func newSerializedChangeProof(
 	db *Database,
 	startRoot, endRoot Hash,
 	startKey, endKey maybe,
-	proofLen uint32,
 ) []byte {
 	r := require.New(t)
 
-	proof, err := db.ChangeProof(startRoot, endRoot, startKey, endKey, proofLen)
+	proof, err := db.ChangeProof(startRoot, endRoot, startKey, endKey, changeProofLenUnbounded)
 	r.NoError(err)
 
 	proofBytes, err := proof.MarshalBinary()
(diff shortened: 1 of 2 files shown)


ffi/proofs_test.go
```diff
@@ -701,6 +700,105 @@ func TestRangeProofFinalizerCleanup(t *testing.T) {
...
+		}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := require.New(t)
+			p := tt.run(r) // must not fatally panic with "finalizer already set"
+			t.Cleanup(func() { _ = p.Free() })
+		})
+	}
```
RodrigoVillar:
```suggestion
	t.Run("create_then_unmarshal", func(t *testing.T) {
		r := require.New(t)
		p, err := db.ChangeProof(startRoot, endRoot, nothing(), nothing(), changeProofLenUnbounded)
		r.NoError(err)
		r.NoError(p.UnmarshalBinary(proofBytes))

		t.Cleanup(func() { _ = p.Free() })
	})

	t.Run("unmarshal_then_unmarshal", func(t *testing.T) {
		r := require.New(t)
		p := new(ChangeProof)
		r.NoError(p.UnmarshalBinary(proofBytes))
		r.NoError(p.UnmarshalBinary(proofBytes))

		t.Cleanup(func() { _ = p.Free() })
	})
```

Since we have just two test cases and each subtest is larger than the table-test scaffolding itself, I think it would be clearer if we declare two subtests directly here.

RodrigoVillar (approved):

ffi/proofs_test.go
demosdemon: Already done in the next PR

RodrigoVillar: Hmm looking at #2209, I see that `TestChangeProofSetFinalizerReset` was unchanged: 

https://github.com/ava-labs/firewood/blob/a45f9ad2b6f0cd9463245ca323bfd57bb766021a/ffi/proofs_test.go#L747-L774

demosdemon: Got it. I confused this for a different test.

PR #2146, opened 2026-07-21
fix(merkle): correct change-proof verification at range boundaries

### LINKED ISSUE #2136: Valid change proof rejected with `EndRootMismatch` after deleting a key just past the end key
### Affected area(s)

proofs

## Summary

`db.change_proof(...)` can produce a proof that the verifier itself then
rejects with `ProofError(EndRootMismatch)`. No tampering involved — generate a
proof, verify it against the same roots and bounds, and verification fails.

It happens when a key **greater than the end key** (i.e. just outside the
proven range) was deleted between the two revisions, and that key sat in the
same branch-child subtree the end bound descends into. The deletion is outside
the proven range, so it is (correctly) not in `batch_ops` — but the verifier's
reconstruction then still contains the deleted key, and the computed root
doesn't match `end_root`.

This is a false rejection of a valid proof, not a soundness hole. Reproduces
in both merkledb and ethhash modes. Found by the change-proof fuzzer
(`FIREWOOD_TEST_SEED=11081895077678951690`), minimized from 244 keys to 3.
Fuzzer is from PR #1951 (not yet merged).

## To reproduce

Save as `firewood/tests/change_proof_repro.rs`, run
`cargo test -p firewood --test change_proof_repro` (with or without
`--features ethhash`):

```rust
use firewood::api::{BatchOp, Db as _, Proposal as _};
use firewood::db::{Db, DbConfig};
use firewood::{NodeHashAlgorithm, verify_change_proof_root_hash, verify_change_proof_structure};

// start trie: { 0xfb00: 0x00 }
// end trie:   { 0xf1: 0x00, 0xf7: 0x00 }   (0xfb00 deleted; f1, f7 added)
// proof range: [0x00, 0xfb] — the delete of 0xfb00 is above the range,
// so batch_ops carries only the two puts.
#[test]
fn valid_change_proof_rejected_after_delete_above_end_bound() {
    let dir = tempfile::tempdir().unwrap();
    let config = DbConfig::builder()
        .node_hash_algorithm(NodeHashAlgorithm::compile_option())
        .build();
    let db = Db::new(dir.path(), config).unwrap();

    db.propose(vec![BatchOp::Put { key: b"\xfb\x00".as_slice(), value: b"\x00".as_slice() }])
        .unwrap().commit().unwrap();
    let start_root = db.root_hash().unwrap();

    let end_batch: Vec<BatchOp<&[u8], &[u8]>> = vec![
        BatchOp::Delete { key: b"\xfb\x00" },
        BatchOp::Put { key: b"\xf7", value: b"\x00" },
        BatchOp::Put { key: b"\xf1", value: b"\x00" },
    ];
    db.propose(end_batch).unwrap().commit().unwrap();
    let end_root = db.root_hash().unwrap();

    let (start, end) = (b"\x00".as_slice(), b"\xfb".as_slice());
    let proof = db
        .change_proof(start_root.clone(), end_root.clone(), Some(start), Some(end), None)
        .expect("change_proof should succeed");

    // Verify the proof against the same bounds and roots it was built for.
    let ctx = verify_change_proof_structure(&proof, end_root, Some(start), Some(end), None)
        .expect("structural verification should pass");
    let parent = db.revision(start_root).unwrap();
    let proposal = db.apply_change_proof_to_parent(&proof, &*parent).unwrap();
    verify_change_proof_root_hash(&proof, &ctx, &proposal)
        .expect("BUG: valid change proof failed verification");
}
```

Output:

```
BUG: valid change proof failed verification: ProofError(EndRootMismatch)
```

## Cause

The end bound `0xfb` doesn't exist in the end trie, so the end proof is an
exclusion proof: a single branch node (path `[f]`) whose child `b` is empty.

To check the root hash, the verifier rebuilds the end state as
start trie + `batch_ops`. Since the delete of `0xfb00` is above the range,
the rebuilt trie still contains `0xfb00` — under exactly that child `b`.
Note `0xfb00` sorts *after* `0xfb` — a longer key extending a shorter one
comes later in byte order — so it is outside the range even though it lives
inside the subtree where the trie walk for `0xfb` ends.

`compute_outside_children` (`firewood/src/merkle/mod.rs`, the
"Terminal is an ancestor of the boundary key" arm) then decides, per child,
whether the root hash uses the rebuilt trie or the proof's stored hashes.
Children past the bound are taken from the proof, but the child the bound
descends into (`b`) is deliberately taken from the rebuilt trie:

```rust
// The on-path child itself is left inside the range so its
// subtree is correctly computed from the applied batch.
```

That assumption is wrong here: everything under child `b` can lie above the
bound (`0xfb00 > 0xfb`), where `batch_ops` says nothing. The stale `0xfb00`
gets hashed in, and the computed root can't match `end_root`, which has no
child `b` at all. No other step removes it: the collapse passes only strip
along proof-node paths, and a delete — unlike a put — has no proof node that
could be reconciled into the rebuilt trie.

Variants confirming the diagnosis: keeping `0xfb00` unchanged passes; widening
the range to `[00, ff]` (so the delete is in `batch_ops`) passes; deleting
`0xfc00` instead (a different branch child) passes; and the bound need not be
a prefix of the deleted key — delete `0xfbff` with bound `0xfbc0` fails the
same way.

## Note for the fix

Simply taking that child from the proof ("it's empty") would be unsound: with
a longer bound, part of that subtree is *inside* the range (e.g. `0xfb00` ≤
bound `0xfbc0`), and an attacker could then drop an in-range delete from
`batch_ops` unnoticed. The fix has to validate the in-range part of the
subtree against the applied ops while adopting the end trie's state for the
out-of-range part.

### LINKED ISSUE #2138: Change-proof verification accepts a forged in-range Delete-to-Put
### Affected area(s)

proofs

### Describe the bug

Change-proof verification accepts a tampered proof. An in-range `Delete` is forged into a `Put`, and `verify_change_proof_root_hash` returns `Ok` even though it was given the correct roots and bounds.

This is the same symptom as #2091 (forged batch ops accepted). #2109 fixed #2091 in one arm of `compute_outside_children`, but not the arm this case exercises (see Root cause). It is also the accept-forgery counterpart of #2136 (a valid proof wrongly rejected); both come from right-edge boundary handling when the end bound is a prefix of a neighbouring key on its own trie path.

Found by the change-proof fuzzer (`FIREWOOD_TEST_SEED=6071609516474822149`), reproduced across large soaks, deterministic in both merkledb and ethhash.

## Reproduction

Save as `firewood/tests/change_proof_forged_delete.rs`, run `cargo test -p firewood --test change_proof_forged_delete` (optionally `--features ethhash`):

```rust
use firewood::api::{BatchOp, Db as _, Proposal as _};
use firewood::db::{Db, DbConfig};
use firewood::{
    ChangeProof, NodeHashAlgorithm, Proof, verify_change_proof_root_hash,
    verify_change_proof_structure,
};

// start trie:  { 0x56: .., 0x5601: .. }
// end trie:    { 0x5601: .. }            (0x56 deleted)
// proof range: [0x00, 0x5600]
// Forge the in-range Delete{0x56} into a Put; verification wrongly accepts.
#[test]
fn forged_in_range_delete_to_put_is_accepted() {
    let dir = tempfile::tempdir().unwrap();
    let config = DbConfig::builder()
        .node_hash_algorithm(NodeHashAlgorithm::compile_option())
        .build();
    let db = Db::new(dir.path(), config).unwrap();

    db.propose(vec![
        BatchOp::Put { key: b"\x56".as_slice(), value: b"\x01".as_slice() },
        BatchOp::Put { key: b"\x56\x01".as_slice(), value: b"\x01".as_slice() },
    ]).unwrap().commit().unwrap();
    let start_root = db.root_hash().unwrap();

    let end_batch: Vec<BatchOp<&[u8], &[u8]>> =
        vec![BatchOp::Delete { key: b"\x56".as_slice() }];
    db.propose(end_batch).unwrap().commit().unwrap();
    let end_root = db.root_hash().unwrap();

    // Honest change proof over [0x00, 0x5600].
    let start = b"\x00".as_slice();
    let end = b"\x56\x00".as_slice();
    let proof = db
        .change_proof(start_root.clone(), end_root.clone(), Some(start), Some(end), None)
        .expect("change_proof should succeed");

    // Forge: turn the in-range Delete{0x56} into a Put{0x56, "forged"}.
    let forged_ops = proof
        .batch_ops()
        .iter()
        .map(|op| match op {
            BatchOp::Delete { key } if key.as_ref() == b"\x56" => BatchOp::Put {
                key: key.clone(),
                value: Box::from(&b"forged"[..]),
            },
            other => other.clone(),
        })
        .collect::<Vec<_>>();
    let forged = ChangeProof::new(
        Proof::new(proof.start_proof().as_ref().into()),
        Proof::new(proof.end_proof().as_ref().into()),
        forged_ops.into_boxed_slice(),
    );

    // The forged proof must be rejected. It is currently accepted (the bug).
    let rejected = match verify_change_proof_structure(
        &forged, end_root.clone(), Some(start), Some(end), None,
    ) {
        Err(_) => true,
        Ok(ctx) => {
            let parent = db.revision(start_root).unwrap();
            let proposal = db.apply_change_proof_to_parent(&forged, &*parent).unwrap();
            verify_change_proof_root_hash(&forged, &ctx, &proposal).is_err()
        }
    };
    assert!(
        rejected,
        "SOUNDNESS BUG: forged Delete{{0x56}} to Put was accepted by change-proof verification"
    );
}
```

The assertion fires: the forged proof verifies. `0x56` is in range (`0x56 < 0x5600`), so forging its deletion must be detected. It is not, because `0x56` is a prefix of the non-existent end bound `0x5600`, and `0x5601` is out of range (`> 0x5600`) sharing the `0x56` path, so the verifier takes `0x56`'s subtree from the proof instead of validating it against the batch.

## Root cause

`compute_outside_children` decides, per child of a boundary proof node, whether that child's hash comes from the proof or is recomputed from the applied batch. Instrumentation shows this bug and #2136 hit different arms of its right-edge terminal handling:

- This bug hits the "divergence within the terminal's key" arm. For end bound `0x5600` the terminal is the `0x5601` leaf (nibbles `[5,6,0,1]`); the bound `[5,6,0,0]` diverges at the last nibble, so the whole terminal subtree is marked outside-range and the forged in-range ancestor `0x56` is never validated against the batch. #2109 did not touch this arm.
- #2136 hits the "terminal is an ancestor of the boundary key" arm, the one #2109 changed (it stopped marking the on-path child outside). That change is what makes #2136 reject a valid proof.

| Bug | Arm of `compute_outside_children` | Status |
|---|---|---|
| #2091 (forged accepted) | ancestor-on-path | fixed by #2109 |
| #2136 (valid rejected) | ancestor-on-path | appeared with #2109's fix (verified below) |
| this bug (forged accepted) | divergence-in-terminal | not addressed by #2109 |

Verified by a controlled revert: re-adding the `.set(on_path_nibble)` call that #2109 removed makes the #2136 valid proof verify again, and leaves this bug reproducing unchanged. So #2109's arm-B change is a soundness-for-completeness trade (it stopped the #2091 forgery but now rejects the valid #2136 proof), and this bug is a separate soundness gap in the divergence arm that #2109 never touched.

Both arms are too coarse: they mark a whole child of the boundary node as either inside or outside the range. The correct fix splits that child's subtree at the boundary, validating its in-range keys against the batch while adopting its out-of-range keys from the proof. Fixing one arm alone is not enough, and, as #2109 showed, flipping a whole child from one side to the other only swaps a soundness bug for a completeness bug. This bug (divergence arm) and #2136 (ancestor-on-path arm) should be fixed together.

### To reproduce

See above

### Expected behavior

The forged proof is rejected: verify_change_proof_root_hash returns Err.

### Additional context

_No response_

bernard-avalabs:
## Why this should be merged
  
Fixes #2136 and #2138.

Change-proof verification mishandled boundary subtrees that straddle the proven range — where an on-path child holds a mix of in-range and out-of-range keys. This caused both completeness and soundness bugs:

- **Completeness:** honest proofs were rejected with `EndRootMismatch` when an out-of-range key just past a bound was deleted — at the end bound (#2136, deleted key extends the bound) and its left-edge mirror (bound extends the deleted key, found by fuzzing).
- **Soundness:** a forged in-range `Delete`→`Put` whose key is a prefix of the end bound was accepted (#2138). Review of the fix also surfaced two further soundness holes, both fixed here: a proof omitting an in-range op could verify (the on-path child was decided from attacker-controlled `batch_ops`), and an unbounded-end proof with an emptied end proof could hide an omitted in-range op (an empty end bound sorted as the minimum key instead of +∞).

## What the omission hole required

The omission hole needs a specific boundary geometry, worth spelling out because it explains the shape of the regression fixtures:

1. An exclusion proof — an inclusion proof terminates at the boundary key, leaving no on-path child.
2. The terminal's key is a strict ancestor of the boundary key, whose next nibble selects the on-path child.
3. A key sits under that on-path child, on the in-range side of the boundary key — at or below an end bound, at or above a start bound. This is stricter than "sorts between the terminal and the boundary key": such a key can live under a different child, where a nibble comparison settles it. How many keys qualify depends on how far the boundary key runs past the terminal — at an end bound one extra nibble admits only the boundary key itself, more nibbles admit every in-range key under the child, and at a start bound the comparison reverses so one extra nibble leaves the whole child in range.
4. That key is absent from `end_root` — the on-path child is empty there, which is why the walk stopped at the terminal — but present in the proposal, because `batch_ops` omits its `Delete`.

Deciding this from `batch_ops` cannot work: an omission leaves no key there to find, so the child would be marked outside, its subtree taken from the (empty) proof, and the un-deleted key would never reach the computed root. That is why the on-path child is resolved against the proposal instead.

## How this works

- `classify_terminal` classifies each boundary terminal once (all-outside / on-path / nothing), and `compute_outside_children` applies it to the outside-children mask. `change_outside_children` is extracted from `verify_change_proof_root_hash` so the change-proof-specific boundary merging lives in one place.
- The straddling on-path child is decided against the proposal, never the `batch_ops`: it is taken from the proof only when the proving trie holds no in-range key under it (`on_path_child_in_range`). Structural verification already requires that child to be absent from `end_root` (`ExclusionProofMissingChild`), and the root-hash step confirms it rather than inheriting it, so the substitution does not depend on call ordering. The ordering contract that does still matter — the context's bounds defining the proven range — is documented on `verify_change_proof_root_hash`.
- `collapse_strip` clears a node's value only when its key is out of range and never flattens a valued node, so in-range values survive to be validated.
- The right bound is threaded as `Option<&[u8]>` (`None` = +∞) through the range checks, distinct from the empty slice (minimum key).

## How this was tested

- Six deterministic regression tests covering each bug class (both boundary edges, forge and omission variants, bounded and unbounded ends), run under both merkledb and `ethhash` hashing.
- Full workspace matrix: `cargo fmt`, `clippy` (dev + `maxperf` profiles, pinned nightly), `nextest` across no-features / `--no-default-features` / `ethhash,logger`; `cargo doc`; no FFI header drift.
- Merged `origin/main`, which brings in the ethhash-shaped proof fuzzer (#2080): `test_slow_ethhash_proof_fuzz` now exercises these changes on every `--profile ci` run.
- An earlier 5-hour change-proof fuzz bake (329,058 fixtures, both hash modes, zero failures) validated the boundary logic and reproduced the unbounded-end omission hole against the unfixed code. It ran against an earlier revision of this branch, and the op-omission mutation it used lives in the fuzzer branch rather than here. The bounded omission hole is pinned by a deterministic regression test.
- Re-baked against the tip (`7f56c42a9`): 1,477,016 fixtures across 3,190 chunks, both hash modes and both the release and dev profiles, zero failures. Like the bake above, it does not exercise op omission.

## Known limitations

Three pre-existing defects surfaced while reviewing this work. None is caused by it, and none is fixed here.

- **#2155** — a bounded change proof is rejected when a truncated change list ends with a delete. `proofs/change.rs` is untouched here, so this PR does not fix it, but it does change which shapes reach it: measured over 720 honestly generated proofs, it clears 53 such false rejections and adds 2.
- **#2154** — a change proof is rejected when every key in the range was deleted. Orthogonal to this work: it needs no limit and no truncation, and reproduces identically with and without this PR.
- **#2153** — an ethhash account's lone storage child is skipped by the storage-trie-root fold when that child is persisted. Surfaced while writing account-depth coverage for this work; `build_branch_parts` is untouched here and it reproduces identically with and without this PR.

## Breaking Changes

None to public APIs. Verification behavior changes as intended: forged/omission proofs described above are now rejected, and honest boundary-deletion proofs now verify.


firewood/src/merkle/tests/change/regression.rs
```diff
@@ -39,46 +105,326 @@ fn test_tampered_right_edge_delete_to_put_is_rejected() {
...
+        Vec::new().into_boxed_slice(),
+    );
+    assert!(
+        matches!(
+            verify(&db, &forged, &start_root, &end_root, Some(sk), None),
+            Err(api::Error::ProofError(crate::ProofError::EndRootMismatch))
+        ),
+        "SOUNDNESS BUG: an unbounded-end change proof omitting the in-range \
+         Delete{{0x53}} was accepted. The key remains in the proposal and its \
+         subtree must be recomputed, not taken from the start proof"
```
demosdemon: This overstates the assertion. Suggested change: state the invariant rather than a bug, for example "an in-range key left in the proposal must force its boundary child to be recomputed, not taken from the proof".

firewood/src/merkle/mod.rs
```diff
@@ -256,32 +257,120 @@ impl EdgeBoundary<'_> {
...
 /// it outside in addition to the usual "children with index > the on-path
 /// child" marking — the hash check then substitutes the proof's stored
 /// child hash there instead of recursing into the proving trie's
 /// (irrelevant) subtree.
+///
+/// The terminal's classification (`terminal`) is computed once by the caller
+/// via [`classify_terminal`] and applied here. When it is
+/// [`TerminalMarking::OnPath`], the straddling child's far side is marked, and
+/// the change-proof caller decides that child itself against the proving trie.
 fn compute_outside_children(
```
demosdemon: `compute_outside_children` takes a precondition it cannot validate. It accepts a `TerminalClass` that must have been derived from the same `boundary`, and the doc asks for this in prose. Returning the classification instead —`Result<(HashMap<..>, Option<TerminalClass>), ProofError>` — makes mismatch unrepresentable, saves four lines at the range-proof call site (`mod.rs:1177-1190`), and removes the need for the `outside_children_mask` test shim. Aside from this the four-way decomposition holds up cleanly, with no responsibilities leaking between `classify_terminal`, `compute_outside_children`, `change_outside_children`, and `boundary_child_source`.

firewood/src/merkle/collapse.rs
```diff
@@ -331,6 +357,58 @@ impl<S: ReadableStorage> Merkle<NodeStore<Mutable<Propose>, S>> {
...
+    /// otherwise — the sound default, since a recompute that mismatches
+    /// `end_root` is a rejection, so nothing is silently taken from the proof.
+    /// Only a child definitely present with no in-range key under it answers
+    /// [`BoundaryChildSource::Proof`].
+    pub(crate) fn boundary_child_source(
+        &self,
+        node_key: &[PathComponent],
+        on_path: PathComponent,
+        start_nib: &[u8],
+        end_nib: Option<&[u8]>,
```
demosdemon: We should continue to carry the `CollapseRange` through instead of decomposing it. Apply here and for `child_in_range`.

firewood/src/merkle/collapse.rs
```diff
@@ -7,10 +7,44 @@ use firewood_storage::{
...
+    merkle::{Merkle, get_helper},
+};
+
+/// The proven range for a collapse operation, in nibbles. Callers pass it
+/// wrapped in an `Option`: `None` skips range checking — off-path children are
+/// stripped freely and branch values cleared unconditionally — while `Some`
+/// enforces the range, where stripping an off-path child that holds an in-range
+/// key is a tamper and in-range values are kept.
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct CollapseRange<'a> {
```
demosdemon, line 24: This asymmetry deserves a mention in `CODE_REVIEW.md`. Suggested section:


  ```markdown
  - **Nibble range bounds**: A nibble-path lower bound uses the empty slice
    for "unbounded" — the empty slice already sorts as the minimum key. An
    upper bound must be `Option<&[u8]>` with `None` for unbounded (+∞);
    reusing the empty slice there judges every key out of range. When either
    bound is threaded through new code, check every comparison site. See
    `merkle::collapse::CollapseRange`.
  ```

demosdemon (commented):
This is RTM as-is but could use a few documentation changes and minor ergonomics changes.

firewood/src/merkle/tests/change/regression.rs
bernard-avalabs: Applied your suggestion (or a slight variation of it) to all five messages in the file rather than just this one.

firewood/src/merkle/mod.rs
bernard-avalabs: Done as suggested. `compute_outside_children` now calls `classify_terminal` itself and returns `Result<(HashMap<..>, Option<TerminalClass>), ProofError>`, so there's no way to pair a classification with a different boundary.

firewood/src/merkle/collapse.rs
bernard-avalabs: Done for both, and for two more: `nibble_in_range`, because `child_in_range` would otherwise take the struct apart again on its first line to call it, and `change_outside_children`, because it's what hands the pair down to `boundary_child_source`.

firewood/src/merkle/collapse.rs
bernard-avalabs: Added your text under `## Rust` in `CODE_REVIEW.md`.

demosdemon (approved):

Issue #2099, opened 2026-06-22
proof deserialization ~100x memory amplification


demosdemon:
### Describe the bug

```rust
// FIXME: we must somehow validate `num_items` matches what is expected
// An incorrect, or unexpectedly large value could lead to DoS via OOM
// or panicing
(0..num_items).map(|_| reader.read_v0_item()).collect()
```

The FIXME is correct, but incomplete in characterising the actual risk. The issue is not only a `num_items` panic — it is a **structural memory amplification**.

### To reproduce

A minimal `ProofNode` on the wire is **5 bytes**:

| Field | Bytes |
|---|---|
| `key` (varint 0, empty path) | 1 |
| `partial_len` (varint 0) | 1 |
| `value_digest` discriminant `0` (None) | 1 |
| `children_map` (zero, no children) | 2 |

A deserialized `ProofNode` in memory is **~528+ bytes** because `child_hashes: Children<Option<HashType>>` is `[Option<HashType>; 16]` regardless of how many children exist. This is a **~100× amplification ratio**. A 50 MB input of tight-packed minimal `ProofNode`s can produce ~5 GB of heap objects.

The same effect applies to `FrozenChangeProof` via the `BatchOp<Key, Value>` collection: a `BatchOp::Delete` with an empty key takes 2 bytes on the wire but requires a separate heap allocation for the `Box<[u8]>` key.

**There is no size bound enforced on `data: &[u8]` at any call site in the
deserialization path.** The only mitigation must come from callers.

> [!NOTE] 
> A large `num_items` varint *alone*, without backing data, does **not** cause OOM. `collect::<Result<Box<[T]>, E>>()` goes through `ResultShunt`, which does not propagate `TrustedLen`, so `Vec::with_capacity` is not seeded with `num_items`. Iteration short-circuits on the first missing item. The amplification attack requires real, parseable data.

PR #2147, opened 2026-07-22
refactor(storage)!: adopt arity-arrays for the U4 index type

demosdemon:
## Why this should be merged

First of three sequenced PRs migrating firewood-storage's hand-rolled trie child/index containers onto the externally-tested `arity-arrays` crate. This one is the small, foundational step: it makes `PathComponent.0` the `arity_arrays::index::U4` that the later containers index by, so the follow-up PRs need no index-conversion shims.

- PR 2 adopts `arity_arrays::PackedArray` for `ProofNode.child_hashes`, closing the #2099 P0 proof-deserialization memory-amplification vulnerability.
- PR 3 rebases `Children` on `arity_arrays::FixedArray`.

Builds on the `arity-arrays = "0.2.0"` dependency. No on-disk or on-wire format changes in this PR.

## How this works

- `U4` and `TryFromIntError` become re-exports of `arity_arrays::index::{U4, TryFromIntError}`. The hand-rolled 16-variant `Repr` enum, the inherent `U4` methods, `TryFrom<u8>`, and the formatting impls are removed — that behavior lives in, and is tested by, `arity-index`. `PathComponent::new_pair`/`join` are reimplemented inline because arity's `U4` exposes no `new_pair`/`join`.
- arity's `TryFromIntError` is `#[non_exhaustive]`, so the two validation sites in `path/component.rs` (`path_from_unpacked_bytes` and `try_from_maybe_u4`) **obtain** the error from `U4::try_from` rather than constructing a literal.
- `PathComponent` is marked `#[repr(transparent)]` with compile-time `size_of` assertions, and its zero-copy `&[u8]`⇄`&[PathComponent]` conversions delegate byte validation (`U4::try_from_slice`) and the `U4`↔`u8` cast (`U4::as_u8_slice`) to `arity-index`, which Miri-tests them.
- Adds a path-scoped Miri CI job that runs the slice-cast tests under `-Zmiri-strict-provenance`, so a layout regression surfaces as UB rather than a silently-passing test.

## How this was tested

- `cargo nextest run --workspace --features ethhash,logger --all-targets` — 737 passed, 16 skipped.
- clippy (default and maxperf profiles, `-D warnings`) and `cargo doc --no-deps` clean.
- `cargo miri test -p firewood-storage path::component` on nightly-2026-07-05 — 15 passed, no undefined behavior or leaks.
- A characterization test guards the `new_pair`/`join` nibble mapping (asymmetric inputs catch an upper/lower swap); a round-trip test covers forward, reverse, reject, and empty-slice conversion.

## Breaking Changes

- `firewood-storage`: `U4` and `TryFromIntError` change identity to re-exported `arity-arrays` types. The crate is beta; direct `U4` consumers are few and mostly reach it through `PathComponent`.

## Follow-up (not blocking)

- The Miri job runs on pull requests but only gates merges once it is added to the branch's required status checks in repository settings.


.github/workflows/ci.yaml
```diff
@@ -381,3 +381,23 @@ jobs:
         run: cargo install cargo-machete
       - name: Check for stale dependencies
         run: cargo machete --with-metadata
+
+  miri:
+    name: miri (path::component slice casts)
+    # NB: this uses macos and not linux intentionally. miri failes due to a dependency
```
RodrigoVillar:
```suggestion
    # NB: this uses macos and not linux intentionally. miri fails due to a dependency
```

.github/workflows/ci.yaml
```diff
@@ -381,3 +381,23 @@ jobs:
...
+      - uses: actions/checkout@v6
+      - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # @master
+        with:
+          toolchain: nightly-2026-07-05
+          components: miri
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # @v2
+        with:
+          shared-key: "miri"
+      - name: Miri (slice reinterpret round-trip)
+        run: cargo miri test -p firewood-storage path::component
```
RodrigoVillar: Would it make sense to add `--locked` here to make sure that the `Cargo.lock` file doesn't drift?

storage/src/path/component.rs
```diff
@@ -7,8 +7,16 @@ use super::{PartialPath, TriePath, TriePathFromUnpackedBytes};
 
 /// A path component in a hexary trie; which is only 4 bits (aka a nibble).
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
 pub struct PathComponent(pub crate::u4::U4);
 
+// `PathComponent` must have `u8` layout for the zero-copy slice conversions below
+// to be sound. `#[repr(transparent)]` over `U4` (itself `#[repr(transparent)]` over
+// a `#[repr(u8)]` enum, guaranteed by arity-index >= 0.1.2) provides it; assert it
+// so a layout regression is a build error, not undefined behavior.
```
RodrigoVillar: I might be reading this wrong, but I don't think the `#[repr(transparent)]` is responsible for these asserts passing (verified by removing the attribute and rebuilding the code).

storage/src/path/component.rs
```diff
@@ -7,8 +7,16 @@ use super::{PartialPath, TriePath, TriePathFromUnpackedBytes};
 
 /// A path component in a hexary trie; which is only 4 bits (aka a nibble).
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(transparent)]
 pub struct PathComponent(pub crate::u4::U4);
```
RodrigoVillar, line 13: (idea) If we derive TransparentWrapper for PathComponent, then we can get rid of the unsafe code that appears down in this file:

```rs
use bytemuck::TransparentWrapper as _;
use bytemuck_derive::TransparentWrapper;

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, TransparentWrapper)]
#[repr(transparent)]
pub struct PathComponent(pub crate::u4::U4);
```

`u4_slice_as_path_components` becomes `PathComponent::wrap_slice(u4s)`, and `path_components_as_byte_slice` reduces to:

```rs
fn path_components_as_byte_slice(components: &[PathComponent]) -> &[u8] {
    crate::u4::U4::as_u8_slice(PathComponent::peel_slice(components))
}
```

That drops this file's unsafe to zero, and both `bytemuck` and `bytemuck_derive` are already dependencies of this crate.

RodrigoVillar (commented):
Mostly LGTM, just a few comments.

storage/src/path/component.rs
demosdemon: I thought I needed to add bytemuck to arity-arrays but it appears I don't. Neat. I didn't know this feature was there. Thanks for teaching me something new.

storage/src/path/component.rs
```diff
@@ -4,11 +4,17 @@
...
+use bytemuck_derive::TransparentWrapper;
 
 /// A path component in a hexary trie; which is only 4 bits (aka a nibble).
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, TransparentWrapper)]
+#[repr(transparent)]
 pub struct PathComponent(pub crate::u4::U4);
 
+const _: () = assert!(size_of::<PathComponent>() == 1);
+const _: () = assert!(size_of::<Option<PathComponent>>() == 1);
```
RodrigoVillar, line 18: (nit) Adding back the comments for why we do these compile-time assertions would be useful.

.github/workflows/ci.yaml
```diff
@@ -431,3 +431,23 @@ jobs:
         run: cargo install cargo-machete
       - name: Check for stale dependencies
         run: cargo machete --with-metadata
+
+  miri:
+    name: miri (path::component slice casts)
+    # NB: this uses macos and not linux intentionally. miri fails due to a dependency
+    # that uses inline assembly (fastrace → fastant → small_ctor) which is not supported
+    # by the Miri interpreter.
+    runs-on: macos-latest
```
RodrigoVillar: (nit) I believe at some point, we switched all our CI jobs to use `macos-26` rather than `macos-latest`.

RodrigoVillar (approved):

PR #2148, opened 2026-07-22
fix(proofs)!: store ProofNode.child_hashes as DenseChildren

### LINKED ISSUE #2099: proof deserialization ~100x memory amplification

demosdemon:
## Why this should be merged

Closes #2099, a P0 proof-deserialization memory-amplification vulnerability: `ProofNode.child_hashes` was a fixed 16-slot `Children<Option<HashType>>` (~500+ bytes per node regardless of occupancy), so a stream of near-empty proof nodes could drive roughly a 100x memory-amplification ratio (a ~50 MB payload allocating multiple GB of heap). This is the second PR of the arity-arrays migration; it builds on PR 1 (the `U4` alias).

## How this works

- `ProofNode.child_hashes` becomes `DenseChildren<HashType>` (`arity_arrays::PackedArray<HashType, Arity16>`): a pointer-sized handle to a heap block sized to exactly the present children, with zero heap allocation for an empty node. Per-node memory is now O(present children) instead of a fixed 16-slot array, so total memory is O(wire bytes), not O(nodes x 16).
- `firewood-storage` gains the `DenseChildren<T>` alias plus `dense_from_children` / `children_from_dense` conversion helpers (temporary; they become native `From` impls in PR 3).
- Deserialization (`proofs/de.rs`) reads the child bitmap and populates a transient, inline `FixedArray` (per-node stack, freed on return), then converts once to the packed representation — a single count-sized allocation, no per-insert reallocation on the DoS-sensitive path.
- The hashing/verification surface stays on `Children`: `Hashable::children()` still returns `Children<Option<HashType>>` and `fix_account_storage_root_value` still takes `&Children<Option<HashType>>`, with `DenseChildren`->`Children` conversion at those seams. Every other consumer (the RLP branch encoding in `eth.rs`, the child-count and present-child walks in `eth_proof.rs` and `merkle/mod.rs`) reads the packed container directly.
- `ChildMask::from_children` (whose sole caller was the old serializer) is removed; the rest of `ChildMask` (the verification/outside-range bitmap) is unchanged.

## How this was tested

- Full workspace suite: 735 passed with `--features ethhash,logger`, 697 passed in the default (SHA-256) hash mode.
- clippy (default and maxperf profiles, `-D warnings`) and `cargo doc --no-deps` clean.
- **The proof wire format is byte-identical.** The proof snapshots (`firewood/src/proofs/snapshots/*`) and the byte-offset tests in `proofs/tests.rs` pass unchanged in BOTH hash modes, with no snapshot regenerated — `DenseChildren`'s little-endian `u16` bitmap and ascending present-child order reproduce the exact bytes the previous `ChildMask` encoding produced.
- New proof gen/verify benchmark: range-proof generation and verification are within noise of the pre-change baseline; single-key proof verification improved ~2.7%.

## Breaking Changes

- `firewood`: `ProofNode.child_hashes` is now `DenseChildren<HashType>` instead of `Children<Option<HashType>>`. `ProofNode` is `#[non_exhaustive]`.
- `firewood-storage`: adds `DenseChildren`, `dense_from_children`, `children_from_dense` to the public surface.
- `firewood` now takes a direct `arity-arrays` dependency (for the deserialization transient); it becomes internal-only again once PR 3 removes the conversion seams.

## Performance note

Single-key proof *generation* regresses ~6.9%: building a proof node now allocates once per non-empty branch (converting the live node's children into the packed form), where the old fixed inline array allocated nothing. This is the inverse of the memory win and is inherent to present-only storage; it is a bounded, one allocation per node (not quadratic), and range-proof generation — the more representative path — stays within noise. A follow-up (a dense-producing `BranchNode::children_hashes()`, or a move-based conversion) can recover it.

### DIFF
--- a/firewood/benches/proofs.rs
+++ b/firewood/benches/proofs.rs
@@ -1,0 +1,78 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! Proof generation and verification throughput.
+//!
+//! Builds a hashed in-memory 16-ary trie of 1000 random 32-byte keys (so branch
+//! nodes near the root are densely populated) and benchmarks single-key and range
+//! proof generation and verification. Guards the `ProofNode.child_hashes`
+//! representation change against throughput regressions.
+
+use std::iter::repeat_with;
+use std::num::NonZeroUsize;
+use std::sync::Arc;
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use firewood::Merkle;
+use firewood::api::DbView as _;
+use firewood::verify_range_proof_structure;
+use firewood_storage::{DeletedNodeTracking, MemStore, NodeStore, SeededRng};
+use rand::{RngExt, distr::Alphanumeric};
+
+#[expect(clippy::unwrap_used, clippy::indexing_slicing)]
+fn bench_proofs(criterion: &mut Criterion) {
+    // Fixture: 1000 random 32-byte keys, 1-byte values, hashed and frozen.
+    let rng = &SeededRng::from_option(Some(1234));
+    let store = Arc::new(MemStore::default());
+    let nodestore = NodeStore::new_empty_proposal(store, DeletedNodeTracking::Enabled);
+    let mut merkle = Merkle::from(nodestore);
+
+    let mut keys: Vec<Vec<u8>> =
+        repeat_with(|| rng.sample_iter(&Alphanumeric).take(32).collect::<Vec<u8>>())
+            .take(1000)
+            .collect();
+    keys.sort();
+    keys.dedup();
+    for key in &keys {
+        merkle.insert(key, Box::new(*b"v")).unwrap();
+    }
+
+    let frozen = merkle.hash();
+    let view = frozen.nodestore();
+    let root_hash = view.root_hash().expect("a non-empty trie has a root hash");
+
+    let first = keys.first().expect("at least one key").as_slice();
+    let last = keys.last().expect("at least one key").as_slice();
+    let mid = keys[keys.len() / 2].as_slice();
+    let limit = NonZeroUsize::new(500);
+
+    let mut group = criterion.benchmark_group("Proofs");
+    group.sample_size(30);
+
+    group.bench_function("single_key_proof/generate", |b| {
+        b.iter(|| view.single_key_proof(mid).unwrap());
+    });
+    group.bench_function("single_key_proof/verify", |b| {
+        let proof = view.single_key_proof(mid).unwrap();
+        b.iter(|| {
+            proof
+                .verify(mid, Some(b"v".as_slice()), &root_hash)
+                .unwrap();
+        });
+    });
+    group.bench_function("range_proof/generate", |b| {
+        b.iter(|| view.range_proof(Some(first), Some(last), limit).unwrap());
+    });
+    group.bench_function("range_proof/verify", |b| {
+        let proof = view.range_proof(Some(first), Some(last), limit).unwrap();
+        b.iter(|| {
+            verify_range_proof_structure(&proof, root_hash.clone(), Some(first), Some(last), limit)
+                .unwrap();
+        });
+    });
+
+    group.finish();
+}
+
+criterion_group!(benches, bench_proofs);
+criterion_main!(benches);
--- a/firewood/src/proofs/ser.rs
+++ b/firewood/src/proofs/ser.rs
@@ -151,7 +151,7 @@ impl WriteItem for ProofNode {
         self.key.write_item(out);
         out.push_var_int(self.partial_len);
         self.value_digest.write_item(out);
-        ChildMask::from_children(&self.child_hashes).write_item(out);
+        out.extend_from_slice(&self.child_hashes.bitmap().to_le_bytes());
         for (_, child) in self.child_hashes.iter_present() {
             child.write_item(out);
         }
--- a/firewood/src/proofs/types.rs
+++ b/firewood/src/proofs/types.rs
@@ -50,9 +50,9 @@
 use crate::proofs::eth::ACCOUNT_DEPTH_NIBBLES;
 use firewood_storage::hash_node_as_storage_trie_root_parts;
 use firewood_storage::{
-    Children, FileIoError, HashType, Hashable, IntoHashType, IntoSplitPath, NibblesIterator, Path,
-    PathBuf, PathComponent, PathIterItem, Preimage, RlpError, SplitPath, TrieHash, TriePath,
-    ValueDigest,
+    Children, DenseChildren, FileIoError, HashType, Hashable, IntoHashType, IntoSplitPath,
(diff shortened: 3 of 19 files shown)


firewood/src/proofs/types.rs
```diff
@@ -310,7 +310,7 @@ pub struct ProofNode {
     /// Otherwise, the node's value or the hash of its value.
     pub value_digest: Option<ValueDigest<Value>>,
     /// The hash of each child, or None if the child does not exist.
```
RodrigoVillar, line 312: (nit) I don't think the None case applies here, unless you're referencing the internals of `DenseChildren` itself: 

```rs
pub struct PackedArray<T, A: Arity>(
    Option<NonNull<Inner<A, T>>>,
    PhantomData<alloc::boxed::Box<T>>,
);
```

firewood/benches/proofs.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use firewood::Merkle;
+use firewood::api::DbView as _;
+use firewood::verify_range_proof_structure;
+use firewood_storage::{DeletedNodeTracking, MemStore, NodeStore, SeededRng};
+use rand::{RngExt, distr::Alphanumeric};
+
+#[expect(clippy::unwrap_used, clippy::indexing_slicing)]
+fn bench_proofs(criterion: &mut Criterion) {
```
RodrigoVillar, line 23: (no need for action) Would it make sense to run this benchmark as a CI job?

firewood/src/proofs/ser.rs
```diff
@@ -151,7 +151,7 @@ impl WriteItem for ProofNode {
         self.key.write_item(out);
         out.push_var_int(self.partial_len);
         self.value_digest.write_item(out);
-        ChildMask::from_children(&self.child_hashes).write_item(out);
+        out.extend_from_slice(&self.child_hashes.bitmap().to_le_bytes());
```
RodrigoVillar, line 154: (idea) The child bitmap is a wire-format contract which the writer/reader must agree on. Before this PR, both directions went through ChildMask, so they couldn't disagree: they were the same code.

With this PR, the write side now gets all three choices from somewhere else - bit order from `arity-bitmap`'s internals, width from the `Arity16` parameter, byte order from u16's inherent `to_le_bytes()` - while the read side (de.rs) still decodes through `ChildMask`. The two implementations agree today, but nothing asserts it; the only guard is the snapshot tests. It also orphans `impl WriteItem for ChildMask` (trait impls escape the dead code linter, and deleting it cascades to `to_le_bytes()`).

Would it make sense to route the write side back through `ChildMask` with a trivial constructor?

```rs
// childmask.rs
/// Builds a mask from a raw bitmap in wire bit order (bit i = child i).
pub(crate) const fn from_bitmap(bits: u16) -> Self {
    Self(bits)
}

// here
ChildMask::from_bitmap(self.child_hashes.bitmap()).write_item(out);
```

`DenseChildren` still owns the data (which children exist), but the encoding of it flows through the one type that documents the format — and the width assumption becomes a compile error instead of a silent break: if `DenseChildren` were ever re-aliased to a wider arity, `bitmap()` would return a `u32`, and today's line would still compile while emitting 4 bytes; `from_bitmap(bits: u16)` makes that fail to build at this exact line.

RodrigoVillar (commented):
Mostly LGTM, just one comment regarding routing both ser/de through `ChildMask`.

firewood/src/proofs/ser.rs
demosdemon: There's no risk of a "silent break" we have snapshot tests that ensure the wire format doesn't change on proofs. I'll give your suggestions some consideration; however, the claim that "nothing asserts it" is false.

firewood/benches/proofs.rs
demosdemon: That already exists here <https://github.com/ava-labs/firewood/blob/main/.github/workflows/benchmarks.yaml>

firewood/src/proofs/types.rs
demosdemon: This comment is technically still true. `child_hashes.get(0)` returns an Option.

firewood/src/proofs/ser.rs
demosdemon: I'd rather not add a wrapper type for what we already verify with snapshot tests. <https://github.com/ava-labs/firewood/blob/main/firewood/src/proofs/snapshot_tests.rs>.

firewood/benches/proofs.rs
RodrigoVillar: Hmm apologies if I wasn't clear here - I meant if we should add `bench_proofs` to the list of benchmarks that we run (i.e. with this PR, the only benchmarks that we run are `hashops`, `defer_persist`, and `serializer`).

RodrigoVillar (approved):

firewood/benches/proofs.rs
demosdemon: I see. I missed that the benchmarks workflow only runs specific benchmarks. That's a footgun waiting to happen.

demosdemon: I'll resolve this with <https://github.com/ava-labs/firewood/pull/2192>.

PR #2149, opened 2026-07-22
refactor(storage)!: back Children with arity_arrays::FixedArray (PR 3/3)

demosdemon:
## Why this should be merged

PRs 1 and 2 moved `U4` and `ProofNode.child_hashes` (the #2099 P0 fix) onto arity-arrays, leaving `Children<T>` as the last hand-rolled container plus a pair of temporary `Children`<->`DenseChildren` conversion helpers (introduced in PR 2 as scaffolding). This PR retires both, so the child containers now have a single backing implementation.

## How this works

- `Children<T>` becomes a newtype over `arity_arrays::FixedArray<T, Arity16>` (was an inline `[T; 16]`). The full public API is reimplemented over FixedArray primitives, so no external call site changes. `iter`/`iter_mut`/`each_ref`/ `each_mut`/`merge` are hand-rolled (FixedArray has no equivalents); the iterator aliases stay a `Zip` of two `array::IntoIter` (built via `core::array::from_fn`) so they remain `Clone` + `DoubleEndedIterator`, which two call sites rely on (`node/mod.rs`, `merkle/changes.rs`).
- PR 2's `dense_from_children`/`children_from_dense` free functions are removed. The `DenseChildren`->`Children` direction becomes a by-reference `From` impl; the `Children`->`DenseChildren` direction becomes an inherent `Children::to_dense` method. A forward `From` impl is orphan-rule-blocked (E0210): the uncovered `T` sits in the foreign `PackedArray` self type ahead of the local `Children`.

## How this was tested

- On-disk and hash-preimage byte formats are unchanged: FixedArray iterators preserve ascending nibble order 0->15, so `BranchNode` serialization and every computed hash stay byte-identical. Node + proof snapshot suites pass with no regeneration in BOTH hash modes (SHA-256 default and Keccak-256 ethhash).
- Full suite: 735/735 (ethhash,logger) and 697/697 (default), all targets. clippy clean (standard + maxperf, both feature sets). cargo doc clean.

## Breaking Changes

- `firewood-storage`: `Children<T>` is no longer `Copy`, and its accessors (`get`/`get_mut`/`replace`/`each_ref`/`each_mut`/`new`/`take`) are no longer `const fn`. No workspace consumer relied on either (verified across all 23 consumer files).
- `firewood-storage`: the `dense_from_children`/`children_from_dense` free functions are removed. They were temporary scaffolding added in PR 2 and never appeared in a released version.

## Follow-ups (optional, not in this PR)

- A `const _: () = assert!(MAX_CHILDREN == <Arity16 as arity_arrays::Arity>::LEN);` would harden the (currently trivially-true) coupling that the iterator reconstruction relies on, at zero runtime cost.
- PR 2's +6.9% single-key proof-generation cost is unchanged here; recovering it needs a dense-producing `BranchNode::children_hashes()`, a separate change.

### DIFF
--- a/firewood/src/proofs/eth.rs
+++ b/firewood/src/proofs/eth.rs
@@ -125,11 +125,8 @@ pub fn proof_node_to_mpt_rlp(node: &ProofNode) -> SmallVec<[Box<[u8]>; 2]> {
     // hasher hashed.
     let inner_bytes: Box<[u8]> = if is_account {
         match value_bytes {
-            Some(v) => fix_account_storage_root_value(
-                v,
-                &firewood_storage::children_from_dense(&node.child_hashes),
-            )
-            .unwrap_or_else(|| Box::from(v)),
+            Some(v) => fix_account_storage_root_value(v, &Children::from(&node.child_hashes))
+                .unwrap_or_else(|| Box::from(v)),
             None => branch_bytes,
         }
     } else {
--- a/firewood/src/proofs/types.rs
+++ b/firewood/src/proofs/types.rs
@@ -373,7 +373,7 @@ impl Hashable for ProofNode {
     }
 
     fn children(&self) -> Children<Option<HashType>> {
-        firewood_storage::children_from_dense(&self.child_hashes)
+        Children::from(&self.child_hashes)
     }
 }
 
@@ -429,7 +429,7 @@ impl From<PathIterItem> for ProofNode {
             key: item.key_nibbles,
             partial_len,
             value_digest,
-            child_hashes: firewood_storage::dense_from_children(&child_hashes),
+            child_hashes: child_hashes.to_dense(),
         }
     }
 }
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -64,7 +64,6 @@ pub use linear::{FileIoError, ReadableStorage, WritableStorage};
 pub use node::path::{NibblesIterator, Path};
 pub use node::{
     BranchNode, Child, Children, ChildrenSlots, DenseChildren, LeafNode, Node, PathIterItem,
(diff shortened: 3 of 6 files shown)


RodrigoVillar (approved):
Mostly LGTM. One comment regarding testing ser/de with distinct children:

The full-branch decode loop (`for (_, child) in &mut children` in `node/mod.rs`) discards the `PathComponent` and relies purely on iteration order - an assumption this PR moves from `[T; 16]` semantics onto `arity-arrays`' iterator contract. But every full-branch fixture in `test_serialize_deserialize` gives all 16 children identical hashes/addresses, so the round-trip `assert_eq!` would pass even if the children came back permuted (the `.expect()`s only guard count, not order).

Making one fixture slot-distinguishable pins this with no new test machinery:

```
 #[test_case(Node::Branch(Box::new(BranchNode {
     partial_path: Path::from(vec![0, 1, 2, 3]),
     value: Some(vec![4, 5, 6, 7].into()),
-        children: Children::from_fn(|_|
-                Some(Child::AddressWithHash(LinearAddress::new(1).unwrap(), std::array::from_fn::<u8, 32, _>(|i| i as u8).into()))
+        // Each child's hash starts with its own nibble so a decode that pairs
+        // slots with the wrong wire entries fails the round-trip comparison.
+        children: Children::from_fn(|pc|
+                Some(Child::AddressWithHash(LinearAddress::new(1).unwrap(), std::array::from_fn::<u8, 32, _>(|i| if i == 0 { pc.as_u8() } else { i as u8 }).into()))
     )})), 652; "full branch node with long partial path and value"
 )]
```

Only byte values change, not widths, so the 652 expected length is untouched.

PR #2150, opened 2026-07-22
refactor(storage): fold HashOrRlp into HashType [4/9]

RodrigoVillar:
## Why this should be merged

After `HashType` becomes the shared hash-or-inline-RLP representation, keeping the older `HashOrRlp` type and the custom `IntoHashType` conversion trait leaves two names and two conversion paths for the same concept.

In the stack, this removes transitional duplication created by reaching the shared representation in PR 3. It leaves one hash-reference type and standard conversion path for the proof and typed-database work in PRs 5 and 6, while remaining a focused API cleanup rather than mixing those higher-level changes into the representation PR.

## How this works

This change moves the remaining `HashOrRlp` implementation directly onto`HashType`, removes the duplicate type and module, and replaces `IntoHashType` call sites with standard `From`/`Into` conversions or direct `HashType` values. It also moves `TrieHash` into its own mode-independent module so digest bytes remain distinct from the representation used for child references.

## How this was tested

CI.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


demosdemon (approved):

RodrigoVillar:
Rebased to handle merge conflicts related to the `DenseChildren` PRs which got merged in.

PR #2161, opened 2026-07-30
docs: establish the design-doc subsystem with date-prefixed filenames

demosdemon:
First of a stack that breaks the mdBook content branch into one pull request per document. This one carries the structural pieces every later PR builds on: the design-doc naming convention, the subsystem's README and template, and the generated-docs link-outs.

## Why this should be merged

The scaffolded site has a design-doc directory holding exactly one design and no documented convention for adding another. This lands the convention, the tooling that implements it, and the RFC-style template, so subsequent designs can be proposed without re-litigating layout.

## How this works

- Designs are named `YYYY-MM-DD-slug.md`, replacing the monotonic `NNNN-slug.md` scheme. A date needs no coordination between concurrent authors, and it is self-describing. The one remaining collision (same slug, same day) is caught by `just new-design` refusing to overwrite.
- The filename records when a design was *proposed*; git history records how *fresh* it is (`just design-age`). A hand-maintained date inside the document is still refused, because only that one can drift. Promotion stays a `status` flip, so the date never means "last touched".
- `category` accepts a list for designs spanning subsystems. Drawbacks survives promotion; only Unresolved questions is proposal-only, since the record of what was accepted is as useful after implementation as before.
- The old `[0-9][0-9][0-9][0-9]-*.md` glob in `design-doc-age.sh` would have matched date-prefixed names by accident, so the tightened glob is a correctness fix, not just prose.
- Generated-docs links live on `reference/README.md` as fully-qualified URLs. The design doc previously claimed `SUMMARY.md` accepts raw-URL sidebar entries; it does not, and a raw URL there fails the build. Full URLs resolve from local `mdbook serve` and forks, and `follow-web-links = false` already skips them, so no `linkcheck2` exclude is needed.
- The design doc is `active` while the site lands one page per PR, so it carries a rollout note and each PR in the stack ticks its own acceptance boxes.
- `AGENTS.md` drops its paragraph directing agents to a local authoring plugin: that tooling is not part of this repository.

## How this was tested

- `just book-build` (mdbook 0.5.3 + mermaid + linkcheck2) builds cleanly; `just ci-lint-markdown` and `shellcheck` pass.
- `just new-design <slug>` scaffolds a dated file and exits non-zero on a second run with the same slug.
- The frontmatter preprocessor strips the renamed design's and the template's YAML blocks; the rendered `<main>` opens on the H1.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,32 @@ The `fwdctl` tool provides command-line operations on databases. See `fwdctl/REA
 
 For more information on coding conventions and constraints, please refer to [CONTRIBUTING.md](./CONTRIBUTING.md)
 
+## Design Documents
+
+Firewood records designs as living documentation in a single flat directory,
+`docs/src/designs/` (rendered in the mdBook site). Every design is
+`YYYY-MM-DD-<slug>.md`, where the date is the day the design was written; it is
+assigned once and never changes. A design's lifecycle state lives in the `status`
+frontmatter field, not in its location or filename. Follow this workflow when writing
+a design document:
+
+- **Propose:** run `just new-design <slug>` to scaffold
+  `docs/src/designs/YYYY-MM-DD-<slug>.md` (with `status: proposed`) from the RFC-style
+  template, fill it in, and open a pull request for review.
+- **Promote:** once the design is implemented, promote it in place — a status flip,
+  not a file move, so its `YYYY-MM-DD-<slug>.md` path stays stable and its date prefix
+  goes on recording when the design was proposed. Following the checklist in
+  `docs/src/designs/README.md`: flip the frontmatter `status` to `active`, drop the
+  proposal-only sections, rewrite future-tense prose into the present tense, add
+  cross-links to the implementing PR(s), and update the index in
+  `docs/src/designs/README.md` and `docs/src/SUMMARY.md`.
+- `active` designs describe what the code does today; keep them up to date as the
+  design changes.
+
+Never write a date into a design's frontmatter or body. The filename records when the
+design was proposed; `just design-age` reports how fresh each design is from git
+history. A hand-maintained date inside a document only drifts out of sync.
+
 ## Commit and PR Title Convention
 
 Commit messages and PR titles **must** follow the
--- a/docs/src/designs/README.md
+++ b/docs/src/designs/README.md
@@ -1,0 +1,76 @@
+# Design Documents
+
+Firewood records its designs as living documentation in a single flat directory.
+Every design is `YYYY-MM-DD-slug.md` — the date the design was first written, followed
+by a short slug. The date is assigned once, when the file is created, and never
+changes afterwards: it records when the design was *proposed*, not when it was last
+touched, so a design's path stays a stable reference in review threads and
+cross-links. Two designs written on the same day are distinguished by their slugs.
+
+A design's lifecycle state lives in its `status` frontmatter field, not in its
+location:
+
+- **`proposed`** — under review, not yet built. A proposal typically lands before
+  development so reviewers comment on the document through normal pull-request review.
+- **`active`** — reflects what the code does today; a living document, updated as the
+  code evolves.
+- **`draft`**, **`superseded`**, **`rejected`** — a work in progress, a design replaced
+  by a newer one, or a design considered and declined; each kept for the record.
+
+> [!NOTE]
+> This is a lightweight convention, not a mandatory gate. A `proposed` design is
+> encouraged for significant or non-obvious work, where writing it down sharpens the
+> discussion; small or low-risk changes do not need one, and nothing blocks a pull
+> request for lacking one. An `active` design may also be written after the fact to
+> document something already built. Optimize for "more designs discussed and
+> recorded," not "process followed."
+
+## Designs
+
+| Design | Proposed | Status |
+| --- | --- | --- |
+| [mdBook documentation site](2026-06-17-mdbook-documentation-site.md) | 2026-06-17 | active |
+
+## Designs still to be written
+
+Several core subsystems are documented only in code, not yet as designs. The
+backlog — revision management, hashing, proposals & commits, state sync &
+reconstruction, and the Go FFI layer — is tracked in
+[ava-labs/firewood#2139](https://github.com/ava-labs/firewood/issues/2139).
+
+## Frontmatter
+
+Every design begins with a YAML frontmatter block, stripped from the rendered site by
+the in-repo `frontmatter-strip` preprocessor:
+
+| Field | Required | Values |
+| --- | --- | --- |
+| `title` | yes | Human-readable design title. |
+| `status` | yes | `draft`, `proposed`, `active`, `superseded`, or `rejected`. |
+| `category` | yes | A Firewood subsystem: `storage`, `ffi`, `revision-management`, `hashing`, `proposals`, `docs`, or `tooling`. |
+| `authors` | yes | List of GitHub handles. |
+| `tracking-issue` | no | An `owner/repo#N` reference to the issue or PR tracking the work. |
+
+The schema has **no date fields**. A design's proposal date is its filename prefix,
+fixed when the file is created; its freshness comes from git history. Neither is a
+hand-maintained date inside the document, which would only drift out of sync with
+reality. Run `just design-age` to list designs by last git-commit date, oldest first,
+to spot stale documents.
+
+## Propose a design
+
+Run `just new-design <slug>` to scaffold `YYYY-MM-DD-<slug>.md` — today's date plus the
+slug — from `template.md`, fill it in, and open a pull request.
+
+## Promote a design
+
+When a proposed design is implemented, promote it in place — a status flip, not a file
+move, so its `YYYY-MM-DD-slug.md` path stays stable and its date prefix goes on
+recording when the design was proposed:
+
+1. Flip the frontmatter `status: proposed` to `status: active`.
+2. Drop the proposal-only sections (Drawbacks, Unresolved questions), folding any
+   surviving content into the living structure.
+3. Rewrite any future-tense prose as present tense — the design now describes reality.
+4. Add cross-links to the implementing pull request(s) and commits.
+5. Update this index's status marker and `SUMMARY.md`.
--- a/docs/src/designs/template.md
+++ b/docs/src/designs/template.md
@@ -1,0 +1,48 @@
+---
+title:
+status: proposed
+category:
+authors: []
+tracking-issue:
+---
+
+# \<Title\>
+
+## Summary
+
+One paragraph describing the change.
+
+## Motivation
+
+Why this is worth doing; the problem it solves.
+
+## Guide-level explanation
+
+Explain the design as if teaching it to another contributor.
+
+## Detailed design
+
+The technical detail: data structures, algorithms, on-disk/runtime behavior,
+edge cases.
+
+## Drawbacks
+
+Reasons this design might not be worth doing. (Proposal-only; drop when promoting to
+`active`.)
+
+## Rationale and alternatives
+
+Why this design over others; what alternatives were considered.
+
+## Prior art
+
+Comparable designs in other systems, if any.
+
+## Unresolved questions
+
+Open questions to settle during review. (Proposal-only; drop when promoting to
+`active`.)
+
+## Future possibilities
+
+Natural extensions, explicitly out of scope here.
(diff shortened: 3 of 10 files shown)


docs/src/designs/README.md
```diff
@@ -0,0 +1,76 @@
...
+cross-links. Two designs written on the same day are distinguished by their slugs.
+
+A design's lifecycle state lives in its `status` frontmatter field, not in its
+location:
+
+- **`proposed`** — under review, not yet built. A proposal typically lands before
+  development so reviewers comment on the document through normal pull-request review.
+- **`active`** — reflects what the code does today; a living document, updated as the
+  code evolves.
+- **`draft`**, **`superseded`**, **`rejected`** — a work in progress, a design replaced
```
rkuris, line 17: nit: kinda strange, why are these three on separate lines but the other two have verbose descriptions?

nit: They also all feel a bit Ai-wordy but it's not terrible.

docs/src/designs/README.md
```diff
@@ -0,0 +1,76 @@
...
+| --- | --- | --- |
+| [mdBook documentation site](2026-06-17-mdbook-documentation-site.md) | 2026-06-17 | active |
+
+## Designs still to be written
+
+Several core subsystems are documented only in code, not yet as designs. The
+backlog — revision management, hashing, proposals & commits, state sync &
+reconstruction, and the Go FFI layer — is tracked in
+[ava-labs/firewood#2139](https://github.com/ava-labs/firewood/issues/2139).
+
```
rkuris, line 40: This should be left for the issue, not checked in here.

Action: Remove this block or comment as to why you think it should be duplicated here.

docs/src/designs/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Frontmatter
+
+Every design begins with a YAML frontmatter block, stripped from the rendered site by
+the in-repo `frontmatter-strip` preprocessor:
+
+| Field | Required | Values |
+| --- | --- | --- |
+| `title` | yes | Human-readable design title. |
+| `status` | yes | `draft`, `proposed`, `active`, `superseded`, or `rejected`. |
+| `category` | yes | A Firewood subsystem: `storage`, `ffi`, `revision-management`, `hashing`, `proposals`, `docs`, or `tooling`. |
```
rkuris, line 50: I can see proposals that span multiple categories: perhaps both revision management and storage, hashing and storage, etc. How is that handled?

Action: To resolve this comment, please answer the question or provide a mechanism for having more than one, then resubmit for review.

docs/src/designs/README.md
```diff
@@ -0,0 +1,76 @@
...
+| `authors` | yes | List of GitHub handles. |
+| `tracking-issue` | no | An `owner/repo#N` reference to the issue or PR tracking the work. |
+
+The schema has **no date fields**. A design's proposal date is its filename prefix,
+fixed when the file is created; its freshness comes from git history. Neither is a
+hand-maintained date inside the document, which would only drift out of sync with
+reality. Run `just design-age` to list designs by last git-commit date, oldest first,
+to spot stale documents.
```
rkuris, line 58: AI wordy
```suggestion
There's no date field because that's included in the filename.
```

I'm not sure we should be cross-referencing the `just design-age` tooling here, especially since there's isn't actually a date field. Perhaps that could be mentioned when we talk about the filename choice though.

Action: Accept this change or explain why you think the extra documentation has value, then resubmit.

docs/src/designs/README.md
```diff
@@ -0,0 +1,76 @@
...
+## Propose a design
+
+Run `just new-design <slug>` to scaffold `YYYY-MM-DD-<slug>.md` — today's date plus the
+slug — from `template.md`, fill it in, and open a pull request.
+
+## Promote a design
+
+When a proposed design is implemented, promote it in place — a status flip, not a file
+move, so its `YYYY-MM-DD-slug.md` path stays stable and its date prefix goes on
+recording when the design was proposed:
```
rkuris, line 69: Step 1 repeats exactly this paragraph in a much more succinct way, so delete this intro. The only possible value here is that you're explaining *why* we don't change the filename, but that was (a) discussed earlier, and (b) doesn't belong as part of the code, but could be part of the commit comments.

Action: Remove this paragraph or simplify it, then resubmit.

docs/src/designs/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Flip the frontmatter `status: proposed` to `status: active`.
+2. Drop the proposal-only sections (Drawbacks, Unresolved questions), folding any
+   surviving content into the living structure.
+3. Rewrite any future-tense prose as present tense — the design now describes reality.
```
rkuris, line 74: Remove the em-dash and the why here.

docs/src/designs/template.md
```diff
@@ -0,0 +1,48 @@
+---
+title:
+status: proposed
+category:
+authors: []
+tracking-issue:
+---
+
+# \<Title\>
```
rkuris, line 9: Why is only `Title` using the greater-than and less-than tags? Is the title supposed to have those markers for some reason?

Action: Answer the question, then resubmit for review.

docs/src/designs/template.md
```diff
@@ -0,0 +1,48 @@
+---
+title:
+status: proposed
+category:
+authors: []
```
rkuris, line 5: An example would be good here, which can avoid the verbose description elsewhere.
```suggestion
authors: ['github-handle-1', 'github-handle-2']
```

docs/src/designs/template.md
```diff
@@ -0,0 +1,48 @@
...
+One paragraph describing the change.
+
+## Motivation
+
+Why this is worth doing; the problem it solves.
+
+## Guide-level explanation
+
+Explain the design as if teaching it to another contributor.
+
```
rkuris, line 22: nit: not sure this is required. I think this section can be listed as "optional"

docs/src/designs/template.md
```diff
@@ -0,0 +1,48 @@
...
+## Detailed design
+
+The technical detail: data structures, algorithms, on-disk/runtime behavior,
+edge cases.
+
+## Drawbacks
+
+Reasons this design might not be worth doing. (Proposal-only; drop when promoting to
+`active`.)
+
```
rkuris, line 32: nit: I'm not sure we should drop the drawbacks we considered during the original design. Can you provide an example of why you'd not want to keep a record of this post-implementation?

I can come up with counter examples: let's say someone proposes making the checker run in parallel, which uses more CPU. Probably should retain the fact that this decision was made despite this drawback instead of dropping it completely.

Not feeling very strong about this one, so no action is required (hence the nit) but it does feel strange to drop the record of drawbacks for implemented solutions.

docs/src/reference/README.md
```diff
@@ -0,0 +1,9 @@
+# Reference
+
+Generated API documentation and dashboards. These are site-absolute paths that
+resolve only on the deployed site; under local `mdbook serve` (or a fork's Pages
+deployment) they 404.
```
rkuris, line 5: The `introduction.md` file takes the opposite approach. of including full links. Let's pick one approach and apply it to both this file and the introduction document.

In fact, this file might be duplicative to what's already in introduction.md.

rkuris (changes requested):

docs/src/designs/template.md
demosdemon: Angle brackets mark values the author replaces; the section headings are fixed. The frontmatter now uses the same convention (`title: <Title>`, `category: <subsystem>`, `authors: [<github-handle-1>, <github-handle-2>]`, `tracking-issue: <owner/repo#N>`), so the H1 is no longer the odd one out.

docs/src/designs/README.md
demosdemon: `category` now accepts a list for designs spanning subsystems, e.g. `category: [storage, hashing]`. Documented in the README table, the design doc's schema table, and the template.

docs/src/designs/template.md
demosdemon: Agreed; the parallel-checker example is exactly the case. Drawbacks now survives promotion and only Unresolved questions is proposal-only. Updated the template, README checklist, design doc, and renamed this design's own "Accepted trade-off" section to Drawbacks.

AGENTS.md
```diff
@@ -119,6 +119,24 @@ The `fwdctl` tool provides command-line operations on databases. See `fwdctl/REA
 
 For more information on coding conventions and constraints, please refer to [CONTRIBUTING.md](./CONTRIBUTING.md)
 
+## Design Documents
```
AminR443, line 122: Last bullet from PR descriptions talks about dropping a paragraph from `AGENTS.md`, but you only add design docs section to it. I'm thinking PR description is out of date.

AGENTS.md
```diff
@@ -119,6 +119,24 @@ The `fwdctl` tool provides command-line operations on databases. See `fwdctl/REA
...
+
+Firewood records designs as living documentation in `docs/src/designs/`, rendered in
+the mdBook site. Every design is `YYYY-MM-DD-<slug>.md`, dated when it was proposed;
+the filename never changes. Lifecycle state lives in the `status` frontmatter field.
+The convention, frontmatter schema, and promotion checklist are documented in
+`docs/src/designs/README.md`.
+
+- **Propose:** run `just new-design <slug>` to scaffold a `status: proposed` design
```
AminR443, line 130: nit: do we really need to mention these in here? I mean if we're already giving a reference to `docs/src/designs/README.md`, most probably agent will read that file anyway and we won't be saving much tokens.

docs/src/designs/README.md
```diff
@@ -0,0 +1,64 @@
...
+last commit date, oldest first.
+
+A design's lifecycle state lives in its `status` frontmatter field, not in its
+location:
+
+- **`draft`**: a work in progress, not yet ready for review.
+- **`proposed`**: under review, not yet built.
+- **`active`**: describes what the code does today; kept up to date as the code evolves.
+- **`superseded`**: replaced by a newer design; kept for history.
+- **`rejected`**: considered and declined; kept for the record.
```
AminR443, line 17: Does "protoyped with AI" deserve to be a separate status in here? definitely we're not gonna merge in the code in that stage, but could help separate "implemented" and "vibe-coded". I'm mentioning this since experimenting with AI is faster, and could help rule out/promote a certain design proposal before diving deep.

docs/src/designs/template.md
```diff
@@ -0,0 +1,48 @@
...
+records what was accepted and why.
+
+## Rationale and alternatives
+
+Why this design over others; what alternatives were considered.
+
+## Prior art
+
+Optional. Comparable designs in other systems.
+
```
AminR443, line 40: Can we add a dedicated optional “Evaluation” section? 

It could already live under Detailed Design, but I think it would be useful to have a place to discuss the design’s impact, why we chose certain parameters, and what the experiments show once it’s implemented/proposed. We could also include evidence for ruling out the alternatives we considered.

docs/src/designs/README.md
```diff
@@ -0,0 +1,64 @@
...
+## Promote a design
+
+When a proposed design is implemented, promote it in place:
+
+1. Flip the frontmatter `status: proposed` to `status: active`.
+2. Drop the Unresolved questions section, folding any settled answers into the design.
+   Keep Drawbacks: it records what was accepted and why.
+3. Rewrite future-tense prose as present tense.
+4. Add cross-links to the implementing pull request(s) and commits.
+5. Update this index's status marker and the sidebar entry in `docs/src/SUMMARY.md`.
```
AminR443, line 64: shouldn't we add an entry to table or summary when proposing as well? why just when promoting? I think the section above this should mention it.

justfile
```diff
@@ -341,8 +341,23 @@ book-serve: book-assets
...
+    out="$dir/$(date +%F)-{{slug}}.md"
+    if [[ -e "$out" ]]; then
+        # Two designs with the same slug on the same day collide; refuse rather than clobber.
+        echo "error: $out already exists" >&2
+        exit 1
+    fi
+    cp "$tmpl" "$out"
+    echo "Created $out"
+
+# List design docs by last git-commit date (oldest/stalest first)
```
AminR443, line 360: I think it'd be a nice idea to validate `slug` and have a convention for it (i.e. lowercase kebab-case, no spaces, no slashes, no underscores). 

I also think that there is a possibility that one can execute shell commands if they include quotes in slug. This could potentially escape checks and approve commands if someone adds sth like `just new-design *` in their agent allowlist.

AminR443 (approved):

PR #2171, opened 2026-07-30
perf(storage): box Child::Node variant to reduce BranchNode footprint

### LINKED ISSUE #2034: Branch nodes are large
**Describe the bug**
Branch Nodes have a footprint of ~1.6kib for merkledb and 1.7kib for ethhash.

A worker configured to hold 100,000 recent revisions pays ≈188 MB just for the in-memory representation of each root node — roughly 1888 bytes per node, of which 1792 bytes are the 16-slot child array. The slot is sized for the widest `Child::Node(Node)` variant even though a cached, persisted branch never contains that variant. Separately, the `Arc<Mutex<MaybePersisted>>` indirection that the `Child::MaybePersisted` variant relies on to communicate post-freeze address updates from the persist worker to the parent's serializer is structurally heavy and used outside its narrow window in ways that obscure the invariant.

**To Reproduce**

**Expected behavior**

**Possible Solutions**
If known, give some ideas of what may be done to fix the problem.

**Additional context**

galadd:
## Why this should be merged

Closes #2034 

Currently, BranchNode is 1624 bytes because the Child enum reserves 96 bytes per slot to fit the unboxed Node variant. Long-lived cached nodes pay this penalty even though they only ever contain AddressWithHash. Boxing the variant drops the enum to 48 bytes and BranchNode to 856 bytes (-47%), which significantly improves cache locality during trie traversal.

## How this works

Changed Child::Node(Node) to Child::Node(Box<Node>). This is a purely mechanical layout change; it doesn't alter ownership semantics, the persistence state machine, or the public API.

`size_of` output before/after:
Child:    96 -> 48 bytes
Branch: 1624 -> 856 bytes

## How this was tested

Existing firewood-storage tests pass.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


rkuris (changes requested):
This adds an additional allocation on the hot path (`read_for_update` now unboxes, `insert_helper` then recurses on an owned `Node`, and re-boxes. Performance testing is likely to show the problem more clearly, but I think this needs a different design.

The build also fails with ethhash enabled; please check the local test instructions before resubmitting.

rkuris:
Additionally let's add some assertions so the build fails if these critical sizes are changed: `const _: () = assert!(size_of::<Child>() == 48);`

rkuris:
Moved to draft to avoid discussions during grooming today

galadd:
Hello @rkuris, thanks for reviewing. Sorry for the gap, I was unavailable this past week. I have fixed the build error and added the assertion

On the allocation regression you flagged: You are right. I measured it with a counting global allocator across a 50k-insert uncommitted-proposal workload, and confirmed the original patch introduced a real hot-path problem:

| Version                                                           | Allocs | Time |
| ---------------------------------------------------------- | --------- | ------- |
| main (unboxed) | ~269k | ~152ms |
| Initial PR (unbox/rebox per level) | ~522k (+94%) | ~195ms |
| Revised (in-place &mut Node mutation) | ~341k (+27%) | ~159ms |

Note: wall-clock numbers are still noisy across runs. Allocation counts are the more stable signal here.

The revised version changes insert_helper to take &mut Node and mutate through the existing Box instead of moving nodes out and reconstructing new boxes on the way back up the recursion. 

The residual ~72k-over-baseline is not traversal churn; I traced it to something structural: every new child placement (leaf splits, new-ancestor inserts) now costs one Box::new per slot, since Child::Node no longer holds a Node inline. That's an inherent cost of the memory-shrink design itself, not leftover inefficiency in the traversal code. 

Happy to push the in-place-mutation diff for review now, or explore closing the remaining gap further (e.g. a small pooled allocator for child boxes) if you'd rather not accept that trade-off. Let me know which you'd prefer before I open the updated PR.

PR #2156, opened 2026-07-28
feat(fwdctl): support hex and Ethereum keys

wenn-id:
## Summary

- add generic `--hex` key decoding to `fwdctl get`, `insert`, and `delete`
- add feature-gated `--account` and `--storage <SLOT>` modes for Ethereum key derivation
- validate hex input lengths, reject conflicting modes, and document the CLI contract
- add integration coverage for round trips, Keccak-derived keys, malformed input, conflicts, and deletion

## Key derivation

- account: `keccak256(address)`
- storage: `keccak256(address) || keccak256(slot)`

Closes #1000.

## Validation

- `cargo fmt --check`
- `cargo test -p firewood-fwdctl --test cli` (19 passed)
- `cargo test -p firewood-fwdctl --features ethhash,logger --all-targets` (25 passed)
- `cargo nextest run --workspace --features ethhash,logger --all-targets` (824 passed, 17 profile-skipped)
- `cargo +nightly-2026-07-05 clippy -p firewood-fwdctl --features ethhash,logger --all-targets -- -D warnings`
- `cargo +nightly-2026-07-05 clippy --profile maxperf -p firewood-fwdctl --features ethhash,logger --all-targets -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p firewood-fwdctl --features ethhash,logger --no-deps`
- `markdownlint-cli2 fwdctl/README.md`

PR #2172, opened 2026-07-31
feat(fwdctl): accept hex and Ethereum key inputs

### LINKED ISSUE #1000: feat(fwdctl): add support for hex input strings and hashing eth accounts
For most users, it seems more helpful to provide hex strings as input for keys. Especially in the ethhash case, it would be best if we could provide accounts (20 bytes) and storage keys (32 bytes) and hash them (note a storage key is account hash, storage key hash, concatenated).

tomatotomata:
Closes #1000

## Summary

- add one shared key selector for `fwdctl get`, `insert`, and `delete`
- accept UTF-8 keys as before, raw hex keys with an optional `0x` prefix, and Ethereum account/storage inputs
- hash 20-byte accounts and 32-byte storage slots with Keccak256, composing storage keys as `keccak256(account) || keccak256(storage_key)`
- validate selector combinations and exact Ethereum input lengths before opening the database
- add CLI round-trip coverage for a binary key and document the new forms

## Validation

- `cargo fmt --check` passes
- `git diff --check` passes
- focused Rust tests were attempted with and without `ethhash`; this Windows checkout cannot compile the existing storage crate because it imports `std::os::unix::fs::FileExt`
- no diagnostics were emitted for the changed `fwdctl` files

The legacy `insert KEY VALUE`, `get KEY`, and `delete KEY` forms remain supported. The change is limited to the CLI key boundary and its tests. The commit is unsigned because this environment has no configured signing key; I can amend if the repository requires a specific signing setup.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1913,6 +1913,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml_ng",
+ "sha3",
  "tempfile",
  "thiserror",
  "tokio",
--- a/fwdctl/Cargo.toml
+++ b/fwdctl/Cargo.toml
@@ -32,6 +32,7 @@ firewood-storage.workspace = true
 hex.workspace = true
 log.workspace = true
 nonzero_ext.workspace = true
+sha3 = { version = "0.12.0", optional = true }
 # Regular dependencies
 csv = "1.4.0"
 humantime = "2.4.0"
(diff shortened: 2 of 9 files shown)


demosdemon:
Hi there. Apologies but I'm closing this in favor of <https://github.com/ava-labs/firewood/pull/2156>.

Issue #2151, opened 2026-07-23
feat: Avoid padding when nodes cannot be deleted


rkuris:
### Problem

Some database configurations (notably "archive mode") never prune anything. This means we never free anything: the free lists are always empty. Therefore, we don't need to pad each node to the nearest AREA_SIZE. We don't need to pad at all, actually. It's fine if the next block starts at an odd address.

To avoid corrupting things, we should designate a sentinel AREA_SIZE (perhaps 255) indicating that this node can never be freed.

This should save considerable space when running in archival mode, particularly for larger nodes like branch nodes.

This adds considerable complexity to the checker, but the space savings probably warrant doing this.

### Proposed solution

Don't pad

galadd:
Implemented this across two PRs:

**#2173** - Reader foundation
- Added `SENTINEL_UNPADPED` (0xFF) and `AreaMetadata` type
- Updated read paths to handle both padded and unpadded formats

**#2175** - Full implementation (this PR)
- Writer: unpadded serialization when `DeletedNodeTracking::Disabled`
- Checker: skips alignment checks for unpadded nodes, tracks unpadded stats
- fwdctl: new "Unpadded (Archive Mode)" section in check report

PR #2173, opened 2026-07-31
Migrate area_index_and_size to AreaMetadata

galadd:
## Why this should be merged

Foundation for #2151. Adds reader support for unpadded node format without changing any existing behavior.

## How this works

Nodes on disk currently start with an AreaIndex byte (0-22) indicating their padded size class. This PR adds support for a SENTINEL_UNPADDED byte (0xFF) that indicates an unpadded node followed by a varint length prefix. The area_index_and_size() function and node readers now handle both formats.

Since no existing code writes the sentinel byte, all current databases continue to work unchanged through the padded code path.

## How this was tested

-  Added unit tests for varint_encoded_size() and SENTINEL_UNPADDED collision check
-  Added unit test for AreaMetadata struct
-  All existing tests pass (no behavior change)

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/storage/src/checker/mod.rs
+++ b/storage/src/checker/mod.rs
@@ -295,14 +295,21 @@ where
         check_area_aligned(subtrie_root_address, StoredAreaParent::TrieNode(parent))?;
 
         // read the node from the disk - we avoid cache since we will never visit the same node twice
-        let (area_index, area_size) =
-            self.area_index_and_size(subtrie_root_address)
-                .map_err(|e| {
-                    vec![CheckerError::IO {
-                        error: e,
-                        parent: StoredAreaParent::TrieNode(parent),
-                    }]
-                })?;
+        let area_meta = self
+            .area_index_and_size(subtrie_root_address)
+            .map_err(|e| {
+                vec![CheckerError::IO {
+                    error: e,
+                    parent: StoredAreaParent::TrieNode(parent),
+                }]
+            })?;
+
+        // TODO(galadd): will send a PR after  this to add proper unpadded handling here
+        // not included in this so that code changes won't be too large
+        // For now, all existing databases are padded
+        let area_index = area_meta.area_index.expect("padded node expected");
+        let area_size = area_meta.total_size;
+
         let (node, node_bytes) = self
             .read_node_with_num_bytes_from_disk(subtrie_root_address)
             .map_err(|e| {
@@ -664,13 +671,22 @@ where
 
         // First attempt to read the valid stored areas from the leaked range
         loop {
-            let (area_index, area_size) = match self.read_leaked_area(current_addr) {
+            let area_meta = match self.read_leaked_area(current_addr) {
                 Ok(area_index_and_size) => area_index_and_size,
                 Err(e) => {
                     warn!("Error reading stored area at {current_addr}: {e}");
                     break;
                 }
             };
+
+            // TODO(galadd): will send a PR after  this to add proper unpadded handling here
+            // not included in this so that code changes won't be too large
+            // For now, all existing databases are padded
+            let Some(area_index) = area_meta.area_index else {
+                warn!("Unpadded node in leaked range at {current_addr}, cannot add to free list");
+                break;
+            };
+            let area_size = area_meta.total_size;
 
             let next_addr = current_addr
                 .advance(area_size)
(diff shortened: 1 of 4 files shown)


storage/src/checker/mod.rs
```diff
@@ -295,14 +295,21 @@ where
...
+            .map_err(|e| {
+                vec![CheckerError::IO {
+                    error: e,
+                    parent: StoredAreaParent::TrieNode(parent),
+                }]
+            })?;
+
+        // TODO(galadd): will send a PR after  this to add proper unpadded handling here
+        // not included in this so that code changes won't be too large
+        // For now, all existing databases are padded
```
rkuris, line 309: This might have been okay if we returned an error. firewood is designed to be panic free and it's possible we might release with just this half of the fix, which would be really bad.

Please figure out how to reverse these two commits. That is, make a change such that we panic in the event we find a padded node and have the framework needed to determine if the node is actually padded or not. Even better, avoid the panic altogether and return an error.

storage/src/checker/mod.rs
```diff
@@ -295,14 +295,21 @@ where
...
+        let area_index = area_meta.area_index.expect("padded node expected");
+        let area_size = area_meta.total_size;
+
```
rkuris, line 312: Why can't we just leave the two-variable assignment from earlier and just adjust area_index accordingly, something like:
```
let area_index = area_index.ok_or(CheckerError::...)?;
```

storage/src/checker/mod.rs
```diff
@@ -664,14 +671,23 @@ where
 
         // First attempt to read the valid stored areas from the leaked range
         loop {
-            let (area_index, area_size) = match self.read_leaked_area(current_addr) {
+            let area_meta = match self.read_leaked_area(current_addr) {
```
rkuris, line 674: Same as previous comment. Since this code seems duplicative to the prior block anyway, perhaps a helper might also make sense.

storage/src/nodestore/alloc.rs
```diff
@@ -263,8 +262,8 @@ impl<'a, S: ReadableStorage> NodeAllocator<'a, S> {
     pub(crate) fn area_index_and_size(
         &self,
         addr: LinearAddress,
-    ) -> Result<(AreaIndex, u64), FileIoError> {
-        area_index_and_size(self.storage, addr)
+    ) -> Result<super::AreaMetadata, FileIoError> {
```
rkuris, line 265: This solidifies my prior comment: just do a PR that changes the return type of area_index_and_size to return an AreaMetadata.

I stopped my review here because this PR should be done in smaller parts and I'm sure to have some comments on just the migration to the new `AreaMetadata` type.

rkuris (changes requested):

storage/src/nodestore/alloc.rs
galadd: Hello @rkuris, thanks for the review. I've changed the PR into just the migration to `AreaMetadata` type. Should I edit the first comment (PR description), so it's not misleading or should I just add a new description at the bottom of this

rkuris: Thanks for doing this! But when I look at this now, you're creating a new type, `AreaMetadata` whose only purpose is to "cache" the total area size. Since an AreaIndex already has an O(1) method to do exactly this, I'm wondering why this is needed and we can't just use `AreaIndex`.

rkuris (changes requested):
I'm not yet convinced this new type is needed as it seems to add little value on top of the existing AreaIndex type.

Also it seems to have a conflict with main at this point that must be resolved.

storage/src/nodestore/alloc.rs
galadd: It is really just in preparation for an eventual fix. I'm trying to resolve #2151, and I intend on adding an is_unpadded to an `Area` info. So it would look like:
```rust
pub(crate) struct AreaMetadata {
    pub area_index: AreaIndex,
    pub total_size: u64,
    pub is_unpadded: bool,
}

galadd:
> I'm not yet convinced this new type is needed as it seems to add little value on top of the existing AreaIndex type.
> 
> Also it seems to have a conflict with main at this point that must be resolved.

I'm having troubles reproducing the conflict you are referring to

PR #2177, opened 2026-08-03
fix(fwdctl): escape JSON dump strings

### LINKED ISSUE #2176: fwdctl dump --output-format json produces invalid JSON when keys/values contain quotes
Title: fwdctl dump --output-format json produces invalid JSON when keys/values contain quotes

## Description

`JsonOutputHandler` in `fwdctl/src/dump.rs` writes keys and values into the JSON document without escaping:

```rust
write!(self.writer, r#"  "{key_str}": "{value_str}""#)?;
```

When a key or value contains a `"` (or `\`, or control characters) and `--hex` is not used, the output is not valid JSON.

## Reproduction

```console
$ fwdctl create --db demo
$ fwdctl insert --db demo 'say"hi' 'x'
$ fwdctl dump --db demo --output-format json --output-file-name dump
$ cat dump.json
{
  "say"hi": "x"
}
$ python3 -c 'import json; json.load(open("dump.json"))'
json.decoder.JSONDecodeError: Expecting ':' delimiter: line 3 column 8 (char 21)
```

## Impact

Any consumer parsing the dump with a real JSON parser fails on such databases. This also affects the upcoming JSON input path for `fwdctl import` (#966), which parses strictly and would reject these dumps. `--hex` output is unaffected since hex strings never need escaping.

## Suggested fix

Serialize the key/value strings with `serde_json` (or escape manually) in `JsonOutputHandler::handle_record`. Happy to submit a PR for this if useful — found while implementing the JSON import path for #966.

xianjianlf2:
Fixes #2176

The JSON dump handler previously interpolated keys and values directly into quoted strings, producing invalid JSON when records contained quotes, backslashes, control characters, or newlines.

This adds a streaming JSON string escaper and covers the special-character cases with an integration test.

Validation:
- cargo fmt --all -- --check
- cargo test -p firewood-fwdctl --test cli
- cargo clippy -p firewood-fwdctl --all-targets -- -D warnings

### DIFF
--- a/fwdctl/src/dump.rs
+++ b/fwdctl/src/dump.rs
@@ -226,6 +226,10 @@ fn key_value_to_string(key: &[u8], value: &[u8], hex: bool) -> (String, String) 
     (key_str, value_str)
 }
 
+fn write_json_string<W: Write>(writer: &mut W, value: &str) -> Result<(), std::io::Error> {
+    write!(writer, "\"{}\"", json_escape::escape_str(value))
+}
+
 fn handle_next_key(next_key: KeyFromStream) {
     match next_key {
         Some(Ok((key, _))) => {
@@ -283,7 +287,10 @@ impl OutputHandler for JsonOutputHandler {
             self.writer.write_all(b",\n")?;
         }
 
-        write!(self.writer, r#"  "{key_str}": "{value_str}""#)?;
+        write!(self.writer, "  ")?;
+        write_json_string(&mut self.writer, &key_str)?;
+        write!(self.writer, ": ")?;
+        write_json_string(&mut self.writer, &value_str)?;
         Ok(())
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1905,6 +1905,7 @@ dependencies = [
  "hex",
  "humantime",
  "indicatif",
+ "json-escape",
  "log",
  "nonzero_ext",
  "num-format",
@@ -2930,6 +2931,15 @@ dependencies = [
  "cfg-if",
  "futures-util",
  "wasm-bindgen",
+]
+
+[[package]]
+name = "json-escape"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef6fd1644b881a0cb8030578c318e14840e543876e98f7bc9215270fd55ca5db"
+dependencies = [
+ "memchr",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


rkuris (changes requested):
Can you use the json-escape crate? It supports streaming and is optimized for performance.

xianjianlf2:
Implemented the requested change in 3ff37c0: replaced the hand-written JSON escaping with json-escape 0.3.1 streaming escaping, kept the special-character regression test, and synced origin/main. Validation: default fwdctl CLI tests 20/20 and ethhash tests 26/26 passed.

xianjianlf2:
已推送 `a11e9520` 同步最新 `main`。现有 `c44da88` 已使用 `json-escape` 的流式 Display 实现，并覆盖特殊字符 JSON 输出；`cargo fmt --check` 和目标测试通过。@rkuris 请复审。

fwdctl/src/dump.rs
```diff
@@ -283,7 +287,10 @@ impl OutputHandler for JsonOutputHandler {
             self.writer.write_all(b",\n")?;
         }
 
-        write!(self.writer, r#"  "{key_str}": "{value_str}""#)?;
+        write!(self.writer, "  ")?;
+        write_json_string(&mut self.writer, &key_str)?;
+        write!(self.writer, ": ")?;
+        write_json_string(&mut self.writer, &value_str)?;
```
rkuris, line 293: This is fine, but it would be a bit cleaner (possibly faster) to just do this inline
```suggestion
use json_escape::escape_str;
        write!(self.writer, r#"  "{}": "{}""#, escape_str(&key_str), escape_str(value_str))?;
```

rkuris (approved):

rkuris:
This is ready to merge, but requires that you sign your commits. We have instructions here: https://github.com/ava-labs/firewood/blob/main/CONTRIBUTING.md#signing-your-commits

xianjianlf2:
Updated the branch to satisfy the repository's signed-commit requirement.

- Rebased the PR change onto the latest `main` (`8489542`).
- Recreated the JSON escaping change as signed commit `6278105d`; GitHub now shows it as **Verified**.
- Kept the streaming `json-escape` implementation and special-character regression test.
- Validation: `cargo fmt --all -- --check` passed; `cargo test -p firewood-fwdctl --test cli` passed (20/20).

The existing inline suggestion is optional; the helper still uses `json_escape::escape_str`'s streaming `Display` implementation.

PR #2178, opened 2026-08-03
ci: run CI and local prepush checks through shared just recipes

RodrigoVillar:
## Why this should be merged

When working with coding agents, especially when building a stack of PRs, a lint or test failure in the base PR causes every PR above it to fail as well. The author then has to fix the base PR and rebase the rest of the stack, creating avoidable work and slowing developer velocity.

We should run as many checks as practical locally before pushing code. CI should be the final gate before merging into `main`, rather than the first place routine lint or unit-test failures are discovered. Although the existing CI jobs are individually reproducible, there was no simple way to run the same collection of checks locally. CI runs many commands across several jobs, making it easy for developers and coding agents to miss a check before pushing.

## How this works

This PR provides a small Justfile API shared by developers, coding agents, and CI:

  - `just lint` runs the macOS-compatible linting and documentation checks.
  - `just test` runs the macOS-compatible unit and FFI test suites.
  - `just prepush` runs both phases in the expected order (on my local macbook, this took ~6.5 minutes to run)

The lower-level `ci-*` recipes are also used by GitHub Actions. Named Rust profiles and their Cargo arguments are defined in one shared script, avoiding duplicated command arguments between the Justfile and CI workflows.

`AGENTS.md` now directs developers and coding agents to these commands before pushing. This gives both groups a consistent, simple interface for reproducing as much of CI as possible locally.

Checks that require a Linux environment, such as the `io-uring` all-features profile and differential fuzz jobs, remain CI-only.

## How this was tested

CI + ran `just` commands locally.

> Note: for this PR to be merged, the `main` branch protection rules will need to be modified to account for the new CI jobs. For example, `rust-lint (debug-ethhash-logger, --features ethhash,logger)` is now just `rust-lint (debug-ethhash-logger)` now that `profile-args` have been removed.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,21 +27,18 @@ jobs:
       matrix:
         include:
           - profile-key: debug-no-default-features
-            profile-args: "--no-default-features"
           - profile-key: debug-no-features
-            profile-args: ""
           - profile-key: debug-no-features
-            profile-args: ""
             os: macos-26
           - profile-key: debug-ethhash-logger
-            profile-args: "--features ethhash,logger"
           - profile-key: debug-all-features
-            profile-args: "--all-features"
           - profile-key: maxperf-ethhash-logger
-            profile-args: "--profile maxperf --features ethhash,logger"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@67729d5c413db75907f0ad1e39bb04b9c868ff60 # @v2 -> v2.85.7
+        with:
+          tool: just@1.46.0
       - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # @master
         with:
           toolchain: stable
@@ -55,14 +52,17 @@ jobs:
       - name: Cargo Fetch (run `cargo generate-lockfile` if this fails)
         run: cargo fetch --locked --verbose
       - name: Check
-        run: cargo check --frozen ${{ matrix.profile-args }} --workspace --all-targets
+        run: ./scripts/run-just.sh ci-rust check ${{ matrix.profile-key }}
       - name: Build
-        run: cargo build --frozen ${{ matrix.profile-args }} --workspace --all-targets
+        run: ./scripts/run-just.sh ci-rust build ${{ matrix.profile-key }}
 
   no-rust-cache-lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@67729d5c413db75907f0ad1e39bb04b9c868ff60 # @v2 -> v2.85.7
+        with:
+          tool: just@1.46.0
       - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # @master
         with:
           toolchain: stable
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,14 +52,17 @@ jobs:
       - name: Cargo Fetch (run `cargo generate-lockfile` if this fails)
         run: cargo fetch --locked --verbose
       - name: Check
-        run: cargo check --frozen ${{ matrix.profile-args }} --workspace --all-targets
+        run: ./scripts/run-just.sh ci-rust check ${{ matrix.profile-key }}
```
JonathanOppenheimer, line 55: Does it make sense to use `just` over `task?. See: https://github.com/ava-labs/avalanchego/blob/master/docs/tasks.md#why-not-make-or-just. If firewood will eventually be grafted, or even if not, IMO we should be consistent across our repos where it makes sense.

JonathanOppenheimer (commented):
This is the premise of the PR so didn't review further. Lmk your thoughts!

.github/workflows/ci.yaml
RodrigoVillar: Reading the link above:

```
Task was chosen over just because the existing dependency on Go tooling made adoption simpler
```

Likewise, `just` is written in Rust and can be installed by `cargo install just`. This should work because any Firewood developer should have `cargo` installed to work in this repo - you don't necessarily need to have `go` installed unless you want to work in the `ffi` package.

JonathanOppenheimer: Will Firewood eventually be grafted into AvalancheGo or nah? I'm actually curious here.

demosdemon: we already use just. switching to task wouldn't make sense for this PR even if we were to switch. 

I don't have an answer for the other question.

.github/workflows/ci.yaml
```diff
@@ -27,21 +27,18 @@ jobs:
...
           - profile-key: debug-ethhash-logger
-            profile-args: "--features ethhash,logger"
           - profile-key: debug-all-features
-            profile-args: "--all-features"
           - profile-key: maxperf-ethhash-logger
-            profile-args: "--profile maxperf --features ethhash,logger"
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@67729d5c413db75907f0ad1e39bb04b9c868ff60 # @v2 -> v2.85.7
```
demosdemon: This arrow syntax is meaningless to dependabot. Dependabot will only update the tail version leaving the first version to go stale. We should remove them or at least not add more. E.g.,

https://github.com/ava-labs/firewood/blob/f024883c8ba95a22ab142746ecfd5139272a2e65/.github/workflows/ci.yaml#L249

demosdemon (approved):
We won't be able to merge this until the repository rules are updated. Changing anything in the matrix is a breaking change there.

.github/workflows/ci.yaml
RodrigoVillar: Good call, removed the arrow syntax here and followed the example linked above here: [e32e18f](https://github.com/ava-labs/firewood/pull/2178/commits/e32e18f0e45716c4ccf358d7d5e2a0fe3929aad4)

.github/workflows/ci.yaml
JonathanOppenheimer: Got it -- yeah I imagine in the future we'd want to align on one tool, but perhaps it may never actually be necessary to align!

PR #2179, opened 2026-08-04
ci: define required merge checks in code

RodrigoVillar:
## Why this should be merged

#2178 changed the CI matrix-generated check names, which required a maintainer to update the repository’s merge rules manually before it could be merged. That process was tedious, error-prone, and created an unnecessary bottleneck even though the PR was already approved and its replacement checks were passing.

Required merge checks should be defined in version-controlled code so changes can be reviewed alongside the workflow changes that require them. This follows the same general approach used by AvalancheGo (ref: https://github.com/ava-labs/avalanchego/blob/9e14e9aa278e65c6caeaafb08ec202e720037134/.github/workflows/ci.yml#L26) and avoids requiring a maintainer to edit GitHub settings whenever CI jobs or matrix values change.

## How this works

This PR adds a single aggregate `tests-required` job. Instead of configuring branch protection with a collection of individual, matrix-generated check names, branch protection only needs to require `tests-required`. The aggregate job:

  - Depends on each merge-blocking CI job through its needs list.
  - Uses `always()` so it still runs when a dependency fails or is skipped.
  - Fails unless every listed dependency succeeds.

After the one-time branch-protection update, adding or removing a merge-blocking check only requires changing `tests-required.needs` in the CI workflow. The change can then be reviewed and merged like any other code change.

This initially expands the effective set of required passing jobs. The existing rules require 14 individual check contexts, while `tests-required` depends on all matrix variants across seven job groups - 28 underlying job executions in total. Branch protection sees only one stable required status, but every one of those underlying jobs must succeed for it to pass. This broader requirement preserves a strong merge gate while moving its definition into code. The dependency list can be trimmed in a follow-up if some matrix variants do not need to be merge-blocking.

## How this was tested

The new tests-required job is exercised by GitHub Actions on this PR. It runs after all jobs in its needs list complete and validates their reported results.

https://github.com/ava-labs/firewood/actions/runs/30950282102/job/92133177781?pr=2179

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,6 +10,40 @@ env:
   RUST_BACKTRACE: full
 
 jobs:
+  # Keep this as the only CI status required by main's branch protection.
+  # To make another job merge-blocking, add its job key to this list. This
+  # keeps the required checks version-controlled instead of requiring a
+  # corresponding branch-protection update in the GitHub UI.
+  tests-required:
+    if: ${{ always() }}
+    needs:
+      - build
+      - no-rust-cache-lint
+      - rust-lint
+      - test
+      - examples
+      - docs
+      - ffi
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail unless all required jobs succeeded
+        env:
+          NEEDS_JSON: ${{ toJSON(needs) }}
+        run: |
+          failed_jobs="$(
+            jq -r '
+              to_entries[]
+              | select(.value.result != "success")
+              | .key
+            ' <<<"$NEEDS_JSON"
+          )"
+
+          if [[ -n "$failed_jobs" ]]; then
+            echo "Required jobs failed:"
+            printf '%s\n' "$failed_jobs"
+            exit 1
+          fi
+
   build:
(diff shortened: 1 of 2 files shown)


demosdemon (commented):
This is a nifty idea I hadn't considered.

We don't actually need to go this far. The `needs` relationship is enough. This can simply do a `true` and drop the `always()`. Without the `always` and if any of the listed needs fail, the job never runs thus the requirement fails.

demosdemon (approved):

demosdemon:
> This is a nifty idea I hadn't considered.
> 
> We don't actually need to go this far. The `needs` relationship is enough. This can simply do a `true` and drop the `always()`. Without the `always` and if any of the listed needs fail, the job never runs thus the requirement fails.

Rodrigo pointed me to <https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks> offline. This explanation matches the code here but doesn't line up with my experience. I'll trust docs in this case.

rkuris (approved):

PR #2181, opened 2026-08-05
feat(proofs)!: add zstd compression for proof body

### LINKED ISSUE #1869: Improve (de)serialization of proofs
Range and Change proofs are serialized and sent across the wire, but they contain information that can easily be derived, and therefore are larger than they need to be.

Only send non-duplicated proof nodes. The deserializer can recreate them. Additionally, the keys sent across contain a common prefix. Remove that common prefix based on how many duplicated nodes are between the two proofs during serialization and add it back in when deserializing.

Another way to do this is to send tries instead of k/v pairs

AminR443:
## Why this should be merged

Adds zstd compression to the proof wire format. Measured on C-Chain mainnet replays, compressed proofs at the production state-sync message size are 46–57% smaller, roughly halving sync bandwidth, for ~3.3 ms/MiB compress and ~0.5 ms/MiB decompress.

## How this works

The wire becomes a 32-byte uncompressed header followed by a single zstd frame compressing the proof body. zstd output isn't bit-stable across versions, so the decompressed body is the canonical form (new `write_body_to_vec`). The decoder enforces all bounds before allocating: exactly one frame spanning the remainder, a mandatory frame header content size capped at 4 MiB and at 128× the frame length (bomb defense). The encoder will error if proof is larger than the cap specified since no consumer will be able to consume it.

Closes #1869.

## How this was tested

Unit tests cover roundtrips, wire shape, and the rejection taxonomy (trailing/skippable/concatenated frames, corruption, truncation, missing content size, over-cap, compression bomb), plus a golden vector pinning decode-side compatibility across zstd upgrades and two fuzz tests (canonical body and wire corruption). Full CI matrix passes locally.

## Breaking Changes

- [x] firewood
- [ ] firewood-storage
- [x] firewood-ffi (C api)
- [x] firewood-go (Go api)
- [ ] fwdctl

Wire format changes incompatibly under version 0, deliberate hard cutover, since v0 had zero users.


firewood/src/proofs/de.rs
```diff
@@ -97,32 +102,7 @@ impl FrozenChangeProof {
     /// Returns a [`ReadError`] if the data is invalid. See the enum variants for
     /// the possible reasons.
     pub fn from_slice(data: &[u8]) -> Result<Self, ReadError> {
```
rkuris, line 105: Why are we just wrapping from_slice to call parse_proof instead?

firewood/src/proofs/de.rs
```diff
@@ -25,6 +25,35 @@ use firewood_storage::{Children, PathBuf, TrieHash, TriePathFromUnpackedBytes, V
...
+    let this = reader.read_v0_item()?;
+    if reader.remainder().is_empty() {
+        Ok(this)
+    } else {
+        Err(reader.invalid_item(
+            "trailing bytes",
+            "no data after the proof",
+            format!("{} bytes", reader.remainder().len()),
+        ))
+    }
```
rkuris:
```rust
if !reader.remainer().is_empty() {
    // error handling here
    // return Err(...)
}

Ok(this)
```

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/// Decompresses a framed proof body (`varint(uncompressed_len) :: zstd
+/// frame`) from the remainder of `reader`, enforcing the decode bounds
+/// described in the module docs. On success the reader is left at the end
+/// of its data; error offsets are reader offsets, so they line up with a
+/// hexdump of the wire bytes.
+pub(super) fn decompress_body(
+    reader: &mut ProofReader<'_>,
+    kind: &'static str,
```
rkuris: This should not be a generic string, use the enum.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,307 @@
...
+// See the file LICENSE.md for licensing terms.
+
+//! # Compressed proof body framing
+//!
+//! Serialized proofs carry their body inside a single zstd frame:
+//!
+//! ```text
+//! Header (32 bytes, uncompressed)
+//! varint(uncompressed_len)     // exact byte length of the canonical body
+//! zstd frame                   // compresses exactly the canonical body bytes
```
rkuris:
```suggestion
//! zstd frame                   // compressd "frame data"
```
What exactly is "frame data" here?

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,307 @@
+// Copyright (C) 2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! # Compressed proof body framing
+//!
+//! Serialized proofs carry their body inside a single zstd frame:
+//!
+//! ```text
+//! Header (32 bytes, uncompressed)
+//! varint(uncompressed_len)     // exact byte length of the canonical body
```
rkuris: Why is this needed?

firewood/src/proofs/frame.rs
rkuris: Looks like "frame data" is anything that implements `Version0`. Is that true? Do we send only a subset of these types across the wire? I think so, so perhaps this type is wrong?

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,307 @@
...
+//! zstd frame                   // compresses exactly the canonical body bytes
+//! ```
+//!
+//! The canonical body encoding (proof nodes, key-value pairs, batch
```
rkuris: This comment is not part of the code, perhaps it belongs in the PR description or notes

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,307 @@
...
+//! operations — see `ser.rs`) is unchanged; compression wraps around it.
+//! Measured on C-Chain mainnet replays (blocks 50M–56.44M and 75M–76M),
```
rkuris: This belongs in "how this was tested"

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,307 @@
...
+//!
+//! ## Determinism and canonicality
+//!
+//! The producer pins the zstd parameters: level [`ZSTD_LEVEL`], single
+//! frame, no dictionary, no checksum. For a fixed zstd library version
+//! the output is deterministic, but zstd does not guarantee bit-identical
+//! output across library upgrades. **The canonical form of a proof is its
+//! decompressed body.** Anything that hashes, signs, or byte-compares
+//! proofs must operate on the body bytes, never on the compressed wire
+//! bytes.
```
rkuris: Just say "Proof bytes after compression are not canonical because zstd does not make this guarantee"

firewood/src/proofs/de.rs
AminR443: removed these changes and kept them for a follow up PR.

firewood/src/proofs/frame.rs
AminR443: removed metrics for now.

firewood/src/proofs/frame.rs
AminR443: added better explanation.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright (C) 2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE.md for licensing terms.
+
+//! # Compressed proof body framing
+//!
+//! Serialized proofs carry their body inside a single zstd frame:
```
rkuris: This took a double-take, and is still very verbose. You can get everything from here to line 14 from this sentence:

"Serialized proofs consist of a `Header` followed by the zstd-compressed range or change proof"

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+//!
+//! The canonical body is one range or change proof's V0 body (see
+//! `write_body_to_vec` in `ser.rs`); the header's proof type says which.
+//!
+//! ## Determinism and canonicality
+//!
+//! The producer pins the zstd parameters: level [`ZSTD_LEVEL`], single
+//! frame, no dictionary, no checksum. zstd output is not bit-identical
+//! across library versions; **the decompressed body is the canonical
+//! form**.
```
rkuris: Still very wordy.
```suggestion
//! While the body of the deserialized proof is canonical, compression is not.
```

The zstd parameters shouldn't be discussed here. Why? They can drift from the actual code. Make the code more readable so if someone actually cares about what zstd compression mode we're using, they can just check the compression function.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use super::reader::ReadError;
+
+/// zstd compression level pinned by the producer rule.
+/// Level 3 is zstd's default and best measured tradeoff.
+pub(super) const ZSTD_LEVEL: i32 = 3;
```
rkuris: Remove this, and use DEFAULT_COMPRESSION_LEVEL from the crate. See https://docs.rs/zstd/0.13.3/zstd/constant.DEFAULT_COMPRESSION_LEVEL.html

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Upper bound on the declared uncompressed body length.
+pub(super) const MAX_DECOMPRESSED_LEN: usize = 32 * 1024 * 1024;
```
rkuris: This probably needs to be:
 - More readable. This is 32 mebibytes? Add "// 32 MiB" right after the value (this avoids drift)
 - Explanation as to why this value was chosen (hint: I think it's the wrong value)
 - Consider making this a configuration option at the DB layer or even per-proof. The former is much easier and probably fine. OK to add a TODO for this along with an associated issue.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Upper bound on the ratio between the declared uncompressed body length
+/// and the compressed frame length. Bounds the decoder's allocation by the
+/// bytes the peer actually sent (decompression-bomb defense).
+pub(super) const MAX_COMPRESSION_RATIO: usize = 128;
+
+/// Appends the zstd frame compressing `body`.
+#[cfg(test)]
```
rkuris: Can we move this further down? It's not an impl, should be in the #[cfg(test)] section scoped only for testing I think.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+    let compressed =
+        zstd::bulk::compress(body, ZSTD_LEVEL).expect("couldn't allocate zstd context");
+    out.extend_from_slice(&compressed);
+}
+
+/// Compresses `out[body_start..]` (the canonical body, serialized in place
+/// after the header) into a zstd frame, replacing it. Avoids a separate
+/// body buffer.
+pub(super) fn compress_body_in_place(out: &mut Vec<u8>, body_start: usize) {
+    debug_assert!(body_start <= out.len());
```
rkuris: It's already going to panic if you fail this assertion, no point checking it twice in debug mode.

debug_assert is also really evil. Remind me sometime to tell you about a bug that was caught during debug_assert but at runtime produced some really scary bugs. Arguably debug_assert should be banned.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[expect(
+        clippy::indexing_slicing,
+        reason = "callers record body_start as out.len() before appending the body"
+    )]
+    let compressed = zstd::bulk::compress(&out[body_start..], ZSTD_LEVEL)
```
rkuris: And this panics at runtime. This can be redesigned to avoid this, let me know if you need some ideas (I see a few ways).

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/// Enforces the pre-allocation bounds (see [`decompress_body`]) and
+/// returns the frame's declared content size.
+fn validate_frame(frame: &[u8], frame_offset: usize) -> Result<usize, ReadError> {
+    // `find_frame_compressed_size` measures only the *first* frame.
+    // Requiring it to span the entire remainder leaves no room for
+    // trailing concatenated or skippable frames, which the raw
+    // decompressor would otherwise silently accept — an
+    // attacker-controlled padding channel.
```
rkuris: Not reviewed, likely AI slop. At a minimum this is a misplaced comment. Readers of `validate_frame` have two reasons to read this function: 1, to get an overview, and 2 to read it in detail. A big block comment before the first line of code is likely to be an overview comment at best, possibly a design-level comment which doesn't belong here either way.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,139 @@
...
+    let frame_size = zstd::zstd_safe::find_frame_compressed_size(frame).map_err(|code| {
```
rkuris: nit: A::B::C is usually code smell inline. It makes reading this code a bit tougher. I'd be okay with `use A::B` then `B::C` or `use A::B::C` and then just `C`, depending on how clearly you want to say where this is coming from.

firewood/Cargo.toml
```diff
@@ -34,6 +34,7 @@ firewood-storage.workspace = true
 hex.workspace = true
 integer-encoding.workspace = true
 thiserror.workspace = true
+zstd.workspace = true
```
rkuris: Why workspace? Only this crate needs it.

rkuris (dismissed):
Partial review, please resubmit for another pass.

firewood/Cargo.toml
AminR443: Yeah no need for workspace. lingered around from experiments. Fixed.

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/frame.rs
AminR443: Not AI slop. I agree on placement being weird. I moved it upper and shortened it, if you disagree with the new position as well, I'm open to suggestions. Using `find_frame_compressed_size` and enforcing a single frame is a little bit unintuitive and I think it deserved more description to remain in code.

firewood/src/proofs/frame.rs
AminR443: Is this about the `expect`? As far as I checked, compression can only fail here if an allocation fails, either for the zstd context or the output buffer. At that point, wouldn't Rust abort anyway?

That's mainly why I didn't propagate the error since it feels like a point of no return. Let me know plz how you'd prefer to handle it and I'll make the changes.

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/frame.rs
AminR443: As I mentioned these numbers are placeholders, because I'm still not sure about the proof size we're targeting,  existing sync code I saw targeted near 2MiB (which I did my evaluations based on), but in some of calls we talked about 8MiB being the actual protocol limit IIRC. This variable should be set in a way that it leaves enough headroom for valid proofs while putting a limit on allocation.

I like the idea of making it a config option and added a TODO and made it more readable.

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/frame.rs
rkuris: Config option, separate PR. With no consumers you can land it before this one.

firewood/src/proofs/tests.rs
```diff
@@ -1178,6 +1223,270 @@ mod box_array_deserialization_tests {
...
 
+fn range_wire() -> Vec<u8> {
+    let (proof, _) = create_valid_range_proof();
+    let mut wire = Vec::new();
+    proof.write_to_vec(&mut wire);
+    wire
+}
+
+fn change_wire() -> Vec<u8> {
+    let (proof, _) = create_valid_change_proof(NodeHashAlgorithm::MerkleDB);
```
RodrigoVillar: q: must this be using the `MerkleDB` scheme, or can this be generalized to use either hashing algorithm (i.e. `MerkleDB` and `Ethhash`)?

RodrigoVillar (commented):
I had one small comment related to the recent hashing runtime changes.

rkuris (commented):
Partial review

firewood/src/proofs/ser.rs
```diff
@@ -117,6 +135,22 @@ impl ProofWriter<'_> {
...
+    proof: &impl WriteItem,
+    proof_type: ProofType,
+    mode: NodeHashAlgorithm,
+    out: &mut Vec<u8>,
+) {
+    let mut w = ProofWriter { out, mode };
+    Header::from((proof_type, mode)).write_item(&mut w);
+    let body_start = w.out.len();
+    proof.write_item(&mut w);
+    super::frame::compress_body_in_place(w.out, body_start);
```
rkuris: This method signature should be rethought to handle the ProofWriter inferface better

firewood/src/proofs/tests.rs
```diff
@@ -10,21 +10,54 @@ use firewood_storage::{
...
 
+/// Builds the 32-byte proof header
+fn raw_header(proof_type: ProofType) -> Vec<u8> {
+    bytemuck::bytes_of(&Header::from((proof_type, DefaultHashMode::ALGORITHM))).to_vec()
+}
+
+/// Compresses the body of uncompressed `header || body` bytes into wire
+/// form, then parses. Inputs shorter than a header parse as-is.
+fn compress_and_parse_range(data: &[u8]) -> Result<FrozenRangeProof, ReadError> {
+    match data.split_at_checked(32) {
```
rkuris: Magic constant 32 should come from the code

firewood/src/proofs/tests.rs
```diff
@@ -10,21 +10,54 @@ use firewood_storage::{
...
+        Some((header, body)) => {
+            let mut wire = header.to_vec();
+            super::frame::write_compressed_body(body, &mut wire);
+            FrozenRangeProof::from_slice(&wire)
+        }
+        None => FrozenRangeProof::from_slice(data),
```
rkuris, line 45: This is strange, we assume that buffers less than 32 bytes can't contain a header, but then we try to decode it using FrozenRangeProof, which is likely to fail.

firewood/src/proofs/tests.rs
```diff
@@ -10,21 +10,54 @@ use firewood_storage::{
...
+        }
+        None => FrozenChangeProof::from_slice(data),
+    }
+}
+
+/// Returns a valid range proof plus its canonical uncompressed bytes
+/// (`header || body`) for the byte-taxonomy tests. Parse the result with
+/// [`compress_and_parse_range`].
 fn create_valid_range_proof() -> (FrozenRangeProof, Vec<u8>) {
     let merkle = crate::merkle::tests::init_merkle((0u8..=10).map(|k| ([k], [k])));
```
rkuris, line 71: This function should include exactly what contents of the range proof are being generated. Even better if it's in the function name. It's not just valid, it's a specific proof (why is this one valuable is helpful to answer here).

firewood/src/proofs/de.rs
```diff
@@ -34,6 +34,15 @@ impl FrozenRangeProof {
     /// Returns a [`ReadError`] if the data is invalid. See the enum variants for
     /// the possible reasons.
     pub fn from_slice(data: &[u8]) -> Result<Self, ReadError> {
```
rkuris, line 37: This method is only used in tests now, let's at a minimum remove pub and make it test only. Better yet, let's get rid of it

firewood/src/proofs/frame.rs
AminR443: #2197

firewood/src/proofs/frame.rs
AminR443: Fixed.

firewood/src/proofs/tests.rs
AminR443: Fixed.

firewood/src/proofs/ser.rs
AminR443: Done.

firewood/src/proofs/tests.rs
AminR443: Done.

firewood/src/proofs/tests.rs
AminR443: Documented the reason.

firewood/src/proofs/de.rs
AminR443: A lot of tests used it. changed it to test only.

firewood/src/proofs/tests.rs
AminR443: This wasn't part of my changes, but I tried to add descriptions to both of them.

ffi/src/proofs/range.rs
```diff
@@ -557,10 +557,12 @@ pub extern "C" fn fwd_range_proof_to_bytes(proof: Option<&RangeProofContext>) ->
 /// - [`RangeProofResult::Err`] containing an error message if the proof could not be parsed.
 #[unsafe(no_mangle)]
 pub extern "C" fn fwd_range_proof_from_bytes(
+    db: Option<&DatabaseHandle>,
     bytes: BorrowedBytes<'_>,
 ) -> RangeProofResult<'static> {
+    let config = db.map(DatabaseHandle::proof_config).unwrap_or_default();
```
demosdemon: this feels unsafe. the user doesn't get to provide the parameters to `validate_frame` which may be different than when the proof was generated.

demosdemon: @alarso16 -- fixing this correctly would require getting rid of the database free deserialization.

demosdemon: my knee jerk reaction says this shouldn't be a database specific option which would allow us to hoist this into a global static. That materially changes #2197 and is a deviation from our normal configuration.

AminR443: changed to const.

firewood/src/proofs/ser.rs
```diff
@@ -80,25 +85,49 @@ impl FrozenRangeProof {
     ///
     /// Variable-length integers are encoded using unsigned LEB128.
     pub fn write_to_vec(&self, out: &mut Vec<u8>) {
+        let header = Header::from((ProofType::Range, self.hash_mode()));
+        out.extend_from_slice(bytemuck::bytes_of(&header));
+        let mut body = Vec::new();
+        self.write_body_to_vec(&mut body);
+        super::frame::write_compressed_body(&body, out);
```
demosdemon: This never consults `MAX_DECOMPRESSED_LEN` or `MAX_COMPRESSION_RATIO`. `fwd_db_range_proof` takes `max_length: u32` where 0 means unbounded (Go tests use exactly that). An unbounded range over real C-Chain state yields a >4 MiB body; `MarshalBinary` returns it with no error, and the receiving peer fails with `expected content size within the maximum, but found 5242880 > 4194304`. The producing node never learns.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,107 @@
...
+use zstd::zstd_safe;
+
+use super::reader::ReadError;
+
+/// Hard cap on a decoded proof body.
+pub(super) const MAX_DECOMPRESSED_LEN: usize = 4 * 1024 * 1024; // 4 MiB
+
+/// Upper bound on the ratio between the uncompressed body length and the
+/// compressed frame length.
+pub(super) const MAX_COMPRESSION_RATIO: usize = 128;
```
demosdemon: MAX_DECOMPRESSED_LEN doesn't match the pr description. And, these values need rationales for their choices. What is the impact of changing and can they be changed once released?

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Enforces the pre-allocation bounds (see [`decompress_body`]) and returns
+/// the frame's declared content size.
+///
+/// Exactly one zstd frame must span all of `frame`: the raw decompressor
+/// accepts concatenated frames and silently skips skippable ones, which
+/// would otherwise be an attacker-controlled padding channel.
+fn validate_frame(frame: &[u8], frame_offset: usize) -> Result<usize, ReadError> {
```
demosdemon, line 36: This passes a lone *skippable* zstd frame. This returns 16 (`== frame.len()`), and `get_frame_content_size` returns `Ok(Some(0))`, not `None` (see zstd_decompress.c:580), so it clears every guard and `decompress_body` returns `Ok(vec![])`. Not exploitable — the empty body fails the parse — but the doc at frame.rs:24-27 claims an invariant the code doesn't hold, the caller gets a body-parser error (`incomplete array length at offset 0`) for a frame-layer violation, and an attacker can stuff arbitrary payload into the skippable frame and have the frame layer accept it. Only the *trailing* skippable case is tested (tests.rs:1324). 

Suggested fix: reject decoded_len == 0, and add the lone case to test_frame_rejects_malformed_wire.

demosdemon (changes requested):

firewood/src/proofs/ser.rs
AminR443: Added errors.

firewood/src/proofs/frame.rs
AminR443: Added docs. will update PR description.

firewood/src/proofs/frame.rs
AminR443: Done.

firewood/src/proofs/frame.rs
```diff
@@ -0,0 +1,114 @@
...
+        found: err.to_string(),
+    })?;
+    Ok(body)
+}
+
+/// Appends the single zstd frame compressing `body` (the canonical
+/// serialized body that follows the header on the wire).
+pub(super) fn write_compressed_body(body: &[u8], out: &mut Vec<u8>) {
+    let compressed = zstd::bulk::compress(body, zstd::DEFAULT_COMPRESSION_LEVEL)
+        .expect("zstd compressor allocation failed");
```
demosdemon: Now that `write_to_vec` is fallible, this should return an error instead.

demosdemon (approved):

PR #2182, opened 2026-08-05
fix(storage): make the io-uring feature a portable no-op off Linux

demosdemon:
## Why this should be merged

`--all-features` could not be run on macOS: the `io-uring` feature pulled in the `io-uring` crate unconditionally, and that crate has no `target_os` gate of its own — `src/sys/mod.rs` dispatches purely on `target_arch`, so on Apple Silicon it selects the `aarch64` bindings and then fails against Linux-only `libc` items. There is no upstream fix to wait for.

The cost had spread past the build: `debug-all-features` was excluded from `just lint` and `just test`, and four documents told humans and agents to skip `--all-features` on macOS, so nothing run locally exercised that configuration.

## How this works

The Cargo feature stays portable and declarable everywhere; a new `cfg(io_uring)` alias emitted by `storage/build.rs` gates the code on the feature _and_ a Linux target.

The non-obvious constraint is that the two halves cannot be separated. Target-gating the dependency alone would make things _worse_: `cfg(feature = "io-uring")` still evaluates true on macOS, so the existing sites would reference a crate no longer in the graph.

`Cargo.lock` is unchanged by design — it records dependencies target-agnostically, and CI runs `--frozen`/`--locked`.

## How this was tested

`just lint` and `just test` pass in full on macOS, now including `debug-all-features` (765/765). Verified in a `rust:1.96-bookworm` container that `--all-features` still compiles `io-uring v0.7.13`, so the ring is not silently disabled — and that substituting a `build.rs` which omits the cfg fails with `error[E0080]`, confirming the new `const` guard is not decorative.

## Follow-ups

- macOS is **not** added to the `debug-all-features` CI matrices, so that configuration is enforced only by local `just prepush`. Adding `os: macos-26` to five entries in `ci.yaml` would make CI authoritative, at the cost of CI minutes per PR.
- A green macOS `--all-features` run does not prove `storage/src/linear/io_uring.rs` compiles — only the Linux jobs build it. Called out in `AGENTS.md` and the review skill so the new checkmark is not over-read.

## Breaking Changes

None. The feature keeps its name and its Linux behavior, and stays off by default everywhere. Downstream consumers who split the dependency by target to work around this can now enable `io-uring` unconditionally.

### DIFF
--- a/justfile
+++ b/justfile
@@ -85,13 +85,14 @@ ci-test-ffi-compat hash_mode:
     cd "$test_dir"
     go test -count=1 -race ./...
 
-# Run macOS-compatible CI lints (all-features/io-uring remains Linux-only).
+# Run macOS-compatible CI lints.
 lint:
     ./scripts/run-just.sh ci-format
     ./scripts/run-just.sh ci-check-todos
     ./scripts/run-just.sh ci-rust clippy debug-no-default-features
     ./scripts/run-just.sh ci-rust clippy debug-no-features
     ./scripts/run-just.sh ci-rust clippy debug-ethhash-logger
+    ./scripts/run-just.sh ci-rust clippy debug-all-features
     ./scripts/run-just.sh ci-rust clippy maxperf-ethhash-logger
     ./scripts/run-just.sh ci-lint-markdown
     ./scripts/run-just.sh ci-docs
@@ -99,11 +100,12 @@ lint:
     ./scripts/run-just.sh ci-check-go-generate
     ./scripts/run-just.sh ci-machete
 
-# Run macOS-compatible CI unit tests (all-features/fuzz remain Linux-only).
+# Run macOS-compatible CI unit tests (differential fuzzing remains Linux-only).
 test:
     ./scripts/run-just.sh ci-rust test debug-no-default-features
     ./scripts/run-just.sh ci-rust test debug-no-features
     ./scripts/run-just.sh ci-rust test debug-ethhash-logger
+    ./scripts/run-just.sh ci-rust test debug-all-features
     ./scripts/run-just.sh ci-rust test maxperf-ethhash-logger
     ./scripts/run-just.sh ci-build-ffi firewood
     ./scripts/run-just.sh ci-test-ffi firewood
--- a/storage/src/linear/mod.rs
+++ b/storage/src/linear/mod.rs
@@ -16,9 +16,24 @@ use std::path::PathBuf;
 
 use crate::{CacheReadStrategy, LinearAddress, MaybePersistedNode, SharedNode};
 pub(super) mod filebacked;
-#[cfg(feature = "io-uring")]
+/// Batched-write backend built on Linux `io_uring`.
+///
+/// Gated on `cfg(io_uring)`, which `build.rs` sets only when the `io-uring`
+/// feature is enabled *and* the target is Linux. The `io-uring` crate does not
+/// build on other targets, so the feature is accepted but inert there and
+/// [`FileBacked`](filebacked::FileBacked) falls back to the serial
+/// [`WritableStorage::write_batch`] default.
+#[cfg(io_uring)]
 mod io_uring;
 pub mod memory;
+
+/// Pins the `cfg(io_uring)` policy that `build.rs` implements.
+///
+/// A regression that stopped emitting the cfg would silently fall back to the
+/// serial write path on Linux without failing anything — the ring has no unit
+/// tests of its own. Conversely, emitting it off Linux would fail to compile
+/// with a confusing missing-crate error. This catches both at build time.
+const _: () = assert!(cfg!(io_uring) == (cfg!(feature = "io-uring") && cfg!(target_os = "linux")));
 
 /// An error that occurs when reading or writing to a [`ReadableStorage`] or [`WritableStorage`]
 ///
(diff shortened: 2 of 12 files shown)


storage/src/linear/mod.rs
```diff
@@ -213,9 +228,14 @@ pub trait WritableStorage: ReadableStorage {
...
-    /// objects at once.
+    /// objects at once. This default writes each object in turn via
+    /// [`Self::write`]; under `cfg(io_uring)`,
+    /// [`FileBacked`](filebacked::FileBacked) overrides it with a single ring
+    /// submission.
     ///
     /// The iterator is expected to be cheap to clone without copying the data.
+    /// The [`Clone`] bound exists for the io-uring override, which re-walks the
+    /// batch when retrying entries.
     fn write_batch<'a, I: IntoIterator<Item = (u64, &'a [u8])> + Clone>(
```
RodrigoVillar, line 239: (out-of-scope) I don't think the `Clone` method is currently load-bearing; I had AI removed the `Clone` bound and ran `cargo check` against the Linux target, which worked.

justfile
```diff
@@ -85,25 +85,27 @@ ci-test-ffi-compat hash_mode:
...
     ./scripts/run-just.sh ci-check-go-generate
     ./scripts/run-just.sh ci-machete
 
-# Run macOS-compatible CI unit tests (all-features/fuzz remain Linux-only).
+# Run macOS-compatible CI unit tests (differential fuzzing remains Linux-only).
 test:
     ./scripts/run-just.sh ci-rust test debug-no-default-features
     ./scripts/run-just.sh ci-rust test debug-no-features
     ./scripts/run-just.sh ci-rust test debug-ethhash-logger
+    ./scripts/run-just.sh ci-rust test debug-all-features
```
RodrigoVillar, line 108: This step fails on my local machine:

```
./scripts/run-just.sh ci-rust test debug-all-features
./scripts/run-rust-ci.sh "test" "debug-all-features"
error: rustc 1.94.0 is not supported by the following packages:
  aws-config@1.9.0 requires rustc 1.94.1
  aws-credential-types@1.3.0 requires rustc 1.94.1
  aws-runtime@1.8.1 requires rustc 1.94.1
  aws-sdk-ec2@1.239.0 requires rustc 1.94.1
  aws-sdk-ssm@1.115.0 requires rustc 1.94.1
  aws-sdk-sso@1.103.0 requires rustc 1.94.1
  aws-sdk-ssooidc@1.105.0 requires rustc 1.94.1
  aws-sdk-sts@1.108.0 requires rustc 1.94.1
  aws-sigv4@1.5.1 requires rustc 1.94.1
  aws-smithy-async@1.3.0 requires rustc 1.94.1
  aws-smithy-http@0.64.0 requires rustc 1.94.1
  aws-smithy-http-client@1.2.0 requires rustc 1.94.1
  aws-smithy-json@0.63.0 requires rustc 1.94.1
  aws-smithy-observability@0.3.0 requires rustc 1.94.1
  aws-smithy-query@0.61.1 requires rustc 1.94.1
  aws-smithy-runtime@1.12.0 requires rustc 1.94.1
  aws-smithy-runtime-api@1.13.0 requires rustc 1.94.1
  aws-smithy-runtime-api-macros@1.1.0 requires rustc 1.94.1
  aws-smithy-schema@0.2.0 requires rustc 1.94.1
  aws-smithy-types@1.6.1 requires rustc 1.94.1
  aws-smithy-xml@0.61.1 requires rustc 1.94.1
  aws-types@1.4.0 requires rustc 1.94.1
  aws-types@1.4.0 requires rustc 1.94.1
  aws-types@1.4.0 requires rustc 1.94.1
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.94.0

error: command `/Users/rodrigo.villar/.rustup/toolchains/stable-aarch64-apple-darwin/bin/cargo test --no-run --message-format json-render-diagnostics --all-features --locked` exited with code 101
error: Recipe `ci-rust` failed on line 7 with exit code 101
error: Recipe `test` failed on line 108 with exit code 101
```

Here's the version of Rust I'm using locally:

```
rustc 1.94.0 (4a4ef493e 2026-03-02)
```

This may be a "you should update your version of Rust" issue but wanted to bring this up nonetheless.

RodrigoVillar (commented):
LGTM, the only important comment I have is regarding `just test` failing on my machine.

justfile
demosdemon: This is a "you should update your version of Rust" problem. All of the aws parts of `fwdctl` require 1.94.1 while firewood requires 1.94.0. Firewood's msrv is independent of that because the aws tooling is an optional feature of fwdctl. But, it deserves documenting

RodrigoVillar: Ah that makes sense, thanks for clarifying here!

RodrigoVillar (approved):

storage/src/linear/mod.rs
demosdemon: You're probably right. The clone binding was from before the io-uring module had a refactor. I'll take a look.

justfile
demosdemon: Added clarification in 3cc322e4e5dc8a1ee8de8d9fd28618bea6030ee2.

PR #2189, opened 2026-08-11
fix(proofs)!: bound the range-proof apply to the proven range

demosdemon:
## Why this should be merged

`Db::merge_key_value_range` is a total replacement: any key in its bounds that
is absent from the supplied key-values is deleted. The FFI apply path passed it
the caller's **requested** `end_key`, while verification only ever anchored at
the proof's own right edge. A truncated reply therefore deleted every local key
in `(proven_edge, end_key]` on the strength of a proof that never covered it.

#1970 established truncated coverage as a valid outcome rather than a
verification error, on the rationale that *"the caller is expected to compare
`key_values.last()` against their requested `last_key`."* Sound — but firewood's
own apply path is that caller, and it wasn't comparing. Verification had no way
to tell it what was proven: `verify_range_proof` computed the right edge, used
it, and returned `()`.

Forward-only sync never populates the affected window, which is why this stayed
latent. Repair, patch, and re-sync do.

## Notes for review

**The load-bearing invariant is `proven_end >= last_kv`, on every branch of
`proven_right_edge`.** That is what makes three otherwise-alarming things safe:
`MergeKeyValueIter` cannot over-apply despite the hardcoded `None` on its
key-values side; `find_next_key_after_range_proof`, resuming at
`key_values.last()`, cannot open a gap; and a bound that is too tight cannot
silently drop proven data.

Both failure directions are tested, and each test was verified to fail without
its fix. The under-deletion side — a bound *tighter* than justified — has no
natural trigger, so it is mutation-tested: force `proven_end()` artificially
tight and confirm the stale key survives.

## Behavior change

**`FindNextKey` can now return `nil` where it previously returned a range.** A
receiver that is genuinely caught up now reaches the target root, so the
existing "already at target" short-circuit fires where it structurally could not
before — the unbounded merge used to corrupt local state and mask it. A Go sync
driver that treats `nil` as unexpected will behave differently.

`TestRangeProofFindNextKey`'s assertion flips from "more to fetch" to `nil` for
that reason: it previously passed *because of* the bug.

Separately, `NextKeyRange` and its accessors documented the fetch range as
`[start, end)` when the implementation returns `(start, end]`. No code changes,
but a Go client that trusted those docs re-requests a key it already holds, or
skips one.

## Deliberately out of scope

`find_next_key_after_range_proof` still resumes from `key_values.last()` and
keeps its stale `TODO(#352)`. It is safe as-is: because `proven_end >= last_kv`,
resuming there re-requests a span already written and already proven — overlap,
never a gap. Worst case is one redundant round trip.

## Breaking Changes

`firewood`:

- `verify_range_proof` returns `ProvenRange` instead of `()`. External callers
  binding `Ok(())` must change. `ProvenRange` is `#[must_use]`, so discarding the
  proven bound — the defect this fixes — is now a compile error rather than
  silent data loss.
- `RangeProofVerificationContext` gains `right_edge_key` and becomes
  `#[non_exhaustive]`.

`firewood-ffi` / `firewood-go`: no signature changes, but a truncated proof
writes a narrower range than before, and `FindNextKey` may return `nil` where it
previously returned a range — see above.

### DIFF
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -58,7 +58,7 @@ typedef struct IteratorHandle IteratorHandle;
 typedef struct ProposalHandle ProposalHandle;
 
 /**
- * FFI context for for a parsed or generated range proof.
+ * FFI context for a parsed or generated range proof.
  */
 typedef struct RangeProofContext RangeProofContext;
 
@@ -815,7 +815,7 @@ typedef struct CreateRangeProofArgs {
   struct Maybe_BorrowedBytes start_key;
   /**
    * The end key of the range to prove. If `None`, the range ends at the end
-   * of the keyspace or until `max_length` items have been been included in
+   * of the keyspace or until `max_length` items have been included in
    * the proof.
    *
    * If provided, end key is inclusive if not truncated. Otherwise, the end
@@ -857,15 +857,9 @@ typedef struct VerifyRangeProofArgs {
    * The upper bound of the key range that the proof is expected to cover. If
    * `None`, the proof is expected to cover to the end of the keyspace.
    *
-   * This is an upper bound on what the proof *may* cover, not an assertion
-   * that it does. A responder may truncate its reply, in which case the
-   * proven range ends at the proof's own right edge instead. Verification
-   * accepts that as a valid partial result.
-   *
-   * The apply path writes over the **proven** range, so keys between the
-   * proven edge and this bound are left untouched rather than deleted. The
-   * caller owns the remainder: use `fwd_range_proof_find_next_key` to get the
-   * next range to request, and keep going until it reports nothing left.
+   * This is a ceiling, not an assertion: a proof carrying a key above this
+   * bound is invalid, but a truncated proof may prove less and anchor at
+   * its own right edge instead.
    */
   struct Maybe_BorrowedBytes end_key;
   /**
@@ -2964,7 +2958,7 @@ struct RangeProofResult fwd_range_proof_from_bytes(BorrowedBytes bytes);
  *
  * - [`ValueResult::NullHandlePointer`] if the caller provided a null pointer.
  * - [`ValueResult::Some`] containing the serialized bytes if successful.
- * - [`ValueResult::Err`] if the caller provided a null pointer.
+ * - [`ValueResult::Err`] containing an error message if serialization panicked.
  */
 struct ValueResult fwd_range_proof_to_bytes(const struct RangeProofContext *proof);
 
--- a/ffi/firewood.h
+++ b/ffi/firewood.h
@@ -857,8 +857,15 @@ typedef struct VerifyRangeProofArgs {
    * The upper bound of the key range that the proof is expected to cover. If
    * `None`, the proof is expected to cover to the end of the keyspace.
    *
-   * This is ignored if the proof is truncated and does not cover the full,
-   * in which case the upper bound key is the final key in the key-value pairs.
+   * This is an upper bound on what the proof *may* cover, not an assertion
+   * that it does. A responder may truncate its reply, in which case the
+   * proven range ends at the proof's own right edge instead. Verification
+   * accepts that as a valid partial result.
+   *
+   * The apply path writes over the **proven** range, so keys between the
+   * proven edge and this bound are left untouched rather than deleted. The
+   * caller owns the remainder: use `fwd_range_proof_find_next_key` to get the
+   * next range to request, and keep going until it reports nothing left.
    */
   struct Maybe_BorrowedBytes end_key;
   /**
--- a/ffi/proofs.go
+++ b/ffi/proofs.go
@@ -113,9 +113,10 @@ type ChangeProof struct {
 	handle *C.ChangeProofContext
 }
 
-// NextKeyRange represents a range of keys to fetch from the database. The start
-// key is inclusive while the end key is exclusive. If the end key is Nothing,
-// the range is unbounded in that direction.
+// NextKeyRange represents a range of keys to fetch from the database,
+// `(startKey, endKey]`: the start key is exclusive because it has already been
+// synchronized, and the end key is inclusive. If the end key is Nothing, the
+// range is unbounded in that direction.
 type NextKeyRange struct {
 	startKey *ownedBytes
 	endKey   Maybe[*ownedBytes]
@@ -201,14 +202,12 @@ func (p *RangeProof) Verify(
 // call to [*Database.VerifyAndCommitRangeProof] will skip verification and commit the
 // prepared proposal.
 //
-// The prepared proposal only covers the range the proof actually proves,
-// which can be narrower than [endKey]: a responder is allowed to truncate
-// its reply, in which case the proof anchors at its own right edge instead
-// of the requested one. Keys beyond that proven edge are left untouched
-// rather than deleted. A sync-client caller must keep calling
-// [*RangeProof.FindNextKey] and requesting, verifying, and applying the
-// ranges it returns until it returns nil, to be sure the full [startKey,
-// endKey] span has actually been synchronized.
+// The proposal replaces state across the range the proof proves: any existing
+// key in that range that the proof does not carry is deleted. The range runs
+// from [startKey] to the proof's right edge, which is [endKey] when the
+// responder covered the whole request and a smaller key when it truncated.
+// Keys past that edge are left as they are; [*RangeProof.FindNextKey] reports
+// where to resume.
 //
 // Because this method prepares a proposal, the database must be kept alive
 // until the proof is committed or freed.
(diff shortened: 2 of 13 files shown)


ffi/src/proofs/range.rs
```diff
@@ -58,8 +58,15 @@ pub struct VerifyRangeProofArgs<'a, 'db> {
...
-    /// in which case the upper bound key is the final key in the key-value pairs.
+    /// This is an upper bound on what the proof *may* cover, not an assertion
+    /// that it does. A responder may truncate its reply, in which case the
+    /// proven range ends at the proof's own right edge instead. Verification
+    /// accepts that as a valid partial result.
+    ///
+    /// The apply path writes over the **proven** range, so keys between the
+    /// proven edge and this bound are left untouched rather than deleted. The
+    /// caller owns the remainder: use `fwd_range_proof_find_next_key` to get the
+    /// next range to request, and keep going until it reports nothing left.
```
rkuris: This is a bit wordy and ambiguous. Which proven edge? The right one. And nobody would suppose that we delete any keys past the proven edge, just like we don't delete keys past `end_key` so that's not all that helpful. Let's focus on what this variable is actually for instead of how the algorithm works outside of this variable, and move those comments there.

Edited: it's already there, this is also duplicative to the actual code that does it.

```suggestion
    /// The requested last key. The proof may not cover keys before this
    /// if it was truncated. A proof containing a key beyond end_key (or the proven right edge)
    /// is invalid.
```

ffi/src/proofs/range.rs
```diff
@@ -144,6 +151,27 @@ impl<'db> RangeProofContext<'db> {
...
+    /// `(right_edge_key, end_key]` carries no proof — merging over the
+    /// caller's requested `end_key` instead of this bound would delete local
+    /// keys the proof never mentioned.
+    ///
+    /// `right_edge_key == None` legitimately means "proven to the end of the
+    /// keyspace", so this must not fall back to `None` when the verification
+    /// context is missing: that would authorise deleting the whole tail
+    /// instead of correctly reporting an unbounded proven range. A successful
+    /// [`Self::verify`] guarantees the context is populated, which is why
+    /// this asserts with `expect` rather than degrading to that default.
```
rkuris: Same issue here. This comment is talking about what the caller's responsibility is and addressing the bug being fixed instead of describing what `proven_end` is actually doing.
```suggestion
    /// Returns the upper bound of the range the proof actually proves.
    ///
    /// Returns `None` when the proof is proven to the end of the keyspace.
    /// Panics
    /// This function panics unless [`Self::verify`] is called first.
```

ffi/src/proofs/range.rs
```diff
@@ -217,8 +247,9 @@ impl<'db> RangeProofContext<'db> {
             && allow_rebase
         {
             // proposal is stale, try rebasing and committing again
+            let proven_end = self.proven_end();
             let proposal_handle = db
-                .merge_key_value_range(start_key, end_key, self.proof.key_values())?
+                .merge_key_value_range(start_key, proven_end, self.proof.key_values())?
```
rkuris, line 240: nit: A comment as to why we choose proven_end here would be okay (but I'm not sure we need it, it seems somewhat readable as-is). Your choice.

ffi/firewood.h
```diff
@@ -857,8 +857,15 @@ typedef struct VerifyRangeProofArgs {
...
-   * in which case the upper bound key is the final key in the key-value pairs.
+   * This is an upper bound on what the proof *may* cover, not an assertion
+   * that it does. A responder may truncate its reply, in which case the
+   * proven range ends at the proof's own right edge instead. Verification
+   * accepts that as a valid partial result.
+   *
+   * The apply path writes over the **proven** range, so keys between the
+   * proven edge and this bound are left untouched rather than deleted. The
+   * caller owns the remainder: use `fwd_range_proof_find_next_key` to get the
+   * next range to request, and keep going until it reports nothing left.
```
rkuris: This comment seems identical, somewhat tautological, and again talks about how the variable is used elsewhere.

ffi/proofs.go
```diff
@@ -250,6 +259,12 @@ func (db *Database) VerifyRangeProof(
 // [rootHash]. If the proof is valid, it is committed to the database and the
 // new root hash is returned. The resulting root hash may not equal the
 // provided root hash if the proof was truncated due to [maxLength].
+//
+// The commit only applies the range the proof actually proves, which can be
+// narrower than [endKey] when the responder truncated its reply; keys past
+// the proven edge are left untouched rather than deleted. The caller owns
+// the remainder: keep calling [*RangeProof.FindNextKey] and requesting,
+// verifying, and committing the ranges it returns until it returns nil.
```
rkuris: This is the right place for the comment, but it's still a bit backwards, instead describing the bug you're fixing instead of the actual behavior. The algorithm for how to do a state sync probably belongs elsewhere.

```suggestion
// Any existing key within the proven range (typically from `start_key` to the end of
// the proven range provided by [...::proven_end]) that is not in the range
// proof is deleted.
```

ffi/proofs_test.go
```diff
@@ -314,23 +315,80 @@ func TestRangeProofFindNextKey(t *testing.T) {
 	// Verify the proof
 	r.NoError(db.VerifyRangeProof(proof, nothing(), nothing(), root, rangeProofLenTruncated))
 
-	// Now FindNextKey should work
+	// This proof was taken from db's own current state and verified against
```
rkuris: I didn't review this comment but on the surface it has the same class of problem as above.

ffi/proofs_test.go
```diff
@@ -314,23 +315,80 @@ func TestRangeProofFindNextKey(t *testing.T) {
...
-	// FindNextKey should still work after commit
 	nextRange, err = proof.FindNextKey()
 	r.NoError(err)
+	r.Nil(nextRange)
+}
+
+// TestRangeProofFindNextKeyDivergentReceiver covers the case
+// TestRangeProofFindNextKey no longer can: a receiver genuinely behind the
+// proof's target root, rather than one self-proving state it already has.
+// A truncated proof applied to an empty target must report more to fetch.
```
rkuris: nit: wordy (but not horrible, so at your discretion)

```suggestion
// Test receivers beyond the proof's target
// A truncated proof applied to an empty target must report more to fetch.
```

ffi/proofs_test.go
```diff
@@ -314,23 +315,80 @@ func TestRangeProofFindNextKey(t *testing.T) {
...
+func TestRangeProofFindNextKeyDivergentReceiver(t *testing.T) {
+	r := require.New(t)
+
+	dbSource := newTestDatabase(t)
+	dbTarget := newTestDatabase(t)
+
+	// kvForTest sorts keys (and their paired vals) lexicographically, so
+	// keys[:rangeProofLenTruncated] are exactly the trie-order prefix the
+	// truncated proof below proves.
```
rkuris: Let's move this to the comments in kvForTest, or better yet, rename this function: `testSortedKvs` then this comment is mostly tautological.

rkuris (commented):
These comments are very verbose and could use some TLC before merging. Please audit for repetitiveness and locality.

As usual, specific comments marked "nit" are optional to fix or comment and may be ignored (they are optionally actionable). The remaining ones require some action: either take the suggestion or explain why the suggestion isn't better.

ffi/proofs_test.go
demosdemon: `kvForTest` has ~80 callsites. I'll skip the rename for now but will follow up.

ffi/firewood.h
demosdemon: firewood.h headers duplicate the docs from rust. this file is auto-generated.

ffi/src/proofs/range.rs
demosdemon: Updated docs on `merge_key_value_range` to explain what should be provided.

firewood/src/merkle/tests/mod.rs
```diff
@@ -35,7 +35,7 @@ fn verify_range_proof<H: ProofCollection<Node = ProofNode>>(
     last_key: Option<impl KeyType>,
     root_hash: &TrieHash,
     proof: &RangeProof<impl KeyType, impl ValueType, H>,
-) -> Result<(), Error> {
+) -> Result<ProvenRange, Error> {
```
AminR443: nit: only `test_dropped_trailing_key_accepted_as_partial_coverage` uses the returned `ProvenRange` and all call sites had to add `let _ =` because of the `must_use`. I wonder if it would be cleaner to keep this as `()`, and call the wrapped function directly in the test that uses it.

AminR443 (approved):

firewood/src/merkle/tests/mod.rs
demosdemon: That's fair. I originally didn't do that because this function doesn't exist at tip of @RodrigoVillar ethhash merge stack.

PR #2191, opened 2026-08-12
fix: add Owner tag to EC2 instances

RodrigoVillar:
## Why this should be merged

Trying to launch an EC2 instance with `fwdctl launch deploy` fails because SecOps requires AWS resources (EC2 instances and EBS volumes, in this case) to have an `Owner` tag.

## How this works

Adds an `Owner` tag to resources deployed by `fwdctl launch deploy`.

## How this was tested

Ran `fwdctl launch deploy` and observed the benchmark run. Doing so off the `main` branch of Firewood currently fails.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/fwdctl/src/launch/ec2_util.rs
+++ b/fwdctl/src/launch/ec2_util.rs
@@ -28,6 +28,8 @@ const UBUNTU_NOBLE_AMI_NAME_PATTERN: &str =
     "ubuntu/images/hvm-ssd-gp3/ubuntu-noble-24.04-{arch}-server-*";
 /// Root EBS volume size (GiB) for launched benchmark instances.
 const ROOT_VOLUME_SIZE_GIB: i32 = 50;
+/// Tag key identifying the AWS caller responsible for the instance.
+const OWNER_TAG_KEY: &str = "Owner";
 /// Tag key used to identify instances managed by `fwdctl`.
 const MANAGED_BY_TAG_KEY: &str = "ManagedBy";
 /// Tag value used to identify instances managed by `fwdctl`.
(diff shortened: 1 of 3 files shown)


fwdctl/src/launch/ec2_util.rs
```diff
@@ -161,6 +164,12 @@ pub async fn launch_instance(
                 .set_tags(Some(tags))
                 .build(),
         );
+    request = request.tag_specifications(
```
demosdemon: This can be chained fluent style. There's no need to rebind this. We also need to tag network interfaces.

fwdctl/src/launch/ec2_util.rs
```diff
@@ -209,6 +223,7 @@ fn build_tags(opts: &DeployOptions, instance_name: &str, username: &str) -> Vec<
     let mut tags = vec![
```
demosdemon, line 228: Any reason we wouldn't want to apply all of these tags?

fwdctl/src/launch/ec2_util.rs
RodrigoVillar: Good catch regarding chaining - I've gone ahead and refactored the code + also tagged network interfaces here: [37a2480](https://github.com/ava-labs/firewood/pull/2191/commits/37a2480e9a2736a243077cf79d3da279a27c03ca).

fwdctl/src/launch/ec2_util.rs
RodrigoVillar: Applying the owner tag to the resource instance/volume was the minimum code change needed for `fwdctl launch deploy` to work again. However, as you've alluded to, there's no reason to not be consistent with what labels we're applying to the resources; I've gone ahead and applied all tags to all resources here: [d5bf6ee](https://github.com/ava-labs/firewood/pull/2191/commits/d5bf6ee80e2a69c510ddffe528facc09131e290b).

demosdemon (approved):

PR #2192, opened 2026-08-12
ci: derive the benchmark matrix from cargo metadata

demosdemon:
## Why this should be merged

`benchmarks.yaml` hand-enumerated three `cargo bench` steps. The workspace declares six bench targets, so `proposal_reconstruct`, `rlp`, and `triehash` had never run in CI. A hand-maintained list drifts silently — nothing fails when a new `[[bench]]` is added and never wired up. Deriving the list from `cargo metadata` closes that gap and lets the targets run in parallel.

## How this works

**`--workspace --bench <name>`, not `-p <package> --bench <name>`.** `-p firewood-triehash --features ethhash,logger,test_utils` is a hard error — that package declares no features at all. Workspace scoping applies each feature to whichever members declare it, so one feature string covers all six targets and no per-package feature computation is needed. It is also the cache-correct choice: Cargo resolves dependency features from the package selection, so `-p` resolves a narrower set than the `--workspace --benches` warming build and recompiles behind a warm tree. Measured under `maxperf`:

| scoping                             | units recompiled |
| ----------------------------------- | ---------------- |
| `--workspace --bench <name>`        | 0, every target  |
| `-p firewood` (first bench)         | 80               |
| `-p firewood-storage` (first bench) | 14               |
| `-p firewood-triehash`              | 28               |

Later benches in a package reuse the narrowed build, but each CI leg restores the cache cold and pays its package's first-invocation cost.

Name-only selection requires bench names to be unique workspace-wide, so `list-bench-targets.sh` asserts it rather than leaving a future duplicate to fail confusingly. Singular `--bench` also errors when a target's `required-features` are unmet, where the plural `--benches` glob silently skips it — the stricter behavior is what we want.

`list-bench-targets.sh` needs only cargo and jq. Because `--no-deps` reads just the workspace manifests, it runs with an empty `CARGO_HOME`, offline, and without `Cargo.lock`. `--frozen` therefore asserts nothing here; it is retained only so the invocation stays offline if it ever grows a form that does resolve dependencies.

**The cache-warming `build` job runs alongside the matrix rather than gating it.** `Swatinem/rust-cache`'s `SAVE_TARGETS` is `{lib, proc-macro}`, so bench binaries and `target/criterion` are pruned before the tarball is written; a leg restores the dependency rlibs and relinks its own binary whether or not `build` ran first. Gating bought the legs nothing while putting a full workspace build on the critical path — measured on CI, a 144s build step ahead of legs that each took 78–98s, roughly half of a 6m14s run. Ungated, `build` refreshes the cache for subsequent runs while the current run's legs proceed on the previous run's cache. It is skipped entirely on a filtered dispatch, where `--workspace --benches` would build all six targets to serve a subset and there is no cache obligation to meet. `benchmarks-required` accordingly tolerates a skipped `build` but still fails on a skipped `bench`, which is how an empty matrix surfaces.

Behavior changes:

- `serializer` and `proposal_reconstruct` now compile with `test_utils` at bench time. Every `cfg(feature = "test_utils")` site is additive — nothing is conditional inside an existing function body. The old Build step already compiled these targets with `test_utils` and the bench step then rebuilt them without it, so this removes a redundant compile rather than adding one.
- Six artifacts instead of one: `upload-artifact` has been immutable since v4, so a shared name across matrix legs fails the run. A merge job recombines them into `benchmark-results`. Nothing in the repo consumes that artifact — `track-performance.yml` reads a different file.
- `ci-rust` becomes variadic. Existing callers pass exactly two arguments and only `bench` takes a trailing one, so this is backward compatible, but extra arguments to other commands are now rejected rather than silently dropped.

## How this was tested

Exercised on CI as both a full sweep and a `bench-filter` subset, which is where the build/leg timings above come from and how the gating change was validated.

Locally, `build-benches` compiles all six targets under the new `maxperf-ethhash-logger-test_utils` profile, with freshness verified per target (table above), and `triehash`, `rlp` (73s), and `proposal_reconstruct` (85s) run end-to-end through `just ci-rust bench …`. `timeout-minutes` are headroom rather than tight bounds, since a cold leg is dominated by the fat-LTO build and not by any individual benchmark.

`list-bench-targets.sh` is covered for matrix shape and single-line output, filter hits, a filter matching nothing (`rc=1`), and the duplicate-name assertion (`rc=1`, exercised against synthetic metadata). The `benchmarks-required` aggregator is covered for all-green, filtered-with-`build`-skipped, `build` failure, empty matrix, leg failure, and `discover` failure. `actionlint`, `shellcheck`, and `markdownlint-cli2` are clean.

## Breaking Changes

CI and developer tooling only; no library, FFI, or CLI surface is affected.

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -20,6 +20,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   # Benchmark results are the point of the run, so do not discard a merged
   # commit's results just because a newer commit arrived.
+  queue: ${{ github.event_name == 'workflow_dispatch' && 'single' || 'max' }}
   cancel-in-progress: false
 
 env:
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -119,7 +119,7 @@ jobs:
       - name: Benchmark
         env:
           BENCH: ${{ matrix.bench }}
-        run: ./scripts/run-just.sh ci-rust bench "$PROFILE_KEY" "$BENCH"
+        run: ./scripts/run-just.sh ci-rust-bench "$PROFILE_KEY" "$BENCH"
       - name: Upload benchmark results
         uses: actions/upload-artifact@v7
         with:
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -4,36 +4,180 @@ on:
   push:
     branches: [main]
   workflow_dispatch:
+    inputs:
+      bench-filter:
+        description: >-
+          Optional regex matched against bench target names (e.g. `proofs`, or
+          `^(rlp|triehash)$`). Leave blank to run every bench target. A filter
+          that matches nothing fails the run.
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Benchmark results are the point of the run, so do not discard a merged
+  # commit's results just because a newer commit arrived.
+  cancel-in-progress: false
 
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: full
+  # One value serving two purposes: the profile name understood by
+  # scripts/run-rust-ci.sh, and the Swatinem/rust-cache shared-key, so the
+  # warming job and the matrix legs cannot drift apart.
+  PROFILE_KEY: maxperf-ethhash-logger-test_utils
 
 jobs:
-  bench:
+  # Derives the bench matrix from cargo metadata rather than a hand-maintained
+  # list, so adding a `[[bench]]` target needs no change here.
+  discover:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      matrix: ${{ steps.list.outputs.matrix }}
     steps:
       - uses: actions/checkout@v6
+      - uses: taiki-e/install-action@67729d5c413db75907f0ad1e39bb04b9c868ff60 # v2.85.7
+        with:
+          tool: just@1.46.0
+      # The script runs `cargo metadata`, so it needs a toolchain. It does not
+      # need rust-cache or `cargo fetch`: `--no-deps` reads only the workspace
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmarks.yaml
```diff
@@ -4,36 +4,180 @@ on:
...
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Benchmark results are the point of the run, so do not discard a merged
+  # commit's results just because a newer commit arrived.
+  cancel-in-progress: false
```
RodrigoVillar, line 24: (AI-generated comment) cancel-in-progress: false only protects the currently running workflow. By default, a newer run still replaces an existing pending run in this concurrency group, so rapid pushes can discard a merged commit’s benchmark results. Since the comment says every merged commit should be retained, should this use queue: max or avoid grouping the runs?

justfile
```diff
@@ -2,9 +2,13 @@
 default:
     ./scripts/run-just.sh --list
 
-# Run a Rust command with a named CI profile.
-ci-rust command profile:
-    ./scripts/run-rust-ci.sh "{{command}}" "{{profile}}"
+# Run a Rust command with a named CI profile; only `bench` takes a trailing arg.
+ci-rust command profile *args:
+    ./scripts/run-rust-ci.sh "{{command}}" "{{profile}}" {{args}}
```
RodrigoVillar: Rather than having an edge case for `bench`, I think it would be cleaner to instead add another recipe for the `bench` command. That is:

```
# Preserve the existing uniform interface.
ci-rust command profile:
    ./scripts/run-rust-ci.sh "{{command}}" "{{profile}}"

# Run one benchmark target.
ci-rust-bench profile target:
    ./scripts/run-rust-ci.sh bench "{{profile}}" "{{target}}"
```

RodrigoVillar (approved):

justfile
demosdemon: I would rather collapse it into a single `{{args}}` and let the shell script handle it. Your suggestion is also a special case handling of the same script.

demosdemon: I've come to realize this better even if it is a bit bespoke because there's no need for the var args in the justfile.

PR #2194, opened 2026-08-12
refactor(proofs)!: privatize verification-context fields

demosdemon:
## Why this should be merged

A verification context is meant to be evidence that verification happened. Neither
of ours could carry that weight, and they failed differently:

- `RangeProofVerificationContext` already had `#[non_exhaustive]`, so it could not
  be *constructed* externally — but every field was `pub`, so a caller holding a
  legitimately obtained context could overwrite `root` or `right_edge_key` and hand
  it back alongside a different proof.
- `ChangeProofVerificationContext` had no such attribute. It was constructible from
  whole cloth via a struct literal, with no verification call involved at all.

Two sibling types with divergent extensibility contracts, one of them accepting
forgeries. This closes both holes with one shape.

The motivation is concrete rather than hygienic: hole detection (#352) adds public
entry points that take a context as a *witness* and skip re-verifying what it
attests. Those signatures cannot go `pub` while a context is forgeable.

## Notes for review

Accessors return borrows except where the type is `Copy`. `TrieHash` derives
`Clone` but not `Copy`, so borrowing avoids a 32-byte copy per read — and that
choice is why four `as_deref()` calls *disappear* at call sites rather than
migrating: `Option<&[u8]>` is what consumers already wanted.

The mutation surface really is closed, not just narrowed: workspace-wide these
types have exactly two `impl` blocks between them, with no `Clone`, `Default`,
`serde`, `From`, `Deref`, or `&mut` accessor. An external holder can read or drop,
and nothing else.

### What this does *not* establish

A context is not bound to the proof that produced it. Both entry points take the
proof and the context as independent arguments, so a caller can verify proof A,
obtain a perfectly legitimate context, and pass it alongside proof B. After this PR
a context witnesses that *some* verification ran with *those parameters* — not that
*this proof* was verified.

That is likely sufficient for hole detection, since the parameters are what bound
the range. It is still narrower than "witness" suggests, so the hole-detection PR
has to pick a resolution deliberately: pair them in one owning type (as
`RangeProofContext` already does on the FFI side), re-derive the binding by checking
`ctx.root()` against the proof, or define a mismatched pair as a caller error.
Recorded here so that decision is made rather than assumed.

### Rejected alternative

Three tests in `merkle/tests/change/attack.rs` build a context by struct literal on
purpose — they drive `verify_change_proof_root_hash` with proofs engineered to
*fail* structural validation, so the real constructor is unreachable by test design.
They now go through a `#[cfg(test)] pub(crate)` escape hatch.

`pub(crate)` fields were the obvious alternative and were rejected: they exist in
*every* build, which would let any in-crate module mutate a context — including the
hole-detection code arriving later in this same stack. That relocates the hazard
inside the crate boundary instead of removing it, and lands it exactly where the
future consumers live. The `#[cfg(test)]` hatch adds zero forgery surface outside
`cargo test`.

### Downstream impact beyond the field access

External consumers lose one legitimate capability: destructuring
`ChangeProofVerificationContext` to move its four boxed keys out zero-copy, which
was possible only because it lacked `#[non_exhaustive]`. Borrows cover every in-tree
use, and an `into_parts()` accessor remains a purely additive, non-breaking addition
if anyone needs it.

No C-ABI change. Neither struct is `#[repr(C)]`, both sit behind opaque FFI handles,
and the `ffi/` changes are Rust-internal — the Go binary contract is untouched.

## How this was tested

Mechanical refactor with no behavior change, so no new tests: the gate is that
existing coverage passes unchanged, with test counts measured before and after
rather than assumed. `just prepush` green; Go suite green in both hash modes.

Two things worth noting about the verification rather than the code:

- A missed reader is a compile error by construction, so a clean
  `cargo check --workspace --all-targets --all-features` is dispositive for
  completeness here, not merely suggestive.
- The rebase onto `HashOrRlp`-folded-into-`HashType` (#2150) forced a conflict
  inside this diff, at the change-proof root-hash equality check. The resolution
  compares `HashType` against `TrieHash` directly instead of converting up first.
  That is behavior-preserving, and the `Rlp` arm is why it needed checking: both
  `PartialEq` impls apply the same Keccak normalization to the mixed case, so no
  outcome changes. Confirmed by the ethhash profile and by the `attack.rs` tests
  that traverse that exact comparison.

## Breaking Changes

- [x] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

BREAKING CHANGE: `RangeProofVerificationContext` and
`ChangeProofVerificationContext` no longer expose public fields. Read them through
the new accessors: `root()` / `end_root()`, `start_key()`, `end_key()`,
`right_edge_key()`, and `max_length()` on the range context.
`ChangeProofVerificationContext` is additionally `#[non_exhaustive]`.

### DIFF
--- a/firewood/src/proofs/change.rs
+++ b/firewood/src/proofs/change.rs
@@ -196,7 +196,6 @@ where
 /// Constructible only by [`verify_change_proof_structure`]; fields are private
 /// so a context cannot be forged or altered after verification.
 #[derive(Debug)]
-#[non_exhaustive]
 pub struct ChangeProofVerificationContext {
     end_root: HashKey,
     start_key: Option<Box<[u8]>>,
--- a/firewood/src/proofs/change.rs
+++ b/firewood/src/proofs/change.rs
@@ -190,11 +190,13 @@ where
 // ── Change proof verification ──────────────────────────────────────────────
 
 /// Verification context captured after structural validation of a change proof.
-/// Stored so that downstream logic (root hash verification, `find_next_key`) can
-/// reference the original verification parameters without re-validating.
+/// Stored so that downstream logic (root hash verification,
+/// [`find_next_key_after_change_proof`]) can reference the original
+/// verification parameters without re-validating.
 ///
-/// Constructible only by [`verify_change_proof_structure`]; fields are private
-/// so a context cannot be forged or altered after verification.
+/// Outside of test builds, constructible only by
+/// [`verify_change_proof_structure`]; fields are private so a context cannot be
+/// forged or altered after verification.
 #[derive(Debug)]
 pub struct ChangeProofVerificationContext {
     end_root: HashKey,
--- a/firewood/src/proofs/change.rs
+++ b/firewood/src/proofs/change.rs
@@ -192,14 +192,37 @@ where
 /// Verification context captured after structural validation of a change proof.
 /// Stored so that downstream logic (root hash verification, `find_next_key`) can
 /// reference the original verification parameters without re-validating.
+///
+/// Constructible only by [`verify_change_proof_structure`]; fields are private
+/// so a context cannot be forged or altered after verification.
 #[derive(Debug)]
+#[non_exhaustive]
 pub struct ChangeProofVerificationContext {
+    end_root: HashKey,
+    start_key: Option<Box<[u8]>>,
+    end_key: Option<Box<[u8]>>,
+    right_edge_key: Option<Box<[u8]>>,
+}
+
+impl ChangeProofVerificationContext {
     /// The expected root hash of the ending revision.
-    pub end_root: HashKey,
+    #[must_use]
+    pub const fn end_root(&self) -> &HashKey {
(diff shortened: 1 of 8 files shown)


firewood/src/proofs/change.rs
```diff
@@ -192,14 +192,37 @@ where
 /// Verification context captured after structural validation of a change proof.
 /// Stored so that downstream logic (root hash verification, `find_next_key`) can
 /// reference the original verification parameters without re-validating.
+///
+/// Constructible only by [`verify_change_proof_structure`]; fields are private
+/// so a context cannot be forged or altered after verification.
 #[derive(Debug)]
+#[non_exhaustive]
```
rkuris: I don't think this is needed. #[non_exhaustive] is a no-op inside the defining crate, it's only used by external crates.

rkuris (approved):

PR #2197, opened 2026-08-17
feat(ffi): proof wire compression config variables

AminR443:
## Why this should be merged
Adds a config knob for proof decode limits (`max_decompressed_len`,`max_compression_ratio`) which we'll use to enforce zstd compression limits in #2181. It has no consumers yet, so it can land first.

## How this works
`ProofConfig` (two `usize` limits, defaults 32 MiB / 128) is added to `DbConfig`, stored on `Db`, and read via `Db::proof_config()`.
  
## How this was tested
TBD

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


firewood/src/db.rs
```diff
@@ -124,6 +143,9 @@ pub struct DbConfig {
     /// Whether to enable `RootStore`.
     #[builder(default = false)]
     pub root_store: bool,
+    /// Size limits applied when serializing/deseralizing proofs.
```
RodrigoVillar:
```suggestion
    /// Size limits applied when serializing/deserializing proofs.
```

ffi/firewood.go
```diff
@@ -185,6 +189,15 @@ func WithNodeCacheSizeInBytes(sizeInBytes uint) Option {
 	}
 }
 
+// WithProofSizeLimits sets the size limits applied when serializing and
```
RodrigoVillar: q: would it make sense to specify the default used here? The other `WithX()` options do so - ref:

```go
// WithNodeCacheSizeInBytes sets the node cache memory limit in bytes.
// Must be non-zero.
// Default: 128 MB
func WithNodeCacheSizeInBytes(sizeInBytes uint) Option {}
```

firewood/src/db.rs
RodrigoVillar: q: the documentation here says that this applies when `serializing/deserializing proofs` but the documentation for `ProofConfig` seems to scope this more narrowly to just deserialization:

```rs
/// Size limits governing how a proof is deserialized.
#[derive(Clone, Copy, TypedBuilder, Debug)]
#[non_exhaustive]
pub struct ProofConfig {}
```

Does the configuration in this PR apply to just deserialization or also serialization as well?

AminR443: Good catches. the "serializing" part was lingering from sized proofs which I changed my mind about it.

ffi/firewood.go
AminR443: good idea.

firewood/src/db.rs
```diff
@@ -177,10 +200,17 @@ impl Db {
         let db = Self {
             manager,
             use_parallel: cfg.use_parallel,
+            proof_config: cfg.proof,
         };
         Ok(db)
     }
 
+    /// The size limits applied when serializing/deserializing proofs.
```
RodrigoVillar:
```suggestion
    /// The size limits applied when deserializing proofs.
```

From my last review pass, I think we agreed that this was just for deserializing proofs?

ffi/src/handle.rs
```diff
@@ -106,9 +106,26 @@ pub struct DatabaseHandleArgs<'a> {
...
+
+    /// The hard cap on a decoded proof body. `0` uses the default.
+    pub proof_max_decompressed_len: usize,
+
+    /// The max ratio of uncompressed body to compressed frame length. `0` uses the default.
+    pub proof_max_compression_ratio: usize,
 }
 
 impl DatabaseHandleArgs<'_> {
+    fn as_proof_config(&self) -> ProofConfig {
```
rkuris: If you keep this, this method (and arguably the next one) should be named `to_proof_config`. See https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv for the reasoning. This is a borrowed->owned conversion.

But, I'd much rather prefer to keep this: as_proof_config should return the `ProofConfig` that should be embedded inside `DatabaseHandleArgs`, not as bespoke separate params. This way, if we add a new one, we add it in one place and much of these conversions vanish.

ffi/src/handle.rs
```diff
@@ -106,9 +106,26 @@ pub struct DatabaseHandleArgs<'a> {
...
+        let mut cfg = ProofConfig::default();
+        if self.proof_max_decompressed_len != 0 {
+            cfg.max_decompressed_len = self.proof_max_decompressed_len;
+        }
```
rkuris: This doesn't seem very idiomatic. Zero seems to have special meaning here; this should be a `NonZero` type combined with an `Option` at a minimum. Better yet, just embed here.

rkuris (changes requested):

ffi/src/handle.rs
AminR443: Fixed

ffi/src/handle.rs
AminR443: Fixed

firewood/src/db.rs
AminR443: Fixed.

demosdemon (approved):

AminR443:
Based on discussions we had, we decided to go with global consts and later make them config variables if necessary.

PR #2198, opened 2026-08-17
feat(merkle): generating sized change/range proofs

AminR443:
## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


firewood/src/merkle/sized.rs
```diff
@@ -0,0 +1,316 @@
...
+//! wire size targets the specified byte budget.
+
+#![expect(
+    clippy::cast_precision_loss,
+    clippy::cast_possible_truncation,
+    clippy::cast_sign_loss,
+    reason = "byte/key counts (well below 2^52) are converted to and from f64 \
+              for the compression-ratio estimate, which is re-checked against \
+              the exact serialized length."
+)]
```
rkuris: This should not be needed. If you absolutely cannot work around this, then putting this expect near the error is preferred.

firewood/src/merkle/sized.rs
```diff
@@ -0,0 +1,316 @@
...
+const DEFAULT_COMPRESSION_RATIO: f64 = 0.52;
+/// Ratio hints and measurements are clamped into `RATIO_MIN..=RATIO_MAX`
+const RATIO_MIN: f64 = 0.05;
+const RATIO_MAX: f64 = 2.0;
+/// Cap on ratio-correction (grow) passes.
+const MAX_GROW: usize = 6;
+
+/// A sized proof `P` with its compressed `wire` bytes.
+#[derive(Debug)]
+pub struct SizedProof<P> {
```
rkuris: Curious why we're increasing the public space with SizedProof. Does someone outside of firewood need this? Even the mod import uses pub and this exposes this whole struct and each member.

rkuris:
Not a complete review, please re-request review once this is further along.

Naming:
"DEFAULT_COMPRESSION_RATIO"  is "INITIAL_COMPRESSION_ESTIMATE". It's not a default, it's a guess as to what we think we'll get. ACCEPT_FLOOR is maybe SUFFICIENT_FILL_FRACTION. It's not actually a floor. RATIO_MIN isn't actually a min, it's an expected min so maybe MIN_EXPECTED_RATIO (same with MAX). MAX_GROW is maybe MAX_RATIO_CORRECTION_PASSES.

Abstractions:
Create a newtype (CompressionRatio) to avoid all the casting and give you a domain specific type. All the clamping, math, and other stuff belongs in this type, not scattered through the function. Also lets you localize these constants to this new type. Implement this with integer math, storing everything with a scaling factor (yet another constant) so "1" in the CompressionRatio means 0.00001 or something.

PR #2204, opened 2026-08-17
perf(storage): use a constant for the empty-trie root instead of rehashing it

AminR443:
## Why this should be merged

`keccak256(NULL_RLP)` is the empty-trie root, but it was recomputed on every leaf hash. This PR defines it as a const to prevent recomputing it.

## How this works

Adds `rlp::EMPTY_TRIE_ROOT`, a `const [u8; 32]`, and replaces the two `Keccak256::digest(NULL_RLP)` call sites with it.

## How this was tested
UT.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


storage/src/rlp.rs
```diff
@@ -14,6 +14,12 @@ use thiserror::Error;
 /// RLP encoding of an empty byte string (`0x80`).
 pub const NULL_RLP: &[u8] = &[0x80];
 
+/// `keccak256(NULL_RLP)` — the root of an empty trie.
+pub const EMPTY_TRIE_ROOT: [u8; 32] = [
```
demosdemon, line 19: please include the hash as a comment as well. it makes it easy to grep for the hash if it's as one hex str.

AminR443: Fixed.

demosdemon (approved):

PR #2209, opened 2026-08-24
refactor(ffi): make the code-hash iterator's proof borrow structural

demosdemon:
## Why this should be merged

#2143 and #2144 each fixed a real use-after-free, and each proved it with a test
that forced up to 50 collections per yielded hash and asserted a
`runtime.AddCleanup` canary had not fired. Those tests could not establish what
they claimed:

- Asserting the GC did *not* collect something has no natural bound. The `50`
  traded runtime for confidence; no value is ever *correct*.
- Nothing verified the canary could fire at all, so one that silently stopped
  working would pass forever.
- The probe lived inside the `for … range seq` body with no yield count
  asserted, so a fixture that stopped producing code hashes would also pass,
  having asserted nothing.

The constant was really approximating a runtime rule the test never stated: the
proof carries both a `Free` finalizer and the cleanup, and `runtime.AddCleanup`
only runs a cleanup *"once it has been finalized and becomes unreachable without
an associated finalizer"* — at least two cycles plus two goroutine hops. Since
only *positive* facts about the collector are observable, this changes what is
asserted rather than how hard the test tries.

## How this works

The borrow becomes a fact the language guarantees: `codeIterator` holds the
proof, and the deferred `Free` in `codeHashSeq` captures the iterator, so the
owner stays reachable for the whole iteration by ordinary liveness rules.

The durable part is that the owner is a required `newCodeIterator` argument
rather than a field the caller assigns — a future refactor that forgets it fails
to compile instead of silently reintroducing the use-after-free. The assertion
only pins that; it no longer depends on GC timing.

`FindNextKey` and `MarshalBinary` keep their bare `runtime.KeepAlive`: no
wrapper object outlives those calls, so there is no structural analogue, and
their single fast cgo call leaves no window a deterministic test could target.

## Notes

**A coverage trade, stated explicitly:** the deleted tests were the only
end-to-end exercise of `CodeHashes` under GC pressure, and that goes away. I
judged it worth making, since they never established the property they claimed.
If behavioural coverage is wanted back, an ordering assertion — "collection
happened *after* iteration finished", two positive events bounded by the test
deadline — layers on without touching the production change.

`TestCodeIteratorRetainsProof` skips in merkledb mode, where
`fwd_*_code_hash_iter` errors and no iterator exists to inspect; that error path
stays covered by the `else` branch of `Test*CodeHashes`.

Mutation-checked: patching `newCodeIterator` to discard the owner fails both
subtests with `*ffi.RangeProof.codeIterator() must retain the proof it borrows`.

## Breaking Changes

n/a

### DIFF
--- a/ffi/proofs.go
+++ b/ffi/proofs.go
@@ -711,22 +711,12 @@ func getNextKeyRangeFromNextKeyRangeResult(result C.NextKeyRangeResult) (*NextKe
 // instead of silently reintroducing the use-after-free described on
 // [codeIterator].
 func newCodeIterator(result C.CodeIteratorResult, owner codeHashSource) (*codeIterator, error) {
-	it, err := getCodeHashIteratorFromCodeHashIteratorResult(result)
-	if err != nil {
-		return nil, err
-	}
-
-	it.owner = owner
-	return it, nil
-}
-
-func getCodeHashIteratorFromCodeHashIteratorResult(result C.CodeIteratorResult) (*codeIterator, error) {
 	switch result.tag {
 	case C.CodeIteratorResult_NullHandlePointer:
 		return nil, errDBClosed
 	case C.CodeIteratorResult_Ok:
 		ptr := *(**C.CodeIteratorHandle)(unsafe.Pointer(&result.anon0))
-		return &codeIterator{handle: ptr}, nil
+		return &codeIterator{handle: ptr, owner: owner}, nil
 	case C.CodeIteratorResult_Err:
 		err := newOwnedBytes(*(*C.OwnedBytes)(unsafe.Pointer(&result.anon0))).intoError()
 		return nil, err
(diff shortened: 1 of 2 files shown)


ffi/proofs.go
```diff
@@ -701,6 +705,21 @@ func getNextKeyRangeFromNextKeyRangeResult(result C.NextKeyRangeResult) (*NextKe
...
+	it, err := getCodeHashIteratorFromCodeHashIteratorResult(result)
+	if err != nil {
+		return nil, err
+	}
+
+	it.owner = owner
+	return it, nil
+}
+
 func getCodeHashIteratorFromCodeHashIteratorResult(result C.CodeIteratorResult) (*codeIterator, error) {
```
rkuris: looks like this method now has only one caller, and leaves you with a codeIterator and a nil owner. This function is short enough that we should just fold it into newCodeIterator.

rkuris (approved):
I realize that  getCodeHashIteratorFromCodeHashIteratorResult is only package-level visibility but it still feels like a footgun. However, this code is better than what was there before. I think the comments overstate where it is though.

PR #2213, opened 2026-08-25
build(just): Velocity: add a prepush-lite recipe for local iteration

rkuris:
> This is a developer-velocity change.

## Why this should be merged

Developer velocity suffers because Claude has been instructed to run
`just prepush` before every handoff, which is a 5+ minute wait to confirm your
changes won't cause any problems. Most of the time, we just need to run a basic
set of checks. The prepush might be helpful if you're in rarer areas of the
code, but I'd estimate >90% of all changes don't need that level of scrutiny.
If you miss anyway, it's going to CI which will run all those for you.

This mostly just boils down to developer preference: pay the extra local time
to be sure CI can't fail, or go faster and accept the occasional CI failure. I
suspect many people aren't running this at all and have their own stuff because
this is "just" too slow (no pun intended).

Why is `just prepush` so expensive? 23 sub-invocations: five clippy profiles,
five nextest profiles, six Go/cgo FFI steps, plus the docs, markdown, machete,
go-generate, FFI-lint, format and TODO checks.
It builds `maxperf-ethhash-logger` (panic=abort, single codegen unit, fat LTO)
twice and pulls the entire AWS SDK in through `debug-all-features` for
`fwdctl`'s `launch` feature.

## How this works

New `prepush-lite` recipe: `ci-format`, `ci-check-todos`, clippy and tests for
the single `debug-ethhash-logger` profile, `ci-lint-markdown`, `ci-docs`. Like
the other aggregates it shells out through `./scripts/run-just.sh`, so it stays
in sync with the CI commands.

Deliberately not included: the other four Rust profiles, the FFI steps, and
`ci-machete`. Those catch real breakage, but only for changes that touch the
FFI, workspace dependencies, or feature combinations `debug-ethhash-logger`
does not cover. The recipe comment and the reworked "PR Strategy" section in
`AGENTS.md` both say to reach for the full `prepush` in those cases.

`AGENTS.md` previously presented `prepush` as the only local gate. It now
presents `prepush-lite` as the default, `prepush` as the deliberate heavier
option, and CI as authoritative. Also notes
that `ci-format` only *checks* formatting, so `cargo fmt` remains a separate
step.

## Timings

Measured on my laptop: MacBook Pro (Mac17,6), Apple M5 Max, 18 cores
(12 performance + 6 efficiency), 64 GB, macOS 26.5.1, rustc 1.98.0,
cargo-nextest 0.9.132, Go 1.26.5, just 1.58.0, ~4 GB warm `target/`.

This is my working laptop, not a benchmark rig: no core pinning, no attempt to
control thermal or scheduler behavior, normal interactive load and background
daemons throughout, and the runs below are single samples rather than averaged
repetitions. Numbers are rounded and should be read as ballpark ratios, not
measurements — expect noise on the order of tens of percent on the longer runs.

After a source edit (`touch storage/src/lib.rs`, rebuilding `firewood-storage`
and everything downstream):

| recipe          | wall clock  | share of `prepush` |
| --------------- | ----------- | ------------------ |
| `prepush`       | ~5–6 min    | 100%               |
| `prepush-lite`  | ~45 sec     | ~13%               |

With nothing changed and every cache warm (floor):

| recipe          | wall clock  | share of `prepush` |
| --------------- | ----------- | ------------------ |
| `prepush`       | ~5 min      | 100%               |
| `prepush-lite`  | ~25 sec     | ~8%                |

For reference, the first `prepush` of a session against a partly warm `target/`
took ~11.5 min. So `prepush-lite` costs roughly a tenth of `prepush`, and the
repeated-run figure was stable across three consecutive runs (42/43/43 sec).

Worth knowing: `ci-docs` dominates `prepush-lite` and is the most
cache-sensitive step in it — it ranged from ~5 sec to ~108 sec depending on
what had run before. The Rust clippy and nextest steps for one profile are
consistently under 30 sec combined.

## How this was tested

`just --list` and `just -n prepush-lite` to check parsing and wiring, then
`just prepush-lite` and `just prepush` repeatedly against these changes — all
runs green, which is where the numbers above come from.

## Breaking Changes

None. `prepush`, `lint`, and `test` are unchanged.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -140,24 +140,44 @@ of the change.
 
 ## PR Strategy
 
-Before submitting or updating a PR, run the local pre-push checks:
+GitHub Actions is authoritative for the full check matrix. Locally, the default
+gate before submitting or updating a PR is the light one:
+
+```bash
+just prepush-lite
+```
+
+That runs `ci-format`, `ci-check-todos`, clippy and tests for the single
+`debug-ethhash-logger` profile, `ci-lint-markdown`, and `ci-docs`. Note that
+`ci-format` only *checks* formatting — run `cargo fmt` yourself to apply it.
+
+The full local matrix is available when a change warrants it — in particular
+when it touches the FFI, workspace dependencies, or code gated behind a feature
+that `debug-ethhash-logger` does not enable:
 
 ```bash
 just prepush
 ```
 
-This runs `just lint` followed by `just test`. These recipes use the same
-lower-level `ci-*` recipes as GitHub Actions, keeping the local and CI commands
-in sync. Run either phase separately while iterating:
+That is `just lint` followed by `just test`: 23 sub-invocations, which build
+`maxperf-ethhash-logger` (fat LTO) twice, drag in the AWS SDK via
+`debug-all-features`, and run six Go/cgo FFI steps. Expect roughly 15 minutes
+with a warm cache and considerably more cold. Run either phase separately while
+iterating:
 
 ```bash
 just lint
 just test
 ```
 
-If `just` is not installed, use `./scripts/run-just.sh prepush`. The wrapper
-uses `just` when available, falls back to Nix when available, and otherwise
-prints installation instructions.
+All of these recipes use the same lower-level `ci-*` recipes as GitHub Actions,
+keeping the local and CI commands in sync. Individual profiles are reachable
+directly — `just ci-rust clippy debug-all-features`, for example — which is
+usually cheaper than a whole phase when you already know what needs checking.
+
+If `just` is not installed, use `./scripts/run-just.sh prepush-lite`. The
+wrapper uses `just` when available, falls back to Nix when available, and
+otherwise prints installation instructions.
 
 The complete set of Rust profile names and Cargo arguments is defined in
 `scripts/run-rust-ci.sh`. Every profile there is portable to macOS, so the Just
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -140,24 +140,44 @@ of the change.
...
 ```
 
-This runs `just lint` followed by `just test`. These recipes use the same
-lower-level `ci-*` recipes as GitHub Actions, keeping the local and CI commands
-in sync. Run either phase separately while iterating:
+That is `just lint` followed by `just test`: 23 sub-invocations, which build
+`maxperf-ethhash-logger` (fat LTO) twice, drag in the AWS SDK via
+`debug-all-features`, and run six Go/cgo FFI steps. Expect roughly 15 minutes
+with a warm cache and considerably more cold. Run either phase separately while
+iterating:
```
RodrigoVillar: I think this is justification for why we shouldn't run `just prepush` on every change. If so, I would remove this since this leaks implementation details and is already justified in the paragraph above.

rkuris: Fixed in 1fd0c3dc — that sentence is back to just "That is `just lint` followed by `just test`."

One deliberate difference from a literal revert: the old text also said "these recipes use the same lower-level `ci-*` recipes as GitHub Actions, keeping the local and CI commands in sync," and this PR already says that once a few paragraphs down, so I left it there rather than duplicating it.

RodrigoVillar (approved):

PR #2215, opened 2026-08-26
feat(metrics): tag metrics with per-database db_tag label

### LINKED ISSUE #1090: feat: have some way to fetch metrics for only one database

AminR443:
## Why this should be merged

A process using multiple Firewood databases (e.g. multiple chains in one node) currently gets all their metrics mixed into the same series, with no way to distinguish between databases. This PR adds a per-database tag so every metric series identifies its database, letting consumers split dashboards and alerts per database.

## How this works

`MetricsContext`, which was added to support expensive metrics and works through TLS, is extended with a `db_tag` field. The `firewood_counter!`/`firewood_gauge!`/`firewood_histogram!` macros inject it as the first label (`db_tag=<tag>`) on every recording, falling back to `db_tag="untagged"` when no tag is set so the label dimension stays consistent across all    series of a family. For FFI side, `DatabaseHandleArgs` gains a `db_tag` field and a new `WithMetricsTag(tag)` option plumbs the tag into `fwd_open_db`. `Gatherer` is unchanged and returns all series; consumers can filter on the `db_tag` label themselves if they need per-database views.

## How this was tested
UT. New Go tests in `ffi/metrics_test.go`: one for tagged filtering and one for concurrency.

  ## Breaking Changes

  - [ ] firewood
  - [ ] firewood-storage
  - [ ] firewood-ffi (C api)
  - [ ] firewood-go (Go api)
  - [ ] fwdctl

Note for dashboard: every Rust-side metric series now carries a `db_tag` label (`"untagged"` by default).

Closes #1090.

### DIFF
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -92,11 +92,12 @@ pub struct DatabaseHandleArgs<'a> {
     /// Expensive metrics are disabled by default.
     pub expensive_metrics: bool,
 
-    /// Optional tag used to separate metrics and logs per database.
-    ///
-    /// If empty, no tag is applied.
+    /// Tag used to separate metrics and logs per database.
     ///
     /// This must be a valid UTF-8 string.
+    ///
+    /// If empty, no tag is applied and this database's metrics are recorded
+    /// with the default `db_tag="untagged"` label.
     pub db_tag: BorrowedBytes<'a>,
 
     /// The hashing mode to use for the database.
@@ -376,10 +377,6 @@ fn parse_db_tag(db_tag: BorrowedBytes<'_>) -> Result<Option<metrics::SharedStrin
         .as_str()
         .map_err(|err| invalid_data(format!("database tag contains invalid utf-8: {err}")))?;
 
-    if db_tag.is_empty() {
-        Ok(None)
-    } else {
-        let db_tag = Arc::<str>::from(db_tag);
-        Ok(Some(metrics::SharedString::from(db_tag)))
-    }
+    // Arc<str> keeps the per-recording clone of the tag a refcount bump.
+    Ok((!db_tag.is_empty()).then(|| metrics::SharedString::from(Arc::<str>::from(db_tag))))
 }
--- a/ffi/src/handle.rs
+++ b/ffi/src/handle.rs
@@ -2,6 +2,7 @@
 // See the file LICENSE.md for licensing terms.
 
 use std::num::{NonZeroU64, NonZeroUsize};
+use std::sync::Arc;
 
 use firewood::{
     DefaultHashMode,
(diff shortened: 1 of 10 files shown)


metrics/src/lib.rs
```diff
@@ -175,9 +176,10 @@ pub fn strip_doc_leading_space(s: &str) -> &str {
 ///
 /// This is set at API boundaries (e.g., FFI entrypoints) and read when deciding
 /// whether to record expensive metrics.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct MetricsContext {
     expensive_metrics_enabled: bool,
+    db_tag: Option<SharedString>,
```
demosdemon, line 183: Add a comment here that `SharedString` is cheap-ish to clone because it uses atomic reference counting over a shared pointer. This is important because it makes all of the `.clone()`s also cheap-ish.

metrics/src/lib.rs
```diff
@@ -208,29 +226,53 @@ pub struct MetricsContextGuard {
...
+            .borrow()
+            .as_ref()
+            .and_then(|context| context.db_tag.clone())
+    })
+}
+
+/// Returns the current metrics db tag, or a default "untagged" label.
+#[must_use]
+pub fn current_db_tag_label() -> SharedString {
+    current_db_tag().unwrap_or_else(|| SharedString::const_str(DEFAULT_DB_TAG))
```
demosdemon: nit: `const_str` is a const fn therefore the result is `const` so the `unwrap_or` is generally more appropriate.

```suggestion
    current_db_tag().unwrap_or(const { SharedString::const_str(DEFAULT_DB_TAG) })
```

You can replace the `const { ... }` by making `DEFAULT_DB_TAG` a `SharedString` above.

ffi/src/handle.rs
```diff
@@ -354,10 +364,19 @@ impl<'db> CView<'db> for &'db crate::DatabaseHandle {
...
     api::Error::IO(std::io::Error::new(std::io::ErrorKind::InvalidData, error))
 }
+
+fn parse_db_tag(db_tag: BorrowedBytes<'_>) -> Result<Option<metrics::SharedString>, api::Error> {
+    let db_tag = db_tag
+        .as_str()
+        .map_err(|err| invalid_data(format!("database tag contains invalid utf-8: {err}")))?;
+
+    // Arc<str> keeps the per-recording clone of the tag a refcount bump.
+    Ok((!db_tag.is_empty()).then(|| metrics::SharedString::from(Arc::<str>::from(db_tag))))
```
demosdemon, line 384: Can you run an experiment where you use `metrics::SharedString::const_str(String::leak(db_tag))` instead of the `Arc`. I'm curious the performance overhead of the Arc versus simply leaking the string once up-front.

demosdemon (approved):

ffi/src/handle.rs
AminR443: Measured and decided to keep as-is. No particular difference in performance.

PR #2216, opened 2026-08-26
feat(logging): tag log lines with per-database db_tag

AminR443:
## Why this should be merged
When multiple Firewood databases run in the same process, their log lines are indistinguishable. Building on the per-database `MetricsContext` (from the #2215 this stacks on), this attaches the same `db_tag` to every log record emitted by the storage layer, so logs can be attributed to a specific database instance.

## How this works
`firewood-metrics` gains an optional logger feature and a `firewood_log!` macro family (`firewood_error!`, ..., `firewood_trace!`). Each macro reads the current `db_tag` from the thread-local `MetricsContext` and emits it as a log key-value (`kv` feature) on the record.

## How this was tested
UT.

## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl

### DIFF
--- a/ffi/src/logging.rs
+++ b/ffi/src/logging.rs
@@ -112,12 +112,14 @@ impl LogArgs<'_> {
         env_logger::Builder::new()
             .filter_level(level)
             .format(move |buffer, record| {
-                let db_tag = record_db_tag(record);
+                // matches env_logger's default format style
+                // TODO(AminR443): adopt slog format like avalanchego
+                // https://github.com/ava-labs/firewood/issues/2227
                 let timestamp = buffer.timestamp_millis();
-                if let Some(db_tag) = db_tag {
+                if let Some(db_tag) = record_db_tag(record) {
                     writeln!(
                         buffer,
-                        "{timestamp} [{}] [{}] [db_tag={db_tag}] {}",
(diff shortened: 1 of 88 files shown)


ffi/src/logging.rs
```diff
@@ -100,6 +111,27 @@ impl LogArgs<'_> {
...
+                } else {
+                    writeln!(
+                        buffer,
+                        "{timestamp} [{}] [{}] {}",
+                        record.level(),
+                        record.target(),
+                        record.args()
+                    )
+                }
+            })
```
demosdemon, line 136: The original format was `[{timestamp} {level:<5} {target}] {args}` with only one open and one close bracket. However, since we're adjusting the format, I wonder if we should adapt the slog format that's already used by [avalanchego](https://github.com/ava-labs/avalanchego/blob/9705d9705a9fb055346b6231489f8e58c5126aa4/graft/evm/log/format.go) for a uniform style. That's a bigger ask then I originally intended so that can be deferred to a follow up.

demosdemon (approved):

ffi/src/logging.rs
AminR443: #2227

Issue #2212, opened 2026-08-25
Deep peer-supplied keys can abort the process during proof verification


bernard-avalabs:
### Affected area(s)

hashing, proofs

### Describe the bug

Verifying a range or change proof from an untrusted peer rebuilds a trie from the peer's data and hashes it. The walks that do this recurse once per trie level. Trie depth is bounded by the peer's key lengths, so the peer decides the recursion depth. A proof of a few kilobytes exhausts the stack.

A Rust stack overflow hits the guard page and aborts the process. It is not a catchable panic, so there is no error to return and nothing upstream can recover. For a syncing node this is a crash reachable with no authentication and no user interaction.

This is an availability issue. Nothing is disclosed or modified.

## Why it aborts rather than returning an error

Stack exhaustion is not unwinding, so `catch_unwind` never sees it. The FFI boundary installs a guard at `ffi/src/lib.rs:72`, and it cannot help here. The `maxperf` profile sets `panic = "abort"` and compiles that guard out regardless.

2 MiB is the operative stack size. It is `std::thread`'s default on all Tier-1 platforms, and firewood sets `stack_size` nowhere. That includes the rayon pool used for parallel proposal application, built at `firewood/src/manager.rs:622` with `stack_size` left unset.

## Where the depth comes from

Depth comes from shared prefixes between keys, not from key count or key length alone. Keys of the form `[0x00; i] ++ 0x11` for increasing `i` each share one more byte than the last, so they build a chain rather than a bush. Fifty-nine such pairs, none longer than 59 bytes, are enough.

Measured on a 2 MiB stack:

| walk | survives | aborts | stack per level |
| --- | --- | --- | --- |
| insertion | 58 levels | 59 | about 35 KB |
| hashing | 140 levels | 160 | about 14 KB |

## Reproduction, range proofs

Build a proof whose sorted keys form the chain above, then verify it. Build the fixture on a deliberately large stack and verify on a small one, so any abort is attributable to verification rather than to setup.

| verify stack | deepest that verifies | first that aborts |
| --- | --- | --- |
| 2 MiB | 58 keys | 59 keys |
| 8 MiB | 230 keys | 240 keys |

`verify_range_proof` takes no count parameter and performs no internal count check, so nothing bounds how many pairs arrive.

## Reproduction, change proofs

Take an honest change proof, keep its boundary proofs, and replace its batch operations with a deep chain followed by the honest final operation.

| forged operations | outcome |
| --- | --- |
[...]

PR #2222, opened 2026-08-28
refactor(storage): make the node hashing walk iterative

bernard-avalabs:
Stack: PR 2 of 7 for #2212, on top of the key-length bound PR.

## Why this should be merged

The live hashing walk recursed once per trie level and aborted at 2,048 nibbles on a 2 MiB stack, exactly the depth the key-length bound admits, so the deepest legal input was fatal.

## How this works

The walk keeps one path buffer and a stack of frames, split across two functions. `hash_helper` owns the stack and does what the call stack did: it remembers which slot is waiting on a child and hashes a node once its children are done. `hash_step` does one node's worth of the old recursive body, preparing the node on its first visit and taking out the next child that needs hashing, so it reads against the recursive function nearly line for line. Each frame records the path length at its own prefix, which replaces the nested `PathGuard` that the recursive version used inside this module (the public `PathGuard` in `path/buf.rs` is unrelated and unchanged). The fold state that recursion passed as arguments lives in the frames. The frame stack has inline capacity for sixteen frames (`FRAME_STACK_INLINE_CAPACITY`), enough for a realistic trie including an `ethhash` storage write, so the common case allocates nothing. Leaf children are hashed in place rather than given frames. The walk's doc gains a `# Panics` section covering the three internal-invariant panics in the driver. Two comments elsewhere, in the merkle crate and in the `ethhash` hasher, named the recursive function and now name the function that does that job.

## Review guide

The recursive `hash_helper_inner` on the left side of the diff (lines 130 to 240 of the old file) is split into four functions and a struct. Each row is one piece of the old body and where it now lives.

| Old code | New home |
|---|---|
| The arguments `node`, `path_prefix`, `fake_root_extra_nibble` and the local `make_fake_root`. | Fields of `HashFrame` with the same names, plus `prefix_len` for the path. The three extra fields, `cursor`, `pending` and `prepared`, record where a call was, which the call stack used to remember. |
| The account-branch block (lines 142 to 180): the depth test, `ethhash_classify_children`, the rehash of a single already-hashed child, and the `make_fake_root` rule. | `prepare_account_branch`. Same code, reading the frame instead of arguments and locals. `hash_step` runs it on a frame's first step, before any child is taken out, which is when the old code ran it. |
| The `for` loop over children up to the recursive call (lines 190 to 212): the skip predicate, `mem::take`, extend by `partial_path`, push the nibble unless folding. | The `while` loop in `hash_step`. Same predicate, same take, same path arithmetic. Where the old code recursed, it returns `Descend { nibble, child }`. |
| The recursive call and the write-back after it (lines 212 to 216). | The driver, `hash_helper`. On `Descend` it takes the fold argument from the frame's `make_fake_root`, as the old loop took it from its local, then pushes a frame for a branch child and records the slot in `pending`. When that frame finishes, the next turn installs `Child::MaybePersisted` into that slot. A leaf child gets no frame and is hashed in place. |
| The tail (lines 220 to 239): `update_account_storage_root`, the fold-or-plain hash, `SharedNode::new`. | `hash_finished_node`, unchanged. The driver calls it on `Finish` for a branch and directly for a leaf child. |
| `PathGuard` extend and restore. | `path.0.truncate(frame.prefix_len)` in the driver, at the points where a guard used to drop: after a child's result comes back, and before the node is hashed. |

Suggested order: compare `hash_step` with the old `for` loop first, since they match nearly line for line. Then read `hash_helper`. It does by hand what the call stack did for the recursive version: pushing a frame is the recursive call, popping one is the return, and `pending` records where the returned hash goes.

## How this was tested

The full storage and firewood suites pass under both feature profiles. Hash outputs are pinned by the merkledb fixture tests in `merkle/tests/triehash.rs` and the geth-reference and account-fold tests in `merkle/tests/ethhash.rs`; all pass. A new test hashes a 4,096-level chain on a 256 KiB thread, which the recursive walk could not do, so a return to recursion aborts it. The proof round-trip guards land later in the stack.

Out of scope here and fixed later in the stack: `Reconstructed::root_hash` still clones the root with the derived `Clone` before calling this walk, and the returned node chain has depth-proportional drop glue.

## Breaking Changes

None.


rkuris:
I'm still thinking about this, but I'm not convinced this is the best way to fix it. I think it introduces a tremendous amount of complexity.

Preliminary analysis shows the frame size for a recursive call at a whopping 2816 bytes. The arguments account for only about 170 bytes, so the locals are where we're running into problems. Here's an AI analysis that appears to make some sense, and seems like a better direction overall:

```
Options
A — Shrink the frame, keep recursion. In rough order of payoff per unit of risk:

Replace the for (nibble, child) in &mut b.children loop with for nibble in PathComponent::ALL + b.children.get_mut(nibble). PathComponent::ALL already exists and is documented as the preferred idiom ([component.rs:47](vscode-webview://1c5jin534i7n04o6up4n5ntrvotdmenc7q6uhckmme90kbi237jb/storage/src/path/component.rs#L47)). Drops the live iterator from 192 B to ~24 B.
Extract the account-depth block ([hash.rs:139-180](vscode-webview://1c5jin534i7n04o6up4n5ntrvotdmenc7q6uhckmme90kbi237jb/storage/src/nodestore/hash.rs#L139-L180)) into an #[inline(never)] helper returning Option<PathComponent>. ClassifiedChildren, the SharedNode, and the fold's argument setup are only reachable at exactly depth 64 — once per root-to-leaf path — but today they occupy every frame.
Change the signature to fn hash_helper_inner(&self, node: &mut Node, ...) -> Result<HashType, FileIoError>, letting the caller own the SharedNode::new wrap and the child-slot store. Removes the 96 B outgoing Node slot, the mem::take temporary, and shrinks sret from 64 B to size_of::<Result<HashType, FileIoError>>().
Optional: act on the existing HashType TODO ([hashtype.rs:30](vscode-webview://1c5jin534i7n04o6up4n5ntrvotdmenc7q6uhckmme90kbi237jb/storage/src/hashtype.rs#L30)) — SmallVec<[u8;32]> with a usize length makes HashType 48 B. A u8 length would take it to ~40 and shrink every return slot and every Children<Option<HashType>> (768 → 656) along with it.
Items 1–3 are local and I'd expect them to land the release frame near 450–500 B and the debug frame well under 1400 — the "about half" you asked for. They do not give you a bound.

B — Explicit heap stack. Iterative post-order with Vec<Frame>, Frame = { node: Node, cursor: u8, make_fake_root: Option<PathComponent>, prefix_len: usize } ≈ 112 B/level on the heap, constant machine stack. The awkward part is that PathGuard's RAII truncation has to become explicit path.0.truncate(frame.prefix_len) on pop, and the ethhash fold at depth 64 has to be re-expressed against the explicit stack. Real rewrite of a subtle function, and it fixes only this function.

C — stacker::maybe_grow at the recursion site. One line, unbounded depth, no rewrite, and it works identically for insert_helper, remove_helper, and Node's drop glue — which is the whole class of bug rather than one instance. Costs a dependency (stacker + psm, both use internal unsafe, so the workspace unsafe_code deny is unaffected) and a branch per level.

Recommendation
Do A1–A3 now, then C, and treat B as unnecessary.

A1–A3 are cheap, individually reviewable, and give you the measurable halving you asked for. But the table says a frame-size fight is unwinnable as a correctness strategy: you'd have to fix insert_helper (3840 B), remove_helper, and drop glue too, and any future local addition silently eats your margin back. C gives you the actual bound for a fraction of B's effort and covers all four call sites.

Two things I'd also fix regardless of which option you pick, since they're one-liners that buy back more headroom than A1–A3 combined:

Set .stack_size(...) on the rayon pool in [manager.rs:629](vscode-webview://1c5jin534i7n04o6up4n5ntrvotdmenc7q6uhckmme90kbi237jb/firewood/src/manager.rs#L629). Parallel hashing is running on a 2 MiB stack today, and that's the configuration your deep-trie tests are most likely to hit.
Decide whether unbounded key length is actually supported. There's no MAX_KEY_LEN anywhere in the tree, and 2048 nibbles is a 1024-byte key against a store whose target workload is 32-byte keys. A documented cap validated at the FFI and fwdctl boundaries may be the honest answer, in which case C stops being load-bearing and A becomes pure headroom.
I haven't touched any source. Tell me which direction and whether your repro really enters via insert_helper or via reconstruction, and I'll write the change.
```

bernard-avalabs:
> C — stacker::maybe_grow at the recursion site. One line, unbounded depth, no rewrite, and it works identically for insert_helper, remove_helper, and Node's drop glue — which is the whole class of bug rather than one instance. Costs a dependency (stacker + psm, both use internal unsafe, so the workspace unsafe_code deny is unaffected) and a branch per level.

I wasn't aware of the stacker crate. stacker::maybe_grows seems like an interesting solution (possibly together with some frame size trimming). I can look into it more.

PR #2224, opened 2026-08-28
refactor(storage): make node cloning iterative

bernard-avalabs:
Stack: PR 3 of 7 for #2212, on top of the hashing-walk PR.

## Why this should be merged

A derived clone of a `Node` descends once per trie level because a node owns its children by value, and a mutable trie serves every get by cloning: resolving the root of a mutable nodestore clones the whole trie, and producing the found node clones everything below it. With the derived clone in place, the survival test in this PR aborts with a stack overflow at 4,096 levels on a 2 MiB thread.

## How this works

Cloning a branch becomes a loop over frames, split across two functions. `clone_step` does one branch's worth of work: it moves a cursor over the child slots, copies each child that owns no subtree into the new branch as it is reached, and returns the first branch child it finds so the driver can give it a frame. `clone_branch` owns the frame stack. It pushes a frame for that child, and when a frame finishes it installs the copy in the parent's slot. This is the shape the hashing walk uses. The frame stack has inline capacity for realistic depths, so a shallow clone allocates nothing beyond the copy itself. The capacity constant moves from the hashing walk's module to the node module, since two walks now share it.

## Review guide

The derive cloned `Node` field by field and recursed through `Child::Node`. Each piece of that has one home here.

| Derived behaviour | New home |
|---|---|
| `Node::Leaf`: `LeafNode::clone`. | `Node::clone`, the leaf arm, unchanged. |
| `Node::Branch(Box<BranchNode>)`: a new box. | `CloneFrame::new`, one box per frame. `clone_branch` returns the root's. |
| `BranchNode::partial_path` and `value`. | `CloneFrame::new`. |
| `children`, slot by slot. | The `match` in `clone_step`, in slot order. Two arms copy in place. `None` slots stay `None`, which `Children::new()` in `CloneFrame::new` already made them. The branch arm returns `Descend`, and the driver installs that child's copy when its frame finishes. |
| The recursion. | The frame stack. One frame and one box per branch reached through `Child::Node`. A frame pops only when every slot is copied. |

Two things about `Node` still recurse once per level: comparing with `==`, which is the derived `PartialEq`, and dropping, which is the compiler-generated code that frees a node and everything it owns. On a 4,096-level chain either would overflow the stack by itself. The survival test therefore never compares its trees and leaks them with `std::mem::forget` rather than letting them drop, so the only recursion it can trip over is in the clone under test.

## How this was tested

Clone unit tests: equality against the derived semantics with whole-subtree `PartialEq` as the oracle, on a single spine and on a fan-out fixture where every branch above the bottom level has four branch children, one of them in the last slot, and every `Child` variant appears; deep-copy independence under mutation; and a 4,096-level chain survival test on an explicit 2 MiB thread. With the derive restored, the survival test aborts and the equality tests still pass, so they are a valid oracle. Full suites under both feature profiles.

## Breaking Changes

None. Clone semantics are identical; only the call-stack shape changes.


bernard-avalabs:

PR #2228, opened 2026-08-31
refactor(merkle): make the insert and read walks iterative

bernard-avalabs:
Stack: PR 4 of 7 for #2212, on top of the clone PR.

## Why this should be merged

`insert_helper`, `insert_branch_helper` and `get_helper` each recursed once per trie level. Insertion aborted at roughly 200 deep keys on a 2 MiB stack, and a get of the deepest key the bound admits aborted even with cloning already iterative. Inserts are reached from proof verification with peer keys; gets are reached from the public API.

## How this works

The insert walks become loops over `Box<BranchNode>` parent frames with the per-level work in `insert_step` / `insert_branch_step` and reattachment shared through `unwind_parents`. The read walk becomes two nested loops that follow in-memory children by reference and only take ownership when a persisted child is resolved. The drivers borrow the caller's key for the first level, so a shallow insert allocates nothing.

## How this was tested

Full suites under both feature profiles. After the conversion, insertion and reads survive 8,192 nibbles on a 2 MiB stack, four times the bound. The depth guards land later in the stack as each path's remaining walks convert.

## Breaking Changes

None.


bernard-avalabs:

PR #2229, opened 2026-08-31
refactor(merkle): make the removal walks iterative

bernard-avalabs:
Stack: PR 5 of 7 for #2212, on top of the insert-and-read PR.

## Why this should be merged

Change-proof verification applies the peer's batch ops, so a `Delete` or `DeleteRange` reaches the removal walks with a peer-chosen key. `remove` aborted at 1,024 nibbles on a 2 MiB stack, half the depth the bound admits.

## How this works

`remove_helper` and `remove_prefix_helper` become loops over `Box<BranchNode>` parent frames with per-level work in `remove_step` / `remove_prefix_step`, sharing `unwind_and_flatten` for the flatten-on-unwind tail. `delete_children` only tallies what it frees, so it becomes a worklist. Its caller pre-counted a valued branch that `delete_children` counts again, so `remove_prefix` over-reported by one there; the pre-count is dropped (callers discard the count).

## How this was tested

Full suites under both feature profiles. A new guard drives the in-memory paths at the deepest structure the bound admits, a spine that branches at every nibble: gets at both ends, a deepest-key removal, and a prefix removal of the whole spine, on an explicit 2 MiB thread. The guard aborts the process with node cloning derived, and again with the read walk recursive, and passes with both converted.

## Breaking Changes

None. `remove_prefix`'s returned count no longer double-counts a valued branch at the prefix top.


bernard-avalabs:

PR #2230, opened 2026-08-31
refactor(merkle): make the proof-side root-hash walk iterative

bernard-avalabs:
Stack: PR 6 of 7 for #2212, on top of the removal PR.

## Why this should be merged

`compute_root_hash_with_proofs` rebuilds the proving trie's root hash during proof verification. It recursed once per trie level. On a 2 MiB stack it aborted at 320 nibbles, which made it the lowest ceiling on the verification path and put it far inside the key-length bound.

## How this works

The walk is now a post-order traversal over heap frames. Each frame holds a cursor into its children and the slot waiting on the child frame below it. The two helper functions it used to call, `compute_root_hash_as_storage_trie_root` and `build_branch_parts`, are folded into it. Whether a node is hashed normally or as an `ethhash` storage-trie root is recorded on its frame as `HashAs`. Each frame looks up its proof node, out-of-range mask, and lone-storage-child slot once, where the recursive version repeated those lookups for every child.

`NodeStore<Mutable<_>>` gains `root_ref`, which borrows the root. Both verification paths used `root_node`, which copies the whole trie for a mutable nodestore. Neither path needs more than a borrow.

This PR also carries the account-depth rule from #2206, which landed on `main` after the stack was cut. At account depth under `ethhash`, a persisted child's stored hash cannot be trusted: it reflects the child count when it was written, and reconcile and collapse can change that count. The walk therefore takes `storage` and returns a `Result`. When it meets a persisted child at account depth, it reads the child back and hashes it with an inner walk. That inner walk cannot recurse further: a node read from storage has only persisted children, and they sit below account depth, so it uses their stored hashes. The tests #2206 added pass through the new walk.

## How this was tested

The full test suites pass under both feature profiles, including the account-depth tests from #2206.

A new guard builds the deepest trie the bound admits, generates a range proof over it, and verifies the proof on a 2 MiB thread. A stack overflow aborts the process, so if the walk ever recurses again this test dies rather than failing an assertion. The `prefix_chain_keys` test helper that builds that trie is added here, since this is its first user in the stack.

## Breaking Changes

None. `root_ref` is a new public method on `NodeStore<Mutable<_>>`.


bernard-avalabs:

bernard-avalabs:

PR #2231, opened 2026-08-31
refactor(merkle): make the collapse walks iterative

bernard-avalabs:
Stack: PR 7 of 7 for #2212, on top of the proof-hash PR.

## Why this should be merged

The collapse walks run during change-proof verification, driven by the peer's proof-node keys: `collapse_strip` aborted between 512 and 768 nibbles and `child_in_range` at roughly 1,400 levels on a 2 MiB stack, both inside the bound. This PR completes #2212: with it, every walk a peer can drive survives the deepest input the bound admits.

## How this works

`collapse_navigate` becomes a loop sharing `unwind_parents` with the insert walks. `collapse_strip` flattens each branch on the unwind through `flatten_collapsed`. `child_in_range` answers one question about a whole subtree, so its descent becomes a worklist with children resolved at push time, matching the recursive version's error surfacing.

## How this was tested

Full suites under both feature profiles. A guard commits a Delete on the deepest key of a chain at the bound, then generates and verifies the change proof on an explicit 2 MiB thread, reaching the removal and collapse walks; with the removal walks reverted to recursion this guard aborts the process. The change-proof fuzzer soaked ten minutes per hash mode against the completed branch, clean.

## Breaking Changes

None


bernard-avalabs:

Issue #1507, opened 2025-12-08
Look into custom `rustfmt` configuration


RodrigoVillar:
Currently, we use the default options for `rustfmt` but we're able to customize it similar to `clippy`. One option I think would be useful is [reorder imports](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#reorder_imports), as our importing style differs per file.

demosdemon:
I am for using some of the nightly rustfmt features; however, it is annoying to do so while also trying to maintain support for stable releases of rust. The feature you linked to reorder imports is already enabled by default. A more significantly useful feature is the one listed right after, `group_imports`. I'm a big proponent of using `group_imports = StdExternalCrate`. However, because that feature is unstable, it is only supported by the nightly toolchain version of `rustfmt` which requires us to use `cargo +nightly fmt` to format the code. It feels like a minor speedbump on the surface, but it becomes a huge pock after some time.

rkuris:
IMO we should wait until the useful nightly options become stable.

PR #2232, opened 2026-09-01
ci: pin the formatting and linting toolchain in rust-toolchain.toml

demosdemon:
## Why this should be merged

The nightly pin added for clippy lived in three places: the `clippy` command in `scripts/run-rust-ci.sh`, the `rust-lint` job in `ci.yaml`, and the `miri` job beside it. Enabling nightly-only rustfmt options would have added a fourth, and would have required contributors to remember `cargo +nightly-<date> fmt` instead of `cargo fmt` -- a wrapper that decays the moment anyone forgets it.

Moving the pin into `rust-toolchain.toml` makes a bare `cargo fmt` resolve to it with no ceremony, and collapses the copies to one. This commit lands the plumbing only; the `rustfmt.toml` options it exists for come next, so the tree is byte-identical afterwards.

This PR enables #1507, which will soon follow.

## How this works

`rust-toolchain.toml` outranks `rustup default`, which is what `dtolnay/rust-toolchain` sets. Introducing the file therefore turns every `toolchain: stable` input in every workflow into a no-op: without further changes, all of CI would silently build and test on the pinned nightly. The fix is `RUSTUP_TOOLCHAIN`, which outranks the toolchain file in turn, applied at three layers so no invocation path is missed:

  - `scripts/run-rust-ci.sh` sets it for check, build, build-benches, bench, test, and the two example commands. It defaults rather than overrides (`${RUSTUP_TOOLCHAIN:-stable}`) so a caller can still aim the same commands at a third toolchain without editing the script.
  - The `justfile` sets it inline on the recipes that invoke Cargo directly. The two formatting recipes instead *clear* it, so an exported value cannot quietly format against a toolchain CI does not check with.
  - Each workflow declares it in an `env:` block -- workflow-level where every job needs stable, per-job in `ci.yaml`, the only workflow containing jobs that want the pin.

The three jobs the pin exists for drop their `dtolnay/rust-toolchain` step entirely and call `rustup toolchain install`, which with no argument installs the active toolchain: channel and components both come from the file. That is what keeps the date in one place. `miri` passes `--component miri` rather than adding it to the file, since no one else needs it.

`lint-nightly.yaml` keeps its explicit `cargo +nightly`, which outranks both the variable and the file. Its schedule comment also named the pinned date and has been reworded, since a second copy of the date is exactly the drift this change removes.

## How this was tested

  - `just prepush-lite`: 836 tests, clippy clean on the pinned nightly, and `cargo fmt --check` clean, confirming the pinned rustfmt at default configuration reformats nothing.
  - `actionlint` over all workflows, compared per file against HEAD: identical.
  - Resolution order checked directly: a bare `cargo` in the repository reports 1.98.0-nightly, the script's build path reports stable 1.98.0, an externally supplied `RUSTUP_TOOLCHAIN` still wins, `cargo +toolchain` outranks the variable, and a bogus value fails loudly instead of falling back to stable.
  - `git grep` confirms the pinned date now appears in one file.

## Breaking Changes

None for consumers of the published crates, which continue to build on stable.

Contributors using rustup will see the pinned nightly install itself on their first Cargo command in the repository, and `target/` grows because stable and nightly artifacts coexist. They do not invalidate each other, so this costs disk rather than rebuilds.

`nix develop` is unaffected and remains outside this scheme: `ffi/flake.nix` names its own stable toolchain and provides no rustup, so `cargo fmt` there will not match `just fmt` once the nightly-only options land.


.github/workflows/attach-static-libs.yaml
```diff
@@ -16,6 +16,12 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: full
+  # The static libraries built here are consumed by Go projects, and on a tag
+  # they are attached to a release, so they come from the stable channel and
+  # never from the nightly rust-toolchain.toml pins for `cargo fmt` and clippy.
+  # That file outranks rustup's default, so the `toolchain: stable` input below
+  # is not on its own enough to guarantee it; RUSTUP_TOOLCHAIN is.
```
rkuris, line 23: Wordy. Suggest:

```suggestion
  # RUSTUP_TOOLCHAIN must be specified here to override rustup's
  # default of `toolchain::stable`. This ensures we use stable builds when
  # building the Go static libraries.
```

.github/workflows/benchmarks.yaml
```diff
@@ -30,6 +30,10 @@ env:
   # scripts/run-rust-ci.sh, and the Swatinem/rust-cache shared-key, so the
   # warming job and the matrix legs cannot drift apart.
   PROFILE_KEY: maxperf-ethhash-logger-test_utils
+  # rust-toolchain.toml pins a nightly for `cargo fmt` and clippy, and it
+  # outranks rustup's default -- so a `toolchain: stable` input alone would not
+  # keep this workflow off the pin. RUSTUP_TOOLCHAIN does.
```
rkuris, line 35: Duplicative. Maybe refer to the other file for reasoning or create a README in .github/workflows that explains this philosopy or something. This one was much shorter and accomplished the same thing as the other one.

.github/workflows/ci.yaml
```diff
@@ -9,6 +9,20 @@ env:
...
+# That file outranks `rustup default`, so the `toolchain:` input of
+# dtolnay/rust-toolchain no longer decides anything on its own -- without
+# `RUSTUP_TOOLCHAIN`, every job here would silently run the pin.
+#
+# So each job that compiles or runs the workspace sets `RUSTUP_TOOLCHAIN:
+# stable`, which does outrank the file. The three jobs the pin exists for --
+# `no-rust-cache-lint`, `rust-lint`, and `miri` -- set no toolchain at all and
+# let `rust-toolchain.toml` supply both the channel and its components, which is
+# what keeps the pinned date in exactly one file.
+
```
rkuris, line 25: Case #3 repeating the same thing. I'm leaning toward a readme.

Action required: Find some way to reduce duplication. This could be referring the reader to a workflows readme or a section of our existing readme. That readme could contain something like "here's how we pick which version we're doing when" or something.

Another option is to pare these down significantly.

Another problem is inconsistent placement. The others had it right before the setting of the environment variable. This one puts it above the jobs tag. Probably not the right place.

Action required: Move the comment to the line (preferred) or move all the other comments.

.github/workflows/ci.yaml
```diff
@@ -97,10 +113,11 @@ jobs:
       - uses: taiki-e/install-action@67729d5c413db75907f0ad1e39bb04b9c868ff60 # v2.85.7
         with:
           tool: just@1.46.0
-      - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # @master
-        with:
-          toolchain: stable
-          components: rustfmt
+      # No toolchain is named here. `rust-toolchain.toml` supplies the channel
+      # and the rustfmt component, and `rustup toolchain install` with no
+      # argument installs whatever that file names.
```
rkuris, line 118: This describes the fix, and is quite verbose. I am not certain it's needed, but if you think it is, then here's a suggestion:

```suggestion
      # Toolchain version comes from `rust-toolchain.toml`
```

Action required: remove this comment or accept the suggestion

.github/workflows/ci.yaml
```diff
@@ -436,10 +470,11 @@ jobs:
     runs-on: macos-26
     steps:
       - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # @master
-        with:
-          toolchain: nightly-2026-07-05
-          components: miri
+      # No toolchain is named here; `rust-toolchain.toml` supplies the channel.
+      # miri is requested per-job rather than added to that file's component
+      # list, because this is the only job that needs it.
```
rkuris, line 475: Same as previous comment. Further blocks like this are not receiving additional comments.

Action required: Be consistent with the action of the other comments and either remove or shorten this comment. Search for additional instances of this comment type and address them in the same way.

.github/workflows/lint-nightly.yaml
```diff
@@ -24,6 +25,9 @@ jobs:
       - uses: taiki-e/install-action@67729d5c413db75907f0ad1e39bb04b9c868ff60 # v2.85.7
         with:
           tool: just@1.46.0
+      # The floating nightly, not the one rust-toolchain.toml pins. The
+      # `clippy-nightly` command names it as `cargo +nightly`, which outranks
+      # that file.
```
rkuris, line 30: Question: Does this mean each time nightly gets updated we have to change this sha?

Action: Answer the question, then resubmit for review, or ping me on slack referring to this comment.

scripts/run-rust-ci.sh
```diff
@@ -96,6 +100,24 @@ case "$profile" in
         ;;
 esac
 
+# Commands that compile or run the workspace use stable, not the nightly pinned
+# by rust-toolchain.toml. RUSTUP_TOOLCHAIN outranks that file, so setting it here
+# is enough to opt out.
```
rkuris, line 105: This is a well-disguised repeat of some prior comments. Remainder of this block comment was not reviewed.

Action required: consider treating this as the other comments; at best this is a referral to a README. Another option is a shorter comment.

scripts/run-rust-ci.sh
```diff
@@ -96,6 +100,24 @@ case "$profile" in
...
+#
+# This defaults rather than overrides, so a caller that exports
+# RUSTUP_TOOLCHAIN itself still gets the toolchain it named -- which is what
+# lets these same commands be pointed at a third toolchain, such as the
+# workspace's declared rust-version, without editing this script.
+#
+# `clippy` is absent on purpose: it is one of the commands the pin exists for.
+# `clippy-nightly` names its toolchain inline, which outranks both this variable
+# and the toolchain file.
```
rkuris, line 114: nit: hard to parse

Action required: None, but I had a hard time reading this so any cleanup would be helpful.

scripts/run-rust-ci.sh
```diff
@@ -131,7 +153,7 @@ case "$command" in
         cargo bench --frozen ${cargo_args[@]+"${cargo_args[@]}"} --workspace --bench "$1" -- --noplot
         ;;
     clippy)
-        cargo +nightly-2026-07-05 clippy --locked ${cargo_args[@]+"${cargo_args[@]}"} --workspace --all-targets -- -D warnings
+        cargo clippy --locked ${cargo_args[@]+"${cargo_args[@]}"} --workspace --all-targets -- -D warnings
```
rkuris, line 156: nit: this one could have used a one-line comment

Action required: none

AGENTS.md
```diff
@@ -184,6 +184,44 @@ changing a CI Rust matrix, update the shared script and both callers as needed.
 
 All tests must pass, and there should be no clippy warnings.
 
+### Toolchain Selection
```
rkuris, line 187: Ah, here is the README entry you can refer to for the toolchain selection. Not 100% sure AGENTS.md is the best place, but it's OK.

Action required: This is only a comment and no action is actually required here. This does not block approval. The comment is provided because I believe it's relevant to the discussion in prior blocks.

AGENTS.md
```diff
@@ -184,6 +184,44 @@ changing a CI Rust matrix, update the shared script and both callers as needed.
...
+  outranks. A job that must run stable says so with `RUSTUP_TOOLCHAIN`.
+- Building on both channels roughly doubles the size of `target/`, because
+  Cargo hashes the compiler version into each unit. It does not cause rebuilds:
+  alternating between `just test` and a rust-analyzer check leaves both sets of
+  artifacts intact.
+
+The Nix flake is outside all of this. `ffi/flake.nix` names
+`rust-bin.stable.latest` and provides no rustup, so it neither reads nor needs
+`rust-toolchain.toml`. Running `cargo fmt` inside `nix develop` therefore uses
+that flake's stable rustfmt and produces different output than `just fmt`.
```
rkuris, line 223: This is scary. So if someone uses nix they get the wrong rustfmt?

Can we fix this?

Action required, answer the question. If it's "no this is hard to fix" and you don't think we should fix it, then we probably want this note somewhere more prominent. If it's "yes" then create a new issue explaining the problem, or fix it in this PR.

CONTRIBUTING.md
```diff
@@ -23,15 +23,26 @@ guidelines for contributing to firewood.
...
 Resolve any warnings or errors before making your PR.
 
+Prefer that over invoking Cargo directly, because two toolchains are in play.
+`rust-toolchain.toml` pins a nightly that formatting, clippy, and miri use;
+everything that builds or runs code uses stable. The `just` recipes select the
+right one per command, so a bare `cargo nextest run` in this repository tests
+against a channel CI never tests against. To format, use:
+
+    ./scripts/run-just.sh fmt
+
```
rkuris, line 45: What effect does this have on most IDEs? Will they run the right toolchain at the right time, or are we deprecating IDE formatting? I often use vscode (insiders).

Action required: Answer the question and then resubmit. If the answer is we're deprecating IDE formatting, this should be a prominent note in the PR description. If it's fully supported, then why do you have to use these special just scripts?

README.md
```diff
@@ -117,6 +117,13 @@ In order to build firewood, the following dependencies must be installed:
 - `cargo` See [installation instructions](https://doc.rust-lang.org/cargo/getting-started/installation.html).
 - `make` See [download instructions](https://www.gnu.org/software/make/#download) or run `sudo apt install build-essential` on Linux.
 
+Firewood is built and tested on the stable Rust channel. The
```
rkuris, line 120: This PR exceeded it's time limit here.

Action required: Address all the other actions and resubmit. This particular comment is not actionable for you, but is actionable for me when I resume reviewing it. I search for this comment and change my review style accordingly.

You might be able to infer additional review comments based on the above ones as well. Many of them were repetitive.

rkuris (changes requested):

AGENTS.md
demosdemon: Yes, someone running `cargo fmt` inside `nix develop` gets stable rustfmt, and once nightly-only options land it will reformat differently. This is intentional because I don't really care about nix and don't intend to support it beyond the minimum needed to support avalanchego's hacky build. I also don't know of anyone that is using `nix develop`. If I did, I might consider otherwise. But as it is we're already planning on removing nix.

.github/workflows/lint-nightly.yaml
demosdemon: No. That SHA pins the `dtolnay/rust-toolchain` action itself, not a Rust version. `toolchain: nightly` floats, and the dated pin lives only in `rust-toolchain.toml`.

CONTRIBUTING.md
demosdemon: Fully supported and actually the point of the change. rust-analyzer and VS Code invoke `cargo fmt` and `cargo clippy` through `rustup` proxies, which read `rust-toolchain.toml`, so format-on-save and inline diagnostics use the pinned nightly automatically. The `just` recipes matter for build and test, which must be steered back to stable. The `fmt` recipe exists only for shells that export RUSTUP_TOOLCHAIN.

AGENTS.md
demosdemon: I opted to keep this in AGENTS. Primarily because code agents need these details the most. They typically won't read the README unless instructed.

PR #2233, opened 2026-09-01
style: enable the nightly rustfmt options that do not touch imports

demosdemon:
## Why this should be merged

This is the payload the toolchain pin in #2232 exists for. The options that rewrite imports (`group_imports`, `imports_granularity`) account for most of the churn and land separately, so this PR carries the remainder: the ones whose effect is local and easy to eyeball.

## How this works

`rustfmt.toml` is new and enables `normalize_comments`, `normalize_doc_attributes`, `format_macro_matchers`, and `use_field_init_shorthand`. The first three are unstable, so they apply only on the nightly that `rust-toolchain.toml` pins. Stable rustfmt warns once per option and skips it; because each defaults to `Preserve` or `false`, stable can never undo this formatting, only fail to apply it to new code, which `just ci-format` catches. The file links to the rustfmt option reference rather than paraphrasing it.

Two candidates were evaluated and left out:

- `overflow_delimited_expr` was enabled in an earlier revision of this PR and accounted for roughly 700 changed lines. On review the result read worse than the default layout: it trades one indentation level for calls whose closing delimiters stack at the end of a long argument list. Dropping it returns every affected call to its prior shape.
- `use_try_shorthand` cannot fire on edition 2024, where `try!` does not exist. A guard against an impossible style is noise.

`normalize_doc_attributes` and `use_field_init_shorthand` change nothing today and are kept as drift guards.

Two places needed a decision rather than a mechanical apply:

`normalize_comments` is bidirectional. Besides dropping empty `///` lines and inserting the space in `//Text`, it rewrites a trailing `//` comment as `/* */` when it shares a line with code that continues past it. It did this to a `// TODO` in `firewood/src/iter.rs`, which reads worse. Moving that comment above the match arm removes the trigger and keeps the option's other effects.

`format_macro_matchers` collapses a multi-line matcher onto one line but omits the space after a comma between named fragments, producing `$hex16:expr,ethereum`. It did this to both arms of `expected_hash!` in `storage/src/tries/kvp.rs`, whose matchers deliberately mirror the multi-line calling convention that macro's own doc comment shows. That macro carries `#[rustfmt::skip]` and a note explaining why. The option is kept because its remaining effect, collapsing `descend!` in `storage/src/tries/iter.rs`, is an improvement and formats correctly.

## How this was tested

- `just fmt` produced the diff; no hunk was hand-edited except the two cases described above, both of which change the input rather than the output.
- `just ci-format` is clean afterwards, so formatting is idempotent and the `#[rustfmt::skip]` holds.
- `just prepush-lite` passed on the revision that still had `overflow_delimited_expr`; removing it is a whitespace-only revert to the layout on the base branch.
- Confirmed the diff introduces no `/* */` comments and no matcher missing a space after a comma.

## Breaking Changes

None. Formatting only; no semantic change to any crate.


firewood/src/merkle/tests/change/edge_cases.rs
```diff
@@ -873,10 +868,9 @@ fn test_truncation_classification_under_a_prefix_key(
             },
         ]
     } else {
-        vec![
-            BatchOp::Delete { key: &b"\x10"[..] },
-            BatchOp::Delete { key: &b"\x20"[..] },
-        ]
+        vec![BatchOp::Delete { key: &b"\x10"[..] }, BatchOp::Delete {
+            key: &b"\x20"[..],
+        }]
```
rkuris, line 873: nit: IMO this is much harder to read

The one later is worse.

No action is strictly required, but adding a #[rustfmt::skip] to this statement and reverting is a good idea if you're going to resubmit anyway.

firewood/src/merkle/tests/change/regression.rs
```diff
@@ -509,16 +489,11 @@ fn test_all_deleted_range_with_survivor_above_end_bound_verifies() {
...
-                key: b"\xe0",
-                value: b"\x01",
-            },
-        ],
-    );
+    let (start_root, end_root) =
+        commit_batch(&db, vec![BatchOp::Delete { key: b"\x20" }, BatchOp::Put {
+            key: b"\xe0",
+            value: b"\x01",
+        }]);
```
rkuris, line 496: This one is even worse. A causal reader can't see this contains two `BatchOp`s, a Delete and a Put, significantly lowering readability.

Fix: Revert this diff, and add #[rustfmt::skip]

firewood/src/merkle/changes.rs
```diff
@@ -1137,7 +1129,7 @@ mod tests {
             (m1.nodestore(), m2.nodestore(), "m1->m2")
         };
 
-        //let diff_stream = DiffMerkleKeyValueStreams::new(tree_left, tree_right, Key::default());
+        // let diff_stream = DiffMerkleKeyValueStreams::new(tree_left, tree_right, Key::default());
```
rkuris, line 1132: nit: not really sure why we have commented out code here; different issue.

No action required for approval, just noting this in case you want to roll this in or open another ticket or ignore it.

firewood/src/iter.rs
```diff
@@ -947,18 +947,16 @@ mod tests {
...
         // Covers case of leaf with no partial path
         let (key, node) = iter.next().unwrap().unwrap();
         assert_eq!(key, vec![0x00, 0xFF].into_boxed_slice());
-        assert_eq!(
-            node.as_leaf().unwrap().clone().value.to_vec(),
-            vec![0x00, 0xFF]
-        );
+        assert_eq!(node.as_leaf().unwrap().clone().value.to_vec(), vec![
+            0x00, 0xFF
+        ]);
```
rkuris, line 959: nit: I also find this less readable.

Again, no action is required. It's only slightly less readable, but a `#[rustfmt::skip]`isn't a terrible idea here either.

firewood/src/iter.rs
```diff
@@ -1187,11 +1183,11 @@ mod tests {
...
-        let key_values = [
-            vec![first_key],
-            vec![intermediate, intermediate],
-            vec![intermediate, intermediate, intermediate],
-        ];
+        let key_values = [vec![first_key], vec![intermediate, intermediate], vec![
+            intermediate,
+            intermediate,
+            intermediate,
+        ]];
```
rkuris, line 1190: This one is pretty bad. The reader no longer realizes there are 3 vecs here.

Action: Add a #[rustfmt::skip] and revert this diff.

storage/src/checker/range_set.rs
```diff
@@ -381,10 +381,10 @@ mod test_range_set {
...
         range_set.insert_range(0..10);
         range_set.insert_range(20..30);
-        assert_eq!(
-            range_set.into_iter().collect::<Vec<_>>(),
-            vec![0..10, 20..30]
-        );
+        assert_eq!(range_set.into_iter().collect::<Vec<_>>(), vec![
+            0..10,
+            20..30
+        ]);
```
rkuris, line 387: Same issue as previous. I can't look at this and see what's being compared in the assert_eq macro without doing a lot of reading here.

Action: #[rustfmt::skip] and revert.

I think another option here is to add a comment to each block; I have noticed rustfmt seems to behave a bit different in this case.

storage/src/checker/range_set.rs
```diff
@@ -431,10 +431,11 @@ mod test_range_set {
...
         assert!(range_set.insert_disjoint_range(12..18).is_ok());
-        assert_eq!(
-            range_set.into_iter().collect::<Vec<_>>(),
-            vec![0..10, 12..18, 20..30]
-        );
+        assert_eq!(range_set.into_iter().collect::<Vec<_>>(), vec![
+            0..10,
+            12..18,
+            20..30
+        ]);
```
rkuris, line 438: Second instance of this same file. Future instances are not called out.

Action: Search for any more instances where the test readability suffers because the reader can't easily see what's being compared any more, and address the remaining ones.

Note that all remaining blocks that suffer from this problem should be addressed before resubmitting for review.

rustfmt.toml
```diff
@@ -0,0 +1,33 @@
+# Most options here are unstable, so they apply only on the nightly channel.
+# `rust-toolchain.toml` pins that nightly and `just fmt` uses it; formatting
+# with stable rustfmt prints a warning per unstable option and silently skips
+# it, which leaves changes `just ci-format` then reports. Only
+# `use_field_init_shorthand` and `use_try_shorthand` are stable and apply
+# everywhere.
```
rkuris, line 6: This is overly verbose, try:

```suggestion
# Rust formatting rules, requires the nightly toolchain
```

If you want to comment each one that requires nightly, you can do that inline, but I think that's a bad idea because as these options become stabilized it will increase churn or leave us with a bad comment.

Action required: Accept the suggestion or explain why the more verbose one makes sense.
Optional, not required for re-review: Add comments on which ones are stable and which ones are not.

rustfmt.toml
```diff
@@ -0,0 +1,33 @@
...
+#
+# Each option below defaults to `Preserve` or `false`, so stable rustfmt never
+# undoes what nightly did here -- it only fails to apply it to new code.
```
rkuris, line 9: Tautological. I don't think we should be explaining how rustfmt works in the code here.

Action: Remove this or explain why you think this comment makes sense.

rustfmt.toml
```diff
@@ -0,0 +1,33 @@
...
+
+# Wrap the arguments of a function call whose last argument is a delimited
+# expression, rather than indenting the whole call one level deeper.
```
rkuris, line 12: nit: do we need verbose descriptions of each item? Seems like we don't; someone can easily look these up if they want to see what they do, and in theory these could slightly change or the explanations could be incorrect.

No action required, but recommend removing the descriptive text. Appears mostly AI generated.

firewood/src/merkle/tests/change/edge_cases.rs
demosdemon: I opted to remove the `overflow_delimited_expr` option since it generated to many instances worse than before.

PR #2235, opened 2026-09-01
style: group imports as std, then external crates, then crate

demosdemon:
## Why this should be merged

Import grouping was the original motivation for issue #1507: the repository had no convention, so each file carried whatever grouping it happened to be written with. `group_imports = "StdExternalCrate"` settles it -- `std`/`core`/ `alloc` first, then external crates, then `crate`/`self`/`super`, one blank line between groups.

This lands separately from the other formatting options because it rewrites import blocks in 95 files, and reviewing that alongside unrelated formatting would hide both.

## How this works

One line added to `rustfmt.toml`; the rest is `just fmt` output, committed without hand edits.

Reordering `use` statements is semantically inert in Rust: imports do not shadow by position, trait method resolution does not depend on import order, and macros imported through `use` are order-independent, unlike the old `#[macro_use] extern crate` form, which this workspace does not use.

The case worth checking is conditional imports, and rustfmt keeps each attribute bound to its own statement. `storage/src/nodestore/persist.rs` is the clearest example: a `#[cfg(test)] use crate::RootReader` and a `#[cfg(not(test))] use super::RootReader` sort into different groups and stay correctly paired.

## How this was tested

  - `just fmt` produced the diff; `just ci-format` is clean afterwards, so the result is idempotent.
  - Clippy passes on all five profiles -- `debug-no-default-features`, `debug-no-features`, `debug-ethhash-logger`, `debug-all-features`, and `maxperf-ethhash-logger`. The full matrix rather than the usual single profile, because an import reshuffle is exactly the change that breaks a feature combination the light gate does not build.
  - `ci-docs`, `ci-lint-markdown`, `ci-check-go-generate`, and `ci-machete` pass. `go generate` leaves tracked files untouched.
  - Import statement count is identical before and after, confirming the change reorders and never drops or merges. Merging is `imports_granularity` and comes separately.

## Breaking Changes

None. Formatting only; no semantic change to any crate.


rkuris (approved):
Mechanical change, LGTM.

Note: no action required, just marking this PR approved and it's only a mechanical change.

PR #2236, opened 2026-09-01
style: split grouped imports into one `use` per item

### LINKED ISSUE #1507: Look into custom `rustfmt` configuration

demosdemon:
## Why this should be merged

`imports_granularity = "Item"` is the last of the formatting options behind issue #1507, and the largest. It gives every import its own line, so adding or removing one is a one-line diff and two branches touching different items of the same module no longer conflict in the same brace list. The cost is verbosity: imports go from 1,075 statements to 2,320.

Closes #1507

## How this works

One line added to `rustfmt.toml`; the reformat is `just fmt` output. Two follow-on source changes were required, neither cosmetic.

Splitting a grouped import copies the group's attributes onto every item it produces. In `storage/src/test_utils.rs`, `use rand::{Rng, SeedableRng, TryRng}` carried a single `#[expect(clippy::disallowed_types)]`, which became three. Only `SeedableRng` is in `clippy.toml`'s `disallowed-types`, so the other two expectations went unfulfilled -- an error under `-D warnings`, and the reason clippy failed on all five profiles before this was addressed.

That failure is the option working as intended. The grouped attribute had been suppressing `disallowed_types` for `Rng` and `TryRng`, which never needed it, and nothing could detect that while the three shared one line. Splitting forced the suppression down to the single import that warrants it, and `#[expect]` refused to compile until it was narrowed. An `#[allow]` would have let the over-broad suppression survive silently.

Second, `Item` granularity writes `use std::io::{self};` when it splits a group that also imported the module itself, in 30 places across 6 modules. The bare `use std::io;` is idiomatic and, confirmed by reformatting afterwards, is left alone by rustfmt -- so all 30 were rewritten rather than committing a form this change would otherwise introduce.

## How this was tested

  - `just fmt` produced the reformat; `just ci-format` is clean afterwards.
  - Clippy passes on all five profiles, and tests pass on all five: 791, 791, 836, 839, and 836 tests.
  - All six FFI steps pass -- build, test, and the compatibility suite, for both hash modes -- plus `ci-docs`, `ci-check-go-generate`, `ci-machete`, `ci-lint-markdown`, and `ci-check-todos`.
  - The set of `pub use` re-exports across every `lib.rs` is unchanged: 143 symbols before and after, none lost or gained. Splitting re-exports is the one way this change could silently alter the public API, so the sets were compared directly rather than assumed equal.

## Breaking Changes

None. Formatting only, and the public API is byte-for-byte the same set.


rkuris (approved):
Mechanical change, LTGM

Note: No action required, just noting this is primarily a mechanical change and marking the PR approved.

PR #2240, opened 2026-09-02
docs: require comments to earn their line, and fix stale references

demosdemon:
## Why this should be merged

Comments here increasingly read as generated rather than written — padded, repeated at every call site, and narrating the change that introduced them ("now uses", "fixed to handle") instead of describing the code as it stands. That last habit is the expensive one: the narration is already stale by the time the next reader arrives, and nothing flags it. No linter can catch any of this, so review is the only gate — which means both the authoring rules and the reviewer's mandate to enforce them have to be written down.

Reading the three governance docs closely enough to amend them surfaced a second problem: several of their references no longer resolve. Those corrections ride along rather than waiting for a separate pass.

## How this works

The authoring rules live in `CONTRIBUTING.md`. The reviewer-facing half extends `CODE_REVIEW.md`'s existing **Comments and messages** bullet instead of adding a parallel one, so the checklist does not accumulate the duplication the new rules prohibit.

The `AGENTS.md` note is deliberately redundant with `CONTRIBUTING.md` rather than a bare link. Agents act on that notes list directly and frequently without following links, so the failure modes are named where they will actually be seen. This is the one place the new "say it once and cross-reference" rule is knowingly broken, and the trade is intentional.

## How this was tested

`just ci-lint-markdown` is clean.

## Breaking Changes

None. Documentation change only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,6 +41,8 @@ firewood/             # Core library and main database implementation
 └── benches/          # Benchmarks
 
 firewood-macros/      # Procedural macros for the project
+metrics/              # Shared metrics utilities
+replay/               # Replay log types and engine for database operations
 storage/              # Storage layer implementation
 triehash/             # Trie hashing functionality
 ffi/                  # Foreign Function Interface (FFI) binding for Golang
@@ -78,9 +80,10 @@ By default, Firewood uses SHA-256 hashing compatible with merkledb. Enable this 
   storage-child count
 - See `storage/src/hashers/ethhash.rs` for implementation details
 
-### `logging`
+### `logger`
 
-Enable for runtime logging. Set `RUST_LOG` environment variable accordingly (uses `env_logger`).
+Enable for runtime logging. Set the `RUST_LOG` environment variable accordingly
+(uses `env_logger`).
 
 ## Common Development Tasks
 
@@ -90,17 +93,21 @@ Building and using the FFI library is a multi-step process. To generate the
 Firewood Rust FFI bindings:
 
 ```bash
-cd ffi/src                                              # Go to Rust binding directory
+cd ffi                                                  # Go to the FFI crate root
 cargo clean                                             # Remove any existing bindings
 cargo build --profile maxperf --features ethhash,logger # Generate bindings
 ```
 
-To then have Golang utilize these new bindings:
+To then have Golang utilize these new bindings, from that same directory:
 
 ```bash
-cd ..                   # Go to ffi directory
 go tool cgo firewood.go # Generate cgo wrappers
 ```
+
+`go tool cgo` refreshes only the wrappers the Go linter reads. The `go:generate`
+directives in `ffi/firewood.go` additionally update the cgo pragmas and
+`LDFLAGS`, and CI checks that they leave the tree unchanged. Run
+`just ci-check-go-generate` when link flags or pragmas drift.
 
 Any tagged enums added to the FFI api where the union body contains a pointer must be defined as `#[repr(C, usize)]` so that the enum tag forces the C struct to have pointer alignment.
 
@@ -274,7 +281,9 @@ brew install markdownlint-cli2
 
 Key dependencies are centrally managed in workspace `Cargo.toml`:
 
-- `firewood`, `firewood-macros`, `firewood-storage`, `firewood-ffi`, `firewood-triehash` (workspace members)
+- Workspace members: `firewood`, `firewood-storage`, `firewood-triehash`,
+  `firewood-macros`, `firewood-metrics`, `firewood-ffi`, `firewood-fwdctl`,
+  `firewood-benchmark`, `firewood-replay`
 - Common deps: `clap`, `thiserror`, `smallvec`, `sha2`, `log`, etc.
 - Test deps: `criterion`, `tempfile`, `rand`, etc.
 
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -311,7 +320,19 @@ Key dependencies are centrally managed in workspace `Cargo.toml`:
...
+   at every call site, and narrating the change that produced them ("now uses",
+   "previously this would", "fixed to handle"). Write comments that stand alone
+   for a reader years from now who has no diff and no review thread: present
+   tense or imperative, consistent terminology, no filler or hype, each piece of
+   context stated once and cross-referenced from everywhere else. The same bar
+   applies to user-facing strings — error messages, log lines, and CLI help. See
+   [Comments and documentation](./CONTRIBUTING.md#comments-and-documentation)
+   and [User-facing strings](./CONTRIBUTING.md#user-facing-strings).
+
+8. **Workspace Awareness**: This is a multi-crate workspace. Changes may affect multiple crates. Check `Cargo.toml` for workspace structure.
```
rkuris: Why do we need this? Seems like ci will fail if someone doesn't notice a change affected some other crate within the repository, or at least should. And even so, if we change an interface that affects another one of our crates, it's possible it might affect a crate from another author somewhere else (breaking change). I'd remove this.

CODE_REVIEW.md
```diff
@@ -16,9 +16,17 @@ providing inline feedback — in addition to general best practices.
...
+  `#[allow(clippy::unwrap_used)]` and `#[expect(clippy::unwrap_used)]` there. Judge test
+  context by location, not by `cfg(test)` alone: unit-test modules and integration tests
+  under `tests/` are exempt, and so are Criterion benches under `benches/`, which are
+  `harness = false` binaries that never compile with `cfg(test)` set.
+- **`expect()`**: The sanctioned alternative to `unwrap` (see
+  [Style Guide](./CONTRIBUTING.md#style-guide--coding-conventions)), and clippy does not
+  enforce `expect_used`, so review is the only gate. The message must state the invariant
+  that makes the panic unreachable — why reaching it would be a bug — rather than name the
+  operation that failed. Reject `.expect("valid")` and `.expect("should not fail")`. Where
+  the caller can act on the failure, returning a `Result` beats any message.
```
rkuris: This is excessively wordy. How about:
```suggestion
- `unwrap`, is prohibited in production code, primarily because we don't want code that panics in production. We allow expect for the few cases that don't have a good way of rewriting the code so it doesn't panic. Always include a reason why this should never happen as an argument. See https://www.ncameron.org/blog/to-panic-or-not-to-panic/ for further discussion of this.
```

CODE_REVIEW.md
```diff
@@ -91,7 +101,11 @@ For lints we leave suppressible:
 
 - **Comments and messages**: Every comment, log message, and error string must be
   accurate, useful, and free of tautologies. Log and error messages must be unique enough
-  to locate their source quickly.
+  to locate their source quickly. Challenge text that reads as generated rather than
+  written — padded, hedged, repeated at every call site, or narrating the change that
+  introduced it. See
+  [Comments and documentation](./CONTRIBUTING.md#comments-and-documentation) and
+  [User-facing strings](./CONTRIBUTING.md#user-facing-strings) for the authoring rules.
```
rkuris: nit: I'm challenging this text that reads as generated. I'm not sure this actually is useful here, but willing to accept it if you think it's warranted.

CONTRIBUTING.md
```diff
@@ -17,20 +17,30 @@ guidelines for contributing to firewood.
...
 ## Testing
 
-After submitting a PR, we'll run all the tests and verify your code meets our submission guidelines. To ensure it's more likely to pass these checks, you should run the following commands locally:
+After submitting a PR, we'll run all the tests and verify your code meets our submission guidelines. To ensure it's more likely to pass these checks, run the same recipes CI runs:
 
     cargo fmt
-    cargo nextest run
-    cargo clippy
-    cargo doc --no-deps
+    just prepush-lite
```
rkuris: This is actually incorrect. If you want *everything* CI runs, you have to use `just prepush`.

Action: Either use `just prepush` or explain that `prepush-lite` catches most but not all CI failures.

CONTRIBUTING.md
```diff
@@ -17,20 +17,30 @@ guidelines for contributing to firewood.
...
+`cargo fmt` applies formatting; the recipes only check it. `prepush-lite` checks
+formatting and TODOs, runs clippy and the tests for one feature profile, and
+lints the Markdown and the documentation build.
+
+CI is authoritative for the full matrix, and a change that touches the FFI,
+workspace dependencies, or a feature that profile does not enable warrants the
+whole thing before you push:
+
+    just prepush
+
```
rkuris: This corrects the previous incorrect information. Someone who stops reading above will incorrectly assume `just prepush-lite` is all they need.

Action: Please reword this so you're discussing how to choose between prepush and prepush-lite first.

AGENTS.md
```diff
@@ -90,18 +93,22 @@ Building and using the FFI library is a multi-step process. To generate the
...
 
 ```bash
-cd ..                   # Go to ffi directory
 go tool cgo firewood.go # Generate cgo wrappers
 ```
 
+`go tool cgo` refreshes only the wrappers the Go linter reads. The `go:generate`
+directives in `ffi/firewood.go` additionally update the cgo pragmas and
+`LDFLAGS`, and CI checks that they leave the tree unchanged. Run
+`just ci-check-go-generate` when link flags or pragmas drift.
```
rkuris: nit: Would prefer to say what actually works rather than what doesn't. `just ci-check-go-generate` works. If you want to quickly mention not to use `go tool cgo` that might be okay too but a full explanation as to why it doesn't work seems excessive.

Action: none required for approval (this comment by itself is not actionable), but this could be cleaned up.

RodrigoVillar (approved):

PR #2251, opened 2026-09-14
Jl/snoopy linus setup

JuanLeon2:
## Breaking Changes

- [ ] firewood
- [ ] firewood-storage
- [ ] firewood-ffi (C api)
- [ ] firewood-go (Go api)
- [ ] fwdctl


benchmark/README.md
```diff
@@ -14,6 +14,9 @@ Firewood has two categories of benchmarks:
 | Rust criterion | Iterating on a specific operation locally; enforces pure-Rust API in CI | Local or CI (`benchmarks.yaml`) |
 | Synthetic workloads | Testing Firewood API patterns without AvalancheGo | Local |
 
+Benchmarks that need bare metal rather than CI or EC2 can run on the on-prem
+servers; see [`infra/onprem/`](../infra/onprem/README.md).
```
rkuris, line 18: This is in a public repo and these resources aren't available for outside contributors. We should make this clear.

infra/onprem/add-ssh-key.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
```
rkuris, line 1: Can't we just use the pre-published list of ssh-keys?

infra/onprem/add-user.sh
```diff
@@ -0,0 +1,357 @@
+#!/bin/bash
```
rkuris, line 1: Similar comment: since we already have a valid list of team members, why can't we just use that?

rkuris (commented):
This feels extremely duplicative for what we have in benchmark/setup-scripts. Can we reuse those, or perhaps modify the benchmark setups to use these scripts if you think they are better?

benchmark/README.md
rkuris: I see we do later, but this is a separate instance.

infra/onprem/add-ssh-key.sh
JuanLeon2: We do.  But the keys need to be added to each user's authorized_keys on snoopy/linus, by snoopy/linus local user root.  This script does that.  It also gives the option of the keys coming from a file like that written by ssh-keygen.  We could remove that option if that is what you are suggesting. add-user.sh, which invokes this, already gives the option of getting the key from the benchmark list of keys.

infra/onprem/add-user.sh
JuanLeon2: launch-stages.yaml is embedded into fwctl at compile time.  In part because of that, the change required to extract the list of users into a more "neutral" location/file is a bit intrusive.  I thought it was too much for this one PR.  

I have stacked a PR (#2252) for that if you think is worth it.  IN any case, this script adds a single user.  The list of users is not assumed to exist in this PR, other than as documentation in SETUP.md.

benchmark/README.md
JuanLeon2: Good point. Thanks.

infra/toolchains/firewood-toolchain.sh
```diff
@@ -0,0 +1,43 @@
...
+FIREWOOD_GO_TOOLS=(
+	# avalanchego drives its build through a root Taskfile.yml, so anyone
+	# working on the FFI against a local avalanchego needs this.
+	github.com/go-task/task/v3/cmd/task@v3.53.1
+	github.com/reteps/dockerfmt@v0.5.4
+	# Moves C-Chain blocks and state snapshots to and from S3. The benchmark
+	# workflow depends on it; .github/workflows/ci.yaml pins the same version.
+	github.com/peak/s5cmd/v2@v2.3.0
+	mvdan.cc/sh/v3/cmd/shfmt@v3.14.1
+)
```
demosdemon, line 43: lets use mise en place for this https://mise.jdx.dev/getting-started.html

infra/onprem/add-user.sh
JuanLeon2: I have now also stacked PR #2263 to also share the list of packages needed in these firewood test hosts, if the impetus to have a single configuration point goes that far-- though I have kept out of .devcontainer which should also share teh configuration info.

JuanLeon2:
Closing: this has been split in two, and neither half is this branch.

The part firewood wants — pinned, checksummed Rust and Go toolchains for the
benchmark hosts, replacing an `install-golang.sh` that installed whatever
release was newest that day and a `curl | sh` rustup — is #2264, rebased and
trimmed of the on-prem conventions that came with it here.

The on-prem provisioning itself (`infra/onprem/`: Snoopy, Linus and the LXD
session containers) has moved to a separate internal repository. It is not
about the database, and there is no reason for firewood's reviewers to carry
it.

PR #2252, opened 2026-09-14
chore(infra): share Firewood user manifest

JuanLeon2:
## Why this should be merged

The benchmark launch config and on-prem provisioning need the same team/user data, but the canonical list previously lived inside benchmark configuration. That made on-prem helpers reach across into benchmark files and left account reconciliation manual.

## How this works

- Moves shared user data into infra/users/firewood-users.yaml.
- Keeps benchmark-specific account fields in that manifest and teaches fwdctl to render embedded benchmark users from it.
- Adds infra/onprem/add-users.sh to loop over local users and call add-user.sh, then install interim SSH public keys.
- Points add-ssh-key.sh at the shared manifest instead of benchmark/launch/launch-stages.yaml.

## How this was tested

- bash -n infra/onprem/add-users.sh infra/onprem/add-ssh-key.sh
- shellcheck -x infra/onprem/add-users.sh infra/onprem/add-ssh-key.sh
- /opt/homebrew/bin/shfmt -d infra/onprem/add-users.sh infra/onprem/add-ssh-key.sh
- /opt/homebrew/bin/markdownlint-cli2 infra/onprem/SETUP.md fwdctl/README.launch.md
- ruby YAML parse for benchmark/launch/launch-stages.yaml and infra/users/firewood-users.yaml
- nix shell nixpkgs#cargo nixpkgs#rustc nixpkgs#rustfmt -c cargo fmt --check
- nix shell nixpkgs#cargo nixpkgs#rustc nixpkgs#rustfmt -c cargo test -p firewood-fwdctl --features launch launch::stage_config::tests::embedded_config_parses


infra/users/firewood-users.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICoGgX8nCin3FPc1V3YYN1M9g039wMbzZSAXZJCqzBt3 cardno:31_786_961 austin"
+  - local_user: aaron
+    benchmark_user: aaron
+    full_name: Aaron
+    benchmark_groups: "users, adm, sudo"
+    benchmark_shell: /usr/bin/bash
+    benchmark_sudo: "ALL=(ALL) NOPASSWD:ALL"
+    lock_passwd: true
+    ssh_authorized_keys:
+      - "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIMj2j6ySwsFx7Y6FW2UXlkjCZfFDQKHWh0GTBjkK9ruV cardno:19_236_959 aaron"
```
demosdemon, line 28: please delete

benchmark/launch/launch-stages.yaml
```diff
@@ -10,57 +12,6 @@ packages:
...
-    lock_pass: true
-    ssh_authorized_keys:
-      - "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICoGgX8nCin3FPc1V3YYN1M9g039wMbzZSAXZJCqzBt3 cardno:31_786_961 austin"
-  - name: aaron
-    groups: "users, adm, sudo"
-    shell: /usr/bin/bash
-    sudo: "ALL=(ALL) NOPASSWD:ALL"
-    lock_pass: true
-    ssh_authorized_keys:
-      - "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIMj2j6ySwsFx7Y6FW2UXlkjCZfFDQKHWh0GTBjkK9ruV cardno:19_236_959 aaron"
```
demosdemon, line 34: we should audit these more often ;)

infra/onprem/add-ssh-key.sh
```diff
@@ -14,162 +14,195 @@ set -o nounset
...
-LAUNCH_STAGES="$SCRIPT_DIR/../../benchmark/launch/launch-stages.yaml"
+USER_MANIFEST="$SCRIPT_DIR/../users/firewood-users.yaml"
 
 DRY_RUN=0
 USERNAME=""
 KEY_FILE=""
 LAUNCH_USER=""
 
 show_usage() {
-    echo "Usage: $0 [OPTIONS] USERNAME [KEYFILE]"
```
demosdemon, line 25: your formatter replaced all of the indention with hard tabs; please revert back to spaces

JuanLeon2:
Closing: the reason for this change no longer exists.

It moved the user list out of `launch-stages.yaml` so the on-prem session hosts
and the EC2 benchmark hosts could share one manifest. The on-prem side has
since moved to a separate internal repository, so there is no second consumer
to share with.

Without one, this is a net loss for firewood: it deletes a plain inline `users:`
block and adds ~60 lines of Rust — `UserManifest`, `SharedUserDefinition`,
`embedded_default()` and a second `include_str!` — to read the same list back
from another file, through a filter whose only job was dropping on-prem-only
accounts. `launch-stages.yaml` keeps its inline list.

PR #2263, opened 2026-09-18
feat(infra): share package lists between benchmark hosts and sessions

JuanLeon2:
> Stacked on #2252 — review that one first. This PR targets `jl/shared-infra-users`,
> so its diff will shrink to just these commits once 2252 merges.

## Why this should be merged

The benchmark hosts and the on-prem session image kept separate package lists in
four places, which is how `task` and s5cmd came to be installed on one and
missing from the other. Rust, Go and the tool pins are already shared through
`firewood-toolchain.sh`, and users through `firewood-users.yaml` (#2252);
packages were the remaining copy.

## How this works

`infra/packages/firewood-packages.yaml` holds `common`, plus `session` and
`host` for what only one environment needs. It sits beside the user manifest and
is read the same way:

- `fwdctl` merges `common` + `host` into the cloud-init package list, exactly as
  it already does for users, so `launch-stages.yaml` loses its `packages:`
  block. `build-environment.sh` then installs nothing and is left with the disk
  setup it is actually for.
- `provision-session.sh` installs `common` + `session`.

The one bash consumer reads the manifest with PyYAML through
`infra/read-list.py`, not with a hand-written awk or sed parser. A YAML file
parsed by approximation behaves differently depending on who reads it. Ubuntu's
cloud images ship PyYAML with cloud-init (6.0.3 on snoopy), and the script
installs `python3-yaml` regardless so it does not depend on that.

`common` is what both environments already installed, not the union: the
benchmark hosts build the FFI without clang, cmake, pkgconf or libssl-dev, so
promoting one of those wants evidence rather than symmetry. The session set is
therefore unchanged at the same 18 packages, and the benchmark set changes only
by dropping `make`, which `build-essential` depends on.

Adds the `# Errors` doc that `embedded_default` was missing, since this extends
that function and it was the only clippy warning in the crate.

## How this was tested

`cargo test -p firewood-fwdctl --features launch` passes, including new
assertions that the benchmark list contains `common` and `host` entries and
excludes session-only ones — which also proves the manifest parses. clippy
clean, `cargo fmt --check` clean, shellcheck -x and markdownlint-cli2 clean.

Exercised the reader against the manifest for each list and for missing file,
missing key and missing argument. Reproduced the layout that `SETUP.md` pushes
into the build container and confirmed the script's relative paths resolve
there, returning the expected 18 packages.

Not yet exercised end to end on a real `fwdctl launch`; the session-image half
lands with the next image build.

## Breaking Changes

None to `firewood`, `firewood-storage`, `firewood-ffi`, `firewood-go` or
`fwdctl`'s public surface. `benchmark/setup-scripts/build-environment.sh` no
longer installs packages — it relies on the cloud-init list `fwdctl` builds, so
running it standalone on a host that never ran cloud-init would skip them.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JuanLeon2:
Closing, for the same reason as #2252, which this was stacked on.

It split the package list into `common`/`session`/`host` so the on-prem session
image and the benchmark hosts could share the parts they agreed on. With
on-prem moved to a separate internal repository, `session` has no reader here
and `common` + `host` is just the package list `launch-stages.yaml` already
had. `infra/read-list.py` had no firewood consumer at all — `fwdctl` parses
YAML through serde.

Worth noting one part that was not a pure refactor: this also removed the
`apt-get install` loop from `benchmark/setup-scripts/build-environment.sh` in
favour of cloud-init having run first, which made that script depend on the
launch path rather than standing on its own. That goes back with the rest.

PR #2264, opened 2026-09-21
build(benchmark): pin and verify the Rust and Go toolchains

JuanLeon2:
## Why this should be merged

`install-golang.sh` installs whatever Go release is newest on the day it runs,
scraped from `go.dev/dl/?mode=json`, with no integrity check. Two benchmark
hosts provisioned a week apart can therefore run different Go versions, and a
benchmark comparison across them is not comparing what it claims to. The launch
config installs Rust the same way, by piping `sh.rustup.rs` into `sh`.

## How this works

`infra/toolchains/firewood-toolchain.sh` holds the versions and checksums, and
is sourced rather than executed. It lives outside `benchmark/` because pinned
toolchains are not a benchmark concept -- any future provisioned environment
should read the same file.

- `install-rust.sh` (new) installs the pinned rustup, stable and nightly
  toolchains, verifying the rustup-init download against its SHA256.
- `install-golang.sh` keeps its existing arch/OS map and pins the version. It
  looks up a per-platform checksum from the manifest and verifies when one
  exists, warning and continuing when it does not, so adding a checksum for a
  new platform needs no change here.
- `launch-stages.yaml` runs `rust-install` after `clone-firewood`, because
  `install-rust.sh` now comes from the clone. This also drops the stale
  `aws-launch.sh: ... (runcmd lines NNN)` cross-references, whose line numbers
  no longer point anywhere useful.
- `build-firewood.sh` calls `install-rust.sh` instead of piping rustup into sh.

The manifest is also sourced from outside this repository, so its variable
names are an interface: renaming one, or dropping an entry because nothing here
reads it, breaks a consumer that is not visible from this checkout.
`FIREWOOD_CARGO_TOOLS` and `FIREWOOD_GO_TOOLS` have no in-repo consumer today,
and the file says as much, so a later reader does not delete them as dead. The
s5cmd entry is the exception with a counterpart here -- it is pinned
independently in `.github/workflows/ci.yaml`.

## How this was tested

`shellcheck` and `bash -n` clean on all four scripts; `markdownlint-cli2` clean
on the README. Verified the per-platform checksum lookup resolves to the pinned
value for linux/amd64 and falls through to the warning path for a platform with
no pin.

Not tested end to end: no benchmark host was provisioned from this branch.

## Breaking Changes

None to `firewood`, `firewood-storage`, `firewood-ffi`, `firewood-go` or
`fwdctl`. `install-golang.sh` now installs the pinned Go rather than the latest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


infra/toolchains/firewood-toolchain.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2034
```
rkuris: Why is this disabled?

If it's a false positive somewhere, silence the warning for that specific line by adding # shellcheck disable=SC2034 right above it. Don't do it across the whole file as it will hide errors.

infra/toolchains/firewood-toolchain.sh
```diff
@@ -0,0 +1,49 @@
...
+# Shared Firewood toolchain pins for provisioned Linux environments.
+#
+# Keep this file side-effect free: provisioning scripts source it so that Rust,
+# Go and developer-tool versions do not drift between environments.
+#
+# It is also sourced from outside this repository, so the variable names below
+# are an interface: renaming one, or dropping an entry because nothing here
+# reads it, breaks a consumer this checkout cannot show you.
+# FIREWOOD_CARGO_TOOLS and FIREWOOD_GO_TOOLS have no in-repo consumer today.
```
rkuris, line 10: Ah this is probably why. These lines should be annotated with the shell check override.

infra/toolchains/firewood-toolchain.sh
```diff
@@ -0,0 +1,49 @@
...
+# The s5cmd pin is the exception -- .github/workflows/ci.yaml pins it too, and
+# the two should agree.
+
+FIREWOOD_RUSTUP_VERSION=1.29.1
+FIREWOOD_RUSTUP_X86_64_LINUX_GNU_SHA256=dda7234360b7f578ca8b0ddcb80145646fa61a67c1720a5abc7051b35c9fcb71
+FIREWOOD_RUST_VERSION=1.94.1
+FIREWOOD_RUST_NIGHTLY=nightly-2026-09-13
```
rkuris: Unless you plan on updating this every day, I don't think we want this pinned :)

infra/toolchains/firewood-toolchain.sh
```diff
@@ -0,0 +1,49 @@
...
+
+FIREWOOD_GO_VERSION=1.26.0
```
rkuris: I am worried this might easily get out of sync with what we test with. Can we set this in one place?

infra/toolchains/firewood-toolchain.sh
```diff
@@ -0,0 +1,49 @@
...
+#
+# It is also sourced from outside this repository, so the variable names below
+# are an interface: renaming one, or dropping an entry because nothing here
+# reads it, breaks a consumer this checkout cannot show you.
+# FIREWOOD_CARGO_TOOLS and FIREWOOD_GO_TOOLS have no in-repo consumer today.
+# The s5cmd pin is the exception -- .github/workflows/ci.yaml pins it too, and
+# the two should agree.
+
+FIREWOOD_RUSTUP_VERSION=1.29.1
+FIREWOOD_RUSTUP_X86_64_LINUX_GNU_SHA256=dda7234360b7f578ca8b0ddcb80145646fa61a67c1720a5abc7051b35c9fcb71
```
rkuris: Nit: this feels like overkill for rustup. It would be very surprising if we had different versions of rustup doing different things here.

No action required unless you feel the same. This does raise maintenance costs slightly as someone now will have to update this or check it periodically.

benchmark/setup-scripts/install-rust.sh
```diff
@@ -0,0 +1,62 @@
...
+download_checked \
+    "https://static.rust-lang.org/rustup/archive/${FIREWOOD_RUSTUP_VERSION}/x86_64-unknown-linux-gnu/rustup-init" \
+    "$FIREWOOD_RUSTUP_X86_64_LINUX_GNU_SHA256" \
+    "$rustup_init"
+chmod 0755 "$rustup_init"
+"$rustup_init" -y --no-modify-path --profile default \
+    --default-toolchain "$FIREWOOD_RUST_VERSION"
+
+rustup component add llvm-tools rust-docs --toolchain "$FIREWOOD_RUST_VERSION"
+rustup toolchain install "$FIREWOOD_RUST_NIGHTLY" --profile minimal
```
rkuris: This is pretty niche, and I doubt nightly builds are kept all that long. See below.

rkuris (commented):
Definitely need to address the nightly pin and the shell check fix. Remainder is optional.

infra/toolchains/firewood-toolchain.sh
JuanLeon2: Since rustup does no verification of the rust it downloads, would not verifying rustup introduce a supply-chain attack vector (granted the rustup server would need to be compromised).  Updates to this -- rustup version and checksum -- can be very infrequent, indeed because rustup does very little and evolves very slowly.  OTOH we have minor supply-chain vulnerabilities elsewhere so is not worth to inisist on this.....

infra/toolchains/firewood-toolchain.sh
JuanLeon2: Yeah totally right thanks

infra/toolchains/firewood-toolchain.sh
JuanLeon2: This was stupid.

infra/toolchains/firewood-toolchain.sh
JuanLeon2: It was already out of sync!  Changed it to match and added to the verify-go-versions CI workflow to check that this matches go.mod.

infra/toolchains/firewood-toolchain.sh
JuanLeon2: Fixed. Thank you.

benchmark/setup-scripts/install-rust.sh
JuanLeon2: Moot since I got rid of nightlies.

